Issue #110, opened 2026-06-13
gda script create + script get: script-group tracer (load → save → read)


qinhaihong-red:
## Parent

#17 — PRD: gda domain command surface — Phase-1 headless increment.

## What to build

Establish the `script` command group, which acts on `.gd` (GDScript) script files on disk (headless, one-shot `godot --headless`). This is the script-group tracer, mirroring the `scene` (#18) and `node` (#53) tracers.

- `gda script create` creates a new script at a target `res://` path — from a built-in template or from supplied content — no-clobber, and returns the created path plus any `class_name`/`extends` it declares as typed JSON.
- `gda script get` reads a script's source back (with its `class_name`/`extends` metadata) as typed JSON, so a `create` is verifiable end-to-end (create → get returns the source).

It establishes **script-file addressing** (how a script is referenced within a project) and the load/save plumbing every other `script` command reuses. Same headless pipeline and command skeleton as the scene/node slices (ADR-0001/0002/0004/0005; the shared `command_class()` + `--schema` flag).

**Scope — GDScript only.** C# (`.cs`) is **deferred**: it requires the .NET build of Godot, whereas ADR-0003 targets the **standard** build (4.4+/4.6 baseline), and supporting it is a decision in its own right (class/base semantics differ from GDScript's leading `class_name`/`extends`). The `script` group is GDScript-only until a `.NET`-build target and a dedicated decision exist; a non-`.gd` path is refused as `invalid_path` rather than half-supported as opaque text. _(Narrowed from the original `.gd`/`.cs` wording during implementation review — the local engine is the standard build, which cannot compile C# at all; documented in the `script` section of `docs/command-catalog.md`.)_

## Acceptance criteria

- [x] `gda script create` creates a `.gd` at a given path from a template or `--content`, refuses to clobber an existing file, and returns the created path + class metadata as typed JSON.
- [x] `gda script get` returns a script's source and `class_name`/`extends` metadata as typed JSON, and round-trips a `create`.
- [x] Script-file addressing — how a script is referenced within the project — is defined and documented (`docs/command-catalog.md`).
- [x] Both ship `--json` and the ADR-0004 `--schema` hard gate via the shared command skeleton.
- [x] Failure modes (path collision on create, script not found on get, invalid target path / non-`.gd` extension) surface as structured `GdaError`s with registered codes (Python registry + ADR-0002 table + GDScript mirror).
[...]

Issue #64, opened 2026-06-11
node add drops sub-scene instance overrides on re-save (data loss)


qinhaihong-red:
## Parent

#53 — gda node add + node list (the node-group tracer). Found in the high-effort code review of PR #60; empirically reproduced on Godot 4.6.

## What to build

`gda node add` mutates a scene via load → `PackedScene.instantiate()` → edit → `PackedScene.pack()` → `ResourceSaver.save()`. The instantiate call uses the default `GEN_EDIT_STATE_DISABLED`, which does not attach instance-state to instanced sub-scenes. On re-pack, the engine cannot diff against the sub-scene baseline, so **property overrides on nodes inside a pre-existing instanced sub-scene are silently dropped** from the saved file. Any future mutating node command built on the same round-trip (#55, #56, #57) inherits the bug.

Make the mutate round-trip preserve a scene's existing instanced sub-scenes and their overrides. The likely fix is instantiating with `GEN_EDIT_STATE_INSTANCE` (available in the editor/tools build `gda` targets). Confirm the chosen edit-state also preserves editable-children and node-path property overrides, not just top-level instance properties.

### Reproduction (verified, Godot 4.6)

`parent.tscn` instances `child.tscn` (a `Node2D` with an inner `Sprite2D`) and overrides the inner node:

```
[node name="ChildInstance" parent="." instance=ExtResource("1_child")]
[node name="Inner" parent="ChildInstance" index="0"]
modulate = Color(1, 0, 0, 1)
```

After `gda node add res://parent.tscn --type Marker2D --name M`, the saved file no longer contains the `Inner` override block — the whole `[node name="Inner" ...] modulate=...` entry is gone, and the instance node gains a spurious `type="Node2D"`. The top-level instance property (e.g. `position` on `ChildInstance` itself) does survive; only overrides on nodes *inside* the instance are lost.

## Acceptance criteria

- [ ] After `node add` on a scene that instances a sub-scene with an overridden inner node, the saved file still contains that override (and still records the node as an instance, not a flattened copy).
- [ ] Editable-children additions/overrides in the original scene survive the round-trip.
- [ ] An e2e regression test (`@pytest.mark.e2e`) builds a parent scene instancing a child scene with an inner-node override, runs `node add`, and asserts the override and the instance reference both survive.
- [ ] The full suite stays green; existing flat-scene round-trip tests still pass.

## Blocked by

None - can start immediately.

## Related

[...]

Issue #53, opened 2026-06-10
gda node add + node list: node-group tracer (load → mutate → save)


qinhaihong-red:
## Parent

#17 — PRD: gda domain command surface.

## What to build

Establish the `node` command group, which acts on nodes **within a scene file** (load → locate → mutate → pack → save), so it stays headless. This is the node-group tracer, mirroring the `scene create` + `scene get` tracer (#18).

- `gda node add` loads a `.tscn`, adds a child node — by Godot type (e.g. `Node2D`), and by a script's `class_name` — under a given parent node path, names it, packs and saves.
- `gda node list` loads a `.tscn` and reports its structured node tree (nestable name/type/path), so an add is verifiable end-to-end (add → list shows the new node).

It establishes **node-path addressing** (how a node is referenced within a scene) and the load→locate→mutate→pack→save round-trip every other `node` command reuses. Same headless pipeline and command skeleton as the scene slice (ADR-0001/0002/0004/0005; the shared `command_class()` + `--schema` flag).

## Acceptance criteria

- [ ] `gda node add` adds a node of a given type (and, where given, a `class_name` script) under a given parent node path, names it, packs and saves; returns the created node's path/name/type as a typed JSON result.
- [ ] Node-path addressing — how a parent/target node is referenced within the scene — is defined and documented.
- [ ] `gda node list` reports the scene's node tree (name/type/path, nestable) as a typed JSON result and round-trips an add.
- [ ] Both ship `--json` and the ADR-0004 `--schema` hard gate via the shared command skeleton.
- [ ] Operation failures (scene not found, bad parent path, invalid type, name collision) surface as structured `GdaError`s with registered operation codes (ADR-0002); new codes added to the Python registry, the ADR-0002 table, and the GDScript mirror.
- [ ] Tests: S2 model/parser unit, S3 CLI-with-FakeRunner (success + failure modes), S1 e2e against real Godot using the project fixture.

**Implementation order within the slice:** `node add` first (establishes addressing + mutate/save), then `node list` (verifies it).

## Blocked by

None - can start immediately. (`scene create` (#18) already produces a target scene for tests.)

Issue #17, opened 2026-06-09
PRD: gda domain command surface — Phase-1 headless increment


qinhaihong-red:
## Parent

#1 — PRD: gda Phase-1 headless foundation. That PRD's "Out of Scope" deferred every domain operation beyond `gda info` to "the next vertical slices built on this foundation". This epic is that body of work.

## Problem Statement

`gda info` proved the full Phase-1 pipeline end-to-end (binary resolution → headless runner → sentinel contract → typed model → JSON → structured errors), but it is the only command that exists. An agent driving Godot cannot yet create a scene, add a node, edit a script, read a project setting, or run an export — i.e. it cannot actually *build a game*. The command surface designed in ADR-0005 is still prose; nothing on disk lets an agent act on a Godot domain object.

## Solution

Grow `gda` across the **Phase-1 headless** domain command groups (`scene`, `node`, `script`, `project`, `resource`, `export`, and the asset-file / static-analysis commands) catalogued in `docs/command-catalog.md`. Each command is fulfilled by a one-shot `godot --headless` process that loads a file from disk, reads or mutates it, and saves it — no running engine required.

Delivery is **incremental, one vertical slice at a time** (ADR-0000, ADR-0005). This epic does not commit to all commands at once: the catalog is the map, and each command becomes a commitment only when its slice is opened as a child issue. Every command ships **self-describing** (`--schema`) as a hard, no-exception merge gate (ADR-0004).

## User Stories

1. As an AI agent, I want `gda scene create`, so that I can start a new `.tscn` with a chosen root node type without a running editor.
2. As an AI agent, I want `gda scene get`, so that I can read a scene's structured node tree from its file on disk.
3. As an AI agent, I want `gda scene list`, so that I can discover the scenes in a project.
4. As an AI agent, I want `gda scene delete`, so that I can remove a scene file I no longer need.
5. As an AI agent, I want `gda scene get-exports`, so that I can learn which `@export` properties a scene's nodes expose before configuring them.
6. As an AI agent, I want `gda node add`, so that I can place a node (by built-in type or `class_name` script) into a scene file.
7. As an AI agent, I want `gda node remove`, so that I can delete a node from a scene.
8. As an AI agent, I want `gda node get`, so that I can read a single node's properties.
9. As an AI agent, I want `gda node list`, so that I can enumerate a scene's nodes, optionally filtered by type or group.
10. As an AI agent, I want `gda node set`, so that I can change a node property with correct type coercion.
[...]

PR #129, opened 2026-06-13
feat(script): script set + script validate + script attach — edit, compile-check, bind (#118)

### LINKED ISSUE #118: gda script set + script validate + script attach: edit, compile-check, bind
## Parent

#17 — PRD: gda domain command surface — Phase-1 headless increment.

## What to build

The write/verify/bind operations on scripts, reusing the script-file plumbing from the script-group tracer (#110).

- `gda script set` edits a script via **search-replace**, **line-range**, or **full replacement**; verifiable via `script get`.
- `gda script validate` syntax/compile-checks a script and reports diagnostics as typed JSON.
- `gda script attach` binds a script to a node within a scene file, reusing the node-path addressing established in #53 (and the scene load/save round-trip).

## Acceptance criteria

- [ ] `gda script set` edits a script via search-replace / line-range / full replacement and round-trips via `script get`.
- [ ] `gda script validate` reports pass/fail with structured diagnostics (line/column/message) as typed JSON.
- [ ] `gda script attach` binds a script to a node by node path in a scene and persists the scene; verifiable by reading the node back.
- [ ] All ship `--json` and the ADR-0004 `--schema` hard gate via the shared skeleton.
- [ ] Failure modes (bad edit range, no search-replace match, compile errors surfaced structurally, node/script not found on attach) surface as structured `GdaError`s with registered codes (Python registry + ADR-0002 table + GDScript mirror).

## Blocked by

- #110 (script-group tracer: script-file addressing + load/save). (`attach` also reuses node-path addressing from the already-merged #53.)

qinhaihong-red:
## What

Closes #118 (parent PRD #17). Rounds out the `script` group's write/verify/bind operations, reusing the script-file plumbing from #110 and the node-path addressing from #53.

- **`gda script set`** — edit an existing `.gd` as **raw text** (never compiles/loads it) via one of three mutually-exclusive modes:
  - search-replace: `--search <old> --replace <new>` (literal, replace-all)
  - line-range: `--start-line N [--end-line M] --content <text>` (1-based, inclusive; `M` defaults to `N`)
  - full: `--content <text>` (overwrite)
  CLI enforces exactly one mode (usage error/exit 2 otherwise). Re-parses the written source's `class_name`/`extends` so an edit round-trips through `script get`. Edits only an existing script — a missing target is `path_not_found`, never a silent create.
- **`gda script attach SCENE --node <path> --script <gd>`** — bind a `.gd` to a node. Reuses the #53 node-path addressing (`.` = root, `A/B` = descendant) and the shared mutate-entry (`_load_for_mutation`, so it honors the #64 mutation-integrity boundary like `node set`): load → resolve node → `set_script` → re-pack → save. Echoes the attached script's `class_name`.
- **`gda script validate PATH`** — syntax/compile-check. Mechanism: read text → `GDScript.new()` + `source_code` + `reload()` (`OK` ⇒ compiles). Compiles but never **instantiates**, so it doesn't run instance code. A **`valid=false` result is a successful op** (exit 0) — it only op-fails for `invalid_path`/`path_not_found`.

## Design notes

- **Error codes (public ABI):** mints two focused operation codes — `no_search_match`, `invalid_line_range` — synced across all three mirrors (Python registry, ADR-0002 table, GDScript `OP_ERROR_*` consts; drift tests enforce). `attach`/`validate` reuse existing codes only.
- **`validate` diagnostics — the hard constraint:** on the standard Godot build there is **no bound API** for structured parse diagnostics. `GDScript.reload()` returns only an `Error`; `is_valid()` is not callable from GDScript; `--check-only` is unreliable (exits 0 on broken scripts). The per-error `line`/`message` exist only in the engine's stderr (**no column**, typically first error only). So `validate` reports `valid`/`error_string` structurally via the sentinel, and a per-command classifier parses **best-effort, advisory** `{line, message}` diagnostics from stderr. ADR-0002 is refined to legitimize this: stderr is still **never** parsed for the success/failure outcome or stable codes — only surfaced as advisory diagnostics.
- `ScriptValidateResult` intentionally exposes `valid` + `error_string` + `diagnostics`; the raw Godot `Error` int is omitted as an unstable engine internal (not a stable gda ABI).

## Tests

All three tiers, behavior-focused (round-trips, on-disk state, untouched-on-failure, descendant addressing). The riskiest path — `validate`'s stderr diagnostics parsing — is covered at S2 (pure parser, incl. ignoring `operations.gd`'s own backtrace frames), S3 (FakeRunner fed the exact engine stderr), and S1 (real engine).

- `uv run pytest -m "not e2e"` → **192 passed** (was 158)
- `uv run pytest -m e2e` → **110 passed** (was 92), real Godot 4.6.3

## Docs

- ADR-0002: 2 new registry rows + a new "stderr as advisory diagnostics" subsection.
- `command-catalog.md`: prose for `script set` (3 modes + the 1-based-over-`\n`-split line rule), `script attach`, `script validate`.

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -91,6 +91,8 @@ source and is checked by tests. GDScript mirrors only the rows whose source is
 | `node_not_found` | `operation` | `operation` | A requested node path does not resolve to a node in the scene. |
 | `unknown_property` | `operation` | `operation` | A requested property does not exist as a settable property on the node. |
 | `uncoercible_value` | `operation` | `operation` | A supplied value cannot be coerced to the property's declared Godot type. |
+| `no_search_match` | `operation` | `operation` | A search-replace script edit found no occurrence of the search string. |
+| `invalid_line_range` | `operation` | `operation` | A line-range script edit specified lines outside the script's bounds, or end before start. |
 | `contract_violation` | `parse` | `parser` | The process claimed success but violated the structured-output contract. |
 
 ## Considered options
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -106,6 +106,7 @@ source and is checked by tests. GDScript mirrors only the rows whose source is
 | `uncoercible_value` | `operation` | `operation` | A supplied value cannot be coerced to the property's declared Godot type. |
 | `no_search_match` | `operation` | `operation` | A search-replace script edit found no occurrence of the search string. |
 | `invalid_line_range` | `operation` | `operation` | A line-range script edit specified lines outside the script's bounds, or end before start. |
+| `script_compile_failed` | `operation` | `operation` | A script could not be attached to a node because it does not compile. |
 | `contract_violation` | `parse` | `parser` | The process claimed success but violated the structured-output contract. |
 
 ## Considered options
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -56,6 +56,19 @@ The GDScript payload owns only `code` and `message`. `gda` owns the public
 `GdaError` wrapper: it validates that the code is registered, assigns the
 `operation` category, preserves the message, and copies stderr into diagnostics.
 
+### stderr as advisory diagnostics
+
+stderr is still **never** parsed for the success/failure *outcome* or for stable
+error codes — those come only from the exit code and the stdout sentinel, as
+above. A command **may**, however, surface engine error text from stderr as
+**advisory, best-effort diagnostics** on its *success* result, when a useful
+detail is available nowhere else. `script validate` (#118) is the established
+case: when it reports `valid=false`, the per-error `line` and `message` exist
+only in the engine's stderr (no bound API exposes them), so `gda` parses them
+into the result's `diagnostics`. This stays within the contract: the diagnostics
+are advisory (they may hold only the first error, and `column` is unavailable on
+the standard build), and they never determine the outcome or a stable code.
+
 ## `GdaError.code` registry
 
 `GdaError.code` values are a public ABI for agents. Their authoritative source is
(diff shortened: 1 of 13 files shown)


qinhaihong-red:
**Review findings**

Requesting changes based on two behavior-level issues reproduced against real Godot:

1. Required: `script validate` does not compile the script with its real resource path.

In `src/gda/ops/operations.gd`, `_op_script_validate` creates `GDScript.new()`, assigns `source_code`, and calls `reload()` without setting the script resource path. That means valid scripts using relative resources are evaluated under a synthetic `gdscript://...` path and can be reported invalid incorrectly. I reproduced this with both `preload("helper.gd")` and `extends "base.gd"`: `script validate` returned `valid:false` with diagnostics such as `gdscript://helper.gd does not exist`, while the equivalent `res://helper.gd` form passed. Since the command contract is `validate PATH`, this should validate the actual `.gd` file context, including relative resource resolution.

2. Required: `script attach` misclassifies node/script native-type incompatibility as a compile failure.

At the bind check after `node.set_script(script)`, any null `node.get_script()` is reported as `script_compile_failed`. I reproduced a script that `script validate` reports as valid (`extends Node3D`) but attach to a `Node2D` fails with `script_compile_failed`; the engine stderr correctly says the script cannot be assigned to an object of type `Node2D`. This gives agents the wrong remediation: the script compiles, but it is not compatible with the target node type.

Verification run locally:

- `uv run pytest -m "not e2e"` -> `196 passed, 111 deselected`
- `git diff --check origin/main...HEAD` -> clean
- Targeted real-Godot repros for both findings above

Verdict: request changes before merge.

qinhaihong-red:
Thanks for the careful review — both findings reproduce and are valid.

Process note: this PR was already squash-merged before the review was actioned (my error — an ambiguous "merge" instruction was read as authorization). The maintainer has chosen to keep it merged, so both findings are tracked as follow-ups rather than changes to this PR:

- **Finding 1** (validate compiles without the script's real `res://` path → relative `preload`/`extends "base.gd"` false-negatives): tracked in #131, now enriched with your `extends "base.gd"` repro and the `gdscript://helper.gd does not exist` diagnostic.
- **Finding 2** (attach reports a compiling-but-type-incompatible script as `script_compile_failed`): a real regression introduced by this PR's attach bind-check — tracked in #136, and being fixed now in a follow-up PR. Independently confirmed against Godot 4.6.3: an `extends Node3D` script (compiles) attached to a `Node2D` is bounced by `set_script` with "Script inherits from native type 'Node3D', so it can't be assigned to an object of type 'Node2D'", which the bind-check mislabeled. `ClassDB.is_parent_class(node.get_class(), script.get_instance_base_type())` cleanly separates the two modes, so the fix reports a distinct `incompatible_script_type` with the correct remediation.

Issue #139, opened 2026-06-15
Refactor: collapse per-command execution wiring in the CLI into one _dispatch() helper


qinhaihong-red:
## What to build

Deepen the `gda` CLI command surface (`src/gda/cli.py`) by collapsing the execution wiring that every domain command repeats today into a single CLI-layer `_dispatch()` helper. This is a behavior-preserving refactor — architecture deepening candidate B (modest, safe scope), from an architecture review of the command surface.

Today each of the ~15 domain command functions repeats the same execution tail when calling `HeadlessCommand.emit()`: project resolution (`resolve_project_dir`), the runner seam (`make_runner=_make_runner`), the `json_output` pass-through, and the JSON-vs-text branch. The runner seam alone is referenced from 16 call sites.

Introduce one `_dispatch(cmd, params, *, json_output, godot, project, render)` helper in the CLI layer that owns this shared tail. Each command function keeps its own Typer signature, params construction (including per-command derivation such as the default scene root name and the default node name), `render` closure, and any pre-dispatch validation (`script create` mutual exclusion, `script set` mode validation) — only the repeated execution tail moves into `_dispatch()`.

Scope boundaries (decided in review):

- Command signatures are unchanged. The four common options (`--json`, `--schema`, `--godot`, `--project`) stay declared per command — Typer parses trailing flags from the signature, so removing them would change the CLI surface. This is the accepted ceiling of candidate B.
- Path normalization stays at the CLI layer (ADR-0006); it is not pushed into models.
- `info` is the lone meta command (no `--project`, ADR-0005); it gets a small meta dispatch path that does not call `resolve_project_dir`.
- No `Common` options struct — Typer cannot auto-inject one, so it would be a shallow pass-through.
- `_dispatch()` lives in the CLI layer (it injects CLI-layer concerns); `headless.py` is untouched.

Out of scope (separate architecture-review candidates, not part of this issue): moving rendering behind the result models, and making the error registry self-describing.

## Acceptance criteria

- [ ] A single `_dispatch()` helper in the CLI layer owns project resolution, the runner seam, and the JSON-vs-text branch for domain commands.
- [ ] All ~15 domain commands route their execution through `_dispatch()`; each command's per-command params construction, `render`, and pre-dispatch validation are preserved.
- [ ] `info` routes through a meta path that does not resolve a project; its behavior (no `--project`) is unchanged.
[...]

PR #143, opened 2026-06-15
refactor(render): extract presentation module, close result-model leaks (#140)

### LINKED ISSUE #140: Refactor: extract a presentation module for human rendering and close the result-model leaks
## What to build

Deepen the `gda` result models and the human-rendering seam — architecture deepening candidate 2 (variant C: a dedicated presentation module), from an architecture review of the structured-output layer. Behavior-preserving for the JSON path and for `--schema`.

Today the human-readable renderers live as free functions in the CLI command file and reach across the seam into result-model internals:

- `json.dumps(...)` is called directly on a model's `.value` field in two renderers (node `get` and node `set`), so the presentation layer encodes the model's value shape.
- One renderer reads `.path` / `.class_name` / `.extends` across a five-way union of script result types, so adding a field means hand-patching the renderer.

Extract a dedicated presentation module that owns human rendering, selected by result type, and close both leaks:

- Route value-to-text formatting through a typed helper instead of a raw `json.dumps` reaching into `.value`.
- Give the script result types a shared metadata interface (a base or protocol) so the renderer reads a typed surface, not a union.

The result models remain pure `--schema` / JSON data contracts (ADR-0004) — presentation does not move into them. The `--schema` output and the `--json` output are unchanged; only the human-readable text path is reorganized.

Composes with the `_dispatch()` refactor (#139): `_dispatch()` already takes a `render` callable, so each command sources its renderer from the new presentation module instead of an inline closure.

## Acceptance criteria

- [ ] Human-rendering functions live in a dedicated presentation module, not in the CLI command file; renderer selection is keyed by result type.
- [ ] No renderer reaches into a model's `.value` with a raw `json.dumps`; value-to-text goes through a typed helper.
- [ ] The script result types share a metadata interface; the script-metadata renderer reads that interface instead of a five-way union.
- [ ] Result models gain no presentation methods; `--schema` and `--json` output are unchanged byte-for-byte.
- [ ] Existing human-output assertions in the command tests pass, updated only where a test imported a now-moved private renderer.
- [ ] Full test suite passes, including the e2e tests against a real Godot engine.

## Blocked by

- #139 (collapse per-command execution wiring into `_dispatch()`) — land first so the `render` callable is already a dispatch parameter and the presentation extraction does not conflict in the CLI command file.

qinhaihong-red:
## What

Extract human rendering out of the CLI into a dedicated presentation module (`src/gda/render.py`), keyed by result type, and close the two leaks the architecture review found. Behavior-preserving — architecture deepening candidate 2 (variant C).

- **Presentation module.** All `_render_*` logic and inline render closures move to `render.py`; `render(result)` dispatches on `type(result)` via a `result-type -> renderer` table. Each command now passes `render=render` to #139's `_dispatch`/`_dispatch_meta`. An unregistered result type raises (loud wiring bug, not silent mis-format).
- **`.value` leak closed.** A single typed `format_value()` owns the JSON projection; no renderer reaches into a model's `.value` with a raw `json.dumps`.
- **Five-way union closed.** A `ScriptMetadata` `Protocol` (fields-only: `path`/`class_name`/`extends`) is the surface the script-metadata renderer reads, instead of a five-way union. A Protocol, not a shared base model — so it imposes nothing on field order and the `--schema` contract stays byte-for-byte identical.

## Why

The renderers lived in the CLI file and reached across the seam into model internals (`json.dumps(.value)` in two places; `.path/.class_name/.extends` over a union). Presentation now lives in one module; the result models stay pure `--schema`/`--json` data contracts (ADR-0004) with no presentation methods.

## Behavior preservation

- `src/gda/models.py` **untouched** (empty diff); no model gained a presentation method.
- `--schema` captured for all 7 affected commands and diffed before/after — **byte-for-byte identical**. `--json` (`model_dump_json`) unchanged.
- Human render text is identical for every command (renderers reproduce the exact prior strings).
- No existing test assertion changed; no test import edit needed. Added `tests/test_render.py` (17 tests) pinning the new typed seams.
- Full suite: **331 passed** (e2e against real Godot included). Targeted schema/render/command tests independently re-run on review: **86 passed**.

## Note

Stacked on #139 (PR #142) — base is `refactor/cli-dispatch-helper`. Rebase onto `main` once #139 merges.

Closes #140.

### DIFF
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -7,7 +7,6 @@ domain group (issue #18). Every command drives the same headless pipeline:
 binary resolution → runner → sentinel parse → typed model → JSON.
 """
 
-import json
 from importlib.metadata import version as package_version
 from pathlib import Path
 from typing import Optional
@@ -28,8 +27,6 @@ from gda.headless import (
 from gda.models import (
     EngineVersion,
     InfoParams,
-    ListedNode,
-    ListedScript,
     NodeAddParams,
     NodeAddResult,
     NodeGetParams,
@@ -46,7 +43,6 @@ from gda.models import (
     SceneGetResult,
     SceneListParams,
     SceneListResult,
-    SceneNode,
     ScriptAttachParams,
     ScriptAttachResult,
     ScriptCreateParams,
@@ -63,6 +59,7 @@ from gda.models import (
     ScriptValidateResult,
 )
 from gda.project import resolve_project_dir
+from gda.render import render
 from gda.runner import GodotRunner
 
 app = typer.Typer(
@@ -320,13 +317,6 @@ def _derive_scene_root_name(path: str) -> str:
     return filename
 
 
-def _render_tree(node: "SceneNode | ListedNode", depth: int = 0) -> str:
-    """Render a node tree as an indented ``name (Type)`` outline for humans."""
-    lines = [f"{'  ' * depth}{node.name} ({node.type})"]
-    lines += (_render_tree(child, depth + 1) for child in node.children)
-    return "\n".join(lines)
-
-
 @scene_app.command(cls=SCENE_CREATE_COMMAND.command_class())
 def create(
     path: str = typer.Argument(..., help="Target .tscn path to write."),
@@ -362,9 +352,7 @@ def create(
         json_output=json_output,
         godot=godot,
         project=project,
-        render=lambda created: (
-            f"created {created.path} (root {created.root_type})"
-        ),
+        render=render,
     )
 
 
@@ -383,21 +371,8 @@ def get(
         json_output=json_output,
         godot=godot,
         project=project,
-        render=lambda scene: _render_tree(scene.root),
+        render=render,
     )
-
-
-def _render_scene_list(listed: "SceneListResult") -> str:
-    """Render the enumerated scenes as ``path (root_name: root_type)`` lines."""
-    if not listed.scenes:
-        return "(no scenes)"
-    lines = []
-    for scene in listed.scenes:
-        if scene.root_name is not None and scene.root_type is not None:
-            lines.append(f"{scene.path} ({scene.root_name}: {scene.root_type})")
-        else:
-            lines.append(f"{scene.path} (unreadable)")
-    return "\n".join(lines)
 
 
 @scene_app.command(name="list", cls=SCENE_LIST_COMMAND.command_class())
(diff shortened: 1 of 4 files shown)


qinhaihong-red:
## Code review

I reviewed PR #143 with the code-review-and-quality checklist.

Verdict: no blocking issues found. The extraction of human rendering into `src/gda/render.py` keeps result models as `--json` / `--schema` contracts, and CLI commands now pass the module-level `render` function through the existing dispatch tail.

Consider: `tests/test_render.py` says existing command tests pin every command human-output path, but command-level human-output assertions currently only cover script validate/attach/set. The new render tests cover key seams and renderer registration, so this is not blocking, but a follow-up parameterized renderer snapshot would make the behavior-preservation claim stronger for scene/node and the remaining script/info renderers.

Verification run locally:
- `uv run pytest tests/test_render.py` -> 17 passed
- `uv run pytest tests/test_render.py tests/test_scene_commands.py tests/test_node_commands.py tests/test_script_commands.py tests/test_info_command.py tests/test_schema_command.py` -> 87 passed
- `uv run pytest -m "not e2e"` -> 218 passed, 113 deselected
- `uv run pytest` -> 331 passed in 189.44s
- `git diff --check origin/refactor/cli-dispatch-helper...HEAD` -> clean

Note: the referenced security/performance checklist files were not present in the skill directory, so I used the main five-axis review checklist.

qinhaihong-red:
Thanks for the review. Addressed the non-blocking follow-up in `b32cf7d`.

Added `tests/test_human_output.py` — a parameterized end-to-end test that drives **every** command in human mode (no `--json`) through the real Typer CLI against a fake runner and asserts the exact stdout (`result.stdout == rendered + "\n"`). 19 cases:

- **scene** create / get / list / list-empty / delete
- **node** add / list / get / set
- **script** create (with metadata) / create-bare / get / list / list-empty / delete / set / attach / validate
- **info**

Edge cases pinned, per your suggestion to harden the behavior-preservation claim:
- `format_value` non-scalar projection — `position (Vector2) = [10.0, 20.0]`, `set Hero.position (Vector2) = [3.0, 4.0]`, and `visible (bool) = true`.
- `render_script_metadata` branches — `class_name`+`extends`, `extends`-only, and bare path (no metadata).
- empty lists — `(no scenes)` / `(no scripts)`, plus the scene-list `(unreadable)` branch.

Test-only; no source changed. All 19 matched the actual CLI output on first run (no behavior discrepancy). `uv run pytest -m "not e2e"` → 237 passed; full `uv run pytest` → 350 passed (e2e vs real Godot included).

PR #144, opened 2026-06-15
refactor(errors): make the error registry self-describing (exit_code on ErrorCodeSpec) (#141)

### LINKED ISSUE #141: Refactor: make the error registry self-describing (exit_code on ErrorCodeSpec)
## What to build

Make the `gda` error registry self-describing — architecture deepening candidate 3, from an architecture review of the error/failure layer. Behavior-preserving: same error codes, same exit codes, same envelope.

Today a single error code is defined across three places: its category lives on `ErrorCodeSpec` (the registry), its exit code is a separate constant, and the binding between code and exit code is hardcoded inline at each failure-construction call site. Adding a code means editing all three with no compile-time check that the wiring is complete — only a registry test catches gaps.

Exit code is genuinely per-code, not per-category: within the ENVIRONMENT category, `binary_not_found` exits 127 while `launch_timeout` exits 124. So the exit code belongs on the per-code spec.

Add an `exit_code` field to `ErrorCodeSpec` and have failure construction derive both category and exit code from the registry by code. The human-facing `message` stays at the call site — it is per-occurrence (it embeds the binary path, the timeout, etc.).

## Acceptance criteria

- [ ] `ErrorCodeSpec` carries `exit_code` alongside its category; a code is fully defined by one registry row.
- [ ] Failure construction takes `(code, message, ...)` and derives category + exit code from the registry; no exit code is hardcoded at a failure-construction call site.
- [ ] The registry remains the single authoritative error ABI (ADR-0002); the existing registry test still validates it, extended to cover exit codes.
- [ ] Error codes, exit codes, and the error envelope shape are unchanged for every existing failure path.
- [ ] Full test suite passes, including the e2e tests against a real Godot engine.

## Blocked by

None - can start immediately.

qinhaihong-red:
## What

Make the error registry self-describing — a single `ErrorCodeSpec` row now fully defines a code's public ABI. Behavior-preserving refactor — architecture deepening candidate 3.

- Add `exit_code: int` to `ErrorCodeSpec`; populate all 31 rows from the `EXIT_*` constants in `exit_codes.py` (no inline magic numbers). Exit code is per-code, not per-category: within ENVIRONMENT, `binary_not_found` exits 127 while `launch_timeout` exits 124.
- `_failure(code, message, stderr)` now derives both `category` and `exit_code` from the registry by code (keeping the existing unregistered-code guard). The per-occurrence `message` stays at the call site; no category/exit-code is hardcoded at any of the 8 call sites.

## Why

A code used to be defined across three places — its category on the spec, its exit code as a separate constant, and the code→exit binding hardcoded inline at each `_failure(...)` call site, with only a test to catch gaps. Now one registry row is the single source of truth, and the call sites read as the taxonomy itself.

## Behavior preservation

- The error-envelope JSON (`GdaError`/`GdaErrorEnvelope`) has no `exit_code` field; `models.py` is untouched, so the wire shape is byte-for-byte unchanged. Exit code lives only on the internal `Failure` dataclass.
- Every failure path keeps its prior exit code — verified by 14 other test files that assert specific exit codes, all passing unchanged.
- `errors.py` now has zero `EXIT_*` references; `exit_codes.py` keeps all five constants (single audit registry, ADR-0003).
- `test_error_registry.py` extended to assert each code's exit code (the registry, not a call site, is the checked source of truth).
- Full suite: **316 passed** (e2e against real Godot included). Error suites independently re-run on review: **80 passed**.

## Scope

Touches `error_codes.py`, `errors.py`, and the registry test only. `cli.py`, `models.py`, `headless.py`, the render layer, `operations.gd`, ADRs, and `CONTEXT.md` untouched. Independent of #142/#143 (branches off `main`).

Closes #141.

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -76,39 +76,45 @@ the Python registry in `src/gda/error_codes.py`; the table below mirrors that
 source and is checked by tests. GDScript mirrors only the rows whose source is
 `operation`, because only those codes can be reported by headless operations.
 
-| Code | Category | Source | Meaning |
-| --- | --- | --- | --- |
-| `binary_not_found` | `environment` | `runner` | The Godot binary could not be launched. |
-| `launch_timeout` | `environment` | `runner` | Godot launched but did not return before the runner timeout. |
-| `unsupported_version` | `version` | `version_gate` | The detected Godot version is below the supported minimum. |
-| `engine_crashed` | `operation` | `classifier` | Godot terminated abnormally, such as by signal death. |
-| `operation_failed` | `operation` | `classifier` | The engine or operation failed without a valid registered operation error envelope. |
-| `usage_error` | `operation` | `operation` | The operation dispatcher was invoked without the required operation name. |
-| `unknown_operation` | `operation` | `operation` | The operation dispatcher received an unknown operation name. |
-| `invalid_params` | `operation` | `operation` | The operation dispatcher received params that are not a JSON object. |
-| `invalid_path` | `operation` | `operation` | A required path parameter is missing or invalid. |
-| `invalid_root_type` | `operation` | `operation` | A requested Godot root node type cannot be instantiated as a `Node`. |
-| `invalid_root_name` | `operation` | `operation` | A requested root node name is empty or would be rewritten by Godot. |
-| `already_exists` | `operation` | `operation` | A create operation target already exists and will not be overwritten. |
-| `save_failed` | `operation` | `operation` | A scene could not be packed or saved. |
-| `delete_failed` | `operation` | `operation` | A file could not be removed from disk. |
-| `project_not_found` | `operation` | `operation` | An operation that enumerates a project's `res://` tree ran without a resolved Godot project. |
-| `path_not_found` | `operation` | `operation` | A requested file does not exist. |
-| `not_a_scene` | `operation` | `operation` | A requested file cannot be loaded as a `PackedScene`. |
-| `parent_not_found` | `operation` | `operation` | A requested parent node path does not resolve to a node in the scene. |
-| `invalid_node_type` | `operation` | `operation` | A requested node type is neither an instantiable `Node` class nor a registered `class_name`. |
-| `invalid_node_name` | `operation` | `operation` | A requested node name is empty or would be rewritten by Godot. |
-| `duplicate_node_name` | `operation` | `operation` | The parent node already has a child with the requested name. |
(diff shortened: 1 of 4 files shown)


qinhaihong-red:
Review findings:

- [P2] `exit_code` is now part of the `ErrorCodeSpec` public ABI, but ADR-0002's mirrored registry table still only has `Code | Category | Source | Meaning`, and `tests/test_error_registry.py` still parses/checks only `(category, source)` from the ADR table. Issue #141 explicitly asks for the existing registry test to be extended to cover exit codes. The new `EXPECTED_EXIT_CODE` map checks the Python registry locally, but it does not prevent drift between ADR-0002 and the Python registry for exit codes. Suggested fix: add an `Exit Code` column to the ADR-0002 registry table and extend `_adr_registry()` / `_python_registry()` to compare it.

- [P3] `src/gda/error_codes.py` says adding a code is "one row, not edits across three sites", but ADR-0002 still requires adding/changing a `GdaError.code` to update the Python registry, ADR registry table, and any GDScript operation-code mirror. Suggested fix: narrow that wording to say failure-construction call sites no longer need to restate category/exit code.

Verification run locally on PR #144 HEAD:

- `uv run pytest tests/test_error_registry.py tests/test_classify_run.py tests/test_info_errors.py -q` -> 32 passed
- `uv run pytest -q` -> 316 passed in 192.71s

qinhaihong-red:
Thanks — both addressed in `79913a0` (P2) and `d69beac` (P3).

**P2 — ADR↔code exit-code drift closed.** Added an `Exit Code` column to the ADR-0002 registry table (31 rows, transcribed from `error_codes.py`: `binary_not_found`→127, `launch_timeout`→124, `unsupported_version`→3, operation/classifier→4, `contract_violation`→5), with a prose note on the per-code rationale. `_adr_registry()` and `_python_registry()` now both return `(category, source, exit_code)`, so `test_adr_registry_matches_python_authoritative_registry` compares exit codes too — the ADR doc is now the documented authority and the cross-check enforces ADR↔code parity on exit codes, not a test-local copy. The redundant `EXPECTED_EXIT_CODE` map and `test_every_code_carries_its_expected_exit_code` are removed (subsumed by the cross-check); `test_failure_derives_exit_code_from_registry` is kept (it tests `_failure`'s derivation behavior).

**P3 — wording narrowed.** The `ErrorCodeSpec` docstring no longer claims "one row, not edits across three sites"; it now says only that failure construction no longer restates a code's category or exit code — both come from the row. (Adding a new code still touches the Python registry, the ADR table, and the GDScript mirror for operation codes.)

Scope: ADR-0002 doc, `tests/test_error_registry.py`, and the `error_codes.py` docstring only — no exit-code value or runtime source changed. Full `uv run pytest` → 315 passed (e2e vs real Godot included).

Issue #135, opened 2026-06-15
operations.gd: extract shared script-read / path-guard / write / pack-save helpers


qinhaihong-red:
## Parent

#118 (PRD #17). Follow-up from the code review of #129. Pure refactor.

## What to build

Several blocks in the headless operations payload are duplicated across the script and node ops; extract them so a change to a load-bearing contract is made once:

- file-read + empty-vs-open-error disambiguation (script get / set / validate) -> a `_read_script_source`-style helper;
- `.gd`-path + exists guard (script get / delete / set / validate) -> a `_require_existing_script`-style helper;
- open / store / get_error / close write tail (script create / set) -> a `_write_script_file`-style helper;
- pack + save + `root.free()` mutate-and-save tail (script attach / node add / node set, most leak-prone) -> a `_repack_and_save`-style helper.

No behavior change; the full e2e suite is the regression gate.

## Acceptance criteria

- [ ] The four duplicated blocks are each behind one shared helper; call sites read as intent, not re-inlined boilerplate.
- [ ] No behavior change: full fast + e2e suites green (incl. scene/node ops sharing the mutate-and-save tail).
- [ ] The empty-vs-unreadable and partial-write guards remain intact (the bugs those blocks prevent do not regress).

## Blocked by

None - can start immediately. Touches the same ops as the validate-path / set-mode / attach-ordering issues; coordinate to minimize rebase churn.

PR #147, opened 2026-06-15
refactor(script): dispatch script set on an explicit mode discriminator (#133)

### LINKED ISSUE #133: script set: single authoritative edit-mode discriminator
## Parent

#118 (PRD #17). Follow-up from the code review of #129.

## What to build

`gda script set` decides its edit mode (search-replace / line-range / full) in **two** places that must stay in lockstep: the CLI `_validate_set_mode` enforces exactly-one-of, and the GDScript op re-derives the mode by presence-precedence. Collapse to a single source of truth — e.g. the CLI resolves and passes an explicit `mode` discriminator the op switches on — so the op never re-infers and precedence can't drift from the CLI's exclusivity rule.

## Acceptance criteria

- [ ] The edit mode is decided once; the op dispatches on an explicit discriminator, not param-presence inference.
- [ ] CLI mutual-exclusion behavior (exit 2 on missing/mixed mode) unchanged.
- [ ] All existing `script set` S2/S3/e2e behavior preserved (search-replace / line-range / full / failure modes).

## Blocked by

None - can start immediately.

qinhaihong-red:
## What

`gda script set` decided its edit mode (search-replace / line-range / full) in **two** places that had to stay in lockstep: the CLI's `_validate_set_mode` exclusivity check, and the GDScript op re-inferring the mode by param-presence precedence. The two could drift. This collapses them to a single source of truth.

Closes #133.

## Change

- **CLI is the one decider** — `src/gda/cli.py`: `_validate_set_mode` → `_resolve_set_mode(...) -> ScriptSetMode`. It still enforces exactly-one-of (same `typer.BadParameter` usage errors, exit 2) but now *returns* the resolved mode; `set_script` stamps it onto `ScriptSetParams(mode=...)`.
- **Explicit discriminator** — `src/gda/models.py`: new `ScriptSetMode(str, Enum)` (`search_replace | line_range | full`) and a required `mode` field on `ScriptSetParams`. Model-derived, so `--schema` (ADR-0004) now enumerates `mode` automatically — no hand-written schema to touch.
- **Op dispatches, never infers** — `src/gda/ops/operations.gd`: `_op_script_set` replaces the param-presence `if/elif/else` with `match mode`. A missing/unknown mode (only reachable via a malformed *direct* op call, never via the CLI) fails with the pre-existing, registered `invalid_params`. Keeps using the #135 shared helpers.

## Behavior preserved

- CLI mutual-exclusion is unchanged: missing or mixed mode still exits 2 with the same messages.
- All `script set` behavior preserved exactly: search-replace (`no_search_match`), line-range (1-based inclusive, `\n`-split, `invalid_line_range`), full overwrite, class_name/extends round-trip. No error code or message changes; no new error code.

## Tests

- Updated the three CLI dispatch tests in `tests/test_script_commands.py` to pin the resolved `mode` the CLI stamps.
- Added one op-level test `tests/test_e2e_op_robustness.py::test_script_set_dispatches_on_the_explicit_mode_not_param_presence`: it drives the op below the CLI with a params dict the CLI's exclusivity would never produce (`mode=full` **with** `start_line=1` also set) and asserts a full overwrite. It **fails** against the old presence-inference and **passes** against `match mode` — the regression net for "the op no longer re-infers".

Fast: `238 passed`. Full incl. e2e: `353 passed` (+1 = the new op test). Reviewer independently re-ran fast (238) + the affected e2e subset (12 passed), and confirmed `invalid_params` is a pre-existing registered code.

### DIFF
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -53,6 +53,7 @@ from gda.models import (
     ScriptGetResult,
     ScriptListParams,
     ScriptListResult,
+    ScriptSetMode,
     ScriptSetParams,
     ScriptSetResult,
     ScriptValidateParams,
@@ -686,11 +687,12 @@ def set_script(
     project: Optional[str] = project_option(),
 ) -> None:
     """Edit a .gd script via search-replace, line-range, or full overwrite."""
-    _validate_set_mode(search, replace, start_line, end_line, content)
+    mode = _resolve_set_mode(search, replace, start_line, end_line, content)
     _dispatch(
         SCRIPT_SET_COMMAND,
         ScriptSetParams(
             path=_normalize_path(path),
+            mode=mode,
             search=search,
             replace=replace,
             start_line=start_line,
@@ -704,19 +706,21 @@ def set_script(
     )
 
 
-def _validate_set_mode(
+def _resolve_set_mode(
     search: Optional[str],
     replace: Optional[str],
     start_line: Optional[int],
     end_line: Optional[int],
     content: Optional[str],
-) -> None:
-    """Enforce that ``script set`` selects exactly one edit mode (issue #118).
+) -> ScriptSetMode:
+    """Resolve ``script set``'s edit mode, the single source of truth (issue #133).
 
-    Validated at the CLI layer (before any dispatch, like ``script create``'s
-    mutual exclusion), so the operation is always handed exactly one well-formed
-    mode and never has to defend against "no mode" or a mixed-mode combination.
-    A violation is a usage error (exit 2).
+    This is the one place the edit mode is decided: it enforces that exactly one
+    of the three mutually-exclusive modes is supplied (a violation is a usage
+    error, exit 2, like ``script create``'s mutual exclusion) and returns the
+    resolved :class:`ScriptSetMode`. The CLI stamps it onto the op params so the
+    operation dispatches on this explicit discriminator instead of re-inferring
+    the mode from which params are present — the two can no longer drift.
     """
     has_search = search is not None or replace is not None
     has_line_range = start_line is not None or end_line is not None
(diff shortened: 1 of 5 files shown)


qinhaihong-red:
Code review result: request changes.

I found one required issue:

- `src/gda/ops/operations.gd:597` still documents `script-set` as choosing among modes by param-presence inference with precedence `search -> line-range -> full`. That is stale after this PR: `_op_script_set` now dispatches on the explicit `mode` discriminator. Since removing presence inference is the core intent of #133, please update this block comment before merge.

No implementation blockers found in correctness, architecture, security, performance, or dependency review.

Verification run locally:

- `UV_CACHE_DIR=/private/tmp/uv-cache-gda-codex uv run pytest -m "not e2e" -q` -> 238 passed, 115 deselected
- `UV_CACHE_DIR=/private/tmp/uv-cache-gda-codex uv run pytest tests/test_e2e_op_robustness.py -q` -> 4 passed
- targeted script/schema tests -> 24 passed
- `git diff --check main...HEAD` -> passed

Full `pytest -q` was also attempted and ended at 324 passed / 29 failed. The failures all reproduced as Godot 4.6.3 headless startup crashes in `RotatedFileLogger` while opening `user://logs/...`, across unrelated scene/node/script e2e tests, so I am treating that as an environment/Godot logging issue rather than a PR #147 regression.

qinhaihong-red:
Thanks — fixed in `ba611ee`.

**Required issue (stale comment) — addressed.** The `_op_script_set` header doc-comment described the edit mode as *"inferred by param presence with precedence search → line-range → full"*, which is exactly the behavior #133 removes. Updated it to describe the explicit `mode` discriminator the CLI resolves and the op dispatches on:

```
# Three mutually-exclusive edit modes; the CLI resolves exactly one and stamps it
# on the explicit `mode` discriminator the op dispatches on (issue #133), never
# re-inferred here from which params are present:
```

I also swept the branch for any other stale wording to be sure this was the only miss: `operations.gd:621` was already the corrected inline comment ("the op never re-infers it"), `operations.gd:922` is unrelated (the UID cache), and `models.py` was already updated in the original commit. `operations.gd:597` was the single remaining drift. Comment-only change, no behavior/test impact.

**On the full-suite failures (324 passed / 29 failed).** Agreed these are environmental, not a #147 regression — the `RotatedFileLogger` crash opening `user://logs/...` on Godot 4.6.3 headless lands across unrelated scene/node/script e2e tests, independent of this change. For reference, the full suite runs green in our environment: `uv run pytest -q` → **353 passed** (the implementing run and an independent reviewer re-run). The fast tier (238) and `test_e2e_op_robustness.py` (which exercises the new `match mode` dispatch) are green in both environments.

Issue #132, opened 2026-06-15
script attach: policy for an already-scripted node, and scene-before-script error ordering


qinhaihong-red:
## Parent

#118 (PRD #17). Follow-up from the code review of #129.

## What to build

Two `gda script attach` refinements, folded together.

**1. Already-scripted-node policy (the HITL decision).** `attach` calls `set_script` unconditionally, so attaching to a node that **already** has a script silently replaces the prior binding with no report of what was displaced. Decide and implement:
- **(a) refuse** when the node already has a script (a registered code, e.g. `node_already_scripted`), forcing an explicit replace; or
- **(b) overwrite-and-report** — keep replacing, but capture the previously-attached script's path (`node.get_script()` before `set_script`) and surface it (e.g. a `replaced_script` field on `ScriptAttachResult`, null when none).

**2. Scene-before-script error ordering (folded in from the former #134).** `attach` validates the `--script` path (extension + existence) **before** loading the scene, so when both the scene and the script are missing it reports the *script* as `path_not_found`, masking the missing scene — diverging from the node group (node get/set load the scene first). Load the scene first (or otherwise surface the scene failure first) so error ordering is consistent across scene-mutating commands. Trade-off: a wrong/missing script then pays the scene load+instantiate cost on the error path.

HITL overall because part 1 is a behavior/contract decision to make before implementing; part 2 then rides along in the same `_op_script_attach` change.

## Acceptance criteria

- [ ] Policy decision recorded (refuse vs overwrite-and-report) with rationale.
- [ ] Attaching to an already-scripted node no longer silently hides the displaced script.
- [ ] With both scene and script missing, attach reports the scene problem first (consistent with node get/set).
- [ ] All existing attach failure modes still map to their codes (missing script, wrong ext, node not found, missing scene, non-compiling script).
- [ ] `--schema`/model + command catalog updated; S3 + e2e cover the already-scripted path and the both-missing ordering.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Policy decision (HITL resolved)

Decided in a grilling session. Records the policy + rationale (closes the first acceptance criterion).

### Part 1 — already-scripted-node policy: **(b) overwrite-and-report**

`script attach` overwrites an existing binding instead of refusing it, but surfaces what it displaced.

[...]

PR #150, opened 2026-06-15
feat(node): connect-signal + disconnect-signal — wire signals to methods (#57)

### LINKED ISSUE #57: gda node connect-signal + disconnect-signal: wire signals to methods
## Parent

#17 — PRD: gda domain command surface.

## What to build

Wire and unwire a node's signal to a target node's method, persisted in the scene file.

- `gda node connect-signal` records a connection from a source node's signal to a target node's method, stored in the `.tscn`.
- `gda node disconnect-signal` removes an existing connection.

This slice carries a design decision: the CLI addressing for (source node, signal, target node, method), and whether the target method must already exist on the target's script. Decide it as part of the slice.

Reuses the node-path addressing from the node tracer (#53).

## Acceptance criteria

- [ ] `gda node connect-signal` records a signal→method connection between two nodes in the scene file and round-trips (a subsequent read shows the connection).
- [ ] `gda node disconnect-signal` removes an existing connection; errors if absent.
- [ ] The CLI addressing for (source node, signal, target node, method) is defined and documented; behavior when the signal or target method does not exist is defined (error vs allow).
- [ ] `--json` + the ADR-0004 `--schema` hard gate via the shared skeleton.
- [ ] Failure modes surface as structured `GdaError`s with registered codes (registry + ADR-0002 table + GDScript mirror).

**Note:** AFK for now. If the signal-addressing / method-existence design proves contentious during implementation, escalate to a `ready-for-human` discussion before committing the contract.

## Blocked by

- #53 (node-path addressing + the node-group plumbing).

qinhaihong-red:
## What

Adds `gda node connect-signal` and `gda node disconnect-signal` — wire/unwire a source node's signal to a target node's method, persisted in the `.tscn`.

Closes #57.

## The connection contract (ratified)

```
gda node connect-signal    SCENE --from <source-path> --signal <name> --to <target-path> --method <name>
gda node disconnect-signal SCENE --from <source-path> --signal <name> --to <target-path> --method <name>
```

- **Four-part addressing**, all flags required, mirroring the `.tscn` `[connection]` keys. The two node paths reuse the node group's strict node-path addressing (#53/#66).
- **Signal must exist** on the source → else `signal_not_found` (a typo / wrong node; cheap `has_signal` check).
- **Target method may dangle** (ratified): connecting to a not-yet-defined method is **allowed**, mirroring the Godot editor (wire first, author the handler later) — a `[connection]` is just persisted data. The asymmetry with the signal ruling is deliberate: a signal is a fixed part of the source node's type; a method lives on the target's script and is commonly written after the wiring.

## Persistence mechanism

`source.connect(signal, Callable(target, method), Object.CONNECT_PERSIST)` on the instantiated tree before re-pack. **`CONNECT_PERSIST` is load-bearing** — a plain `connect()` is runtime-only and `PackedScene.pack` drops it; only a persisting connection serializes to `[connection ...]`. Verified against Godot 4.6.3: the round-trip (re-read shows the `[connection]`, `is_connected()` true) and disconnect-removal hold.

- Duplicate connect → `already_connected` (guarded by `is_connected`, since a plain re-connect errors noisily).
- Disconnect of an absent connection → `connection_not_found` (guarded by `has_signal`/`is_connected`, since `disconnect()` errors on an absent connection).

## Error codes

- **New (registered in all three places — Python registry, ADR-0002 table, GDScript `OP_ERROR_*` mirror; `test_error_registry` drift test green):** `signal_not_found`, `already_connected`, `connection_not_found`.
- **Reused:** `node_not_found` (message labels the source/target endpoint), `path_not_found`, `not_a_scene`, `missing_dependency`, `save_failed`.

## Tests

- S2 model/parser (`tests/test_models.py`), S3 CLI-with-FakeRunner (`tests/test_node_commands.py`, `tests/test_node_errors.py`), S1 e2e (`tests/test_e2e_node.py`): connect round-trips, disconnect removes, disconnect-absent errors, signal-not-found, already-connected, the dangling-method allowance, and human-output rendering.
- `--json` + the ADR-0004 `--schema` hard gate (model-derived); `tests/test_schema_command.py` extended.

Fast: `254 passed`. Full incl. e2e: `380 passed`. Reviewer independently re-ran fast (254) + the signal e2e subset (9 passed) and audited `root.free()` on all paths.

## Notes

- `docs/command-catalog.md` node section gains a "Signal wiring" subsection (addressing, the signal-exists / method-may-dangle contract, the `CONNECT_PERSIST` mechanism, failure modes).
- `from` is a Python keyword, so the model field is `from_node` aliased to the wire key `from` (`populate_by_name` + `serialize_by_alias`), keeping `from` consistent across params / `--json` / `--schema`.


qinhaihong-red:
Review verdict: Request changes.

I found one required contract issue before this should merge.

`disconnect-signal` does not pin the missing-signal behavior consistently with the public contract. In `src/gda/ops/operations.gd`, `disconnect-signal` treats `not source.has_signal(signal_name)` the same as an absent connection and returns `connection_not_found`. The command catalog, however, describes source signal existence as part of the signal-wiring contract and says a missing or typoed signal is `signal_not_found`.

Evidence:
- `src/gda/ops/operations.gd`: `disconnect-signal` currently combines `not source.has_signal(signal_name)` with `not source.is_connected(...)` and reports `connection_not_found`.
- `src/gda/ops/operations.gd`: `connect-signal` reports the same missing-source-signal condition as `signal_not_found`.
- `docs/command-catalog.md`: the signal wiring contract says the source signal must exist and missing signals are `signal_not_found`.
- Current tests cover connect with an unknown signal and disconnect with a valid signal but absent connection, but do not cover `disconnect-signal --signal no_such_signal`.

Please choose and pin one contract:

1. If `disconnect-signal` should also treat an unknown source signal as `signal_not_found`, update the implementation and add S3 + S1 tests for `disconnect-signal --signal no_such_signal`.
2. If disconnect intentionally collapses unknown signal into `connection_not_found`, update the command catalog / ADR wording to scope `signal_not_found` to connect only, and add the same tests so the ABI is explicit.

Verification I ran locally:

```text
uv run pytest -m "not e2e"
254 passed, 126 deselected

uv run pytest tests/test_e2e_node.py -k "connect_signal or disconnect_signal"
9 passed, 45 deselected
```

Other reviewed paths looked sound: the `from` alias is consistent across JSON/schema/runner params, `CONNECT_PERSIST` is used for persistence, and the `root.free()` paths are covered through the shared mutation save path.

qinhaihong-red:
Thanks — valid catch, fixed in `2307e2b`.

**Agreed it's a real inconsistency.** `disconnect-signal` collapsed an unknown source signal into `connection_not_found`, while `connect-signal` reports the same condition as `signal_not_found` and the catalog's signal-wiring contract states a missing source signal is `signal_not_found`.

**Chose option 1 (symmetric `signal_not_found`).** Split the disconnect guard so a missing `has_signal` is `signal_not_found` and only a genuinely absent connection (signal present, not wired to this target/method) is `connection_not_found`:

```gdscript
if not source.has_signal(signal_name):
    _fail(OP_ERROR_SIGNAL_NOT_FOUND, "source node " + from_path + " has no signal: " + signal_name)
    return
if not source.is_connected(signal_name, callable):
    _fail(OP_ERROR_CONNECTION_NOT_FOUND, "no such connection: ...")
    return
```

Rationale: the signal's existence is a type-level property of the source node, independent of direction, so `signal_not_found` (fix the signal name) vs `connection_not_found` (signal valid, this wiring isn't there) are genuinely different remediations. Option 1 makes the code match both `connect-signal` and the **already-written** catalog, so **no doc change** was needed — only the code had drifted.

**Tests added** (both requested):
- S1 e2e `test_node_disconnect_signal_missing_signal_yields_signal_not_found` — `disconnect-signal --signal no_such_signal` → `signal_not_found`, file untouched (red against the old collapsed guard).
- S3 `test_disconnect_signal_missing_signal_maps_to_signal_not_found_code`.

Also rebased onto current `main` (it now includes #132 / #149) — clean, no conflicts. Local: fast `256 passed`; signal e2e subset `10 passed`.

PR #151, opened 2026-06-15
feat(node): remove + duplicate + move — structural node-tree edits (#56)

### LINKED ISSUE #56: gda node remove + node move + node duplicate: structural node-tree edits
## Parent

#17 — PRD: gda domain command surface.

## What to build

Structural mutations of the node tree within a scene file (load → locate → restructure → pack → save).

- `gda node remove` deletes a node and its subtree.
- `gda node move` reparents a node (with its subtree) under a new parent.
- `gda node duplicate` duplicates a node (and subtree) under a parent with a fresh, non-colliding name.

Reuses the node-path addressing from the node tracer (#53).

## Acceptance criteria

- [ ] `gda node remove` removes a node and its subtree; defined behavior (error) for a missing node or removing the scene root.
- [ ] `gda node move` reparents a node (with subtree) to a target parent; errors on an invalid or cyclic target.
- [ ] `gda node duplicate` duplicates a node + subtree under a parent with a fresh unique name; returns the new node path.
- [ ] Each returns a typed JSON result and ships `--json` + the ADR-0004 `--schema` hard gate.
- [ ] Failure modes surface as structured `GdaError`s with registered codes (registry + ADR-0002 table + GDScript mirror).

**Implementation order within the slice:** `remove` → `duplicate` → `move` (increasing complexity).

## Blocked by

- #53 (node-path addressing + the node-group plumbing).

qinhaihong-red:
## What

Three structural node-tree edits (load → locate → restructure → pack → save), reusing the node group's strict node-path addressing (#53/#66). Delivered as three green vertical slices in increasing complexity: `remove` → `duplicate` → `move`.

Closes #56.

```
gda node remove    <scene> --node <path>
gda node duplicate <scene> --node <path>
gda node move      <scene> --node <path> --to <new-parent-path>
```

- **remove** — deletes a node and its subtree (`remove_child` + `free`).
- **duplicate** — copies a node + subtree under the same parent with a fresh, non-colliding name (`Hero`→`Hero2`→`Hero3`…, skipping any taken name); `node.duplicate()` then `_reown_subtree` so the unowned copy serializes; returns the new node path.
- **move** — reparents a node + subtree under a new parent (manual `remove_child`/`add_child` + `_reown_subtree`).

## Failure modes

- **Reused codes:** `node_not_found` (missing source), `parent_not_found` (move's bad/non-canonical `--to`), `duplicate_node_name` (move name collision at target), plus `missing_dependency`/`path_not_found`/`not_a_scene` via the shared load ladder.
- **New codes** (registered in all three places — Python registry, ADR-0002 table, GDScript `OP_ERROR_*` mirror; `test_error_registry` drift test green):
  - `cannot_target_root` — remove/duplicate/move targeting the scene root (it has no parent).
  - `cyclic_target` — move under self or own descendant (`target == node or node.is_ancestor_of(target)`).

`root.free()` runs on every error path of all three ops; result fields are captured off the live tree before `_repack_and_save` frees it (a freed-node read in move's collision branch was caught by its e2e RED and fixed).

## Tests

- S2 model round-trips (`tests/test_models.py`), S3 dispatch + failure-mode mapping (`tests/test_node_commands.py`, `tests/test_node_errors.py`), S1 e2e (`tests/test_e2e_node.py`): round-trips + every failure mode (missing node, remove-root, duplicate fresh-name/nested-subtree, cyclic-under-self, cyclic-under-descendant, move name-collision, move-to-root).
- `--json` + the ADR-0004 `--schema` hard gate (model-derived); `tests/test_schema_command.py` extended. Renderers added in `src/gda/render.py`.

Fast: `256 passed`. Full incl. e2e: `390 passed`. Reviewer independently re-ran fast (256) + the structural-edit e2e subset (22 passed) and audited `root.free()` across all three ops.

## Notes

`docs/command-catalog.md` node section gains a "Structural edits" block (the shared root rule, the remove-subtree rule, the duplicate fresh-naming rule, the move cyclic-target rule). This PR shares the node command surface / error registry / catalog node section with #57 (signal wiring) and #132 (attach) — whichever merges later will need a trivial rebase.


qinhaihong-red:
## Review: Request changes

I found two required correctness issues in `node move`.

1. `node move` to the current parent is not a no-op.

In `src/gda/ops/operations.gd`, the implementation treats an existing child under the target as non-colliding, but still executes `remove_child` followed by `add_child`. That appends the node to the end of the same parent, changing sibling order even though the requested parent is already satisfied.

Repro from a temporary project:

```json
{
  "command": "gda node move main.tscn --node A --to . --json",
  "before": ["A", "B", "C"],
  "after": ["B", "C", "A"],
  "reported_path": "A"
}
```

Sibling order is meaningful in Godot scenes, so this should either return success without remove/add/save when `node.get_parent() == target`, or preserve the original index explicitly.

Relevant code: `src/gda/ops/operations.gd` lines 645-660.

2. `node move` rewrites ownership inside instanced sub-scenes.

The move path calls `_reown_subtree(node, root)`. That is appropriate after `duplicate()`, because the duplicate is unowned, but it is not safe for move. Moving an existing subtree should preserve ownership. With the existing editable-instance fixture, moving `ChildInstance` under `Target` kept `instance=ExtResource(...)` and `[editable ...]`, but rewrote inherited/override children into locally owned nodes with `type=` and new `unique_id` values. That changes instance inheritance semantics and violates the #64 mutation integrity boundary.

Observed after moving `ChildInstance`:

```text
[node name="ChildInstance" parent="Target" unique_id=... instance=ExtResource("1_child")]
[node name="Inner" type="Sprite2D" parent="Target/ChildInstance" unique_id=...]
[node name="Deep" type="Node2D" parent="Target/ChildInstance/Inner" unique_id=...]
[editable path="Target/ChildInstance"]
```

The original override-style inherited nodes should not be converted into locally owned nodes. Please avoid recursively changing owner on move, and add an e2e regression test that moves an editable instanced scene while preserving the instance reference, editable marker, and override serialization.

Relevant code: `src/gda/ops/operations.gd` line 660.

Verification I ran locally:

- `uv run pytest -m "not e2e"` -> `256 passed`
- `uv run pytest tests/test_e2e_node.py -m e2e -k "node_remove or node_duplicate or node_move"` -> `17 passed`
- Manual temporary-project probes reproduced both issues above

qinhaihong-red:
Thanks — both `node move` bugs are real; fixed in `289a4f4` (and the branch is now merged up to current `main`, which includes #57 and #132).

**Bug 1 — same-parent move is now a no-op.** When `node.get_parent() == target`, move short-circuits to a successful no-op: it reports the node's current path/name/type and returns **without reparenting and without re-saving**, so siblings are never reordered and the file is untouched. The check sits **before** the name-collision check, so a node already under the target isn't mistaken for a collision with itself.

**Bug 2 — instance ownership is preserved.** Replaced the manual `remove_child`/`add_child` + `_reown_subtree(node, root)` with **`node.reparent(target, false)`**. Empirically (Godot 4.6.3, editable-instance fixture):
- The old path corrupted the instance: `_reown_subtree` reached into the instance's `Inner`/`Deep` override nodes and re-owned them to root, rewriting them as local `type=Sprite2D`/`type=Node2D` nodes.
- `remove_child` did **not** clear the moved node's own `owner`, so only the descendant reown was destructive.
- `reparent()` preserves the moved node's **and its descendants'** owners → `instance=ExtResource(...)`, `[editable path=...]`, and the `parent=… index=` override nodes all serialize intact.
- `keep_global_transform=true` (the default) rewrites the local transform; passing **`false`** keeps it, making move a purely structural edit with no transform churn. `_reown_subtree` is retained for `node duplicate`, which genuinely needs it (the copy is unowned).

**Regression tests** (both red against the pre-fix merged code, green after):
- `test_node_move_to_current_parent_is_a_noop_preserving_sibling_order` — `[A,B,C]`, `move A --to .`, saved order stays `[A,B,C]`.
- `test_node_move_preserves_editable_instance_overrides` — moves an editable `ChildInstance` under a new parent; asserts `instance=ExtResource(`, `[editable path=...]`, and the `Inner`/`Deep` override form (`parent=… index=`, no `type=`) survive.

Local: fast `274 passed`; full `tests/test_e2e_node.py` `74 passed` (structural edits + signal wiring + the two regressions coexist after the merge). The catalog's `move` paragraph now documents the same-parent no-op and the instance-preservation contract.

(Note: `reparent()` emits a benign "make owner inconsistent" WARNING to stderr; harmless — diagnostics route to stderr and the packed result is correct.)

PR #60, opened 2026-06-10
feat(node): gda node add + node list — the node-group tracer

qinhaihong-red:
## Summary

Implements #53: establishes the `node` command group — the node-group tracer mirroring the `scene` tracer (#18) — with the load → locate → mutate → pack → save round-trip every other `node` command will reuse.

- **`gda node add`** loads a `.tscn`, adds a child node under a parent node path, names it, packs and saves. `--type` resolves first as a built-in Godot `Node` class, then as a `class_name` from the project's global class list (instantiating the script and attaching it). `--name` defaults to the type name, `--parent` to the scene root. Returns the created node's `path`/`name`/`type` (+ `script_class` for class_name additions) as a typed JSON result.
- **`gda node list`** reports the scene's node tree (name/type/path, nestable) from `SceneState` without instantiating — the structured verifier of `node add` (add → list round-trips), preserving the no-code-execution read property (#30).

## Node-path addressing (established by this slice)

A node within a scene is addressed by its node path **relative to the scene root**: `.` is the root itself, `Player/Arm` a nested node; absolute paths are rejected. `node list` reports every node's path in exactly the form `node add --parent` accepts. Documented in the command catalog and in the `--schema` field descriptions.

## Error contract (ADR-0002)

New operation codes registered in the Python registry + ADR-0002 table + GDScript mirror (drift tests cover all three): `parent_not_found`, `invalid_node_type`, `invalid_node_name`, `duplicate_node_name`. Scene-file failures reuse `path_not_found` / `not_a_scene` / `invalid_path`; failed mutations leave the scene file untouched (asserted e2e).

## Tests (TDD, red→green per slice)

- **S2**: model round-trips for `NodeAddResult` / nested `NodeListResult` trees.
- **S3** (CLI + FakeRunner): success paths, defaults, all five failure modes, `--schema` hard gate (model-derived contracts, sample payloads validate, no Godot spawned).
- **S1** (e2e, real Godot): add → list round-trip, nested parent addressing, default name, all failure modes end-to-end, and class_name resolution against an imported project fixture (`godot --headless --import`).

Full suite: 128 passed (31 e2e against Godot 4.6).

## Known limitation

Mutation requires instantiating the scene, so `node add` on a scene that already contains scripted nodes runs those scripts' `_init` (unlike the read ops, which stay on `SceneState`). Noted in `operations.gd`; inherent to headless file mutation.

Closes #53

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #154, opened 2026-06-15
refactor(node): node-group cleanup + test hardening (#67)

### LINKED ISSUE #67: node group cleanup + test hardening (post-#53 polish)
## Parent

#53 — gda node add + node list. Low-priority, non-functional polish surfaced by the code review of PR #60. None of these change behavior; bundle them as a single cleanup slice.

## What to build

Cleanup:

1. **Remove the dead rewrite guard in `_op_node_add`** — the `actual_name != node_name` check after assigning `.name` to a parentless node is unreachable: `_is_valid_node_name` already rejects the chars Godot sanitizes, and a parentless node never gets the `@`-dedup suffix (that only happens in `add_child`, already guarded by the duplicate check). It copies scene-create's guard into a context where it cannot fire. (If a real rewrite case is found, it belongs in the validator, not as a post-assignment recheck.)
2. **Make `ListedNode` subclass `SceneNode`** — `ListedNode` currently duplicates the whole recursive model to add one `path` field, forcing a `"SceneNode | ListedNode"` union in `_render_tree`. Prefer `class ListedNode(SceneNode): path: str` so the tree shape has one source of truth.
3. **Extract a shared pack-and-save helper** — `_op_node_add` and `_op_scene_create` repeat the same `PackedScene.new()` / `pack()` / save-failed / `ResourceSaver.save()` / `free` / `_save_failure_message` tail verbatim; a `_pack_and_save(root, path)` helper removes the drift risk.

Test hardening:

4. **Assert `script_class` is null e2e for a built-in type** — currently only asserted in the model round-trip and fake-runner dict; a broken `_script_class_of` returning garbage for a plain Sprite2D would pass all e2e success tests.
5. **Cover `node list` with multiple siblings at one level** — every fixture has at most one child per parent, so sibling enumeration and child ordering are unverified.

## Acceptance criteria

- [ ] Dead rewrite guard removed; node-add behavior unchanged (tests still green).
- [ ] `ListedNode` reuses `SceneNode`; `--json` / `--schema` output for `node list` unchanged.
- [ ] Pack-and-save logic lives in one helper used by both `node add` and `scene create`.
- [ ] e2e asserts `script_class is None` for a built-in-type add.
- [ ] A `node list` test asserts a parent with >1 child reports all siblings in tree order.

## Blocked by

None - can start immediately. Best landed after #64 (which rewrites part of `_op_node_add`) to avoid churn, but not strictly blocked.

qinhaihong-red:
## What

Non-functional cleanup + test hardening for the node group, surfaced by the PR #60 review. No behavior change except the two added tests.

Closes #67.

## Cleanup

1. **Removed the dead rewrite guard in `_op_node_add`** — the `actual_name != node_name` recheck after assigning `.name` to a parentless node was unreachable (`_is_valid_node_name` already rejects the sanitized chars, and the `@`-dedup suffix only happens in `add_child`, already guarded by the duplicate-name check). Replaced with a comment; the result now uses the validated `node_name`.

2. **`ListedNode` now subclasses `SceneNode`** (`class ListedNode(SceneNode): path: str`, per the issue) — collapses the `"SceneNode | ListedNode"` union in `render_node_tree` to a single tree type, so the tree shape has one source of truth.
   - **One honest caveat on the "output unchanged" criterion:** the `node list` `--schema`/`--json` are **semantically identical** (same fields, same `required` `["name","type","path"]`, same field definitions and `$ref` recursion; `$defs` holds only `ListedNode`, no `SceneNode` leak) — but the **property declaration order** of `ListedNode` changes: `path` now follows `children` instead of preceding it, because a subclass inherits `name`/`type`/`children` first and appends `path`. This is the inherent consequence of the issue-prescribed `class ListedNode(SceneNode): path: str`. JSON property order is insignificant (agents parse by key; MCP generation from `--schema` is order-insensitive), and the whole suite — including the `--schema` hard-gate — stays green. If strict byte-stability of the schema string is required, the fields can instead be re-declared on `ListedNode` in the original order (at the cost of duplicating the field defs); flagging the trade-off rather than deciding it silently.

3. **Consolidated the pack-and-save tail** — re-scoped: #135 already extracted `_repack_and_save(root, path)` and routed the node ops through it, so **no new helper was created**. `_op_scene_create` still inlined the same tail; routed it through `_repack_and_save` too. `_ensure_parent_dirs` runs **before** the shared tail (pack is purely in-memory, so creating dirs first is equivalent), freeing `root` explicitly on the dir-creation failure path. `created_dirs` and the `save_failed` diagnostics are unchanged.

## Test hardening

4. **e2e asserts `script_class is None` for a built-in-type add** (`node add --type Sprite2D`) — previously only the model round-trip / fake-runner dict asserted it, so a broken `_script_class_of` returning garbage for a plain node would pass every e2e success test.
5. **e2e covers `node list` with >1 sibling at one level** — three distinct-typed children under the root, asserting all are reported with correct `(name, type, path)` in tree order (sibling enumeration / child ordering was previously unverified).

## Tests

Fast: `274 passed`. Full incl. e2e: `425 passed` (+2 = the new e2e tests). Reviewer independently re-ran fast (274) + the scene/node e2e subset (19 passed), verified the dead guard is gone, and diffed the `node list --schema` against `main` (semantically identical; only the `path`/`children` property order differs, as noted above).

### DIFF
--- a/src/gda/models.py
+++ b/src/gda/models.py
@@ -310,7 +310,7 @@ class NodeAddResult(BaseModel):
     )
 
 
-class ListedNode(BaseModel):
+class ListedNode(SceneNode):
     """One node of ``gda node list``'s tree: name, type, node path, children.
 
     Like ``SceneNode`` but each node also carries its node path relative to the
@@ -318,8 +318,12 @@ class ListedNode(BaseModel):
     other node commands (e.g. ``node add --parent``).
     """
 
-    name: str
-    type: str
+    # Subclasses ``SceneNode`` so the recursive tree shape (name/type/children)
+    # has one source of truth; ``children`` is re-declared as ``list[ListedNode]``
+    # — not the inherited ``list[SceneNode]`` — so a listed node's children are
+    # themselves listed nodes (carrying ``path``), keeping the tree self-recursive
+    # on ``ListedNode`` and the ``--json``/``--schema`` output byte-for-byte
+    # unchanged.
     path: str = Field(
         description=(
             "The node's node path relative to the scene root: '.' for the root "
--- a/src/gda/models.py
+++ b/src/gda/models.py
@@ -322,8 +322,13 @@ class ListedNode(SceneNode):
     # has one source of truth; ``children`` is re-declared as ``list[ListedNode]``
     # — not the inherited ``list[SceneNode]`` — so a listed node's children are
     # themselves listed nodes (carrying ``path``), keeping the tree self-recursive
-    # on ``ListedNode`` and the ``--json``/``--schema`` output byte-for-byte
-    # unchanged.
+    # on ``ListedNode``. The ``--json``/``--schema`` output stays SEMANTICALLY
+    # unchanged (same fields, same ``required``, same recursion) but is NOT byte-
+    # identical: inheriting name/type/children and appending ``path`` orders the
+    # properties name,type,children,path rather than the prior name,type,path,
+    # children. JSON property order is insignificant — consumers parse by key and
+    # the ``--schema`` → MCP generation is order-insensitive — so this is a safe
+    # trade for the single-source-of-truth tree shape.
     path: str = Field(
         description=(
             "The node's node path relative to the scene root: '.' for the root "
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -177,21 +177,17 @@ func _op_scene_create(params: Dictionary) -> void:
 		_fail(OP_ERROR_INVALID_ROOT_NAME, "Godot rewrote root_name from " + root_name + " to " + actual_root_name)
 		return
 
-	var packed := PackedScene.new()
-	var pack_err := packed.pack(root)
-	if pack_err != OK:
-		root.free()
-		_fail(OP_ERROR_SAVE_FAILED, "failed to pack scene: " + error_string(pack_err))
-		return
+	# Create any missing parent dirs BEFORE packing-and-saving. pack() works purely
+	# in memory on root and writes nothing, so making the directories first (rather
+	# than between pack and save) is behavior-equivalent and lets scene-create reuse
+	# the one shared pack-and-save tail (_repack_and_save) the node ops use (#135).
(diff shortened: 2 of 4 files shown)


qinhaihong-red:
Review result: request changes for one documentation/contract wording issue.

Finding:
- `src/gda/models.py`: the `ListedNode` comment says the `node list --json` / `--schema` output is “byte-for-byte unchanged”, but that is not true after `ListedNode` inherits from `SceneNode`. I compared `node list --schema` against the PR base: the parsed JSON schema is semantically equal, but the observable property order changes from `name,type,path,children` to `name,type,children,path`. Please either update the comment to describe semantic/order-insensitive compatibility, or preserve the old field order if issue #67 requires byte-stable output.

Verification I ran locally in a temporary worktree:
- `uv run pytest -m 'not e2e'` -> 274 passed
- Targeted e2e coverage for the new node tests plus scene-create save paths -> 4 passed

I did not find behavior, architecture, security, or performance blockers beyond the inaccurate comment / contract wording.

qinhaihong-red:
Thanks — correct, the comment was wrong. Fixed in `bca86a9` (option 1: corrected the wording; kept the `SceneNode` subclassing).

The `ListedNode` comment claimed the output is "byte-for-byte unchanged"; it is not — as you measured, inheriting `name`/`type`/`children` and appending `path` orders the properties `name,type,children,path` instead of the prior `name,type,path,children`. The comment now states the output is **semantically** unchanged (same fields, same `required`, same recursion) but **not** byte-identical, and why the property-order change is safe (JSON property order is insignificant — consumers parse by key, and `--schema` → MCP generation is order-insensitive).

I kept the subclassing rather than preserving byte order (option 2) because the order change is semantically null and the single-source-of-truth tree shape is item 2's actual goal — preserving the exact byte order would require re-declaring `name`/`type`/`children` on `ListedNode`, re-introducing the duplication the change removes. The PR description already carried this trade-off; the stale code comment was the gap, now closed. (If strict byte-stability of the schema string is a hard requirement, say so and I'll switch to the re-declared-fields form.)

Local: fast `274 passed`.

PR #156, opened 2026-06-15
fix(classify): distinguish deep scene trees from contract violations

### LINKED ISSUE #37: Deep scene trees must not be misclassified as a contract violation
## Parent

#17 — PRD: gda domain command surface. Found during the high-effort code review of #18's implementation (merged via #41).

## What to build

The recursive \`SceneNode\` model is validated by pydantic-core, which has an internal recursion ceiling (~255 levels). A legitimately deep scene tree from \`scene get\` therefore raises \`ValidationError\` on depth, which \`classify_run\` reports as \`parse / contract_violation\` (exit 5) — blaming the engine payload for a wrapper-side limitation. Verified: depth 254 validates, depth 255 fails. The human-readable \`_render_tree\` has the same unbounded recursion (pydantic fails first today).

Make deep-but-valid trees succeed, or fail with an accurate structured error that is not "the engine violated the output contract". Options: raise/configure the recursion limit, render/validate iteratively, or surface a distinct \`tree_too_deep\` operation-side limit.

## Acceptance criteria

- [ ] A valid scene nested past the current ~255 limit either returns its tree or fails with an accurate, distinct structured error — not \`parse / contract_violation\`.
- [ ] The human-readable (non-\`--json\`) render path does not raise an unstructured \`RecursionError\`.
- [ ] A unit test exercises a tree at/above the previous failing depth.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Problem

A legitimately deep scene tree from `scene get` nests past pydantic-core's hardcoded recursive-validation ceiling (~255 levels). The resulting `ValidationError` on depth was reported by `classify_run` as `parse / contract_violation` (exit 5) — blaming the engine payload for a wrapper-side limit. The recursive `_render_tree` / `render_node_tree` had the same unbounded recursion on the non-`--json` path (pydantic failed first today, but the latent `RecursionError` is real).

Verified locally: depth 255 validates, 256 fails (boundary varies slightly by pydantic-core version; the issue reported 254/255). pydantic-core reports the depth ceiling with the `recursion_loop` error type — the *same* type used for genuine cyclic references — which gives a clean, version-stable discriminator.

## Fix

Route chosen: **distinct structured error for the classification/`--json` path + iterative render for the human path.** Making validation "succeed" via a higher limit is not possible — pydantic-core's recursion guard is a hardcoded Rust limit, unaffected by `sys.setrecursionlimit`; and even if validation were bypassed via `model_construct`, `model_dump_json` hits the *same* ceiling on serialization. So fully succeeding the `--json` path would mean replacing pydantic's validator *and* serializer with custom iterative code — fighting the framework on both ends. A precise, distinct error is the honest fix.

- `classify_run` now detects a `ValidationError` whose errors are **all** `recursion_loop` and surfaces a distinct `tree_too_deep` failure whose message names the limit as gda's own — never `contract_violation`. A mix of `recursion_loop` with other error types stays a genuine `contract_violation` (a real shape violation that merely happens to also be deep).
- `render_node_tree` is now an explicit-stack pre-order walk, so the human (non-`--json`) path renders an arbitrarily deep tree without raising an unstructured `RecursionError`; output order is unchanged (existing render tests pass untouched).

## New error code

`tree_too_deep` — category `parse`, source `classifier`, exit code `6` (new `EXIT_TREE_TOO_DEEP`). Registered in the Python registry (`error_codes.py`), the ADR-0002 table, and `exit_codes.py`. **No GDScript mirror** — it is a classifier-assigned code, not an operation-reported one, so the registry-drift test correctly does not require it in `operations.gd`.

A distinct exit code (6, not 5) lets a shell consumer tell "tree too deep for the wrapper" apart from "the engine violated the output contract", matching the ADR's per-code exit-code rule (precedent: ENVIRONMENT spans 127 and 124).

## Tests

- `tests/test_classify_run.py`: a 1000-deep valid `SceneGetResult` payload classifies as `tree_too_deep`, not `contract_violation`, with an accurate wrapper-side message.
- `tests/test_render.py`: a 2000-deep `SceneNode` chain and a 2000-deep `ListedNode` chain each render without `RecursionError`, one indented line per node.
- Full suite green: **427 passed** (incl. real-Godot e2e). CI-equivalent `pytest -m "not e2e"` + `uv build` pass.

## Acceptance criteria

- [x] A valid scene nested past the ~255 limit fails with an accurate, distinct structured error — not `parse / contract_violation`.
- [x] The human-readable render path does not raise an unstructured `RecursionError`.
- [x] A unit test exercises a tree at/above the previous failing depth.

Closes #37

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -120,6 +120,7 @@ operation, and parse codes the CLI assigns).
 | `already_connected` | `operation` | `operation` | `4` | A signal is already connected to the target node's method. |
 | `connection_not_found` | `operation` | `operation` | `4` | A requested signal-to-method connection does not exist on the source node. |
 | `contract_violation` | `parse` | `parser` | `5` | The process claimed success but violated the structured-output contract. |
+| `tree_too_deep` | `parse` | `classifier` | `6` | The engine emitted a valid result tree that nests past gda's recursion limit; the payload is contract-conformant, the limit is wrapper-side. |
 
 ## Considered options
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -120,7 +120,7 @@ operation, and parse codes the CLI assigns).
 | `already_connected` | `operation` | `operation` | `4` | A signal is already connected to the target node's method. |
 | `connection_not_found` | `operation` | `operation` | `4` | A requested signal-to-method connection does not exist on the source node. |
 | `contract_violation` | `parse` | `parser` | `5` | The process claimed success but violated the structured-output contract. |
-| `tree_too_deep` | `parse` | `classifier` | `6` | The engine emitted a valid result tree that nests past gda's recursion limit; the payload is contract-conformant, the limit is wrapper-side. |
+| `tree_too_deep` | `parse` | `classifier` | `5` | The engine emitted a valid result tree that nests past gda's recursion limit; the payload is contract-conformant, the limit is wrapper-side (shares the `parse` exit code; the `code` distinguishes it from `contract_violation`). |
 
 ## Considered options
 
--- a/src/gda/error_codes.py
+++ b/src/gda/error_codes.py
@@ -14,6 +14,7 @@ from gda.exit_codes import (
     EXIT_OPERATION,
     EXIT_PARSE,
     EXIT_TIMEOUT,
+    EXIT_TREE_TOO_DEEP,
     EXIT_VERSION,
 )
 from gda.models import ErrorCategory
@@ -306,6 +307,14 @@ ERROR_CODES: tuple[ErrorCodeSpec, ...] = (
         EXIT_PARSE,
         ErrorCodeSource.PARSER,
         "The process claimed success but violated the structured-output contract.",
+    ),
+    ErrorCodeSpec(
+        "tree_too_deep",
+        ErrorCategory.PARSE,
+        EXIT_TREE_TOO_DEEP,
+        ErrorCodeSource.CLASSIFIER,
+        "The engine emitted a valid result tree that nests past gda's recursion"
+        " limit; the payload is contract-conformant, the limit is wrapper-side.",
     ),
 )
 
--- a/src/gda/error_codes.py
+++ b/src/gda/error_codes.py
@@ -14,7 +14,6 @@ from gda.exit_codes import (
     EXIT_OPERATION,
     EXIT_PARSE,
     EXIT_TIMEOUT,
-    EXIT_TREE_TOO_DEEP,
     EXIT_VERSION,
 )
 from gda.models import ErrorCategory
@@ -311,7 +310,7 @@ ERROR_CODES: tuple[ErrorCodeSpec, ...] = (
     ErrorCodeSpec(
         "tree_too_deep",
         ErrorCategory.PARSE,
-        EXIT_TREE_TOO_DEEP,
+        EXIT_PARSE,
         ErrorCodeSource.CLASSIFIER,
         "The engine emitted a valid result tree that nests past gda's recursion"
         " limit; the payload is contract-conformant, the limit is wrapper-side.",
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -95,6 +95,20 @@ def _failure(code: str, message: str, stderr: str) -> Failure:
 M = TypeVar("M", bound=BaseModel)
 
 
+def _is_too_deep(exc: ValidationError) -> bool:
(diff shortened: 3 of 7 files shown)


qinhaihong-red:
# PR #156 Review

Reviewed from GitHub PR metadata and diff only, using PR base `main` as the fixed point.

Diff source: `gh pr diff 156 --patch`
Commits source: `gh pr view 156 --json commits`
Spec source: PR body plus closing issue #37 (`gh issue view 37 --json number,title,body,comments,state,labels,url`)
Standards sources: `AGENTS.md`, `RULES.md`, `CONTEXT.md`, `docs/agents/*.md`, relevant ADRs (`docs/adr/0002-headless-structured-output-contract.md`, `docs/adr/0004-schema-flag-self-description.md`), and `pyproject.toml` as machine-enforced context.

## Standards

No hard documented-standard violation found.

Consider: `src/gda/exit_codes.py` says "Exit codes stay at category granularity; finer detail is the `code` field's job", while ADR-0002 and `src/gda/error_codes.py` frame each `GdaError.code` row as the public ABI and explicitly say exit codes are carried per code, not simply per category. This PR's actual registry choice is internally consistent (`tree_too_deep` and `contract_violation` both map to `EXIT_PARSE=5`), but the prose can be read as weakening the per-code ABI rule. I would tighten that sentence to say the parse-category codes intentionally share exit 5 here, while the registry row remains authoritative. This is a judgement call, not a blocker.

## Spec

No missing, partial, or incorrect requirement found against issue #37.

Issue #37 asks that a deep-but-valid scene tree either succeed or fail with an accurate distinct structured error, not `parse / contract_violation`; the diff adds `tree_too_deep` in the Python error registry and routes recursion-depth-only validation failures there. The issue also asks that the human-readable render path not raise an unstructured `RecursionError`; `render_node_tree` now uses an explicit stack. The issue asks for a unit test at or above the previous failing depth; the PR adds classifier coverage at depth 1000 and renderer coverage at depth 2000 for both `SceneNode` and `ListedNode`.

Summary: Standards found 0 hard violations and 1 non-blocking wording concern; Spec found 0 issues. Worst issue: non-blocking exit-code wording clarity in `src/gda/exit_codes.py`.

## code-review-and-quality

### Context

This change addresses deep scene trees that exceed pydantic-core's recursive validation ceiling. It introduces a distinct `tree_too_deep` classifier error while keeping the public process exit code at `EXIT_PARSE=5`, and it rewrites human tree rendering to avoid recursive calls.

### Tests Reviewed First

The new tests cover the core behavior before the implementation details:

- `tests/test_classify_run.py` builds a valid 1000-deep `SceneGetResult` payload and asserts it becomes `tree_too_deep`, not `contract_violation`.
- The same test file checks that the message describes a wrapper-side depth/materialization limit and that `tree_too_deep` shares `EXIT_PARSE`.
- `tests/test_render.py` constructs 2000-deep `SceneNode` and `ListedNode` chains with `model_construct` to isolate the renderer from pydantic validation and verify the iterative renderer produces one line per node.

These tests are behavior-focused and would catch regressions in both the classifier mapping and the render recursion fix.

### Correctness

No required correctness issue found. The implementation only maps a `ValidationError` to `tree_too_deep` when every pydantic error has type `recursion_loop`, so mixed shape errors still fall through to `contract_violation`. Because `classify_run` starts from parsed JSON, genuine Python object cycles are not a realistic payload source here, which makes the discriminator reasonable for this boundary.

### Readability / Simplicity

Nit: The new renderer comment says "pre-order, in-order traversal". For a general tree this is just pre-order left-to-right traversal; "in-order" is binary-tree terminology and slightly muddies an otherwise clear comment.

### Architecture

No required architecture issue found. The new code follows the existing split: `src/gda/error_codes.py` remains the authoritative error registry, `src/gda/errors.py` owns classification, ADR-0002 mirrors the public error table, and `src/gda/render.py` keeps presentation logic out of the data models. Not mirroring `tree_too_deep` in GDScript is consistent with it being classifier-sourced rather than operation-reported.

### Security

No security issue found. The change does not parse stderr for stable outcomes, does not introduce new command execution, and continues treating the structured stdout payload as data validated at the Python boundary.

### Performance

No performance blocker found. The renderer moves from recursive calls to an explicit stack, removing call-stack depth risk. It still materializes the full rendered output in a list before joining, but that matches the existing function's output contract and is acceptable for this slice.

### Verification

I did not checkout the branch or run local tests, per the review constraint to prefer no checkout. GitHub reports `Python unit tests and package build` passing for PR #156. The `Godot e2e tests` check is currently skipped. The PR body also reports local full-suite verification, including real-Godot e2e, but I did not independently rerun that.

### Verdict

Approve. I found no required changes. The only comments are non-blocking documentation/comment clarity items.

Issue #121, opened 2026-06-13
gda export run: run an export preset via the headless CLI


qinhaihong-red:
## Parent

#17 — PRD: gda domain command surface — Phase-1 headless increment.

## What to build

`gda export run` exports the project via the headless CLI for a named preset, producing artifacts at the preset's configured output path, and reports the result (output path, warnings) as typed JSON. Builds on the export discovery from #114 (an agent checks template-install status via `export get` before running).

## Acceptance criteria

- [ ] `gda export run` exports a given preset to its output path and reports the result (output path, warnings) as typed JSON.
- [ ] Ships `--json` and the ADR-0004 `--schema` hard gate via the shared skeleton.
- [ ] Failure modes surface as structured `GdaError`s with registered codes. The preset-resolution failure (unknown preset) is the operation-source `export_preset_not_found` from the phase-1 `export get` op (Python registry + ADR-0002 table + GDScript mirror). The export-execution failures (`export_templates_missing`, `export_failed`, `export_path_unset`) are **classifier-source**, because `export run` executes via the engine's native CLI mode rather than an `operations.gd` op (ADR-0010): registered in the Python registry + ADR-0002 table and, per ADR-0002, **not** GDScript-mirrored (only operation-source codes are mirrored; the drift test enforces this). _(Narrowed from the original blanket "+ GDScript mirror" — see PR #169 review.)_
- [ ] Tests: S2 model/parser unit, S3 CLI-with-FakeRunner, S1 e2e (e2e may require installed export templates; auto-skip / flag clearly when absent, consistent with the e2e engine-presence policy).

## Blocked by

- #114 (`export list` + `export get`: preset discovery + template-install status).

PR #172, opened 2026-06-16
fix(runner): map launch failures + non-UTF-8 output to structured errors (#33)

### LINKED ISSUE #33: runner: map all process-launch failures and non-UTF-8 output to structured errors
## Parent

#17 — PRD: gda domain command surface. Found during the high-effort code review of #18's implementation (merged via #41).

## What to build

Several engine-launch failure modes escape the classification layer as raw Python tracebacks (exit 1, empty stdout, **no \`{"error": …}\` envelope**) — violating #18's acceptance that every command failure maps to a structured error with a registry exit code. All verified:

- \`--godot ""\` → bare \`ValueError\` from binary resolution escapes (rich traceback).
- \`--godot\` pointing at a directory (e.g. \`Godot.app\`, a natural \`$GDA_GODOT\` mistake) or a non-executable file → \`PermissionError\` escapes; the runner only catches \`TimeoutExpired\` and \`FileNotFoundError\`.
- \`subprocess.run(text=True)\` decodes engine stdout with the **locale** encoding, but Godot's \`JSON.stringify\` emits raw UTF-8. \`scene get\` is the first op whose payload carries user content (node names, echoed paths), so on a non-UTF-8 locale (Windows cp1252/cp936) a non-ASCII node name either raises \`UnicodeDecodeError\` (uncaught) or silently mojibakes the names in the result.

Make the runner/binary layer map all of these to the existing structured error taxonomy.

## Acceptance criteria

- [ ] \`--godot ""\`, \`--godot <directory>\`, and \`--godot <non-executable file>\` each produce a structured \`{"error": …}\` envelope with an environment-category code and a registry exit code — no raw traceback.
- [ ] Engine output is decoded as UTF-8 explicitly, so a scene with a non-ASCII node name round-trips correctly regardless of host locale.
- [ ] Unit tests cover the empty/directory/non-executable binary paths and a non-ASCII node-name decode.

## Blocked by

None - can start immediately.

qinhaihong-red:
Closes #33

## What

Three engine-launch failure modes escaped `gda`'s classification layer as raw Python tracebacks (exit 1, no `{"error": …}` envelope), violating ADR-0002's contract that every command failure maps to a structured `GdaError` with a registry exit code. This maps all three into the existing environment-category taxonomy.

- **`--godot <directory>` / `--godot <non-executable file>`** — `subprocess.run` raised an uncaught `PermissionError`; the runner only caught `TimeoutExpired` and `FileNotFoundError`. Broadened the catch to `OSError` (covers missing / directory / non-executable / any `exec` failure) → `LaunchFailure.NOT_FOUND`. `TimeoutExpired` is a `SubprocessError`, not an `OSError`, so the timeout path is preserved.
- **`--godot ""` (empty)** — binary resolution raises `ValueError` *before* a runner is built. `HeadlessCommand.run` now catches it and emits `binary_not_found` through the public failure channel (new `errors.unresolvable_binary_failure` helper).
- **Non-UTF-8 host locale** — `subprocess.run(text=True)` decoded engine stdout with the locale encoding, but Godot's `JSON.stringify` emits raw UTF-8; a non-ASCII node name mojibaked or raised `UnicodeDecodeError` on a non-UTF-8 locale (cp1252/cp936). The runner now captures **bytes** and decodes UTF-8 explicitly (`errors="replace"`: never crash on engine output; a malformed result then surfaces as a structured `contract_violation` downstream).

## Error code

All three reuse the existing **runner-source `binary_not_found`** code (exit 127) — no new code minted (ADR-0002: reuse the exit code, discriminate via the envelope). `error_codes.py`, the ADR-0002 table, and `operations.gd` are therefore untouched.

## Tests

TDD red→green. New unit tests (no real engine needed):

- `test_runner.py`: directory binary, non-executable file → `LaunchFailure.NOT_FOUND` / exit 127; engine output decoded as UTF-8 round-trips a `日本語` node name + stderr regardless of locale (driven by returning raw UTF-8 bytes from a fake subprocess).
- `test_headless_command.py`: `--godot ""` → structured `binary_not_found` envelope (exit 127), no runner ever built.

Updated the pre-existing `test_timeout_is_passed_through_to_subprocess` fake to return bytes, matching the real (no-`text=True`) subprocess contract.

## How verified

`uv run pytest` full suite green (unit + real-engine e2e). Each new test was confirmed failing before the fix and passing after.

### DIFF
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -125,6 +125,24 @@ def _operation_error_from_payload(result: RunResult) -> tuple[str, str] | None:
     return envelope.error.code, envelope.error.message
 
 
+def unresolvable_binary_failure(reason: str) -> Failure:
+    """The ``binary_not_found`` failure when the binary cannot even be resolved (issue #33).
+
+    Binary resolution runs *before* a runner is built, so an unresolvable
+    ``--godot`` value (an empty ``--godot ""`` / ``$GDA_GODOT`` mistake) raises
+    instead of producing a launchable path. There is no engine to run — the same
+    environment outcome the runner reports as ``LaunchFailure.NOT_FOUND`` — so it
+    reuses the ``binary_not_found`` code rather than minting a new one (ADR-0002:
+    reuse the exit code; discriminate via the envelope). Kept here beside the
+    other environment failures so the whole taxonomy reads from one place.
+    """
+    return _failure(
+        "binary_not_found",
+        f"Godot binary could not be resolved: {reason}",
+        "",
+    )
+
+
 def classify_run(result: RunResult, binary: Path, output_model: type[M]) -> M | Failure:
     """Classify a raw headless run into the command's typed model or a ``Failure``.
 
--- a/src/gda/export_runner.py
+++ b/src/gda/export_runner.py
@@ -92,8 +92,12 @@ class SubprocessExportRunner:
             cmd += ["--path", str(self.project)]
         cmd += [flag, preset, output_path]
         try:
+            # Capture raw bytes (no ``text=True``): like the sentinel runner,
+            # the native export channel must not decode with the host locale,
+            # which mojibakes or raises ``UnicodeDecodeError`` on a non-UTF-8
+            # locale. We decode UTF-8 explicitly below (issue #33).
             proc = subprocess.run(
-                cmd, capture_output=True, text=True, timeout=self.timeout, cwd=cwd
+                cmd, capture_output=True, timeout=self.timeout, cwd=cwd
             )
         except subprocess.TimeoutExpired:
             return ExportRunOutput(
@@ -102,15 +106,29 @@ class SubprocessExportRunner:
                 exit_code=EXIT_TIMEOUT,
                 launch_failure=LaunchFailure.TIMEOUT,
             )
-        except FileNotFoundError:
+        except OSError as exc:
+            # The configured binary could not be launched: ``FileNotFoundError``
+            # (missing), ``PermissionError`` (a directory like ``Godot.app`` or a
+            # non-executable file), and any other ``OSError`` from ``exec`` are
+            # one environment failure — there was no engine to run (issue #33).
+            # Mirrors ``SubprocessGodotRunner`` so both runners close the same
+            # raw-launch-failure surface. ``OSError`` does not subsume
+            # ``TimeoutExpired`` (a ``SubprocessError``), so the timeout path
+            # above is preserved.
             return ExportRunOutput(
                 stdout="",
-                stderr=f"gda: Godot binary not found: {self.binary}\n",
+                stderr=f"gda: Godot binary could not be launched: {self.binary} ({exc})\n",
                 exit_code=EXIT_NOT_FOUND,
                 launch_failure=LaunchFailure.NOT_FOUND,
             )
         return ExportRunOutput(
-            stdout=proc.stdout, stderr=proc.stderr, exit_code=proc.returncode
+            # Decode the engine's bytes as UTF-8 with a replacement policy, like
+            # the sentinel runner: a well-behaved export emits valid UTF-8, so
+            # ``replace`` only fires on genuinely malformed bytes and the runner
+            # never crashes on engine output (issue #33).
+            stdout=proc.stdout.decode("utf-8", errors="replace"),
+            stderr=proc.stderr.decode("utf-8", errors="replace"),
+            exit_code=proc.returncode,
         )
 
 
--- a/src/gda/headless.py
+++ b/src/gda/headless.py
@@ -18,7 +18,7 @@ from pydantic import BaseModel
 from typer.core import TyperCommand
 
 from gda.binary import resolve_godot_binary
-from gda.errors import Failure, classify_run
+from gda.errors import Failure, classify_run, unresolvable_binary_failure
 from gda.models import CommandSchema, GdaErrorEnvelope
 from gda.runner import GodotRunner, RunResult, SubprocessGodotRunner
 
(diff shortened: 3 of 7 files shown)


qinhaihong-red:

Reviewed PR #172 against `main` using `gh pr diff 172 --patch --color never`.

## Standards

No hard documented-standards violations found in the changed hunks.

Judgment call: `src/gda/runner.py` now hardens `SubprocessGodotRunner.run` for broad `OSError` launch failures and explicit UTF-8 byte decoding, but the native export runner remains unchanged with `subprocess.run(..., text=True)` and only `FileNotFoundError` handling. ADR-0010 says to "Minimise the divergence" and keep as much as possible on the default rails (`docs/adr/0010-headless-operation-execution-mechanism.md`), and says the native runner reuses the existing runner / `GdaError` machinery with the same agent-facing contract. Because `src/gda/export_runner.py` is outside this PR's diff, I would treat this primarily as a spec coverage gap rather than a standards-only violation, but the divergence is real.

No ADR-0002 registry issue found: the PR reuses the existing `binary_not_found` code, whose registered meaning is "The Godot binary could not be launched"; ADR-0002 only requires registry/ADR/GDScript mirror updates when adding or changing codes.

## Spec

Required: PR #172 does not cover the export runner path. Issue #33 says: "Make the runner/binary layer map all of these to the existing structured error taxonomy." `SubprocessExportRunner` is also a Godot engine-launch runner and still catches only `FileNotFoundError`; directory/non-executable binaries can still escape as raw `PermissionError`. I reproduced this with `SubprocessExportRunner(Path(tempdir)).run(...)`, which raises `PermissionError` instead of returning a launch failure for classification.

Required: export output still uses locale decoding. Issue #33 says: "Engine output is decoded as UTF-8 explicitly..." but `src/gda/export_runner.py` still calls `subprocess.run(..., text=True)`, leaving the native export channel exposed to host-locale decoding behavior.

Potential / clarify: `src/gda/runner.py` decodes with `errors="replace"`. This is probably acceptable for the issue's main requirement, because the body describes Godot's valid UTF-8 output being decoded with the wrong host locale, not malformed engine bytes. It would still be worth confirming that replacement-on-malformed-bytes is the intended behavior if a malformed-but-parseable JSON payload appears.

Summary: Standards found 0 hard violations and 1 judgment call. Spec found 2 required gaps and 1 clarification. Worst issue: `export run` still retains the same raw launch-failure and locale-decoding surface that #33 is meant to close.

qinhaihong-red:

### Findings

The `export run` native runner still has the same failure surface this PR closes for the default runner. `src/gda/export_runner.py:95-113` still calls `subprocess.run(..., text=True, ...)` and only catches `FileNotFoundError`; `PermissionError` from a directory or non-executable `--godot` value still escapes as a raw traceback instead of becoming `LaunchFailure.NOT_FOUND` -> `binary_not_found`. I reproduced the directory case locally against the current code path. Because issue #33 is framed as "the runner/binary layer" and ADR-0010 says the native export runner mirrors `classify_run`'s environment/crash prefix, this should be fixed in the same PR or the issue scope should be narrowed explicitly.

The same `export_runner.py` path also still uses locale decoding via `text=True`. That leaves native export stdout/stderr on the old host-locale behavior, while this PR's main runner correctly captures bytes and decodes UTF-8 explicitly. Even if export output does not currently carry `scene get` node names, it is still a Godot engine output path under the same agent-facing failure contract.

### Other Axes

Correctness: the changed `SubprocessGodotRunner` and `HeadlessCommand` paths look coherent for `--godot ""`, directory binaries, non-executable binaries, timeout preservation, and UTF-8 output from the sentinel runner.

Readability / architecture: the new helper and runner changes fit the existing `Failure` / `LaunchFailure` taxonomy and reuse the existing `binary_not_found` registry row instead of minting a new code.

Security: no new secret handling, injection surface, or trust-boundary change found.

Performance: no material performance issue found; byte capture was already what `subprocess` does underneath.

Verification: GitHub shows Python unit tests/package build passing and Godot e2e skipped. I did not run the PR-head full suite locally; I only ran the targeted export-runner reproduction described above.

### Verdict

Request changes: please harden `SubprocessExportRunner` the same way, with coverage for directory/non-executable binary launch failures and byte/UTF-8 decoding, or explicitly split/narrow #33 so `export run` is tracked by a follow-up issue.

qinhaihong-red:
## Resolution — hardened the native export runner too (commit `248c096`)

Good catch — the gap is real and now fixed in this PR rather than deferred. `SubprocessExportRunner` (the second engine-launch runner, added in #121) carried exactly the surface #33 closes, and #33 is framed as "the runner/binary layer ... map **all** of these", so it belongs here. ADR-0010 also mandates the native runner reuse the same env/crash classification — so this brings it into compliance, not scope creep.

**Both required findings fixed**, mirroring `SubprocessGodotRunner` 1:1:
1. **Launch failures** — `SubprocessExportRunner.run` now catches `OSError` (not just `FileNotFoundError`), so a directory (`Godot.app`) or non-executable `--godot` becomes `LaunchFailure.NOT_FOUND` → `binary_not_found` instead of an escaping `PermissionError`. `TimeoutExpired` (a `SubprocessError`, not `OSError`) stays on its own path.
2. **UTF-8 decode** — dropped `text=True`; captures bytes and decodes UTF-8 explicitly (`errors="replace"`), so native export output no longer rides host-locale decoding.

New tests in `tests/test_export_runner.py` cover the directory and non-executable cases (→ `binary_not_found`) and the byte/UTF-8 decode (asserts `text` is not set), mirroring the sentinel-runner tests; the existing `_RecordingRun` double was updated to return bytes to match the real subprocess contract.

**On the `errors="replace"` clarification:** intended, and now consistent across both runners. Godot emits valid UTF-8, so `replace` only fires on genuinely malformed bytes — in which case the runner does not crash; the malformed payload then fails to parse and surfaces as a structured `contract_violation` downstream (ADR-0002), rather than an escaping `UnicodeDecodeError`. We prefer a structured failure over a traceback for unexpected engine bytes.

### Verification
Orchestrator independently re-ran the **full suite incl. e2e**: `uv run pytest` → **707 passed, 1 skipped** (the skip = the templates-absent configured-path export e2e). `operations.gd` / `error_codes.py` / ADR-0002 remain untouched — both runners reuse the existing `binary_not_found` row.

PR #174, opened 2026-06-16
feat(export): run — add --mode + --output overrides (#170)

### LINKED ISSUE #170: gda export run: --mode (release/debug/pack) + --output override
## Parent

#17 — PRD: gda domain command surface — Phase-1 headless increment. Follow-up de-scoped from #121 / PR #169.

## What to build

`gda export run` (#121) exports a named preset in **release** mode to the preset's **configured** output path. This rounds it out with two optional overrides that were de-scoped from #121 to keep that slice to its requested surface:

- `--mode release|debug|pack` — select the export mode (Godot's native `--export-release` / `--export-debug` / `--export-pack`); default `release`.
- `--output PATH` — export to `PATH` instead of the preset's configured `export_path`. A filesystem path normalized once at the CLI layer (ADR-0006: `~` expanded), like every other path-taking command.

The reported result keeps the same shape (preset, platform, mode, output_path, warnings): `mode` reflects the selected mode and `output_path` reflects the effective destination.

## Acceptance criteria

- [ ] `--mode release|debug|pack` selects the export mode (default `release`), reflected in both the native invocation and the result's `mode`.
- [ ] `--output PATH` overrides the configured `export_path`; the result's `output_path` reports the effective destination.
- [ ] `--output` is normalized at the CLI layer per ADR-0006 (`~` expanded), covered by a literal-tilde test.
- [ ] Both options ship via `--json` and the ADR-0004 `--schema` hard gate through the shared skeleton.
- [ ] Tests: S2 model, S3 CLI-with-FakeRunner, S1 e2e (the export-success e2e may auto-skip when export templates are absent, consistent with the e2e template-presence policy).

## Blocked by

- #121 (export run base — named preset → configured path, release mode; PR #169).

qinhaihong-red:
## What

Rounds out `gda export run` (#121) with the two optional overrides de-scoped from that slice:

- `--mode release|debug|pack` — selects the export flavor (maps to Godot's native `--export-release`/`--export-debug`/`--export-pack`). A closed Typer `Enum`, so an unknown value is a usage error (exit 2) before reaching the runner. Default `release`, preserving #121's behavior when omitted.
- `--output PATH` — overrides the preset's configured `export_path`. Normalized **once** at the CLI layer via `_normalize_path` (ADR-0006: `~` expanded), like every other path-taking command.

The result keeps its shape: `mode` reflects the selected flavor; `output_path` reports the effective destination (override when given, else the configured path).

## How

- `ExportRunParams` gains `mode` (default `release`) and `output` (default `None`) so the ADR-0004 `--schema` gate advertises both.
- Changes are localized to the `run_export` command body: `mode`/`output_path` were already the locals the native invocation + result consume, so this only changes where they come from. The `export_path_unset` preflight now fires only when there is neither an override nor a configured path.
- No shared registries, enums, dispatch tables, or other commands touched.

## Tests (TDD: failing tests first, then implementation)

- **S2** (`test_models.py`): mode round-trip per flavor; result reports the overridden `output_path`.
- **S3** (`test_export_run_commands.py`, `FakeExportRunner`): default mode is release; mode selects the flavor; unknown mode rejected (exit 2); output overrides configured path; **leading-`~` expansion** (ADR-0006); output overrides even an unset configured path; `--schema` now advertises `mode` + `output`.
- **S1 e2e** (`test_e2e_export_run.py`): `--mode pack --output …` writes to the overridden path, following the existing template-presence auto-skip policy.

## Results & known limitation

- Engine-free full suite: **489 passed**. `--schema` ADR-0004 gate verified (emits `mode`/`output` in `input.properties`, no engine spawned).
- e2e on the real engine: the two **success-path** e2e tests (configured-path and the new `--output`/`--mode pack` override) **auto-skip** because this machine has no export templates for the running engine version installed — each verifies the structured `export_templates_missing` preflight live before skipping. So the on-disk-artifact assertions could not be exercised here; reviewer with templates installed can confirm.

Closes #170.

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -122,8 +122,8 @@ operation, and parse codes the CLI assigns).
 | `invalid_resource_type` | `operation` | `operation` | `4` | A requested resource type cannot be instantiated as a `Resource`. |
 | `export_presets_not_found` | `operation` | `operation` | `4` | The project has no export_presets.cfg, so it defines no export presets. |
 | `export_preset_not_found` | `operation` | `operation` | `4` | No export preset with the requested name exists in export_presets.cfg. |
-| `export_path_unset` | `operation` | `classifier` | `4` | An export run targeted a preset that has no configured export_path. |
-| `export_templates_missing` | `operation` | `classifier` | `4` | An export run needs the export templates for the running engine version, which are not installed. |
+| `export_path_unset` | `operation` | `classifier` | `4` | An export run has no destination — neither a `--output` override nor a configured `export_path` (#170). |
+| `export_templates_missing` | `operation` | `classifier` | `4` | A release/debug export needs the export templates for the running engine version, which are not installed (pack needs no platform templates and is exempt; #170). |
 | `export_failed` | `operation` | `classifier` | `4` | A native Godot export run failed (the engine reported the export did not complete). |
 | `invalid_uid` | `operation` | `operation` | `4` | A requested `uid://` value is not a syntactically valid resource UID. |
 | `unknown_uid` | `operation` | `operation` | `4` | A syntactically valid resource UID is not registered in the engine's UID cache. |
--- a/docs/adr/0010-headless-operation-execution-mechanism.md
+++ b/docs/adr/0010-headless-operation-execution-mechanism.md
@@ -58,11 +58,15 @@ native run:
 - **Structured preflight (classifier, no native run).** `export-get` already reports
   the preset's configured `export_path` and, structurally, its template readiness
   (`templates_installed` — the readiness field built for exactly this check). From
-  those: an empty configured `export_path` is `export_path_unset`, and uninstalled
-  templates for the running engine version are `export_templates_missing`. Both are
-  decided here, from structured fields, with **no** native export spawned — in
-  particular `export_templates_missing` is *not* inferred from the engine's "due to
-  configuration errors" stderr (which would violate ADR-0002 and also misfires for a
+  those: no **effective destination** — neither a `--output` override (#170) nor a
+  configured `export_path` — is `export_path_unset`; and uninstalled templates for the
+  running engine version are `export_templates_missing`. The destination check applies
+  to **every** mode, but the template check is for **release/debug only** — `pack`
+  produces project data (a PCK/ZIP via Godot's native `--export-pack`) and needs no
+  platform templates, so it is exempt (#170). Both checks are decided here, from
+  structured fields, with **no** native export spawned — in particular
+  `export_templates_missing` is *not* inferred from the engine's "due to configuration
+  errors" stderr (which would violate ADR-0002 and also misfires for a
   merely-misconfigured preset).
 - **Native run (classifier on exit code).** Only the export execution itself uses the
   native CLI mode. A clean exit is the typed success result (with any advisory
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -1625,31 +1625,47 @@ def run_export(
         "--preset",
         help="The export preset's display name, as 'gda export list' reports it.",
     ),
-    # NOTE: --mode (release/debug/pack) and --output (path override) are
-    # deferred to follow-up issue #170. Issue #121 asks only to export a named
-    # preset in RELEASE mode to its CONFIGURED export_path, so this command
-    # exposes neither flag; the result still carries `mode` (always "release")
-    # to document the mode that ran.
+    # --mode (#170): select the export flavor. A closed Enum so an unrecognized
+    # value is a Typer usage error (exit 2) rather than reaching the runner;
+    # release is the default, preserving #121's behavior when --mode is omitted.
+    mode: ExportRunMode = typer.Option(
+        ExportRunMode.RELEASE,
+        "--mode",
+        help="The export flavor to run (release/debug/pack); default release.",
+    ),
+    # --output (#170): override the preset's configured export_path. A filesystem
+    # path normalized ONCE at the CLI layer (ADR-0006: ~ expanded), like every
+    # other path-taking command.
+    output: Optional[str] = typer.Option(
+        None,
+        "--output",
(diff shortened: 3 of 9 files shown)


qinhaihong-red:
## Review Skill Pass (`review`)

### Standards

Required:

- `src/gda/cli.py` and the new tests change the `export_path_unset` semantics so `--output` bypasses the failure when the preset's configured `export_path` is empty. That matches issue #170's requested override behavior, but the accepted docs still say otherwise: ADR-0010 defines the structured preflight as "an empty configured `export_path` is `export_path_unset`", and ADR-0002 defines `export_path_unset` as "An export run targeted a preset that has no configured export_path." If #170 intentionally redefines this from configured destination to effective destination, please update the ADR/error-code documentation in the same PR so the public contract and implementation stay synchronized.

No other standards findings. The schema remains model-driven, `--output` is normalized at the CLI layer, and native export failure classification still avoids parsing stderr for stable error codes.

### Spec

Required:

- `src/gda/cli.py:1704` still performs an unconditional `if not got.templates_installed` preflight before the native export runs. That blocks `--mode pack` in projects without export templates, even though issue #170 says `--mode release|debug|pack` selects Godot's native `--export-release` / `--export-debug` / `--export-pack`. Godot's `--export-pack` exports project data only and does not need platform templates; I confirmed locally with Godot 4.6.3 that a minimal `--export-pack` invocation produced a `.pck` without installed export templates. The new e2e at `tests/test_e2e_export_run.py:195` instead expects `export_templates_missing` and skips in exactly this case, so it codifies the wrong behavior rather than verifying the required pack path.

Summary: 1 Standards finding, 1 Spec finding. Worst issue: `--mode pack` is not fully implemented because template-less projects never reach Godot's native pack export.

qinhaihong-red:
## Code Review and Quality Pass

Verdict: Request changes.

### Correctness

Required:

- `src/gda/cli.py:1704` applies the missing-template preflight to every export mode. That is incorrect for `pack`: `ExportRunMode` documents that `pack` produces project data only and needs no platform templates, and Godot's native `--export-pack` path does not perform the same platform-template check as release/debug exports. I verified this locally with Godot 4.6.3: a minimal direct `Godot --headless --export-pack Linux/X11 dist/packed.pck` run produced a `.pck` without installed export templates. The PR head currently returns `export_templates_missing` for the same `gda export run --mode pack --output dist/packed.pck` scenario and writes no artifact, so template-less environments cannot use the mode that should work there.

- `tests/test_e2e_export_run.py:195` reinforces the bug by expecting `export_templates_missing` and skipping when templates are absent. The regression test should instead prove that `--mode pack --output ...` reaches the native pack export and writes the artifact without templates. Keep the template preflight for `release`/`debug`; do not apply it to `pack`.

### Other Axes

- Readability: The change is localized and generally easy to follow, but the new comments around the e2e test and preflight now encode the wrong `pack` template requirement.
- Architecture: Reuses the existing enum/model/schema/export-runner structure and adds no new dependency. No additional architecture concern beyond the contract/docs sync called out in the separate `review` pass.
- Security: No new secret handling, auth, injection, or untrusted external-data flow beyond the existing trusted-project export surface.
- Performance: No new unbounded work or hot-path overhead.

### Verification

I ran:

- `uv --cache-dir /private/tmp/gda-uv-cache run pytest tests/test_models.py tests/test_export_run_commands.py tests/test_classify_export_run.py -q` -> 73 passed
- `env HOME=/private/tmp/gda-e2e-home GDA_GODOT=/Users/lelej/Applications/Godot.app/Contents/MacOS/Godot uv --cache-dir /private/tmp/gda-uv-cache run pytest tests/test_e2e_export_run.py -q -m e2e` -> 2 passed, 2 skipped
- `env GDA_GODOT=/Users/lelej/Applications/Godot.app/Contents/MacOS/Godot uv --cache-dir /private/tmp/gda-uv-cache run pytest -m 'not e2e' -q` -> 489 passed, 227 deselected
- Manual repro: direct Godot `--export-pack` produced a `.pck` without templates, while PR-head `gda export run --mode pack --output ...` returned `export_templates_missing`.

qinhaihong-red:
Both findings addressed in `eb0b9f6` (thanks for the careful pack-mode catch and the local repro).

**Correctness/Spec — `--mode pack` blocked by the template preflight:** the preflight is now gated `if mode is not ExportRunMode.PACK and not got.templates_installed:`. `pack` produces project data only (native `--export-pack`) and needs no platform templates, so it is exempt; `export_path_unset` still applies to every mode. 
- The e2e (`tests/test_e2e_export_run.py`) is rewritten: `test_export_run_pack_writes_pck_without_templates` runs `--mode pack --output dist/packed.pck` against the real engine and asserts **exit 0 + the `.pck` on disk**, with **no skip** — the on-disk verification a release export can't give here. Confirmed this machine is genuinely template-less: the release test `test_export_run_writes_to_configured_export_path` still auto-skips, so pack provably ran without templates. (Fixture needed a trivial `main.gd` so `all_resources` has content, and empty `include_filter`/`exclude_filter` keys to clear config noise.)
- Added two S3 guards (`test_export_run_commands.py`): with `templates_installed=False`, `--mode pack` proceeds to the runner (exit 0, native call recorded); `--mode release`/`debug` still fail fast with `export_templates_missing` (exit 4, no native run).

**Standards — docs sync:** `--output` changed `export_path_unset` from "configured path empty" to "no effective destination", and pack now skips the template preflight. Updated to match (ADRs are amended in place here — ADR-0007 uses inline supersede notes, the ADR-0002 table already carries inline issue refs):
- ADR-0002 table: `export_path_unset` → "no destination — neither a `--output` override nor a configured `export_path`"; `export_templates_missing` → "release/debug … pack needs no platform templates and is exempt". Both tagged `(#170)`.
- ADR-0010 "Structured preflight" bullet: destination check is every-mode; template check is release/debug only, pack exempt.
- `export_path_unset_failure` message: `"has no configured export_path"` → `"has no destination: pass --output or set the preset's export_path"` (builder test updated).

Engine-free suite: **491 passed** (+2 from the new guards). Export-run e2e: **3 passed, 1 skipped** (pack passes with the on-disk `.pck`; release skips, templates absent).

Issue #136, opened 2026-06-15
script attach: distinguish node/script type incompatibility from compile failure


qinhaihong-red:
## Parent

#118 (PRD #17). Raised in the #129 review (maintainer-reproduced) and independently confirmed against Godot 4.6.3.

## What to build

`gda script attach`'s post-bind check reports **any** `node.get_script() == null` after `set_script` as `script_compile_failed`. But `set_script` also silently rejects a script that **compiles fine** yet whose native base type is **incompatible** with the target node — e.g. an `extends Node3D` script attached to a `Node2D` node, where the engine logs: `Script inherits from native type 'Node3D', so it can't be assigned to an object of type 'Node2D'.` Reporting that as `script_compile_failed` gives an agent the wrong remediation (it would try to fix syntax, but the script is valid — the target node type is wrong).

Distinguish the two failure modes:
- script does **not compile** -> `script_compile_failed` (existing);
- script **compiles** but its base type is **incompatible** with the node -> a new `incompatible_script_type` code, whose message names both the script's native base type and the node's class so the agent knows to attach to a compatible node (or change the script's `extends`).

Detection without extra engine work: `ClassDB.is_parent_class(node.get_class(), script.get_instance_base_type())` is **false** for the incompatible case and **true** for a compatible one (verified: `extends Node3D` -> base `Node3D`, `is_parent_class("Node2D","Node3D")` == false).

## Acceptance criteria

- [ ] A compiling script whose `extends` base is incompatible with the node yields `incompatible_script_type` (not `script_compile_failed`), with a message naming the script base type and the node class.
- [ ] A genuinely non-compiling script still yields `script_compile_failed`.
- [ ] A compatible script still attaches successfully (no regression).
- [ ] The new code is synced across the Python registry, the ADR-0002 table, and the GDScript mirror; S3 + e2e cover both failure modes.

## Blocked by

None - can start immediately.

PR #176, opened 2026-06-16
fix(script): preserve sibling scripts on re-pack in unimported projects (#164)

### LINKED ISSUE #164: script attach: re-attaching in an unimported project can drop a sibling node's script on re-pack
## Context

Discovered while building #58's e2e fixtures (reported in PR #158). Pre-existing, intermittent, and independent of `scene get-exports`. The defect lives in the `script attach` path (#118).

## What to build

`gda script attach` binds a `.gd` to a node, then re-packs and saves the `.tscn`. In a freshly-created project whose `res://` script has **not yet been imported** by the engine (no import metadata / assigned UID), attaching a script to a second node can **intermittently drop a *sibling* node's existing `script =` line** on the re-pack/save — silently removing an unrelated node's script binding and corrupting the scene.

The fix makes `script attach` deterministically preserve every other node's existing script binding on re-pack, regardless of whether the referenced scripts have been imported yet.

## Acceptance criteria

- [ ] Attaching a script to one node never removes or alters another node's existing `script =` binding on the re-pack/save — including in a freshly-created project whose scripts have not yet been imported.
- [ ] A **deterministic** regression test reproduces the original drop (e2e, against the real engine, in an unimported-project fixture) and verifies all sibling script bindings survive.
- [ ] Root cause is documented in the PR (working hypothesis: an unimported script lacks a stable UID / `ext_resource` reference at pack time, so `PackedScene.pack` serializes the scene's script references inconsistently); the fix addresses the cause, not the symptom.
- [ ] No regression to existing `script attach` behavior — displacement reporting (#132), compile/type-incompatibility verdicts (#136, #137), and the registry drift tests stay green.

## Blocked by

None - can start immediately.

qinhaihong-red:
## The bug

`gda script attach` (and any mutating re-pack op) could **intermittently drop a sibling node's `script =` line** when re-packing a `.tscn` in a freshly-created project whose `.gd` scripts have not yet been imported — silently corrupting the scene (the sibling's external script is dropped or re-embedded as an inline `sub_resource`).

## Root cause (confirmed against the Godot 4.6.x source)

The editor build's text scene saver dedups `ext_resource` entries through a **path-keyed** cache (`ResourceCache::resource_path_cache`), not object identity. An **unimported** `.gd` carries no `uid://` (headless never runs the import pass that assigns one), so its `res://` path string is its only identity. The engine's `GDScriptCache` upgrades a shallow script to a full one via `Resource::set_path(take_over=true)`, which **evicts the previously cached object without freeing it** — a sibling node still holds the evicted orphan, so two distinct in-memory `Script` objects share one `res://` path. On save the path-keyed dedup collapses them and the sibling's binding is lost. Intermittent because it depends on whether the second same-path object is alive at save time.

## Fix (gda-layer hardening, no engine patch)

Re-anchor every external script to the **single canonical cached resource** for its `res://` path before `pack()`:

- `_load_for_mutation` calls `_capture_external_scripts(root)` the instant after instantiate — snapshotting `{NodePath → script res:// path}` for every scripted node, **before** any op-specific load can evict a sibling (the orphan loses its `resource_path` on eviction, so the path must be captured while intact).
- `_repack_and_save` calls `_reanchor_external_scripts(root)` before packing: for each captured node still carrying its captured script (current path equals captured, **or empty** — the orphan signature), re-load the path with `CACHE_MODE_REPLACE` (sole cache owner) and `set_script` it back.

**Preserves existing semantics:** a node the op *intentionally* re-scripted to a **different** path is left untouched (`script attach` overwrite-and-report, #132); a vanished node (remove/move) is skipped; already-canonical is idempotent. This protects every mutating re-pack op, not just `script attach`.

## Test — deterministic regression

`tests/test_e2e_script.py::test_script_attach_preserves_sibling_script_on_repack_when_unimported` (`@pytest.mark.e2e`, real engine, unimported `godot_project` fixture). The pure `gda` CLI cannot reproduce the drop deterministically (each call is a fresh process with a clean cache — measured 0 drops in 30+ runs), so the test drives gda's **own** `operations.gd` payload (real `_load_scene` / `_capture_external_scripts` / `_reanchor_external_scripts` / pack / save) and injects the engine precondition deterministically via `take_over_path` (== `set_path(take_over=true)`). `has_method` guards keep the harness loadable against the pre-fix payload, so red→green is on behavior:

- **RED** (fix stashed): node A's `res://a.gd` ext_resource dropped and re-embedded as `sub_resource type="GDScript"` — the exact corruption.
- **GREEN** (fix applied): both nodes keep clean external bindings; 0 sub_resources.

## Results

- Full suite: **708 passed, 1 skipped** (the 1 skip is a pre-existing, unrelated template-presence skip).
- Displacement reporting (#132), compile/type-incompatibility verdicts, and registry-drift tests stay green.

Closes #164.

### DIFF
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -3158,6 +3158,16 @@ func _repack_and_save(root: Node, path: String) -> bool:
 	# dropping the sibling's `script = ExtResource(...)` line (or re-embedding it as
 	# a sub_resource). Re-anchoring repoints every node at the single cache owner
 	# for its path, so the saver sees one consistent ext_resource per path.
+	#
+	# This is the CENTRAL shared-mutation hardening point. Because it runs from the
+	# shared pack-and-save tail, it hardens every mutating re-pack op (node
+	# add/set/remove/duplicate/move, signal connect/disconnect) — not only `script
+	# attach` (issue #164's reported path). That broader reach is correct, not
+	# overreach: _reanchor_external_scripts only repoints a node that STILL carries
+	# its captured script (siblings preserved); it leaves a node the op
+	# intentionally re-scripted to a different non-empty path alone; and it skips
+	# nodes that vanished since capture (remove/move). `node add` is covered by
+	# test_node_add_preserves_sibling_script_on_repack_when_unimported.
 	_reanchor_external_scripts(root)
 	var repacked := PackedScene.new()
 	var pack_err := repacked.pack(root)
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -3126,6 +3126,10 @@ func _load_for_mutation(params: Dictionary) -> Node:
 		_fail(OP_ERROR_MISSING_DEPENDENCY, "scene nodes vanished or degraded on load: "
 				+ ", ".join(unmaterialized) + " — re-saving would silently drop or downgrade them; check the scene's dependencies and --project")
 		return null
+	# Snapshot every node's external script path NOW — the instant after
+	# instantiate, before any op-specific load can evict a sibling's script object
+	# (issue #164). _repack_and_save re-anchors from this snapshot on the way out.
+	_capture_external_scripts(root)
 	return root
 
 
@@ -3141,6 +3145,20 @@ func _load_for_mutation(params: Dictionary) -> Node:
 # before calling, as the tree is gone once this returns; for scene create the caller
 # also creates any missing parent dirs first, since this tail only packs and saves.
 func _repack_and_save(root: Node, path: String) -> bool:
+	# Re-anchor every external script captured at load time to the one canonical
+	# cached resource for its res:// path BEFORE packing (issue #164). On the
+	# editor build gda drives, the text scene saver dedups ext_resources through a
+	# PATH-keyed cache (ResourceCache::resource_path_cache), not object identity. A
+	# re-attach can leave two distinct in-memory Script objects sharing one res://
+	# path: the engine's GDScriptCache upgrades a shallow script to a full one via
+	# set_path(take_over=true), which evicts the previously cached object WITHOUT
+	# freeing it, so a sibling node still holds the evicted orphan. With an
+	# UNIMPORTED script the path string is the only identity (no uid://), so the
+	# path-keyed dedup collapses the two same-path objects on save — silently
+	# dropping the sibling's `script = ExtResource(...)` line (or re-embedding it as
+	# a sub_resource). Re-anchoring repoints every node at the single cache owner
+	# for its path, so the saver sees one consistent ext_resource per path.
+	_reanchor_external_scripts(root)
 	var repacked := PackedScene.new()
 	var pack_err := repacked.pack(root)
 	if pack_err != OK:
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Review skill pass

Reviewed PR #176 against `main...95b907b9b22759e640b7d8ee211965fdd7dae475`.

Commit reviewed:
- `95b907b` `fix(script): preserve sibling scripts on re-pack for unimported projects (#164)`

Spec source: #164, "script attach: re-attaching in an unimported project can drop a sibling node's script on re-pack".

## Standards

No findings.

Hard violations: none.
Judgement calls: none.

The change stays within ADR-0010's default GDScript op-dispatch path and does not add a new stdout/result contract, error-code registry entry, schema surface, or CLI/path-resolution surface. The new test is marked `e2e` and starts Godot, which matches the documented expectation that integration-point changes run a real integration layer (`RULES.md`) and the existing `e2e` marker definition (`pyproject.toml`).

Residual risk tied to a documented standard: ADR-0003 declares Godot 4.4+ support with 4.6 as the test baseline. This PR and its test are explicitly framed around 4.6.x engine behavior; I did not verify the same cache/take-over behavior on 4.4. That is a compatibility risk, not a documented standards violation in this diff.

## Spec

1. Required: the deterministic regression test does not exercise the actual product wiring.

   Spec quote: "A deterministic regression test reproduces the original drop (e2e, against the real engine, in an unimported-project fixture) and verifies all sibling script bindings survive."

   The harness uses the real engine and unimported fixture, but it manually calls `_load_scene` plus `instantiate`, then manually calls `_capture_external_scripts`, manually calls `_reanchor_external_scripts`, and directly packs/saves (`tests/test_e2e_script.py:1196`, `tests/test_e2e_script.py:1203`, `tests/test_e2e_script.py:1218`, `tests/test_e2e_script.py:1221`). It does not prove the new production calls from `_load_for_mutation` and `_repack_and_save` are wired (`src/gda/ops/operations.gd:3129`, `src/gda/ops/operations.gd:3147`). The test would still pass if those two production wiring calls were deleted while the helpers remained. Please drive `_load_for_mutation(params)`, inject the same `take_over_path` precondition, attach B, then call `_repack_and_save(live, "res://main.tscn")` so the regression fails if either wiring point regresses.

2. Consider: the implementation scope is broader than the issue text.

   Spec quote: "`gda script attach` binds a `.gd` to a node, then re-packs and saves the `.tscn`" and "The fix makes `script attach` deterministically preserve every other node's existing script binding on re-pack".

   The diff hooks `_reanchor_external_scripts(root)` into shared `_repack_and_save`, so it affects node add/set/remove/duplicate/move and signal connect/disconnect in addition to `script attach`. That may be the right central hardening point, but it is broader than the issue asks for. A short PR rationale or a small shared-tail coverage case would make the broader surface easier to accept.

3. FYI: CI does not currently demonstrate the real-engine side of the acceptance criteria.

   Spec quote: "No regression to existing `script attach` behavior - displacement reporting (#132), compile/type-incompatibility verdicts (#136, #137), and the registry drift tests stay green."

   The PR checks show "Python unit tests and package build" passing, but the "Godot e2e tests" job is skipped. I locally verified the affected paths with Godot 4.6.3; see the separate quality-review comment for commands and results.

Summary: Standards 0 findings; Spec 1 required finding plus 2 non-blocking risks. Worst issue: the new regression test can pass without the actual `_load_for_mutation` / `_repack_and_save` wiring.

qinhaihong-red:

### Findings

Required: `tests/test_e2e_script.py:1196` does not exercise the production wiring added in `src/gda/ops/operations.gd:3129` and `src/gda/ops/operations.gd:3147`.

The implementation wires capture into `_load_for_mutation` and re-anchor into `_repack_and_save`, but the new e2e harness bypasses both wiring points: it calls `_load_scene`/`instantiate`, then manually invokes `_capture_external_scripts`, manually invokes `_reanchor_external_scripts`, and packs/saves directly. That means the test validates the helper behavior and the deterministic engine precondition, but it would still pass if `_load_for_mutation` stopped capturing or `_repack_and_save` stopped re-anchoring. This misses the acceptance target for a regression test that protects `gda script attach`'s re-pack path.

Please adjust the harness so it calls the real product path around the deterministic injection point: `ops.call("_load_for_mutation", params)`, inject the same `take_over_path("res://a.gd")` precondition, attach `b.gd` to B, then call `ops.call("_repack_and_save", live, "res://main.tscn")` before asserting the saved scene. That would fail if either production wiring point regresses.

Consider: `_reanchor_external_scripts(root)` now runs from the shared `_repack_and_save` tail, so it covers more than `script attach` (node add/set/remove/duplicate/move and signal connect/disconnect also pass through it). I did not find a current correctness bug there: `node set` cannot coerce resource/script values, removed nodes are skipped, and intentionally re-scripted nodes with a different non-empty `resource_path` are left alone. Still, because the hook is intentionally broader than #164's `script attach` path, one small shared-tail coverage case or a short rationale in the PR body would reduce review risk.

### Five-axis notes

Correctness: the fix logic is plausible and the targeted local e2e passes, but the regression test does not guard the actual wiring, which is the required change above.

Readability: the implementation is heavily commented. The comments are long, but the root cause is engine-specific enough that the extra context is useful.

Architecture: centralizing in `_load_for_mutation` / `_repack_and_save` matches the existing shared mutation boundary.

Security: no new untrusted boundary. This remains inside the documented trusted-project execution model.

Performance: the new tree walk plus script reload happens on mutating pack/save paths. It is bounded by scene node count and existing mutating ops already instantiate the scene; I did not see a hot-path concern.

### Verification

- GitHub checks: `Python unit tests and package build` passed; `Godot e2e tests` skipped.
- Local PR-head verification in `/tmp/gda-pr176-review-clone`:
  - `uv --cache-dir /tmp/gda-pr176-uv-cache run pytest -m 'not e2e' -q` -> `482 passed, 227 deselected`
  - `env HOME=/tmp/gda-pr176-home TMPDIR=/tmp/gda-pr176-tmp GDA_GODOT=/Users/lelej/Applications/Godot.app/Contents/MacOS/Godot uv --cache-dir /tmp/gda-pr176-uv-cache run pytest -q --basetemp=/tmp/gda-pr176-pytest-targeted tests/test_e2e_script.py::test_script_attach_binds_script_to_node_and_scene_references_it tests/test_e2e_script.py::test_script_attach_overwrites_and_reports_the_displaced_script tests/test_e2e_script.py::test_script_attach_non_compiling_script_yields_script_compile_failed tests/test_e2e_script.py::test_script_attach_incompatible_node_type_yields_incompatible_script_type tests/test_e2e_script.py::test_script_attach_both_scene_and_script_missing_reports_the_scene_first tests/test_e2e_script.py::test_script_attach_preserves_sibling_script_on_repack_when_unimported` -> `6 passed`

Verdict: request changes for the test coverage gap above. The implementation itself looks close; the main issue is that the new regression test does not prove the production wiring that makes the fix active.

qinhaihong-red:
Both addressed in `fe7aef6` — thanks, this was the right call; the test now actually guards the fix.

**Required — regression test didn't exercise the production wiring:** rewritten so the harness drives the real mutate entry points and nothing else:
1. `ops.call("_load_for_mutation", params)` — captures the external-script snapshot internally; returns the live root.
2. inject `take_over_path("res://a.gd")` (the orphan-eviction) **after** `_load_for_mutation` returned, matching real ordering.
3. attach `b.gd` to node B.
4. `ops.call("_repack_and_save", live, "res://main.tscn")` — re-anchors internally before pack/save.

The harness no longer calls `_capture_external_scripts` / `_reanchor_external_scripts` itself (the `has_method` guards are gone), so it now guards the **wiring**. Verified red→green on **both** points independently: deleting `_capture_external_scripts(root)` in `_load_for_mutation` → FAIL with the `sub_resource type="GDScript"` corruption signature; deleting `_reanchor_external_scripts(root)` in `_repack_and_save` → FAIL the same way; both restored → pass.

**Consider — broader-than-`script attach` surface:** did both. Added `test_node_add_preserves_sibling_script_on_repack_when_unimported` — a NON-attach mutation (`node add`) over the same unimported two-node scene with the same eviction, asserting the scripted sibling survives the shared re-pack (it also goes red under both wiring-removal probes). And added a comment at the `_reanchor_external_scripts` call site explaining it's the central shared-mutation hardening point and why the broader reach is correct (siblings preserved; intentionally re-scripted nodes left alone; vanished nodes skipped). The fix logic in `operations.gd` is otherwise unchanged.

**FYI — CI e2e job skipped:** agreed, the `Godot e2e tests` job not running in CI is a real gap, but it's a CI-config matter beyond this PR. Locally: `tests/test_e2e_script.py -m e2e` → **50 passed** (incl. displacement / compile-failure / type-incompatibility — no regression); engine-free → **482 passed**.

PR #182, opened 2026-06-17
test(e2e): isolate fixtures from user:// log contention (#180)

### LINKED ISSUE #180: Make the e2e suite robust against shared user:// log-dir contention (engine_crashed flakiness)
## What to build

The e2e tier intermittently fails with `engine_crashed` (`RotatedFileLogger::rotate_file()` / `Failed to open 'user://logs/godot….log'`). It looks like a product/engine crash but is **test-infra contention**, not a `gda` bug.

Root cause (verified): Godot desktop enables file logging by default (a rotating logger that keeps ~5 files). `user://` resolves to `$HOME/Library/Application Support/Godot/app_userdata/<project-name>/`. The shared `godot_project` conftest fixture names **every** project `gda-e2e-fixture`, so every e2e test that uses it writes to **one** `user://logs/` directory. When e2e runs overlap — parallel agents, a CI matrix, or a manual run while another is going — the concurrent Godot processes race in `rotate_file()` against that shared dir and abort. (Observed: a full-suite run hit 78/227 `engine_crashed` while other e2e activity ran against the same real `HOME`.)

The current workaround is undocumented tribal knowledge: run the full suite with an isolated state — `env HOME=<fresh tmp> TMPDIR=<fresh tmp> GDA_GODOT=<abs Godot path> uv run pytest` (absolute Godot path, since a `~`-relative `$GDA_GODOT` re-expands against the overridden `HOME`). Replace that workaround with a real fix in the test fixture so the suite is reliable by default.

Make the e2e launches stop contending on one log dir. Two viable approaches — **(a)** is preferred as the most targeted (touches only the test fixture, not product behavior):

- **(a)** Disable file logging in the fixture's `project.godot` (a `[debug]` `file_logging/enable_file_logging=false` setting), so no `user://logs` is written at all during e2e.
- **(b)** Give each e2e test a unique project name / user-data dir so `user://logs` is isolated per test.

Note this is the test-fixture's logging, not a change to how `gda` runs the engine for real users; keep product behavior untouched.

## Acceptance criteria

- [ ] The full e2e suite (`uv run pytest`, e2e tier included) runs reliably against a normal, lived-in `HOME` — **no `fresh-HOME` workaround needed** and no `engine_crashed` / `RotatedFileLogger` failures.
- [ ] The suite stays green even when **two e2e runs execute concurrently** against the same `HOME` (the parallel-subagent / CI-matrix scenario that surfaced this) — demonstrate the isolation holds under overlap.
- [ ] The fix is confined to test infrastructure (the `godot_project` fixture / `conftest`); `gda`'s production engine-invocation behavior is unchanged.
- [ ] A short note in the e2e/test docs (or the fixture) records that e2e launches are isolated from `user://` log contention, so the rationale survives.

## Blocked by

None - can start immediately.

qinhaihong-red:
## What

Eliminates the `engine_crashed` / `RotatedFileLogger::rotate_file()` flakiness in the e2e tier. Root cause: Godot desktop enables file logging by default; `user://` resolves to `…/app_userdata/<config/name>/`, and the e2e fixtures shared one project name, so concurrent e2e launches (parallel agents / CI matrix) raced in `rotate_file()` against one shared `user://logs` dir and aborted. It looked like a product crash but was test-infra contention. This removes the log write at the source so the e2e tier is reliable by default — no `fresh-HOME` workaround needed.

## How

- New `project_godot(name="gda-e2e-fixture", extra="")` builder in `tests/conftest.py` — the single source of truth for every e2e `project.godot`. It always appends a `[debug]` section disabling file logging; `extra` appends test-specific sections (`[autoload]`, etc.) verbatim. `PROJECT_GODOT` is rebuilt through it.
- Routed every e2e test that writes its own `project.godot` through the builder: `gda-refgraph-fixture`/`gda-classname-fixture`/`gda-binary-fixture` (project-analysis) and `gda-e2e-autoload-fixture` (scene-security). (Export tests write only `export_presets.cfg` over the shared fixture, already covered.)
- Confined to test infra; **no production code changed**. `gda`'s real engine-invocation behavior is unchanged.

## Key finding — the `.pc` override is the operative key

Setting `debug/file_logging/enable_file_logging=false` alone is a **no-op on desktop**: Godot's base key defaults to `false`, but the `.pc` (desktop) feature-tagged override defaults to `true` and wins at startup (`main/main.cpp`). The builder sets **both** `file_logging/enable_file_logging=false` **and** `file_logging/enable_file_logging.pc=false`. Verified empirically: base-key-only still wrote `godot.log`; with the `.pc` override, none.

## Proof & results

- **Deterministic "no log dir" proof:** ran the full e2e suite against a clean throwaway `HOME`, then scanned `app_userdata/` — across all 7 fixture project dirs, `find` reported **0** `godot.log` files and **0** `logs/` directories. The contention source is structurally removed (stronger than reproducing the race).
- **Full suite (e2e included):** **724 passed, 1 skipped, 0 failed** — no `engine_crashed`/`RotatedFileLogger`.
- **Engine-free:** **491 passed**.

Closes #180.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,23 @@ domain slices (node, script, …) reuse it rather than growing their own.
 tier (issue #106): any ``e2e``-marked test selected without a resolvable Godot
 binary fails loudly here, naming the resolved path, rather than silently
 skipping per-module.
+
+``project_godot`` (issue #180) is the single builder for every e2e
+``project.godot``. It always disables Godot's default desktop file logging, so
+**no** e2e engine launch writes a ``user://logs/godot.log``. Why this matters:
+Godot resolves ``user://`` to ``$HOME/Library/Application Support/Godot/
+app_userdata/<config/name>/`` and, on desktop, attaches a ``RotatedFileLogger``
+that races in ``rotate_file()`` when concurrent launches share one log dir.
+Overlapping e2e runs (parallel agents / a CI matrix) then abort with what looks
+like ``engine_crashed`` but is purely test-infra contention. Removing the log
+write at the source makes the e2e tier reliable by default — no ``fresh-HOME``
+workaround needed. The operative key is the **platform override**
+``debug/file_logging/enable_file_logging.pc``: the base key defaults to
+``false`` but Godot's ``.pc`` feature-tagged override defaults to ``true`` on
+desktop and wins at startup (godot ``main/main.cpp``), so disabling the base
+alone is a no-op — the ``.pc`` override must be set too. Every e2e test that
+needs its own ``project.godot`` must build it through this helper (passing its
+extra sections via ``extra``) so the logging stays disabled.
 """
 
 import pytest
@@ -34,14 +51,37 @@ def _require_godot_engine(request):
         )
 
 
-# The minimal project.godot a Godot 4 engine accepts as a project root.
-PROJECT_GODOT = """\
-config_version=5
+# Disables Godot's default desktop file logging so an e2e launch writes no
+# ``user://logs/godot.log`` (issue #180). The ``.pc`` line is the operative one:
+# on desktop the ``.pc`` feature override (default ``true``) wins over the base
+# key (default ``false``), so both are set off.
+_DEBUG_NO_FILE_LOGGING = """\
+[debug]
 
-[application]
-
-config/name="gda-e2e-fixture"
+file_logging/enable_file_logging=false
+file_logging/enable_file_logging.pc=false
 """
+
+
+def project_godot(name: str = "gda-e2e-fixture", extra: str = "") -> str:
+    """Build a ``project.godot`` with file logging disabled (issue #180).
+
+    The single source of truth for every e2e ``project.godot``. ``name`` sets
+    ``config/name`` (which is also the ``user://`` dir name); ``extra`` appends
+    further sections (``[autoload]``, ``run/main_scene``, ``[editor_plugins]``,
+    …) verbatim. File logging is always disabled so no e2e launch writes to a
+    shared ``user://logs`` dir.
+    """
+    body = f'config_version=5\n\n[application]\n\nconfig/name="{name}"\n'
+    if extra:
+        body += "\n" + extra.rstrip("\n") + "\n"
+    return body + "\n" + _DEBUG_NO_FILE_LOGGING
+
+
+# The minimal project.godot a Godot 4 engine accepts as a project root, with
+# e2e file logging disabled (issue #180). Built through ``project_godot`` so the
+# logging-disable rationale lives in exactly one place.
+PROJECT_GODOT = project_godot()
 
 
 @pytest.fixture
--- a/tests/test_e2e_project_analysis.py
+++ b/tests/test_e2e_project_analysis.py
@@ -18,17 +18,17 @@ import pytest
 
 from gda.binary import resolve_godot_binary
 
+from .conftest import project_godot
+
 GODOT = resolve_godot_binary()
 
 
 # A project.godot with an autoload, a main scene, and an enabled editor plugin —
 # the project-level references and the statistics fields the commands report.
-PROJECT_GODOT = """\
-config_version=5
-
-[application]
-
-config/name="gda-refgraph-fixture"
+# Built through ``project_godot`` so e2e file logging stays disabled (issue #180).
+PROJECT_GODOT = project_godot(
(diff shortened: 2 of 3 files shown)


qinhaihong-red:
## `$review` pass: Standards / Spec

Fixed point: `main`; PR head: `test/180-e2e-log-isolation`; reviewed diff equivalent to `git diff main...HEAD` for PR #182.

## Standards

No Standards findings.

I checked the documented project standards from `AGENTS.md`, `RULES.md`, `CONTEXT.md`, `docs/agents/*`, the ADRs, and `pyproject.toml`. The diff is confined to e2e test infrastructure and routes `project.godot` generation through a shared builder. It does not alter `gda` production engine invocation behavior or the documented command / structured-output contracts.

Standards-grounded verification note: `RULES.md` says integration-point changes must run the test layer that actually exercises that integration point, and `pyproject.toml` defines `e2e` as the real-Godot tier. I ran that tier locally; results are below.

## Spec

No implementation findings against issue #180.

The PR implements the preferred fixture-level approach: `project_godot(...)` always disables Godot file logging, including the desktop `.pc` override, and the visible e2e `project.godot` writers are routed through it. The rationale note is present in `tests/conftest.py`, and the change is confined to test infrastructure.

One residual scope note: issue #180 asks to demonstrate that the suite stays green when two e2e runs execute concurrently against the same `HOME`. I ran a concurrent e2e smoke, not two complete concurrent e2e suites.

## Verification

- CI: Python unit tests and package build passed; Godot e2e job was skipped.
- Local PR-head checkout in `/private/tmp`: `uv run pytest -m 'not e2e'` passed (`491 passed, 234 deselected`).
- Local PR-head checkout in `/private/tmp`, normal `HOME`: `uv run pytest -m e2e` passed (`233 passed, 1 skipped, 491 deselected`), with no `engine_crashed` / `RotatedFileLogger` failures.
- Local same-`HOME` concurrent smoke: two overlapping runs of `uv run pytest -m e2e tests/test_e2e_project.py tests/test_e2e_scene_security.py` both passed (`18 passed` each).

Summary: Standards findings 0; Spec implementation findings 0. Worst remaining issue is verification scope only: I did not run two full e2e suites concurrently.

qinhaihong-red:
## `$agent-skills:code-review-and-quality` pass

### Verdict

No required code changes found.

This is a focused test-infrastructure fix: it centralizes e2e `project.godot` generation, disables Godot's desktop file logger at the fixture level, preserves production `gda` engine invocation behavior, and the relevant real-Godot validation is green locally.

### Correctness

No correctness findings.

- `project_godot(...)` sets both `file_logging/enable_file_logging=false` and the desktop override `file_logging/enable_file_logging.pc=false`.
- The shared `godot_project` fixture uses the builder.
- The visible e2e tests that write custom `project.godot` files now use the builder as well: the project-analysis fixtures and the scene-security autoload fixture.
- The fix stays on the fixture/test side; production code is untouched.

### Readability & Simplicity

No required readability findings. The rationale in `tests/conftest.py` is longer than a usual fixture comment, but it earns its size here: the `.pc` override is the non-obvious part, and without that note the next cleanup pass could easily remove the operative setting.

### Architecture

No architecture findings. Centralizing the e2e project scaffold in one builder fits the issue: the problem was duplicated fixture state and shared `user://logs` behavior, not command behavior. The change keeps the boundary clean by changing test infrastructure only.

### Security

No security findings. This PR does not alter trusted-project semantics, subprocess invocation for real users, input parsing, dependency surface, or secret handling.

### Performance

No performance findings. The change removes file-log writes from e2e Godot launches, which should reduce a test-infra contention point without adding meaningful overhead.

### Verification

- CI: Python unit tests and package build passed; Godot e2e job was skipped.
- Local PR-head checkout in `/private/tmp`: `uv run pytest -m 'not e2e'` passed (`491 passed, 234 deselected`).
- Local PR-head checkout in `/private/tmp`, normal `HOME`: `uv run pytest -m e2e` passed (`233 passed, 1 skipped, 491 deselected`) with no `engine_crashed` / `RotatedFileLogger` failures.
- Local same-`HOME` concurrent smoke: two overlapping runs of `uv run pytest -m e2e tests/test_e2e_project.py tests/test_e2e_scene_security.py` both passed (`18 passed` each).

Residual risk: I did not run two complete e2e suites concurrently, so that exact acceptance scenario is still covered by code structure plus the smaller concurrent smoke, not by a full duplicate-suite run.

PR #191, opened 2026-06-17
docs(adr): record gda-mcp architecture (ADR-0011..0014)

qinhaihong-red:
## What

Records the architecture decisions for `gda-mcp`, aligned in a grill-with-docs session and anchoring PRD #8.

- **ADR-0011 — integration mechanism:** `gda-mcp` wraps the `gda` CLI as a subprocess (public ABI only, no in-process import). `gda-daemon` sits below `gda`; `gda-mcp` always wraps the CLI and never speaks to the daemon directly, so it follows `gda` into Phase 2 with zero per-phase work.
- **ADR-0012 — tool-surface generation:** runtime introspection of a single whole-surface `gda` schema dump (incl. descriptions); one MCP tool per command (`<group>_<command>`); full surface, no hand-curated profiles (config-driven group filtering reserved if needed).
- **ADR-0013 — packaging & launch:** one distribution, MCP behind a `gda[mcp]` extra; `gda-mcp` stdio console script on the `mcp` low-level Server; per-agent registration recipes as a deliverable.
- **ADR-0014 — project-context resolution:** target project is server context, not a tool parameter (honors ADR-0006); resolved by a portable precedence (`GDA_PROJECT` → `CLAUDE_PROJECT_DIR` → `roots/list` → cwd) because cwd is unreliable across agents.

## Why

These ADRs are the authoritative anchors referenced by **PRD #8** (`gda-mcp`). Landing them keeps the issue / doc / code reference chain consistent before the implementation slices are filed via `/to-issues`.

Docs-only change.

Refs #8

### DIFF
--- a/docs/adr/0011-gda-mcp-integration-mechanism.md
+++ b/docs/adr/0011-gda-mcp-integration-mechanism.md
@@ -22,11 +22,27 @@ gda-daemon exists in Phase 2 — the topology of {gda-mcp, gda, gda-daemon}.
 ## Decision
 
 **gda-mcp wraps `gda` as a subprocess.** It shells out to the installed `gda`
-console script (`gda <group> <command> --json …`, and `gda <group> <command>
---schema` for tool generation) and consumes only `gda`'s **public ABI**: the
-ADR-0002 sentinel-delimited JSON result / error envelope on stdout, the ADR-0002
-exit-code categories, and the ADR-0004 `--schema` self-description. It does **not**
+console script and consumes only `gda`'s **public CLI ABI** — *not* the internal
+sentinel protocol ADR-0002 defines **between** `gda` and its headless operation
+subprocesses (which `gda` already parses away before emitting its public output, and
+which ADR-0010's native-CLI-mode operations do not emit at all). That public ABI is:
+the `--json` success output, the structured `GdaError` envelope on a non-zero exit
+(the uniform `GdaErrorEnvelope` of ADR-0004), the exit-code categories, and the
+`--schema` self-description (ADR-0004). Operations are invoked per-command as
+`gda <group> <command> --json …`; the **tool surface is enumerated from the aggregate
+schema dump** (ADR-0012), *not* a per-command `--schema` fan-out. gda-mcp does **not**
 import `gda`'s Python modules or depend on any internal symbol.
+
+**Error mapping is part of this mechanism, and is lossless.** gda-mcp keys on `gda`'s
+exit code: exit 0 is success — the `--json` result becomes the tool's
+`structuredContent` (validated against its `outputSchema`); any non-zero exit becomes
+`CallToolResult(is_error=True)` carrying the **full** `GdaError` envelope
+`{code, category, message, diagnostics}` **losslessly** as JSON in the result content.
+The envelope is **never** flattened to a prose string — the stable `code` exists
+precisely so an agent branches on it without parsing prose — and the `error` schema
+stays out of `outputSchema` (ADR-0004). All non-zero categories (environment /
+version / operation / parse / contract_violation, including launch failures) map
+uniformly to `isError`; `category` / `code` distinguish them.
 
 **Phase-2 topology — gda-daemon sits *below* `gda`, not beside gda-mcp.** This ADR
 records the load-bearing assumption that makes ADR-0001's "gda-mcp follows `gda` into
@@ -36,7 +52,7 @@ Phase 2 automatically" literally true:
   them, internally, to gda-daemon over its IPC channel; gda-daemon owns the
   persistent engine and therefore the cross-call state (the subject of #5).
 - Each `gda` invocation stays a one-shot, stateless RPC to the daemon and emits the
-  **same** ADR-0002 contract a headless op does.
+  **same** public `--json` / `GdaError` contract a headless op does.
 - **gda-mcp always wraps the CLI and never speaks to gda-daemon directly.** The
   headless/live distinction stays invisible to gda-mcp exactly as ADR-0005 keeps it
   invisible in the command tree. gda-mcp therefore needs no IPC client and no
@@ -47,7 +63,7 @@ Phase 2 automatically" literally true:
 
 - **In-process import (rejected).** gda-mcp imports `gda` and calls its
   functions/models directly. Rejected because: (1) it couples gda-mcp to `gda`'s
-  internals instead of the frozen public ABI that `--schema` / ADR-0002 exist to
+  internals instead of the frozen public CLI ABI that `--schema` (ADR-0004) exists to
   expose; (2) its only real upside is per-call speed, which is illusory in Phase 1 —
   every call's latency is dominated by Godot's own `--headless` spawn (60 s / 600 s
   timeouts), beside which a `gda` process start is noise; (3) it gets *worse* under
@@ -78,6 +94,6 @@ Phase 2 automatically" literally true:
   design (#5, #7): the daemon backs the CLI, and "state consistency" (#5) is a
   property of the daemon holding the engine across one-shot CLI calls — gda-mcp does
   not participate in it.
-- gda-mcp's packaging, tool-generation strategy, transport, and error-mapping
-  fidelity are deliberately out of scope here; they are downstream of this mechanism
-  and decided separately.
+- This ADR owns the integration mechanism **and its error mapping**. gda-mcp's
+  packaging, tool-generation strategy, and transport are downstream and decided
+  separately (ADR-0013, ADR-0012, and the PRD respectively).
--- a/docs/adr/0011-gda-mcp-integration-mechanism.md
+++ b/docs/adr/0011-gda-mcp-integration-mechanism.md
@@ -1,0 +1,83 @@
+---
+status: accepted
+---
+
+# gda-mcp integration mechanism: a subprocess adapter over the gda CLI
+
+ADR-0000 fixes the component order `gda → gda-mcp → gda-daemon` and frames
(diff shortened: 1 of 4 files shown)


qinhaihong-red:
## Review pass 1/2: Standards + Spec

Reviewed PR #191 against base `main` / head `d96a4ec06e38` with:

- Standards sources: `RULES.md`, `CONTEXT.md`, `AGENTS.md`, and existing ADRs `0000`, `0001`, `0002`, `0004`, `0005`, `0006`, `0008`, `0009`, `0010`
- Spec source: issue #8 (`PRD: gda-mcp — a generated MCP adapter over the gda CLI`)

Tests/e2e were not run because this is a docs-only ADR review and the request explicitly scoped verification to design review.

## Standards

### Required

`docs/adr/0011-gda-mcp-integration-mechanism.md` currently says `gda-mcp` consumes the “ADR-0002 sentinel-delimited JSON result / error envelope on stdout” as `gda`'s public ABI.

That crosses the documented boundary. ADR-0002 defines the sentinel protocol as the GDScript/headless subprocess contract that `gda` extracts before emitting the public wrapper. ADR-0004 defines the adapter-facing contract as `--json` success output plus a structured error envelope on non-zero exit, and ADR-0010 explicitly notes native CLI-mode operations emit no ADR-0002 sentinel while the agent-facing contract remains typed models, `--json`, `--schema`, and registered error codes.

Please revise ADR-0011 so `gda-mcp` consumes `gda`'s public CLI ABI (`--json`, exit code, structured `GdaError` envelope, `--schema` / aggregate schema), not the internal sentinel protocol between `gda` and operation subprocesses.

### Consider

`docs/adr/0014-gda-mcp-project-context-resolution.md` says the resolved path is passed as `gda --project <dir>` to every subprocess, including values derived from `roots/list` / cwd fallback.

ADR-0006 distinguishes explicit project context from cwd fallback: explicit `--project` / `$GDA_PROJECT` must be a Godot project, while cwd only counts as a project if it contains `project.godot`; otherwise `gda` runs projectless. If `gda-mcp` intentionally does not support projectless operation, the ADR should say that and require candidate validation. As written, it risks turning an unreliable cwd/root into an explicit `--project`, changing ADR-0006 path semantics.

## Spec

### Required

`docs/adr/0011-gda-mcp-integration-mechanism.md` records the wrong schema-enumeration path. Issue #8 requires an aggregate dump: “whole surface as one JSON document”; `gda-mcp` introspects that once at startup. ADR-0011 instead says tool generation shells out to `gda <group> <command> --schema`, which contradicts both the spec and ADR-0012 and reintroduces the per-command fan-out the spec rejected.

Please align ADR-0011 with ADR-0012: per-command `--schema` is the source model for individual commands, but `gda-mcp` enumerates the surface through the aggregate schema dump.

### Required

`docs/adr/0014-gda-mcp-project-context-resolution.md` weakens a required precedence step. Issue #8 specifies the portable precedence `GDA_PROJECT -> CLAUDE_PROJECT_DIR -> MCP roots/list -> cwd`. The ADR lists that order, but then says the startup `roots/list` read “may itself be staged after the env-based resolution in the first slice.” That deferral is not in the PRD and makes `roots/list` support partial.

Please either keep `roots/list` in the first-slice contract or update issue #8/PRD first so the ADR does not silently narrow the accepted requirement.

### Required

The ADR set only partially records the error-mapping contract. Issue #8 requires non-zero `gda` exits to map to `CallToolResult(is_error=True)` while carrying the full error envelope `{code, category, message, diagnostics}` losslessly as JSON, and explicitly says the envelope is “never flattened to a prose message.”

The new ADRs mention `isError`, but ADR-0011 leaves error-mapping fidelity out of scope and no ADR captures the exit-code-to-`CallToolResult(is_error=True)` / lossless JSON content contract. Please capture that contract in one of these ADRs, or link to a dedicated ADR if it is intentionally separated.

Summary: Standards found 1 required issue and 1 judgement-call concern. Spec found 3 required issues. Worst issue: ADR-0011 currently exposes the internal sentinel protocol as `gda-mcp`'s public ABI and also states the wrong schema-enumeration path.

qinhaihong-red:
## Review pass 2/2: Code Review and Quality

Reviewed PR #191 as a docs-only architecture change. I did not run tests or e2e per the review request.

### Verdict

Request changes. The ADRs improve the design record, but a few wording choices would make the future implementation target the wrong contract.

### Correctness

Required: `docs/adr/0011-gda-mcp-integration-mechanism.md` conflates `gda`'s public CLI ABI with the lower-level ADR-0002 sentinel protocol. Lines 24-28 say `gda-mcp` consumes the sentinel-delimited result/error envelope on stdout, but that sentinel is the contract between `gda` and its headless operation subprocesses. The MCP adapter should consume the public `gda` surface: command exit code, `--json` success output, structured `GdaError` envelope, and schema self-description. This matters because ADR-0010 already allows native CLI-mode operations that emit no ADR-0002 sentinel while preserving the same agent-facing contract.

Required: ADR-0011 also describes tool generation as `gda <group> <command> --schema`. That undercuts ADR-0012 and issue #8, which require one whole-surface aggregate schema dump at startup. If this lands as-is, implementation agents can reasonably build the rejected per-command fan-out path.

Required: the error relay contract is not anchored cleanly. Issue #8 requires non-zero `gda` exits to become `CallToolResult(is_error=True)` with the full `{code, category, message, diagnostics}` envelope carried losslessly as JSON and never flattened into prose. The new ADRs mention `isError`, but ADR-0011 then declares error-mapping fidelity out of scope. Either this ADR set should record the lossless mapping, or it should link to the document that owns that decision.

### Architecture

Required: `docs/adr/0014-gda-mcp-project-context-resolution.md` should preserve ADR-0006's distinction between explicit project context and projectless fallback. Passing every resolved cwd/root as `gda --project <dir>` turns a fallback signal into an explicit project assertion. The ADR should say whether `gda-mcp` validates the candidate as a Godot project, omits `--project` when no valid project is found, or intentionally rejects projectless mode.

Required or update the PRD first: ADR-0014 lists `roots/list` in the required precedence, then says the startup `roots/list` read may be staged after env-based resolution in the first slice. That is a scope reduction from issue #8's accepted precedence. If staged delivery is intended, it should be reflected in the PRD/work breakdown, not introduced only in the ADR consequence.

### Readability and Simplicity

The split into four ADRs is understandable and mostly clean: mechanism, surface generation, packaging, and project context are separated well. The main readability problem is that ADR-0011 reaches into topics that ADR-0012 or a missing error-mapping ADR own, then declares some of them out of scope. Tightening those boundaries will make the decision chain easier for implementation agents to follow.

### Security

No secrets or auth-sensitive code are introduced. Design risk to keep visible: project resolution determines which trusted project gets executed via `gda --project`. That makes the ADR-0014 validation/fallback ambiguity worth fixing before implementation slices rely on it.

### Performance

The chosen aggregate schema dump avoids startup-time per-command process fan-out and is the right performance shape. The only performance concern is the ADR-0011 wording that still points implementers toward per-command `--schema` calls; fixing that removes the risk.

### Verification

Not run: tests/e2e, by request. Reviewed PR metadata, issue #8, the PR diff, and existing ADRs.

qinhaihong-red:
Thanks — all findings were valid; each was a real inconsistency I introduced. Addressed in `e1a8999`.

## Spec / Correctness

- **ADR-0011 exposed the internal sentinel as the public ABI** *(Required)* — Fixed. ADR-0011 now states gda-mcp consumes only `gda`'s **public CLI ABI** (`--json` output, exit code, `GdaError`/`GdaErrorEnvelope` per ADR-0004, `--schema`), explicitly *not* the ADR-0002 sentinel (which `gda` parses away, and which ADR-0010's native-CLI-mode ops don't emit). Also corrected two lingering "ADR-0002 contract"/"`--schema` / ADR-0002" phrasings elsewhere in the ADR.
- **ADR-0011 pointed at per-command `--schema` fan-out** *(Required)* — Fixed. Tool-surface enumeration now reads "from the aggregate schema dump (ADR-0012), *not* a per-command `--schema` fan-out." Per-command `--json` remains only for *invoking* an operation.
- **Error-mapping contract un-anchored** *(Required)* — Fixed by recording it in ADR-0011 (it is part of the integration mechanism): exit 0 → `structuredContent` validated against `outputSchema`; non-zero → `CallToolResult(is_error=True)` carrying the full `{code, category, message, diagnostics}` envelope losslessly as JSON, never flattened. The "out of scope / decided separately" disclaimer is removed; PRD #8's error-mapping section now notes ADR-0011 as owner. (Kept as part of ADR-0011 rather than a new ADR, per a prior decision not to split it out.)
- **ADR-0014 silently narrowed the `roots/list` precedence** *(Required / or update PRD)* — Fixed by keeping `roots/list` in the first-delivery precedence (the staging hedge is gone); only the *dynamic* `roots/list_changed` re-resolution stays deferred, which matches PRD #8's Out of Scope. The ADR no longer dictates slice timing.

## Standards / Architecture

- **ADR-0014 turned an unvalidated cwd/root into an explicit `--project`** *(Required)* — Fixed. A candidate is asserted as `gda --project <dir>` **only when it is a valid Godot project** (contains `project.godot`); otherwise gda-mcp forges no `--project` and lets `gda` apply its own ADR-0006 cwd/projectless resolution. gda-mcp does not reject projectless mode; a project-requiring op surfaces `gda`'s own typed error, relayed unchanged. This preserves ADR-0006's explicit-vs-fallback distinction and removes the security-relevant ambiguity about which trusted project gets executed.

PRD #8 was updated for consistency (project-context wording + error-mapping ownership). No code/e2e in this PR (docs-only); the implementation slices filed next via `/to-issues` carry the e2e gate.

Issue #198, opened 2026-06-18
docs(adr): structured params-input ABI for gda commands


qinhaihong-red:
## Parent

PRD #8 — gda-mcp. Prerequisite for #193 (gda-mcp tracer bullet). Orthogonal to ADR-0004 (`--schema` self-description) and anchored to ADR-0011/0012 (gda-mcp consumes gda's public CLI ABI).

## What to build

A new ADR (next number: **ADR-0015**) recording **how a caller supplies a command's params as one structured object** — the gda-side prerequisite that lets gda-mcp dispatch parametrized tools without reconstructing argv. The decision was settled in a design review; this issue records it authoritatively.

**Decision to record:**

- gda gains a **uniform structured params-input channel**: `gda <group> <command> --params-json <inline-json | ->` where `-` reads the object from **stdin**. The object is deserialized into that command's **existing Pydantic input model** (`model_validate_json`). Command selection stays in argv (reuses Typer routing); only the params *source* changes. (**Form X**, chosen over a top-level `gda --parse-from-json {command, params}` that would re-implement command routing.)
- The **Pydantic params model is the single source of truth** for BOTH `gda schema`'s emitted `inputSchema` (`model_json_schema`) AND input parsing (`model_validate_json`). The MCP-facing input schema and the input-parse format are therefore aligned **by construction** — no second encoding of CLI binding, no drift.
- **Normalization currently done in CLI command bodies** (e.g. path normalization, derived defaults like scene-create's root-name) is pushed **down into the params models**, so the argv path and the `--params-json` path produce identical params.
- **Why not argv reconstruction:** the dump's `inputSchema` cannot encode CLI binding — gda has 34 positional `Argument`s + 42 `--kebab` `Option`s, a required field may be either, and positionals have no `--name` form. So gda-mcp cannot mechanically rebuild argv from the schema. A verbatim JSON passthrough keeps gda-mcp a pure schema→tool transform (ADR-0011/0012) and mirrors the existing gda→engine `params_json` convention.
- **Large payloads** (e.g. `script create` / `shader create` `content`) use stdin (`-`) to avoid argv length limits and process-listing leakage.

## Acceptance criteria

- [ ] A new ADR (status `accepted`) records the `--params-json` (Form X) decision, stdin support, the model-as-single-source invariant (inputSchema emission + input parsing both derive from the one model), and normalization-in-model.
[...]

Issue #193, opened 2026-06-17
feat(gda-mcp): generated stdio MCP server over the gda CLI (tracer bullet)


qinhaihong-red:
## Parent

PRD #8 — gda-mcp. Anchored by ADR-0011 (subprocess integration mechanism + lossless error mapping), ADR-0012 (tool-surface generation), ADR-0013 (packaging & launch), and ADR-0015 (#198 — structured params-input ABI; the prerequisite for parametrized dispatch).

## What to build

The gda-mcp server: an `mcp` low-level **stdio** server that, at startup, introspects the aggregate schema dump (#192) and registers one MCP tool per command — name `<group>_<command>`, `description` ← command help, `inputSchema` ← input schema, `outputSchema` ← output schema. On a tool call it shells out to the installed `gda` — invoked as `python -m gda` for a deterministic, same-environment binary (Design decision 3) — forwarding the tool input **verbatim** via `gda <group> <command> --params-json -` (the object on stdin; ADR-0015 / #198), consuming `gda`'s **public CLI ABI only** (`--json` output, exit code, `GdaError` envelope — never the internal sentinel; no in-process import).

**Error mapping (ADR-0011), mechanical off the exit code:**
- exit 0 → success: the `--json` result becomes `structuredContent`, validated against the tool's `outputSchema` (by the SDK — see Design decision 5).
- non-zero → `CallToolResult(is_error=True)` carrying the full `{code, category, message, diagnostics}` envelope **losslessly** as JSON in content; **never flattened** to a prose string; `error` stays out of `outputSchema`.

Packaged behind an optional `gda[mcp]` extra exposing a `gda-mcp` console script (also runnable via `uvx --from "gda[mcp]" gda-mcp`); SDK = official `mcp`, **low-level Server** (not FastMCP).

## Why low-level Server, not FastMCP

The `mcp` SDK offers two ways to build a server, and they differ in **who authors the tool schema**:

- **FastMCP** (high-level, decorator-based) derives a tool's `name` / `description` / `inputSchema` / `outputSchema` *from a decorated Python function's signature, type hints, and docstring*. Its value is generating JSON Schema **from** Python — it assumes "one tool = one hand-written Python function."
- **low-level `Server`** (`mcp.server.lowlevel.Server`) takes **explicit** `types.Tool(...)` objects from a `list_tools()` handler and routes invocations to a `call_tool()` handler you write. You supply the schemas; the SDK just speaks the protocol.

[...]

Issue #187, opened 2026-06-17
refactor(export): lift export run's recipe into a deep ExportRun operation


qinhaihong-red:
## What to build

`gda export run` is the only command that does not follow the thin-command pattern. Its
CLI function (`run_export`, ~115 lines) hand-orchestrates three phases —

1. **resolve** the preset via the `export-get` sentinel op (reusing #114's clean
   preset / project errors);
2. **structured preflight** (effective destination + template readiness, ADR-0010), which
   fails fast on `export_path_unset` / `export_templates_missing` with no native run;
3. the **native `--export-<mode>` run**, classified by `classify_export_run`;

— and then **re-implements** the `if json_output: … else render(…)` emit branch that
`HeadlessCommand.emit` already owns.

The orchestration recipe — the genuinely interesting, multi-phase, bug-prone logic — has
no **locality**: it lives inside a Typer function and can only be exercised through Typer's
CliRunner.

**Goal:** give the recipe a home. Extract a deep **ExportRun operation** that owns the
resolve → preflight → native-run → classify recipe behind one interface
(`run(params, …) -> ExportRunResult | Failure`), so the CLI command shrinks to the same
thin shape as every other command, and the recipe gets its own test surface (driven with
the two fakes, asserting phase sequencing and each preflight branch — no CliRunner).
**Pure refactor — zero behavior change.**

ADR-0010 anticipates that *"future editor-only capabilities (other `--export-*` variants,
Android build-template install) follow this precedent"* — this turns that precedent into a
reusable operation shape and removes the lone special-case command from the CLI.

## Acceptance criteria

- [ ] An `ExportRun` operation owns the resolve → preflight → native-run → classify recipe
      behind one interface returning `ExportRunResult | Failure`.
- [ ] The two runner seams (the op runner for `export-get`; the native export runner) are
      injected, so the operation is fully testable without a real engine.
- [ ] `run_export` shrinks to a thin command (build params → invoke the operation → emit via
      the shared emit path); the re-implemented `if json_output … else render(…)` branch is
      deleted.
- [ ] The recipe has direct tests (phase sequencing; both preflight branches
      `export_path_unset` / `export_templates_missing`; success and native-failure) that do
      NOT go through CliRunner.
- [ ] Existing `export run` command tests and e2e pass unchanged.
- [ ] No change to `--json`, `--schema`, exit codes, or the error envelope
      (ADR-0002 / 0004 / 0010 contract intact).
[...]

PR #201, opened 2026-06-18
feat(gda): uniform --params-json structured params-input + normalization parity

### LINKED ISSUE #199: feat(gda): uniform --params-json structured params-input + normalization parity
## Parent

PRD #8 — gda-mcp. Anchored by **ADR-0015** (from #198) and ADR-0004. The gda-side prerequisite that lets gda-mcp (#193) forward an MCP tool's input object to gda verbatim.

## What to build

Implement the structured params-input ABI decided in #198: every command accepts its params as **one JSON object** via `gda <group> <command> --params-json <inline-json | ->` (`-` reads from stdin), deserialized into the command's existing Pydantic input model and dispatched through the **same execution/emit path** as the argv form.

- The mechanism is **centralized** (like `--schema`) in the `HeadlessCommand` command class, so it lands uniformly across every command with near-zero per-command cost.
- **Normalization parity:** normalization/derivation currently done in CLI command bodies is moved into the params models, so the two input paths — individual argv args vs `--params-json` — produce **byte-identical** params.

This unblocks gda-mcp's parametrized tool dispatch: gda-mcp maps an MCP tool's input object straight onto `--params-json`, with the params model as the single source for both the emitted `inputSchema` and input parsing.

## Acceptance criteria

- [ ] Every command accepts `--params-json <inline-json>` and `--params-json -` (stdin), deserializing into its input model; the outcome is identical to invoking the command with the equivalent individual argv args (**normalization parity**).
- [ ] Invalid JSON or schema-invalid input yields gda's uniform structured error envelope (a clean `GdaError`, not a traceback), with a stable category/code.
- [ ] `--params-json` and individual positional/option args are **mutually exclusive** — supplying both is a clean usage error.
- [ ] The `--params-json`-vs-individual-args conflict (and invalid-JSON / schema-invalid input) surfaces as a **registered** `GdaError.code` per ADR-0002 (`error_codes.py` + the ADR-0002 table) — either reuse `usage_error`, broadening its registry description to cover CLI-side invocation conflicts (today it reads "operation dispatcher … without the required operation name"), or register a new code; never an unregistered / ad-hoc code. The error taxonomy stays single-sourced in ADR-0002 (ADR-0015 defers the exact code here).
- [ ] Normalization/derivation previously done in CLI bodies (e.g. scene-create path normalization and root-name derivation) lives in the params models and is exercised by both paths.
- [ ] `--params-json` composes with `--json` (structured result); a bare `--schema` still emits the contract and ignores params (precedence preserved).
- [ ] Fast tests assert round-trip parity (model → JSON → `--params-json` ≡ argv path) and the error / mutual-exclusivity behavior via the existing fake-runner seam, across a representative set of commands, **plus a full-coverage check** that every command supports `--params-json`.
- [ ] e2e: at least one real command with params (e.g. `scene create`) driven end-to-end via `--params-json` against the real engine, asserting the real outcome (file created). Per RULES.md DoD, fake/mock tests do not replace this e2e.

## Blocked by

- #198 (structured params-input ABI ADR)

qinhaihong-red:
## What & why

Implements **ADR-0015** (#198): a uniform structured params-input ABI so a caller can supply a command's params as **one JSON object** via `gda <group> <command> --params-json <json | ->` (`-` reads the object from stdin), instead of the individual CLI arguments. This is the gda-side prerequisite that lets **gda-mcp (#193)** forward an MCP tool's input object **verbatim** — no argv reconstruction (the dump's `inputSchema` deliberately doesn't encode CLI binding). Closes #199.

## Mechanism

- The command class (which already owns `--schema`) intercepts `--params-json`, builds the input model with `model_validate_json`, and dispatches it through the **same** project-resolution + runner seam the argv path uses (a hook `gda.cli` registers, so `gda.headless` needn't import `gda.cli`). The body is bypassed, staying a thin argv→model adapter.
- **Single source of truth:** normalization moved into the models via a reusable `NormalizedPath = Annotated[str, BeforeValidator(normalize_path)]`, plus scene-create's root-name derivation. The argv and `--params-json` paths therefore produce **identical** params — proven by an argv≡JSON parity test driven with `~/…` paths.
- **`gda schema`'s `inputSchema` is exactly the model `--params-json` accepts**, by construction (ADR-0004's input/output/error separation preserved — `--params-json` adds only an input channel).

## Structured errors (ADR-0002, single-sourced)

- `--params-json` + individual args → `usage_error` (mutually exclusive); malformed/schema-invalid JSON → `invalid_params`. Both are **registered** `GdaError` envelopes (never tracebacks), reusing existing codes; their `error_codes.py` + ADR-0002 descriptions were broadened to note the CLI origin. The ADR-0002 ↔ registry guard test confirms consistency.
- A bare `--schema` still wins (emit-only); `--json` composes with `--params-json`.

## Surface

Every command exposes `--params-json` (coverage test walks the live Typer tree), with the single documented exception of `gda schema` (the parameterless manifest emitter, no operation to drive). `export run` is special-cased: it runs a native-export recipe (`run_export_operation`, #187), not the sentinel pipeline, so its `--params-json` path mirrors that recipe rather than `cmd.emit`.

## Tests

- `tests/test_params_json.py`: mechanism, stdin, structured-error paths, flag composition, a **coverage** test (every command exposes `--params-json`), and an argv≡JSON **parity** table (`~/…` paths, asserts identical dispatch + that `~` actually expanded).
- `tests/test_export_run_commands.py`: regression test that `export run --params-json` drives the **native export runner** (guards the special-case).
- `tests/test_e2e_params_json.py`: **e2e gate** — `scene create --params-json` (inline + stdin) creates a loadable `.tscn` through the real engine.

Verification: `pytest -m "not e2e"` → **571 passed**; full `pytest -m e2e` → **235 passed, 1 skipped** (no regression from moving normalization into models); `uv build` clean.

## Acceptance criteria (#199)

- [x] Every command accepts `--params-json` (inline + stdin); identical to the argv path (normalization parity)
- [x] Invalid/conflicting input → registered structured `GdaError` (not a traceback)
- [x] `--params-json` mutually exclusive with individual args
- [x] Normalization lives in the params models, exercised by both paths
- [x] `--params-json` composes with `--json`; `--schema` precedence preserved
- [x] Fast round-trip parity + full-coverage tests; one real-engine e2e

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -89,9 +89,9 @@ operation, and parse codes the CLI assigns).
 | `unsupported_version` | `version` | `version_gate` | `3` | The detected Godot version is below the supported minimum. |
 | `engine_crashed` | `operation` | `classifier` | `4` | Godot terminated abnormally, such as by signal death. |
 | `operation_failed` | `operation` | `classifier` | `4` | The engine or operation failed without a valid registered operation error envelope. |
-| `usage_error` | `operation` | `operation` | `4` | The operation dispatcher was invoked without the required operation name. |
+| `usage_error` | `operation` | `operation` | `4` | The command was invoked incorrectly: the operation dispatcher received no operation name, or the CLI received `--params-json` together with the individual arguments (ADR-0015). |
 | `unknown_operation` | `operation` | `operation` | `4` | The operation dispatcher received an unknown operation name. |
-| `invalid_params` | `operation` | `operation` | `4` | The operation dispatcher received params that are not a JSON object. |
+| `invalid_params` | `operation` | `operation` | `4` | Params do not match the command's contract: the operation dispatcher received non-object params, or a `--params-json` object was malformed or schema-invalid (ADR-0015). |
 | `invalid_path` | `operation` | `operation` | `4` | A required path parameter is missing or invalid. |
 | `invalid_root_type` | `operation` | `operation` | `4` | A requested Godot root node type cannot be instantiated as a `Node`. |
 | `invalid_root_name` | `operation` | `operation` | `4` | A requested root node name is empty or would be rewritten by Godot. |
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -619,9 +619,14 @@ PROJECT_STATISTICS_COMMAND: HeadlessCommand[ProjectStatisticsResult] = HeadlessC
 )
 
 
-# Path normalization now lives in the models (ADR-0015), the single home shared
-# by the argv and ``--params-json`` paths. Commands not yet migrated to model-side
-# normalization still call this alias in their bodies.
+# Path normalization now lives in the models (ADR-0015) via the NormalizedPath
+# field type, the single home shared by the argv and ``--params-json`` paths —
+# every domain command's body passes its raw path straight to the params model.
+# The one exception is ``export run`` (issue #187): its argv body does NOT build
+# an ``ExportRunParams`` (it passes preset/mode/output straight to
+# ``run_export_operation``), so the model's NormalizedPath on ``output`` only
+# fires on the ``--params-json`` path. The argv ``--output`` is therefore still
+# normalized here, via this alias, to keep both export-run paths consistent.
 _normalize_path = normalize_path
 
 
@@ -664,13 +669,14 @@ def get(
     path: str = typer.Argument(..., help="The .tscn scene file to read."),
     json_output: bool = json_option(),
     schema: bool = SCENE_GET_COMMAND.schema_option(),
+    params_json: Optional[str] = params_json_option(),
     godot: Optional[str] = godot_option(),
     project: Optional[str] = project_option(),
 ) -> None:
     """Read a scene file and report its structured node tree."""
     _dispatch(
         SCENE_GET_COMMAND,
-        SceneGetParams(path=_normalize_path(path)),
+        SceneGetParams(path=path),
         json_output=json_output,
         godot=godot,
         project=project,
@@ -682,13 +688,14 @@ def get_exports(
     path: str = typer.Argument(..., help="The .tscn scene file to read."),
     json_output: bool = json_option(),
     schema: bool = SCENE_GET_EXPORTS_COMMAND.schema_option(),
+    params_json: Optional[str] = params_json_option(),
     godot: Optional[str] = godot_option(),
     project: Optional[str] = project_option(),
 ) -> None:
     """List the @export properties a scene's nodes' scripts declare, per node path."""
     _dispatch(
         SCENE_GET_EXPORTS_COMMAND,
-        SceneGetExportsParams(path=_normalize_path(path)),
+        SceneGetExportsParams(path=path),
         json_output=json_output,
         godot=godot,
         project=project,
@@ -699,6 +706,7 @@ def get_exports(
 def list_scenes(
     json_output: bool = json_option(),
     schema: bool = SCENE_LIST_COMMAND.schema_option(),
+    params_json: Optional[str] = params_json_option(),
     godot: Optional[str] = godot_option(),
     project: Optional[str] = project_option(),
 ) -> None:
@@ -717,13 +725,14 @@ def delete(
     path: str = typer.Argument(..., help="The .tscn scene file to delete."),
     json_output: bool = json_option(),
     schema: bool = SCENE_DELETE_COMMAND.schema_option(),
+    params_json: Optional[str] = params_json_option(),
     godot: Optional[str] = godot_option(),
     project: Optional[str] = project_option(),
 ) -> None:
     """Delete a scene file and report what was removed."""
     _dispatch(
         SCENE_DELETE_COMMAND,
-        SceneDeleteParams(path=_normalize_path(path)),
+        SceneDeleteParams(path=path),
         json_output=json_output,
         godot=godot,
         project=project,
@@ -756,6 +765,7 @@ def add(
     ),
     json_output: bool = json_option(),
     schema: bool = NODE_ADD_COMMAND.schema_option(),
+    params_json: Optional[str] = params_json_option(),
     godot: Optional[str] = godot_option(),
     project: Optional[str] = project_option(),
 ) -> None:
@@ -763,7 +773,7 @@ def add(
     _dispatch(
         NODE_ADD_COMMAND,
         NodeAddParams(
-            path=_normalize_path(path),
+            path=path,
             parent=parent,
             type=node_type,
             name=name if name is not None else node_type,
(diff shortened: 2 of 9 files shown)


qinhaihong-red:
## `review` skill pass: Standards / Spec

Reviewed PR #201 against `origin/main` using `git diff origin/main...HEAD`.
Spec sources: issue #199 and ADR-0015.

## Standards

Required: `node add` still derives the default node name in the CLI body instead of in the params model. ADR-0015 says normalization/derivation belongs in the params model so argv and `--params-json` produce identical params. Today `src/gda/cli.py:792` builds `NodeAddParams(..., name=name if name is not None else node_type)`, while `src/gda/models.py:453` leaves `NodeAddParams.name` as `None` when omitted. I confirmed the divergence: argv dispatches `name="Sprite2D"`, but equivalent `--params-json {"path": "...", "type": "Sprite2D"}` dispatches `name=None`, which is not the same operation input.

Required: several input-contract checks/derivers remain CLI-only after adding `--params-json`, so JSON input can bypass or diverge from the documented command contract. ADR-0015 says the params model is the input single source of truth and CLI functions should become thin argv-to-model adapters. The remaining CLI-only logic includes `script create`'s `--content`/`--extends` conflict at `src/gda/cli.py:1104`, `shader create`'s `--content`/`--shader-type` conflict at `src/gda/cli.py:1455`, and `script set` / `shader set` mode resolution at `src/gda/cli.py:1222` and `src/gda/cli.py:1782`. I confirmed `script create --params-json` accepts a JSON object containing both `content` and `extends_type`, while the argv form rejects `--content` plus `--extends`; the same bypass exists for `shader create`. For set commands, JSON can also supply a `mode` that conflicts with the other edit fields because `_resolve_set_mode` is not enforced by the model.

Judgement calls: none beyond the hard violations above.

## Spec

Required: issue #199 requires argv and `--params-json` outcomes to be identical, including normalization/derivation parity. The `node add` default-name case violates that acceptance criterion: the argv path derives `name` from `--type`, but the JSON path leaves the optional schema field as `null`.

Required: issue #199 also requires schema-invalid input to surface as a structured `GdaError`, and ADR-0015 makes the params model the single accepted input contract. Because the model does not encode the `script create` / `shader create` mutually-exclusive pairs, JSON input accepts combinations that have no valid argv equivalent and silently lets `content` win in the operation layer.

Summary: Standards found 2 required violations. Spec found 2 required violations. Worst issue: the model is not yet the single source of truth for all params-input semantics, so `--params-json` is not fully equivalent to argv.

qinhaihong-red:
## `code-review-and-quality` pass

Verdict: Request changes.

### Correctness

Required: schema-invalid JSON for any `NormalizedPath` field can still escape as a Python traceback instead of the structured `invalid_params` envelope required by issue #199. `src/gda/models.py:177` defines `normalize_path(path: str)`, but it is registered as a `BeforeValidator` at `src/gda/models.py:197`, so it receives raw JSON values before Pydantic has proven they are strings. For a non-string path, `src/gda/models.py:188` evaluates `"://" in path` and raises `TypeError`, which is not caught by `src/gda/headless.py:204` because that only handles `ValidationError`.

Repro:

```console
$ uv run gda scene create --params-json '{"path": 123, "root_type": "Node2D"}'
Traceback ...
TypeError: argument of type 'int' is not iterable
```

Expected per #199: a non-zero exit with `{"error":{"category":"operation","code":"invalid_params",...}}`, not a traceback. This affects every field converted to `NormalizedPath`, including `path`, `target`, `script`, and `output`. A regression test should cover a wrong-type path value, not just missing required fields.

Required: the params models do not yet encode all input invariants that the argv path enforces. `script create --params-json` accepts both `content` and `extends_type`; `shader create --params-json` accepts both `content` and `shader_type`; set commands can receive a `mode` inconsistent with the other edit fields. This is the same root issue as the Standards/Spec pass: after `--params-json`, model validation has to own those invariants or JSON dispatch is a wider API than argv.

### Other Axes

Readability/architecture: the central hook in `HeadlessCommand` is a reasonable fit for the existing `--schema` pattern, but the hard-coded model/CLI split above keeps the abstraction incomplete.

Security: no new secret-handling issue found. The stdin path for large payloads is present.

Performance: no performance issue found.

### Verification

Local checks run in an isolated `/tmp` clone:

- `uv run pytest tests/test_params_json.py tests/test_export_run_commands.py -q` -> 73 passed
- `GDA_GODOT=/Users/lelej/Applications/Godot.app/Contents/MacOS/Godot uv run pytest tests/test_e2e_params_json.py -q` -> 2 passed
- `uv run pytest -m 'not e2e' -q` -> 571 passed, 236 deselected

GitHub checks at review time: Python unit tests/package build passed; Godot e2e job was skipped by workflow conditions.

qinhaihong-red:
## Addressed (commit 2ccb6b2)

Both passes were right — `--params-json` was equivalent to argv only for path normalization; the model wasn't yet the single source for the *rest* of the input semantics, so JSON diverged from / was wider than the argv contract. Fixed by moving it all into the models. The argv error ergonomics are unchanged (bodies still pre-validate → Click exit 2); the model now owns each invariant, so the JSON path is validated identically.

**Correctness — wrong-typed `NormalizedPath` → traceback.** Fixed: `NormalizedPath` now uses `AfterValidator` (not `BeforeValidator`), so the value is validated as a `str` *first*. Your exact repro now returns the structured envelope:
```
$ gda scene create --params-json '{"path": 123, "root_type": "Node2D"}'
{"error":{"category":"operation","code":"invalid_params","message":"--params-json is not a valid params object: 1 validation error for SceneCreateParams\npath\n  Input should be a valid string ...
```
Regression test added (wrong-type path), covering all `NormalizedPath` fields' shape.

**`node add` default-name parity.** `NodeAddParams` now derives `name` from `type` model-side; the CLI body is thin (`name=name`). `--params-json` omitting `name` now dispatches `name="Sprite2D"`, matching argv.

**`script create` / `shader create` mutual exclusivity.** Now enforced by model validators (`content` × `extends_type` / `content` × `shader_type`), so `--params-json` rejects both-supplied with `invalid_params` rather than letting `content` silently win.

**`script set` / `shader set` mode.** The rule moved to `gda.models.resolve_set_mode` (single source); `ScriptSetParams`/`ShaderSetParams` derive `mode` from the edit fields (so a `--params-json` caller omits it, like argv) and reject inconsistent combinations. `cli._resolve_set_mode` now wraps that helper, translating its `ValueError` → Click `BadParameter` so **argv keeps exit 2** (your existing argv usage-error tests are unchanged), while `--params-json` surfaces `invalid_params`.

A note on the argv exit-code: I deliberately **kept argv usage errors at Click exit 2** (rather than unifying to `invalid_params`/exit 4) to avoid changing documented argv ABI — existing tests assert exit 2 for these. The model is the single source of the *invariant*; the argv body is a thin pre-validator that translates it to the established Click usage error. Happy to unify to structured exit-4 errors as a separate change if you'd prefer.

**Verification**: `pytest -m "not e2e"` → **577 passed** (6 new regression tests); full `pytest -m e2e` re-running to confirm no regression from the model changes; `import gda.cli` OK.

qinhaihong-red:
Full e2e regression run is green: `pytest -m e2e` → **235 passed, 1 skipped** (no regression from moving the input semantics into the models). Both tiers confirmed.

Issue #195, opened 2026-06-17
docs(gda-mcp): per-agent registration recipes (user + project scope)


qinhaihong-red:
## Parent

PRD #8 — gda-mcp. Anchored by ADR-0013 (packaging & launch).

## What to build

Shipped **registration recipes** for gda-mcp across the major agents, at BOTH user- and project-scope:

- **Claude Code** — `.mcp.json` (project) / `~/.claude.json` (local, user); `claude mcp add --scope user|project -- …`.
- **Codex** — global `~/.codex/config.toml` and repo-local `.codex/config.toml` (`[mcp_servers.gda-mcp]`).
- **Claude Desktop** — `claude_desktop_config.json` (single per-user file).
- **Cursor** — `.cursor/mcp.json` (project) / `~/.cursor/mcp.json` (global).

Each recipe gives a copy-pasteable config plus the launch command (`gda-mcp`, or zero-install `uvx --from "gda[mcp]" gda-mcp`).

**Document the known cross-agent constraint:** GUI-launched agents (Claude Desktop, Cursor) spawn servers with a **minimal PATH**, so their recipes use an **absolute path** to `uvx` / `gda-mcp` or inject `PATH` via `env`. Show how each agent pins the target project consistent with project-context resolution (#194), using only the neutral primitives `GDA_PROJECT` / `roots/list` / `cwd`: `GDA_PROJECT` in any agent's `env`/TOML, `${workspaceFolder}` injected into `GDA_PROJECT` on Cursor, `cwd`/TOML on Codex, and the client-advertised `roots` where available.

## Acceptance criteria

- [ ] User- and project-scope recipes exist for Claude Code, Codex, Claude Desktop, and Cursor, each with a copy-pasteable config and the launch command.
- [ ] The minimal-PATH constraint for GUI-launched agents is documented with the absolute-path / `PATH`-via-env mitigation.
- [ ] Each recipe shows (or links) how the target project is pinned for that agent, consistent with #194.

## Blocked by

- #193 (gda-mcp core)

qinhaihong-red:
## Reconcile (agent-neutral project resolution)

Project-pinning guidance updated per the revised ADR-0014: pin via the neutral primitives `GDA_PROJECT` / `roots/list` / `cwd` (no `CLAUDE_PROJECT_DIR` as a core input). Vendor names remain here — this is the adapter/recipes layer — and coverage stays multi-agent (Claude Code, Codex, Claude Desktop, Cursor).

Issue #194, opened 2026-06-17
feat(gda-mcp): project-context resolution via portable precedence


qinhaihong-red:
## Parent

PRD #8 — gda-mcp. Anchored by ADR-0014 (project-context resolution), ADR-0006 (project / path resolution).

## What to build

Resolve the target Godot project as **server context** — NOT a per-call tool parameter (the tool surface stays a faithful mirror of `--schema`; no synthesized `project` field). Resolve by the **portable precedence** (first hit wins), because the launch cwd is unreliable across MCP clients. The precedence names only agent-neutral primitives (gda's own env, the MCP protocol's own workspace signal, cwd); per-agent config belongs to the recipes (#195 / ADR-0013), not the core:

1. explicit `GDA_PROJECT` env (the neutral, agent-independent anchor),
2. MCP `roots/list` (read on the first tool call — a server→client request needs a live session — then cached; the protocol-native workspace signal),
3. process cwd (last-resort fallback).

Hand the resolved project to `gda` through its own **`GDA_PROJECT`** env channel (ADR-0006) — set on the subprocess, **not** a `--project` flag (which meta commands like `info` reject) — **only when the resolved candidate is a valid Godot project** (contains `project.godot`). Otherwise set no `GDA_PROJECT` and let `gda` apply its own ADR-0006 cwd / projectless resolution over the inherited env. Do not reject projectless operation; an op that requires a project surfaces `gda`'s own typed error, relayed unchanged.

## Acceptance criteria

- [ ] Resolution follows the precedence `GDA_PROJECT` → `roots/list` → cwd, first hit wins — the **full** precedence (including the `roots/list` read, snapshotted on the first tool call then cached) is implemented. The resolver names **no** agent-specific env var (agent-neutral; ADR-0014).
- [ ] A resolved candidate is passed as `gda --project <dir>` only if it contains `project.godot`; otherwise no `--project` is passed and `gda`'s own ADR-0006 resolution applies.
- [ ] No `project` field is injected into any tool's `inputSchema`.
- [ ] Fast tests cover the precedence ordering and the project-validation / no-forge behavior as pure functions.
- [ ] e2e: a project-scoped resolution drives a real tool against the resolved project (real engine).

Use the project's TDD skill. Per RULES.md DoD, fake/mock tests do not replace the e2e assertion above.

## Blocked by

- #193 (gda-mcp core)

qinhaihong-red:
## Reconcile (agent-neutral project resolution)

[...]

Issue #192, opened 2026-06-17
feat(gda): add aggregate-schema meta command (whole-surface dump)


qinhaihong-red:
## Parent

PRD #8 — gda-mcp. Anchored by ADR-0012 (tool-surface generation), ADR-0004 (`--schema`), ADR-0005 (command taxonomy).

## What to build

A new `gda` **meta command** that emits, in a single process, the whole command surface as one JSON document — for every command: `{name, description, input, output, error}` schemas — reusing the existing per-command `--schema` machinery (no engine spawn). This is the machine-readable manifest gda-mcp introspects once at startup to generate its tool surface with zero-touch sync.

Per ADR-0005 it is a **meta command** (top-level, ungrouped — sibling of `gda info`); implement as `gda schema` unless a strong reason favors a top-level `--schema-all` flag (record the chosen form in the PR).

Also **backfill weak command help / Pydantic `Field(description=...)`** where missing — these descriptions become agent-facing contract (they flow into MCP tool and parameter descriptions downstream).

## Acceptance criteria

- [ ] A single invocation emits one JSON document covering every command in every group, each entry carrying `name`, `description`, input schema, output schema, and error schema.
- [ ] `name` matches the deterministic MCP mapping basis (`<group> <command>`); `description` is present and non-empty for every command and every input field.
- [ ] No Godot engine process is spawned (pure schema emission, like `--schema`).
- [ ] Fast tests assert the aggregate JSON shape and full command coverage (mirroring existing `--schema` tests).
- [ ] One e2e test runs the real command and asserts the manifest covers the live command tree (every command reachable from the CLI appears).
- [ ] Backfilled help / field descriptions land for any command or field that lacked them.

Use the project's TDD skill. Per RULES.md DoD, fake/mock tests do not replace the e2e assertion above.

## Blocked by

None — can start immediately.

PR #203, opened 2026-06-18
feat(gda-mcp): generated stdio MCP server over the gda CLI (#193)

### LINKED ISSUE #193: feat(gda-mcp): generated stdio MCP server over the gda CLI (tracer bullet)

qinhaihong-red:
## What

The **gda-mcp tracer bullet** (#193): an `mcp` low-level **stdio** server that, at startup, introspects gda's aggregate schema dump (`gda schema`, #192) and registers **one MCP tool per command** — name `<group>_<command>` (`-`→`_`), `description` ← command help, `inputSchema`/`outputSchema` ← the command's schemas. A faithful mirror of the installed gda (ADR-0012), no codegen.

Each tool call shells out to the **same-distribution** gda via `[sys.executable, "-m", "gda"]` (Design decision 3 — never a PATH lookup), forwarding the tool input **verbatim** through `gda <group> <command> --params-json -` (object on stdin; ADR-0015), and maps the outcome mechanically off gda's exit code (ADR-0011):

- **exit 0** → the `--json` result dict; the SDK validates it against the tool's `outputSchema` and wraps it as `structuredContent`. gda-mcp does **not** re-validate — gda's result and its `outputSchema` share one Pydantic model, so conformance is by construction (Design decision 5).
- **non-zero** → `CallToolResult(isError=True)` carrying the full `{category, code, message, diagnostics}` `GdaError` envelope **verbatim and losslessly** as JSON content — never flattened to prose, kept out of `structuredContent`/`outputSchema`. The **can't-run / non-envelope edge** (e.g. `-m gda` import failure, a Click usage error) is synthesized by gda-mcp itself.

### Why low-level `Server`, not FastMCP
Schemas come **from gda** (not Python signatures), tools are discovered at runtime and served by **one generic dispatcher**, and the result/error channels need direct control — the inverse of FastMCP's "one tool = one decorated function" assumption.

### Packaging (ADR-0013)
Behind an optional `gda[mcp]` extra exposing a `gda-mcp` console script (also `uvx --from "gda[mcp]" gda-mcp`). `mcp` is **also** a dev dependency so `uv run pytest` has the SDK. gda **core never imports `mcp`**, and the entry point guards the import with an actionable "install `gda[mcp]`" message. Adds `gda/__main__.py` for the `-m gda` invocation.

## Architecture
- `src/gda/mcp/runner.py` — the single subprocess seam (`GdaRunner`), shared by startup introspection **and** per-tool dispatch; fast tests inject a fake (mirroring gda's own `FakeRunner`, one layer up).
- `src/gda/mcp/server.py` — introspect → register `types.Tool`s → one generic `call_tool` (`validate_input=False`, so gda owns input validation per ADR-0015).
- `src/gda/mcp/__init__.py` — lean guarded `main()`.
- `src/gda/__main__.py` — `python -m gda`.

## Slices (TDD)
| Slice | Form | Status |
|---|---|---|
| S1 info tracer real chain | e2e (in-memory client → in-process server → real gda → real Godot) | ✅ |
| S2 registration coverage | fast (fake seam + real dump) | ✅ |
| S3 error channel + edge | fast | ✅ |
| S4 packaging guard | fast | ✅ |
| S5 parametrized dispatch | fast | ✅ |
| S6 **e2e stdio gate** (non-negotiable) | e2e (real `gda-mcp` console script over stdio → real gda → real Godot: `info` + `scene_create`) | ✅ |

## Tests
- **598 fast** (21 new mcp fast tests) ✅
- **238 e2e** (1 pre-existing skip), incl. the new S1/S6 gates ✅ — run serially (~7 min)
- `uv build` produces a wheel containing `gda/mcp/*`, `gda/__main__.py`, and both `gda` + `gda-mcp` console scripts ✅

Per RULES.md DoD, the fake-`gda` seam does **not** count toward the e2e gate.

## Scope notes
Project-context resolution (#194) and per-agent registration recipes (#195) are **separate open issues** — out of scope here. The `info` half of S6 is #199-independent; `scene_create` exercises `--params-json` (#199, merged in 0.1.21).

Closes #193.

### DIFF
--- a/docs/adr/0011-gda-mcp-integration-mechanism.md
+++ b/docs/adr/0011-gda-mcp-integration-mechanism.md
@@ -21,15 +21,21 @@ gda-daemon exists in Phase 2 — the topology of {gda-mcp, gda, gda-daemon}.
 
 ## Decision
 
-**gda-mcp wraps `gda` as a subprocess.** It shells out to the installed `gda`
-console script and consumes only `gda`'s **public CLI ABI** — *not* the internal
-sentinel protocol ADR-0002 defines **between** `gda` and its headless operation
-subprocesses (which `gda` already parses away before emitting its public output, and
-which ADR-0010's native-CLI-mode operations do not emit at all). That public ABI is:
-the `--json` success output, the structured `GdaError` envelope on a non-zero exit
-(the uniform `GdaErrorEnvelope` of ADR-0004), the exit-code categories, and the
-`--schema` self-description (ADR-0004). Operations are invoked per-command as
-`gda <group> <command> --json …`; the **tool surface is enumerated from the aggregate
+**gda-mcp wraps `gda` as a subprocess.** It shells out to `gda` — invoked as
+`[sys.executable, "-m", "gda", …]`, the *same-distribution* binary paired with the
+running gda-mcp (the `[mcp]` extra), which is deterministic and cannot resolve a
+*wrong global* `gda`, rather than a PATH lookup of the console script (#193 Design
+decision 3; an optional `$GDA_BIN` escape hatch overrides it, ADR-0013) — and
+consumes only `gda`'s **public CLI ABI** — *not* the internal sentinel protocol
+ADR-0002 defines **between** `gda` and its headless operation subprocesses (which
+`gda` already parses away before emitting its public output, and which ADR-0010's
+native-CLI-mode operations do not emit at all). That public ABI is: the `--json`
+success output, the structured `GdaError` envelope on a non-zero exit (the uniform
+`GdaErrorEnvelope` of ADR-0004), the exit-code categories, and the `--schema`
+self-description (ADR-0004). Operations are invoked per-command as
+`python -m gda <group> <command> --params-json - --json`, forwarding the tool's
+input object **verbatim** on stdin (ADR-0015's structured params channel — so
+gda-mcp reconstructs no argv); the **tool surface is enumerated from the aggregate
 schema dump** (ADR-0012), *not* a per-command `--schema` fan-out. gda-mcp does **not**
 import `gda`'s Python modules or depend on any internal symbol.
 
--- a/docs/adr/0012-gda-mcp-tool-surface-generation.md
+++ b/docs/adr/0012-gda-mcp-tool-surface-generation.md
@@ -32,6 +32,22 @@ server starts — no codegen step, no release-pipeline coupling, no drift.
   flag — is a taxonomy detail left to the PRD). gda-mcp **never parses `gda --help`
   prose** to enumerate the surface.
 
+**The dump is the dispatchable-operation surface (refinement, #193).** It carries
+one entry per command that accepts the `--params-json` structured-input channel
+(ADR-0015) — i.e. has a backing operation gda-mcp can actually invoke. A
+*non-dispatchable* meta command (no backing operation, hence no `--params-json`)
+is excluded **at the source**, inside `gda`'s own surface walk, keyed on the one
+fact the command's registration already carries (its backing operation, `None`
+for such a command). `gda schema` is the only such command today: a pure
+self-describer, excluded from the surface it describes (re-listing it would be
+circular), while it still self-describes under its own `--schema`. This keeps the
+sole authority for "is this an MCP-dispatchable operation" in `gda`, so gda-mcp
+stays a pure transform that registers exactly what the dump reports and never
+advertises a tool it cannot fulfil — it holds no exclusion list of its own. This
+exclusion is a **soundness** property (gda-mcp can only serve what it can
+dispatch), distinct from the optional, user-facing group/tool *filtering* below
+(a surface-*size* control).
+
 **Expose the full surface; do not ship hand-curated profiles.** gda-mcp registers
 every tool the dump reports. It explicitly does **not** provide named tiers (full /
 lite / minimal). If a tool-count-limited client ever needs a smaller surface, the
--- a/docs/adr/0013-gda-mcp-packaging-and-launch.md
+++ b/docs/adr/0013-gda-mcp-packaging-and-launch.md
@@ -59,3 +59,15 @@ plus repo-local `.codex/config.toml`; `claude_desktop_config.json`; `.cursor/mcp
   (python-sdk#1520).
 - The `gda-mcp` entry point exists in every install; without the `[mcp]` extra it
   errors on a missing `mcp` import with an actionable message.
+- **Which `gda` gda-mcp shells out to, and the `GDA_BIN` escape hatch (#193).** By
+  default gda-mcp invokes `[sys.executable, "-m", "gda", …]` (ADR-0011) — the
+  interpreter running gda-mcp, hence the *same install*. This preserves this ADR's
+  single-version guarantee **structurally**: the adapter and the CLI are one
+  distribution, so no CLI/adapter skew is possible. An optional `GDA_BIN` env var
+  overrides that command for unusual deployment layouts (e.g. gda installed under a
+  different interpreter); it is an **explicit escape hatch**, analogous to `--godot`
+  overriding engine resolution (ADR-0006). When set, it can point at a *different*
+  `gda` and therefore transfers responsibility for CLI/adapter version alignment to
+  the operator — the structural no-skew guarantee holds only for the default. A
+  binary the override cannot launch is surfaced as a structured `isError`, never an
+  escaping exception (ADR-0011's can't-run edge).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,8 +13,21 @@ dependencies = [
     "typer>=0.26.7",
 ]
 
+[project.optional-dependencies]
+# The MCP adapter (gda-mcp) ships inside this one distribution, gated behind an
+# optional extra so a plain `gda` install stays lean — only `gda[mcp]` pulls the
+# MCP SDK (ADR-0013). Pinned to the v1 low-level Server API (decorator handlers,
+# camelCase Tool/CallToolResult fields) that gda-mcp targets; <2 fences off the
+# different v2 constructor API.
+mcp = [
+    "mcp>=1.12,<2",
+]
+
 [project.scripts]
 gda = "gda.cli:app"
+# Declared in every install (ADR-0013); without the `[mcp]` extra it errors on a
+# missing `mcp` import with an actionable "install gda[mcp]" message.
+gda-mcp = "gda.mcp:main"
 
 [build-system]
 requires = ["uv_build>=0.11.19,<0.12.0"]
@@ -24,6 +37,9 @@ build-backend = "uv_build"
 dev = [
     "jsonschema>=4.26.0",
     "pytest>=9.0.3",
+    # The MCP SDK is also a dev dependency (not just the `[mcp]` extra) so
+    # `uv run pytest` has it for the gda-mcp fast/e2e tiers (ADR-0013).
+    "mcp>=1.12,<2",
 ]
 
 [tool.pytest.ini_options]
--- a/src/gda/__main__.py
+++ b/src/gda/__main__.py
@@ -1,0 +1,14 @@
+"""Enable ``python -m gda`` (and thus ``sys.executable -m gda``).
+
+gda-mcp invokes the CLI as ``[sys.executable, "-m", "gda", …]`` for a
+deterministic, same-environment binary (ADR-0011, Design decision 3): it runs
+the *exact* gda paired with the running gda-mcp, never a wrong global ``gda`` a
+PATH lookup might resolve. This module is the ``-m gda`` entry point that makes
+that invocation work; it shares the one Typer ``app`` with the ``gda`` console
+script, so both entry points behave identically.
+"""
+
+from gda.cli import app
+
+if __name__ == "__main__":
+    app()
--- a/src/gda/mcp/__init__.py
+++ b/src/gda/mcp/__init__.py
@@ -1,0 +1,48 @@
+"""gda-mcp: a generated stdio MCP server over the gda CLI (issue #193).
+
+gda's MCP adapter (ADR-0011/0012/0013): it introspects gda's aggregate schema
+dump and exposes one MCP tool per command, shelling out to the installed gda for
+each call over gda's public CLI ABI. It ships inside the one ``gda`` distribution
+behind the optional ``gda[mcp]`` extra — gda core never imports ``mcp``, and this
+package's :func:`main` entry point guards the import so ``gda-mcp`` without the
+extra fails with an actionable message rather than an opaque ImportError.
+
+This module stays deliberately lean (no ``mcp`` import at load time): the import
+guard in :func:`main` must run *before* any SDK-dependent code, so all of that
+lives in :mod:`gda.mcp.server`, imported only after the guard passes.
+"""
+
+import sys
+
+# The actionable failure when the optional ``mcp`` SDK is absent (ADR-0013): the
+# ``gda-mcp`` console script is declared in every install, but only ``gda[mcp]``
+# installs the SDK it needs. Point the user at the fix instead of letting a bare
+# ImportError escape.
+_MISSING_MCP_MESSAGE = (
+    "gda-mcp requires the optional 'mcp' extra, which is not installed.\n"
+    "Install it with:    pip install 'gda[mcp]'\n"
+    "or run without installing:    uvx --from 'gda[mcp]' gda-mcp\n"
+)
+
+
+def main() -> int:
+    """The ``gda-mcp`` console-script entry point (ADR-0013).
+
+    Lean by design: it guards the optional ``mcp`` import (so a plain ``gda``
+    install fails with an actionable message, not an opaque traceback), then
+    hands off to the server, which owns all the SDK-dependent logic. Returns the
+    process exit code; ``console_scripts`` wraps the return value in
+    ``sys.exit``.
+    """
+    try:
+        import mcp  # noqa: F401
+    except ImportError:
+        sys.stderr.write(_MISSING_MCP_MESSAGE)
+        return 1
+    # Imported lazily, AFTER the guard: gda.mcp.server imports the SDK at module
+    # load, so importing it before the guard would surface the bare ImportError
+    # this guard exists to replace.
+    from gda.mcp.server import run_stdio
+
+    run_stdio()
+    return 0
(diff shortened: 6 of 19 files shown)


qinhaihong-red:
## Review pass: `review` skill

### Standards

Hard violation / standards-doc drift:

- `src/gda/mcp/runner.py:49-59` and `src/gda/__main__.py:1-14` default gda-mcp to `[sys.executable, "-m", "gda"]`. ADR-0011 still says gda-mcp "shells out to the installed `gda` console script" and invokes commands as `gda <group> <command> --json ...` (`docs/adr/0011-gda-mcp-integration-mechanism.md`, Decision). Issue #193's Design decision 3 now asks for `python -m gda`, so this looks like a real ADR/spec drift: either ADR-0011 should be updated to record the same-distribution `python -m gda` launch ABI, or the implementation should follow the ADR.

Judgement call:

- `src/gda/mcp/runner.py:56-59` adds `GDA_BIN` as an override. That may be useful operationally, but ADR-0013's packaging invariant is one distribution with "single version" and no CLI/adapter version skew (`docs/adr/0013-gda-mcp-packaging-and-launch.md`, Decision). If this override is intentional, the ADR should document the exception and its version-skew tradeoff.

### Spec

Required:

- `src/gda/mcp/server.py:173-189` registers the aggregate dump's `schema` command as an MCP tool, but `src/gda/mcp/server.py:132-137` dispatches every tool through `--params-json - --json`. `gda schema` is the documented `--params-json` exception, so the generated `schema` tool is exposed but not callable through the required ABI. I reproduced this locally: calling MCP tool `schema` returns `isError=True` with adapter code `gda_invocation_failed` and diagnostics ending in `No such option: --params-json`. This conflicts with issue #193's requirements to "register one MCP tool per command" and to forward calls through the public CLI ABI.

- `src/gda/mcp/runner.py:73-87` only returns a `GdaResult` when `subprocess.run()` itself succeeds. If the gda subprocess cannot be launched, for example a bad `GDA_BIN` override, the exception escapes instead of being converted into `CallToolResult(isError=True)` with a structured JSON envelope. Issue #193 says the "can't-run / non-envelope edge" should be synthesized by gda-mcp itself.

### Verification

- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_mcp_registration.py tests/test_mcp_dispatch.py tests/test_mcp_errors.py tests/test_mcp_packaging.py` -> 21 passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_e2e_mcp_tracer.py tests/test_e2e_mcp_stdio.py -m e2e` -> 3 passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv build` -> built sdist and wheel; wheel contains `gda/mcp/*`, `gda/__main__.py`, and both `gda` + `gda-mcp` console scripts

Summary: Standards found 2 docs-alignment findings; Spec found 2 correctness gaps. Worst issue: the registered `schema` MCP tool currently cannot execute successfully.

qinhaihong-red:

### Correctness

Required: the registered `schema` MCP tool is not actually callable.

`build_server()` registers all 44 commands from the aggregate manifest, including `schema`, but `dispatch()` always calls:

```text
[*argv, "--params-json", "-", "--json"]
```

That works for normal headless commands, but `gda schema` is the documented `--params-json` exception. I reproduced this locally by calling MCP tool `schema`; it returns `isError=True` with adapter code `gda_invocation_failed` and diagnostics ending in `No such option: --params-json`. Either do not expose `schema` as a tool, or special-case its dispatch so the registered tool succeeds against its advertised output schema.

Required: real subprocess launch exceptions are not converted into MCP error results.

`SubprocessGdaRunner.run()` lets `subprocess.run()` exceptions escape. With `GDA_BIN=/does/not/exist`, `dispatch(..., ["info"], {})` raises `FileNotFoundError` instead of returning `CallToolResult(isError=True)` with structured JSON content. Since this PR adds `GDA_BIN`, the runner should either catch launch failures and return a raw failure result that `dispatch()` can synthesize, or remove the override.

### Readability / Architecture

The main shape is coherent: a single `GdaRunner` seam, generated registration from `gda schema`, and one dispatcher match the intended adapter boundary. No blocking readability findings beyond the two edge-case holes above.

### Security

No new high-risk secret/auth/data-flow issue found. The subprocess boundary is still the intended design for this trusted-project toolchain.

### Performance

No blocking performance issue found. Startup does one schema subprocess, and per-tool calls mirror the accepted subprocess-adapter design.

### Verification

Good coverage exists for registration, normal dispatch, structured error relay, packaging guard, and the real stdio `info` / `scene_create` gate. Please add regression coverage for:

- calling the generated `schema` MCP tool, or explicitly proving it is intentionally filtered out;
- a real `SubprocessGdaRunner` launch failure path, such as a bad `GDA_BIN`, returning structured `isError` instead of raising.

Local verification run:

- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_mcp_registration.py tests/test_mcp_dispatch.py tests/test_mcp_errors.py tests/test_mcp_packaging.py` -> 21 passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_e2e_mcp_tracer.py tests/test_e2e_mcp_stdio.py -m e2e` -> 3 passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv build` -> succeeded

Verdict: request changes for the two correctness gaps above.

qinhaihong-red:
## Follow-up (commit bc5d0dd): restrict the dump to the dispatchable-operation surface

A review question surfaced: the `schema` meta command was being registered as an MCP tool (faithful mirror of the dump), but `gda schema` has no `--params-json`, so the tool (a) duplicated the protocol's own `list_tools` and (b) always failed when called.

**Decision (Plan A):** exclude non-dispatchable meta commands **at the source** — gda's surface walk now skips a leaf whose backing operation is `None` (`gda_command is None`, the same fact the `--params-json` dispatch keys on; verified 1:1 with `--params-json` option presence across all 44 commands). So `gda schema` now emits the **dispatchable-operation surface**.

Why this shape:
- **gda-mcp code is unchanged** — it still registers exactly what the dump reports, now only commands it can invoke; it holds no exclusion list and stays a pure schema→tool transform. The authority for "is this an MCP-dispatchable operation" lives solely in gda.
- It's a **soundness** property (never advertise a tool you can't fulfil), distinct from the optional group/tool *filtering* ADR-0012 reserves for surface size.
- `gda schema` still self-describes under its own `--schema`; it's just no longer an entry in the surface it describes (re-listing the describer would be circular).

**Touch points:** `surface.py` (the skip + docs), `test_schema_aggregate.py` (coverage cross-check now keyed on the independent `--params-json`-option signal + a focused exclusion test), and an ADR-0012 refinement. Tests: **599 fast** + **3 mcp e2e** green; gda-mcp's own tests pass unchanged.

qinhaihong-red:
## Review response — all 4 findings addressed

Thanks for the thorough pass. Resolutions below (3 new commits + the earlier `bc5d0dd`).

### Standards

**Std-1 — ADR-0011 drift (`-m gda` / `--params-json`).** Valid. ADR-0011's Decision still described the *sketch* ("installed `gda` console script", `gda <group> <command> --json`). Amended to record the implemented launch ABI: same-distribution `[sys.executable, "-m", "gda", …]` (Design decision 3) + verbatim `--params-json - --json` dispatch (ADR-0015). → `docs(adr)` 6830b0f

**Std-2 — `GDA_BIN` vs ADR-0013 single-version.** Kept `GDA_BIN` (it's in #193 Design decision 3) and documented it in ADR-0013 as an explicit escape hatch (analogous to `--godot`): the default `-m gda` preserves the single-version guarantee *structurally*; `GDA_BIN` can point at a different gda and so transfers version-alignment responsibility to the operator. → `docs(adr)` 6830b0f

### Spec

**Spec-1 — `schema` tool exposed but uncallable.** Fixed at the source in `bc5d0dd` (pushed before this review pass): non-dispatchable meta commands are excluded from the dump, so gda-mcp never registers `schema` (it stays a pure transform — no exclusion list of its own; the authority lives in gda). I took your *"do not expose `schema`"* option. Added the MCP-layer regression you asked for (assert `schema` is not among registered tools), alongside the dump-level exclusion test. → `test(gda-mcp)` 9d72f0e

**Spec-2 — launch exception escapes.** Valid bug. `SubprocessGdaRunner.run` now catches `OSError` and returns a non-zero `GdaResult` naming the command (mirroring gda's own `gda.runner.launch`), so `dispatch` synthesizes `CallToolResult(isError=True)` instead of letting a traceback cross the MCP boundary. Regression tests cover both the seam and the full `dispatch` path with a bad `GDA_BIN`. → `fix(gda-mcp)` 8736879

### Verification
- `uv run pytest -m "not e2e"` → **602 passed** (+3: 2 launch-failure + 1 schema-exclusion)
- `uv run pytest tests/test_e2e_mcp_tracer.py tests/test_e2e_mcp_stdio.py -m e2e` → **3 passed**

PR #205, opened 2026-06-19
feat(gda-mcp): project-context resolution via portable precedence (#194)

### LINKED ISSUE #194: feat(gda-mcp): project-context resolution via portable precedence

qinhaihong-red:
## Summary

Resolves the gda-mcp server's target Godot project (ADR-0014) and hands it to the `gda` subprocesses, so a project-scoped registration drives operations against the right project automatically.

Closes #194.

### Two ADR-0014 corrections (both reconciled across PRD #8 / #194 / #195)

1. **Agent-neutral resolution.** Dropped `CLAUDE_PROJECT_DIR` (a Claude Code-specific env var, redundant with the MCP-native `roots/list`). The precedence is now the agent-neutral **`GDA_PROJECT` → `roots/list` → cwd**, and the ADR's normative text names no specific agent. Per-agent specifics belong to the registration recipes (ADR-0013), which stays multi-agent. gda-mcp is a consumer-agnostic service; vendor coupling in the core was the wrong shape.
2. **Mechanism corrected to the `GDA_PROJECT` env channel.** The prior "pass `--project` to every subprocess" wording was impossible — meta commands like `info` reject `--project` outright. The resolved project now rides gda's own `GDA_PROJECT` env channel (ADR-0006): project-taking domain commands consume it, meta commands ignore it, and gda-mcp needs **zero per-command knowledge**.

## How it works

- A pure resolver `resolve_project_dir(env, roots, cwd)` owns the full precedence + `project.godot` validation. An explicitly pinned but invalid `GDA_PROJECT` resolves to `None` without falling through (ADR-0006 strict), so gda surfaces its own typed error.
- `roots/list` is a server→client request, so resolution is deferred to the first tool call (needs a live session) and cached — one server targets one project. It is capability-gated and fail-safe to "no roots", never a hard dependency.
- The resolved dir is set on the subprocess env via the `GdaRunner` seam; `project=None` leaves the inherited env untouched.
- No `project` field is synthesized into any tool's inputSchema — the surface stays a faithful mirror of `--schema` (ADR-0012/0014).

## Built TDD in four serial layers (red→green→refactor, commit per layer)

| Layer | What | Tests |
|---|---|---|
| L1 | pure resolver (precedence + validation + no-forge) | 7 fast |
| L2 | inject via `GDA_PROJECT` env through the runner seam | 5 fast |
| L3 | `roots/list` live wiring (lazy, cached, capability-gated) | 4 fast |
| L4 | e2e gate: real MCP → gda → engine | 3 e2e |

## Acceptance criteria (#194)

- [x] Full precedence `GDA_PROJECT → roots/list → cwd`, first hit wins (incl. the live roots/list read)
- [x] Candidate passed only when it holds `project.godot`; otherwise gda's own ADR-0006 resolution applies
- [x] No `project` field injected into any tool's inputSchema (explicit guard test)
- [x] Fast pure-function tests for precedence + validation / no-forge
- [x] e2e: project-scoped resolution drives a real tool against the resolved project (real engine)

## Testing

- **619 fast tests** pass (`pytest -m "not e2e"`).
- **6 mcp e2e tests** pass against a real Godot engine (3 new + 3 existing, no regression). The strongest is roots → real engine: gda cannot read MCP roots itself, so a `res://` write landing in the advertised-root project proves gda-mcp resolved + injected. A meta command (`info`) tolerating a pinned project validates mechanism D where a `--project` flag would have errored.

### DIFF
--- a/docs/adr/0014-gda-mcp-project-context-resolution.md
+++ b/docs/adr/0014-gda-mcp-project-context-resolution.md
@@ -16,8 +16,13 @@ breaking either decision.
 **The target project is server context, not a per-call tool parameter.** gda-mcp does
 **not** inject a `project` field into any tool's inputSchema; the tool surface stays a
 faithful mirror of `--schema` (ADR-0012), and `gda` still receives the project exactly
-as ADR-0006 specifies. gda-mcp resolves one target project for the server and passes
-it as `--project` to every `gda` subprocess it spawns.
+as ADR-0006 specifies. gda-mcp resolves one target project for the server and hands it
+to its `gda` subprocesses through gda's own **`GDA_PROJECT`** channel (ADR-0006) — set
+on the subprocess environment, **not** a `--project` flag. The env channel applies
+uniformly: project-taking domain commands consume it, while meta commands (`info`) that
+take no project simply ignore it — so gda-mcp needs no per-command knowledge of which
+commands accept a project, and a `--project` flag (which meta commands reject outright)
+is never forged.
 
 **Resolution is by a portable precedence, with cwd only as a last resort** — because
 the launch working directory is not reliably the project across MCP clients (clients
@@ -32,16 +37,18 @@ gda-mcp resolves, first hit wins:
 2. the MCP **`roots/list`** the client advertises — the protocol-native workspace signal,
 3. the process **cwd**, as a last-resort fallback.
 
-**A candidate is asserted as an explicit project only when it is a valid Godot
-project** (it contains `project.godot`). gda-mcp does **not** promote an unvalidated
-cwd or root into an explicit `--project`, because ADR-0006 gives `--project` /
-`$GDA_PROJECT` strict semantics (must be a real project) while leaving cwd lenient.
-So an explicit `GDA_PROJECT`, and any validated `roots`/cwd candidate, are passed as
-`gda --project <dir>`; if nothing resolves to a valid
-project, gda-mcp passes **no** `--project` and lets `gda` apply its own ADR-0006 cwd
-resolution — including running projectless when the cwd is not a project. gda-mcp
-does not itself reject projectless operation; an op that requires a project surfaces
-`gda`'s own typed error, relayed unchanged.
+**A candidate is asserted as the project only when it is a valid Godot project** (it
+contains `project.godot`). gda-mcp does **not** promote an unvalidated cwd or root,
+because ADR-0006 gives `$GDA_PROJECT` strict semantics (must be a real project) while
+leaving cwd lenient. So a validated `roots` or cwd candidate is exported as
+`GDA_PROJECT=<dir>` on the subprocess; if nothing resolves to a valid project, gda-mcp
+sets **no** `GDA_PROJECT` and lets `gda` apply its own ADR-0006 resolution over the
+inherited environment — including running projectless when neither `$GDA_PROJECT` nor
+the cwd is a project. An explicitly set but **invalid** `GDA_PROJECT` is left untouched
+in the inherited environment (not silently shadowed by a roots/cwd candidate), so `gda`
+surfaces its own strict typed error for project-taking commands. gda-mcp does not itself
+reject projectless operation; an op that requires a project surfaces `gda`'s own typed
+error, relayed unchanged.
 
 ## Considered options
 
--- a/docs/adr/0014-gda-mcp-project-context-resolution.md
+++ b/docs/adr/0014-gda-mcp-project-context-resolution.md
@@ -68,8 +68,12 @@ error, relayed unchanged.
 - A **user-scoped** server with nothing pinned falls back through `roots/list` / cwd;
   it works for the single-active-project workflow but cannot, in this first delivery,
   follow a user juggling several projects in one session.
-- The startup `roots/list` read (precedence 2) is part of the resolution contract,
-  not a staged extra — the first delivery implements the full precedence above.
+- The `roots/list` read (precedence 2) is part of the resolution contract, not a
+  staged extra — the first delivery implements the full precedence above. Because
+  `roots/list` is a server→client request that needs a live session, it cannot run
+  at process startup; resolution is **snapshotted on the first tool call** and then
+  cached for the server's lifetime (one server : one project). Re-resolving when the
+  client's active root changes is the deferred `roots/list_changed` path below.
 - **No vendor coupling in the core.** This ADR names no specific agent; per-agent
   registration — how each agent's config sets the server's `env` / launch command and
   pins the project — is the registration recipes' concern (ADR-0013), which must cover
--- a/docs/adr/0014-gda-mcp-project-context-resolution.md
+++ b/docs/adr/0014-gda-mcp-project-context-resolution.md
@@ -20,25 +20,24 @@ as ADR-0006 specifies. gda-mcp resolves one target project for the server and pa
 it as `--project` to every `gda` subprocess it spawns.
 
 **Resolution is by a portable precedence, with cwd only as a last resort** — because
-the launch working directory is not reliably the project across agents (Claude Code
-ignores a configured `cwd`, claude-code#17565; Cursor / Claude Desktop leave it at the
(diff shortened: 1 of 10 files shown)


qinhaihong-red:
## Review pass 1: `review` (Standards + Spec)

### Standards

Hard violations: none found.

Judgment calls: none found.

The diff aligns with the documented standards I checked: project context remains server context rather than a tool parameter per `docs/adr/0014-gda-mcp-project-context-resolution.md`; project injection uses `GDA_PROJECT` rather than forged `--project`; `gda-mcp` keeps the schema-derived tool surface intact per `docs/adr/0012-gda-mcp-tool-surface-generation.md`; dispatch still forwards tool inputs verbatim over `--params-json -` per `docs/adr/0015-structured-params-input-abi.md`; and the adapter does not import `gda` internals, consistent with `docs/adr/0011-gda-mcp-integration-mechanism.md`.

I did not re-check formatting or other machine-enforced tooling rules; `git diff --check` was clean.

### Spec

Finding: `roots/list` is cached, but not read at startup.

Spec quote: “MCP `roots/list` (static read at startup; the protocol-native workspace signal)” and the ADR text in this PR still says “The startup `roots/list` read (precedence 2) is part of the resolution contract.”

The implementation resolves roots lazily in `call_tool` via `_ProjectResolver.resolve(server.request_context.session)`, then caches the first result. That gives a single static result after the first tool call, but it is not a startup read; any root change before the first tool call would affect the selected project. This is partial against the “static read at startup” requirement and leaves the updated ADR wording inconsistent with behavior.

No other spec gaps found. The diff appears to preserve server-context project resolution, avoids a synthesized `project` inputSchema field, removes agent-specific env vars, validates candidates by `project.godot`, and uses `GDA_PROJECT` as the corrected handoff mechanism.

Spec ambiguity: the original issue says resolved candidates are passed as `gda --project <dir>`, while the updated ADR/code use subprocess `GDA_PROJECT`; because the PR explicitly corrects that mechanism to the env channel, I did not count that as a finding.

Summary: Standards: 0 findings. Spec: 1 finding. Worst issue: the implementation and ADR/spec wording disagree on whether `roots/list` is read at startup or lazily on first tool call.

qinhaihong-red:
## Review pass 2: `agent-skills:code-review-and-quality`

### Findings

Required: Align the `roots/list` timing contract with the implementation before merge.

`docs/adr/0014-gda-mcp-project-context-resolution.md:71` still says the “startup `roots/list` read” is part of the resolution contract, and issue #194 specifies a “static read at startup.” The code in `src/gda/mcp/server.py:213` instead waits until the first tool call, reads `roots/list` through the live request context, then caches that first-call result. That is a reasonable implementation shape for the MCP SDK, but it is a different lifecycle contract: a root change between server initialization and the first tool call changes the selected project, even though the ADR says the project is snapped at startup. Either move resolution to an initialization/startup point, or update ADR-0014/#194 wording to explicitly define a first-tool-call snapshot and add a regression test for that timing.

### Multi-axis Review

Correctness: The core project precedence, validation, `GDA_PROJECT` subprocess injection, inherited invalid `GDA_PROJECT` behavior, and “no synthesized `project` inputSchema field” are covered and behaved correctly in the tests I ran. The only correctness concern is the timing contract above.

Readability & simplicity: The new resolver is small and pure; the runner seam extension is straightforward; comments are mostly explaining contract boundaries rather than restating code.

Architecture: The change fits ADR-0011/0012/0014 well: gda-mcp still wraps gda as a subprocess, keeps per-command knowledge out of dispatch, and uses gda’s own env channel rather than forging `--project`.

Security: No secrets or new external dependency risk found. Root paths are accepted only as local file roots from the MCP SDK’s `Root.uri` type and are only promoted when `project.godot` exists.

Performance: The roots request is cached after resolution, so this does not add per-call `roots/list` overhead after the first dispatch.

### Verification

- `git diff --check`
- `uv run pytest -m "not e2e" -q` → 619 passed
- `GDA_GODOT=/Users/lelej/Applications/Godot.app/Contents/MacOS/Godot uv run pytest tests/test_e2e_mcp*.py -q -m e2e` → 6 passed
- Targeted pre-checks also passed: 26 MCP fast tests and 3 new project-context e2e tests.

### Verdict

Request changes until the `roots/list` timing contract is made consistent between docs/spec, tests, and implementation.

qinhaihong-red:
## Addressed: `roots/list` timing contract (commit 9d159c1)

Thanks — the finding is correct, and I took **option (b): make the spec describe the actual lifecycle**, for these reasons:

- **Option (a) — a true startup read — isn't cleanly possible.** `roots/list` is a *server→client* request, so it needs a live session. There is no pre-first-call hook in the low-level SDK to block on before serving; the session is only reachable from within a request (`request_context`). The lazy-on-first-call shape (which the review acknowledges as "a reasonable implementation shape for the MCP SDK") is the earliest the protocol allows.
- **Neither snapshot moment handles a root changing before the snapshot.** The only fully-correct answer to a changing active root is dynamic re-resolution via `roots/list_changed`, which ADR-0014 **explicitly defers** out of the first delivery. So the disagreement was purely about *which moment* we snapshot and *how we describe it* — not about behavior.
- The **code already implemented and documented** the first-call snapshot + cache (`_ProjectResolver` docstring, `build_server` comment). Only the ADR-0014 / #194 wording lagged — "static read at startup" was leftover phrasing I carried over during the earlier reconcile.

What changed:

1. **ADR-0014** and **#194** now state the lifecycle precisely: `roots/list` is read **on the first tool call** (a server→client request needs a live session) and **cached for the server's lifetime** (one server : one project); re-resolving on a changed root is the deferred `roots/list_changed` path.
2. **New regression test** `test_project_is_snapshotted_on_first_call_then_cached` pins the contract: a second tool call advertising a *different* root reuses the first snapshot.
3. While there, fixed a residual `--project` flag phrasing in #194's "What to build" (the mechanism-D `GDA_PROJECT` correction the PR already made in the ADR/code) — the spec ambiguity you noted.

Tests: 620 fast + 6 mcp e2e green (no src change, so e2e behavior is unchanged).

PR #210, opened 2026-06-20
ci(release): publish gda to PyPI via Trusted Publishing (#207)

### LINKED ISSUE #207: chore(release): publish gda (incl. [mcp] extra) to PyPI
## What to build

`gda` is not published to PyPI (`pypi.org/pypi/gda` → 404). The release pipeline (release-please, ADR-0007) builds wheels with `uv build` and attaches them to a GitHub Release only — there is no PyPI publish step. As a result the canonical zero-install / registration commands assumed by ADR-0013 and #195 — `uvx --from "gda[mcp]" gda-mcp` and `pip install "gda[mcp]"` — do **not** work today; #195's recipes ship using a git-source install for now (`uvx --from "gda[mcp] @ git+https://github.com/aigengame/godot-agent" gda-mcp`).

Extend the existing release-please publish flow to also publish the built distributions (sdist + wheel) to PyPI via **Trusted Publishing (OIDC, no long-lived token)**, idempotent and re-runnable like the current GitHub-release upload. Record the decision in a distribution ADR. Once live, the canonical commands work and the recipes can drop the git-source form.

Split out from #195 (option B); phase-orthogonal packaging work, not a delivery phase.

## Acceptance criteria

- [ ] A distribution ADR records the decision to publish to PyPI and the auth mechanism (Trusted Publishing / OIDC; no long-lived token stored in CI).
- [ ] The `gda` project name is claimed on PyPI (or an alternative name is decided if it is taken/squatted).
- [ ] The release workflow publishes sdist + wheel to PyPI on a cut release, idempotent / re-runnable (mirroring the current GitHub-release publish).
- [ ] From a clean environment after a release, `pip install "gda[mcp]"` and `uvx --from "gda[mcp]" gda-mcp` both work.
- [ ] ADR-0013 and the #195 registration recipes are reconciled from the git-source form back to the canonical `uvx --from "gda[mcp]"` / `pip install "gda[mcp]"` form.

## Blocked by

None — can start immediately. Requires human action (claiming the PyPI name and configuring the Trusted Publisher), hence `ready-for-human`.

qinhaihong-red:
## What

Add a PyPI publish step to the release pipeline and record the decision in **ADR-0016**, so the canonical install/registration commands promised by ADR-0013 (`pip install "gda[mcp]"`, `uvx --from "gda[mcp]" gda-mcp`) can work. Today `gda` is not on PyPI and the #195 recipes ship a git-source stopgap.

Split out of #207 (the workflow + ADR slice). **Does not close #207** — the PyPI-side human setup, first publish, and doc reconciliation remain.

## Changes

- **`.github/workflows/release.yml`** — in the existing `github-release` job:
  - `pypa/gh-action-pypi-publish@release/v1` step with `skip-existing: true`, inserted **after `uv build` and before the GitHub Release is un-drafted**.
  - `id-token: write` permission + `environment: pypi` to scope the OIDC trust.
- **`docs/adr/0016-publish-to-pypi-via-trusted-publishing.md`** — the distribution decision.

## Key design points

- **Trusted Publishing (OIDC), no stored token** — honours ADR-0007's "no long-lived credential" stance.
- **Ordering** — publish runs before the un-draft that creates the git tag, so PyPI falls under ADR-0007's *tag-as-commit-point* recovery model: a failure leaves a tag-less draft that the release-PR gate flags and **"Re-run failed jobs"** converges.
- **Idempotent** — `skip-existing: true` (PyPI files are immutable; a re-run from the same tagged commit skips what already landed), mirroring the `--clobber` GitHub upload.

## ⚠️ Merge ordering

This step is **inert until the PyPI pending publisher is configured** (one-time human setup, `ready-for-human` on #207). Complete the PyPI/GitHub-Environment setup **before** merging — otherwise the next cut release fails at this step and leaves a recoverable but noisy tag-less draft.

One-time human setup (also in ADR-0016):
1. PyPI → **Add a pending publisher** for project `gda`: owner `aigengame`, repo `godot-agent`, workflow `release.yml`, environment `pypi`.
2. Create GitHub Environment `pypi` (no protection rules → stays fully automated).

## Follow-up (after first successful publish)

Reconcile the git-source form back to the canonical commands in `docs/gda-mcp-registration.md`, `README.md`, and ADR-0013's prose — done only after a real PyPI release exists so docs never advertise a 404 command.

Refs: #207

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -138,12 +138,22 @@ jobs:
   publish-pypi:
     name: Publish to PyPI (Trusted Publishing)
     runs-on: ubuntu-latest
-    needs: build-release
+    needs:
+      - cut-release
+      - build-release
+    # Run on any non-cancelled post-cut outcome — the guard step below decides
+    # pass/fail — instead of letting a failed build-release SKIP this job.
+    # "Re-run failed jobs" reliably re-runs FAILED jobs and their dependents, but
+    # whether it re-runs a job that was SKIPPED because an upstream failed is
+    # undocumented; relying on it would strand ADR-0007's recovery path. Turning
+    # an upstream failure into an explicit failure of THIS job keeps the whole
+    # post-cut tail re-runnable. Gated on release_created so ordinary pushes that
+    # cut nothing still skip it.
+    if: ${{ !cancelled() && needs.cut-release.outputs.release_created == 'true' }}
+    #
     # Runs BEFORE publish-github-release. Un-drafting the GitHub Release (the next
-    # job) is what creates the git tag, and ADR-0007's recovery model treats the
-    # tag as the atomic "release succeeded" signal: every load-bearing step runs
-    # before it, so a PyPI failure leaves a tag-less draft that the release-PR gate
-    # flags and "Re-run failed jobs" converges (ADR-0016).
+    # job) is what creates the git tag — ADR-0007's atomic "release succeeded"
+    # signal — so a PyPI failure here means no tag is minted (ADR-0016).
     #
     # This job holds id-token: write and NOTHING else, and runs no project, test,
     # or build code — it only downloads the prebuilt artifact and publishes it.
@@ -159,6 +169,16 @@ jobs:
       name: pypi
       url: https://pypi.org/p/gda
     steps:
+      # Fail (don't silently skip) when the build did not succeed, so this is a
+      # FAILED job that "Re-run failed jobs" re-runs — not a skipped one whose
+      # re-run is undocumented.
+      - name: Require the build to have succeeded
+        if: ${{ needs.build-release.result != 'success' }}
+        run: |
+          echo "build-release did not succeed (result: ${{ needs.build-release.result }})." >&2
+          echo "Failing so 'Re-run failed jobs' re-runs this job once the build is repaired." >&2
+          exit 1
+
       - name: Download distributions
         uses: actions/download-artifact@v8.0.1
         with:
@@ -182,13 +202,30 @@ jobs:
     needs:
       - cut-release
       - publish-pypi
-    # Only contents: write here, no id-token. Runs after PyPI publish succeeds;
-    # the un-draft below creates the tag — ADR-0007's commit point.
+    # Same run-on-failure rationale as publish-pypi: run on any non-cancelled
+    # post-cut outcome so a failed publish-pypi makes THIS job fail (via the guard
+    # below) rather than skip — keeping it in the set "Re-run failed jobs" re-runs,
+    # so recovery drives through to tag creation once PyPI is repaired (ADR-0007 /
+    # ADR-0016). Gated on release_created so ordinary pushes skip it.
+    if: ${{ !cancelled() && needs.cut-release.outputs.release_created == 'true' }}
+    # Only contents: write here, no id-token. The un-draft below creates the tag
+    # — ADR-0007's commit point — and the guard ensures it runs only after PyPI
+    # has actually published.
     permissions:
       contents: write
     env:
       RELEASE_TAG: ${{ needs.cut-release.outputs.tag_name }}
     steps:
+      # Fail before the un-draft when PyPI did not publish: this both prevents a
+      # tag from being minted without a PyPI release AND makes this a FAILED (not
+      # skipped) job that "Re-run failed jobs" reliably re-runs.
+      - name: Require the PyPI publish to have succeeded
+        if: ${{ needs.publish-pypi.result != 'success' }}
+        run: |
+          echo "publish-pypi did not succeed (result: ${{ needs.publish-pypi.result }})." >&2
+          echo "Failing before the un-draft so no tag is created; 'Re-run failed jobs' re-runs this once PyPI is repaired." >&2
+          exit 1
+
       - name: Download distributions
         uses: actions/download-artifact@v8.0.1
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,6 +67,18 @@ jobs:
       needs.cut-release.outputs.release_created == 'true'
     permissions:
       contents: write
+      # Trusted Publishing mints a short-lived, scoped PyPI upload token from an
+      # OIDC token instead of a stored API secret (ADR-0016). The publish step
+      # below needs to request it.
+      id-token: write
+    # Scope the OIDC trust to a named GitHub Environment so only this job — not an
+    # arbitrary workflow in the repo — can publish to PyPI; the PyPI trusted
+    # publisher is configured with this exact environment name. No protection
+    # rules: the release stays fully automated and the human gate remains the
+    # Release PR merge (ADR-0007 / ADR-0016).
+    environment:
+      name: pypi
+      url: https://pypi.org/p/gda
 
     env:
       RELEASE_TAG: ${{ needs.cut-release.outputs.tag_name }}
@@ -117,6 +129,23 @@ jobs:
 
       - name: Build distributions
         run: uv build
+
+      # Publish to PyPI BEFORE the GitHub Release is un-drafted. Un-drafting is
+      # what creates the git tag, and ADR-0007's recovery model treats the tag as
+      # the atomic "release succeeded" signal: everything load-bearing runs before
+      # it, so a failure here leaves a tag-less draft that the release-PR gate
+      # flags and "Re-run failed jobs" converges (ADR-0016). The publish is
+      # idempotent via skip-existing — PyPI files are immutable, so a re-run from
+      # the same tagged commit skips whatever already landed and uploads the rest,
+      # mirroring the --clobber idempotency of the gh release upload below.
+      # Trusted Publishing (OIDC) is used; configuring the PyPI pending publisher
+      # is the one-time human step tracked on #207. Until that is done this step
+      # fails the release — intended: it is the gate that proves the publisher is
+      # configured before any tag is minted.
+      - name: Publish to PyPI (Trusted Publishing)
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          skip-existing: true
 
       - name: Publish GitHub Release
         env:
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Review Skill Pass: Standards / Spec

Fixed point: `main`  
Diff reviewed: `git diff main...HEAD`  
Commit reviewed: `2c40524 ci(release): publish gda to PyPI via Trusted Publishing`

## Standards

Hard violations: none found.

Judgement calls: none found.

Per changed file:

- `.github/workflows/release.yml`: no documented-standard violation. The PyPI publish step stays inside the existing release workflow/job, after release-commit checkout, tests, tag validation, and `uv build`, and before GitHub release publication. That preserves ADR-0007's same-workflow, tested-before-publish, exact-release-commit, and rerun-recovery invariants, and ADR-0008's Release-PR-only / single-version-authority rule. OIDC/no stored token also aligns with ADR-0007's no-long-lived-credential rule.

- `docs/adr/0016-publish-to-pypi-via-trusted-publishing.md`: no documented-standard violation. The ADR uses project vocabulary correctly (`gda`, `gda-mcp`, `gda[mcp]`) and records the PyPI decision without silently contradicting existing ADRs. It also keeps ADR-0013's one distribution / `[mcp]` extra / eventual PyPI command direction intact.

## Spec

No findings for this workflow + ADR slice.

I did not count the remaining #207 acceptance criteria as misses because the PR body explicitly scopes them out: "Does not close #207 - the PyPI-side human setup, first publish, and doc reconciliation remain." That covers the PyPI name claim, post-release install validation, and ADR-0013/#195 recipe reconciliation.

Scope creep: none found. The added GitHub Environment/OIDC setup is directly in service of the spec line: "Trusted Publishing (OIDC, no long-lived token)."

Implemented-but-wrong requirements: none found. The diff appears to implement the scoped requirement to extend the existing release-please publish flow so it also publishes the built distributions to PyPI via Trusted Publishing, with publish after `uv build`, no stored PyPI token, and `skip-existing: true`.

Summary: Standards findings: 0. Spec findings: 0. Worst issue: none.

qinhaihong-red:
## Code Review and Quality Pass

Verdict: request changes.

### Findings

1. Required: Keep PyPI OIDC publishing privilege out of the build/test steps.

   `.github/workflows/release.yml` now grants `id-token: write` and attaches the `pypi` environment to the whole `github-release` job before checkout, setup, unit tests, Godot e2e, tag validation, and `uv build` run (`.github/workflows/release.yml:68-148`). In GitHub Actions, that OIDC permission is job-scoped, so any earlier shell/code path in the test or build environment can request an OIDC token and exchange it for a PyPI upload credential before the intended `pypa/gh-action-pypi-publish` step. This is exactly the supply-chain privilege boundary the PyPA action warns about: only set `id-token: write` in the publishing job, and try to separate building from publishing so injected build/test scripts cannot elevate privileges ([pypa/gh-action-pypi-publish README](https://github.com/pypa/gh-action-pypi-publish#trusted-publishing)).

   The ADR rejects a separate publish job because it says artifact hand-off would run after `github-release`, after the tag exists (`docs/adr/0016-publish-to-pypi-via-trusted-publishing.md:86-92`). That is not inherent. The workflow can preserve ADR-0007's tag-as-commit-point model with a job graph like: verify/build artifacts without OIDC -> PyPI publish job with `environment: pypi` and `id-token: write` -> GitHub Release upload/un-draft job. If PyPI publish fails, the final un-draft still never runs, so the release remains a tag-less draft, but the build/test job never receives PyPI publishing authority.

2. Required: Correct the pending-publisher name-claim statement in ADR-0016.

   ADR-0016 says the pending publisher "reserves the name" and describes the project name as "claimed via a pending publisher" (`docs/adr/0016-publish-to-pypi-via-trusted-publishing.md:57-62`). PyPI's own Trusted Publishing docs say the opposite: a pending publisher does not create a project or reserve the project name until it is actually used to publish, and if another user registers the name first, the pending publisher is invalidated ([PyPI docs](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/)).

   This matters because #207 has an explicit acceptance criterion for claiming the `gda` project name or deciding an alternative. The PR body correctly scopes that human setup out, but the accepted ADR currently records a stronger and false guarantee. Please reword this section to say that `gda` is currently unregistered, pending-publisher setup does not reserve it, and the name is only claimed on first successful publish; if it is taken before then, ADR-0013's canonical commands need the documented alternate distribution name.

### Other Axes

Correctness, readability, architecture, security, and performance were otherwise clean for this scoped workflow/ADR slice. The `skip-existing: true` input exists on `pypa/gh-action-pypi-publish@release/v1`, the default `dist` package directory matches `uv build`, `attestations` default to enabled for Trusted Publishing, and PR checks show the Python unit/package-build job passing. I did not rerun the full test suite locally because this change is workflow/docs-only and the relevant PR check already passed; Godot e2e was skipped by CI for this PR.

qinhaihong-red:
Both findings addressed in `aa517d5`.

**1. OIDC privilege isolation (Required) — fixed.** Agreed; the job-scoped `id-token: write` was a real privilege bleed. The single `github-release` job is split into three, exactly along the lines you proposed:

| job | permissions | env | runs |
|---|---|---|---|
| `build-release` | `contents: read` | — | full suite + `uv build`, uploads `dist/` as a run-scoped artifact |
| `publish-pypi` | `id-token: write` **only** | `pypi` | downloads artifact, publishes — no project/test/build code |
| `publish-github-release` | `contents: write` | — | downloads artifact, uploads to the GitHub Release, un-drafts (creates the tag) |

The tag-as-commit-point recovery model (ADR-0007) is preserved by **job ordering**, not co-location: `publish-pypi` runs before the un-draft, so a PyPI failure skips `publish-github-release` → tag-less draft → "Re-run failed jobs" converges (build-release's artifact persists for the run, so the re-run re-publishes without rebuilding). The build/test environment never holds publishing authority.

**2. Pending-publisher name-claim wording (Required) — fixed.** You're right, and I confirmed against the packaging docs: a pending publisher does **not** reserve the name; PyPI creates the project (claims the name) only on the **first successful publish**, and the pending publisher is invalidated if another account registers `gda` first. ADR-0016's "Project name" bullet is reworded to say the name is claimed by *landing the first release*, not by configuring the publisher, with the alternative-name fallback if it's taken before then.

ADR-0016's Decision + Considered options are updated to match (the separate isolated job is now the *chosen* option; the single-job-with-OIDC variant is the rejected one, citing the privilege bleed).

qinhaihong-red:
## Follow-up Recheck

I rechecked the latest head, including `aa517d5`, against the two earlier review findings.

The original two findings are addressed:

- **OIDC privilege isolation:** fixed. `build-release` now runs checkout/tests/`uv build` with only `contents: read`, while `publish-pypi` is the only job with `id-token: write` and the `pypi` environment. That keeps PyPI publishing authority out of the build/test environment and matches the PyPA Trusted Publishing guidance.
- **Pending publisher name claim wording:** fixed. ADR-0016 now correctly says a pending publisher does not reserve `gda`; the project name is claimed only by the first successful publish, and an alternate distribution name is needed if `gda` is registered by someone else first.

One new release-flow issue remains:

Required: make the recovery path preserve the documented "Re-run failed jobs" convergence.

After the split, a PyPI failure makes `publish-pypi` fail and leaves `publish-github-release` skipped because it depends on `publish-pypi`. GitHub's rerun behavior for failed jobs reruns failed jobs and their dependencies; it does not clearly guarantee rerunning downstream jobs that were skipped because a dependency failed. That means the ADR-0007/ADR-0016 recovery story may no longer converge with **Re-run failed jobs**: the rerun can repair PyPI publish but still never run the downstream un-draft/tag-creation job unless the maintainer reruns the whole workflow or manually reruns the skipped job.

Suggested fix: keep the three-job privilege split, but make `publish-github-release` run after `publish-pypi` completes and fail explicitly when PyPI did not succeed. For example, give it an `if` that allows it to run after a non-cancelled release attempt, then add an early shell guard like:

```yaml
if: ${{ !cancelled() && needs.cut-release.outputs.release_created == 'true' }}

steps:
  - name: Require PyPI publish success
    if: ${{ needs.publish-pypi.result != 'success' }}
    run: exit 1
```

Then the un-draft job itself becomes failed instead of skipped when PyPI fails, so **Re-run failed jobs** includes it and can continue through to tag creation once `publish-pypi` succeeds. This keeps the security fix intact while preserving the release recovery contract.

qinhaihong-red:
Addressed in `b72374e`. Agreed — and I confirmed the GitHub docs on "Re-run failed jobs" do **not** specify whether a job skipped because an upstream failed is re-run, so the recovery contract shouldn't depend on it.

Fix applied, slightly broader than the recheck named: I made **both** downstream jobs run-on-failure + explicit-fail, not just `publish-github-release`. The same skip-vs-rerun ambiguity also affects `build-release → publish-pypi` (a build/e2e failure is itself an ADR-0007 recovery point the original single job converged from), so guarding only the PyPI→GitHub hop would still strand recovery after a build failure.

Each post-cut downstream job now:
- runs on `if: ${{ !cancelled() && needs.cut-release.outputs.release_created == 'true' }}` (so an upstream failure makes it *run*, not skip), and
- has a first guard step that `exit 1`s when its upstream `result != 'success'`.

```
build-release ──fail──▶ publish-pypi: "Require the build to have succeeded" → exit 1  (FAILED, not skipped)
publish-pypi  ──fail──▶ publish-github-release: "Require the PyPI publish to have succeeded" → exit 1  (FAILED; un-draft never runs → no tag)
```

Result: every post-cut job is **success-or-failed, never skipped-because-upstream-failed**, so `Re-run failed jobs` re-runs the whole failed tail and drives through to tag creation — matching the pre-split single-job behavior. The `publish-github-release` guard also doubles as a correctness check (no tag is minted unless PyPI published).

Traced all paths:

| scenario | publish-pypi | publish-github-release | tag? | Re-run failed jobs |
|---|---|---|---|---|
| no release cut | skipped | skipped | — | n/a (release-pr maintains) |
| happy path | success | success | created | n/a |
| PyPI publish fails | **failed** | **failed** (guard) | none | re-runs both → converges |
| build/e2e fails | **failed** (guard) | **failed** (guard) | none | re-runs all three → converges |
| run cancelled | skipped | skipped | — | n/a |

ADR-0016's ordering/recovery bullet now documents the run-on-failure guards and why the contract is kept independent of the undocumented skip-rerun behavior.

PR #211, opened 2026-06-20
chore(release): force 0.1.24 to verify the PyPI publish path

qinhaihong-red:
## Why

Actively verify the new PyPI publish chain from #210 end-to-end, and claim the `gda` name on PyPI via the configured pending publisher. The first publish only fires on a versioned release, and there is no pending `feat`/`fix` to bump the version naturally (only `ci`/`docs` landed since 0.1.23).

## What

Set `release-as: "0.1.24"` in `release-please-config.json` — ADR-0007/0008's sanctioned force-version path. (`Release-As:` commit footers don't survive this repo's squash policy, which blanks the squash body, so the config field is the reliable trigger.)

## Sequence

1. **Merge this PR** → release-please opens a Release PR for **0.1.24**.
2. **Merge the 0.1.24 Release PR** → the release run executes `build-release → publish-pypi → publish-github-release`; `publish-pypi` performs the **first PyPI publish** (creates the `gda` project, claims the name).
3. Post-publish follow-up PR: **remove `release-as`** (otherwise it pins every future release to 0.1.24) **+ reconcile docs** from the git-source form to the canonical `pip install "gda[mcp]"` / `uvx --from "gda[mcp]"` commands — closing out #207.

## Watch points on the release run

- `publish-pypi` goes green and `https://pypi.org/project/gda/` appears.
- If `publish-pypi` fails, `publish-github-release` fails at its guard (no tag minted) → recover with **Re-run failed jobs** once the cause is fixed (per ADR-0016).

Refs: #207

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -20,6 +20,8 @@
     { "type": "ci", "section": "Continuous Integration", "hidden": true }
   ],
   "packages": {
-    ".": {}
+    ".": {
+      "release-as": "0.1.24"
+    }
   }
 }


qinhaihong-red:
## Review skill pass

## Standards

No hard violations found.

The only diff adds `packages["."].release-as = "0.1.24"` to `release-please-config.json`. ADR-0008 keeps release-please as the single version authority and explicitly routes forced versions through release-please's one-off `release-as` mechanism, rather than hand-editing version files or pushing tags. ADR-0007's superseded escape-hatch note says the same. So this is the sanctioned path for forcing a release version, not a bypass of the release model.

Judgment call: the cleanup risk is real, but it is not a current standards violation. ADR-0008's "one-off config entry" boundary means this setting must be removed after `0.1.24` is published; leaving it in place would turn a one-time override into stale release policy. ADR-0016 also supports the PR's sequencing: docs should only move from the git-source install form back to canonical PyPI commands after a real PyPI release exists.

## Spec

No blocking spec defect found.

The PR matches the stated intent: force release-please to open a `0.1.24` Release PR so the new PyPI Trusted Publishing chain from #210 can run and perform the first PyPI publish for `gda`. Current state supports the target: latest release is `v0.1.23`, `pyproject.toml` and `.release-please-manifest.json` remain at `0.1.23` on this PR head, and `https://pypi.org/pypi/gda/json` still returns 404.

There is no scope creep in this diff. It does not edit `pyproject.toml`, `.release-please-manifest.json`, `uv.lock`, or tags directly; release-please should produce those changes in the generated Release PR.

The remaining #207 end state is intentionally downstream of this PR: merge this PR, merge the generated `0.1.24` Release PR, verify the first PyPI publish, then remove `release-as` and reconcile ADR-0013 / registration docs from the git-source form back to `pip install "gda[mcp]"` and `uvx --from "gda[mcp]" gda-mcp`.

Summary: Standards 0 hard findings; Spec 0 blocking findings. Worst issue is process risk only: `release-as` must not survive past the one-off `0.1.24` release.

qinhaihong-red:

No blocking findings.

The forced root-package `"release-as": "0.1.24"` is reasonable for this PR's stated purpose. It keeps release-please as the version authority, avoids manual tag/version-file edits, and should cause the next Release PR to carry the `0.1.24` version that exercises the new PyPI Trusted Publishing path.

Findings: none.

FYI: `release-as` bypasses normal conventional-commit version calculation while it remains in config. That is the intended short-term effect here, but it must be removed after the `0.1.24` Release PR is merged and the first PyPI publish is verified. Leaving it in place would turn a one-off release override into stale release policy and could make future releases ignore normal commit-driven versioning.

FYI: `pyproject.toml` and `.release-please-manifest.json` staying at `0.1.23` on this PR head is expected. release-please should update those in the generated Release PR, not in this forcing PR.

Consider: issue #207 is already closed while first PyPI publish and docs reconciliation are still pending. Keep an explicit follow-up tracker until `gda` exists on PyPI, `pip install "gda[mcp]"` and `uvx --from "gda[mcp]" gda-mcp` are verified, `release-as` is removed, and ADR-0013 / registration recipes are reconciled back to canonical package commands.

Security, performance, and readability impact is negligible: this is a small config-only change with no runtime surface. Verification looks appropriate for the scope: package build passed, and the remaining meaningful verification happens only after release-please opens the `0.1.24` Release PR and the publish workflow runs.

qinhaihong-red:
Thanks — both passes land as no-blocking-findings, and the diff stays as-is (config-only force via the sanctioned `release-as` path).

Handling the non-diff points:

- **`release-as` must not survive past 0.1.24** (both FYIs): agreed and already in the plan. The post-publish follow-up PR removes `release-as` **and** reconciles the docs in one change, merged right after the first publish is verified so the override never becomes stale policy. Removing it simply returns release-please to normal commit-driven computation (next `feat`/`fix` → 0.1.25; `ci`/`docs`/`chore` → no release) — no auto-bump, manifest stays authoritative.
- **`pyproject.toml` / `.release-please-manifest.json` at 0.1.23 on this head**: expected — release-please writes 0.1.24 into the generated Release PR, not into this forcing PR.
- **"#207 is already closed while work remains" (Consider):** good catch. It had been closed as COMPLETED when #210 merged, but #210 only shipped the workflow+ADR slice (`Refs:`, not a closing keyword). I've **reopened #207** as the single tracker and checked off only the ADR criterion; the name claim, first-publish verification, clean-env install check, and docs reconciliation remain open there. It closes only once all are green.

No changes pushed to this PR. Ready for merge to kick off the `0.1.24` Release PR and the first publish.

PR #232, opened 2026-06-22
feat: surface execution kind in --schema (#230)

qinhaihong-red:
## What

Add each command's execution **`kind`** (`headless` / `export` / `live`) to its `--schema` self-description, derived from the command descriptor's single source of truth (`HeadlessCommand.kind`, an `ExecutionKind`). Additive and backward-compatible.

Closes #230.

## Why

`--schema` already reports a command's input/output/error JSON Schemas but not its execution channel, so an agent — and gda-mcp, which consumes `--schema` mechanically (ADR-0012) — could not tell a live command from a headless one from the schema alone.

This is the **one** central implementation of the "`--schema` describes the live kind" acceptance line (stories 14/15/24, ADR-0004) that was otherwise scattered across the live slices (#220 / #221 / #222) and #7 — implemented once here so those slices reference it rather than each re-touching the schema-emission path (a known append-conflict hotspot).

## How

- `src/gda/models.py` — add additive `kind: str | None = None` to `CommandSchema` and `CommandManifestEntry`; `CommandSchema.of(...)` accepts and threads `kind`.
- `src/gda/headless.py` — the single `--schema` emission point (`_SchemaCommand.parse_args`, serving both the argv and `--params-json` forms) passes the backing `command.kind` (`None` for the bare `gda schema` meta command).
- `src/gda/surface.py` — the aggregate manifest builder passes `gda_command.kind`, so the aggregate and per-command forms agree.
- `src/gda/mcp/server.py` is **untouched** — gda-mcp maps only `name`/`description`/`input`/`output`, so the new field is ignored (ADR-0012); adding it is backward-compatible.

`ExecutionKind` subclasses `str`, so it serializes as the lowercase value (`"headless"`/`"export"`/`"live"`), never `"ExecutionKind.HEADLESS"`.

## Tests

- `tests/test_schema_command.py` — per-command `kind` for HEADLESS (`scene get`), EXPORT (`export run`), LIVE (`game tree`, asserted without a daemon since `--schema` is intercepted before execution), and argv ↔ `--params-json` parity.
- `tests/test_schema_aggregate.py` — every manifest entry carries a `kind` that matches that command's own `--schema`; all three kinds present.
- `tests/test_export_run_commands.py` — updated one pre-existing exact-key-set assertion to include the additive `kind`.

[...]

Issue #230, opened 2026-06-21
Feature: surface the execution `kind` in `--schema` (story 24)


qinhaihong-red:
## Parent

PRD #6 — Phase 2 — live operations via gda-daemon

## What to build

Surface each command's execution **kind** in its `--schema` self-description (PRD
story 24, ADR-0004). Every command now carries a static `ExecutionKind` (HEADLESS /
EXPORT / LIVE) on its descriptor (added by the gda-daemon bootstrap, #7); `--schema`
already reports a command's input / output / error JSON Schemas but **not** its
kind, so an agent — and gda-mcp, which consumes `--schema` mechanically (ADR-0012) —
cannot tell a live command from a headless one from the schema alone.

Add `kind` to the `--schema` payload, populated from the command's own descriptor at
emission, **additive and backward-compatible** so gda-mcp keeps mapping
input/output unchanged and simply gains the kind. This is the **one** cross-cutting
schema-ABI change that satisfies the "`--schema` describes … the live kind"
acceptance line currently scattered across the live slices (#220 / #221 / #222) and
#7 — implemented once here; those slices reference it rather than each re-doing it.

## Acceptance criteria

- [ ] `gda <group> <command> --schema` includes the command's execution `kind` (HEADLESS / EXPORT / LIVE) alongside input / output / error
- [ ] The kind is derived from the command's own descriptor (single source of truth), not restated per command
- [ ] Verified for a HEADLESS (e.g. `scene get`), an EXPORT (`export run`), and a LIVE (`game tree`) command
- [ ] Additive / backward-compatible: gda-mcp still maps input → inputSchema / output → outputSchema unchanged (ADR-0012); the `--schema` ABI gains a field and breaks nothing
- [ ] The kind appears consistently across the argv and `--params-json` self-description paths

## Blocked by

- None — can start immediately (builds on #7's merged `ExecutionKind`; origin: #7 / PR #229)

qinhaihong-red:
[...]

PR #234, opened 2026-06-22
feat: atomic headless writes with file_changed_externally guard (#226)

### LINKED ISSUE #226: Feature: concurrent-editor safety — atomic headless writes with file_changed_externally
## Parent

## What to build

Concurrent-editor safety on the **headless** side (independent of the daemon, ADR-0018):
make file-mutating operations atomic and guard them with an optimistic mtime check. If
the target `.tscn` / `.gd` / `.tres` changed on disk since `gda` last read it, the
write fails with `file_changed_externally` rather than clobbering the external edit.
Document the single-driver stance — a concurrent external editor's writes are out of
scope; the editor's own "file changed on disk, reload?" prompt is the reciprocal guard.

Also registers the `file_changed_externally` code in the `src/gda/error_codes.py`
registry and the ADR-0002 table (per ADR-0002 — this is where a candidate code becomes
real ABI).

## Acceptance criteria

- [ ] Headless mutating ops detect external modification since read and fail with `file_changed_externally`
- [ ] `file_changed_externally` is registered in `error_codes.py` + the ADR-0002 table
- [ ] Writes are atomic — a failure leaves no partial/torn file
- [ ] Same `--json` / `GdaError` contract; `--schema` unaffected
- [ ] e2e: external edit between gda's read and write → typed failure, file left unclobbered

## Blocked by

- None — can start immediately (does not depend on #7)

qinhaihong-red:
## What

Concurrent-editor safety for **headless** file-mutating operations (ADR-0018): writes are now **atomic**, and read-modify-write ops carry an **optimistic staleness guard** — if a target `.tscn` / `.gd` / `.tres` / `.gdshader` changes on disk in the in-process read→write window, the write is refused with a typed `file_changed_externally` error instead of clobbering the external edit.

Closes #226.

## Why

`gda` is assumed the project's sole driver (ADR-0018, extending ADR-0009), but a human may casually open the editor to view/verify output. A blind headless write could silently clobber such an edit, and a torn write could leave a half-written file. ADR-0018 calls for "two cheap safeguards": atomic writes + an optimistic mtime check.

## How (Design A — engine-side, no new CLI ABI)

Writes happen engine-side in GDScript (`gda` spawns headless Godot), so the guard and atomicity live there.

- **Atomic write.** `ResourceSaver.save()` / `FileAccess` truncate the destination in place, and Godot's atomic mode (`FileAccess::set_backup_save`) is not bound to GDScript — so we replicate it: write to a **same-directory sibling temp**, then `DirAccess.rename_absolute()` (atomic same-filesystem POSIX rename). On any failure the target is left byte-untouched and the temp removed. *Finding:* `ResourceSaver.save()` selects its saver by the destination's recognized extension, so the temp **preserves the original extension** (`.gda-<pid>-<file>.tmp.<ext>`) — a bare `.tmp` failed every `.tscn`/`.tres` save with "File unrecognized". PID in the name avoids collisions between parallel one-shot headless processes.
- **Staleness guard.** Capture a `{mtime, size}` token right after the read; recheck right before the write; a difference → `file_changed_externally`, write refused. The token is **mtime + size**, not mtime alone: `FileAccess.get_modified_time` is whole-second granularity, so size catches a same-second edit. Covers **11 read-modify-write ops** (node add/set/remove/duplicate/move, connect/disconnect-signal, script-attach via the shared load/save tail; resource-set, script-set, shader-set wired individually) — the AC's `.tscn`/`.gd`/`.tres` (+ `.gdshader`). The 7 create blind-writes keep their `already_exists` check and still get atomicity.
- **Error code.** `file_changed_externally` registered as `category=operation`, `source=operation`, exit `4` — it is emitted engine-side from `operations.gd` via `_fail(...)`, so it must be operation-source (the registry mirror test enforces GDScript-const ↔ operation-source parity). ADR-0002 table row added; ADR-0018's line-84 candidate annotation corrected from "classifier-source" to "operation-source" (minimal correction + a realization note; decision prose untouched).
- **Deterministic e2e seam.** The in-process window is sub-second, so a test can't naturally race an edit into it (and pre-setting mtime fails, since capture-at-load captures the pre-set value). A narrow, **production-inert** env seam `GDA_TEST_PERTURB_BEFORE_SAVE` perturbs the target's size between capture and recheck. It is gated solely by `OS.has_environment(...)`; `runner.py` spawns Godot with no `env=` and never sets it, so it is dead code in production.

## Tests

TDD across three slices: (1) register the code [fast tier], (2) route writes through the atomic helpers [e2e regression], (3) the staleness guard [headline e2e].

New e2e (real Godot): `file_changed_externally` for node-add / script-set / resource-set (seam armed → typed error + the mutation did not land + no `.gda-*` temp orphan), plus an atomicity check that a successful `node add` leaves no temp orphan.

**Verification (independently re-run by the reviewer):**
- `uv run --frozen pytest -m "not e2e"` → **647 passed**
- Full real-Godot e2e (`GDA_GODOT=…`, serial, `uv run`): `pytest tests/test_e2e_node.py tests/test_e2e_script.py tests/test_e2e_resource.py tests/test_e2e_asset_file.py` → **165 passed in 377s**

> Note: CI's "Godot e2e tests" job is gated (manual/schedule `workflow_dispatch`) and **skips** on a normal PR, so the real-Godot e2e above — run locally — is this PR's e2e verification. The unit job (`-m "not e2e"`) runs in CI.

## Coordination

Write-disjoint from #232 (#230): this PR touches `operations.gd` / `error_codes.py` / ADR-0002 / ADR-0018 / `test_e2e_*`; #232 touches `models.py` / `surface.py` / `headless.py` / ADR-0004 / ADR-0012 / `test_schema_*`. The two merge cleanly in either order.


qinhaihong-red:
## Review: Standards / Spec

### Standards

No documented-standards findings.

The diff is consistent with the relevant standards checked: ADR-0002's registered operation-code and sentinel-output contract, ADR-0010's default `operations.gd` headless op-dispatch path, ADR-0018's atomic headless write plus optimistic staleness guard, and ADR-0006's path-normalization boundary.

Residual risk: I did not run tests in this review pass. The new mtime+size guard remains optimistic, so same-second same-size edits or a check-to-rename race may still be missed, but ADR-0018 frames this as an optimistic safeguard rather than a full concurrency protocol.

### Spec

- **Required:** Scene mutation guard captures too late. Issue #226 says: "If the target `.tscn` / `.gd` / `.tres` changed on disk since `gda` last read it, the write fails with `file_changed_externally`," and the design comment includes the shared `.tscn` mutation path in the guard scope. In `src/gda/ops/operations.gd`, `_load_scene()` reads the `.tscn` via `ResourceLoader.load()` around line 3104, but `_load_for_mutation()` does not capture the staleness token until after `packed.instantiate()`, dependency checks, and `_capture_external_scripts()` around line 3152. If an external editor changes the scene after `ResourceLoader.load()` but before the token capture, `_check_unchanged()` around line 3195 compares against the already-modified file and allows the stale in-memory scene to overwrite it. That leaves a real unguarded read-to-write window for node add/set/remove/duplicate/move, signal connect/disconnect, and script attach. Capture the token immediately after the `.tscn` read, before instantiation and any other work, so the recheck covers the full window promised by the spec.

No scope creep or registry/schema-contract mismatch found.

Summary: Standards findings: 0. Spec findings: 1. Worst issue: the shared `.tscn` mutation path can still clobber an external edit made after the scene read but before token capture.

qinhaihong-red:
## Code Review and Quality

Verdict: **Request changes**

### Findings

- **Required (Correctness):** The shared `.tscn` mutation path still has an unguarded read-to-write window. `_load_scene()` reads the scene from disk with `ResourceLoader.load()` around `src/gda/ops/operations.gd:3104`, but `_load_for_mutation()` waits until around `src/gda/ops/operations.gd:3152` to capture the staleness token, after `packed.instantiate()`, the unmaterialized-node guard, and `_capture_external_scripts()`. Any external edit that lands after the scene load but before that capture becomes the new token baseline, so `_check_unchanged()` around `src/gda/ops/operations.gd:3195` will pass and the stale in-memory scene can overwrite the external edit. That breaks the main correctness guarantee for node add/set/remove/duplicate/move, signal connect/disconnect, and script attach. Capture the token immediately after the `.tscn` load succeeds, before instantiation or other work, and add a regression test that exercises this earlier window; the current seam only perturbs immediately before `_check_unchanged()`, so it cannot catch this gap.

### Other Axes

- Readability / architecture: The helper split is understandable and follows the existing `operations.gd` style.
- Security: No new secret/auth boundary found. The target project remains trusted per ADR-0009.
- Performance: The added size/mtime checks and same-directory temp writes are bounded per operation.
- Tests: The new e2e coverage exercises the recheck path, but misses the load-to-token-capture gap above.

### Verification

- CI status observed on the PR: `Python unit tests and package build` passed; `Godot e2e tests` was skipped by workflow gating.
- I ran: `UV_CACHE_DIR=/tmp/gda-pr234.G8t5AO/uv-cache uv run --frozen pytest tests/test_error_registry.py -q` -> `7 passed`.
- I did not rerun the full real-Godot e2e suite.

qinhaihong-red:
Good catch — fixed in `fee7558`.

**[Required · correctness] Scene mutation captured the staleness token too late.** Confirmed: `_load_for_mutation` captured the `{mtime,size}` baseline only after `packed.instantiate()` (which runs the project's script `_init`, ADR-0009, and can take real time), the unmaterialized-node guard, and `_capture_external_scripts()`. An external edit landing in that read→instantiate window became the baseline, so `_check_unchanged()` passed and the stale in-memory scene could clobber it.

Fix: moved `_capture_staleness_token()` to **immediately after `_load_scene()` succeeds, before `instantiate()`** or any other work — the baseline now reflects the file gda actually read. (The other read-modify-write ops — `resource set`, `script set`, `shader set` — already captured right after their own read and were not affected; verified.)

**Regression test for the earlier window:** added a second production-inert seam `GDA_TEST_PERTURB_AFTER_LOAD` that perturbs the target *after the read but before `instantiate()`*, and a test asserting the recheck still fires `file_changed_externally`. The pre-existing `GDA_TEST_PERTURB_BEFORE_SAVE` only covered the write-time window; this one fails on the old late-capture placement and passes on the fix.

**Residuals (intentionally not changed):** the mtime+size token can still miss a same-second *same-size* edit, and a check→rename TOCTOU window remains. Per ADR-0018 these are accepted as an *optimistic* safeguard, not a full concurrency protocol — matching the Standards pass's own framing.

Verification (real Godot, `uv run` in the worktree; CI's e2e job is gated/skips):
- `pytest tests/test_e2e_node.py tests/test_e2e_script.py tests/test_e2e_resource.py` → **154 passed**
- both window regression tests pass by name (`..._in_window...`, `..._during_instantiate...`)

Branch rebased onto current `main` (post-#232 / 0.1.28); still write-disjoint, so no conflicts.

Issue #224, opened 2026-06-20
Feature: runtime diagnostics — errors and output log of the running game


qinhaihong-red:
## Parent

PRD #6 — Phase 2 — live operations via gda-daemon

## What to build

Read the running game's runtime diagnostics via the LIVE channel: its runtime errors
and its output log. The introspection-only counterpart to `perf` — what the running
game reports, not its frame metrics.

## Acceptance criteria

- [ ] Read the running game's runtime errors as structured output
- [ ] Read the running game's output log
- [ ] Lives under its domain command group, `kind = LIVE` (ADR-0019)
- [ ] Same `--json` / `GdaError` contract; `--schema` describes input/output and live kind (stories 14, 15, 24)
- [ ] No daemon / no session → typed errors
- [ ] Real-engine e2e: produce a known runtime error / log line and read it back (e2e serial; not a fresh empty HOME)

## Blocked by

- #7 (gda-daemon bootstrap)

qinhaihong-red:
**Design resolved (grill).** The capture mechanism is settled, and it makes `diag` a **daemon-served** live op — independent of the #221/#223 harness work.

**Mechanism — daemon-owned per-session log file.** GDScript can't hook engine/script errors or `print()` in memory (`OS.add_logger` is C++-only), so capture is via the engine log file. Instead of the shared `user://logs/godot.log` (the #180 contention source), the daemon launches each session with Godot's **`--log-file <session-scoped path>`** (under the daemon runtime dir). Verified in the engine: `--log-file` forces file logging on even when the project disables it, writes **both output and errors** to the exact path, and truncates per launch (`RotatedFileLogger`, `max_files=1`). So one file captures errors + output; the isolated path sidesteps #180 by *isolation* (not by disabling logging); truncate-per-launch makes it session-bound (ADR-0020).

**Served daemon-side, crash-resilient.** The daemon (Python) reads/parses that file to serve `diag` — no harness round-trip. It serves **even after the session process has died**, so a crash is diagnosable (the file persists); it requires only that a session was launched this daemon lifetime (else `engine_session_not_running`). Caveats: unflushed lines at the crash boundary may be lost (best-effort); a relaunch truncates the file, so read before relaunching.

**Surface.**
- Group `diag` (gda-framed, like `daemon`; ADR-0019 — not under `game`).
[...]

Issue #221, opened 2026-06-20
Feature: `input` runtime input simulation into the running game


qinhaihong-red:
## Parent

PRD #6 — Phase 2 — live operations via gda-daemon

## What to build

Runtime input simulation into the running game via the LIVE channel to the gda
harness: key, mouse (click / move), input action, and multi-event sequence injection.
A sequence is collected harness-side and applied across frames, returned as one result
(one-shot RPC, ADR-0017).

## Acceptance criteria

- [ ] `gda input` injects key / mouse / action events into the running engine session
- [ ] Sequence injection supports multiple events with frame delays in a single blocking call
- [ ] Lives under an `input` command group, `kind = LIVE` (ADR-0019)
- [ ] Same `--json` / `GdaError` contract; `--schema` describes input/output and live kind (stories 14, 15, 24)
- [ ] No daemon / no session → typed errors
- [ ] Real-engine e2e: inject input, then observe the effect (e.g. via `game get`) (e2e serial; not a fresh empty HOME)

## Blocked by

- #7 (gda-daemon bootstrap)

qinhaihong-red:
The "`--schema` describes … the live kind (story 24)" part of this slice's acceptance is a one-time cross-cutting change now tracked in #230 — reference it rather than re-implementing kind-in-schema here.

qinhaihong-red:
**`--schema` `kind` is centrally delivered by #230 — do not re-implement it in this slice.**

The acceptance line *"`--schema` describes input/output and the live kind (stories 14/15/24)"* is satisfied **once, centrally** by #230 (surface the execution `kind` in `--schema`), which adds `kind` to the `--schema` payload derived from each command's `ExecutionKind` descriptor (single source of truth).

So this slice should:
- **rely on #230** for the `kind` field — it appears automatically once the command declares `kind=ExecutionKind.LIVE` on its descriptor;
- **NOT touch the schema-emission path** (`models.py` `CommandSchema`/`CommandManifestEntry`, `headless.py` `_SchemaCommand`, `surface.py`) — that path is a **merge hotspot** shared by every live slice, and re-editing it per slice is exactly the append-conflict we want to avoid.

Net: this slice's schema/kind work reduces to *declare `kind=ExecutionKind.LIVE`* and *assert the emitted schema reports `"kind": "live"`*.

References #230.

qinhaihong-red:
[...]

Issue #220, opened 2026-06-20
Feature: `game` runtime node properties — get/set on the running game


qinhaihong-red:
## Parent

PRD #6 — Phase 2 — live operations via gda-daemon

## What to build

Add the running game's runtime node-property reads and writes: `gda game get` and
`gda game set`, operating on a node in the live runtime tree of the engine session
held by `gda-daemon` (post-`_ready`, dynamic state). These route through the LIVE
execution channel (`kind = LIVE`) to the daemon → gda harness and back, returning the
same public contract a headless command does. The on-disk counterparts stay under
`node` (ADR-0019).

## Acceptance criteria

- [ ] `gda game get <node> [--property …]` returns runtime property values from the running engine session
- [ ] `gda game set <node> <property> <value>` mutates a runtime property, applied at a frame boundary (ADR-0020)
- [ ] Lives under the `game` command group, sibling to headless `node` commands, `kind = LIVE` (ADR-0019)
- [ ] Same `--json` success / `GdaError` envelope as headless; `--schema` describes input/output and the live kind (stories 14, 15, 24)
- [ ] No daemon → typed `daemon_not_running`; no running session → typed error
- [ ] Real-engine e2e: `daemon start` → engine session → `game set` then `game get` asserts the change (run e2e serially; not a fresh empty HOME)

## Blocked by

- #7 (gda-daemon bootstrap)

qinhaihong-red:
The "`--schema` describes … the live kind (story 24)" part of this slice's acceptance is a one-time cross-cutting change now tracked in #230 — reference it rather than re-implementing kind-in-schema here.

qinhaihong-red:
**`--schema` `kind` is centrally delivered by #230 — do not re-implement it in this slice.**

The acceptance line *"`--schema` describes input/output and the live kind (stories 14/15/24)"* is satisfied **once, centrally** by #230 (surface the execution `kind` in `--schema`), which adds `kind` to the `--schema` payload derived from each command's `ExecutionKind` descriptor (single source of truth).

So this slice should:
- **rely on #230** for the `kind` field — it appears automatically once the command declares `kind=ExecutionKind.LIVE` on its descriptor;
- **NOT touch the schema-emission path** (`models.py` `CommandSchema`/`CommandManifestEntry`, `headless.py` `_SchemaCommand`, `surface.py`) — that path is a **merge hotspot** shared by every live slice, and re-editing it per slice is exactly the append-conflict we want to avoid.

[...]

PR #239, opened 2026-06-22
feat: perf — runtime performance monitoring + the time-windowed live base (#223)

### LINKED ISSUE #223: Feature: `perf` runtime performance monitoring of the running game
## Parent

## What to build

Runtime observation of the running game's performance: instantaneous performance
monitors (FPS, memory, draw calls, …) and time-windowed property/signal monitoring
over N frames. Each is collected harness-side and returned as a **single** payload,
preserving the one-shot RPC model (ADR-0017) and frame-coherent snapshots (ADR-0020).

## Acceptance criteria

- [ ] `gda perf monitors` returns the current performance monitors of the running game
- [ ] Property/signal monitoring over N frames returns one collected timeline in a single blocking call (story 23)
- [ ] Lives under a `perf` command group, `kind = LIVE` (ADR-0019)
- [ ] Samples are frame-coherent (ADR-0020); no daemon / no session → typed errors
- [ ] Real-engine e2e (e2e serial; not a fresh empty HOME)

## Blocked by

- #7 (gda-daemon bootstrap)

qinhaihong-red:
## What

`gda perf` (LIVE): `perf monitors` (instantaneous `Performance` snapshot) and `perf monitor --node <abs> --property <p> | --signal <s> --frames N` (a per-frame timeline over N frames, returned as one blocking result). This PR also lands the **shared time-windowed multi-frame base** in the harness that #221's input-sequence reuses.

Closes #223.

## Why

After `game tree/get/set` (#220) could read/poke runtime state, there was no way to read the running game's performance, nor any harness mechanism for time-windowed ops — both prescribed by ADR-0017/0020 but unimplemented.

## How

**The shared base (`gda_harness.gd`).** `_run` now returns `Variant`: a `String` (single-frame op → reply this frame, behaviour-preserving for game tree/get/set + `perf monitors`) or `null` (a multi-frame handler opened a *window* via `_begin_window(frames, sample, finalize)`). While `_window_state` is set, `_process` advances one sample per frame (single-writer order preserved), finalizing once and replying — with a hard `WINDOW_FRAME_LIMIT` guard → `live_perf_timeout` so a hung game can't stall the RPC. **A new live op adds one `match` arm + a handler that calls `_begin_window` — no `_process` change** (this is exactly how #221's sequence plugs in).

**perf.** `perf monitors` snapshots 16 `Performance` monitors (fps, process/physics time, static memory, object/node/orphan counts, draw calls/objects/primitives, video mem, physics-2D/3D + nav active counts). `perf monitor` reuses #220's `_resolve_runtime_node`/`_jsonify`/`_property_type` for property timelines, and `get_signal_list` + `connect` for signal-emission capture. Group `perf`; follows the #220 live-op pattern (models/cli/errors/`classify_live`/render).

**Error codes (LIVE/classifier/6):** `live_perf_node_not_found`, `live_perf_property_not_found`, `live_perf_signal_not_found`, `live_perf_timeout` — registered + ADR-0002 rows + a harness-const mirror test.

## Docs

`> Outcome (#223)` notes on ADR-0017 + ADR-0020 (the time-windowed multi-frame mechanism is now implemented); `command-catalog.md` marks `perf` shipped.

## Tests (independently re-run by the reviewer)

- `uv run pytest -m "not e2e"` → **698 passed**
- Full real-Godot e2e (`GDA_GODOT=…`, serial) → **257 passed, 1 skipped** — includes the new `perf` snapshot + property-timeline e2e AND the full game/headless suite confirming the `_run`→`Variant` base refactor broke nothing.

> CI's Godot e2e job is gated/skips on a normal PR, so the local e2e above is this PR's e2e verification.

## Notes for the reviewer

- `_RENDERERS` + `cli.py` `add_typer` + `error_codes.py` + ADR-0002 + `models.py` appends overlap with the parallel #224 — resolved by serial merge (this lands first; #224 rebases).
- Signal recorder is bounded to ≤4 args (`_signal_arg_count`); signals with >4 args lose the tail (acceptable for a monitor).

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -143,7 +143,6 @@ operation, and parse codes the CLI assigns).
 | `live_perf_node_not_found` | `live` | `classifier` | `6` | A live perf monitor's node path does not resolve to a node in the running scene tree (Phase 2, #223). |
 | `live_perf_property_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a property the running node does not expose for reading (Phase 2, #223). |
 | `live_perf_signal_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a signal the running node does not declare (Phase 2, #223). |
-| `live_perf_timeout` | `live` | `classifier` | `6` | A live perf time-windowed collection did not complete within the harness's frame budget (Phase 2, #223). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 
 ## Considered options
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -140,6 +140,10 @@ operation, and parse codes the CLI assigns).
 | `live_node_not_found` | `live` | `classifier` | `6` | A live game operation's node path does not resolve to a node in the running scene tree (Phase 2, #220). |
 | `live_unknown_property` | `live` | `classifier` | `6` | A live game get or set targeted a property the running node does not expose for the requested operation (no such readable storage property for get, or no such settable property for set) (Phase 2, #220). |
 | `live_uncoercible_value` | `live` | `classifier` | `6` | A live game set value cannot be coerced to the running property's declared Godot type (Phase 2, #220). |
+| `live_perf_node_not_found` | `live` | `classifier` | `6` | A live perf monitor's node path does not resolve to a node in the running scene tree (Phase 2, #223). |
+| `live_perf_property_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a property the running node does not expose for reading (Phase 2, #223). |
+| `live_perf_signal_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a signal the running node does not declare (Phase 2, #223). |
+| `live_perf_timeout` | `live` | `classifier` | `6` | A live perf time-windowed collection did not complete within the harness's frame budget (Phase 2, #223). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 
 ## Considered options
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -57,10 +57,12 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > per frame (frame-coherent, ADR-0020), and replies **once** with the whole timeline
 > when the budget is met. This keeps the **one-shot RPC** this ADR prescribes — the
 > client still issues one request and blocks for one reply — while the collection
-> itself is multi-frame. A bounded frame ceiling fails a stalled/crashed window with
-> `live_perf_timeout` (the time-windowed analogue of the boot guard), so a hung
-> engine cannot hang the RPC forever. The base is general: #222's viewport capture
-> reuses it by supplying its own sampler/finalizer, with no `_process` change.
+> itself is multi-frame. The window finalizes on its sample count and the requested
+> frame count is bounded model-side (`PerfMonitorParams`, ADR-0015), so the window
+> has no timeout of its own; a genuinely stalled engine — which never advances the
+> window at all — is caught by the daemon-level `live_timeout`, the real
+> stalled-engine guard. The base is general: #222's viewport capture reuses it by
+> supplying its own sampler/finalizer, with no `_process` change.
 
 ## Decision
 
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -47,6 +47,21 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > operations.gd OP_ERROR mirror; a separate test mirrors them against the harness
 > consts.
 
+> **Outcome (2026-06-22, #223) — the time-windowed multi-frame mechanism is now
+> IMPLEMENTED in the gda harness, realizing this ADR's one-shot RPC contract for a
+> multi-frame op.** Until now every live op replied on a single frame. `perf monitor`
+> (over N frames) needed a handler that spans frames, so the harness `_process` loop
+> gained a multi-frame base: a handler that cannot finish in one frame opens a
+> *window* (a frame budget + a per-frame sampler + a finalizer) instead of returning
+> a payload; the loop advances the window one frame per tick, accumulates one sample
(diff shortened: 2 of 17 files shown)


qinhaihong-red:
## Review pass: `review`

### Standards

Required: `perf monitor` declares an exclusive selector, but the params model does not enforce it. The CLI/model text says callers must pass exactly one of `--property` or `--signal`, and ADR-0015 requires the params model to be the input-side source of truth for both `--schema` and `--params-json`. In this PR, `PerfMonitorParams` makes both fields optional with no validator (`src/gda/models.py:2535`), the CLI forwards both values unchanged (`src/gda/cli.py:653` and `src/gda/cli.py:688`), and the harness silently prioritizes `signal` when both are present (`src/gda/harness/gda_harness.gd:359`). I also probed both argv and `--params-json`; both reached the live path instead of producing a structured usage/invalid-params error. This leaves the public schema and gda-mcp input ABI accepting ambiguous shapes the command contract says are invalid.

Consider: `perf monitors` is a noun-style subcommand while ADR-0005 generally frames commands as actions on domain objects. ADR-0019 does explicitly allow a `perf` / `monitor` area, so I would treat this as naming drift rather than a blocker.

### Spec

Required: `live_perf_timeout` is registered and documented, but the harness state machine cannot actually emit it for the current window implementation. Issue #223's design asks for a frame budget/window timeout and the `live_perf_timeout` code. `_begin_window` clamps every request to at most `MAX_WINDOW_FRAMES` (`src/gda/harness/gda_harness.gd:156`), `_advance_window` increments `elapsed`, appends one sample, and finalizes as soon as `samples.size() >= budget` (`src/gda/harness/gda_harness.gd:172` and `src/gda/harness/gda_harness.gd:189`). Because the timeout threshold is `MAX_WINDOW_FRAMES + 60`, the window completes before `elapsed > WINDOW_FRAME_LIMIT` can be reached. If the engine stops processing frames entirely, this guard also will not run; the daemon-level timeout would win instead. The current tests only inject a fake `live_perf_timeout` sentinel, so they prove classification, not the real harness timeout behavior.

Required test/spec gap: issue #223 asks for property/signal monitoring over N frames plus real-engine e2e coverage. The new e2e file covers `perf monitors`, a property timeline, missing node, and no-daemon behavior, but it does not run a real signal monitor through Godot's `get_signal_list` / `connect` / record / disconnect path (`tests/test_e2e_perf.py:76`). The signal path is currently covered only by a canned fake runner payload in `tests/test_perf_commands.py:139`, so the highest-risk Godot integration path for signal monitoring is not exercised.

Clarify: the issue design spells the command as `perf monitor --node <abs> ...`; this implementation uses a positional node argument (`src/gda/cli.py:649`). If the positional form is intentional to match existing live command style, please update the issue/PR-facing command spelling so the spec and public surface agree.

Summary: Standards found 1 required issue and 1 judgement call. Spec found 2 required issues and 1 clarification. The worst issue is the documented `live_perf_timeout` path being unreachable in the real harness.

qinhaihong-red:

Verdict: request changes.

Required: `perf monitor` accepts ambiguous selector input even though the command contract says exactly one selector is valid. `PerfMonitorParams` has `property: str | None` and `signal: str | None` with no model validator (`src/gda/models.py:2535`), and the CLI forwards both unchanged (`src/gda/cli.py:688`). The harness then chooses `signal` first (`src/gda/harness/gda_harness.gd:359`), so `gda perf monitor NODE --property position --signal hit ...` silently ignores the property selector. The same ambiguity is present through `--params-json`, which matters because gda-mcp will use the model/schema input ABI. This should be rejected model-side, with regression tests for argv and `--params-json` covering both “both set” and “neither set”.

Required: the timeout/error path for time-windowed perf collection is not a real behavior today. `_begin_window` clamps the budget to `MAX_WINDOW_FRAMES` (`src/gda/harness/gda_harness.gd:156`), then `_advance_window` adds one sample per tick and finalizes as soon as the clamped budget is reached (`src/gda/harness/gda_harness.gd:187`). That means `elapsed > WINDOW_FRAME_LIMIT` cannot be reached for normal samplers, while a truly stalled engine will not run `_advance_window` at all. The tests only inject `live_perf_timeout` through a fake runner, so they do not prove the harness can emit it. If the intended behavior is clamping, expose/reject the max consistently; if the intended behavior is a timeout, make it reachable and test the real path. Also watch the signal finalizer: it returns the original requested `frames` (`src/gda/harness/gda_harness.gd:437`) even when the window budget was clamped.

Required test gap: signal monitoring is only covered through canned Python fake-runner output (`tests/test_perf_commands.py:139`). The real Godot path that validates `get_signal_list`, `connect`, emission recording, and disconnect is not exercised in `tests/test_e2e_perf.py`. Given that signal recording is the most Godot-specific branch of this feature, add a real-engine e2e that declares/emits a signal and asserts the recorded emissions.

Verification I ran:
- `git diff --check origin/main...HEAD`
- `UV_CACHE_DIR=/tmp/gda-pr239-uv-cache uv run pytest tests/test_perf_commands.py tests/test_params_json.py::test_argv_and_params_json_dispatch_identically tests/test_schema_command.py::test_perf_commands_schema_report_kind_live_and_are_model_derived` -> 20 passed
- Probed argv and `--params-json` with both `property` and `signal`; both reached the live runner path instead of failing input validation.

I did not find a separate Critical security issue. The main performance concern is tied to the timeout/clamp semantics above rather than an additional hot-path bottleneck.

qinhaihong-red:
All 3 required findings addressed (`069c56f`, `661c6a8`, `081133b`).

**[Required] `perf monitor` ambiguous selector → rejected model-side.** `PerfMonitorParams` gains a `@model_validator(mode="after")` enforcing exactly one of `--property` / `--signal` (neither and both both error), plus `frames` bounded `ge=1, le=MAX_WINDOW_FRAMES` — following the existing `ScriptCreateParams._content_xor_extends` pattern, so the model is the single input source of truth (ADR-0015). `--params-json` with both/neither/out-of-range → structured `invalid_params` envelope (exit 4); argv → `typer.BadParameter` usage error (exit 2), matching `script create`'s dual pattern. Regression tests added for both paths (both-set, neither-set, frames over/under range).

**[Required] `live_perf_timeout` was unreachable — removed.** You're right: one-sample-per-frame finalizes at the budget before `elapsed > WINDOW_FRAME_LIMIT`, and a truly stalled engine never runs `_advance_window` (the daemon-level `live_timeout` covers that). Removed the code spec, ADR-0002 row, harness const, `WINDOW_FRAME_LIMIT`, the dead `_advance_window` branch, the silent `clampi` (the bound is now enforced model-side), and the fake-runner routing test; corrected the ADR-0017 Outcome note + catalog to name `live_timeout` as the stalled-engine guard. Registry/mirror tests stay green.

**[Required] signal monitor — real-engine e2e + finalizer fix.** The signal finalizer now reports the actual window length (`samples.size()`), consistent with the property finalizer (was reporting the requested `frames`). Added `tests/test_e2e_perf.py::test_daemon_serves_a_signal_timeline_over_a_window`: a node declares `signal ticked(n)`, emits once per frame; `perf monitor … --signal ticked --frames 5` asserts the recorded emissions — exercising the real `get_signal_list` / `connect` / record / `disconnect` path.

**[Clarify] positional `<NODE>`** is intentional — it matches the merged `game get` / `node get` style (#220); keeping it for surface consistency. **[Judgement] `perf monitors`** noun-subcommand left as-is (ADR-0019 allows the `perf` area).

Verification (independently re-run on the pushed head): `uv run pytest -m "not e2e"` → **705 passed**; full real-Godot e2e → **258 passed, 1 skipped** (incl. the new signal e2e; no game/headless regression).

PR #240, opened 2026-06-22
feat: diag — runtime diagnostics via a daemon-owned per-session log (#224)

### LINKED ISSUE #224: Feature: runtime diagnostics — errors and output log of the running game

qinhaihong-red:
## What

`gda diag errors` (structured) + `gda diag log` (raw lines), both `--limit N`: read the running game's runtime errors and output log. A **daemon-served** live op — the daemon reads a Session log it owns, not the harness.

Closes #224.

## Why

GDScript can't hook engine/script errors or `print()` in memory. The only capture is the engine log file, but the shared `user://logs/godot.log` caused the #180 contention crash (mitigated by disabling file logging in tests).

## How (resolved by design grill — see #224 / ADR-0022)

- **Daemon launches each session with Godot's `--log-file <session-scoped path>`** (under the daemon runtime dir). Verified in-engine: `--log-file` forces file logging on even when the project disables it, writes BOTH output and errors to the exact path, and truncates per launch. → captures errors+output in one file; sidesteps #180 by **isolation** (not by disabling logging); session-bound (ADR-0020).
- **Daemon serves `diag` daemon-side** by reading/parsing that file (`daemon/diag.py`) — no harness round-trip. Served **even after the session process has died**, so a crash is diagnosable (the file persists); requires only that a session launched this daemon lifetime (else `engine_session_not_running`).
- `diag errors` parses Godot's two-line `<TYPE>: <msg>` + `   at: <fn> (<file>:<line>)` best-effort → `{level, message, function?, file?, line?}`, **warnings included** (level-distinguished). `diag log` returns raw lines. New code `live_log_unavailable`; empty log → empty result.

## Docs

New **ADR-0022** (runtime diagnostics via a daemon-owned per-session log); CONTEXT **Session log** term; ADR-0002 row; `command-catalog.md` + README marked.

## Tests (independently re-run by the reviewer)

- `uv run pytest -m "not e2e"` → **709 passed** (incl. 12 log-parser unit tests + daemon-diag + diag-command + schema/render tests).
- Real-Godot e2e: the full suite ran **253 passed / 1 skipped** with the diag round-trip initially failing on a timing race; fixed by polling the Session log (the first `diag` launches the cold session, so the game must run before the marker appears — best-effort per ADR-0022). The diag e2e (`tests/test_e2e_diag.py`) now passes consistently (3/3) against a real engine: `daemon start` → `diag errors` reads back a `push_error` structured + `diag log` reads back a `print` line.

> CI's Godot e2e job is gated/skips on a normal PR; the local e2e above is this PR's e2e verification.

## Coordination

Disjoint from the parallel #223 (perf) on the daemon side (`daemon/session.py`, `server.py`, `diag.py` are #224-only); overlaps only the append hotspots (`cli.py`, `render._RENDERERS`, `error_codes.py`, ADR-0002, `models.py`). **Merge #223 first**, then this rebases onto it.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -84,6 +84,14 @@ that typed reason, not on the overloaded exit code. Both channels return the one
 `RunResult` shape.
 _Avoid_: run output, export output
 
+**Session log**:
+The per-`Engine session` capture of the running game's output and error stream,
+written by the engine to a daemon-owned path (via `--log-file`) and read by
+`gda-daemon` to serve runtime diagnostics; bound to the session (truncated each
+launch), survives the session process so a crash stays diagnosable until relaunch
+(ADR-0022).
+_Avoid_: console output, stdout dump
+
 ### Failure reporting
 
 **Gda error code**:
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ for the full picture.
   progress. Live commands are placed by domain object (a narrow `game` group for the runtime
   scene graph); the editor context is out of scope. Phase-2 live needs **Godot 4.6+** and is
   **macOS/Linux only** (it uses Unix domain sockets); Phase-1 headless is unaffected — still
-  4.4+ and cross-platform. See ADR-0017–0021.
+  4.4+ and cross-platform. See ADR-0017–0022.
 
 **Out of scope for now**
 
@@ -593,7 +593,7 @@ drives a real engine. Everything between the CLI and the runner runs as real cod
 | ----------------- | ------------------------------------------------------------------------- | ------ |
 | **Phase 1** | `gda` serving *headless operations* standalone: `info`, structured errors, `--schema`, and the domain command groups `scene`, `node`, `script`, `project` (incl. static-analysis), `resource`, `export`, `shader`, `theme`. | ✅ Surface complete |
 | **`gda-mcp`** | A thin MCP adapter generated mechanically from `--schema` — first on top of Phase 1, following `gda` forward automatically. | ✅ Shipped |
-| **Phase 2** | `gda` also serving *live operations* through `gda-daemon` and a live *engine session* (requires Godot 4.6+, macOS/Linux only; headless stays 4.4+ and cross-platform — ADR-0021). | 🚧 Runtime scene graph shipped (`gda daemon` + `gda game tree` / `get` / `set`); rest of live catalogue in progress |
+| **Phase 2** | `gda` also serving *live operations* through `gda-daemon` and a live *engine session* (requires Godot 4.6+, macOS/Linux only; headless stays 4.4+ and cross-platform — ADR-0021). | 🚧 Runtime scene graph shipped (`gda daemon` + `gda game tree` / `get` / `set`) and runtime diagnostics (`gda diag errors` / `log`); rest of live catalogue in progress |
 
 Track progress and proposals on the [issue tracker](https://github.com/aigengame/godot-agent/issues).
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -140,6 +140,7 @@ operation, and parse codes the CLI assigns).
 | `live_node_not_found` | `live` | `classifier` | `6` | A live game operation's node path does not resolve to a node in the running scene tree (Phase 2, #220). |
 | `live_unknown_property` | `live` | `classifier` | `6` | A live game get or set targeted a property the running node does not expose for the requested operation (no such readable storage property for get, or no such settable property for set) (Phase 2, #220). |
 | `live_uncoercible_value` | `live` | `classifier` | `6` | A live game set value cannot be coerced to the running property's declared Godot type (Phase 2, #220). |
+| `live_log_unavailable` | `live` | `classifier` | `6` | A live engine session was launched but its diagnostics log file is missing or unreadable, so `gda diag` cannot read the running game's errors/output (Phase 2, #224). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 
 ## Considered options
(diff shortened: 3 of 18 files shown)


qinhaihong-red:

## Standards

Hard violation:

`src/gda/daemon/server.py` has `_handle_diag()` call `_ensure_session()` when no session exists, so the first `gda diag` can create the engine session it then reads. That violates the new ADR-0022 decision that `diag` "requires only that a session was launched this daemon lifetime; with none it is `engine_session_not_running`" (`docs/adr/0022-runtime-diagnostics-via-daemon-owned-session-log.md`). The method docstring also says "diag does NOT relaunch", but the implementation immediately does the equivalent no-session launch path.

`tests/test_e2e_diag.py` codifies the same violation by saying "the FIRST diag call is what launches the (cold) session." That locks in behavior that contradicts ADR-0022's no-session rule.

Judgement calls: none.

## Spec

P1: No-session behavior is implemented as implicit session launch. Issue #224 says "No daemon / no session -> typed errors", and the design-resolution comment says `diag` requires only a session launched this daemon lifetime, otherwise `engine_session_not_running`. The implementation instead calls `_ensure_session()` when `self._session is None`, so a first `gda diag errors` / `gda diag log` can launch the game and return success instead of the required typed no-session error. That is both a partial miss of the typed-error acceptance criterion and extra behavior not asked for: `diag` should read an already-launched current Session log, not create the session it introspects.

Summary: Standards findings: 1 hard violation. Spec findings: 1 P1 finding. Worst issue: `gda diag` implicitly starts a session when the accepted design says no-session must return `engine_session_not_running`.

qinhaihong-red:
## Code Review and Quality pass

Verdict: request changes.

Findings:

- `src/gda/daemon/server.py:134` changes the no-session case into an implicit session launch. `_handle_diag()` calls `_ensure_session()` when `self._session is None`, so `gda diag errors` / `gda diag log` can start the game and then read the just-created Session log. That breaks the documented contract in this PR's own CLI docstring and ADR-0022: no daemon should be `daemon_not_running`, daemon-with-no-session should be `engine_session_not_running`, and a remembered session with a missing log should be `live_log_unavailable`. It also gives `diag` a hidden project-code execution side effect: a read-only diagnostic command starts the trusted project and runs autoloads / `_ready` code instead of only observing an already-launched session.

- `tests/test_e2e_diag.py:74` codifies the same incorrect lifecycle: the test explicitly says the first `diag` call launches the cold session, then polls until the marker appears. That means the regression suite will continue to pass if the no-session typed-error requirement remains broken. Please change the e2e setup so a non-diagnostic live op (or an explicit session-launching setup step) creates the session first, then `diag` reads it; add a unit/CLI regression for daemon-running-but-no-session returning `engine_session_not_running`.

Optional:

- Consider constraining `--limit` to non-negative values in the params model or Typer option. Today `--limit -1` is accepted and treated as "no limit" by `parse_errors()` / `parse_log()`, which is surprising for a tail limit and is not reflected in the schema.

Quality notes:

- Correctness: one required lifecycle/contract issue above.
- Readability/architecture: the parser, models, renderer, and live classifier fit the existing command patterns.
- Security: no new secret handling or untrusted-code boundary issue beyond the lifecycle side effect above.
- Performance: no blocking finding; `diag log` is intentionally a current-session log read in v1.
- Verification: GitHub shows Python unit tests/package build passing and Godot e2e skipped. I ran the focused new non-e2e tests locally (`tests/test_daemon_diag.py`, `tests/test_diag_commands.py`, `tests/test_diag_log_parser.py`) and they passed: 30 passed. I also ran the focused real-Godot e2e (`tests/test_e2e_diag.py -m e2e`) and it passed: 1 passed. These tests currently pass while preserving the lifecycle mismatch above.

qinhaihong-red:
All findings addressed (`482ab9c`, `d8c942a`, `75f06b6`).

**[Required — the serious one] `gda diag` no longer launches a session.** Confirmed: `_handle_diag` removed the `_ensure_session()` call, so a read-only diagnostic never starts the trusted project / runs its autoloads + `_ready` (ADR-0009 `Project-code execution surface`). It now reads only the session the daemon already holds — **alive OR dead** (the crash-diagnosable path stays, ADR-0022). With no session launched this daemon lifetime → typed `engine_session_not_running`, whose remediation tells the user to launch one via a live op (e.g. `gda game tree`), not via diag. Verified at runtime with a `launch_session` trip-wire (no session spawned, exit 6) and a parametrized regression test for both `diag errors` / `diag log`.

**[Required] e2e lifecycle corrected.** `tests/test_e2e_diag.py` no longer relies on diag to launch the session. It now: `daemon start` → `gda game tree` (a non-diag live op that gates on `current_scene != null`, so the scene's `_ready` has run + flushed the markers) → THEN `diag errors`/`diag log` read them. The remaining poll is a flush belt-and-suspenders only. (This also corrects the earlier premise that "the first diag launches the cold session" — it shouldn't, and now doesn't.)

**[Required] regression for daemon-running-but-no-session** added (`test_diag_with_no_session_launched_is_engine_session_not_running`, both ops, asserts `engine_session_not_running` + no launch), alongside the retained crash-diagnosable test (session dead → log still read).

**[Optional, done] `--limit` bounded `ge=1`** in `DiagErrorsParams`/`DiagLogParams` + the Typer option (`min=1`); `--limit 0/-1` → usage error (argv, exit 2) / `invalid_params` (`--params-json`); `--schema` now reflects `minimum: 1`.

Verification (independently re-run): `uv run pytest -m "not e2e"` → **714 passed**; full real-Godot e2e → **254 passed, 1 skipped**; targeted `test_daemon_diag.py` + `test_diag_commands.py` + `test_e2e_diag.py` → **24 passed** (no-session typed error, dead-session readable, limit bound, corrected lifecycle).

Issue #225, opened 2026-06-20
Feature: gda harness lifecycle — version self-sync and paired uninstall


qinhaihong-red:
## Parent

PRD #6 — Phase 2 — live operations via gda-daemon

## What to build

Complete the gda harness lifecycle beyond #7's minimal install (ADR-0018):
- **Version self-sync**: on `gda daemon start`, re-materialize the installed harness
  when its version differs from the running `gda`, and report it — so an agent never
  hits a stale-harness mismatch.
- **Paired uninstall**: `gda daemon uninstall` removes the `[autoload]` entry **and**
  the harness files together (a dangling autoload entry crashes an exported game).
- **Inertness verification**: confirm the runtime gate keeps the harness resident-inert
  (opening nothing, never `free()`/`queue_free()`-ing itself) in a human editor run, a
  plain `godot --path` run, and an exported build.

## Acceptance criteria

- [ ] `daemon start` re-materializes the harness when its version != `gda`'s, and reports the effect
- [ ] `gda daemon uninstall` atomically removes the autoload entry + harness files (paired)
- [ ] Verified: a plain run and an exported build with the harness present open no socket and do nothing (resident-inert)
- [ ] `daemon` command group; same `--json` / `GdaError` contract; `--schema`
- [ ] e2e covering install → version-bump → re-sync, and install → uninstall (e2e serial; not a fresh empty HOME)

## Blocked by

- #7 (gda-daemon bootstrap)

qinhaihong-red:
## Implementation spec (planned, ready for fan-out)

Resolved design decisions from the Phase-2 planning pass (TDD, red→green; mandatory real-engine e2e).

### D1 — Version source of truth: `HARNESS_VERSION` const (not the package version)
The harness changes far less often than `gda` ships, so coupling to the package version would churn it every patch release. The *installed* version is read from a leading GDScript comment `# gda-harness-version: <N>` that `_materialize()` prepends. Because that header is part of `_materialize`'s existing idempotent content-compare, **version self-sync falls out of the existing compare** — no separate branch; re-materialize only on mismatch (never unconditional overwrite, which would trip the concurrent-editor mtime prompt). Report via two **additive** `DaemonStartResult` fields: `harness_synced` (true only on a real version-mismatch rewrite) and `harness_version` (keep existing `installed_harness`).

### D2 — Uninstall ordering & safety
[...]

Issue #222, opened 2026-06-20
Feature: `screen` runtime viewport capture of the running game


qinhaihong-red:
## Parent

PRD #6 — Phase 2 — live operations via gda-daemon

## What to build

Capture the running game's viewport via the LIVE channel: a single screenshot and a
multi-frame capture, returned as image data. Requires a **windowed** engine session —
`--headless` uses the dummy `DisplayServer` and cannot capture pixels (ADR-0017); on a
headless server/CI this needs a virtual framebuffer (e.g. Xvfb).

## Acceptance criteria

- [ ] `gda screen` returns a viewport screenshot of the running game
- [ ] Multi-frame capture returns a collected sequence in one blocking call (time-windowed, ADR-0017)
- [ ] Lives under a `screen` command group, `kind = LIVE` (ADR-0019)
- [ ] Same `--json` / `GdaError` contract; `--schema` describes input/output and live kind (stories 14, 15, 24)
- [ ] Typed error when the session has no display (e.g. `display_unavailable`) or no daemon
- [ ] Real-engine e2e on a windowed session; CI path documented (Xvfb) (e2e serial; not a fresh empty HOME)

## Blocked by

- #7 (gda-daemon bootstrap)

qinhaihong-red:
Note from #7 (gda-daemon bootstrap, PR #229): the daemon currently launches the engine session **headless** — `gda.daemon.session.launch_session` passes `--headless` — because the tracer op `game tree` only reads the runtime `SceneTree` and needs no viewport. This slice (viewport capture) must therefore **switch the session to windowed** (or make the launch mode per-op / configurable) before pixels can be captured, since `--headless` uses the dummy `DisplayServer` as this issue already notes. The session launch is the single spot to change.

Separately: the "`--schema` describes … the live kind (story 24)" acceptance line here is now tracked once, cross-cutting, in #230 — reference it rather than re-implementing kind-in-schema in this slice.

qinhaihong-red:
**`--schema` `kind` is centrally delivered by #230 — do not re-implement it in this slice.**

The acceptance line *"`--schema` describes input/output and the live kind (stories 14/15/24)"* is satisfied **once, centrally** by #230 (surface the execution `kind` in `--schema`), which adds `kind` to the `--schema` payload derived from each command's `ExecutionKind` descriptor (single source of truth).

So this slice should:
[...]

PR #245, opened 2026-06-22
feat: surface live-stack constraints in --schema (#233)

### LINKED ISSUE #233: Feature: surface the live-stack constraint (platform + min Godot version) in --schema
## Parent

PRD #6 — Phase 2: live operations via gda-daemon. Sibling of #230 (which added the execution `kind` field): this is the second cross-cutting `--schema` self-description addition, split out per #230's scope note ("fold in … or split a sibling if it grows" — it grows).

## What to build

Surface the **live-stack constraint** — the platform and minimum Godot version a command needs to use gda's daemon/live stack (macOS/Linux only via Unix domain sockets, Godot 4.6+ per ADR-0021) — as a **structured field** in `--schema`, instead of the prose it is today (only in `--help` text and the aggregate manifest *description*).

**Scope: every command that depends on the live stack — not only `kind == LIVE`.** That set is:
- the **LIVE-channel** domain commands (`game …`), and
- the **`daemon` lifecycle group** (`daemon start` / `daemon stop` / `daemon status`), which are `kind = headless` (they run a process-management recipe, not the engine) but still require the same UNIX + Godot-4.6+ stack and today state that only in help/manifest prose.

> Keying the field purely on `kind == LIVE` would leave the daemon lifecycle commands prose-only and create a **new drift source** — exactly what surfacing the constraint is meant to remove. (Caught in PR #232's follow-up recheck.)

Source the constraint from a **single authority** that marks live-stack dependence (e.g. a per-command / per-group marker, or an explicit live-stack predicate covering LIVE-kind commands plus the daemon group) — the same single-source-of-truth discipline `kind` follows (#230), so help/manifest prose and the structured field never drift. Emit it additively in both the per-command `--schema` and the aggregate manifest, reusing the field-plumbing #230 established. Commands that do **not** depend on the live stack (plain headless domain ops like `scene get`, `export …`) carry `null` (mirroring how `kind` is null for a backing-less self-description). gda-mcp consumes it mechanically with no per-command knowledge (ADR-0012); the addition must be backward-compatible (ADR-0004).

Recommended shape (ABI is PR-reviewed; refine if a better shape emerges in implementation):

```json
"constraints": { "platforms": ["linux", "macos"], "min_godot_version": "4.6" }
```

Note a finer point for the single authority to handle: `platforms = [linux, macos]` (UDS) is uniform across the whole live-stack set, while the `min_godot_version` floor applies where a command actually launches/uses the engine (`game …`, `daemon start`); `daemon stop` / `daemon status` only talk to an already-running daemon over UDS. Model this in the single authority rather than hardcoding one uniform value per command.

This makes the precondition machine-discoverable: an agent — or gda-mcp — can tell from the schema alone that a command needs a UNIX platform (and, where relevant, Godot 4.6+) before attempting it, pairing with the existing typed `live_unsupported_platform` error rather than parsing help prose.

## Acceptance criteria

- [ ] Live-stack-dependent commands' `--schema` includes a structured `constraints` field (platform set, and a Godot-version floor where the command launches/uses the engine), sourced from a single authority — not restated per command
- [ ] The set explicitly **includes the `daemon` lifecycle group** (`daemon start`/`stop`/`status`), not only `kind == LIVE` commands
- [ ] The aggregate `gda schema` manifest carries the same field for those entries; commands with no live-stack dependence (e.g. `scene get`, `export …`) carry `null`
- [ ] Additive / backward-compatible: gda-mcp still maps input → inputSchema / output → outputSchema unchanged (ADR-0012, ADR-0004); the `--schema` ABI gains a field and breaks nothing
- [ ] The field appears consistently across the argv and `--params-json` self-description paths (as `kind` does)
- [ ] `gda schema --schema` self-description reflects the field's type/shape
- [ ] The constraint stated in `--help` / manifest descriptions and the structured field agree — single source of truth, no drift (incl. the daemon lifecycle commands)
- [ ] Verified for: a LIVE command (`game tree`), a daemon lifecycle command (`daemon start`, `kind = headless` but live-stack-dependent → carries the constraint), and a plain headless command (`scene get` → `null`)

## Blocked by

- #230 (PR #232) — reuses the `--schema` field-plumbing it adds (`CommandSchema` / `CommandManifestEntry`) and avoids concurrent edits to the same schema-emission path.

---

**Out of scope (future):** per-command *additional* constraints beyond the live-stack base — e.g. viewport capture (#222) additionally requires a windowed session / display (ADR-0017, recorded there as an environment constraint). A follow-up can add a per-command constraint-extension mechanism layered on this base field.

qinhaihong-red:
## Summary

Surfaces the **live-stack constraint** (platform set + minimum Godot version) as a structured `constraints` field in both per-command `--schema` and the aggregate `gda schema` manifest, replacing the prose that previously lived only in `--help` / manifest descriptions. Makes the precondition machine-discoverable: an agent (or gda-mcp) can tell from the schema alone that a command needs a UNIX platform — and, where relevant, Godot 4.6+ — before attempting it.

Closes #233.

## Design — single-authority predicate (no per-command marker)

A pure leaf function `live_stack_constraints(kind, operation)` in `src/gda/execution.py` is the single source both emission paths call, so the per-command schema and the aggregate manifest cannot drift:

- `platforms = ["linux","macos"]` when `kind is LIVE` **or** `operation.startswith("daemon-")`
- `min_godot_version = "4.6"` only where the engine is launched/used (`kind is LIVE` **or** `operation == "daemon-start"`) — `null` for `daemon stop`/`daemon status` (UDS-only)
- `null` otherwise

This keys on facts the command descriptors already carry, so it **touches zero `cli.py`** and auto-covers future LIVE commands (e.g. #222's `screen`) and daemon commands (#225's `daemon uninstall`) with no per-command work. `MIN_LIVE_VERSION = (4,6)` moved down into `execution.py` as the single source (`daemon_ops.py` re-exports it, backward-compatible).

## ABI

```json
"constraints": { "platforms": ["linux","macos"], "min_godot_version": "4.6" }   // LIVE + daemon start
"constraints": { "platforms": ["linux","macos"], "min_godot_version": null }     // daemon stop/status
"constraints": null                                                              // plain headless / export
```

Additive / backward-compatible (ADR-0004); gda-mcp maps input→inputSchema / output→outputSchema unchanged (ADR-0012) — no mcp files touched.

## Tests

`uv run pytest -m "not e2e"` → **788 passed** (independently re-run by the reviewer). New tests at the CLI ABI seam in `tests/test_schema_command.py` (5) and `tests/test_schema_aggregate.py` (3): `game tree` (4.6), `daemon start` (4.6 despite `kind=headless`), `daemon stop`/`status` (version null), `scene get` / `export run` (null), argv↔`--params-json` byte-parity, aggregate↔per-command agreement, and `gda schema --schema` self-describing the nullable `$ref`. No e2e in this slice.

## Follow-up at merge

ADR-0004 / ADR-0012 outcome notes (the `constraints` additive superset) + ADR-0021 cross-reference to be appended when this lands.

### DIFF
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -14,6 +14,16 @@ status: accepted
 > in the aggregate manifest (ADR-0012) every dispatchable entry's `kind` is
 > required and enum-constrained.
 
+> **Outcome (2026-06-22, #233 / PR #245):** the per-command `--schema` object
+> gained a fifth, additive key — `constraints`, the command's
+> `LiveStackConstraints` (platform set + minimum Godot version) for
+> live-stack-dependent commands, `null` otherwise. The platform/version values
+> are decided in ADR-0021 and now surfaced **structurally** rather than in
+> `--help` / manifest prose, sourced from the single
+> `gda.execution.live_stack_constraints` predicate so the structured field and
+> the surrounding prose cannot drift. Like `kind`, gda-mcp ignores it (ADR-0012),
+> so it stays a strict superset — backward compatible.
+
 ADR-0000 lists `--schema` as a core capability without defining it. We fix its
 semantics here, and deliberately scope out an overloaded interpretation.
 
--- a/docs/adr/0012-gda-mcp-tool-surface-generation.md
+++ b/docs/adr/0012-gda-mcp-tool-surface-generation.md
@@ -13,6 +13,14 @@ status: accepted
 > backward compatible. On the aggregate entry `kind` is required and
 > enum-constrained, so a consumer of `gda schema --schema` can rely on it.
 
+> **Outcome (2026-06-22, #233 / PR #245):** each manifest entry gained an
+> additive `constraints` field (the command's `LiveStackConstraints` — platform
+> set + minimum Godot version, ADR-0021; `null` for non-live-stack commands), so
+> the entry above is now `{name, description, input, output, error, kind,
+> constraints}`. gda-mcp's mapping is unchanged — it ignores `kind` /
+> `constraints` — so the addition is backward compatible. On the aggregate entry
+> the `constraints` key is always present, its value nullable.
+
 ADR-0011 fixes [gda-mcp](../../CONTEXT.md) as a subprocess adapter that consumes
 `gda`'s public ABI. ADR-0004 / ADR-0005 make each command self-describing
 (`--schema` → `{input, output, error}`) and give a deterministic CLI→MCP name map
--- a/docs/adr/0021-gda-daemon-transport-discovery-and-live-version-floor.md
+++ b/docs/adr/0021-gda-daemon-transport-discovery-and-live-version-floor.md
@@ -4,6 +4,13 @@ status: accepted
 
 # gda-daemon transport and discovery: Unix domain sockets end-to-end, per-project socket + pidfile discovery, and a Phase-2 live floor of Godot 4.6
 
+> **Outcome (2026-06-22, #233 / PR #245):** the Phase-2 live floor decided here
+> (macOS/Linux via Unix domain sockets, Godot 4.6+) is now surfaced as a
+> structured `constraints` field in per-command `--schema` and the aggregate
+> manifest (ADR-0004, ADR-0012) — machine-discoverable, no longer `--help` prose.
+> `gda.execution.MIN_LIVE_VERSION` is the single source the daemon-start version
+> gate and the schema predicate share.
+
 ADR-0017 fixed [gda-daemon](../../CONTEXT.md)'s shape — a per-project supervisor + IPC
 broker holding a transient [engine session](../../CONTEXT.md) with the
 [gda harness](../../CONTEXT.md), routed by a static `kind` channel selector — but
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -266,7 +266,7 @@ app.add_typer(theme_app, name="theme")
 # `node`. It is a domain-object group named after the running game, not a phase
 # group — the headless/live split is carried by `kind`, never by the tree.
 game_app = typer.Typer(
-    help="Act on the running game (live; macOS/Linux only, needs `gda daemon start`).",
+    help="Act on the running game (live; needs `gda daemon start`).",
     no_args_is_help=True,
 )
 app.add_typer(game_app, name="game")
@@ -279,7 +279,7 @@ app.add_typer(game_app, name="game")
 # diagnosable. From the CLI's side it routes like any live command (kind = LIVE
 # -> the daemon socket); the daemon recognizes the diag op names.
 diag_app = typer.Typer(
-    help="Read the running game's runtime diagnostics (live; macOS/Linux only, needs `gda daemon start`).",
+    help="Read the running game's runtime diagnostics (live; needs `gda daemon start`).",
     no_args_is_help=True,
 )
 app.add_typer(diag_app, name="diag")
@@ -291,7 +291,7 @@ app.add_typer(diag_app, name="diag")
 # multi-frame harness base). Like `game`, a domain-object group marked live by
 # `kind`, not by the tree (ADR-0019).
 perf_app = typer.Typer(
-    help="Monitor the running game's runtime performance (live; macOS/Linux only).",
+    help="Monitor the running game's runtime performance (live).",
     no_args_is_help=True,
 )
 app.add_typer(perf_app, name="perf")
@@ -307,7 +307,7 @@ app.add_typer(perf_app, name="perf")
 # `gda <group> <command>` → `<group>_<command>` dispatch + gda-mcp tool-name mapping
 # (ADR-0005/0011/0012).
 input_app = typer.Typer(
-    help="Inject input into the running game (live; macOS/Linux only).",
+    help="Inject input into the running game (live).",
     no_args_is_help=True,
 )
 app.add_typer(input_app, name="input")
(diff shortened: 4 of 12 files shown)


qinhaihong-red:
## `review` skill pass: Standards + Spec

Compared `origin/main...HEAD` at `51aa66f` (`feat/233-schema-constraints`). Spec source: issue #233.

## Standards

Hard violation:

- The branch adds a public `constraints` key to per-command `--schema` output and to each aggregate `gda schema` manifest entry, but the accepted decision docs still describe these public contracts without that field. ADR-0004 defines per-command `--schema` as the `{input, output, error}` contract, and ADR-0012 defines aggregate entries as `{name, description, input, output, error}`. The PR makes `constraints` a tested ABI guarantee in `CommandSchema`, `CommandManifestEntry`, and `test_self_described_manifest_describes_the_nullable_constraints_field`, so the ADRs will drift if this lands as-is. This also conflicts with the repo rule that design/decision changes keep issues, ADRs/docs, and code synchronized. Please update ADR-0004 / ADR-0012, and the ADR-0021 cross-reference if needed, in this PR or make the merge gate explicitly include those outcome notes before the ABI change lands.

Judgement call:

- Centralizing `MIN_LIVE_VERSION` and `live_stack_constraints()` in `src/gda/execution.py` is consistent with the leaf-module direction; the standards issue is the undocumented public ABI projection, not the predicate itself.

## Spec

Required:

- Issue #233 says to “Use a single authority that marks live-stack dependence ... so help/manifest prose and structured field never drift” and its acceptance criteria require “the constraint stated in `--help` / manifest descriptions and the structured field agree — single source of truth, no drift.” The implementation uses `live_stack_constraints()` as the authority for the structured field, but the prose still comes from independent `src/gda/cli.py` docstrings that feed both `--help` and aggregate manifest descriptions. Evidence from the PR branch: `gda game tree --schema` emits `{"platforms":["linux","macos"],"min_godot_version":"4.6"}`, but `gda game tree --help` and the `gda schema` description only mention macOS/Linux and a running daemon, not the Godot 4.6+ floor. The same mismatch appears for other LIVE commands such as `diag errors` and `perf monitors`. This leaves the drift source that #233 was meant to remove. Please either generate/synchronize those prose fragments from the same predicate, or remove the duplicated constraint prose and rely on the structured field.

No clear scope creep found.

Summary: Standards found 1 hard violation; Spec found 1 missing/partial requirement. Worst issue: the new structured constraint field is correct, but the public prose/ADR surfaces are still able to drift from it.

Verification run in an isolated `/private/tmp` clone: `env UV_CACHE_DIR=/private/tmp/godot-agent-pr245-review-RxnL8g/uv-cache uv run pytest -m "not e2e" -q` -> 788 passed, 266 deselected.

qinhaihong-red:
## `code-review-and-quality` pass

I found one required issue before merge.

- `src/gda/execution.py`, `src/gda/headless.py`, `src/gda/surface.py`, and `src/gda/cli.py`: `live_stack_constraints()` is the single authority for the structured schema/manifest values, but it is not the authority for the prose that still feeds `--help` and aggregate manifest descriptions. That is a correctness/architecture issue because #233 explicitly exists to eliminate the prose-only drift source. Actual output from this branch shows the problem: `game tree`, `diag errors`, and `perf monitors` all emit `constraints.min_godot_version == "4.6"`, while their descriptions omit the Godot 4.6+ floor; `daemon start` does include it. Future platform/version changes can still leave the public description and machine-readable field inconsistent. Please make the prose derive from the same constraint helper, or remove the duplicated constraint prose and rely on the structured field, with a regression test for representative help/manifest descriptions.

Consider:

- `LiveStackConstraints.min_godot_version` has a default `None`, so the self-described JSON Schema allows a non-null `constraints` object that omits the `min_godot_version` key entirely. The emitted objects currently include the key, including `null` for daemon stop/status. If the intended ABI is “key always present, value nullable,” make the field required (`str | None` without a default) and extend the self-description test accordingly. If omission is intentionally acceptable for consumers, no change needed.

No security or performance concerns found in this slice. The change size is reasonable, and the new tests cover the main ABI paths: live command, daemon lifecycle commands, non-live commands, aggregate/per-command agreement, and argv vs `--params-json` parity.

Verification run in an isolated `/private/tmp` clone: `env UV_CACHE_DIR=/private/tmp/godot-agent-pr245-review-RxnL8g/uv-cache uv run pytest -m "not e2e" -q` -> 788 passed, 266 deselected.

qinhaihong-red:
Thanks — all three addressed in `d283512`. `uv run pytest -m "not e2e"` → 789 passed.

**Standards (hard violation — ADR drift):** added dated Outcome notes to **ADR-0004** and **ADR-0012** for the additive `constraints` key (mirroring the #230 `kind` notes), plus a cross-reference Outcome note in **ADR-0021** (which decides the platform/version values now surfaced structurally). The decision docs and the shipped `--schema` ABI are back in sync, in this PR rather than at merge. (Applying the same in-PR discipline to the sibling PRs #247/#248.)

**Spec (single source / no drift):** took your second option — **removed the duplicated constraint prose and rely on the structured field**. The hand-written "macOS/Linux only" / "Godot 4.6+" claims are gone from the live-command `--help`/docstrings and group help; the structured `constraints` field (sourced from the single `live_stack_constraints` predicate) is now the sole source, so the drift you found (`game tree` etc. stating the platform but omitting the 4.6 floor their structured field carried) is structurally impossible. Behavioural prose is kept (daemon remediation, `daemon_not_running` / `live_unsupported_platform`). New regression test `test_live_command_descriptions_do_not_restate_the_structured_constraint` guards `game tree` / `perf monitors` / `daemon start` / `daemon stop`: their descriptions carry no `macOS`/`Linux`/`4.6` literal, while their `constraints` field is non-null. (I chose removal over generating the prose because it best matches #233's intent — make the precondition machine-discoverable *instead of* prose; happy to switch to generated prose if you'd rather keep it human-readable in `--help`.)

**code-review consider (key always present, value nullable):** made `LiveStackConstraints.min_godot_version` a required key with a nullable value (`str | None`, no default), so the self-described JSON Schema marks it required — matching the emitted objects (which always carry the key, `null` for `daemon stop`/`status`). The self-description test now asserts `min_godot_version` is in the model's `required` set.

PR #247, opened 2026-06-22
feat: gda harness lifecycle — version self-sync and paired uninstall (#225)

### LINKED ISSUE #225: Feature: gda harness lifecycle — version self-sync and paired uninstall

qinhaihong-red:
## Summary

Completes the gda harness lifecycle beyond #7's minimal install (ADR-0018):
- **Version self-sync** — `gda daemon start` re-materializes the installed harness when its version differs from the running `gda`, and reports it, so an agent never hits a stale-harness mismatch.
- **Paired uninstall** — `gda daemon uninstall` removes the `[autoload]` entry and the harness files together (a dangling autoload crashes an exported game).
- **Inertness** — verified the runtime gate keeps the harness resident-inert in a plain run and an exported build.

Closes #225.

## Design

**Version source = `HARNESS_VERSION` const, not the package version** (the harness changes far less often than `gda` ships). `_materialize()` prepends a `# gda-harness-version: <N>` header; because the header rides the existing idempotent content-compare, version self-sync **falls out of the existing compare** — re-materialize happens only on a version/body mismatch (no unconditional overwrite → no mtime bump that would trip the concurrent-editor prompt). Reported via two **additive** `DaemonStartResult` fields: `harness_synced` (true only on a real re-materialize) and `harness_version`.

**Uninstall is crash-safe** — strip the `[autoload]` entry first (atomic `write_text`), then delete files, so a mid-failure leaves only a harmless stray inert `.gd`, never a dangling autoload. Refused while a daemon is running (new `daemon_running` code, exit 6); idempotent no-op when not installed.

## ABI

- `daemon start` payload gains `harness_synced: bool` + `harness_version: str` (existing `installed_harness` / `already_running` unchanged).
- `daemon uninstall` → `DaemonUninstallResult{removed: bool}`; refused-while-running → `daemon_running` `GdaError` (category `live`, exit 6).
- **API note for #222's rebase:** `install_harness()` now returns `HarnessInstall` (was `bool`); direct callers updated to use `.changed`/`.synced`.

## Tests

`uv run pytest -m "not e2e"` → **800 passed** (+20; independently re-run by the reviewer). e2e → **7 passed** against real Godot 4.6.3 (run serially): start re-sync after a version bump, install→uninstall (paired/atomic/idempotent), uninstall-refused-while-running, strengthened plain-run inert boot (asserts no socket activity), and a new exported-PCK inert boot (`export run --mode pack` → run with no `gda-daemon` marker → clean inert boot — the exact failure ADR-0018 guards).

## Follow-up at merge

ADR-0018 outcome note recording the realized lifecycle (version self-sync + paired uninstall + inertness verification) to be appended when this lands. The `daemon_running` error-table row is already in ADR-0002 (kept in sync with `error_codes.py`).

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -137,6 +137,7 @@ operation, and parse codes the CLI assigns).
 | `engine_session_not_running` | `live` | `classifier` | `6` | The daemon is running but holds no live engine session to serve the live operation (Phase 2). |
 | `engine_disconnected` | `live` | `classifier` | `6` | The engine session disconnected before the live operation returned — the game crashed or the harness connection dropped (Phase 2). |
 | `live_timeout` | `live` | `classifier` | `6` | A live operation did not return from the engine session before the daemon's timeout (Phase 2). |
+| `daemon_running` | `live` | `classifier` | `6` | A daemon-lifecycle command was refused because a gda-daemon is running for the project; stop it first with `gda daemon stop` (Phase 2, #225). |
 | `live_node_not_found` | `live` | `classifier` | `6` | A live game operation's node path does not resolve to a node in the running scene tree (Phase 2, #220). |
 | `live_unknown_property` | `live` | `classifier` | `6` | A live game get or set targeted a property the running node does not expose for the requested operation (no such readable storage property for get, or no such settable property for set) (Phase 2, #220). |
 | `live_uncoercible_value` | `live` | `classifier` | `6` | A live game set value cannot be coerced to the running property's declared Godot type (Phase 2, #220). |
--- a/docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md
+++ b/docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md
@@ -39,6 +39,22 @@ so two instances can touch the project at once.
 > grows, the treatment to revisit is per-module fragments + an automatic summary
 > (the same direction RULES.md flags for central registries), under its own ADR.
 
+> **Outcome (2026-06-22, #225 / PR #247) — the harness lifecycle is complete:
+> version self-sync and paired uninstall.** Point 1's "self-syncs the harness to the
+> running `gda` version" is realized: `_materialize` prepends a `# gda-harness-version:
+> <N>` header (from a `HARNESS_VERSION` const — NOT the package version, since the
+> harness changes far less often than `gda` ships), so the version check rides the
+> existing idempotent content-compare: a mismatch re-materializes, a match is a no-op
+> (never an unconditional overwrite, which would bump mtime and trip the concurrent-
+> editor prompt of point 4). `gda daemon start` runs this self-sync **whether or not a
+> daemon is already up** (PR #247 review), so upgrading `gda` never strands a stale
+> harness; `harness_synced` is true only on a real stale→current rewrite, distinct from
+> a first install (`installed_harness`). The paired `gda daemon uninstall` strips the
+> `[autoload]` entry **first**, then deletes the files — crash-safe ordering, so a
+> mid-failure leaves a harmless stray inert `.gd`, never a dangling autoload (point 3);
+> the autoload edit is scoped to the `[autoload]` section so a same-named key elsewhere
+> is never touched. Verified resident-inert in a plain run and an exported PCK.
+
 ## Decision
 
 **1. The harness is an installed autoload, not a runtime injection.** `gda` bundles
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -19,6 +19,7 @@ from gda.daemon_ops import (
     run_daemon_start_operation,
     run_daemon_status_operation,
     run_daemon_stop_operation,
+    run_daemon_uninstall_operation,
 )
 from gda.errors import (
     Failure,
@@ -65,6 +66,8 @@ from gda.models import (
     DaemonStatusResult,
     DaemonStopParams,
     DaemonStopResult,
+    DaemonUninstallParams,
+    DaemonUninstallResult,
     DiagErrorsParams,
     DiagErrorsResult,
     DiagLogParams,
@@ -1096,12 +1099,23 @@ DAEMON_STATUS_COMMAND: HeadlessCommand[DaemonStatusResult] = HeadlessCommand(
     output_model=DaemonStatusResult,
 )
 
+DAEMON_UNINSTALL_COMMAND: HeadlessCommand[DaemonUninstallResult] = HeadlessCommand(
+    operation="daemon-uninstall",
+    input_model=DaemonUninstallParams,
+    output_model=DaemonUninstallResult,
+)
+
 # The daemon lifecycle commands run a process-management recipe (gda.daemon_ops),
 # not the sentinel pipeline — the same shape as `export run`. They carry no Godot
 # execution channel, so they are routed by command identity, shared by the argv
 # bodies and the --params-json path so both forms drive the SAME recipe.
 _DAEMON_COMMANDS = frozenset(
-    {DAEMON_START_COMMAND, DAEMON_STOP_COMMAND, DAEMON_STATUS_COMMAND}
+    {
+        DAEMON_START_COMMAND,
+        DAEMON_STOP_COMMAND,
+        DAEMON_STATUS_COMMAND,
+        DAEMON_UNINSTALL_COMMAND,
+    }
 )
 
 
@@ -1118,6 +1132,8 @@ def _daemon_dispatch(
         outcome = run_daemon_start_operation(resolved, godot)
     elif cmd is DAEMON_STOP_COMMAND:
         outcome = run_daemon_stop_operation(resolved)
+    elif cmd is DAEMON_UNINSTALL_COMMAND:
+        outcome = run_daemon_uninstall_operation(resolved)
     else:
         outcome = run_daemon_status_operation(resolved)
     if isinstance(outcome, Failure):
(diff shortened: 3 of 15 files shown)


qinhaihong-red:
## Review pass: `review` (Standards / Spec)

Fixed point: `origin/main` at `89a07f2`; PR head: `1d14c4a`.

## Standards

Hard violations: none found.

The diff keeps the documented registry contract: `daemon_running` is added to both `src/gda/error_codes.py` and ADR-0002's table, matching ADR-0002's rule that `GdaError.code` changes update the Python registry and mirrored ADR table. The new `gda daemon uninstall` command also has typed params/result models, `--schema`, renderer coverage, and `--params-json` routing, consistent with ADR-0004's self-description gate.

Judgement calls:

- `src/gda/daemon_ops.py:168`: the already-running `daemon start` branch explicitly "syncs nothing". ADR-0018 says `gda daemon start` auto-installs when missing and self-syncs the harness to the running `gda` version. This becomes a concrete Spec concern below, but as a Standards point it is a judgement call because the daemon is already live and repeated `start` has historically been idempotent.
- `src/gda/cli.py:1102`: `DAEMON_UNINSTALL_COMMAND` inherits the default `kind=headless` even though the nearby lifecycle comment says daemon commands are process-management recipes with no Godot execution channel. ADR-0004/0017 define `kind` as the static execution channel, but the repo has no documented `daemon`/`local` kind and the existing daemon lifecycle commands already use this shape, so I would not block this PR on it.

## Spec

Required changes:

- `src/gda/daemon_ops.py:168-179`: `daemon start` does not self-sync when a daemon is already running. The spec says: "on `gda daemon start`, re-materialize the installed harness when its version differs from the running `gda`, and report it — so an agent never hits a stale-harness mismatch." In this branch the code returns before `install_harness(project)`, so a stale installed header remains stale and the result reports `harness_synced: false`. I verified this with a local probe: an installed `# gda-harness-version: stale-old` plus `daemon_pid != None` returns `harness_version: "stale-old"` and does not rewrite the file. Either `start` needs to reject/force restart/sync in this state, or the spec needs an explicit already-running exception with a safe remediation.
- `src/gda/harness/install.py:168`: `harness_synced` is reported too broadly. The implementation sets `synced=materialized`, so a first install reports `harness_synced: true`; the tests lock this in at `tests/test_e2e_daemon.py:171`. The implementation spec says `harness_synced` is "true only on a real version-mismatch rewrite" and "not merely adding the autoload entry." A first install should remain visible through `installed_harness: true`, but `harness_synced` should be reserved for an already-installed stale/body-mismatched harness unless the public API text is deliberately changed.
- `src/gda/harness/install.py:181-188`: uninstall removes any line whose stripped text starts with `GdaHarness=`, regardless of section. The spec asks to remove the `[autoload]` entry and the harness files together. A project setting or other section line with the same key would be removed even though it is not the autoload entry. Please scope removal to the `[autoload]` section.

Implemented/covered: CLI/schema/error-code surface, paired uninstall ordering, plain-run inertness, and exported-PCK inertness coverage are present.

Summary: Standards has 0 hard violations and 2 judgement calls; Spec has 3 required findings. The worst issue is the already-running `daemon start` path leaving a stale harness unsynced.

qinhaihong-red:

Verdict: request changes.

### Correctness

Required findings:

- `src/gda/daemon_ops.py:168-179`: `daemon start` skips harness self-sync when a daemon is already running. That leaves the exact stale-harness upgrade path unresolved: if `gda` is upgraded while an old daemon remains up, a subsequent `gda daemon start` returns `already_running: true`, reports the stale installed version, and never calls `install_harness(project)`. I reproduced the branch locally with an installed `# gda-harness-version: stale-old`: the result was `harness_synced: false`, `harness_version: "stale-old"`, and the file stayed stale. This undermines the PR's stated guarantee that an agent never hits a stale-harness mismatch. Please add a concrete policy for this state: refuse with a remediation, force restart/resync safely, or otherwise make the stale condition impossible to miss.
- `src/gda/harness/install.py:168`: `HarnessInstall.synced` is currently just `materialized`, so a first install reports `harness_synced: true`. The new field is meant to tell agents that an already-installed harness was brought back in sync; first install is already represented by `installed_harness: true`. As written, consumers cannot distinguish "installed for the first time" from "stale harness was corrected", and tests assert the broader behavior. Please reserve `harness_synced` for an actual stale/body-mismatch rewrite, or update the API/spec wording intentionally.
- `src/gda/harness/install.py:181-188`: `_remove_autoload()` removes every stripped line that starts with `GdaHarness=`, independent of the current section. `daemon uninstall` should remove the `GdaHarness` row from `[autoload]`, not a same-named key elsewhere in `project.godot`. Please scope this parser to the `[autoload]` section so uninstall does not delete unrelated project data.

### Readability / Simplicity

The new lifecycle code is straightforward and matches the surrounding recipe style. The dataclass return values make the daemon payload mapping easier to follow than the previous bool return.

### Architecture

The implementation mostly fits the existing boundaries: harness file operations stay in `gda.harness.install`, lifecycle orchestration stays in `daemon_ops.py`, CLI dispatch remains centralized, and result rendering stays type-dispatched. One non-blocking concern: `daemon uninstall` appears in schema as `kind: headless`, inherited from the current daemon-command pattern, even though daemon lifecycle commands are process-management recipes rather than true headless engine operations. I would not block this PR on that because the existing daemon commands already do it, but it is worth resolving when the command-kind vocabulary is revisited.

### Security

No new security issue found. This continues to operate inside the documented trusted-project / single-driver boundary.

### Performance

No performance concern found. The extra file read/write work happens on explicit daemon lifecycle commands and is bounded.

### Verification

I ran:

- `uv run pytest tests/test_harness_install.py tests/test_daemon_lifecycle_payload.py tests/test_daemon_ops.py tests/test_error_registry.py tests/test_render.py` -> 59 passed
- `uv run pytest -m "not e2e"` -> 800 passed, 270 deselected
- Manual probe for the already-running stale-harness branch described above

I did not rerun the real-Godot e2e suite in this pass because the non-e2e suite already reproduces the changed contract surface and the remaining blockers are uncovered logic/contract cases.

qinhaihong-red:
Thanks — all 3 Required findings fixed in `e8cc7c9`. Fast suite → 802 passed; daemon e2e → 10 passed (real Godot 4.6.3).

**Finding 1 — already-running `daemon start` left the harness unsynced (the worst one).** The already-running branch now runs `install_harness(project)` instead of returning early, so upgrading `gda` while an old daemon stays up re-materializes a stale harness rather than leaving it stale. New test `test_already_running_start_resyncs_a_stale_harness` reproduces your probe (`# gda-harness-version: stale-old` + a live pid → `harness_synced: true`, `harness_version` current, file rewritten on disk). `harness_synced` is true only on a real stale→current rewrite, so a steady-state repeat start still reports false and writes nothing.

> On the running session specifically: the harness is read at session launch, and sessions launch lazily / relaunch when the prior one dies (ADR-0017). So a `daemon start` before any live op — the common flow — fully resyncs the next session; re-materializing the on-disk file is the correct sync point. A session already live keeps its loaded autoload until it relaunches (inherent to the session-reuse model), which is out of scope for this slice; the on-disk guarantee is now solid.

**Finding 2 — `harness_synced` too broad.** Now `synced = existed and materialized`: a first install is reported by `installed_harness` / `changed`, and `harness_synced` is reserved for rewriting an **already-installed** harness (stale→current). Updated the tests that locked in the old "first install IS a sync" semantics.

**Finding 3 — `_remove_autoload` not section-scoped.** Removal is now scoped to the `[autoload]` section via a shared `_section_of` helper; I applied the same scoping symmetrically to `_ensure_autoload`'s re-point so install can't touch a same-named key elsewhere either. New test `test_uninstall_scopes_removal_to_the_autoload_section` (a decoy `GdaHarness="keep-me"` in another section survives uninstall while the `[autoload]` entry is removed).

**Judgement call — `daemon uninstall` shows `kind: headless`.** Left as-is per your "would not block" note: it matches the existing daemon lifecycle commands, and a dedicated `daemon`/`local` kind is a separate command-kind-vocabulary decision rather than something to introduce in this slice.

qinhaihong-red:
Added the **ADR-0018** Outcome note (`8e019eb`) recording the realized harness lifecycle — version self-sync (run whether or not a daemon is already up) and the crash-safe, `[autoload]`-scoped paired uninstall — so the decision doc reflects what this PR ships.

qinhaihong-red:
Round-2 fix in `a45fa6d`. Dropped `_ensure_autoload`'s global `if line in text` early return — it no-op'd whenever the exact GdaHarness autoload line appeared in **any** section (a same-named decoy elsewhere), wrongly skipping the real `[autoload]` entry; that was the asymmetry left after round 1 scoped `_remove_autoload`/the re-point. The "already present" decision is now scoped to `[autoload]`: only an exact GdaHarness line **inside** `[autoload]` is a no-op, otherwise re-point in place or insert — fully symmetric with `_remove_autoload`. New regression test `test_install_is_not_fooled_by_the_exact_line_outside_autoload` (an exact decoy line in another section no longer suppresses the install). Fast: 803 passed; daemon e2e: 10 passed (real Godot 4.6.3).

PR #248, opened 2026-06-22
feat: screen — runtime viewport capture of the running game (#222)

### LINKED ISSUE #222: Feature: `screen` runtime viewport capture of the running game

qinhaihong-red:
## Summary

Adds **`screen` runtime viewport capture** of the running game over the LIVE channel: a single screenshot and a multi-frame capture, returned as PNG image data. Closes the build→run→**observe**→verify loop with on-screen pixels.

Closes #222.

> **Stacked on #225** (base = `feat/225-harness-lifecycle`). Built on #225's evolved `daemon start` surface (added `windowed` *alongside* `harness_synced`/`harness_version`, no conflict). **Merge after #225 (PR #247).** GitHub will retarget this PR to `main` once #225 merges.

## Design

**Windowed session = a start-time declared mode** — `gda daemon start --windowed`. `--headless` uses the dummy `DisplayServer` and cannot capture pixels, so the daemon launches a windowed session only when explicitly asked. A screen op on a headless session fails with the typed `live_display_unavailable` (self-revealing remediation). This **refines ADR-0017's** anticipated "switch to windowed on the first viewport op" (a lazy per-op relaunch) — rejected because a mid-session relaunch throws away runtime state and violates ADR-0020 session-bound consistency — while keeping #7's cheap `--headless` default for non-visual live ops. (ADR-0017 outcome note to be appended at merge.)

**Two sibling commands** under a new `screen` group (the viewport is the domain object, ADR-0019), mirroring `perf monitors`/`perf monitor`:
- `gda screen capture --output PATH [--inline]` → `{path, width, height, bytes, format, inline}`. Writes the PNG and returns the path; `--inline` additionally embeds base64.
- `gda screen frames --frames N --output-dir DIR` → `{count, frames:[{path,width,height,bytes,format}]}`, path-only (an N-frame base64 would blow the agent's context). `--frames` bounded 1..600.

Both are `kind = LIVE`; the wire stays the existing UTF-8 sentinel JSON (harness base64-encodes the PNG; the CLI-side recipe in `screen_ops.py` decodes and writes the file). **GPU-capture timing:** both ops sample **inside a window tick** (reusing #223's `_begin/_advance/_finish_window`, no `_process` change) so `get_viewport().get_texture().get_image()` reads a rendered, non-empty frame; a `_capture_frame()` guard defends against an empty image. The headless guard fires before any window opens.

Schema-emission path (`headless.py`/`surface.py`/`CommandSchema`) untouched — declaring `kind=ExecutionKind.LIVE` makes `"kind":"live"` appear automatically, and #233's predicate auto-covers the `screen` constraints.

## Tests

`uv run pytest -m "not e2e"` → **818 passed** (independently re-run by the reviewer). Real-engine **windowed** e2e (`tests/test_e2e_screen.py`, run serially with a real display) → **5 passed**: single-frame capture (asserts PNG magic + dims>0), inline base64, 3-frame window, headless-session → `live_display_unavailable` (exit 6), no-daemon → `daemon_not_running`. Linux-CI Xvfb path documented in the module docstring.

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -148,6 +148,7 @@ operation, and parse codes the CLI assigns).
 | `live_invalid_key` | `live` | `classifier` | `6` | A live input key event named a key the engine could not resolve to a keycode (Phase 2, #221). |
 | `live_unknown_action` | `live` | `classifier` | `6` | A live input action targeted an action the running game's InputMap does not declare (Phase 2, #221). |
 | `live_invalid_event_spec` | `live` | `classifier` | `6` | A live input sequence event has a type the harness does not recognize (Phase 2, #221). |
+| `live_display_unavailable` | `live` | `classifier` | `6` | A live `screen` capture ran on a headless engine session (the dummy DisplayServer cannot read pixels); start the daemon windowed with `gda daemon start --windowed` (Phase 2, #222). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 
 ## Considered options
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -64,6 +64,20 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > stalled-engine guard. The base is general: #222's viewport capture reuses it by
 > supplying its own sampler/finalizer, with no `_process` change.
 
+> **Outcome (2026-06-22, #222 / PR #248) — the windowed session is a start-time
+> declared mode, refining the #7 note above.** #222 adds `screen capture` / `screen
+> frames` (viewport capture), the first ops needing a real `DisplayServer`. The #7 note
+> anticipated the daemon "switches to a windowed session when the first capturing op
+> lands"; #222 does **not** switch mid-session. A relaunch would discard the session's
+> accumulated runtime state and silently re-run autoloads, breaking ADR-0020's
+> session-bound consistency and this ADR's "deliberate, declared effect". Instead the
+> display mode is declared once at `gda daemon start --windowed`; the `--headless`
+> session stays the cheap default (decision 2's "windowed by default" becomes an
+> explicit per-session opt-in), and a capture op on a headless session fails with the
+> typed `live_display_unavailable` naming the remediation. The capture handlers reuse
+> the #223 time-windowed base — a 1-frame window for a single shot, an N-frame window
+> for a sequence — so the one-shot RPC contract holds.
+
 ## Decision
 
 **1. An execution-channel selector, chosen per command by a static `kind`.**
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -146,6 +146,10 @@ from gda.models import (
     ResourceUidParams,
     ResourceUidResult,
     SchemaAllParams,
+    ScreenCaptureParams,
+    ScreenCaptureResult,
+    ScreenFramesParams,
+    ScreenFramesResult,
     SceneCreateParams,
     SceneCreateResult,
     SceneDeleteParams,
@@ -185,6 +189,10 @@ from gda.models import (
 )
 from gda.project import resolve_project_dir
 from gda.runner import GodotRunner
+from gda.screen_ops import (
+    run_screen_capture_operation,
+    run_screen_frames_operation,
+)
 from gda.surface import build_surface_manifest
 
 app = typer.Typer(
@@ -314,6 +322,17 @@ input_app = typer.Typer(
     no_args_is_help=True,
 )
 app.add_typer(input_app, name="input")
+
+# The screen command group (Phase 2, ADR-0019): the running game's VIEWPORT is the
+# domain object (not under `game`, whose object is the runtime scene graph). Both
+# commands are LIVE (kind = LIVE), routed through gda-daemon to a WINDOWED engine
+# session (`gda daemon start --windowed`); on a headless session a capture is the
+# typed `live_display_unavailable` (#222).
+screen_app = typer.Typer(
+    help="Capture the running game's viewport (live; macOS/Linux only, windowed session).",
+    no_args_is_help=True,
+)
+app.add_typer(screen_app, name="screen")
 
 # The daemon command group (Phase 2, ADR-0017): gda's own per-project daemon
 # lifecycle — a deliberate extension of ADR-0005's domain-object grouping to an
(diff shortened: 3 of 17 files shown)


qinhaihong-red:
## `review` skill pass for PR #248

Diff reviewed: `git diff base-225...pr-248`

Commits reviewed:
- `fedc5c7` `feat(daemon): start --windowed declares engine session display mode (#222)`
- `ed7430f` `feat(errors): register live_display_unavailable for screen capture (#222)`
- `09d1ca9` `feat(screen): screen capture/frames -- LIVE viewport capture (#222)`
- `3641faa` `feat(harness): screen-capture/screen-frames ops with headless guard (#222)`
- `8c9d0d0` `test(e2e): screen windowed capture/frames, headless guard, no-daemon (#222)`

## Standards

1. **Hard violation: `screen` input schema is not the command contract.**

   Both new `screen` descriptors use input models that omit required operational CLI inputs: `screen capture` requires `--output` (`src/gda/cli.py:1313`), and `screen frames` requires `--output-dir` (`src/gda/cli.py:1362`). The models explicitly treat the output path as not a harness param (`src/gda/models.py:3070`), but that also removes it from the public input contract.

   This violates ADR-0004's schema contract for command input and ADR-0015's rule that the params model is the single source of truth for both emitted `input` schema and `--params-json` parsing (`docs/adr/0015-structured-params-input-abi.md:24`).

2. **Hard violation: `--params-json` is not a complete params source for `screen`.**

   The new params-json dispatcher tries to recover `output` / `output_dir` from `ctx.params` (`src/gda/cli.py:528`), but ADR-0015 says `--params-json` supplies one JSON object of the command params and is mutually exclusive with individual operation options (`docs/adr/0015-structured-params-input-abi.md:18`, `docs/adr/0015-structured-params-input-abi.md:44`). Since those paths are absent from the model, a valid JSON-only invocation cannot express where files should be written.

Judgment call: the default-headless `daemon start --windowed` design appears to drift from ADR-0017's earlier “game runs windowed by default” language, but issue #222 explicitly refines that decision, so I would treat that as ADR follow-up rather than a blocking standards violation.

## Spec

1. **Partial/wrong: `DaemonStartResult.windowed` does not reliably surface the session mode.**

   Spec quote: “Surface `windowed` on `DaemonStartResult` so an agent can see the mode.”

   In the already-running path, `run_daemon_start_operation(..., windowed=True)` always returns `windowed=False` (`src/gda/daemon_ops.py:175`), and `tests/test_daemon_lifecycle_payload.py:168` codifies that behavior. That means an agent cannot tell whether the existing session is actually windowed; `gda daemon start --windowed` can report `false` even if the already-running daemon was launched windowed.

2. **Partial/wrong: the `--params-json` path is wired but cannot successfully produce screen image output.**

   Spec quotes: “Same `--json` / `GdaError` contract” and “Image returned as a written PNG file path by default.”

   The screen commands opt into `HeadlessCommand` `--params-json` wiring, but `src/gda/headless.py:197` forbids operation options alongside `--params-json`; meanwhile `src/gda/cli.py:528` expects `ctx.params["output"]` / `output_dir`, and `src/gda/screen_ops.py:107` writes to that path. A valid successful `screen capture --params-json '{}'` has no way to provide the required output path; if the live op succeeds, it falls out of the typed `GdaError` contract into Python misuse.

No additional scope-creep findings: `headless.py` and `surface.py` were not modified, and `kind=ExecutionKind.LIVE` is used rather than reimplementing schema kind emission.

Summary: Standards found 2 hard violations; Spec found 2 partial/wrong requirements. Worst issue: `screen`'s public input contract omits the required output paths, breaking schema / MCP / `--params-json` parity and causing an unstructured failure on a successful structured invocation.

qinhaihong-red:
## `agent-skills:code-review-and-quality` pass for PR #248

Verdict: **Request changes**

### Findings

1. **Correctness / API contract: `screen` cannot be used through the structured input path.**

   `ScreenCaptureParams` is empty (`src/gda/models.py:3070`) and `ScreenFramesParams` only carries `frames` (`src/gda/models.py:3100`), so the generated `--schema` / MCP input contract has no way to provide `output` or `output_dir`. At the same time, the runtime path still requires those values from `ctx.params` (`src/gda/cli.py:528`) and writes through them (`src/gda/screen_ops.py:107`). Because the central `--params-json` machinery rejects individual operation options alongside `--params-json`, `screen capture --params-json '{}' --output shot.png` is a structured `usage_error`, while a successful `screen capture --params-json '{}'` reaches `_write_png(..., None)` and raises `AttributeError`.

   This breaks the agent-facing structured input ABI and the issue #222 “Same `--json` / `GdaError` contract” requirement. The likely fix is to put `output` / `inline` into `ScreenCaptureParams` and `frames` / `output_dir` into `ScreenFramesParams` (using the existing normalized path type for filesystem paths), then have both argv and `--params-json` construct the same params model. The recipe can still send only the harness-relevant fields over the live op.

2. **Correctness: `DaemonStartResult.windowed` reports a value the implementation does not know.**

   The already-running branch always returns `windowed=False` (`src/gda/daemon_ops.py:175`), and the test locks that in (`tests/test_daemon_lifecycle_payload.py:168`). That makes `gda daemon start --windowed` an unreliable way for an agent to confirm the current session mode: an already-running windowed daemon is reported as non-windowed, and an already-running headless daemon is indistinguishable from the “unknown” case. Issue #222 explicitly asked to surface `windowed` so an agent can see the mode, so this should either persist/query the actual daemon session mode or avoid reporting `false` as if it were known.

3. **Standards / path handling: new output paths skip the repo's filesystem path normalization rule.**

   `screen capture --output` and `screen frames --output-dir` are accepted as raw `Path` options (`src/gda/cli.py:1313`, `src/gda/cli.py:1362`). ADR-0006 says filesystem paths get `~` expanded once at the CLI/model boundary, and `export run --output` already does this explicitly (`src/gda/cli.py:2676`). As written, `--output ~/shot.png` is treated as a literal relative `~` path. Moving these fields into params models with `NormalizedPath` would also close this gap.

### Verification

- Ran `uv run pytest tests/test_screen_commands.py tests/test_daemon_lifecycle_payload.py tests/test_error_registry.py -q` in a clean `/tmp` clone: **35 passed**.
- Reproduced the successful-`--params-json` failure with an injected live runner: no `--output` raises `AttributeError`; adding `--output` is rejected as `usage_error`.

No additional security or performance issues stood out in this pass.

qinhaihong-red:
Added the **ADR-0017** Outcome note (`3e35bf9`) recording that the windowed session is a **start-time declared mode** (`daemon start --windowed`), refining the #7 note's anticipated "switch to windowed on the first capturing op" — a mid-session switch would break ADR-0020 session-bound state. Keeps the decision doc consistent with the code.

qinhaihong-red:
Thanks — all four findings fixed in `1ea3bac`. Fast suite → 823 passed; screen e2e → 5 passed (real Godot 4.6.3).

**Finding 1 (hard) + Spec 2 — `screen` input contract / `--params-json`.** Moved `output` + `inline` into `ScreenCaptureParams` and `output_dir` into `ScreenFramesParams` as the single source of truth for both the emitted `--schema` `input` and `--params-json` parsing (ADR-0004/ADR-0015). Both the argv body and the `--params-json` dispatch now build the **same** model — `ctx.params` recovery is gone — while the recipe still writes the file CLI-side and the live op carries no path. So:
- `screen capture --params-json '{}'` → structured `invalid_params` (was an `AttributeError` on `_write_png(..., None)`).
- `screen capture --params-json '{"output":"shot.png"}'` → writes the PNG, returns the path.
- the output path is now in `--schema` / the MCP input contract.
New tests cover each (params-json supplies output/output_dir, missing output → `invalid_params`, `--schema` carries the contract).

**Finding 3 — path normalization.** `output` / `output_dir` are typed `NormalizedPath`, so `~` is expanded once at the model boundary (ADR-0006), the same rule `export run --output` follows. (Test asserts `~/shot.png` is not kept literal.)

**Finding 2 / Spec 1 — `DaemonStartResult.windowed`.** Now `Optional[bool]`. A fresh start reports the launched mode; an idempotent **already-running** start reports `None` ("not determined here") instead of a misleading `False` — a daemon launched windowed no longer reads as headless. I took your *"avoid reporting false as if known"* option rather than persist/query, because:
- The CLI `daemon status` op is currently **pidfile-only** — it never round-trips to the daemon, so it has no mode to report either. The daemon *does* already reply `windowed` over `STATUS_OP`, but nothing on the CLI side reads it.
- Surfacing a *running* daemon's actual mode therefore belongs on `daemon status` (make it query the daemon and add `windowed` to `DaemonStatusResult`), not on the start result. That's a clean, self-contained follow-up — happy to do it if you'd prefer the agent be able to read a running daemon's mode directly; this PR keeps the start result honest in the meantime.

**Judgement call — `--windowed` vs ADR-0017's "windowed by default".** Recorded in the ADR-0017 Outcome note (`3e35bf9`, the comment above).

(Note: this PR is stacked on #225, which advanced during its review; it still needs a rebase onto the final #225 at merge — the only conflict is the `daemon_ops` already-running branch, which now carries #225's harness self-sync and this PR's `windowed=None`.)

qinhaihong-red:
Recheck follow-up in `145427a`: cleaned the stale "`--output-dir` is mutually exclusive with `--params-json`" comment in the over-range-frames test — `output_dir` is now a required params field carried IN `--params-json`. Also added a valid `output_dir` to that test's JSON so the over-range `frames` is the **sole** failing constraint (it was otherwise also tripping on the now-missing required `output_dir`, passing for a second unintended reason). Test-only; the assertion still pins `invalid_params`. `tests/test_screen_commands.py` → 16 passed.

PR #262, opened 2026-06-23
fix: correct outdated README and refresh package description

qinhaihong-red:
## What & why

The README still read like a mid-Phase-1 snapshot and **understated the shipped product**. This rewrites the public-facing copy (README + `pyproject.toml` description/keywords) around three weighted principles — **user-facing onboarding (40%) · feature highlight (35%) · SEO/discoverability (25%)** — and corrects factual drift.

### Correctness fix (why this is `fix:`)
The old *Roadmap*/*Project status* claimed the live catalogue (input simulation, viewport capture, performance/signal monitoring) was *"in progress"* — **it ships today**, with e2e tests. The PR documents `game` / `diag` / `perf` / `input` / `screen` + the `daemon` lifecycle as a first-class command group, and adds the previously-undocumented `gda schema` meta command. It also refreshes the stale, understated package `description`/`keywords`.

### Highlights
- Reframe internal **Phase 1/Phase 2** → user-facing **Headless / Live** modes (CONTEXT.md / ADRs keep the Phase vocabulary).
- Add a **platform & version support matrix** (Headless 4.4+ macOS/Linux, Windows untested-but-likely; Live 4.6+ macOS/Linux via Unix domain sockets) + a platform badge.
- Highlight **CLI ⇄ MCP interchangeability** (one tool, two mechanically-identical surfaces).
- Delete stale/dead sections (Project status, Roadmap, Acknowledgements, Requirements); keep one honest **pre-1.0** signal.
- Re-sequence install → quick start (ends on a live demo) → how-it-works → command reference → configuration; collapse deep internals into `<details>`.
- Cut cognitive load: drop the leaky `godot --headless` mechanism, the misleading macOS-only "default path", and internal issue links from the user-facing copy; keep ADR/CONTEXT refs only behind folds and in Contributing.
- Refresh the outdated dev module layout.

Net: **README −90 lines**, deep detail folded (not lost).

## Release impact
This is a `fix:` → release-please will cut **`0.1.36`** (patch). Per the repo's `bump-patch-for-minor-pre-major: true`, `feat` would also yield 0.1.36 pre-1.0; a `0.2.0` milestone bump would need a breaking change or `Release-As:` and is intentionally **not** done here (version tracks code; this PR is docs/metadata only).

## Verification
- Documented command surface cross-checked against `gda schema` (62 commands / 14 groups) — exact match.
- No dead links / stale anchors; all referenced ADR/docs paths exist.
[...]

PR #264, opened 2026-06-24
docs: restore per-agent user-scope MCP registration recipes

qinhaihong-red:
## What & why

The README rewrite (#262) reorganized the MCP registration recipes into per-agent `<details>` blocks but **dropped the user-scope variants for Codex and Cursor** (Claude Code's was retained). That left the three agents inconsistent and diverged from the canonical `docs/gda-mcp-registration.md` and the v0.1.35 README.

## Changes
- **Codex** — restore user scope: `codex mcp add gda-mcp --env GDA_PROJECT=... -- uvx --from "gda[mcp]" gda-mcp`.
- **Cursor** — restore user scope: `~/.cursor/mcp.json` with an absolute `GDA_PROJECT`, plus the Settings → MCP UI note.
- Tidy: drop a stray trailing space after `<summary>Cursor</summary>`; even out the blank line before the Codex block's `</details>`.

All three agents now show **both project and user scope**, matching `docs/gda-mcp-registration.md`. Verified: user-scope commands match the canonical recipes doc; `<details>` blocks balanced; renders correctly on GitHub.

Type is `docs:` → no release. If you want this reflected on the PyPI README (currently 0.1.36 lacks it), say so and I'll retitle to `fix:` for a 0.1.37 patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,13 @@ uvx --from "gda[mcp]" gda-mcp
 workspace `roots`, otherwise the working directory. An explicitly set
 `GDA_PROJECT` that is not a valid project is reported as an error, not silently replaced.
 
-<details>
-<summary>Register with Claude Code / Codex / Cursor</summary>
 
-**Claude Code** — project scope, `.mcp.json` at the repo root (auto-detects the project via `roots`):
+#### Register with Coding Agents
+
+<details>
+<summary>Claude Code</summary>
+
+Project scope, `.mcp.json` at the repo root (auto-detects the project via `roots`):
 
 ```json
 {
@@ -123,7 +126,12 @@ User scope (every project) — the CLI, which writes `~/.claude.json`:
 claude mcp add --scope user gda-mcp -- uvx --from "gda[mcp]" gda-mcp
 ```
 
-**Codex** — project scope, `.codex/config.toml` at the repo root (the project must be trusted):
+</details>
+
+<details>
+<summary>Codex</summary>
+
+Project scope, `.codex/config.toml` at the repo root (the project must be trusted):
 
 ```toml
 [mcp_servers.gda-mcp]
@@ -134,7 +142,20 @@ args = ["--from", "gda[mcp]", "gda-mcp"]
 GDA_PROJECT = "/absolute/path/to/your/godot/project"
 ```
 
-**Cursor** — project scope, `.cursor/mcp.json` at the repo root (`${workspaceFolder}`
+User scope (every project) — the same table in `~/.codex/config.toml`, or add it with the
+CLI (Codex has no workspace variable, so `GDA_PROJECT` stays an absolute path):
+
+```bash
+codex mcp add gda-mcp --env GDA_PROJECT=/absolute/path/to/your/godot/project -- \
+  uvx --from "gda[mcp]" gda-mcp
+```
+
+</details>
+
+<details>
+<summary>Cursor</summary>
+
+Project scope, `.cursor/mcp.json` at the repo root (`${workspaceFolder}`
 tracks the open project):
 
 ```json
@@ -152,6 +173,10 @@ tracks the open project):
   }
 }
 ```
+
+User scope (every project) — the same config in `~/.cursor/mcp.json`, but set `GDA_PROJECT`
+to an absolute path (`${workspaceFolder}` is only reliable in the project-level file).
+Cursor has no `mcp add` command — register via the JSON above or the Settings → MCP UI.
 
 > GUI-launched clients (Cursor, Claude Desktop) start with a minimal `PATH`, so a bare
 > `uvx` may not resolve — the Cursor snippet repairs it in `env`; if `uvx` is still not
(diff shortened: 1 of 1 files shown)


qinhaihong-red:
## Review skill pass

## Standards

- `README.md`'s Cursor project-scope quick recipe now uses `"command": "uvx"` with no absolute command and no repaired `PATH` in `env`. ADR-0013 records the GUI-client constraint for Cursor: a bare `uvx` may not resolve, so recipes should use an absolute executable path or inject a full `PATH`. The caveat below the snippet helps, but the visible copy-paste recipe still violates that documented registration standard. Please either make the Cursor snippet use an absolute placeholder command, as the canonical registration guide does, or include the PATH repair directly in the snippet.

- Judgement call: the MCP project-resolution summary says roots are the auto-detected stand-in for CLI `--project`. The precedence is correct, but the analogy blurs ADR-0006 and ADR-0014: CLI project resolution is flag/env/cwd, while `gda-mcp` uses server context with env/`roots/list`/cwd. For user understanding, it would be clearer to describe roots as the protocol workspace fallback, not as a replacement for an explicit CLI flag.

## Spec

- The Cursor quick recipe is still copy-paste fragile against the canonical recipe. `docs/gda-mcp-registration.md` says Cursor is GUI-launched and should use an absolute executable path or repaired PATH; its Cursor project recipe uses an absolute `command`. This README change removes the previous PATH injection but leaves the snippet as `"command": "uvx"`. The follow-up note is useful, but the primary recipe no longer matches the safe canonical shape.

- The project-resolution explanation omits the validation rule for fallback candidates. The canonical registration guide and ADR-0014 say roots/cwd candidates are only used when they are real Godot projects containing `project.godot`; only an explicitly invalid `GDA_PROJECT` is treated as a strict error. README now mentions the invalid `GDA_PROJECT` behavior, but not that invalid roots/cwd are skipped, so the fallback can read more unconditional than the actual contract.

Summary: Standards found 1 required issue and 1 judgement call; Spec found 2 docs-completeness issues. Worst issue: the Cursor quick recipe is not reliably copy-pasteable for GUI-launched Cursor.

qinhaihong-red:

Verdict: request changes for documentation clarity. This is README-only; no runtime tests are needed. I verified the diff, checked `<details>` balance, ran `git diff --check`, checked local `codex mcp add --help` / `claude mcp add --help`, and confirmed PyPI currently publishes `gda` 0.1.36.

Correctness / readability:

- Required: `README.md:153` and `README.md:184` describe the Codex/Cursor configs as "User scope (every project)", but both snippets require an absolute `GDA_PROJECT`. For Codex and Cursor, the canonical guide says a global/user registration does not reliably follow each open project; it is pinned to the single `GDA_PROJECT` you set, and multi-project users should prefer project-scope registration. The current wording can make users think one global config automatically works across every Godot project. Please rephrase these two paragraphs along the lines of: "User/global scope - available from every workspace, but pins this one Godot project via absolute `GDA_PROJECT`; use project scope for several projects."

- Consider: `README.md:100` says the server takes the project and binary "from the recipe's `env`", but the very next bullet says the project may come from MCP roots or the server cwd, and the Claude Code project recipe intentionally has no env block. A small wording change like "The server resolves two pieces of context" would better match the actual resolution model.

Architecture/security/performance: no concerns for this docs-only change.

qinhaihong-red:
Thanks — all findings verified against ADR-0013/0006/0014 and `docs/gda-mcp-registration.md`, addressed in 456847e.

**Standards/Spec — Cursor snippet fragile (required):** Correct. GUI-launched Cursor needs an absolute executable (ADR-0013), and the canonical Cursor recipe uses an absolute `command`. Removing the macOS `PATH` blob earlier left a bare `uvx`. Fixed: the Cursor project snippet now uses a `/path/to/uvx` placeholder (filled via `which uvx`), and the caveat explains why. (Claude Code / Codex keep bare `uvx` — they're CLI-launched with a normal PATH.)

**Standards — roots/`--project` analogy (judgement):** Agreed. Reframed so roots is the **protocol's workspace fallback**, not a stand-in for the CLI flag: "The CLI resolves a project differently — `--project` → `GDA_PROJECT` → cwd; the MCP server has no flags, so `GDA_PROJECT` is the explicit override and the protocol's workspace `roots` are the auto-detected fallback."

**Spec — fallback validation rule omitted:** Correct (ADR-0014). Added: "A roots or cwd candidate is used only if it is a real Godot project (has `project.godot`), else resolution moves on; a *set-but-invalid* `GDA_PROJECT`, by contrast, is a reported error, not a silent fallback."

**Code-review — "User scope (every project)" misleading (required):** Correct. For Codex/Cursor an absolute `GDA_PROJECT` pins one project, so reworded both to "User scope (available everywhere, but pinned to one project) … use project scope for several projects." Claude Code keeps "every project" — it follows the open project via `roots`, no pinned `GDA_PROJECT`.

**Code-review — "from the recipe's env" (consider):** Adopted — now "The server resolves two pieces of context", since the project can come from roots/cwd and the Claude Code recipe has no `env` block.

Issue #278, opened 2026-06-24
feat(daemon): run a specific scene in a live session via `daemon start --scene`


qinhaihong-red:
## What to build

`gda`'s daemon-owned live `Engine session` always boots the project's `main_scene`; there is no way to run an arbitrary scene without first mutating `application/run/main_scene`. Godot exposes this natively as `godot --headless --path <proj> --scene <path|UID>` (verified this session: it runs the chosen scene and leaves `main_scene` untouched).

Thread an **optional scene selector** from the CLI into the daemon's session launch so the engine boots a chosen scene. Surface it as **`gda daemon start --scene <path|UID>`** (the mechanism). With no `--scene`, behaviour is unchanged (runs `main_scene`). Once the session runs the chosen scene, the entire existing live surface (`game tree/get/set`, `diag`, `perf`, `input`, `screen`) applies unchanged — that is the payoff: "run a specific scene" plus full live observation, for one small change.

Anchored by **ADR-0017** (live execution mechanism — see its amendment recording scene selection as an extension of Decision 2) and **ADR-0025** (surface inclusion principle: agent-value + structured-op fit over engine-CLI parity, which is why `run` is in scope while most engine flags are not). Consistent with **ADR-0005** (reserved `play`/`run` verbs) and **ADR-0019** (live-command placement by domain object). Trust is unchanged: running a scene executes `_ready`/game code — the same trusted-project boundary mutating ops already cross (ADR-0009). The `--json` / `--schema` / stable-error contract is identical to existing daemon commands.

An ergonomic `scene play <path>` / `game run` wrapper (a live op under its real domain group per ADR-0019) is a possible **follow-up**, not part of this slice.

## Acceptance criteria

- [ ] `gda daemon start --scene <path|UID>` launches the session on the given scene; the engine receives `--scene <path|UID>` as an engine option (before `--path`).
- [ ] With no `--scene`, the session still runs the project's `main_scene` — no behaviour change to existing callers.
- [ ] Accepts either a scene **path** or a **UID** (per Godot `--scene`); a non-existent scene yields a typed `GdaError`, not a silent fallback.
- [ ] e2e (real Godot 4.6+): with scenes A (main) and B, `daemon start --scene res://B.tscn` then `game tree` reports **B**'s tree (not A), and `project.godot` `main_scene` is **unchanged**.
- [ ] Unit coverage of the launch-argv construction: the `--scene` flag is inserted in the correct position when set, and omitted when not.
[...]

PR #279, opened 2026-06-24
docs(adr): surface-inclusion principle (ADR-0025) + ADR-0017 scene-selection amendment

qinhaihong-red:
Prerequisite decisions for the run-a-specific-scene slice (#278).

## ADR-0025 — surface inclusion principle (new)
Records *which* engine capabilities earn a `gda` command (ADR-0005 fixed naming, not inclusion). Criterion: **agent value + structured-operation fit, not engine-CLI parity.** Puts running a project/scene in scope; keeps editor/dev-tooling flags and **interactive** debugging (break/step/eval) out, while leaving read-only observability (errors+callstacks, perf) eligible. Gives the recurring "should gda support feature X?" question a durable, one-paragraph answer.

## ADR-0017 — amendment (scene selection)
Records that the live `Engine session` may boot a **chosen scene** via a `--scene <path|UID>` passthrough — an **extension** of Decision 2's running-game scope (not a reversal): same gda-owned game, headless by default, same harness and live surface; only the entry scene changes. Default (no `--scene`) unchanged. Editor "current scene" remains out of scope.

## Notes
- Docs-only (no package change); does not trigger a release.
- Unblocks #278 (run-a-scene slice), which implements the `--scene` passthrough.

### DIFF
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -82,14 +82,29 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > `main_scene`.** Decision 2 holds the running *game* but never fixed *which* scene it boots; the
 > session has so far run the project's configured `main_scene`. To serve "run a specific scene" —
 > the F6-equivalent the engine exposes natively as `godot --scene <path|UID>` (verified to run that
-> scene without mutating `main_scene`) — the session launch gains an **optional scene selector**,
-> threaded from the CLI to the daemon's session launch and passed to the engine as
-> `--scene <path|UID>` (an engine option, before `--path`). This is a deliberate **extension** of
-> Decision 2's running-game scope, not a reversal: still a gda-owned game, still headless by
-> default, still the same harness and live surface (ADR-0019 / 0020), only with a chosen entry
-> scene. It does **not** reach the out-of-scope editor context (there is no editor "current
-> scene"). The selector-less default is unchanged. Realized by the run-a-scene slice (#278); the
-> surface-inclusion rationale (why `run` is in scope at all) is recorded in ADR-0025.
+> scene without mutating `main_scene`) — the session launch gains an **optional scene selector**.
+>
+> **Public surface — `gda daemon start --scene <path|UID>`.** It is a **start-time daemon option**:
+> the daemon holds the value and passes it to the engine as `--scene <path|UID>` (an engine option,
+> before `--path`) when the session is **lazily launched on the first live op** (the #7 note above).
+> With no `--scene`, behaviour is unchanged (runs `main_scene`). The selector accepts a scene **path
+> or UID** (per Godot's `--scene`). Any `scene play` / `game run` ergonomic wrapper is a **separate
+> follow-up**, not part of this amendment.
+>
+> **Failure semantics — typed error, never a silent fallback.** A missing or non-existent scene
+> selector **must** surface a typed `GdaError` (a `LIVE` classifier-source code minted by the slice,
+> per decision 6 below); it must **not** silently fall back to `main_scene`. A typed failure beats a
+> wrong positive path the agent cannot detect.
+>
+> **Trust boundary unchanged.** Launching a chosen scene runs that scene's `_ready` and game code —
+> the **same trusted-project / project-code execution surface** that running `main_scene` and every
+> mutating op already cross (ADR-0009, ADR-0018), not a new boundary. It does **not** reach the
+> out-of-scope editor context (there is no editor "current scene").
+>
+> This is a deliberate **extension** of Decision 2's running-game scope, not a reversal: still a
+> gda-owned game, headless by default, same harness and live surface (ADR-0019 / 0020), only with a
+> chosen entry scene. The selector-less default is unchanged. Realized by the run-a-scene slice
+> (#278); the surface-inclusion rationale (why `run` is in scope at all) is recorded in ADR-0025.
 
 ## Decision
 
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -78,6 +78,19 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > the #223 time-windowed base — a 1-frame window for a single shot, an N-frame window
 > for a sequence — so the one-shot RPC contract holds.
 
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Review skill pass

Fixed point: `origin/main...HEAD`
Commit reviewed: `712a5ed docs(adr): add ADR-0025 surface-inclusion principle + amend ADR-0017 for scene selection`

## Standards

Required: ADR-0025 accepts "run a project / a specific scene" as in scope and names #278 as the first slice (`docs/adr/0025-surface-inclusion-principle.md:33`, `:59`), but the command-surface roadmap is not updated with the corresponding `daemon start --scene` territory or an explicit deferral. That leaves a sync gap with the documented project standard that design decisions and issue/docs references stay consistent, complete, and synced, and with ADR-0005's rule that command territory is tracked/refined in `docs/command-catalog.md`. Please either add the placement/provenance to the catalog, or explicitly state why #278's accepted surface is intentionally not catalogued until the implementation PR.

Judgement call: ADR-0025 says it does not change naming/grouping (`docs/adr/0025-surface-inclusion-principle.md:63`), but blessing `run` without pointing to its command group leaves future placement ambiguous against ADR-0005 / ADR-0019's domain-object grouping rules. This becomes less risky if the catalog or ADR text makes clear that this PR only accepts the inclusion principle, while `gda daemon start --scene` is the #278 mechanism and any `scene play` / `game run` wrapper remains a separate follow-up.

## Spec

Required: ADR-0017 records an optional scene selector, but never names the public surface `gda daemon start --scene <path|UID>` or explains how this start-time option applies to the lazy session launch. The spec says: "Surface it as **`gda daemon start --scene <path|UID>`** (the mechanism)." Please pin that exact surface in the amendment so the follow-up implementation cannot drift to a different entry point.

Required: ADR-0017 does not pin the invalid-selector behavior. The spec says: "a non-existent scene yields a typed `GdaError`, not a silent fallback." The amendment should explicitly say a bad scene selector must fail with a typed error and must not silently run `main_scene`.

Required: The trust-boundary rationale is only implicit. The spec says: "Trust is unchanged: running a scene executes `_ready`/game code". Since ADR-0009 is part of the rationale, please state that selected-scene launch stays inside the existing trusted-project / project-code execution boundary.

Scope risk: ADR-0025's worked classification names several future surfaces (`.cs`, asset import, `--script`, `--check-only`, `--import`, remote debug/LSP/DAP, codegen, `--build-solutions`, `--convert-3to4`). The principle itself was requested, but these specific future classifications are broader than #278 needs. If kept, mark them clearly as examples/non-commitments so they do not become accidental roadmap decisions.

Summary: Standards: 1 required finding + 1 judgement call. Spec: 3 required findings + 1 scope-risk note. Worst issue: ADR-0017 does not yet pin the requested `gda daemon start --scene <path|UID>` contract that #278 is supposed to implement.

qinhaihong-red:

### Findings

Required: `docs/adr/0017-gda-daemon-live-execution-mechanism.md:81` records a generic optional scene selector but does not pin the public contract from #278: `gda daemon start --scene <path|UID>`. For an ADR that unblocks the implementation slice, this is the load-bearing API decision. Please name the exact CLI surface and state that it is a start-time daemon option whose value is used when the lazy engine session is launched.

Required: `docs/adr/0017-gda-daemon-live-execution-mechanism.md:81` also needs the failure semantics from #278: a missing/invalid scene selector must produce a typed `GdaError` and must not silently fall back to `main_scene`. Without that, the follow-up implementation can pass the positive path while violating the agent-facing error contract.

Required: The selected-scene amendment should explicitly preserve the existing trust boundary. Running a chosen scene executes the chosen game's code just like running `main_scene`; the issue calls this out, and ADR-0009 / ADR-0018 are the existing authority. Right now the amendment says the editor context remains out of scope, but it does not say that selected-scene launch stays inside the trusted-project / project-code execution surface.

Consider: ADR-0025's worked classification (`docs/adr/0025-surface-inclusion-principle.md:31`) reaches beyond #278 into future calls about `.cs`, asset import, `--script`, `--check-only`, `--import`, remote debugging, codegen, conversion, and build flags. The inclusion principle is useful, but those examples should be framed as illustrative/non-binding unless this PR intends to make each of those surface decisions now.

Consider: If ADR-0025 keeps "run a project / a specific scene" as accepted command territory, update `docs/command-catalog.md` with the `daemon start --scene` provenance or an explicit deferral. Otherwise the ADR and the command-surface roadmap will start out of sync.

### Axis notes

- Correctness: Request changes. The direction is sound, and local Godot 4.6.3 help confirms `--scene <path>` accepts a path or UID, but the ADR misses key #278 contract details.
- Readability: The new ADR is readable and concise; the only clarity issue is over-broad future classification.
- Architecture: The "agent value + structured-operation fit" criterion fits the existing `--json` / `--schema` / `GdaError` design, but the command placement and failure semantics need to be anchored to prevent drift.
- Security: No new security issue in a docs-only PR, but the trust-boundary wording should be explicit because selected-scene launch runs project code.
- Performance: No performance concern.
- Verification: I did not run the test suite because this is docs-only. I verified the Godot CLI premise against Context7 Godot 4.6 docs and the local Godot 4.6.3 `--headless -h` output.

### Verdict

Request changes before merge.

qinhaihong-red:
## Review fixes — all findings addressed (`4e8a3d7`)

Both reviews were on point; accepted and fixed every required finding (docs-only).

### ADR-0017 amendment
1. **Pinned the public surface** — now names `gda daemon start --scene <path|UID>` as a **start-time daemon option** whose value is applied when the **lazy** session launches (per the #7 note), accepting a scene **path or UID**. The `scene play` / `game run` wrapper is called out as a separate follow-up.
2. **Failure semantics pinned** — a missing/non-existent selector **must** surface a typed `GdaError` (a `LIVE` classifier-source code minted by the slice, per decision 6) and **must not** silently fall back to `main_scene`.
3. **Trust boundary explicit** — selected-scene launch runs the chosen scene's `_ready`/game code: the **same** trusted-project / project-code execution surface as `main_scene` and mutating ops (ADR-0009 / 0018), not a new boundary.

### ADR-0025
4. **Worked classification reframed as illustrative / non-binding** — only run-a-scene is *committed* (via #278); export is shipped; every other in/out example is explicitly a non-binding illustration, each its own future decision under the criterion (no accidental roadmap).
5. **Catalog sync (took the explicit-deferral option)** — added a consequence: the concrete command placement and any `command-catalog.md` entry land with the implementation slice **#278**, not this principle ADR (the catalog is a non-binding feature map; committed status lives in the issue tracker / milestone). Also pinned the run entry point (`daemon start --scene`) + wrapper-as-follow-up in the naming/grouping consequence.

Docs-only; no code/tests touched. Ready for re-review.

Issue #282, opened 2026-06-24
feat(logger): `gda_log()` opt-in protocol for rich structured records


qinhaihong-red:
## What to build

Layer the **opt-in rich protocol** on top of the passive `logger` channel (#281, read via `gda logger tail`): let the running game emit fully structured, field-carrying log records.

The gda harness exposes `gda_log(level, message, fields = {})`, which emits **one sentinel-delimited JSON line** into the engine log using a marker **distinct** from ADR-0002's result sentinel — `<<<GDA:LOG>>>{json}` — so a log line can never be mistaken for an op result. The daemon's log parser recognises these lines and turns them into `LogRecord`s carrying the supplied `level` and `fields`. Composes with the passive layer: a `<<<GDA:LOG>>>` line → rich record; engine error/warning → typed record; else → plain `info`.

Anchored by ADR-0026.

## Acceptance criteria

- [ ] The harness exposes `gda_log(level, message, fields={})` callable from project code in a live session.
- [ ] Each call emits exactly one `<<<GDA:LOG>>>{json}` line into the session log; the JSON is single-line (newlines escaped).
- [ ] `gda logger tail --json` returns those as `LogRecord`s with the supplied `level`, `message`, and `fields`.
- [ ] The `<<<GDA:LOG>>>` marker is distinct from `<<<GDA:RESULT>>>`: a plain print of result-shaped text is never misparsed as a log record, and a log line is never misparsed as an op result.
- [ ] e2e: a game calling `gda_log(...)` + `print(...)` + `push_error(...)` yields rich / info / error records respectively.

## Blocked by

- #281 (passive `logger` channel — `gda logger tail`)
- ADR-0026 (PR #280)

qinhaihong-red:
Per the ADR-0026 review (PR #280): the read command is **`gda logger tail`** (ADR-0005 `<group> <command>` shape); `gda_log()` records are retrieved via `gda logger tail --json`.

PR #285, opened 2026-06-24
feat(daemon): run a chosen scene via daemon start --scene

### LINKED ISSUE #278: feat(daemon): run a specific scene in a live session via `daemon start --scene`

qinhaihong-red:
## What & why

Threads an optional **`gda daemon start --scene <path|UID>`** selector through the daemon into the live `Engine session` launch, so the session boots a chosen scene via Godot's native `--scene` engine option (inserted **before** `--path`). With no `--scene`, behaviour is unchanged (runs `main_scene`). A missing/non-existent scene — `res://` path **or** `uid://` — surfaces a typed `live_scene_not_found`, **never** a silent fallback to `main_scene`.

Realizes the **ADR-0017 amendment (2026-06-24, #278)** (session scene selection as an extension of Decision 2) under **ADR-0025** (surface-inclusion). Trust boundary unchanged (running a scene runs its `_ready` — the same trusted-project surface mutating ops already cross, ADR-0009).

## Acceptance criteria

- [x] `gda daemon start --scene <path|UID>` launches the session on the given scene; engine receives `--scene` before `--path`.
- [x] No `--scene` → still runs `main_scene` (no behaviour change for existing callers).
- [x] Accepts a path **or** a UID; a non-existent scene (either form) → typed `live_scene_not_found` (`LIVE`/classifier code, exit 6), not a silent fallback.
- [x] Unit coverage of launch-argv construction (`--scene` inserted in the right position when set, omitted when not).
- [x] `--schema` gate passes; `--json` / error-envelope shapes unchanged.
- [x] e2e (real Godot 4.6): `daemon start --scene res://B.tscn` → `game tree` reports **B**, `main_scene` unchanged; non-existent `res://` and bad `uid://` → typed error.

## Scene validation — at the launch boundary (post-review rework)

The selector is validated **once, at session launch** (not per-request), by what *actually loaded* — so an alive session is never re-validated against later disk changes (ADR-0020 session-bound). Two engine behaviours, one typed outcome:

- **`res://` / filesystem path** — checked for existence at the launch boundary **before** spawning (Godot fails to launch on a missing `res://` rather than fall-back-and-run, so it must be caught pre-launch) → `live_scene_not_found`.
- **`uid://`** — can't be resolved daemon-side, so the **harness verifies at launch**: it compares the actually-loaded `current_scene.scene_file_path` against the selector (resolving the uid via `ResourceUID`). Godot silently runs `main_scene` for a bad uid; the harness detects the mismatch → the daemon tears the session down → `live_scene_not_found`. **No silent fallback.**
- **Already-running daemon + `--scene`** — `--scene` is a start-time selector; a repeat start with `--scene` on a live daemon returns a typed **`daemon_already_running`** (guides `gda daemon stop` then start), instead of silently ignoring it. Already-running with no `--scene` stays idempotent success.

## Tests

- Fast tier: **929 passed** (`pytest -m "not e2e"`). Schema gate: **96 passed**.
- Real-Godot e2e (`tests/test_e2e_daemon.py`): **11 passed**, incl. chosen-scene runs / `main_scene` unchanged, missing `res://` → typed, bad `uid://` → typed (not silent `main_scene`), already-running `--scene` → typed refusal, and a verified session surviving deletion of the scene file mid-session (no per-request re-validation).

## Notes

- New `LIVE` classifier-source codes: `live_scene_not_found`, `daemon_already_running` (registered in `error_codes.py` + the ADR-0002 table).
- **Merge note:** bumps `HARNESS_VERSION` `0`→`1` (harness gains launch-time scene verification). PR #288 (#282) also bumps it to `1`; the two will be reconciled to one coherent value when the second of them merges.
- Disjoint from the diag/logger CLI surface; touches the daemon-launch stack + harness handshake.

Closes #278

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -150,6 +150,7 @@ operation, and parse codes the CLI assigns).
 | `live_unknown_property` | `live` | `classifier` | `6` | A live game get or set targeted a property the running node does not expose for the requested operation (no such readable storage property for get, or no such settable property for set) (Phase 2, #220). |
 | `live_uncoercible_value` | `live` | `classifier` | `6` | A live game set value cannot be coerced to the running property's declared Godot type (Phase 2, #220). |
 | `live_log_unavailable` | `live` | `classifier` | `6` | A live engine session was launched but its diagnostics log file is missing or unreadable, so `gda diag` cannot read the running game's errors/output (Phase 2, #224). |
+| `live_scene_not_found` | `live` | `classifier` | `6` | A `gda daemon start --scene` selector names a scene that does not exist in the project, so no engine session can be launched on it (it is never silently replaced by the project's main_scene) (Phase 2, #278). |
 | `live_perf_node_not_found` | `live` | `classifier` | `6` | A live perf monitor's node path does not resolve to a node in the running scene tree (Phase 2, #223). |
 | `live_perf_property_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a property the running node does not expose for reading (Phase 2, #223). |
 | `live_perf_signal_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a signal the running node does not declare (Phase 2, #223). |
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -146,11 +146,12 @@ operation, and parse codes the CLI assigns).
 | `engine_disconnected` | `live` | `classifier` | `6` | The engine session disconnected before the live operation returned — the game crashed or the harness connection dropped (Phase 2). |
 | `live_timeout` | `live` | `classifier` | `6` | A live operation did not return from the engine session before the daemon's timeout (Phase 2). |
 | `daemon_running` | `live` | `classifier` | `6` | A daemon-lifecycle command was refused because a gda-daemon is running for the project; stop it first with `gda daemon stop` (Phase 2, #225). |
+| `daemon_already_running` | `live` | `classifier` | `6` | A `gda daemon start --scene` was refused because a gda-daemon is already running for the project; `--scene` only takes effect at start, so stop it with `gda daemon stop` then start again with `--scene` (Phase 2, #278). |
 | `live_node_not_found` | `live` | `classifier` | `6` | A live game operation's node path does not resolve to a node in the running scene tree (Phase 2, #220). |
 | `live_unknown_property` | `live` | `classifier` | `6` | A live game get or set targeted a property the running node does not expose for the requested operation (no such readable storage property for get, or no such settable property for set) (Phase 2, #220). |
 | `live_uncoercible_value` | `live` | `classifier` | `6` | A live game set value cannot be coerced to the running property's declared Godot type (Phase 2, #220). |
 | `live_log_unavailable` | `live` | `classifier` | `6` | A live engine session was launched but its diagnostics log file is missing or unreadable, so `gda diag` cannot read the running game's errors/output (Phase 2, #224). |
-| `live_scene_not_found` | `live` | `classifier` | `6` | A `gda daemon start --scene` selector names a scene that does not exist in the project, so no engine session can be launched on it (it is never silently replaced by the project's main_scene) (Phase 2, #278). |
+| `live_scene_not_found` | `live` | `classifier` | `6` | A `gda daemon start --scene` selector did not load: the launched session ran a different scene (Godot silently falls back to main_scene for a missing/invalid path or UID), verified by the harness at launch — gda never falls back (Phase 2, #278). |
 | `live_perf_node_not_found` | `live` | `classifier` | `6` | A live perf monitor's node path does not resolve to a node in the running scene tree (Phase 2, #223). |
 | `live_perf_property_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a property the running node does not expose for reading (Phase 2, #223). |
 | `live_perf_signal_not_found` | `live` | `classifier` | `6` | A live perf monitor targeted a signal the running node does not declare (Phase 2, #223). |
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -489,7 +489,10 @@ def _make_live_runner(binary: Optional[Path], project: Optional[Path]) -> GodotR
 
 def _daemon_start_recipe(params, *, project, godot):
     return run_daemon_start_operation(
-        resolve_project_dir(project), godot, windowed=params.windowed
+        resolve_project_dir(project),
+        godot,
+        windowed=params.windowed,
+        scene=params.scene,
     )
 
 
@@ -1385,6 +1388,16 @@ def daemon_start(
             "headless host (#222)."
         ),
     ),
+    scene: Optional[str] = typer.Option(
+        None,
+        "--scene",
+        help=(
+            "Boot the engine session on this scene (a res:// path or uid:// value) "
+            "instead of the project's main_scene; default runs main_scene. A "
+            "non-existent scene is a typed `live_scene_not_found` error, never a "
+            "silent fall back (#278)."
+        ),
+    ),
     json_output: bool = json_option(),
     schema: bool = DAEMON_START_COMMAND.schema_option(),
     params_json: Optional[str] = params_json_option(),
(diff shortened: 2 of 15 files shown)


qinhaihong-red:

Fixed point: `origin/main...HEAD` (`976845a`, `70d95d5`).
Spec source: #278.

## Standards

1. Required: `src/gda/daemon/server.py` validates the `--scene` selector before every project live op, even when the daemon already holds an alive engine session.

This ties an existing runtime session back to current disk state. I reproduced this with the PR branch: start `daemon --scene res://B.tscn`, run `game tree` successfully (`root.name == "B"`), delete `B.tscn`, then run `game tree` again. The second call returns `live_scene_not_found` instead of serving the still-alive session.

That conflicts with ADR-0020's session-bound model: live consistency is about the engine session's runtime state, while disk/runtime coherence is out of scope and on-disk edits are picked up by relaunching a session. It also conflicts with ADR-0017's statement that externally edited `.tscn` / `.gd` files are not reloaded by an already-running game.

2. Required: `uid://` selectors are exempt from daemon-side validation, and a bad UID silently falls back to `main_scene`.

ADR-0017 says `--scene <path|UID>` accepts paths or UIDs, and missing/non-existent selectors must surface a typed LIVE `GdaError`, never silently fall back to `main_scene`. ADR-0002 registers `live_scene_not_found` for this failure. I reproduced `daemon start --scene uid://does-not-exist` on the PR branch: `game tree` returns success with `root.name == "Main"` instead of a typed error.

Other checked standards look aligned: the field is model-backed, CLI dispatch still goes through the command descriptor/recipe path, and the Python registry plus ADR-0002 table were updated.

## Spec

1. Required: `--scene` is ignored when a daemon is already running.

Spec line: "`gda daemon start --scene <path|UID>` launches the session on the given scene". In `run_daemon_start_operation`, the existing-daemon branch returns `already_running=True` before forwarding `scene` into the daemon. I reproduced: start the daemon without `--scene`, then run `daemon start --scene res://B.tscn`; it returns `already_running: true`, and the following `game tree` still reports `Main`.

2. Required: missing UID handling does not meet the typed-error/no-fallback acceptance criterion.

Spec line: "Accepts either a scene path or a UID (per Godot `--scene`); a non-existent scene yields a typed `GdaError`, not a silent fallback." As above, a bad `uid://` currently succeeds by running `main_scene`, so the UID half of the selector contract is not satisfied.

No obvious scope creep found.

Summary: Standards found 2 required issues; Spec found 2 required issues. Worst issue: an invalid `uid://` selector silently runs `main_scene`, exactly the fallback the spec and ADR prohibit.

qinhaihong-red:

Verdict: Request changes.

### Findings

Required: Invalid `uid://` selectors silently run `main_scene` instead of failing with a typed error.

`src/gda/daemon/server.py:195-197` skips validation for every `uid://` selector and relies on Godot to handle it. In practice, on this PR branch, `gda daemon start --scene uid://does-not-exist` followed by `gda game tree` succeeds and reports `root.name == "Main"`. That is the dangerous failure mode for an agent: it believes it is observing the requested scene but is actually observing the default `main_scene`. The tests only prove that a UID is passed through in argv/fake-server paths; they do not prove bad UID behavior against real Godot.

Required: `daemon start --scene ...` is a no-op when the daemon is already running, with no conflict or remediation.

`src/gda/daemon_ops.py:183-206` returns the existing daemon before the new `scene` value reaches the daemon process. I reproduced: start the daemon without `--scene`, then run `daemon start --scene res://B.tscn`; the second start returns `already_running: true`, and the next `game tree` still reports `Main`. If `scene` is a start-time selector fixed for the daemon/session lifetime, a mismatched repeat start should not look successful without making clear that the requested scene is not active.

Required: Scene existence validation runs before every project live op, so an alive session can be broken by later disk changes.

`src/gda/daemon/server.py:131-134` calls `_scene_not_found_reply()` before `_ensure_session()` can reuse the existing session. I reproduced: start with `--scene res://B.tscn`, run `game tree` successfully (`B`), delete `B.tscn`, then run `game tree` again. The second call returns `live_scene_not_found` even though the engine session was already alive. This puts disk-state validation on the request path instead of the session launch/relaunch boundary, which makes the live layer less session-bound and can fail unrelated reads after the scene is already running.

### Coverage gaps

The added tests cover the happy path, missing `res://` path, argv ordering, schema output, and registry updates. They should also cover:

- bad `uid://` with real Godot or an equivalent integration seam that proves no fallback to `main_scene`;
- repeated `daemon start --scene` when a daemon already exists, especially mismatched scene selectors;
- serving an existing session after the selected scene file is deleted or changed on disk.

### Other axes

Readability is generally clear, and the additive model/CLI plumbing follows the existing descriptor/recipe pattern. I did not find a separate security or performance issue beyond the correctness risks above.

### Verification I ran

- `pytest -q` on the targeted daemon scene/unit tests and error registry: 25 passed.
- Focused schema tests: 3 passed.
- Real Godot e2e for the two new PR tests: 2 passed.
- Manual real-Godot probes for bad UID, already-running daemon, and post-launch scene deletion; these produced the failures described above.

qinhaihong-red:
## Review addressed — `f5966e1` + `b5bbbea`

All three Required findings were real and shared one root cause: scene validation was in the wrong layer (per-request disk pre-check, with `uid://` exempted). Reworked to **verify the scene at the launch boundary**, by what *actually loaded*.

### ✅ Finding 1 — per-request validation broke an alive session
Validation no longer runs in `_handle` per op. It runs **once at launch** inside `_ensure_session`; a verified session is cached and reused without re-checking. Deleting the scene file after launch no longer breaks live ops.
*(e2e: `test_scene_verified_once_at_launch_survives_deleting_the_file`.)*

### ✅ Finding 2 — bad `uid://` silently ran `main_scene`
The harness now verifies, at launch, that the **actually-loaded** `current_scene.scene_file_path` matches the requested selector (resolving `uid://` via `ResourceUID`). Godot's silent fallback to `main_scene` is detected → `SceneMismatch` → typed `live_scene_not_found`, and the half-alive session is torn down. A missing `res://` path is a *different* engine behaviour (Godot fails to launch rather than fall-back-and-run, so the harness never connects) — handled by a `res://` existence check **also at the launch boundary** (not per-request). Two engine behaviours, one boundary, one typed error.
*(e2e: `test_daemon_start_nonexistent_uid_is_typed_not_silent_main_scene`, `test_daemon_start_nonexistent_scene_is_typed_live_scene_not_found`.)*

### ✅ Finding 3 — `daemon start --scene` was a silent no-op on a running daemon
`run_daemon_start_operation` now returns a typed **`daemon_already_running`** (new `LIVE` classifier-source code, registered in `error_codes.py` + the ADR-0002 table) when a daemon is already up **and** `--scene` is given — message guides `gda daemon stop` then start. Already-running with no `--scene` stays idempotent success.
*(e2e: `test_daemon_start_scene_against_a_running_daemon_is_a_typed_refusal`.)*

### Mechanism
- `launch_session` threads the selector into the harness arg tail and reads a post-token verification frame `{scene_ok, current}`; returns the session, raises `SceneMismatch` (uid/loaded mismatch), or `None` (generic launch failure → `engine_session_not_running`, unchanged).
- Harness waits for `current_scene != null` (bounded) before serving any op, then reports the match.
- `_ensure_session` rejects a missing `res://` selector pre-launch (`SceneMismatch`, never spawning the engine).

### Validation
- `uv run --frozen pytest -m "not e2e"` → **929 passed**; schema gate → **96 passed**.
- Real-Godot e2e (`tests/test_e2e_daemon.py`, run locally) → **11 passed**, incl. all four scenarios above + your manual probes (bad uid, already-running, post-deletion survival).

### Merge note
`HARNESS_VERSION` bumped `0`→`1` (harness body changed). PR #288 (#282) also bumps it to `1`; I'll reconcile to one coherent value when merging the second of the two.

PR #297, opened 2026-06-25
feat: a shipped build never carries the gda harness

### LINKED ISSUE #298: Shipped builds must never carry the gda harness (export strip + inert-in-template gate)
> **Status: implemented — tracking/review.** Delivered in #297. This issue is the
> tracking + review record; acceptance criteria are checked against that PR.

## What was built

A shipped build must carry **zero** daemon-related files/config — as if the live
mechanism never existed — and this **must not depend on the developer remembering
`gda daemon uninstall`**. `gda daemon start` installs the harness as a `project.godot`
`[autoload]` + `res://addons/gda_harness/`, and `gda export run` previously had no
harness awareness, so a forgotten harness shipped verbatim.

## Why the obvious fixes don't work (engine research against `../godot`)

- **An export cannot strip an autoload** — `project.binary` is serialized whole from
  `ProjectSettings` *after* every `EditorExportPlugin` hook, with no hook to alter it
  (`editor/export/editor_export_platform.cpp` `save_custom`).
- **`override.cfg` re-leaks** — it is loaded via `set()` into the same `props` map that
  `save_custom` serializes, so any export by a binary that loads it (the editor binary
  `gda export run` uses) writes the autoload back into `project.binary`.
- **`exclude_filter` removes only files**, not the autoload entry; a file-excluded
  dangling autoload then logs `ERR_CONTINUE` and is skipped at startup (`main/main.cpp`)
  — startup error spam in a shipped game, not a hard crash (corrects ADR-0018's wording).
- **feature-tagged autoloads are unsupported** (`autoload/X.editor` registers an autoload
  literally named `X.editor`).

⇒ The only reliable guarantee is that the autoload is **gone before the export reads the
project**.

## Delivered mechanisms (two, complementary)

1. **Transactional strip in `gda export run`** — paired-uninstall the harness (autoload
   entry + files, crash-safe ordering) before the native export and restore it in a
   `finally`. The gda export path is **physically clean and forget-proof**; the dev
   project is untouched; a mid-export crash leaves it harness-*absent* (the safe
   direction).
2. **Runtime build-type gate in the harness** — `_ready()` returns immediately under
   `OS.has_feature(\"template\")` (true only in an exported build, false on the editor
   binary every session runs on). Defense-in-depth so a harness reaching a build via a
   **non-gda** route (editor GUI, raw `godot --export`) is **provably inert**.

**Coverage decision:** `gda export run` = physically clean · every other route = provably
inert. Extending *physical* cleanliness to non-gda routes would need an ephemeral install
that reopens ADR-0018's \"no per-launch mutation\" decision; **deferred** (open a follow-up
only if GUI/raw-CLI exports later need physical, not just inert, cleanliness).

## Acceptance criteria

- [x] `gda export run` produces an artifact with **no** `gda_harness.gd` and no
      `GdaHarness` autoload, with the harness still installed afterward (dev project
      untouched), needing no manual `gda daemon uninstall`.
- [x] The harness stays inert in an exported build regardless of launch args
      (`OS.has_feature(\"template\")` gate).
- [x] Unit tests: strip-during-export, restore-after, restore-on-exception,
      no-op-when-absent.
- [x] e2e: exported archive omits the harness and the project is restored; the
      \"exported pck runs inert\" e2e packs via raw `godot --export-pack` (since
      `gda export run` strips) and asserts the harness is genuinely in the pack so
      \"inert\" stays meaningful.
- [x] Full suite green (970 unit + 287 e2e; 1 e2e skipped — export templates absent).
- [x] ADR-0028 added (export-cleanliness decision + rejected alternatives); ADR-0018
      shrunk to a pointer keeping the point-3 `ERR_CONTINUE` correction; README clarified.

## PR

- #297

## Blocked by

- None — implemented.

qinhaihong-red:
## Problem

A shipped build must carry **zero** daemon-related files/config — as if the live mechanism never existed — and this **cannot depend on the developer remembering `gda daemon uninstall`**. Today `gda export run` has no harness awareness, so a forgotten harness (installed by `gda daemon start` into `project.godot`'s `[autoload]` + `res://addons/gda_harness/`) ships verbatim.

## Why the obvious fixes don't work (engine research against `../godot`)

- **An export cannot strip an autoload.** `project.binary` is serialized whole from `ProjectSettings` *after* every `EditorExportPlugin` hook, with no hook to alter it (`editor/export/editor_export_platform.cpp` `save_custom`).
- **`exclude_filter` removes only files**, not the autoload entry; a file-excluded dangling autoload then logs `ERR_CONTINUE` spam at startup (`main/main.cpp` autoload loop) — note: logged + skipped, **not** the hard crash ADR-0018 originally claimed.
- **`override.cfg` re-leaks**: it's loaded via `set()` into the same `props` map that `save_custom` serializes, so any export by a binary that loads it (the editor binary `gda export run` uses) writes the autoload right back into `project.binary`.
- **feature-tagged autoloads are unsupported** (`autoload/X.editor` registers an autoload literally named `X.editor`).

⇒ The only reliable guarantee is that the autoload is **gone before the export reads the project**.

## Approach (two complementary mechanisms)

1. **Transactional strip in `gda export run`** (`src/gda/export_run.py`): paired-uninstall the harness (autoload entry + files, crash-safe ordering) before the native export and restore it in a `finally`. The gda export path is **physically clean and forget-proof**; the dev project is left untouched; a mid-export crash leaves it harness-*absent* (the safe direction). Reuses the existing `uninstall_harness`/`install_harness`.
2. **Runtime build-type gate** (`src/gda/harness/gda_harness.gd`): `_ready()` returns immediately under `OS.has_feature("template")` — true only in an exported build, false on the editor binary every session runs on. Defense-in-depth so a harness reaching a build via a **non-gda** route (editor GUI, raw `godot --export`) is **provably inert**. `HARNESS_VERSION` 2→3 so installed copies re-materialize.

**Coverage:** gda export = physically clean · every other route = provably inert. (Physical cleanliness on non-gda routes would need an ephemeral install reopening ADR-0018's "no per-launch mutation"; deferred.)

## Tests

- **Unit** (`tests/test_export_run_operation.py`): strip-during-export, restore-after, restore-on-exception, no-op-when-absent.
- **e2e** (`tests/test_e2e_export_run.py`): install harness → `export run --mode pack --output …zip` → archive omits `gda_harness.gd` and the project is restored.
- **Reconciled** (`tests/test_e2e_harness_install.py`): the "exported pck runs inert" boot now packs via a **raw `godot --export-pack`** (since `gda export run` strips), asserting the harness is genuinely in the pack so "inert" stays meaningful.
- Full suite green: **970 unit + 287 e2e** (1 e2e skipped — export templates absent).

## Docs

New **ADR-0028** records the export-cleanliness decision and the rejected alternatives (override.cfg re-leak, file-only exclude_filter, feature-tagged autoload, ephemeral install); ADR-0018 shrinks to a pointer and keeps the corrected point-3 (`ERR_CONTINUE`, not a crash) note. README: the harness is dev-only (export strips it / it self-disables in shipped builds); `daemon uninstall` clarified.

---
Tracking/review issue: Closes #298

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ one you just built, then start the daemon (macOS/Linux, Godot 4.6+):
 
 ```bash
 gda project set application/run/main_scene --value res://scenes/main.tscn --json
-gda daemon start             # start the daemon for $GDA_PROJECT (installs the harness)
+gda daemon start             # start the daemon for $GDA_PROJECT (installs the in-game harness)
 gda game tree --json         # the runtime scene tree, after _ready
 gda perf monitors --json     # live engine counters: fps, memory, node count
 gda daemon stop
@@ -438,7 +438,7 @@ flags — `gda --help` is the authoritative list of what is installed.
 
 | Command | What it does |
 | ------- | ------------ |
-| `daemon start` | Start the per-project daemon and install the harness; the engine session launches on the first live op (`--windowed` for `screen` capture). |
+| `daemon start` | Start the per-project daemon and install the in-game harness; the engine session launches on the first live op (`--windowed` for `screen` capture). |
 | `daemon stop` | Stop the project's daemon and any running engine session. |
 | `daemon status` | Report the daemon's state (running, windowed mode, session). |
 | `daemon uninstall` | Remove the in-game `gda` harness (autoload entry + files) from the project — an explicit dev-tooling teardown; `gda export run` already strips it from exported artifacts automatically. |
--- a/README.md
+++ b/README.md
@@ -302,6 +302,10 @@ command — register via the JSON above or the Settings → MCP UI.
   in-game harness, and brokers requests over a Unix domain socket (runtime tree, input,
   screenshots, performance, diagnostics).
 
+The harness is **dev-only**: `gda export run` strips it from the exported artifact, and it
+self-disables in any exported build — so a shipped game carries and runs nothing
+daemon-related.
+
 **Platform & version support:**
 
 | Mode | Godot | Platforms |
@@ -437,7 +441,7 @@ flags — `gda --help` is the authoritative list of what is installed.
 | `daemon start` | Start the per-project daemon and install the harness; the engine session launches on the first live op (`--windowed` for `screen` capture). |
 | `daemon stop` | Stop the project's daemon and any running engine session. |
 | `daemon status` | Report the daemon's state (running, windowed mode, session). |
-| `daemon uninstall` | Remove the in-game `gda` harness autoload from the project. |
+| `daemon uninstall` | Remove the in-game `gda` harness (autoload entry + files) from the project — an explicit dev-tooling teardown; `gda export run` already strips it from exported artifacts automatically. |
 
 **`game`** — the running game's runtime scene graph
 
--- a/README.md
+++ b/README.md
@@ -302,9 +302,9 @@ command — register via the JSON above or the Settings → MCP UI.
   in-game harness, and brokers requests over a Unix domain socket (runtime tree, input,
   screenshots, performance, diagnostics).
 
-The harness is **dev-only**: `gda export run` strips it from the exported artifact, and it
-self-disables in any exported build — so a shipped game carries and runs nothing
-daemon-related.
+The in-game harness `gda-daemon` injects is **dev-only**: `gda export run` strips it from the
+exported artifact, and it self-disables in any exported build — so a shipped game carries and
+runs nothing daemon-related.
 
 **Platform & version support:**
 
--- a/README.md
+++ b/README.md
@@ -303,8 +303,9 @@ command — register via the JSON above or the Settings → MCP UI.
   screenshots, performance, diagnostics).
 
 The in-game harness `gda-daemon` injects is **dev-only**: `gda export run` strips it from the
-exported artifact, and it self-disables in any exported build — so a shipped game carries and
-runs nothing daemon-related.
+artifact entirely, and built any other way (editor GUI, raw `godot --export`) it still
+self-disables in the exported game — so a shipped game never *runs* anything daemon-related
+(and via `gda export run`, doesn't even carry it).
 
 **Platform & version support:**
 
--- a/README.md
+++ b/README.md
@@ -296,8 +296,8 @@ command — register via the JSON above or the Settings → MCP UI.
 | **`gda-mcp`**    | An MCP server exposing the same operations as tools, from `--schema`. |
 | **`gda-daemon`** | A per-project process supervising a running game for live operations. |
 
-- **Headless operations** need no running engine — `gda` runs them one-shot with nothing
-  to install (create a scene, edit a script, export, analyze).
+- **Headless operations** run one-shot — no daemon, nothing to install (create a scene, edit
+  a script, export, analyze).
 - **Live operations** require a running game — `gda-daemon` launches it, injects an inert
   in-game harness, and brokers requests over a Unix domain socket (runtime tree, input,
   screenshots, performance, diagnostics).
@@ -572,7 +572,7 @@ or agent can branch on the failure **category without parsing the JSON error**:
 | `3`       | `version`     | The detected Godot version is below the supported minimum.            |
 | `4`       | `operation`   | The engine ran but the operation failed — a registered operation error, an engine crash, or an unstructured non-zero exit. |
 | `5`       | `parse`       | The process claimed success but violated the structured-output contract. |
-| `6`       | `live`        | A live operation failed — e.g. no running daemon/session, or a live timeout (Phase-2 live). |
+| `6`       | `live`        | A live operation failed — e.g. no running daemon/session, or a live timeout. |
 
 These values are the public ABI; their authoritative source is
 [`src/gda/exit_codes.py`](src/gda/exit_codes.py). The `{"error": {category, code, …}}`
--- a/docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md
+++ b/docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md
@@ -66,26 +66,17 @@ so two instances can touch the project at once.
 > **asymmetric** — implicit install on `start`, explicit `uninstall` — not a symmetric
 > install/uninstall pair. Decision 1's wording is preserved as the point-in-time record.
 
-> **Outcome (2026-06-25) — a shipped build never carries the harness: `gda export run`
-> strips it, and the harness self-disables in any exported build.** The bottom line is
-> that a shipped build must carry zero daemon-related files/config and must not depend on
-> the developer remembering `gda daemon uninstall`. Engine research settled the mechanism:
-> an export **cannot** remove a `project.godot` autoload after the fact — `project.binary`
-> is serialized whole from `ProjectSettings` *after* every `EditorExportPlugin` hook, with
-> no hook to alter it (`editor/export/editor_export_platform.cpp` `save_custom`), and the
-> entry re-leaks through `override.cfg` (loaded via `set()` into the same `props` map that
-> `save_custom` writes), so the only reliable guarantee is that the autoload is gone
-> **before** the export reads the project. So **`gda export run` paired-strips the harness
-> transactionally** (uninstall → native export → reinstall in a `finally`): the gda export
-> path is harness-free and forget-proof, the dev project is left untouched, and a mid-export
-> crash leaves it harness-*absent* (safe). As **defense in depth** for export routes that
-> bypass gda (editor GUI, raw `godot --export`), the harness `_ready()` returns immediately
-> under `OS.has_feature("template")` — true only in an exported build, false on the editor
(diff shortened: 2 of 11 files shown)


qinhaihong-red:
## Review pass 1/2: `review` skill — Standards + Spec

Base: `main`  
Diff reviewed: `git diff main...HEAD`

## Standards

Hard violations:

1. `src/gda/export_run.py:157-162` conflicts with the project-state guarantee documented in ADR-0028. ADR-0028 Decision 1 says the dev project is left "byte-identical", and ADR-0018's new outcome says the dev project is "untouched". This implementation restores by calling the normal `install_harness(project)`, not by restoring the exact previous files/config. `install_harness` can rewrite stale harness bytes/version, repoint or insert the canonical autoload, and materialize a fresh harness body. A stale installed harness, a stray harness file without an autoload, or a noncanonical prior autoload is therefore not restored byte-for-byte. The related tests only assert "file exists + autoload name exists", so they miss this documented standard.

2. `README.md:305-307` overstates ADR-0028's coverage. ADR-0028 Decision 3 says physical cleanliness is guaranteed for `gda export run`; other export routes may still physically carry the harness, but it is inert. The README says "a shipped game carries and runs nothing daemon-related", which reads as all shipped builds, including editor GUI/raw `godot --export` routes. It should distinguish "`gda export run` artifacts are physically clean" from "other export routes may carry the harness but it self-disables".

## Spec

1. Partial: inert exported-build coverage does not exercise the required gate. The spec says `_ready()` returns immediately under `OS.has_feature("template")`, true only in an exported build, and that this works regardless of launch args. The implementation adds the gate in `src/gda/harness/gda_harness.gd:92-100`, but `tests/test_e2e_harness_install.py:130-131` runs the `.pck` with the editor/tools binary via `--main-pack` and no `gda-daemon` marker. That proves the old "no marker" inert path, not `template == true`, and not "regardless of launch args".

2. Partial: artifact autoload absence is not directly verified. The spec requires an artifact with no `gda_harness.gd` and no `GdaHarness` autoload, and says the autoload must be gone before export reads the project. The e2e checks only the ZIP listing lacks `gda_harness.gd` (`tests/test_e2e_export_run.py:263-270`), while the unit strip test observes only file absence during export (`tests/test_export_run_operation.py:254-256`). The behavior likely follows from `uninstall_harness`, but the acceptance check for packed `project.binary` / `GdaHarness` is missing.

3. Requirement looks implemented but can be wrong for noncanonical installed state. The spec says `gda export run` should restore the harness in a `finally` and leave the dev project untouched. `src/gda/export_run.py:157-162` restores by calling fresh `install_harness(project)`, not by replaying the exact prior state. I reproduced a concrete edge case: if the project has only a leftover `addons/gda_harness/gda_harness.gd` file and no `[autoload]` entry before export, `uninstall_harness()` reports `removed=True`, the finally calls `install_harness()`, and the project ends with a newly added `GdaHarness` autoload. That is not "untouched" and violates the no-op-when-absent intent.

Summary: Standards found 2 hard conflicts; Spec found 3 partial/incorrect-coverage issues. Worst issue: `gda export run` does not preserve the exact pre-export dev-project state required by ADR-0028/spec.

qinhaihong-red:
## Review pass 2/2: `code-review-and-quality`

Verdict: request changes. The main `gda export run` happy path looks directionally right, and the focused unit/e2e paths pass locally, but this is a security/build-hygiene change and the current implementation plus verification still leave important gaps.

### Findings

1. **Required — `src/gda/export_run.py:157-162`: the strip/restore transaction can mutate the developer project instead of restoring it.** The code treats any `uninstall_harness(...).removed` result as proof that a full harness should be reinstalled. That is not equivalent to "the harness was installed before export." I reproduced a concrete edge case: a project with only a leftover `addons/gda_harness/gda_harness.gd` file and no `[autoload]` entry exports cleanly, then the `finally` calls `install_harness(project)` and adds a new `GdaHarness` autoload that was not present before. The same restore path can also rewrite stale harness bytes or canonicalize a noncanonical pre-existing autoload. For the stated guarantee ("dev project untouched"/byte-identical), capture and restore the exact pre-export state, or at least make the uninstall result distinguish "autoload was present" from "some file was removed" and cover stale/noncanonical cases in tests.

2. **Required — `tests/test_e2e_harness_install.py:130-131`: the defense-in-depth exported-template gate is not actually verified.** The new harness guard is `if OS.has_feature("template"): return`, but this e2e runs a `.pck` with the editor/tools binary via `--main-pack` and no `gda-daemon` marker. That proves the old "no marker" inert path still works; it does not prove the `template` branch executes, and it does not prove the "regardless of launch args" claim. This should be tested with an actual exported template binary, or with another explicit proof that the `template` branch ran before marker/socket handling. CI currently skips the Godot e2e job, so this cannot be left as an assumed property.

3. **Required — `README.md:305-307`: the public docs overpromise physical absence for non-`gda export run` builds.** ADR-0028 deliberately chooses "`gda export run` = physically clean; every other route = provably inert." The README sentence says a shipped game "carries and runs nothing daemon-related", which reads as all shipped builds. For editor GUI/raw `godot --export` routes, the harness may still be physically present; it should say those builds self-disable the harness rather than carrying nothing.

4. **Required verification gap — `tests/test_e2e_export_run.py:263-270`: the artifact check only proves the script file is absent.** The acceptance criteria also require no `GdaHarness` autoload/config in the artifact. Because the Godot risk here is specifically `project.binary` serializing `ProjectSettings`, the e2e should directly check that packed settings/autoload state does not contain `GdaHarness`, not just that `gda_harness.gd` is absent from the ZIP listing.

5. **Nit — stale comments still say dangling autoloads crash exported games.** Examples: `src/gda/harness/install.py:25`, `src/gda/harness/install.py:252`, and `tests/test_harness_install.py:194-195`. ADR-0028 corrects this to `ERR_CONTINUE` + skip, not a hard crash. Since this PR is explicitly reconciling that wording, the remaining comments should be updated to avoid reintroducing the old mental model.

### Verification I ran

- `env PATH=/private/tmp/godot-agent-pr297-review-20260626/.venv/bin:... UV_CACHE_DIR=/private/tmp/godot-agent-pr297-uv-cache .venv/bin/pytest -m 'not e2e' -q` — `970 passed, 288 deselected`.
- `env PATH=/private/tmp/godot-agent-pr297-review-20260626/.venv/bin:... .venv/bin/pytest tests/test_e2e_export_run.py::test_export_run_pack_omits_installed_harness_and_restores_it tests/test_e2e_harness_install.py::test_exported_pck_with_harness_runs_inert -q` — `2 passed`.
- Focused repro for finding 1: before export, project had a stray harness file and no `GdaHarness` autoload; after `run_export_operation`, the file existed and `project.godot` contained a newly added `GdaHarness="*res://addons/gda_harness/gda_harness.gd"` entry.

No new dependency, performance, or broad architecture concerns beyond the transaction semantics above.

qinhaihong-red:
Thanks — all five findings were valid; addressed in `754a697` (fix) and `b47ea21` (test+docs).

**1 (Required — strip/restore could mutate the dev project).** Confirmed bug against ADR-0028's "byte-identical". Replaced the `install_harness`-based restore with a **byte-exact snapshot/replay**: capture the exact pre-export bytes of `project.godot` + the harness file, strip via the paired `uninstall_harness`, then restore the snapshot in `finally` (writing only when changed). Your repro is now a regression test — `test_export_does_not_add_autoload_for_a_stray_harness_file` (stray file → no autoload synthesized), plus `..._restores_a_stale_harness_body_unchanged...` (not re-materialized) and a byte-identical assertion on the happy path.

**4 (Required verification gap — artifact autoload not checked).** The export e2e now reads the packed **`project.binary`** from the archive and asserts it declares no `GdaHarness` autoload — the specific `ProjectSettings`-serialization risk — not just that `gda_harness.gd` is absent.

**2 (Required — template gate not actually verified).** Right: the pck e2e runs on the editor binary (`template == false`), so it proves the *no-marker* path. The behavioural `template == true` proof needs a real exported **template** binary, which needs installed export templates **and** the Godot e2e job (skipped on PRs) — so I added a **CI-runnable static guarantee** that `if OS.has_feature("template"): return` is the **first statement of `_ready()`** (`test_ready_gates_on_template_feature_as_its_first_statement`). That runs in the default PR gate, so the property is *checked*, not assumed — directly answering the "CI skips e2e" concern. The pck e2e comment is re-scoped to say what it actually proves. (A full template-binary behavioural e2e remains a follow-up gated on a runner with export templates.)

**3 (Required — README overpromises).** Reworded to distinguish the two: `gda export run` strips it from the artifact entirely; built any other way it self-disables — so a shipped game never *runs* anything daemon-related (and via `gda export run`, doesn't even carry it).

**5 (Nit — stale "crash" comments).** Corrected to `ERR_CONTINUE` + skip (error spam, not a hard crash) in `install.py`, `cli.py`, and the test comments. ADR-0018 Decision 3's point-in-time text is intentionally left as-is; its Outcome note carries the correction.

Verification: `973 unit` + the affected e2e (`test_e2e_export_run.py`, `test_e2e_harness_install.py`) green serially against real Godot (1 e2e skipped — export templates absent).

PR #303, opened 2026-06-26
test(e2e): behaviourally prove the harness template gate (#301)

### LINKED ISSUE #301: test(e2e): behavioural proof of the harness `template` gate (needs export templates)
## What to build

PR #297 added `if OS.has_feature("template"): return` as the **first statement** of the harness `_ready()` (ADR-0028), so an exported build self-disables **regardless of launch args**. That gate is currently proven only **statically** — `tests/test_harness_install.py::test_ready_gates_on_template_feature_as_its_first_statement` asserts it is the first body statement — because the behavioural `template == true` branch needs a **real exported template binary**.

Add the behavioural e2e: export a debug (or release) **template** binary, run it headless **with** a `gda-daemon <socket> <token>` marker pointing at a real **listening** Unix socket, and assert the harness **does NOT connect** — i.e. the `template` gate short-circuits `_ready()` *before* the marker/socket handling. This is the only test that exercises the `template` branch end-to-end (the existing exported-pck e2e runs on the editor binary, where `template` is false, so it proves the *no-marker* path, not the gate).

> **Requires a runner with Godot export templates installed** for the engine version under test. The default PR CI both **skips the Godot e2e job** and lacks export templates, so this test must `pytest.skip` (loudly) when templates are absent and only runs where they exist. PR #297 adds a **tripwire** e2e (`test_template_gate_behavioural_proof_is_due_once_templates_exist`) that warns the moment templates become available, so this work is not silently forgotten — implementing this issue should replace that tripwire.

## Acceptance criteria

- [ ] An e2e exports a real template binary and runs it headless with a `gda-daemon` marker + a live listening socket; asserts the harness never connects (the `template` gate fired before marker handling).
- [ ] Skips **loudly** (clear reason) when export templates are not installed; runs automatically where they are.
- [ ] Replaces the PR #297 tripwire test once landed.
- [ ] Documents the runner requirement (export templates) wherever the e2e job is configured.

## Blocked by

- None functionally — but depends on a CI runner / dev env with **export templates** to actually execute. Independent of #297 (which adds the static proof + tripwire).

### LINKED ISSUE #304: bug: gda mis-detects installed export templates (Linux dir case + .0-release patch name)
`gda export get` reports `templates_installed: false` even when the matching export
templates **are** installed, and `gda export run` (release/debug) then refuses the
export with `export_templates_missing`. Two independent defects in how
`src/gda/ops/operations.gd` reconstructs the templates directory path, both masked
until now (dev ran on case-insensitive macOS at a non-`.0` patch; CI had no
templates):

## Facet 1 — wrong user-dir case on Linux

`_export_templates_installed` hardcoded a capital `"Godot"`:

```gdscript
var templates_root := OS.get_data_dir().path_join("Godot").path_join("export_templates")
```

But the engine's per-platform user-dir name is **lowercase `"godot"` on Linux and the
BSDs** — only macOS/Windows capitalize it (C++ `OS::get_godot_dir_name()`; its default
is lowercase, only those two override). On case-sensitive Linux the capital path never
resolves. Fixed via a `_godot_user_dir_name()` helper.

## Facet 2 — patch component included for `.0` releases

`_export_templates_version_dir` built `"%d.%d.%d.%s"` unconditionally → `"4.6.0.stable"`.
But Godot **omits the patch when it is 0** — the per-version dir and the official
`.tpz`'s `version.txt` are `"4.6.stable"` for 4.6.0 (and `"4.6.3.stable"` for 4.6.3),
exactly as `Engine.get_version_info()`'s version string does (`engine.cpp` only appends
`".patch"` when non-zero). So on any `.0` release gda looked for `"4.6.0.stable"` and
missed the real `"4.6.stable"`. Platform-independent. Fixed by omitting the patch when 0.

Confirmed on a Linux CI runner via a diagnostic: installed dir `…/godot/export_templates/4.6.stable/`,
yet `gda export get` reported `templates_installed:false, templates_version:"4.6.0.stable"`.

## Impact

- `export get` → `templates_installed` wrong (Linux always; any platform on a `.0` release).
- `export run` release/debug → fails fast as `export_templates_missing` even when a real
  export would succeed (the structured preflight keys on the bad signal — `src/gda/export_run.py`).

## Fix

Both resolved in **#303**. No behaviour change on macOS/Windows at a non-`.0` patch.

## Acceptance criteria

- [x] Templates dir resolved with the OS-correct user-dir name (lowercase on Linux).
- [x] Version dir omits the patch component for `.0` releases.
- [x] Exercised end-to-end by the export-family e2e on a Linux runner (4.6-stable, a `.0` release) with templates installed — `gda export get` reports `templates_installed:true` and the #301 proof runs.

## Blocked by

None — fix already in #303.

qinhaihong-red:
## What

Implements #301 — the **behavioural** proof of the gda harness's ADR-0028 `template` gate — and fixes the export-template detection bug (#304) that wiring the proof into Linux CI surfaced.

- **New e2e** `test_template_feature_gates_the_harness_only_in_exported_builds`: exports a real **debug template** binary that ships the harness, runs it headless with a `gda-daemon <socket> <token>` marker against a **live listening** Unix socket, and asserts the harness **never connects** — the `template` gate short-circuits `_ready()` before marker/socket handling.
- **Positive control**: the editor binary (template false) runs against the *identical* socket/marker/token rig and DOES connect (token frame verified), isolating `OS.has_feature("template")` as the only variable — so the template binary's silence is the gate firing, not a dead rig. Anti-vacuous: also asserts the harness is genuinely packed into the exported artifact.
- Host-parameterized: **Linux/X11** (CI) and **macOS** (dev). Skips **loudly** where the host is unsupported or templates are absent (the static first-statement proof in `tests/test_harness_install.py` still guards the gate everywhere).
- **Replaces** the PR #297 tripwire `test_template_gate_behavioural_proof_is_due_once_templates_exist`.
- Promotes `templates_installed` to `tests/support.py` (single source of the e2e template-presence policy, now preset-parameterized).
- **CI**: extends the shared `install-godot` action to download+cache the matching `export_templates.tpz` so the scheduled/dispatch `godot-e2e` job (and `release.yml`) actually run the proof. The version dir is named from the `.tpz`'s `version.txt` (authoritative), and `-rs` is added to the e2e run so a silently-skipped test is visible in the log.
- Installing templates **un-skips** `test_export_run_writes_to_configured_export_path` (gave it exportable content) and tightened `test_e2e_export`'s version-format regex — both branches had never run on a Linux/.0 runner before.

## Fixes #304 — gda mis-detected installed export templates

Running the proof on Linux CI surfaced a real, pre-existing gda bug in how `operations.gd` reconstructs the export-templates path — two facets, both masked because dev ran on case-insensitive macOS at a non-`.0` patch and CI had no templates:

1. **Wrong user-dir case on Linux** — hardcoded capital `"Godot"`, but the engine's user-dir is lowercase `"godot"` on Linux/BSD (only macOS/Windows capitalize it). Case-sensitive Linux never resolved it.
2. **Patch included for `.0` releases** — built `"4.6.0.stable"`, but Godot omits the patch when 0 (`"4.6.stable"`), as `Engine.get_version_info()`'s version string does.

Either made `export get` report `templates_installed=false` and `export run` refuse every Linux/`.0` export as `export_templates_missing`. Without this fix the #301 proof would silently **skip** on Linux CI instead of running. No behaviour change on macOS/Windows at a non-`.0` patch.

## Why

PR #297 added the gate but proved it only **statically** (it is the first `_ready()` statement). The `template == true` branch had never been executed end-to-end — this is the only test that does.

## Verification (real engine, no mocks)

- `pytest -m "not e2e"` → **973 passed** (incl. the static first-statement proof).
- **macOS local** (4.6.3 templates installed): full e2e suite **289 passed / 0 failed**. The editor binary connects; the exported **universal** template binary stays silent under the marker.
- **Linux CI** (`workflow_dispatch`, `run-e2e=true`): the `install-godot` template steps succeed and, after the #304 fix, `gda export get` reports `templates_installed:true` — the export-family e2e (incl. the new proof) **runs and passes**: **286 passed, 3 skipped, 0 failed** (the 3 skips are the headless-Linux windowed-capture tests, unrelated).

Closes #301.

### DIFF
--- a/.github/actions/install-godot/action.yml
+++ b/.github/actions/install-godot/action.yml
@@ -46,23 +46,23 @@ runs:
         chmod +x "${GDA_GODOT}"
         "${GDA_GODOT}" --headless --version
 
-    # Export templates for the SAME pinned version. Required only by the #301
+    # Export templates for the SAME pinned engine, required only by the #301
     # template-gate behavioural e2e (it exports a real template binary); every other
-    # e2e needs only the editor binary above. Templates MUST match the engine version
-    # exactly — Godot keys them on its full version string with a `.` separator
-    # (4.6-stable -> 4.6.stable) — so both derive from the one `godot-version` input.
-    # A local dev on a different patch (e.g. 4.6.3.stable) needs that patch's .tpz.
+    # e2e needs only the editor binary above. Godot keys templates on a per-version
+    # directory under its user-data dir — lowercase "godot" on Linux — whose name is
+    # the engine's FULL_CONFIG (e.g. "4.6.0.stable", patch ALWAYS included). That name
+    # is authoritative in the .tpz's templates/version.txt, so the version dir is named
+    # from THAT, not guessed from the release tag (which omits the patch).
     - name: Resolve export-template paths
       shell: bash
       run: |
-        version="$(echo '${{ inputs.godot-version }}' | sed 's/-/./')"
-        echo "GODOT_TEMPLATE_DIR=${HOME}/.local/share/godot/export_templates/${version}" >> "$GITHUB_ENV"
+        echo "GODOT_TEMPLATES_ROOT=${HOME}/.local/share/godot/export_templates" >> "$GITHUB_ENV"
 
     - name: Restore cached export templates
       id: templates-cache
       uses: actions/cache@v4
       with:
-        path: ${{ env.GODOT_TEMPLATE_DIR }}
+        path: ${{ env.GODOT_TEMPLATES_ROOT }}
         key: godot-templates-${{ inputs.godot-version }}-linux
 
     - name: Download export templates
@@ -72,14 +72,16 @@ runs:
         curl --fail --location --show-error \
           --output "${RUNNER_TEMP}/templates.tpz" \
           "https://github.com/godotengine/godot/releases/download/${{ inputs.godot-version }}/Godot_v${{ inputs.godot-version }}_export_templates.tpz"
-        # A .tpz is a zip whose top-level templates/ holds the per-platform files;
-        # Godot expects them flattened into the version dir.
+        # A .tpz is a zip whose top-level templates/ holds the per-platform files plus
+        # version.txt (the engine's FULL_CONFIG); flatten them into <root>/<version>/.
         unzip -q "${RUNNER_TEMP}/templates.tpz" -d "${RUNNER_TEMP}/tpz"
-        mkdir -p "${GODOT_TEMPLATE_DIR}"
-        mv "${RUNNER_TEMP}/tpz/templates/"* "${GODOT_TEMPLATE_DIR}/"
+        version="$(cat "${RUNNER_TEMP}/tpz/templates/version.txt")"
+        dest="${GODOT_TEMPLATES_ROOT}/${version}"
+        mkdir -p "${dest}"
+        mv "${RUNNER_TEMP}/tpz/templates/"* "${dest}/"
 
     - name: Verify export templates
       shell: bash
       run: |
-        # The template `--export-debug` uses for Linux/X11 (#301).
-        test -f "${GODOT_TEMPLATE_DIR}/linux_debug.x86_64"
+        # The Linux/X11 debug template the #301 proof's --export-debug uses.
+        find "${GODOT_TEMPLATES_ROOT}" -name linux_debug.x86_64 | grep -q .
--- a/.github/actions/install-godot/action.yml
+++ b/.github/actions/install-godot/action.yml
@@ -50,13 +50,15 @@ runs:
     # template-gate behavioural e2e (it exports a real template binary); every other
     # e2e needs only the editor binary above. Godot keys templates on a per-version
     # directory under its user-data dir — lowercase "godot" on Linux — whose name is
-    # the engine's FULL_CONFIG (e.g. "4.6.0.stable", patch ALWAYS included). That name
-    # is authoritative in the .tpz's templates/version.txt, so the version dir is named
-    # from THAT, not guessed from the release tag (which omits the patch).
+    # the engine's version-config string (e.g. "4.6.stable" for 4.6.0, "4.6.3.stable"
+    # for 4.6.3 — the patch is OMITTED when 0). That name is authoritative in the
+    # .tpz's templates/version.txt, so the version dir is named from THAT, not guessed
+    # from the release tag. The root honors XDG_DATA_HOME so it matches where the
+    # engine (and gda's export-get) actually look.
     - name: Resolve export-template paths
       shell: bash
       run: |
-        echo "GODOT_TEMPLATES_ROOT=${HOME}/.local/share/godot/export_templates" >> "$GITHUB_ENV"
+        echo "GODOT_TEMPLATES_ROOT=${XDG_DATA_HOME:-$HOME/.local/share}/godot/export_templates" >> "$GITHUB_ENV"
 
     - name: Restore cached export templates
       id: templates-cache
--- a/.github/actions/install-godot/action.yml
+++ b/.github/actions/install-godot/action.yml
@@ -1,7 +1,9 @@
 name: Install Godot
 description: >-
-  Download (and cache) a headless Godot binary and export GDA_GODOT for later
-  steps. Single source of truth for the Godot version shared by CI and Release.
+  Download (and cache) a headless Godot binary AND its matching export templates,
+  and export GDA_GODOT for later steps. Single source of truth for the Godot version
+  shared by CI and Release; the export templates let the e2e job run the #301
+  template-gate behavioural proof, which exports a real template binary.
 
 inputs:
   godot-version:
@@ -43,3 +45,41 @@ runs:
       run: |
         chmod +x "${GDA_GODOT}"
         "${GDA_GODOT}" --headless --version
+
+    # Export templates for the SAME pinned version. Required only by the #301
+    # template-gate behavioural e2e (it exports a real template binary); every other
+    # e2e needs only the editor binary above. Templates MUST match the engine version
+    # exactly — Godot keys them on its full version string with a `.` separator
+    # (4.6-stable -> 4.6.stable) — so both derive from the one `godot-version` input.
+    # A local dev on a different patch (e.g. 4.6.3.stable) needs that patch's .tpz.
+    - name: Resolve export-template paths
+      shell: bash
+      run: |
+        version="$(echo '${{ inputs.godot-version }}' | sed 's/-/./')"
+        echo "GODOT_TEMPLATE_DIR=${HOME}/.local/share/godot/export_templates/${version}" >> "$GITHUB_ENV"
+
+    - name: Restore cached export templates
+      id: templates-cache
+      uses: actions/cache@v4
+      with:
+        path: ${{ env.GODOT_TEMPLATE_DIR }}
+        key: godot-templates-${{ inputs.godot-version }}-linux
+
+    - name: Download export templates
+      if: steps.templates-cache.outputs.cache-hit != 'true'
+      shell: bash
+      run: |
+        curl --fail --location --show-error \
+          --output "${RUNNER_TEMP}/templates.tpz" \
+          "https://github.com/godotengine/godot/releases/download/${{ inputs.godot-version }}/Godot_v${{ inputs.godot-version }}_export_templates.tpz"
+        # A .tpz is a zip whose top-level templates/ holds the per-platform files;
+        # Godot expects them flattened into the version dir.
+        unzip -q "${RUNNER_TEMP}/templates.tpz" -d "${RUNNER_TEMP}/tpz"
+        mkdir -p "${GODOT_TEMPLATE_DIR}"
+        mv "${RUNNER_TEMP}/tpz/templates/"* "${GODOT_TEMPLATE_DIR}/"
+
+    - name: Verify export templates
+      shell: bash
+      run: |
+        # The template `--export-debug` uses for Linux/X11 (#301).
+        test -f "${GODOT_TEMPLATE_DIR}/linux_debug.x86_64"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,10 @@ on:
         default: false
         type: boolean
   schedule:
-    - cron: "24 3 * * 1"
+    # Nightly (03:24 UTC) — the automatic, dependable trigger for the godot-e2e job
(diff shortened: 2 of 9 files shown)


qinhaihong-red:
## Code-review follow-ups (xhigh multi-agent review)

A workflow-backed review (10 finder angles, independent verifier per finding) ran on this branch. Disposition:

**Fixed (commit `a0a8b4f`):**
- **action.yml comment was backwards** — said the version dir "always includes the patch"; corrected (it's OMITTED for `.0`, matching the operations.gd fix in this PR).
- **`-rs` added to `release.yml`** too (visibility parity — a silently-skipped e2e now shows a reason on the release path as well).
- **`_SocketProbe` token race** — `connected` is now set AFTER the token frame is read, removing a read-before-write race in the positive control.
- **single-source / perf** — `_exported_tree_has_harness` now scans only `*.pck` (not the multi-hundred-MB template binary) via the `HARNESS_FILE` constant; macOS `_locate_exe` skips dotfiles; negative-control grace `wait(1)`→`wait(0.5)`; `templates_installed` docstring corrected (verdict is preset-independent).

**Fixed (commit `9ca1e4f`):**
- **[9] restored lost live coverage** — new `test_export_run_without_templates_yields_export_templates_missing` (XDG_DATA_HOME→empty dir, Linux-only) re-exercises the `export_templates_missing` preflight that installing templates into CI had made dead. Runs+passes on Linux CI.
- **[10] pinned the version-dir format** — `test_e2e_export` now asserts `templates_version` EXACTLY (derived from `gda info`, patch omitted when 0) so a format regression fails RED instead of slipping past a loose regex.
- **[0] documented** the `.mono`/`.double` build-variant limitation (no engine API to reconstruct the suffix; gda targets standard builds) — see #304.

**Decided, not changed:** heavy export stays on the release path (acceptable: +~1min over the light tier, and a release-time safety net); the proof's self-gating skip is by design (covered by the static proof + `-rs`); arm64-Linux arch is latent (CI is x86_64).

## e2e CI trigger decision

- **Nightly** (was weekly) + **dispatch** drive the `godot-e2e` tier; **no** push-to-main (too frequent for a ~9-min suite). Release keeps the full e2e (templates included).

## Final verification

- `pytest -m "not e2e"` → 973 passed.
- macOS local full e2e → 289 passed / 0 failed.
- Linux CI dispatch (latest) → **287 passed, 3 skipped, 0 failed** (the 3 skips are headless windowed-capture only; the #301 proof + the restored missing-templates test both run and pass).

qinhaihong-red:
## Review pass 1: `$review` (Standards + Spec)

Compared `main...f0b142c773d8803e756ef4bdd22d37748716552b`.

## Standards

Hard violations: none found.

Judgement call: `src/gda/ops/operations.gd` now documents and implements a limitation to **standard official builds** when deriving the export-template version dir (`.mono` / `.double` build variants may under-report templates). The documented support boundary in `docs/adr/0003-target-godot-version.md` is still "Godot 4.x, minimum 4.4, tested against 4.6"; `docs/adr/0021-gda-daemon-transport-discovery-and-live-version-floor.md` only narrows **live** operations to Godot 4.6+ on Unix. If this narrowed export-template support boundary is intentional, please reflect it in the support ADR/docs; otherwise the path logic should avoid turning those builds into false `export_templates_missing` results.

## Spec

1. The #304 verification is still too indirect for the stated acceptance criterion. Issue #304 asks that the fix be "exercised end-to-end by the export-family e2e on a Linux runner (4.6-stable, a `.0` release) with templates installed — `gda export get` reports `templates_installed:true` and the #301 proof runs." The diff installs/verifies template files in CI, but the e2e suite does not assert that `gda export get` must report `templates_installed:true` in that template-installed Linux context. `tests/test_e2e_export.py` only checks that the value is a bool plus the expected version string, while `test_template_feature_gates_the_harness_only_in_exported_builds` still skips if `templates_installed` is false. A regression in the Linux user-dir case or `.0` version-dir logic could therefore turn the required #301 proof into a loud skip instead of a failing #304 acceptance test. Please add a CI/template-installed assertion that `templates_installed` is true before the proof is allowed to skip/pass.

2. Scope note: changing the `godot-e2e` schedule from weekly to nightly goes beyond issue #301's acceptance text, which asks to document the export-template runner requirement wherever the e2e job is configured. This may be a reasonable product decision, but it should be explicitly accepted in the PR/issue trail or split from the proof/fix if the cadence change is not part of this scope.

Otherwise, the #301 behavioural proof is implemented: it exports a real debug template binary, runs it with a `gda-daemon <socket> <token>` marker against a live Unix socket, asserts no connection from the template binary, includes an editor-binary positive control, verifies the harness is actually packed, and replaces the old tripwire.

Summary: Standards: 0 hard violations, 1 judgement call. Spec: 1 required verification gap, 1 scope note. Worst issue: the #304 e2e acceptance can still become a skip rather than a failure if template detection regresses.

Verification I ran locally in an isolated `/tmp` clone:

- `env UV_CACHE_DIR=/tmp/gda-pr303-uv-cache uv run --frozen pytest tests/test_export_commands.py tests/test_export_run_commands.py tests/test_harness_install.py -q` -> 36 passed
- `git diff --check main...HEAD` -> clean

Observed GitHub checks: latest PR-head CI passed the Python job and skipped the PR e2e job by workflow condition; the last workflow-dispatch e2e run I saw was on `9ca1e4f`, before the final `f0b142c` schedule-only commit.

qinhaihong-red:
## Review pass 2: `agent-skills:code-review-and-quality`

Findings first, by requiredness.

1. Required: make the #304 template-detection acceptance fail, not skip, on template-installed Linux CI.

   `tests/test_e2e_export.py` currently asserts `templates_installed` is a bool and that `templates_version` has the expected string, but it does not assert `templates_installed is True` in the CI context where `.github/actions/install-godot/action.yml` has just installed matching templates. The new behavioural proof in `tests/test_e2e_harness_install.py` still gates itself with:

   ```python
   if not templates_installed(gda, preset=target.preset):
       pytest.skip(...)
   ```

   That preserves a correctness hole in the #304 regression story: if the Linux case-sensitive `godot` / `.0` version-dir detection breaks again, the e2e proof can become a visible skip while the suite remains green. Since #304's acceptance criterion explicitly says the Linux `.0` runner should report `templates_installed:true` and run the #301 proof, CI needs a hard assertion in the template-installed path. For example, add a Linux/CI e2e assertion after `install-godot`'s template install, or make the #301 proof fail rather than skip when the action-installed template file is present but `gda export get` reports false.

2. Consider: document the new standard-build-only export-template boundary outside the code comment.

   `src/gda/ops/operations.gd` now explicitly accepts that `.mono` / `.double` non-standard Godot builds may under-report templates because the suffix is not reconstructible from `Engine.get_version_info()`. That is a reasonable engineering tradeoff, but it narrows practical support for export-template detection while `docs/adr/0003-target-godot-version.md` still describes the support boundary only as Godot 4.x / 4.4+ / tested 4.6. Please either reflect this export-template caveat in the support docs/ADR, or link the issue comment from a durable doc so future agents do not treat this as an accidental false-negative bug.

3. FYI verification caveat: the latest PR head (`f0b142c`) has no completed workflow-dispatch e2e run in the run list I checked; the latest head PR CI passed Python and skipped Godot e2e by workflow condition. The last workflow-dispatch e2e success I saw was on `9ca1e4f`, before the final schedule-only commit. That is probably acceptable because `f0b142c` only changes the cron cadence, but the PR's verification note should be precise about which SHA the full e2e evidence belongs to.

No security or performance concerns stood out in the diff. The new socket probe/test structure is readable and has the right anti-vacuous checks: positive editor control, packed-harness assertion, and a negative template-binary connection assertion.

Verification I ran locally in an isolated `/tmp` clone:

- `env UV_CACHE_DIR=/tmp/gda-pr303-uv-cache uv run --frozen pytest tests/test_export_commands.py tests/test_export_run_commands.py tests/test_harness_install.py -q` -> 36 passed
- `git diff --check main...HEAD` -> clean

Verdict: request changes for finding 1; findings 2-3 are non-blocking but worth tightening before merge.

qinhaihong-red:
## Response to Review pass 1 & 2

**[Required — both passes] The #304 acceptance could become a silent SKIP instead of a FAILURE — FIXED (`b96617f`).**
Added `test_templates_installed_is_true_when_host_templates_are_on_disk` (`tests/test_e2e_export_run.py`): it globs the engine's real user data dir for a platform template file **independently of gda's own path logic** (the code #304 fixed), and where that file is present it asserts `gda export get` reports `templates_installed=true` — **hard**. So a regression in the Linux `godot`/`Godot` case OR the `.0` version-dir name now fails **RED** (templates on disk but reported missing → disagreement → assertion fails), rather than letting the #301 proof skip green. It skips only where templates are genuinely absent (a dev box without them); **CI always has them**, so the assertion always runs there. Verified on the Linux runner: see below — it runs (not skips) and passes. (I left the #301 proof's own absent-templates skip as-is; this dedicated assertion is the cleaner of the two options you offered.)

**[Standards judgement call / pass 2 #2] Reflect the standard-build-only export-template boundary in the support docs — DONE (`b96617f`).**
Added a dated consequence to `docs/adr/0003-target-godot-version.md` (#304): export-template detection is standard-official-builds only — a `.mono`/`.double` build names its real dir with a suffix `Engine.get_version_info()` exposes no field to reconstruct, so it may under-report on those variants; intentional within the 4.x/4.4+ target, not an accidental false-negative. The code comment at `_export_templates_version_dir` and #304 (with a tracking comment) now have a durable doc reference.

**[Pass 1 Spec #2 — scope note] weekly→nightly cadence.**
Intentional, explicitly accepted: the cadence was a deliberate decision in the review thread (the heavy tier ran only weekly + forgettable dispatch, so a real-engine regression could sit up to a week). It is a 1-line cron change cohesive with the e2e tier this PR is already wiring up, does not touch the proof/fix, and can be reverted independently. Recorded here in the trail; happy to split it to its own commit/PR if you'd rather keep #301 cron-neutral.

**[Pass 2 #3 — FYI, verification SHA] Resolved.**
Re-ran the full workflow-dispatch e2e on the exact current head **`b96617f`** (not the earlier `9ca1e4f`): **success — 288 passed, 3 skipped, 0 failed** (run 28240664189). The 3 skips are the headless windowed-capture tests; the #301 proof, the restored missing-templates test, AND the new template-installed hard assertion all run and pass on Linux. `pytest -m "not e2e"` → 973 passed; macOS local full e2e → 290 passed / 0 failed.

Thanks for both passes — the required gap (skip-vs-fail) was exactly the failure mode this whole change set was chasing, so it's good to have it nailed shut with a red-on-regression assertion.

Issue #309, opened 2026-06-27
ci: pyright strictness ratchet + Astral ty migration watch


qinhaihong-red:
## What to build

A decision/tracking issue for tightening the type gate over time, once the `basic`-mode gate
(#307) is in place.

Two threads:
- **Strictness ratchet (config-only).** Incrementally enable stricter pyright rules beyond
  `basic` (e.g. selected `reportUnknown*` / `strict` list members), one rule at a time, each
  with its own small cleanup — rather than flipping `typeCheckingMode = "strict"` and absorbing
  hundreds of findings at once.
- **Astral `ty` watch.** `ty` aligns with the repo's uv + ruff (ADR-0029) toolchain and already
  runs here (65 diagnostics today), but is pre-release. Track it and migrate the gate from
  pyright to `ty` once it stabilizes, if the alignment pays off.

This is **HITL** — it carries judgement calls (how strict, whether/when to switch tools), not a
mechanical change. Likely **deferred**; filed so the decision isn't lost.

## Acceptance criteria

- [ ] A decision is recorded (in the issue and, if it changes the gate, an ADR update) on: which stricter rules to adopt and on what cadence, and whether/when to migrate to `ty`.
- [ ] Any adopted strictness step keeps `uv run pyright` (or `ty`) at **0 errors** and the `type-check` job green.

## Blocked by

- #307 (Stage 1 — pyright tooling + `src/` gate)

qinhaihong-red:
## Assessment & decision (2026-06-27) — DEFERRED

Measured on `main` @ `9379791` (after Stage 2, #311 merged; gate = pyright `basic`, `include = ["src", "tests"]`, **0 findings**). Recording the tradeoff for future reference; **not actioning Stage 3 now.**

### Strictness ratchet — measured

| Config | findings | composition |
| --- | --- | --- |
| full `strict` (src + tests) | **7560** (src 313 / tests 7247) | ~97% is the *annotate-everything* family: `reportUnknownArgumentType/MemberType/ParameterType/VariableType` + `reportMissingParameterType` (≈7335) |
| `strict`, src only | 313 | still ~88% the same annotate-everything noise |
| curated "quality" rules (8 added, excluding the annotate family) | **107** | `reportMissingTypeArgument` 100 + `reportUnusedFunction` 4 + `reportIncompatibleMethodOverride` 2 + `reportIncompatibleVariableOverride` 1 |
[...]

Issue #308, opened 2026-06-27
ci: extend the pyright gate to tests/


qinhaihong-red:
## What to build

Extend the pyright type-check gate (introduced for `src/` in #307) to cover `tests/` as well,
so the whole repo is type-checked on every PR.

`tests/` currently has ~109 pyright findings, overwhelmingly **union-narrowing noise**: a test
accesses `.text` (or similar) on an MCP content union (`TextContent | ImageContent | AudioContent
| ResourceLink | EmbeddedResource`) whose concrete shape the test already knows. The fix is one
small shared, typed helper — an assert-and-return (or `assert isinstance(...)`) at the access
sites that narrows the union once and is reused — rather than scattering `# pyright: ignore`
across the suite. Any genuinely incorrect test assertions surfaced along the way are fixed.

Once `tests/` is clean, `[tool.pyright]` `include` is widened to cover it and the `type-check`
CI job checks the whole repo.

## Acceptance criteria

- [ ] A single shared, typed narrowing helper is introduced and used at the union-access sites (no scattered per-line ignores).
- [ ] `[tool.pyright]` `include` widened to cover `tests`; the `type-check` CI job runs against the whole repo.
- [ ] `uv run pyright` reports **0 errors** across the entire repo.
- [ ] `uv run --frozen pytest -m "not e2e"` still passes.

## Blocked by

- #307 (Stage 1 — pyright tooling + `src/` gate)

## Note

"Type-check `src/` only, forever" is a legitimate end-state. If the test-suite churn proves not
worth it, this issue may be closed as `wontfix` and the `src/`-only gate from #307 stands.

PR #310, opened 2026-06-27
ci: pyright type-check gate for src/ (basic mode)

### LINKED ISSUE #307: ci: pyright type-check gate for src/ (basic mode)
## What to build

A blocking CI type-check gate for the shipped library (`src/`), using **pyright** in `basic`
mode — the type-level sibling of the existing ruff `lint` gate (ADR-0029). CI today checks
lint + tests + build but nothing checks types, even though `src/` is heavily annotated (0
`type: ignore`) and the structured-output Pydantic models are the public ABI (ADR-0002 /
ADR-0015). A type gate guards None-safety and ABI-shape drift that ruff's F-rules cannot.

End-to-end: pyright becomes a pinned dev dependency, `[tool.pyright]` is configured to check
`src/` at `basic`, the 21 current `src/` findings are driven to zero, and a parallel
`type-check` CI job enforces it on every PR. The decision is recorded as an ADR and the
workflow is documented for contributors.

The 21 findings resolve in three honest buckets:
- **~4 `reportRedeclaration`** in the CLI command module — the Typer same-named-subcommand
  idiom (`create`/`get` attached to different sub-apps), the same false positive ruff's `F811`
  already exempts (ADR-0023). Suppress consistently with a reason.
- **~5 framework-friction** cases (Typer command-tree introspection on an `object`-typed node,
  a Click `parse_args` override, a Pydantic field narrowed on inheritance) — intentional; each
  gets a targeted `# pyright: ignore[<rule>]` carrying a one-line reason, not a blanket ignore.
- **~12 Optional / argument-type** flow findings (e.g. a method call on a possibly-`None`
  socket, `None` passed where `str` is expected) — **fixed properly** (guards / tightened
  annotations). Any that prove to be genuine latent None bugs are called out in the PR. This is
  where the gate earns its keep.

Mode is `basic`, not `strict` (a strict ratchet is Stage 3). Tool is pyright, not mypy
(needs the pydantic plugin, slower) nor Astral `ty` (pre-release — recorded as a future
migration in Stage 3).

## Acceptance criteria

- [ ] pyright pinned in the dev dependency group and locked in `uv.lock`; `uv sync --frozen --dev` provides it.
- [ ] `[tool.pyright]` added (`include = ["src"]`, `pythonVersion = "3.13"`, `typeCheckingMode = "basic"`).
- [ ] `uv run pyright` reports **0 errors** on `src/`; suppressions carry a reason; the Optional/arg-type findings are fixed, not blanket-ignored.
- [ ] A parallel `type-check` CI job runs `uv run --frozen pyright`, green, mirroring the ruff `lint` job (reuses `setup-python-env`, `save-cache: false`).
- [ ] `uv run --frozen pytest -m "not e2e"` still passes.
- [ ] A negative test confirms the gate blocks a real type error.
- [ ] Decision recorded as **ADR-0030**; README Development/Contributing document `uv run pyright` + the gate (**no ADR link in the README**).

## Blocked by

None - can start immediately.

## Follow-on stages

Part of a staged pyright rollout: **Stage 2** extends the gate to `tests/`, **Stage 3** considers
a strictness ratchet + an `ty` migration. Both depend on this issue.

qinhaihong-red:
Closes #307

## Why

CI gates lint + format (ruff, ADR-0029), tests, and build — but **nothing checks types**, even though `src/` is heavily annotated (0 `type: ignore`) and the structured-output Pydantic models are the public ABI (ADR-0002 / ADR-0015). A type gate guards None-safety and ABI drift that ruff's F-rules can't.

## What

Stage 1 of the staged pyright rollout (#307–#309): a blocking `type-check` CI job over `src/` in `basic` mode, the type-level sibling of the ruff `lint` job.

**Two commits:**

1. **`fix(types):` make `src/` pass pyright** — resolve the 18 `basic` findings honestly (not blanket-ignored):
   - *Real None-safety / narrowing:* guard a connection-less session with a structured `engine_disconnected`; `assert isinstance(op, str)` on the daemon control message (ADR-0021); `assert cmd.recipe is not None` for the recipe-dispatch invariant (ADR-0023); restructure an export guard to narrow `Path | None`; widen `render_node_tree` to `SceneNode | GameNode` (both carry `name`/`type`/`children`); use the `getattr` idiom in `surface.py`.
   - *Annotation widening:* the `classify_*` helpers that only interpolate `binary` into a message take `Path | None` (live ops legitimately pass `None`).
   - *Two reasoned suppressions:* file-wide `reportRedeclaration=false` in `cli.py` (the Typer same-named-subcommand idiom, the analogue of the ruff `F811` ignore, ADR-0023); two `# pyright: ignore[reportArgumentType]` on the shared `RunnerFactory`/`Classifier` seam where `binary` is `None` only for live ops whose runner/classifier ignore it (ADR-0017) — a precise type there would cascade across ~15 sites.

2. **`ci:` add the gate** — pin pyright as a dev dep (locked in `uv.lock`), `[tool.pyright]` (`include=src`, `pythonVersion 3.13`, `typeCheckingMode basic`), the parallel `type-check` job, ADR-0030, and README docs (no ADR link, per the README-is-onboarding convention).

## Scope / non-goals

- `tests/` (109 union-narrowing findings) is **Stage 2** (#308); `strict` ratchet + an `ty` migration watch is **Stage 3** (#309).

## Verification

- `uv run --frozen pyright` → **0 errors** on `src/`.
- Negative test: a deliberate `int`-as-`str` makes `pyright` exit non-zero; reverts to 0.
- `uv run --frozen pytest -m "not e2e"` → **973 passed** (the two added guards change no existing behavior).
- `uv run --frozen ruff check .` / `ruff format --check .` → clean.
- `ci.yml` parses; jobs = `lint`, `type-check`, `python`, `godot-e2e`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,25 @@ jobs:
       - name: Check formatting
         run: uv run --frozen ruff format --check .
 
+  type-check:
+    name: Type check (pyright)
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Set up Python environment
+        uses: ./.github/actions/setup-python-env
+        with:
+          # The python job already populates the dependency cache for this key;
+          # this job only reads it, so skip a redundant (racing) cache save.
+          save-cache: false
+
+      - name: Type check
+        # Scope (src) and mode (basic) come from [tool.pyright] in pyproject (ADR-0030).
+        run: uv run --frozen pyright
+
   python:
     name: Python unit tests and package build
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -599,6 +599,7 @@ uv run pytest -m e2e          # only the end-to-end tests (needs Godot 4.4+ on t
 
 uv run ruff check .           # lint
 uv run ruff format .          # auto-format (append --check to verify without writing)
+uv run pyright                # type-check (src/, basic mode)
 ```
 
 The `e2e` tier runs by default with `uv run pytest`, and **fails loudly** — naming the
@@ -610,6 +611,10 @@ place of flake8 + black + isort, configured under `[tool.ruff]` in `pyproject.to
 pinned via `uv.lock` so local and CI agree. CI's `lint` job runs `ruff check .` and
 `ruff format --check .` on every PR; run `uv run ruff format .` before committing to stay
 green.
+
+Types are checked by [pyright](https://microsoft.github.io/pyright/) in `basic` mode, scoped to
+`src/` and configured under `[tool.pyright]` in `pyproject.toml` (also pinned via `uv.lock`).
+CI's `type-check` job runs `uv run pyright` on every PR.
 
 ```
 src/gda/
@@ -644,8 +649,9 @@ Contributions are welcome. Read [`CONTEXT.md`](CONTEXT.md) to align with the pro
 shared language, and review the relevant [ADRs](docs/adr/) for the area you're touching.
 Issues and PRDs live as [GitHub issues](https://github.com/aigengame/godot-agent/issues).
 Commits follow the [Conventional Commits](https://www.conventionalcommits.org/) specification.
-Python code is linted and formatted with [ruff](https://docs.astral.sh/ruff/), enforced in
-CI — run `uv run ruff format .` before committing (see **Development** above).
+Python code is linted and formatted with [ruff](https://docs.astral.sh/ruff/) and type-checked
+with [pyright](https://microsoft.github.io/pyright/), both enforced in CI — run
+`uv run ruff format .` and `uv run pyright` before committing (see **Development** above).
 
 > **Working with an AI coding agent?** This project is built to be agent-navigable —
 > [`AGENTS.md`](AGENTS.md) is the entry point for coding agents, wiring in the project's
--- a/docs/adr/0030-type-check-gate.md
+++ b/docs/adr/0030-type-check-gate.md
@@ -1,0 +1,81 @@
+---
+status: accepted
+---
+
+# Type-check gate: pyright (`basic`), `src/` first, enforced in CI
+
+CI gates lint + format (ruff, [ADR-0029](0029-lint-and-format-gate.md)), tests, and build —
+but nothing checks types. There is no type-checker config, no type-checker dev dependency,
+and no `py.typed`. Yet `src/` is heavily annotated (0 `type: ignore`) and the structured-output
+Pydantic models **are the public ABI** ([ADR-0002](0002-headless-structured-output-contract.md),
+[ADR-0015](0015-structured-params-input-abi.md)). So nothing enforces the type discipline the
+code already practices, and nothing guards None-safety or ABI-shape drift that ruff's F-rules
+cannot catch. This ADR records the type-level sibling of the ruff gate.
+
+## Decision
+
+**Enforce a type-check gate with pyright in `basic` mode, scoped to `src/`, in CI.** Rolled out
+in stages (issues #307–#309); this ADR covers the Stage 1 `src/` gate.
+
+- **Tool: pyright (pinned dev dependency, locked in `uv.lock`).** Zero-config, fast, and it
+  infers Pydantic/Typer out of the box; CI runs `uv run --frozen pyright`. Pinned because a
+  pyright upgrade can add new checks — so the bump is a deliberate `uv.lock` change, reviewed on
+  its own, not a surprise PR failure (same stance the ruff pin takes).
+
+- **A parallel `type-check` CI job**, mirroring the ruff `lint` job — reuses the
+  `setup-python-env` composite with `save-cache: false` (the `python` job populates that key),
+  fails fast, and gives an independent signal.
+
+- **Mode `basic`, scope `src/`.** `[tool.pyright]` sets `include = ["src"]`,
+  `pythonVersion = "3.13"`, `typeCheckingMode = "basic"`. `basic` is the high-signal starting
+  point; `strict` would add hundreds of `reportUnknown*` findings at once. The 109 `tests/`
+  findings (mostly union-narrowing noise) are out of scope here — extending the gate to `tests/`
+  is Stage 2 (#308), a strictness ratchet is Stage 3 (#309).
+
+- **The 21 `src/` findings were resolved in three honest ways**, not blanket-ignored:
+  - *Real None-safety / narrowing fixes* (the majority): a guard returning a structured
+    `engine_disconnected` when a session has no connection; an `assert isinstance(op, str)` on the
+    daemon control message ([ADR-0021](0021-gda-daemon-transport-discovery-and-live-version-floor.md));
+    an `assert cmd.recipe is not None` documenting the recipe-dispatch invariant
+    ([ADR-0023](0023-command-descriptor-single-registration.md)); a guard restructured to narrow a
+    `Path | None`; `render_node_tree` widened to `SceneNode | GameNode` (the two share
+    `name`/`type`/`children`); and `surface.py` switched to the `getattr` idiom it already uses.
+  - *Honest annotation widening*: the `classify_*` functions that only interpolate `binary` into
+    a message take `Path | None`, since live ops legitimately pass `None`.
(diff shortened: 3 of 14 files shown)


qinhaihong-red:
Review pass: `review` skill (Standards / Spec)

Compared `main...HEAD`.

Commits reviewed:
- `624240b` `fix(types): make src/ pass pyright basic (None-safety + typed seams)`
- `77439e6` `ci: add a pyright type-check gate for src/ (basic mode)`

## Standards

No Standards findings.

I checked the documented standards sources before reading the diff: `AGENTS.md`, `CLAUDE.md`, `CONTEXT.md`, ADRs 0002, 0015, 0017, 0021, 0023, 0029, and `pyproject.toml`.

Against those standards, the PR is consistent with the documented conventions: the pyright gate follows the existing CI/tooling pattern from ADR-0029, the type fixes preserve the descriptor-driven command surface from ADR-0023, and the live/headless seams remain aligned with ADR-0017/0021's channel split. I skipped issues that the new pyright/ruff tooling would enforce directly.

Hard violations: none.
Judgment calls: none.

## Spec

Finding: Missing/partial negative-test evidence for the type gate.

Spec quote: "A negative test confirms the gate blocks a real type error."

The diff adds the pyright dependency/config, CI `type-check` job, ADR, README docs, and source fixes, but it does not add a checked-in test file, CI step, script, or fixture that preserves the deliberately-bad-type-error proof. `docs/adr/0030-type-check-gate.md` and the PR body state that the negative test was verified, but the repository diff only records that claim.

If the intended acceptance criterion is manual PR verification, the PR body may be enough; if it is meant as durable repo/CI evidence, this is still missing.

No scope-creep findings found. No incorrectly implemented requirements found beyond the negative-test-evidence gap.

Summary: Standards 0 findings; Spec 1 finding. Worst issue: the negative type-gate test is claimed but not preserved as durable evidence.

qinhaihong-red:
Review pass: `agent-skills:code-review-and-quality`

## Findings

`src/gda/daemon/server.py:116`

The new `assert isinstance(op, str)` turns a malformed daemon IPC request into an uncaught `AssertionError`. `_accept_loop()` does not catch exceptions from `_handle()`, so a bad frame such as `{}` or `{"op": 1}` can terminate the daemon serve loop instead of failing just that request. I reproduced this directly in the PR checkout:

```text
DaemonServer(None)._handle({})
...
AssertionError
```

This is also the wrong construct for runtime validation: Python can remove `assert` under optimized execution, which would let the invalid value flow onward instead of enforcing the protocol. The daemon socket is private to the user, but it is still an IPC boundary, and the protocol reader accepts JSON from another process. Please replace this with an explicit boundary check that returns a structured live/daemon error, or at least drops the malformed request without allowing the exception to escape `_accept_loop()`.

## Review Notes

Correctness/readability/architecture otherwise look good: the pyright config is scoped to `src/`, the CI `type-check` job mirrors the existing ruff job, the descriptor/live-runner suppressions are localized and explained, and the source fixes mostly narrow existing invariants without broad refactors.

Security/performance: no new secrets, external services, unbounded loops, or hot-path cost issues found. The assertion issue above is the only hardening concern I found because it makes malformed IPC input able to kill the daemon.

Verification I ran:

- `uv run --frozen pyright` -> `0 errors, 0 warnings, 0 informations`
- temporary negative probe under `src/` -> pyright failed with `reportArgumentType`, then the probe was removed and pyright returned to green
- focused pytest subset for touched surfaces -> `113 passed`
- `uv run --frozen pytest -m "not e2e"` -> `973 passed, 291 deselected`
- `uv run --frozen ruff check .` -> passed
- `uv run --frozen ruff format --check .` -> passed
- GitHub checks: `Lint and format (ruff)`, `Python unit tests and package build`, and `Type check (pyright)` are passing; `Godot e2e tests` is skipped for this PR trigger

qinhaihong-red:
Thanks — both findings addressed in `9d936fe`.

**Correctness (the real one): `assert isinstance(op, str)` could kill the daemon.** You're right — it was the wrong construct at an IPC boundary (a malformed frame raised an `AssertionError` that escapes `_handle`/`_accept_loop`, and `assert` strips under `-O`). My "consistent with the existing `assert self._listener`" reasoning was wrong: the listener is an internal invariant, whereas `op` is read from another process's JSON frame.

Fixed by replacing it with an explicit boundary guard that **drops** the frame (`return None`): the serve loop survives, the connection closes with no reply, and the client already maps that to a structured `engine_disconnected` (`DaemonRunner._request`: `reply is None → engine_disconnected`). I chose drop over a server-side error envelope because none of the daemon's LIVE codes fit a malformed-control-frame, and `unknown_operation` is an OPERATION-category code (wrong envelope) — registering a new code for a defensive can't-normally-happen case is scope creep. The guard still narrows `op` to `str`, so pyright stays green.

Added a regression test reproducing your exact case: `_handle({})` and `_handle({"op": 1})` now return `None` and don't raise.

**Spec (negative-test evidence).** Clarified the wording in ADR-0030: the negative test is a **PR-time probe** (a deliberate type error → pyright exits non-zero), not checked in — a permanently-broken file would have to be excluded from the very gate it tests. The gate runs on every PR, so it is itself the durable regression mechanism (same model as the ruff gate, ADR-0029). Happy to add a durable gate-efficacy meta-test if you'd prefer that as a follow-up, but I'd keep it out of this PR for parity with the ruff gate.

Verification after the fix: `pyright` 0 on `src/`; `pytest -m "not e2e"` → **974 passed** (+1 regression test); ruff check/format clean.

qinhaihong-red:
Targeted recheck after `9d936fe`.

Thanks for the fix. The direction is right, but the daemon IPC hardening is not complete yet.

## Still Not Fully Fixed

`src/gda/daemon/server.py:115`

The original repro cases are fixed:

- `_handle({})` now returns `None`
- `_handle({"op": 1})` now returns `None`

But `_accept_loop()` passes `_handle()` whatever JSON value `read_message()` decoded. A valid JSON frame is not necessarily an object. Since `_handle()` still starts with `request.get("op")`, non-dict JSON values still escape as uncaught `AttributeError` and can kill the daemon serve loop:

```text
[] RAISED AttributeError 'list' object has no attribute 'get'
'x' RAISED AttributeError 'str' object has no attribute 'get'
```

So the fix closes the exact test cases but not the boundary. Please guard `request` itself before calling `.get()`, e.g. drop non-dict decoded frames as malformed too, or isolate per-request decode/handle failures in `_accept_loop()` so malformed IPC cannot terminate the daemon.

## Documentation Drift

`docs/adr/0030-type-check-gate.md` still says the finding was resolved with:

```text
assert isinstance(op, str) on the daemon control message
```

That is stale after this fix. The ADR should describe the explicit boundary guard/drop behavior instead.

## Verified

- Latest head checked: `9d936fe4fc9164dce7fece0e06debaa02a1737e9`
- `uv run --frozen pyright` -> 0 errors
- `uv run --frozen pytest tests/test_daemon_server.py -q` -> 12 passed
- `uv run --frozen pytest -m "not e2e"` -> 974 passed
- `uv run --frozen ruff check .` / `ruff format --check .` -> passed

The negative type-gate clarification in ADR-0030 looks acceptable to me: documenting it as a PR-time probe, with CI pyright as the durable regression mechanism, matches the ruff gate model.

qinhaihong-red:
Right on both counts — fixed in `5cda8a7`.

**Boundary wasn't fully closed.** My guard checked `op` but still called `request.get(...)` on whatever `read_message` decoded, and a JSON frame need not be an object — so `[]` / `"x"` / `1` still raised `AttributeError`. Closed it completely:

- `_handle` now takes `object` (honest — `read_message` returns `Any | None`) and **drops a non-dict frame** before touching `.get`, in addition to the missing/non-string-`op` drop. Still narrows `op` to `str`.
- `_accept_loop` now **isolates each request in a try/except**, so a decode failure (bad JSON bytes — which a `_handle` guard can't catch) or any deeper per-request failure drops that frame and the daemon keeps serving (`KeyboardInterrupt`/`SystemExit` still propagate). This is the "malformed IPC cannot terminate the daemon" guarantee at the loop level.
- Regression test extended to your exact non-dict repros: `_handle([])`, `_handle("x")`, `_handle(1)` all return `None`.

**Doc drift.** Replaced the stale `assert isinstance(op, str)` line in ADR-0030 with the boundary-drop + serve-loop-isolation description.

Verification: `pyright` 0 on `src/`; `pytest tests/test_daemon_server.py` and `pytest -m "not e2e"` → **974 passed**; ruff check/format clean.

PR #347, opened 2026-07-01
feat(resource): resolve class_name Resource in resource create (#342)

### LINKED ISSUE #342: gda resource create cannot instantiate a GDScript class_name Resource
## Problem

`gda resource create --type <T>` cannot instantiate a custom GDScript `class_name` Resource — only built-in ClassDB classes resolve.

```
$ gda resource create --type GameConfig --path res://data/generated/boot_config.tres --project <proj> --json
# exit 4, error.code = invalid_resource_type
# "not an instantiable Resource class: GameConfig"  (ClassDB: Cannot get class 'GameConfig')
```

`GameConfig` is a normal `class_name GameConfig extends Resource` script registered as a global class. `resource create` resolves the `--type` through `ClassDB` only, which doesn't know GDScript global classes, so any project-defined Resource subclass is unreachable.

## Why it matters

Data-driven projects (e.g. the gADR-0000 JSON→Resource pattern in `examples/platformer/panda-adventure`) author their config as **custom-typed** Resources. Because `resource create` can't make one, the JSON→Resource converter has to **hand-emit `.tres` text** (an `ext_resource` to the script + `[resource]` field assignments) instead of dogfooding gda. `resource get`/`set` work fine on the resulting file — only creation of a custom-typed resource is blocked.

## Suggested direction

Let `gda resource create` target a project script, e.g. accept a script `res://…/foo.gd` path (or a registered global `class_name`) for `--type`, instantiate via the script, and write the `.tres` with the correct `script`/`script_class`. Subject to the ADR-0025 surface-inclusion criterion.

## Acceptance criteria

- [ ] `gda resource create` can produce a `.tres` whose root type is a project-defined `class_name extends Resource`.
- [ ] The generated `.tres` loads and round-trips through `gda resource get`/`set` with correct field types.
- [ ] A clear, structured error remains for a genuinely unknown/non-Resource type.

## Found while

Building the Panda Adventure S0 walking skeleton (#329, PR #341) — gda dogfooding.

qinhaihong-red:
## What

`gda resource create --type <T>` could only instantiate a Godot built-in Resource class via `ClassDB` — a project-defined GDScript `class_name Foo extends Resource` was rejected with `invalid_resource_type`. This makes `resource create` resolve `--type` the same way `node add` already resolves it: ClassDB first, then the project's global class list (issue #342).

## How

All in `src/gda/ops/operations.gd`, mirroring the existing node-side resolvers:

- `_op_resource_create` now resolves `--type` through a new `_instantiate_resource_type(type)` instead of gating on `ClassDB` alone.
- `_instantiate_resource_type` / `_instantiate_resource_script_class` are the Resource-side twins of `_instantiate_node_type` / `_instantiate_script_class` (ClassDB → `ProjectSettings.get_global_class_list()` → load script → guarded construction → validate the instance `is Resource`).
- **Reuses** the shared, type-agnostic `_new_script_instance(script)` guard — no duplicated constructor-isolation logic.
- Failure split mirrors node add: an unknown / non-Resource `class_name` → `invalid_resource_type`; a registered-but-broken / non-instantiable script → `uninstantiable_script`. No new error code (both already exist and have GDScript mirror constants).

A script-backed Resource runs the script's `_init` at construction — project code within the Trusted-project assumption (ADR-0009), same as node add.

## Scope

No Python / `cli.py` / `models.py` / `error_codes.py` change: `ResourceCreateParams.type` is already free-form, and no `script_class` field was added to `ResourceCreateResult` (out of scope for this issue). The create result echoes the requested type in its existing `type` field.

## Tests

Added to `tests/test_e2e_resource.py` (real headless Godot, scaffold → `--import` → invoke, patterned on `test_e2e_node.py`):

- **Positive** — `class_name PandaStats extends Resource` with typed `Color` / `Vector2` / `float` exports: `create` → `get` round-trips each field with its declared type and value.
- **Negative** — a non-Resource `class_name` (`extends Node2D`) → `invalid_resource_type` (message names the script + "not a Resource-derived script").
- **Negative** — a registered-but-broken script → `uninstantiable_script`.

`uv run pytest tests/test_e2e_resource.py -rs` → **27 passed** (3 new + 24 existing).

Closes #342

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -137,8 +137,10 @@ _Avoid_: safe project, sandboxed project
 **Project-code execution surface**:
 The set of points where a single `gda` run triggers the target project's own
 code to run: autoload constructors at engine startup (every `--project` op), the
-`_init` of scripts on nodes that an instantiating operation constructs, and — via
-`gda script run` (ADR-0031) — the **full execution of a named project script**.
+`_init` of scripts on nodes *or resources* that an instantiating operation
+constructs (a `class_name` node via `node add`, or a script-backed `class_name`
+Resource via `resource create`), and — via `gda script run` (ADR-0031) — the
+**full execution of a named project script**.
 All stay within the `Trusted project` assumption (ADR-0009); `script run` widens
 this surface without adding a new trust axis.
 _Avoid_: attack surface, code-execution risk
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -115,7 +115,7 @@ operation, and parse codes the CLI assigns).
 | `invalid_node_name` | `operation` | `operation` | `4` | A requested node name is empty or would be rewritten by Godot. |
 | `duplicate_node_name` | `operation` | `operation` | `4` | The parent node already has a child with the requested name. |
 | `missing_dependency` | `operation` | `operation` | `4` | A scene's declared nodes vanished or degraded on load — an unresolvable instanced sub-scene or an unavailable node class; re-saving would silently drop or downgrade them. |
-| `uninstantiable_script` | `operation` | `operation` | `4` | A registered `class_name`'s script can no longer be loaded, compiled, or constructed, so it cannot be instantiated as a node. |
+| `uninstantiable_script` | `operation` | `operation` | `4` | A registered `class_name`'s script can no longer be loaded, compiled, or constructed, so it cannot be instantiated as a node or a resource. |
 | `node_not_found` | `operation` | `operation` | `4` | A requested node path does not resolve to a node in the scene. |
 | `cannot_target_root` | `operation` | `operation` | `4` | A structural edit targeted the scene root, which has no parent to be removed from, duplicated alongside, or reparented out of. |
 | `cyclic_target` | `operation` | `operation` | `4` | A node move targeted the node itself or one of its own descendants, which would detach the moved subtree from the scene. |
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -2598,7 +2598,11 @@ def create(
     resource_type: str = typer.Option(
         ...,
         "--type",
-        help="Godot resource class of the new .tres (e.g. Gradient, Curve).",
+        help=(
+            "Resource type of the new .tres: a built-in Resource class (e.g. "
+            "Gradient, Curve) or a registered Resource class_name (a GDScript "
+            "class_name Foo extends Resource)."
+        ),
     ),
     json_output: bool = json_option(),
     schema: bool = RESOURCE_CREATE_COMMAND.schema_option(),
--- a/src/gda/error_codes.py
+++ b/src/gda/error_codes.py
@@ -224,7 +224,7 @@ ERROR_CODES: tuple[ErrorCodeSpec, ...] = (
         EXIT_OPERATION,
         ErrorCodeSource.OPERATION,
         "A registered class_name's script can no longer be loaded, compiled,"
-        " or constructed, so it cannot be instantiated as a node.",
+        " or constructed, so it cannot be instantiated as a node or a resource.",
     ),
     ErrorCodeSpec(
         "node_not_found",
--- a/src/gda/models.py
+++ b/src/gda/models.py
@@ -1445,16 +1445,19 @@ class ResourceCreateParams(BaseModel):
 
     ``path`` is the target ``.tres`` resource file, addressed by its ``res://``
     or filesystem path (resource-file addressing — by file path). ``type`` is
-    the Godot resource class to instantiate and save (e.g. ``Gradient``,
-    ``Curve``); it must be an instantiable ``Resource`` subclass, mirroring
-    ``scene create``'s ``root_type`` check against ``Node``.
+    the Resource type to instantiate and save: a built-in Resource class (e.g.
+    ``Gradient``, ``Curve``) OR a project-defined ``class_name`` (a GDScript
+    ``class_name Foo extends Resource``), resolved the same way ``node add``
+    resolves ``--type`` (issue #342) — mirroring ``scene create``'s ``root_type``
+    check against ``Node``.
     """
 
     path: NormalizedPath = Field(description="Target .tres resource path to write.")
     type: str = Field(
         description=(
-            "The Godot resource class to create (e.g. Gradient, Curve). Must be "
-            "an instantiable Resource subclass."
+            "The Resource type to create: a built-in Resource class (e.g. "
+            "Gradient, Curve) or a registered Resource class_name (a GDScript "
+            "class_name Foo extends Resource)."
         )
     )
 
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -1466,6 +1466,11 @@ func _op_script_validate(params: Dictionary) -> void:
 # Resource runs the script's _init at construction (its constructor is project
 # code, within the Trusted project assumption, ADR-0009); a built-in class
 # constructs an engine class and runs none.
+#
+# The no-clobber check runs BEFORE construction, so an existing target is refused
+# without ever executing a script-backed type's _init: a broken constructor over
+# an existing file stays already_exists (no-clobber), never uninstantiable_script,
+# and no _init side effect runs against a target that would not be written anyway.
 func _op_resource_create(params: Dictionary) -> void:
 	_diag("running operation: resource-create")
 	var path := _string_param(params, "path")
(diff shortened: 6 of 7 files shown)


qinhaihong-red:
## Review Skill: Standards / Spec

Reviewed diff: `git diff main...HEAD`

Commits reviewed:
- `5153fe4` test(resource): e2e cover resource create by class_name
- `f85c615` feat(resource): resolve class_name Resource in resource create

## Standards

Hard violations:

- `src/gda/ops/operations.gd:1465-1467` / `3659-3666` adds a new project-code execution point: constructing a script-backed `Resource` runs that resource script's `_init`. The documented `Project-code execution surface` only covers autoload startup, scene-node script `_init`, and `script run` (`CONTEXT.md:137-143`), while ADR-0009 says trusted-project execution is accepted and documented (`docs/adr/0009-trust-boundary-trusted-project.md:28-32`, `38-49`). The trust-boundary docs are now incomplete.

- `src/gda/ops/operations.gd:3647-3664` reports `uninstantiable_script` for resource-script load/compile/constructor failures. That widens a public `GdaError.code` whose registry/ADR meaning is still node-specific: "cannot be instantiated as a node" (`src/gda/error_codes.py:222-227`, `docs/adr/0002-headless-structured-output-contract.md:118`). ADR-0002 treats error codes as public ABI and requires registry/ADR/GDScript mirror updates when changing them (`docs/adr/0002-headless-structured-output-contract.md:82-85`, `184-186`). Either generalize the code's documented meaning or introduce a resource-specific code.

- `src/gda/ops/operations.gd:1459-1464` / `3631-3637` changes accepted `resource create --type` inputs to include project `class_name`s, but the model-driven public input schema still describes `type` only as a Godot resource class / instantiable `Resource` subclass (`src/gda/models.py:1443-1458`), and CLI help has the same narrower wording (`src/gda/cli.py:2597-2602`). ADR-0004/0015 make the params model the source for schema and `--params-json` semantics (`docs/adr/0004-schema-flag-self-description.md:47-53`, `docs/adr/0015-structured-params-input-abi.md:24-28`). Update the model/help/schema tests with the new `class_name` resolution contract.

Judgement calls: none beyond the hard drift above.

## Spec

- [P1] The `get/set` round-trip requirement is only partially covered. The spec says the generated `.tres` should "round-trip through `gda resource get`/`set` with correct field types (e.g. Color / Vector2 / float fields preserved)." The new custom-resource e2e only does create -> get and checks initial exported values; existing `resource set` e2e still uses built-in `Gradient` int/String properties. A script-backed Resource is never set and reloaded.

- [P2] The public `--type` contract still describes only engine resource classes. The spec says `gda resource create --type <T>` should accept a registered GDScript global `class_name`; runtime now does, but schema/help text still says "Godot resource class" / "instantiable Resource subclass" with only `Gradient`/`Curve` examples. Agents using `--schema` or CLI help will not learn that `class_name` is valid.

- [P2] `resource create` now instantiates before the existing-target check. The spec asks to instantiate `class_name` Resources for new `.tres` creation, but this ordering calls `script.new()` and can run `_init` even when the target already exists and the command will fail with `already_exists`; a broken constructor can also change the failure from no-clobber to `uninstantiable_script`. That side effect/error-ordering was not requested by the issue scope.

Summary: Standards found 3 hard drift issues; Spec found 3 issues. Worst issue: `resource create` now runs script constructors before the no-clobber `already_exists` check.

Verification run locally on the PR branch:
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run pytest tests/test_e2e_resource.py -rs` - 27 passed
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run ruff check .` - passed
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run ruff format --check .` - passed
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run pyright` - 0 errors

qinhaihong-red:
## Code Review and Quality

### Findings

- `src/gda/ops/operations.gd:1478-1483` instantiates the requested resource before the no-clobber check. For a script-backed `class_name` Resource, `_instantiate_resource_type()` reaches `script.new()` before checking whether `path` already exists. That means an `already_exists` call can still run project `_init` code, and a broken constructor can return `uninstantiable_script` instead of preserving the no-clobber behavior. Please move the existing-target check before script construction, or split resolution from construction so existing targets are refused without executing constructors.

- `tests/test_e2e_resource.py:221-272` does not cover the full custom-resource round trip required by #342. The new positive test creates a `PandaStats` resource and verifies `resource get`, but it never uses `resource set` on a script-backed Resource and reloads it. Existing `resource set` e2e coverage only exercises built-in `Gradient` fields. Add a create -> set -> get e2e for at least one custom exported field so the `get/set` acceptance criterion is pinned.

- `src/gda/models.py:1443-1458` and `src/gda/cli.py:2597-2602` still describe `--type` as a Godot resource class with only `Gradient`/`Curve` examples. Because `--schema` is generated from the params model, agents will still see the old public contract even though runtime now accepts registered GDScript `class_name`s. Update the model docstring/field description, CLI help, and schema/help tests to describe "built-in Resource class or registered Resource `class_name`".

- `CONTEXT.md:137-143`, `src/gda/error_codes.py:222-227`, and `docs/adr/0002-headless-structured-output-contract.md:118` still document the old boundaries. This PR intentionally adds Resource-script constructor execution and reuses `uninstantiable_script` outside node construction, so the trust-boundary glossary and public error-code registry need to be generalized or otherwise updated with the new meaning.

### Review Axes

- Correctness: blocked by the no-clobber ordering issue above.
- Tests: strong e2e coverage for create/get and error splits, but missing the custom `resource set` path.
- Architecture: implementation mirrors `node add` well and reuses `_new_script_instance`; public ABI/docs need to move with the behavior.
- Security/trust: no new untrusted-project model is required under ADR-0009, but the project-code execution surface must be documented.
- Performance: no material concern; this is a small global-class-list fallback on a command path.

Verification run locally on the PR branch:
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run pytest tests/test_e2e_resource.py -rs` - 27 passed
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run ruff check .` - passed
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run ruff format --check .` - passed
- `env UV_CACHE_DIR=/private/tmp/uv-cache-pr347 uv run pyright` - 0 errors

qinhaihong-red:
## Review findings addressed — `b01a788`

Thanks for the review. All six findings (3 Standards + 3 Spec) are addressed in `b01a788`, on top of `5153fe4` / `f85c615`.

### Standards

- **`uninstantiable_script` public ABI (node-specific meaning).** Generalized the documented meaning instead of adding a new code (the name is already script-generic). Updated the `ErrorCodeSpec` description in `src/gda/error_codes.py` and the ADR-0002 registry row (`docs/adr/0002-…md:118`) from "…as a node." → "…as a node or a resource." Code name/category/source/exit unchanged; the drift test (`test_adr_registry_matches_python_authoritative_registry`) stays green.

- **`--type` public input schema/help drift.** `ResourceCreateParams` docstring + `type` field description (`src/gda/models.py`) and the CLI `--type` help (`src/gda/cli.py`) now document *"a built-in Resource class (e.g. Gradient, Curve) or a registered Resource `class_name` (a GDScript `class_name Foo extends Resource`)"*, so `--schema` (ADR-0004/0015) and CLI help reflect the widened contract. No test asserted the old prose (the schema test is a self-consistency check against `model_json_schema()`).

- **Trust-boundary glossary incomplete.** `CONTEXT.md` "Project-code execution surface" now enumerates the `_init` of scripts on nodes **or resources** that an instantiating operation constructs (a `class_name` node via `node add`, or a script-backed `class_name` Resource via `resource create`). No ADR-0009 change — the trust model is unchanged; only the surface enumeration was incomplete.

### Spec

- **[Worst / P2] `resource create` ran constructors before the no-clobber check.** Reordered: the `file_exists`/`already_exists` check now runs **before** `_instantiate_resource_type`, so an existing target is refused without ever calling `script.new()`. A broken-constructor script-backed `class_name` over an existing file now stays `already_exists` (never `uninstantiable_script`), and no `_init` runs. Chose the simple reorder over a resolve/construct split because no test pins "invalid-type-before-already_exists" precedence (verified by grep). New e2e `test_resource_create_over_existing_target_refuses_before_constructing` (a registered `NeedsArgsStats` whose `_init` requires args, over an occupied path → asserts `already_exists` + file untouched; getting `already_exists` rather than `uninstantiable_script` is itself the proof the constructor never ran).

- **[P1] get/`set` round-trip only partially covered.** Added e2e `test_resource_create_by_class_name_round_trips_a_set_via_get` — create `PandaStats` → `resource set speed 9.5` → `resource get` asserts the `float`/`9.5` persisted on a script-backed Resource.

- **[P2] `--type` self-description** — same fix as the Standards item above.

### Verification (from the PR branch)
- `uv run pytest tests/test_e2e_resource.py tests/test_resource_errors.py tests/test_error_registry.py -rs` → **56 passed** (incl. the drift test + the new no-clobber test).
- `uv run pytest tests/test_schema_command.py tests/test_models.py tests/test_params_json.py` → **216 passed**.
- `ruff check` / `ruff format --check` / `pyright` → clean, 0 errors. CI on `b01a788` is green.

### Note on the shared error-code registry
Fixing the `uninstantiable_script` meaning means this PR now edits `error_codes.py` + ADR-0002 — the same files as #350, but on **different rows** (`error_codes.py`: L224 here vs L615 there; ADR-0002: L115 here vs L160 there), so the two merge cleanly in either order. This is deliberate: the public ABI/docs move with the behavior rather than being kept artificially disjoint.

Issue #343, opened 2026-06-30
No headless GDScript runner (gda script run) for logic-seam tests


qinhaihong-red:
## Problem

gda has no way to **execute an arbitrary GDScript headless and capture structured output**. There is no `gda script run` (or `gda headless run`). `gda script` only does `create`/`get`/`set`/`list`/`validate`/`attach` (text + compile-check), and `HeadlessCommand` is an internal execution channel, not a user-facing runner.

## Why it matters

A data-driven game's **pure logic seam** — Controller/formula logic over Resource data, asserted headless (the core later reused by a Monte-Carlo balancing sim) — needs to run a `.gd` (a `SceneTree`/`MainLoop` script) and read its result. With no gda runner, the test must **shell out to `godot --headless --path <proj> --script res://…`** directly, bypassing gda entirely (no structured result, no engine-error classification, manual binary resolution). A first-class runner would let logic tests dogfood gda end to end and get the usual structured-output + error-envelope treatment.

## Suggested direction

A `gda script run res://path.gd` (and/or a project-scoped headless eval) that runs the script via `godot --headless --script`, captures stdout/stderr, classifies engine failures into the standard error envelope, and returns a structured result. Evaluate against the ADR-0025 surface-inclusion criterion (agent value + structured-op fit) and the existing `gda logger tail` / `diag errors` observability surface.

## Acceptance criteria

- [ ] gda can run a project `.gd` headless and return a structured result (exit status + captured output), without the caller spawning Godot themselves.
- [ ] Engine/launch failures are classified into the standard error envelope (not raw passthrough).
- [ ] The logic-seam test in the example game can use the gda command instead of a raw `godot --script` subprocess.

## Found while

Building the Panda Adventure S0 walking skeleton (#329, PR #341) — gda dogfooding.

qinhaihong-red:
> *This was generated by AI during triage.*

A `grill-with-docs` session settled the open design questions. The authoritative contract now lives in **ADR-0031** (`docs/adr/0031-headless-script-run-passthrough-execution.md`), with an outcome note on ADR-0010 and a `--script` confirmation note on ADR-0025. This brief is the implementation spec; ADR-0031 is the contract.

## Agent Brief

**Category:** enhancement
**Summary:** Add `gda script run res://path.gd` — a first-class headless runner for an arbitrary project GDScript that returns a structured passthrough result and classifies only launch/crash failures (per ADR-0031).

**Current behavior:**
[...]

PR #348, opened 2026-07-01
fix(export): absolutize project so a relative --project works (#344)

### LINKED ISSUE #344: gda export run fails with a relative --project (path applied twice)
## Problem

`gda export run` fails when `--project` is a **relative** path, while the equivalent native `godot --headless --path <rel> --export-<mode>` succeeds.

```
$ cd <repo-root>
$ gda export run --preset macOS --mode pack --project examples/platformer/panda-adventure --json
# exit 4, error.code = export_failed, empty diagnostics
$ gda export run --preset macOS --mode pack --project "$PWD/examples/platformer/panda-adventure" --json
# exit 0  (absolute path works)
```

Confirmed for both `release` and `pack`.

## Root cause

`SubprocessExportRunner.run` (`src/gda/export_runner.py`) spawns the engine with **both** `cwd=project` **and** `--path project` in the argv. When `project` is relative, the path is applied twice: the child process is already CWD'd into the project dir, then Godot resolves `--path <relative>` *against that dir*, pointing at a non-existent nested path, so the engine can't find `project.godot` and the export fails (with empty diagnostics, which also obscures the cause).

The sentinel op-dispatch channel (e.g. `gda export get`) is unaffected because it runs with `cwd=None`.

## Suggested fix

Absolutize the project path before spawning (or pass `--path` resolved against gda's original CWD, not the child CWD). Add a regression test running `export run` with a relative `--project`. Consider surfacing a non-empty diagnostic when the engine reports a missing project.

## Acceptance criteria

- [x] `gda export run` succeeds with a relative `--project` (same result as an absolute one). *(delivered in #348)*
- [x] A regression test covers a relative-`--project` export run. *(delivered in #348)*
- [ ] ~~A missing/invalid project yields a structured, non-empty diagnostic rather than an empty `export_failed`.~~ **→ Split to #353.** This turned out to be a pre-existing **cross-cutting** gap — an invalid `--project` raises an uncaught `ValueError` in *every* channel (sentinel and recipe alike), not export-specific — so it is tracked and fixed in #353 (a single shared conversion, blocked on #343), not band-aided into #348.

> **Scope update (2026-07-01):** #348 is scoped to the relative-`--project` fix (AC#1 + AC#2, both delivered). AC#3's general form moved to #353. See the PR #348 review thread.

## Found while

Building the Panda Adventure S0 walking skeleton (#329, PR #341) — gda dogfooding. Workaround in use: absolute `--project`.

qinhaihong-red:
## Root cause

`gda export run` fails with a **relative** `--project` while an absolute one works.

`SubprocessExportRunner.run` (`src/gda/export_runner.py`) spawns the native export with **both** `cwd = <project>` **and** `--path <project>`. `resolve_project_dir` never absolutizes the project (it only expands `~`), so for a relative `--project` the child process CWDs into `<project>`, and the engine then re-resolves `--path <project>` against that new CWD → `<project>/<project>` → `project.godot` not found → `export_failed` with an **empty** diagnostics.

The sentinel channel is immune because it spawns with `cwd=None` (`src/gda/runner.py`); only the export channel sets `cwd`, so only it must keep `cwd` and `--path` in agreement.

## Fix (minimal, single file)

In `SubprocessExportRunner.run`, absolutize the project **once** via `Path.absolute()` and use the same path for **both** `cwd` and `--path`, so a relative `--project` reaches the exact resolution an absolute one does.

- Used `Path.absolute()` rather than `Path.resolve()`: it fully fixes the bug (a relative path becomes absolute), keeps the codebase's symlink-agnostic path handling (`resolve_project_dir` only `~`-expands; the sentinel runner passes paths verbatim), and passes an absolute `--project` through unchanged — so the existing `test_export_runs_with_project_as_cwd` (which hardcodes `/tmp/proj`) stays green rather than being rewritten to `/private/tmp/proj` on macOS.
- The shared `launch()` primitive (`runner.py`) is **untouched** — the double-application was in the export caller's argument, not the primitive.
- The class docstring is updated to explain why `--path` must be absolute when `cwd` is the project.

## Tests

- **Unit** (`tests/test_export_runner.py`): a relative `--project` yields an absolute `--path` equal to `cwd`; relative vs absolute produce the identical spawn.
- **E2e** (`tests/test_e2e_export_run.py`): a `--mode pack` export driven from the project's **parent** with a **relative** `--project` runs to exit 0 and writes the `.pck` inside the project. Pack needs no export templates, so it runs **deterministically** on a template-less host (no silent skip masquerading as a pass).
- Both suites were confirmed to **fail without the fix**.

Test commands run (from the worktree, via `uv run`):
- `uv run pytest tests/ -k export -rs -m "not e2e"` → **83 passed**
- `GDA_GODOT=… uv run pytest tests/ -k export -rs -p no:xdist` (incl. e2e) → **105 passed, 1 skipped** (the 1 skip is the pre-existing Linux-only XDG templates test, unrelated)
- ruff check + format + pyright on changed files → clean

Fixes #344

### DIFF
--- a/src/gda/export_runner.py
+++ b/src/gda/export_runner.py
@@ -55,13 +55,22 @@ class SubprocessExportRunner:
     these flags entirely and the engine reports it as a usage error, surfaced as a
     generic export failure by the classifier.
 
-    The process is spawned with ``cwd = <project>`` when a project is resolved.
-    Godot writes the export artifact via ``FileAccess``/``DirAccess`` with NO
-    project globalization (verified against the engine source), so a *relative*
-    output path — the common case, since a preset's configured ``export_path`` is
-    stored project-relative (e.g. ``build/game.x86_64``) — resolves against the
-    process CWD. Running from the project root makes that relative path land
-    inside the project, exactly as the editor resolves it (issue #121).
+    The project is absolutized once and used for BOTH the spawn's
+    ``cwd = <project>`` and the ``--path <project>`` flag when a project is
+    resolved. This matters twice:
+
+    - Godot writes the export artifact via ``FileAccess``/``DirAccess`` with NO
+      project globalization (verified against the engine source), so a *relative*
+      output path — the common case, since a preset's configured ``export_path``
+      is stored project-relative (e.g. ``build/game.x86_64``) — resolves against
+      the process CWD. Running from the project root makes that relative path land
+      inside the project, exactly as the editor resolves it (issue #121).
+    - Because the spawn's ``cwd`` is the project, ``--path`` MUST be absolute. A
+      *relative* ``--project`` would otherwise be re-resolved by the engine
+      against that child CWD — ``<project>/<project>`` — so ``project.godot`` is
+      not found and the export fails with an empty ``export_failed``. Absolutizing
+      the path keeps ``cwd`` and ``--path`` naming the same directory, so a
+      relative ``--project`` resolves exactly as an absolute one (issue #344).
     """
 
     binary: Path
@@ -74,8 +83,15 @@ class SubprocessExportRunner:
         # UTF-8-decode handling to the shared launch primitive (#185).
         flag = EXPORT_MODE_FLAGS[mode]
         args: list[str] = []
-        if self.project is not None:
-            args += ["--path", str(self.project)]
+        # Absolutize the project ONCE and use the same path for cwd and --path:
+        # this channel sets cwd = <project>, so a relative --path would be
+        # re-resolved against that child CWD (<project>/<project>) and Godot would
+        # not find project.godot (#344). ``absolute()`` (not ``resolve()``) keeps
+        # the codebase's symlink-agnostic path handling — resolve_project_dir only
+        # expands ``~`` — so an absolute --project is passed through verbatim.
+        project = self.project.absolute() if self.project is not None else None
+        if project is not None:
+            args += ["--path", str(project)]
         args += [flag, preset, output_path]
         # Pass the export-specific timeout label: on a timeout the primitive's
         # stderr is carried into GdaError.diagnostics and serialized in
@@ -84,7 +100,7 @@ class SubprocessExportRunner:
         return launch(
             self.binary,
             args,
-            cwd=self.project,
+            cwd=project,
             timeout=self.timeout,
             timeout_label="Godot export",
         )
--- a/tests/test_e2e_export_run.py
+++ b/tests/test_e2e_export_run.py
@@ -236,6 +236,61 @@ def test_export_run_pack_writes_pck_without_templates(godot_project):
 
 
 @pytest.mark.e2e
+def test_export_run_pack_accepts_a_relative_project(godot_project):
+    # Regression for #344: `gda export run` with a RELATIVE --project must resolve
+    # the SAME project as an absolute one. The native export runner spawns Godot
+    # with cwd = <project>, so before the fix a relative --path was re-resolved by
+    # the engine against that child CWD (<project>/<project>) — project.godot not
+    # found → empty export_failed. Pack needs no export templates, so this RUNS
+    # deterministically on a template-less host and proves the real bug is gone on
+    # disk (an absolute --project already works via test_..._pack_writes_pck_...).
+    (godot_project / "export_presets.cfg").write_text(
+        EXPORT_PRESETS_CFG, encoding="utf-8"
+    )
+    # all_resources needs at least one exportable file (see the pack test above).
+    (godot_project / "main.gd").write_text(
+        "extends Node\n\nfunc _ready() -> void:\n\tpass\n", encoding="utf-8"
+    )
+    override_rel = "dist/packed.pck"
+    artifact = godot_project / override_rel
+    artifact.parent.mkdir(parents=True, exist_ok=True)
+
+    # Drive gda from the project's PARENT with a RELATIVE --project (the failing
+    # case) — NOT the absolute path _gda_project passes.
+    run = subprocess.run(
+        [
+            *GDA_CMD,
+            "export",
+            "run",
+            "--preset",
+            "Linux/X11",
+            "--mode",
+            "pack",
+            "--output",
+            override_rel,
+            "--json",
+            "--godot",
+            str(GODOT),
+            "--project",
+            godot_project.name,  # RELATIVE; resolved against the cwd below
+        ],
+        capture_output=True,
+        text=True,
+        cwd=str(godot_project.parent),
+    )
+
+    # No skip: pack must RUN to completion regardless of template presence. Before
+    # the fix this returned 4 with an empty export_failed (project.godot not found).
+    assert run.returncode == 0, run.stdout + run.stderr
+    data = json.loads(run.stdout)
+    assert data["mode"] == "pack"
+    assert data["output_path"] == override_rel
+    # The .pck landed INSIDE the project — the relative --project resolved to the
+    # same directory an absolute one would, exactly as the editor resolves it.
+    assert artifact.exists(), f"expected .pck at {artifact} for a relative --project"
+
+
+@pytest.mark.e2e
 def test_export_run_pack_omits_installed_harness_and_restores_it(godot_project):
     # ADR-0018: `gda export run` must NEVER carry the dev-only harness into the
     # artifact — and without the developer having to `gda daemon uninstall` first.
--- a/tests/test_export_runner.py
+++ b/tests/test_export_runner.py
@@ -62,6 +62,64 @@ def test_export_runs_with_project_as_cwd(monkeypatch):
     assert rec.cmd[-3:] == ["--export-release", "Linux/X11", "build/game.x86_64"]
 
 
+def _path_arg(cmd: list[str]) -> str:
+    """The value following ``--path`` in a spawned argv."""
+    return cmd[cmd.index("--path") + 1]
+
+
+def test_relative_project_is_absolutized_so_cwd_and_path_agree(monkeypatch):
+    # Regression for #344: this channel spawns with cwd = <project>, so a RELATIVE
+    # --project passed as --path verbatim would be re-resolved by the engine
+    # against the child CWD (<project>/<project>) — project.godot not found →
+    # empty export_failed. The runner must absolutize the project so --path and
+    # cwd name the SAME absolute directory (reaching the resolution an absolute
+    # --project reaches).
+    rec = _RecordingRun()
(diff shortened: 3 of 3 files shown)


qinhaihong-red:
## Review Skill: Standards / Spec

Fixed point: `origin/main...HEAD` (merge-base `0b89cf3e92d015b21e6c106f1e6a25115e752370`; head `ce4e83504d5a78987852b6966c0edb7c8df84a74`).

## Standards

No documented Standards violations found.

Judgement call checked, not a violation: `src/gda/export_runner.py:92` now absolutizes `project` inside the native export runner. ADR-0006 says path normalization happens once at the CLI layer, but this hunk does not re-select or re-validate the project; it converts the already-resolved project into an absolute spawn representation so `cwd` and `--path` still name the same directory. That matches ADR-0010's export-specific native CLI path, where only the irreducible native export step diverges from the default runner, and it preserves ADR-0028's pre-native-run harness strip/restore ordering because `run_export_operation` still snapshots/restores around the native run.

The added tests in `tests/test_export_runner.py` and `tests/test_e2e_export_run.py` align with the relevant documented standards: ADR-0006 requires deterministic project context, and ADR-0010 requires `pack` to avoid template gating and classify native export outcomes without stderr-code parsing. Tool-enforced Ruff/Pyright concerns were skipped here by design.

## Spec

Required change: the PR does not satisfy the third acceptance criterion from #344:

> A missing/invalid project yields a structured, non-empty diagnostic rather than an empty `export_failed`.

The core relative-`--project` behavior is fixed: `src/gda/export_runner.py:92` makes `cwd` and `--path` use the same absolute project path, and `tests/test_e2e_export_run.py:239` adds a real relative-`--project` pack regression test. However, invalid project input still does not produce a structured JSON error. `_export_run_recipe` calls `resolve_project_dir(project)` directly at `src/gda/cli.py:547`; invalid projects raise `ValueError` from `src/gda/project.py:35`, and `_dispatch_recipe` does not convert that into a `GdaError` envelope at `src/gda/cli.py:669`.

Repro on this PR head:

```bash
env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache \
  uv run python -m gda export run \
  --preset Linux/X11 \
  --mode pack \
  --project /tmp/definitely-not-a-godot-project \
  --json
```

Actual result: exit 1 with a Rich/Python traceback and no JSON envelope. Since this PR closes #344, either this acceptance criterion should be implemented with a regression test, or #344/this PR should explicitly narrow the scope so the missing/invalid-project diagnostic requirement is not claimed as complete.

Validation run:

- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run pytest tests/ -k export -rs -m "not e2e"` -> 83 passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run pytest tests/test_e2e_export_run.py::test_export_run_pack_accepts_a_relative_project -rs -p no:xdist` -> 1 passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run ruff check src/gda/export_runner.py tests/test_export_runner.py tests/test_e2e_export_run.py` -> passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run ruff format --check src/gda/export_runner.py tests/test_export_runner.py tests/test_e2e_export_run.py` -> passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run pyright src/gda/export_runner.py tests/test_export_runner.py tests/test_e2e_export_run.py` -> 0 errors

Summary: Standards 0 findings; Spec 1 required finding. Worst issue: #344's missing/invalid-project structured diagnostic acceptance criterion is still unmet.

qinhaihong-red:
## Code Review and Quality

### Findings

Required: #344's invalid-project acceptance path is still not structured.

Issue #344 requires: "A missing/invalid project yields a structured, non-empty diagnostic rather than an empty `export_failed`." The PR fixes the valid relative-`--project` case, but invalid project input still bypasses the public `GdaError` envelope. In this path, `_export_run_recipe` calls `resolve_project_dir(project)` at `src/gda/cli.py:547`; `resolve_project_dir()` raises `ValueError` from `src/gda/project.py:35`; `_dispatch_recipe` at `src/gda/cli.py:669` does not catch it and emit a structured failure. Repro on `ce4e835`:

```bash
env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache \
  uv run python -m gda export run \
  --preset Linux/X11 \
  --mode pack \
  --project /tmp/definitely-not-a-godot-project \
  --json
```

Actual result: exit 1 with a Rich/Python traceback, no JSON `{ "error": ... }`, and no structured non-empty diagnostics. Because the PR body says `Fixes #344`, this should either be implemented and tested before merge, or the issue/PR scope should be narrowed so this acceptance criterion is not treated as completed.

### Five-axis review

- Correctness: the core relative-`--project` fix is sound. `SubprocessExportRunner.run()` now uses the same absolute project for `cwd` and `--path`, and the new e2e pack test proves the real native export succeeds from the project parent with a relative `--project`. The remaining blocker is the unmet invalid-project diagnostic requirement above.
- Readability/Simplicity: the implementation is small and localized. The comments are longer than usual, but here they document a subtle Godot `cwd`/`--path` interaction and match the existing style in this export surface.
- Architecture: this keeps the native export runner seam intact and does not move classification out of `gda.errors`. Using an absolute spawn representation inside the export runner is reasonable because it does not re-resolve or re-validate project context.
- Security: no new secrets, auth surface, or untrusted-data execution surface introduced beyond the existing trusted-project model.
- Performance: no meaningful performance risk; one `Path.absolute()` per export run is negligible.

### Verification

I ran:

- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run pytest tests/ -k export -rs -m "not e2e"` -> 83 passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run pytest tests/test_e2e_export_run.py::test_export_run_pack_accepts_a_relative_project -rs -p no:xdist` -> 1 passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run ruff check src/gda/export_runner.py tests/test_export_runner.py tests/test_e2e_export_run.py` -> passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run ruff format --check src/gda/export_runner.py tests/test_export_runner.py tests/test_e2e_export_run.py` -> passed
- `env UV_CACHE_DIR=/tmp/godot-agent-pr348-uv-cache uv run pyright src/gda/export_runner.py tests/test_export_runner.py tests/test_e2e_export_run.py` -> 0 errors
- Invalid-project repro above -> still traceback / no structured JSON error

### Verdict

Request changes: the implementation is otherwise healthy, but the PR should not close #344 while one of its acceptance criteria is still unmet.

qinhaihong-red:
## Re: the AC#3 finding — root cause is cross-cutting, split to #353

Thanks — the finding is valid, and I dug into it. The core fix here (relative-`--project` double-application, AC#1) and its regression test (AC#2) stand. But the invalid-project traceback you reproduced is **not** an export-run / recipe-channel issue — it's a **pre-existing gap in every channel**.

Reproduced on `main` (before this PR):

```
gda scene list --project /tmp/not-a-project --json                        # SENTINEL → ValueError traceback
gda export run --preset X --mode pack --project /tmp/not-a-project --json  # RECIPE   → ValueError traceback
```

`resolve_project_dir()` raises `ValueError` for an explicit `--project` that isn't a Godot project (`src/gda/project.py:35,61`), and **no** channel — sentinel `_dispatch` nor recipe `_dispatch_recipe` — converts it into a `GdaError` envelope. So `scene list`, `node add`, etc. traceback exactly the same way.

Given that, patching only `_export_run_recipe` would be a band-aid: it would leave every other `--project` command still tracebacking, and it would duplicate the per-command handling #343 already added in `script_run.py` (`ValueError → project_not_found`). The right fix is a single shared conversion, folding #343's handling in — which is a coordinated cross-cutting change, best done on its own after #343 lands.

**So I've split AC#3's general form into #353** ("an invalid `--project` raises an uncaught `ValueError` instead of a structured `GdaError`, across all channels"), blocked on #343. This PR stays scoped to the relative-`--project` fix it's named for (AC#1 + AC#2), which is complete and CI-green. #344's AC#3 is tracked there rather than claimed complete here.

qinhaihong-red:
## Follow-up Review

I rechecked the proposed resolution for the AC#3 finding.

The split itself looks technically reasonable. I verified the author's key claim: on `origin/main`, invalid `--project` input already tracebacks in both a sentinel command and a recipe command, e.g.:

```bash
uv run python -m gda scene list --project /tmp/not-a-project --json
uv run python -m gda export run --preset Linux/X11 --mode pack --project /tmp/not-a-project --json
```

Both fail with an uncaught `ValueError` rather than a structured JSON error. So AC#3 is not caused by this PR's export-run change; it is a pre-existing cross-cutting gap in shared project resolution / dispatch. Tracking the general fix in #353 is the right shape: #353 covers sentinel + recipe commands, empty `--project`, a single shared conversion layer, error-code / ADR-0002 consistency, and regression tests. That is better than band-aiding only `_export_run_recipe`.

One remaining issue: the GitHub closure semantics are still not clean. This PR body still says `Fixes #344`, while #344's issue body still includes AC#3. If this PR merges as-is, GitHub will close #344 even though part of the issue's acceptance criteria has moved to #353. The explanatory PR/issue comments help, but they do not change the automatic-close semantics.

Recommendation: keep the #353 split, but also update the tracked scope metadata before merge. For example:

- change this PR footer from `Fixes #344` to `Refs #344`, or otherwise avoid automatically closing #344 until the issue body is reconciled; and/or
- edit #344's body to mark AC#3 as moved to #353, leaving #348 responsible only for AC#1 + AC#2.

After that metadata cleanup, I would consider the original AC#3 review finding reasonably addressed by scope split rather than by code in this PR.

qinhaihong-red:
## Metadata reconciled — closure semantics now clean

Thanks for the recheck and for verifying the cross-cutting root cause. Addressed the remaining closure-semantics point:

- **Edited #344's body**: AC#1 + AC#2 marked delivered; AC#3 struck through with **→ Split to #353**, plus a dated scope-update note. So #344's remaining scope is exactly what this PR delivers.
- **Kept `Fixes #344`**: now correct — with AC#3 relocated in the issue body, auto-closing #344 on merge is honest (its remaining ACs, AC#1 + AC#2, are complete). #353 (blocked on #343) owns the general shared-project-resolution fix.

No code change to this PR; it remains the relative-`--project` fix, CI-green. #353 carries the general `invalid --project → structured GdaError` work across all channels.

PR #349, opened 2026-07-01
feat: `gda script run` — user-script passthrough execution (ADR-0031)

qinhaihong-red:
Closes #343. Follows **ADR-0031** (the third execution shape: a user-script passthrough run).

## What

`gda script run res://path.gd [--project …]` runs the user's own script as a one-shot `godot --headless --path <project> --script <res://…>` and returns a structured result. The outcome **bifurcates by whose failure it is**:

- **Clean engine exit (`exit_code >= 0`)** → SUCCESS `{exit_status, stdout, stderr}` passed through **verbatim, even when `exit_status != 0`**. This is the ONLY gda command whose success result can carry a non-zero `exit_status` — a deliberate `quit(1)` (e.g. an assertion-failed logic-seam test) is *data the agent reads*, not a gda failure.
- **gda-/engine-level failure** (binary not launchable / timeout / signal crash, `exit_code < 0`) → Error envelope via the **shared `classify_launch_or_crash`** already used by `export run`, reusing the existing codes `binary_not_found` / `launch_timeout` / `engine_crashed`. No new GDScript-mirrored codes.
- **non-`res://`/absolute path** → `invalid_path`; **no resolved project** → `project_not_found` — both structured, decided *before* any launch.

## How (mirrors `export run` as the template)

- `ExecutionKind.SCRIPT_RUN` — the fourth `kind`, **self-description only** (dispatch routes by `recipe` per ADR-0023, so it adds no runner-selection branch). Moves as one unit with the `--schema`/manifest `kind` enum and the schema tests that pin the set.
- `ScriptRunResult` — the **thin public promotion of the internal Raw run** (`gda.runner.RunResult`): built by dropping `launch_failure` (→ the Error envelope) and renaming `exit_code`→`exit_status`. A code comment notes a 2nd passthrough consumer should later promote this to a shared `RawRunResult` — not built now (one consumer today; `export run` returns a different domain shape).
- `src/gda/script_run.py` — the recipe: validate `res://` + require a resolved project, then **reuse the deep module `gda.runner.launch()`** for all spawn/timeout/UTF-8-decode/launch-failure normalization (`cwd=None`, mirroring the sentinel `SubprocessGodotRunner` — not the export runner's `cwd=project`).
- `gda script run --schema` self-describes the success result + the uniform error envelope (ADR-0004); `surface.py` carries `kind` automatically; **no `mcp/server.py` edit** needed.

## Tests

- `tests/test_script_run_operation.py` — engine-free bifurcation via an injected launch seam.
[...]

PR #350, opened 2026-07-01
fix(daemon): session-launch diagnostics + typed live_windowed_unavailable (#345)

### LINKED ISSUE #345: gda: session-launch failure reports generic engine_session_not_running with empty diagnostics
## Problem

When a live **Engine session** fails to launch, `gda` reports a generic `engine_session_not_running` with an **empty `diagnostics`**, even when the failure is deterministic and the engine already wrote a crash trace. The caller can't root-cause from the gda result alone.

Concrete repro (the trigger that surfaced this): `gda daemon start --windowed` then a first live op (`gda screen capture`) on a **macOS host without an active Aqua GUI session** (SSH / CI / sandbox). The windowed Godot process aborts (`SIGABRT`) in `NSApplication` init / window-server registration — there is no usable DisplayServer, and macOS has no `xvfb` equivalent. gda returns:

```json
{"error":{"category":"live","code":"engine_session_not_running",
 "message":"the gda-daemon could not launch an engine session (no Godot binary, or the harness did not connect)",
 "diagnostics":""}}
```

> Note: the *crash itself* is not a gda bug — a windowed Godot genuinely cannot run without a window server (a Godot/macOS platform limit). This issue is purely about gda's **error reporting / classification quality** for that failure.

## Two gaps

1. **Empty `diagnostics`.** The child died with a clear signal (`SIGABRT`) and Godot wrote a crash report, but gda surfaces none of it — you must read the OS crash log to understand why. gda already owns a daemon-owned Session log + diagnostics surface (ADR-0022), so the information exists; it just isn't carried into the error envelope. (Same "empty diagnostics" class already noted on the `export run` bug, #344.)
2. **Overloaded code.** `engine_session_not_running` conflates "no Godot binary", "harness didn't connect", and "windowed session can't come up because the host has no usable DisplayServer". gda already has a typed `live_display_unavailable` (returned when a windowed capture is requested on a *headless* session); the host-has-no-display case has no typed code, so callers must string-match a generic one.

## Suggested direction

- On a session-launch failure, populate `diagnostics` with the child's exit/abort signal and the tail of the Session log.
- Give the "windowed session unavailable because the host has no usable DisplayServer" case a typed code (extend `live_display_unavailable`, or add `windowed_session_unavailable`) so a caller can branch/skip cleanly.

## Acceptance criteria

- [ ] A session-launch failure returns a non-empty `diagnostics` that names the proximate cause (exit/abort signal and/or Session-log tail).
- [ ] A windowed launch that fails for lack of a host DisplayServer is reported with a typed, documented code distinct from the "no binary / harness didn't connect" cases.
- [ ] A caller can branch on the typed code without string-matching prose or reading an OS crash report.

## Found while

Building the Panda Adventure S0 walking skeleton (#329, PR #341) — gda dogfooding. The game's windowed screenshot e2e currently keys its environment-skip on the generic `engine_session_not_running` code precisely because no typed display code exists for this case.

qinhaihong-red:
Closes #345.

When a live Engine session fails to launch, the daemon returned a generic
`engine_session_not_running` with EMPTY `diagnostics`, even when the child died
deterministically — and that one code conflated three causes, one being "a
windowed session can't come up because the host has no usable DisplayServer".
Delivered in two sequenced parts. No new ADR (this applies three existing
patterns).

## Part A — thread diagnostics into the live Error envelope

- `launch_session` now polls the child at each failure boundary BEFORE terminating
  it (the child is spawned with `stderr=DEVNULL`, so its liveness is the one cause
  signal we keep): a child already dead by a negative return code names its signal
  (a windowed-no-DisplayServer abort is `SIGABRT`); a child still alive is the
  "engine up, harness never connected" case. This distinguishes a crashed windowed
  process from a hung harness.
- The daemon best-effort tails the daemon-owned Session log via the DETERMINISTIC
  `server._session_log_path()` (no live session object required — extends ADR-0022's
  read path to the failed-launch case) and threads the combined detail into the
  `engine_session_not_running` reply.
- The detail rides the **existing** reply `stderr` → `RunResult.stderr` →
  `GdaError.diagnostics` path. **The wire envelope shape `{stdout, stderr, exit_code}`
  / `{error:{code,message}}` is unchanged** — staying within ADR-0002's
  advisory-stderr carve-out. Out of scope: redirecting the child stderr to a pipe
  (deadlock risk).

## Part B — typed code + pre-launch host-display precondition

- New error code **`live_windowed_unavailable`** (category **environment**, exit
  **127**, source **classifier**, NOT GDScript-mirrored), mirroring
  `live_unsupported_platform`. Registered in `error_codes.py` **with** the matching
  ADR-0002 registry row in the same commit (the registry drift test needs code +
  table together); deliberately not added to the LIVE-category tuples.
- New `gda.display` module — a platform-dispatched host-display probe reusable by
  `src` and both test suites: macOS `CGSessionCopyCurrentDictionary` (an on-console
  GUI session — NOT `launchctl managername`, which lies "Aqua" on headless macOS),
  Linux `$DISPLAY` / `$WAYLAND_DISPLAY`, and a documented Windows stub seam (live is
  UNIX-only today). Adapted from the ad-hoc probe in the panda-adventure screenshot
  e2e.
- `run_daemon_start_operation` refuses a windowed start pre-spawn (and
  pre-harness-install) when no display is usable, via an injectable `display_check`
  seam.
- Reuse the helper to fix the windowed e2e gates: `tests/test_e2e_screen.py` now
  skips correctly on headless macOS too (not just Linux without `$DISPLAY`), before
  spawning Godot; the example `test_e2e_screenshot.py` keys on
  `live_windowed_unavailable`.

## Doc artifacts (land with the code)

- **ADR-0002**: the `live_windowed_unavailable` registry row (same commit as
  `error_codes.py`).
- **ADR-0022**: a dated `> Outcome` note — the Session-log read path now also serves
  the failed-launch case via the deterministic path (point-in-time decision prose
  preserved).

## Tests

- `uv run pytest tests/test_daemon_server.py tests/test_daemon_diag.py tests/test_error_registry.py -rs` → **42 passed**. Diagnostics are now non-empty and distinguish "child died (signal X)" vs "harness did not connect"; the drift test passes with the new row.
- A windowed launch on a host with no usable DisplayServer returns
  `live_windowed_unavailable` (environment/127) WITHOUT spawning Godot — added
  (`test_windowed_start_without_a_display_is_live_windowed_unavailable`).
- `tests/test_e2e_screen.py` skips the windowed tests on a headless host before any
  spawn (the `skipif` mark resolves at collection; verified).
- Full non-e2e suite: **991 passed**. `ruff` and the `pyright` gate (ADR-0030) both
  clean.

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -160,6 +160,7 @@ operation, and parse codes the CLI assigns).
 | `live_invalid_event_spec` | `live` | `classifier` | `6` | A live input sequence event has a type the harness does not recognize (Phase 2, #221). |
 | `live_display_unavailable` | `live` | `classifier` | `6` | A live `screen` capture ran on a headless engine session (the dummy DisplayServer cannot read pixels); start the daemon windowed with `gda daemon start --windowed` (Phase 2, #222). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
+| `live_windowed_unavailable` | `environment` | `classifier` | `127` | A windowed live session (`gda daemon start --windowed`) was requested but the host has no usable DisplayServer (no on-console GUI session / no `$DISPLAY`), so the session cannot come up; refused before spawning Godot (Phase 2, #345). |
 
 ## Considered options
 
--- a/docs/adr/0022-runtime-diagnostics-via-daemon-owned-session-log.md
+++ b/docs/adr/0022-runtime-diagnostics-via-daemon-owned-session-log.md
@@ -113,4 +113,10 @@ parse.
 > the `engine_session_not_running` error's advisory `diagnostics`. NOTE: a
 > windowed-no-`DisplayServer` abort happens BEFORE Godot installs its file logger, so
 > for that case the tail is usually empty (the child-signal reason carries it); it
-> carries content for a post-logger crash. The decision above is otherwise unchanged.
+> carries content for a post-logger crash. To keep that tail HONEST, `launch_session`
+> now truncates the Session log at the START of each launch attempt, before spawning —
+> so "truncated each launch" holds even for a pre-logger abort (Godot's own
+> `FileAccess::WRITE` truncation only happens once its logger installs, which a
+> pre-logger crash never reaches). A failed pre-logger launch therefore reads EMPTY
+> rather than a PREVIOUS session's stale output; a post-logger failure reads only the
+> current session. The decision above is otherwise unchanged.
--- a/docs/adr/0022-runtime-diagnostics-via-daemon-owned-session-log.md
+++ b/docs/adr/0022-runtime-diagnostics-via-daemon-owned-session-log.md
@@ -102,3 +102,15 @@ parse.
   project's own semantics — so the ADR-0009 / ADR-0017 trust and editor-context
   boundaries are unchanged: `diag` introspects the running game's reported output,
   not the editor.
+
+> **Outcome (2026-07-01, #345) — the read path also serves the FAILED-LAUNCH case.**
+> The `diag` / `logger tail` read path above observes an already-launched session by
+> reading `session.log_file` off the remembered session object. #345 reuses the SAME
+> daemon-owned Session log for a session that never came up: when `launch_session`
+> returns no session, the daemon best-effort tails that log via the DETERMINISTIC
+> path (`server._session_log_path()`, computed from the project slug — no live
+> session object required) and threads it, alongside the child-liveness reason, into
+> the `engine_session_not_running` error's advisory `diagnostics`. NOTE: a
+> windowed-no-`DisplayServer` abort happens BEFORE Godot installs its file logger, so
+> for that case the tail is usually empty (the child-signal reason carries it); it
+> carries content for a post-logger crash. The decision above is otherwise unchanged.
--- a/examples/platformer/panda-adventure/tests/test_e2e_screenshot.py
+++ b/examples/platformer/panda-adventure/tests/test_e2e_screenshot.py
@@ -17,7 +17,6 @@ an active window-server session) and runs for real on a genuine desktop.
 
 from __future__ import annotations
 
-import ctypes
 import json
 import os
 import shutil
@@ -28,6 +27,7 @@ import sys
 import pytest
 
 from gda.binary import resolve_godot_binary
+from gda.display import windowed_unavailable_reason
 
 import build_config
 
@@ -46,9 +46,12 @@ _COPY_IGNORE = shutil.ignore_patterns(
     "tests", ".godot", "build", "generated", "__pycache__"
 )
 
-# Daemon error codes that mean "this environment cannot show a window" (the live
-# session never came up / has no display) — a skip signal, not a test failure.
-_NO_DISPLAY_CODES = {"engine_session_not_running", "live_display_unavailable"}
+# Daemon error codes that mean "this environment cannot show a window" (the windowed
+# start was refused pre-launch, or the live session has no display) — a skip signal,
+# not a test failure. `live_windowed_unavailable` is the typed pre-launch refusal gda
+# now returns for `daemon start --windowed` on a display-less host (#345), replacing
+# the generic `engine_session_not_running` this used to key on.
+_NO_DISPLAY_CODES = {"live_windowed_unavailable", "live_display_unavailable"}
 
 
 def _error_code(stdout: str) -> str | None:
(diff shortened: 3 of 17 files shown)


qinhaihong-red:
## Review pass 1: Standards / Spec

Reviewed PR #350 against `origin/main...HEAD` (`6ef389f`) using the `$review` split:

- Standards sources checked: `AGENTS.md`, `CLAUDE.md`, `CONTEXT.md`, agent docs, ADR-0002, ADR-0017, ADR-0021, ADR-0022, ADR-0029/0030, and `pyproject.toml`.
- Spec source: issue #345 and its triage comment.
- Local verification: `env UV_CACHE_DIR=/private/tmp/godot-agent-pr350.ePYwoL/uv-cache uv run pytest tests/test_daemon_server.py tests/test_daemon_diag.py tests/test_daemon_lifecycle_payload.py tests/test_daemon_ops.py tests/test_error_registry.py -rs` -> `77 passed`.

### Standards

- Required: `src/gda/daemon/server.py` can attach stale session-log contents to a failed-launch diagnostic. `_launch_failure_diagnostics()` always reads `_session_log_path()` after `launch_session()` returns `None`, but ADR-0022's session-log contract depends on Godot's `--log-file` truncating the file on each launch so diagnostics are bound to the current Engine session. This PR also notes that a windowed no-DisplayServer abort can happen before Godot installs its file logger. In that case the deterministic log path may still contain the previous session's log, and `_read_session_log_tail()` will report stale output as if it belonged to the failed launch. Please clear/replace the log before each attempted launch, or only include the tail when it is known to belong to the attempted launch.

### Spec

- Required: `live_windowed_unavailable` is not enforced at the actual lazy Engine session launch boundary. Issue #345 says the pre-launch host-display check should run "when launching a session with windowed requested" and calls the "Guard location" authoritative at session-launch, while `daemon start --windowed` fail-fast is optional. The diff puts the only check in `run_daemon_start_operation()` (`src/gda/daemon_ops.py`), before spawning the daemon process. The actual lazy Godot spawn remains `DaemonServer._ensure_session()` -> `launch_session()` (`src/gda/daemon/server.py`, `src/gda/daemon/session.py`) with no display precondition, so a windowed daemon that reaches that path can still spawn Godot and then return generic `engine_session_not_running` rather than the typed environment/127 code. Please move or duplicate the authoritative check to the session-launch boundary, and add a regression proving it returns `live_windowed_unavailable` without calling `launch_session()`/spawning Godot there. If that typed code is returned over the live daemon reply path, it will also need to be accepted by `classify_live` like `live_unsupported_platform`.

Summary: 1 Standards finding, 1 Spec finding. Worst issue: the typed no-display precondition is missing from the authoritative session-launch boundary requested by #345.

qinhaihong-red:
## Review pass 2: Code Review and Quality

Verdict: request changes.

### Findings

1. Required (correctness / architecture): the typed no-display refusal is implemented at the daemon-start recipe, not at the component that actually launches the Engine session. `src/gda/daemon_ops.py` refuses `daemon start --windowed`, but `DaemonServer._ensure_session()` still calls `launch_session()` directly and `launch_session()` still reaches `subprocess.Popen()` with `windowed=True` before any host-display precondition. That means the real session-launch owner can still spawn a doomed windowed Godot and report generic `engine_session_not_running` instead of `live_windowed_unavailable`. This also leaves the tests aimed at the wrong boundary: `tests/test_daemon_lifecycle_payload.py` proves the daemon-start seam, but there is no regression proving a windowed `DaemonServer` refuses before calling `launch_session()`. Please put an injectable display check at the daemon/session launch boundary and add a test that fails if `launch_session()` is called when the display check returns unavailable. If the fix returns `live_windowed_unavailable` through the daemon/live reply path, add it to `_LIVE_CLIENT_CODES` in `src/gda/errors.py`; today only `live_unsupported_platform` is whitelisted as an environment-category live client code.

2. Required (correctness): failed-launch diagnostics can include stale session-log output. `_launch_failure_diagnostics()` always reads the deterministic session log path after a failed launch, but the PR's own comments say the no-DisplayServer abort can happen before Godot installs the file logger. If a previous session wrote that same path, the failed launch may not truncate it, and `_read_session_log_tail()` will attach the previous session's log to the current failure. That undermines the goal of making diagnostics trustworthy. Please clear/truncate the log before each attempted launch, or gate the tail on proof that it belongs to the current attempt; add a regression with a pre-existing log file and a failed pre-logger launch.

### Other Axes

- Readability: the new helper and diagnostics path are understandable, though the comments are carrying a lot of design intent because the ownership boundary is still split.
- Security: no new security issue found; the display probe uses fixed OS APIs/env vars and does not introduce user-controlled execution.
- Performance: no material performance issue found in the reviewed paths.
- Verification: local focused suite passed (`77 passed`), and PR checks are green. The missing coverage is specifically the actual session-launch no-display boundary and stale-log case above.

qinhaihong-red:
## Both Required findings addressed — `1ed515c`

Thanks for the two-pass review. Both fixed and pushed; CI green (ruff / unit / pyright), MERGEABLE/CLEAN. You were right on both counts — the guard was at the optional boundary, not the authoritative one.

### 1. No-display guard moved to the authoritative session-launch boundary
The typed refusal now lives where the lazy spawn actually happens — `DaemonServer._ensure_session`, not only the optional `daemon start` fail-fast:
- `DaemonServer` gained an injectable `display_check` ctor param (defaults to `gda.display.windowed_unavailable_reason`), mirroring the daemon_ops seam.
- New typed `WindowedDisplayUnavailable` (mirrors `SceneMismatch`): `_ensure_session` raises it **before `launch_session` is ever called** when `self.windowed` and the display check returns a reason; `_handle` catches it beside the `SceneMismatch` catch and returns `error_reply("live_windowed_unavailable", …, diagnostics=reason)`.
- `live_windowed_unavailable` added to `_LIVE_CLIENT_CODES` in `errors.py` — it now flows through `classify_live` to its environment/127 Failure instead of misrouting.
- Kept the `daemon start --windowed` check as the optional fail-fast (defense-in-depth).
- Regression: `test_windowed_no_display_is_live_windowed_unavailable_without_launching` patches `launch_session` to raise if invoked — it never is; `test_live_windowed_unavailable_flows_through_classify_live` asserts `_LIVE_CLIENT_CODES` membership + environment/127 resolution.

### 2. Failed-launch diagnostics can no longer be stale
`launch_session` now truncates the Session log (`log_file.write_bytes(b"")`, best-effort) at the **start of each attempt, before `subprocess.Popen`** — so a pre-logger abort (which never reaches Godot's own `--log-file` truncation) reads EMPTY instead of a previous session's tail; a post-logger failure reads only the current session. Consistent with ADR-0022's "truncated each launch" (the outcome note documents the pre-launch truncation). Regression: `test_failed_launch_diagnostics_excludes_stale_session_log` pre-writes stale content, drives a real pre-logger SIGABRT launch, and asserts the stale string is absent.

### Verification (orchestrator, independent)
- Ran the 3 new regressions from the branch → 3 passed.
- Daemon suite `test_daemon_server / test_daemon_diag / test_daemon_ops / test_daemon_lifecycle_payload / test_error_registry` → 82 passed; full non-e2e → 996 passed; ruff/pyright clean.

_(Note: this fix now edits `errors.py` `_LIVE_CLIENT_CODES` — a different region (~L398) from PR #349's failure constructors (~L627), so the two auto-merge cleanly.)_

PR #352, opened 2026-07-01
docs(skill): deep-module guardrail for subagent-worktree-parallel (#351)

### LINKED ISSUE #351: docs(skill): add a deep-module guardrail to subagent-worktree-parallel (disjoint ≠ architecture goal)
## What to build

Add a reverse guardrail to the `subagent-worktree-parallel` skill. The skill today emphasizes *"identify append hotspots, keep slices file-disjoint"* but lacks the counter-principle: **file-disjointness is a merge-cost heuristic, not an architecture goal.** When keeping slices disjoint would conflict with a sound design decision, the design wins.

**Why (exposed during the #342–#345 parallel wave):** while planning that fan-out, the goal of "keep slices disjoint" once drove an architecture/scope framing — e.g. "deliberately avoid touching `models.py` to keep slices decoupled", and treating #343's public result model as newly-invented rather than first confirming it reuses the deep module `launch()` / `RunResult`. Both were corrected during planning, but the skill should encode the lesson so it does not recur.

**The principle (two facets):**
1. Never let "keep slices disjoint / avoid the append hotspot" suppress or distort a sound architecture/scope decision — a legitimate shared-module edit, deep-module reuse, or a single source for a public shape. **Serialize that slice's merge instead** of sacrificing reusability.
2. Slice boundaries must not tempt each subagent to **re-implement / duplicate deep-module logic in isolation** — worktree isolation hides the duplication until merge/review. Bake *"did this slice reuse the existing deep module?"* into subagent DoD / review.

This complements (does not contradict) REFERENCE §8 *"kill the hotspots"*: §8 is about splitting shared files so slices stop colliding; this guardrail is about not degrading a deep module for the sake of that disjointness in the first place.

## Acceptance criteria

- [ ] `SKILL.md` "Core principles (non-negotiable)" carries a one-line statement of the guardrail (disjoint ≠ architecture goal; serialize the merge rather than sacrifice reusability).
- [ ] `REFERENCE.md` carries the detail (both facets), placed near §1 dependency analysis or §8 architectural fix, and explicitly framed as complementary to §8.
- [ ] The addition does not contradict the existing §8 guidance.

## Blocked by

None - can start immediately. (Being implemented by the main agent immediately; decoupled from the #342–#345 feature slices.)

qinhaihong-red:
## Summary

Adds the reverse guardrail the `subagent-worktree-parallel` skill was missing: **file-disjointness is a merge-cost heuristic, not an architecture goal.** When keeping slices disjoint would suppress or distort a sound design decision, the design wins — serialize that slice's merge instead of degrading the architecture.

Two facets:
1. Never let "keep slices disjoint / avoid the append hotspot" suppress a legitimate shared-module edit, deep-module *reuse*, or a single source for a public shape.
2. Worktree isolation hides re-implemented/forked deep-module logic until merge/review — bake "did this slice reuse the existing deep module?" into subagent DoD.

- `SKILL.md`: one line under "Core principles (non-negotiable)".
- `REFERENCE.md`: detail in §1, explicitly framed as the **complement of §8** ("kill the hotspots"), not a contradiction.

## Why

Exposed while planning the #342–#345 parallel wave: "keep slices disjoint" briefly drove an architecture/scope framing (avoiding `models.py`; treating a slice's result model as newly-invented rather than a thin projection of the deep module `launch()`/`RunResult`). Both were corrected in planning; this encodes the lesson.

Closes #351

### DIFF
--- a/.agents/skills/subagent-worktree-parallel/REFERENCE.md
+++ b/.agents/skills/subagent-worktree-parallel/REFERENCE.md
@@ -29,6 +29,28 @@ edits, where all the integration cost concentrates:
 A wave whose slices all hammer the same hotspot is *not* a good parallel candidate —
 expect a conflict-heavy serial merge, or fix it structurally first (§8).
 
+**Guardrail: disjointness is a merge-cost heuristic, not an architecture goal.** When
+slicing for disjointness would suppress or distort a sound design decision, the design
+wins:
+
+- **Don't avoid a shared file just to keep a slice disjoint.** If a legitimate change
+  belongs in a registry, a model module, or a *deep module*, put it there and
+  **serialize that slice's merge** — don't sacrifice reusability (or bend scope) to
+  manufacture disjointness. "I deliberately kept this out of `models.py` to decouple the
+  slices" is the smell.
+- **Don't let isolation hide a re-implemented deep module.** Worktree isolation defers
+  duplicated/forked logic to merge or review, so parallel fan-out actively tempts each
+  subagent to re-build shared logic in its own worktree. Before calling a slice's
+  addition "new", confirm it isn't a thin projection of an existing deep module — *reuse
+  the module*; a per-slice boundary DTO layered over it is fine, a second copy of its
+  logic is not. Bake "did this reuse the existing deep module?" into the DoD and the
+  review.
+
+This is the **complement of §8**: §8 *splits* hotspots so slices stop colliding; this
+guardrail stops you from *degrading* a deep module to fake disjointness in the first
+place. Splitting a shared file is a structural fix; suppressing a sound change to dodge
+it is not.
+
 ## 2. Wave sizing
 
 - **≤ ~5 independent slices per wave.** (A run that tried ~9 at once hit two failure
--- a/.agents/skills/subagent-worktree-parallel/REFERENCE.md
+++ b/.agents/skills/subagent-worktree-parallel/REFERENCE.md
@@ -34,7 +34,8 @@ slicing for disjointness would suppress or distort a sound design decision, the 
 wins:
 
 - **Don't avoid a shared file just to keep a slice disjoint.** If a legitimate change
-  belongs in a registry, a model module, or a *deep module*, put it there and
+  belongs in a registry, a model module, or a *deep module* (a unit with a small
+  interface over substantial implementation that is meant to be reused), put it there and
   **serialize that slice's merge** — don't sacrifice reusability (or bend scope) to
   manufacture disjointness. "I deliberately kept this out of `models.py` to decouple the
   slices" is the smell.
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Review: Standards / Spec

Reviewed PR #352 against `origin/main...HEAD` (`a0fcd19`) and issue #351.

## Standards

No hard Standards findings.

Judgement call: `.agents/skills/subagent-worktree-parallel/REFERENCE.md` now uses `deep module` and `boundary DTO` in project-agnostic Skill guidance, but neither term is defined in the Skill itself. The intent is understandable from context, and the new text is consistent with §8: §8 says to split append hotspots structurally, while this new guardrail says not to degrade or reimplement reusable module logic just to fake disjointness. Still, a one-sentence definition of `deep module`, or a more neutral phrase than `DTO`, would make the guidance less jargon-dependent for future agents.

## Spec

Finding:

- [P1] The new DoD/review guardrail is not wired into the executable templates. Issue #351 facet 2 explicitly says to bake `"did this slice reuse the existing deep module?"` into subagent DoD / review. The PR states that principle in `SKILL.md` and `REFERENCE.md` §1, but the reusable dispatch prompt in `REFERENCE.md` §3 still defines DoD only as the integration-tier test, and the §9 pre-launch checklist still lacks the deep-module reuse / reimplementation check. For a Skill, agents are most likely to copy those operational surfaces; leaving them unchanged makes the requirement look implemented while the actual prompt/checklist can still drive the old behavior. Please add the reuse check to the dispatch prompt DoD and/or §9 checklist.

No other spec gaps found. The core solution is reasonable: it states that disjointness is a merge-cost heuristic rather than an architecture goal, covers both requested facets, and frames the new rule as complementary to §8 rather than contradictory.

Summary: Standards: 0 hard findings, 1 non-blocking clarity note. Spec: 1 finding. Worst issue: the new guardrail is present in prose but missing from the Skill's operational prompt/checklist.

qinhaihong-red:
## Both points addressed — `8054056`

Thanks — good catch that the guardrail lived only in prose, not the surfaces agents actually copy.

### [P1] Guardrail wired into the operational surfaces
It's no longer prose-only. Baked the deep-module reuse check into the two places subagents actually execute (issue #351 facet 2):
- **§3 dispatch prompt** — added a Definition-of-Done bullet: *if your change belongs in a shared/deep module, REUSE it — don't re-implement its logic in your worktree; a thin per-slice wrapper is fine, a second copy is not; if the right change belongs in a shared file, make it there and flag it — don't dodge it to stay disjoint.* The "when done, report…" line now also asks the agent to report any deep module reused or shared file it had to touch.
- **§9 pre-launch checklist** — added: *"Each subagent's DoD includes deep-module reuse — no re-implementing shared logic, and no dodging a legitimate shared-file edit, to fake disjointness (§1)."*
- SKILL.md's core-principles bullet now points at §1, §3, and §9 (and says "wired into the dispatch-prompt DoD and the pre-launch checklist, not just prose").

### [clarity] Less jargon
- Glossed *deep module* at first use in §1: "a unit with a small interface over substantial implementation that is meant to be reused".
- Replaced "boundary DTO" with "a thin per-slice data shape (a DTO/wrapper)".

CI green; MERGEABLE/CLEAN.

PR #357, opened 2026-07-01
fix(cli): structured project_not_found for an invalid --project, all channels (#353)

### LINKED ISSUE #353: gda: an invalid --project raises an uncaught ValueError (traceback) instead of a structured GdaError, across all channels
## What to build

`resolve_project_dir()` raises `ValueError` when an explicit `--project` is empty or is not a Godot project (no `project.godot`) — `src/gda/project.py:35,61`. **No CLI channel converts that into a structured `GdaError` envelope**, so `gda <cmd> --project <bad> --json` prints a Rich/Python traceback and no `{"error": …}` JSON, violating the structured-output contract (ADR-0002). This affects BOTH the sentinel channel (`_dispatch`) and the recipe channel (`_dispatch_recipe` → each recipe's own `resolve_project_dir`).

Repro (on `main`):
```
gda scene list --project /tmp/not-a-project --json          # sentinel → ValueError traceback
gda export run --preset X --mode pack --project /tmp/not-a-project --json   # recipe → ValueError traceback
```

This is the **general form of #344's third acceptance criterion** ("a missing/invalid project yields a structured, non-empty diagnostic rather than an empty `export_failed`"). #344 (PR #348) fixed the relative-`--project` double-application — its core bug — and its regression test; AC#3's real content turned out to be this **cross-cutting** gap (not export-specific), split out here.

## Suggested direction

Convert the invalid-`--project` `ValueError` into a structured `GdaError` at a **single shared layer** (so every `--project` command benefits), not per-command.

- Note: #343 (`gda script run`, PR #349) already added its OWN per-command handling in `src/gda/script_run.py` (catches the `ValueError` → `project_not_found`). This issue should **fold that into the shared layer** once #343 merges, so the logic lives in one place (avoid duplicating it per recipe/command).
- Decide the error code: either generalize `project_not_found` (currently documented narrowly as "an operation that enumerates a project's res:// tree ran without a resolved Godot project", `src/gda/error_codes.py:164`) to cover "an explicit `--project` that is not a Godot project", or add a distinct `invalid_project` code. Either way update the ADR-0002 registry row + keep the drift test green.

## Acceptance criteria

- [ ] `gda <cmd> --project <not-a-project> --json` returns a structured `{"error": …}` envelope with a non-empty diagnostic and a non-zero exit — NOT a traceback — for BOTH a sentinel command and a recipe command.
- [ ] The empty-`--project` case is likewise structured.
- [ ] The conversion is handled ONCE at a shared layer; #343's per-command project handling in `script_run.py` is folded in (no duplicated logic).
- [ ] The error code is registry / ADR-0002 / drift-test consistent.
- [ ] Regression tests cover a sentinel and a recipe command.

## Blocked by

- #343 (PR #349) merging first, so the shared fix can subsume its per-command project handling rather than duplicate it.

## Found while

Reviewing PR #348 (#344) — the relative-`--project` fix. AC#3's general form was traced to this pre-existing cross-cutting gap.

qinhaihong-red:
## Summary

`resolve_project_dir()` raises `ValueError` for an explicit `--project`/`$GDA_PROJECT` that is empty or is not a Godot project. No channel converted it, so `gda <cmd> --project <bad> --json` leaked a Rich/Python traceback instead of a `{"error": …}` envelope — violating the ADR-0002 structured-output contract, on **every** `--project` command.

- New `_resolve_project_or_fail()` — the **one shared** project-resolution point on the CLI dispatch path (ADR-0006). It catches the `ValueError` and emits a structured `project_not_found` (`invalid_project_failure`).
- `_dispatch` (sentinel) and `_dispatch_recipe` (recipe) both route through it. `_dispatch_recipe` resolves **once** and passes the already-resolved project to the recipe.
- All 8 recipes (`export run`, `script run`, daemon ×4, screen ×2) now receive a resolved project and no longer resolve themselves. `script run`'s former per-command `ValueError` handling (#343) is **folded into the shared layer** — no duplicated logic.
- `project_not_found`'s registry + ADR-0002 `Meaning` generalized to the widened sense (drift test green; category/source/exit unchanged).

Before → after (`--json`):
```
gda scene list  --project /tmp/not-a-project --json   # sentinel: traceback → {"error":{"code":"project_not_found",…}}
gda export run  --project /tmp/not-a-project --json   # recipe:   traceback → structured
gda daemon status --project /tmp/not-a-project --json # (bonus)   traceback → structured
```

## Tests

New `tests/test_invalid_project_is_structured.py`: sentinel + recipe with an invalid `--project`, empty `--project`, and a bad `$GDA_PROJECT` all yield the structured envelope. #343's existing `script run` bad-project tests are the DRY regression guard (still green after removing its try/except). Full non-e2e suite: 1023 passed; ruff/pyright clean.

Closes #353

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -107,7 +107,7 @@ operation, and parse codes the CLI assigns).
 | `save_failed` | `operation` | `operation` | `4` | A scene could not be packed or saved. |
 | `delete_failed` | `operation` | `operation` | `4` | A file could not be removed from disk. |
 | `file_changed_externally` | `operation` | `operation` | `4` | A read-modify-write operation's target file changed on disk between the read and the write, so the write was refused to avoid clobbering the external edit. |
-| `project_not_found` | `operation` | `operation` | `4` | gda ran without a usable resolved Godot project: an operation needed one and none was resolved, or an explicit `--project`/`$GDA_PROJECT` was empty or is not a Godot project (no `project.godot`). |
+| `project_not_found` | `operation` | `operation` | `4` | gda ran without a usable resolved Godot project: an operation needed one and none was resolved, or an explicit `--project` was empty, or a `--project`/`$GDA_PROJECT` does not name a Godot project (no `project.godot`). |
 | `path_not_found` | `operation` | `operation` | `4` | A requested file does not exist. |
 | `not_a_scene` | `operation` | `operation` | `4` | A requested file cannot be loaded as a `PackedScene`. |
 | `parent_not_found` | `operation` | `operation` | `4` | A requested parent node path does not resolve to a node in the scene. |
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -107,7 +107,7 @@ operation, and parse codes the CLI assigns).
 | `save_failed` | `operation` | `operation` | `4` | A scene could not be packed or saved. |
 | `delete_failed` | `operation` | `operation` | `4` | A file could not be removed from disk. |
 | `file_changed_externally` | `operation` | `operation` | `4` | A read-modify-write operation's target file changed on disk between the read and the write, so the write was refused to avoid clobbering the external edit. |
-| `project_not_found` | `operation` | `operation` | `4` | An operation that enumerates a project's `res://` tree ran without a resolved Godot project. |
+| `project_not_found` | `operation` | `operation` | `4` | gda ran without a usable resolved Godot project: an operation needed one and none was resolved, or an explicit `--project`/`$GDA_PROJECT` was empty or is not a Godot project (no `project.godot`). |
 | `path_not_found` | `operation` | `operation` | `4` | A requested file does not exist. |
 | `not_a_scene` | `operation` | `operation` | `4` | A requested file cannot be loaded as a `PackedScene`. |
 | `parent_not_found` | `operation` | `operation` | `4` | A requested parent node path does not resolve to a node in the scene. |
--- a/docs/adr/0023-command-descriptor-single-registration.md
+++ b/docs/adr/0023-command-descriptor-single-registration.md
@@ -57,6 +57,16 @@ absorbs the two missing facts.**
   frozensets** and the per-command identity-branching inside the old daemon
   dispatch (each daemon command now carries its own recipe).
 
+  > **Outcome (2026-07-01, #353/#357):** project resolution moved OUT of the recipe
+  > into the shared `_dispatch_recipe` tail — it resolves once (via the same
+  > `_resolve_project_or_fail` the sentinel `_dispatch` uses) and hands the recipe an
+  > ALREADY-resolved project, so an invalid `--project` is a structured
+  > `project_not_found` on the recipe channel exactly as on the sentinel one, with no
+  > per-recipe resolution. A recipe now **produces the outcome from a resolved project**
+  > (it no longer resolves). A `projectless: bool` descriptor field excludes pure meta
+  > emitters (e.g. `gda skill`, ADR-0024) from resolution, so an inherited invalid
+  > `$GDA_PROJECT` cannot make a projectless meta command fail.
+
 **2. The render map and dispatch routing are projections of the descriptor, not
 parallel registries.** On the `cmd.emit` path the descriptor is already in hand, so
 rendering and channel selection read off `cmd` directly. Where a whole-surface view
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -716,16 +716,21 @@ def _dispatch_recipe(
     renders identically to a sentinel one; only outcome production differs. Shared by
     the argv bodies and the ``--params-json`` path, so the two forms are
     indistinguishable downstream (ADR-0015). Project resolution stays CLI-side
-    (ADR-0006) and happens HERE, once, for every recipe — so an invalid
-    ``--project`` yields the structured ``project_not_found`` envelope on this
-    channel exactly as on the sentinel one, and no recipe re-resolves (#353).
+    (ADR-0006) and happens HERE, once, for every PROJECT-USING recipe — so an
+    invalid ``--project`` yields the structured ``project_not_found`` envelope on
+    this channel exactly as on the sentinel one, and no recipe re-resolves (#353).
+    A ``projectless`` recipe (a pure meta emitter like ``gda skill``, ADR-0024) is
+    NOT resolved: it takes no project, so an inherited invalid ``$GDA_PROJECT``
+    must not make it fail (#357).
     """
     # A recipe command always carries a recipe channel — that is what routes it
-    # here rather than to the sentinel ``cmd.emit`` path (ADR-0023). The recipe
-    # receives the ALREADY-resolved project (or a structured project_not_found is
-    # emitted before it runs); it never touches ``resolve_project_dir`` itself.
+    # here rather than to the sentinel ``cmd.emit`` path (ADR-0023). A project-using
+    # recipe receives the ALREADY-resolved project (or a structured project_not_found
+    # is emitted before it runs); a projectless meta recipe receives None and never
+    # touches ``resolve_project_dir``.
     assert cmd.recipe is not None
-    outcome = cmd.recipe(params, project=_resolve_project_or_fail(project), godot=godot)
+    resolved = None if cmd.projectless else _resolve_project_or_fail(project)
+    outcome = cmd.recipe(params, project=resolved, godot=godot)
     if isinstance(outcome, Failure):
         emit_failure(outcome)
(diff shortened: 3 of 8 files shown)


qinhaihong-red:
## Review pass: `review` (Standards / Spec)

### Standards

Required: `_dispatch_recipe()` now resolves project context for every recipe before invoking it (`src/gda/cli.py:728`). That also affects `gda skill`, even though `skill()` passes `project=None` (`src/gda/cli.py:3371`) and ADR-0024 defines `gda skill` as a pure emitter meta-command, sibling of `info`/`schema`, with no Godot spawn. ADR-0006 also says meta commands take no project and run projectless. On this PR head I reproduced:

```bash
GDA_PROJECT=/tmp/not-a-godot-project uv run python -m gda skill --json
```

It exits 4 with `project_not_found`. That is a behavior regression for a projectless meta command and for any environment where a project context is inherited through `$GDA_PROJECT` while fetching the bundled skill.

Required: the new `project_not_found` wording in `docs/adr/0002-headless-structured-output-contract.md` and `src/gda/error_codes.py` says an explicit `--project`/`$GDA_PROJECT` that is empty is structured as `project_not_found`, but `resolve_project_dir()` still ignores an empty env var via `if env_value:` (`src/gda/project.py:64`) and falls through to cwd/projectless behavior. Please either implement empty `$GDA_PROJECT` as an explicit invalid input, or narrow the registry/ADR wording to the behavior that is actually implemented.

Consider: ADR-0023 still describes a recipe as producing the outcome by resolving the project plus running the CLI-side operation. This PR moves recipe project resolution into `_dispatch_recipe()` instead. If that is the intended architecture, please amend ADR-0023 so the documented command-descriptor contract matches the new ownership.

### Spec

The core #353 acceptance criteria for project-using commands are mostly implemented: `_dispatch()` and `_dispatch_recipe()` both route through the shared resolver, `script run` no longer carries its own invalid-project try/except, the ADR/Python error-code registry drift test passes, and the new tests cover invalid sentinel, invalid recipe, empty `--project`, and bad `$GDA_PROJECT` cases.

However, the same unconditional `_dispatch_recipe()` resolution is scope creep relative to #353. The issue asks for invalid `--project` / `$GDA_PROJECT` to become structured for commands that resolve a project; it does not ask projectless recipe/meta commands such as `gda skill` to start failing when an inherited `$GDA_PROJECT` is invalid. Please keep the shared structured failure for project-using commands while preserving meta-command projectlessness.

Verification I ran:

```text
env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_invalid_project_is_structured.py tests/test_error_registry.py
# 17 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_script_run_commands.py -q
# 7 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_skill_command.py -q
# 31 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_skill_surface_sync.py tests/test_schema_command.py -q
# 83 passed
```

Summary: Standards has 2 required findings plus 1 ADR follow-up; Spec has 1 required scope-creep finding. Worst issue: `gda skill --json` now fails under an invalid inherited `$GDA_PROJECT` even though it is a projectless meta command.

qinhaihong-red:

Verdict: request changes.

### Correctness

Required: `gda skill` now fails when an inherited `$GDA_PROJECT` is invalid. `gda skill` is a pure local/meta command: it emits or installs the bundled Skill and should not require or validate a Godot project. In this PR, `_dispatch_recipe()` calls `_resolve_project_or_fail(project)` for every recipe command (`src/gda/cli.py:728`); `skill()` passes `project=None` (`src/gda/cli.py:3371`), but `_resolve_project_or_fail(None)` still consults `$GDA_PROJECT`. Repro on this PR head:

```bash
GDA_PROJECT=/tmp/not-a-godot-project uv run python -m gda skill --json
```

Actual result: exit 4 with `project_not_found`. Expected result: the same Skill JSON that `gda skill --json` emits without any project context. This can break Skill discovery/install flows in shells, MCP subprocess environments, or agent runners that carry a stale or intentionally project-pinned `$GDA_PROJECT` while invoking meta commands.

Please keep the shared structured failure for project-using recipes, but exclude projectless/meta recipes from project resolution. A focused regression test belongs in `tests/test_skill_command.py`, for example setting `GDA_PROJECT` to a non-project and asserting `gda skill --json` still succeeds.

Required: the error-code/ADR wording now says empty `$GDA_PROJECT` is part of the generalized `project_not_found` case, but the resolver still ignores empty env values (`src/gda/project.py:64`, `if env_value:`). That means the implementation can fall through to cwd/projectless behavior instead of treating the empty env as the documented explicit invalid input. Either implement the documented behavior with a test, or narrow the wording so it does not promise empty `$GDA_PROJECT` handling.

### Readability and Architecture

Consider: the refactor changes ownership of recipe project resolution from the recipe seam to `_dispatch_recipe()`, but ADR-0023 still describes recipes as producing the outcome by resolving the project plus running the CLI-side operation. If this new ownership is intentional, update ADR-0023 and consider making the dispatcher contract explicit, such as a project-context requirement on the descriptor, so future meta recipes do not repeat the `gda skill` regression.

### Security

No security issue found. This change does not add new trust boundaries, external input sinks, secrets, or command execution paths beyond existing CLI project resolution.

### Performance

No performance issue found. The added resolution work is constant-time path checking, but the unconditional application to meta recipes is a correctness/architecture problem rather than a performance concern.

### Verification

```text
env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_invalid_project_is_structured.py tests/test_error_registry.py
# 17 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_script_run_commands.py -q
# 7 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_skill_command.py -q
# 31 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_skill_surface_sync.py tests/test_schema_command.py -q
# 83 passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run ruff check .
# All checks passed

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run ruff format --check .
# 148 files already formatted

env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pyright
# 0 errors, 0 warnings, 0 informations
```

The focused #353 tests and static checks pass, but the meta-command regression above still needs to be fixed before merge.

qinhaihong-red:
## All findings addressed — `8b031b0`

Thanks — both required findings were real, and the `gda skill` one was scope creep I introduced. Fixed and pushed; CI green (ruff / unit / pyright), MERGEABLE/CLEAN.

### [Required] `gda skill` regression — projectless meta recipe no longer resolves
Root cause was exactly as you diagnosed: `_dispatch_recipe` resolved a project for *every* recipe, so `skill()`'s hardcoded `project=None` still consulted `$GDA_PROJECT`. Rather than special-case skill, I made the dispatcher contract explicit (your suggestion):
- New `projectless: bool = False` field on `HeadlessCommand`. `_dispatch_recipe` now does `resolved = None if cmd.projectless else _resolve_project_or_fail(project)`.
- `SKILL_COMMAND` is marked `projectless=True`. Future meta recipes set the flag instead of re-hitting this.
- Regression test `test_skill_ignores_an_invalid_inherited_gda_project` in `tests/test_skill_command.py`: `GDA_PROJECT=<non-project> gda skill --json` → exit 0, emits the skill. Verified:
```
$ GDA_PROJECT=/tmp/not-a-godot-project gda skill --json
{"name":"gda","version":"0.1.48","content":"---\nname: gda…      # not project_not_found
$ GDA_PROJECT=/tmp/not-a-godot-project gda export run --preset X --mode pack --json
{"error":{"category":"operation","code":"project_not_found",…}}  # project-using recipe still structured
```

### [Required] Empty `$GDA_PROJECT` wording vs implementation
You're right — `resolve_project_dir` ignores an empty env (`if env_value:`), so empty `$GDA_PROJECT` falls through to cwd/projectless and is NOT `project_not_found`; only an empty `--project` raises. I narrowed the wording (not the behavior — leaving an empty env as "unset-ish" is the safer choice) in both `error_codes.py` and the ADR-0002 registry: *"an operation needed one and none was resolved, or an explicit `--project` was empty, or a `--project`/`$GDA_PROJECT` does not name a Godot project."* Drift test still green.

### [Consider] ADR-0023 ownership
Added a dated `> Outcome (2026-07-01, #353/#357)` note to ADR-0023: recipe project resolution moved into `_dispatch_recipe` (resolved once, shared with the sentinel path); a recipe now produces the outcome from an already-resolved project; the `projectless` flag excludes meta recipes. (Kept the point-in-time decision prose per our ADR-record convention.)

### Verification
`uv run pytest -m "not e2e"` → **1024 passed**; `ruff check` / `ruff format --check` / `pyright` clean; the two repros above.

Issue #365, opened 2026-07-01
Inline sub-resource authoring (add-subresource + nested addressing) [deferred]


qinhaihong-red:
> *This was generated by AI during triage.*

## Parent

#361 (origin/context). Explicitly **deferred** by **ADR-0033**.

## What to build

Inline sub-resource authoring: a new `add-subresource` verb that creates an **embedded `SubResource`** on an Object/Resource-typed property (no separate `.tres` file), plus **nested addressing** so the sub-resource's own fields can then be set via `node set` / `resource set`.

**Status: deferred.** ADR-0033 delivers the external `.tres`-reference workflow (#363) as the high-value path first. This inline variant needs new command surface and a nested-addressing design that is not yet fixed. Build only when a concrete need appears that the external-reference workflow cannot meet (i.e. the sub-resource **must** be inline and cannot be an external `.tres`).

## Acceptance criteria

- [ ] (Design pending) An `add-subresource` verb creates an inline `SubResource` of a chosen Resource class on an Object/Resource-typed property of a node or resource.
- [ ] (Design pending) The inline sub-resource's own fields are addressable and settable via the existing `set` commands.
- [ ] Structured errors on a bad type / type-mismatched sub-resource, consistent with #363.

## Blocked by

None technically — but deliberately deferred per ADR-0033 until a concrete "must be inline" need appears. Needs a design pass (command surface + nested addressing) before it is agent-ready.

qinhaihong-red:
> *This was generated by AI during triage.*

## Triage Decision

**Category:** `enhancement` · **State:** `ready-for-human` (removed `needs-triage`)

**Why `ready-for-human` (not `ready-for-agent`):** Per ADR-0033's "Considered options", this inline sub-resource variant was **deferred, not rejected**. Before it can be delegated to an AFK agent it needs a human **design pass** — an architectural decision on (a) the `add-subresource` command surface and (b) the **nested-addressing** scheme for `set`-ing the embedded `SubResource`'s own fields afterward. That design is not yet fixed, so there is no agent-ready contract to hand off.

**Gating / trigger condition:** deliberately deferred until a concrete **"must be inline"** need appears that the external `.tres`-reference workflow (ADR-0033 → #363, now landed) cannot meet. Until then it stays parked as a designed-but-not-scheduled backlog item. It is **not** `wontfix` — a deferral is not a rejection, so it is intentionally *not* recorded in `.out-of-scope/`.

[...]

Issue #364, opened 2026-07-01
gda skill: document the scene-authoring command set (discoverability)


qinhaihong-red:
> *This was generated by AI during triage.*

## Parent

#361 (origin/context). Related design: **ADR-0033**.

## What to build

Close the discoverability gap that #361 surfaced: an agent tried `node set --property script` (and other Object-typed properties) and hit a dead end, unaware that dedicated commands already exist. Document the **scene-authoring command set** in the gda skill guidance so the right command is discoverable:

- `script attach` — the one authoritative way to bind a `.gd` script to a node (#118).
- `resource create` / `resource set` — author and populate a Resource (built-in or project-local `class_name`).
- `node set` / `resource set --value res://….tres` — assign an existing Resource to an Object-typed property (the #363 capability).
- **Property value coercion forms** — document the accepted `--value` string forms so they are discoverable, notably the **comma-separated** form for `Vector2`/`Color` (`--value "48,72"`, `--value "0.2,0.6,1,1"`) rather than a JSON array or a Godot constructor literal. *(Folded in from #360's secondary note: the accepted comma-form is correct but not discoverable from `--help`.)*

Behavioral, not a code change to the engine surface: the actionable error messages themselves ship with #363; this slice makes the whole workflow legible from the skill (and `--help`) so agents reach for the correct command and value form in the first place.

## Acceptance criteria

- [ ] The gda skill guidance describes, as a coherent scene-authoring workflow, how to attach a script (`script attach`), create/populate a Resource (`resource create` / `resource set`), and assign a Resource to an Object-typed property (`node set` / `resource set --value res://…`).
- [ ] The guidance makes clear that `node set --property script` is NOT the way to attach a script (point to `script attach`).
- [ ] The accepted `--value` coercion forms — in particular the comma-separated form for `Vector2`/`Color` — are discoverable from the relevant command `--help` and/or the skill guidance (not only from source).
- [ ] The guidance is emitted by the version-locked skill channel (so it stays in step with the installed CLI, per ADR-0024).

## Blocked by

- #363 (the guidance must accurately describe the shipped `res://` Resource-assignment capability).

qinhaihong-red:
> *This was generated by AI during triage.*

Folded in the `--help` discoverability nit for property **value coercion forms** (the comma-separated `Vector2`/`Color` form) from #360's secondary note — it is the same discoverability theme as this issue. Added to *What to build* and as an acceptance criterion.

Issue #363, opened 2026-07-01
node set / resource set: assign an existing Resource by res:// path to an Object-typed property


qinhaihong-red:
> *This was generated by AI during triage.*

## Parent

#361 (origin/context). Design fixed by **ADR-0033**.

## What to build

Let `gda node set` and `gda resource set` assign an **existing** Resource — referenced by a `res://….tres` path — to an **Object-typed property that expects a Resource (sub)class** (e.g. a `CollisionShape2D`'s `shape`). End to end: the command detects that the target property's declared type is Object, loads the `res://` path as a Resource, type-checks the loaded resource against the property's declared **engine** class, and assigns it as an **external `ext_resource` reference** (not inlined), then saves.

This completes the external sub-resource authoring workflow with **no new command**:

```
gda resource create res://shapes/box.tres --type RectangleShape2D
gda resource set    res://shapes/box.tres --property size  --value 32,64
gda node set scene.tscn --node Col --property shape --value res://shapes/box.tres
```

Design constraints from ADR-0033:
- The shared `_coerce_value` (byte-identical mirror in the gda harness, used by live `game set`) stays **unchanged**. Object resolution is a **separate, headless-only** step in `node set` / `resource set` that reads the property's expected-class hint from its property-list entry.
- The `script` property is **excluded** from this generic Object path and routed to `script attach` (#118): setting `--property script` returns an actionable error pointing at `script attach`, never a second script-binding entry.
- Type-check scope is **engine-class-typed** Object properties. A property typed as a script `class_name` is out of scope here (its validation will reuse ADR-0032's resolver).
- Assigning a Resource on a live `game set` is out of scope.
- The concrete `GdaError` codes for the failure modes are **deferred to this slice** by ADR-0033 (they are public ABI, sourced from the error registry per ADR-0002): mint them here and register them, keeping the GDScript↔Python code-mirror in sync.

## Acceptance criteria

- [ ] `node set <scene> --node <n> --property <objProp> --value res://x.tres` assigns the loaded Resource to an engine-class-typed Object property and the saved scene loads with the resource attached as an `ext_resource`.
- [ ] `resource set <tres> --property <objProp> --value res://x.tres` does the same for a resource-on-resource assignment.
- [ ] The full `resource create` → `resource set` → `node set` external sub-resource workflow above wires a `CollisionShape2D.shape` (or equivalent) end to end.
[...]

Issue #361, opened 2026-07-01
gda scene authoring: node/resource set can't set Object-typed properties (attach a script, create/assign an inline sub-resource)


qinhaihong-red:
## What to build

`gda node set` / `gda resource set` coerce **value-typed** properties fine (scalars, `Vector2`, `Color`, … with the comma-separated form). They cannot set **Object-typed** properties, so an agent can add placeholder nodes but cannot make them *do* anything. The two capabilities that close this — and turn gda from "adds empty nodes" into "authors a wired, physical scene":

1. **Attach a script to a node.** Point a node's `script` property at a `res://…gd` file, so an added `CharacterBody2D` actually runs `player_controller.gd`.
2. **Create + assign an inline sub-resource.** Give a property that expects a `Resource` a freshly-created sub-resource of a chosen type (e.g. a `CollisionShape2D`'s `shape` = a `RectangleShape2D`), then set that sub-resource's own fields.

Shared root cause: both `script` and `shape` are `Object`-typed properties, and `set` rejects any value for them.

## Why (ADR-0025: agent value + structured-op fit — NOT editor parity)

- These are the difference between an empty blockout and a functioning scene: **any** node with behavior (a script) or physics (a shape) currently must be hand-authored as `.tscn` text. That's most real nodes — a central gap for gda's stated mission (agents building Godot games).
- Both are clean, typed, structured ops that fit gda's model: `set-script <node> <res://x.gd>`; `add-subresource <node|resource> <property> --type <ResourceClass>` (then `set` its fields).
- Hand-writing `.tscn`/`.tres` text is fragile: exact serialization, `ext_resource`/`SubResource` id bookkeeping, `load_steps`, and format drift across Godot versions. (Building Panda Adventure S1 required reverse-engineering the `InputEventKey` serialization via `var_to_str` and the comma value-form — see PR #359.)
- **Scope guard:** this is the two ops above, not "author any scene / editor parity" (ADR-0025 rejects parity). Highest value is mutating existing scenes and format-agnostic, validated authoring — the agent can still write greenfield text when it prefers.

## Reproduction (gda 4.6-era, this repo's demo)

```
gda node set  scene.tscn --node Player --property script --value res://src/controllers/player_controller.gd
  → uncoercible_value: cannot coerce value res://…gd to Object for property script
gda node set  scene.tscn --node Col --property shape  --value RectangleShape2D
  → uncoercible_value: cannot coerce value RectangleShape2D to Object for property shape
```

[...]

PR #366, opened 2026-07-01
docs(adr): add ADR-0032 (class_name static-scan fallback) and ADR-0033 (Object-typed property assignment)

qinhaihong-red:
## Summary

Records the two design decisions produced while triaging the Panda Adventure S1 dogfooding feedback (#360, #361). Docs only — no code changes; the implementation lands in the tracked child issues.

- **ADR-0032** — resolve a project-local GDScript `class_name` for `node add` / `resource create` / `find-references` in a **headless, editor-never-opened** project. gda adds a gda-owned **static scan** of the project's `.gd` sources as a **cache-independent fallback**, after (1) built-in engine classes and (2) the editor global class list, invoked only when those miss. Cache-first (editor-opened projects unchanged); a duplicate `class_name` is rejected with a new `ambiguous_class_name` error; all three sites share one resolver. Rejects the `--editor --quit` scan and the doc-only options. Captures the **#360** fix (extends #342, under ADR-0009/0010).
- **ADR-0033** — `node set` / `resource set` accept a `res://….tres` path to assign an **existing Resource** to an **Object-typed property** (external `ext_resource` reference, not inlined). The shared `_coerce_value` harness mirror stays byte-identical; the `script` property is routed to `script attach` (#118); inline sub-resources are **deferred**. Captures the **#361** design, under ADR-0025.

## Follow-up issues (implementation)

- #360 — `class_name` static-scan fallback *(ready-for-agent, ADR-0032)*
- #363 — assign an existing Resource by `res://` path to an Object-typed property *(ready-for-agent, ADR-0033)*
- #364 — gda skill scene-authoring discoverability docs *(ready-for-agent, blocked by #363)*
- #365 — inline sub-resource authoring *(deferred)*
- #361 — closed as split into #363/#364/#365

## Test plan

- [ ] N/A — documentation only (two new ADR files); no code or behavior change.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -139,10 +139,12 @@ The set of points where a single `gda` run triggers the target project's own
 code to run: autoload constructors at engine startup (every `--project` op), the
 `_init` of scripts on nodes *or resources* that an instantiating operation
 constructs (a `class_name` node via `node add`, or a script-backed `class_name`
-Resource via `resource create`), and — via `gda script run` (ADR-0031) — the
+Resource via `resource create`), the `_init` of a **script-backed Resource loaded
+as a value** assigned to an Object-typed property (`node set` / `resource set
+--value res://…`, ADR-0033), and — via `gda script run` (ADR-0031) — the
 **full execution of a named project script**.
-All stay within the `Trusted project` assumption (ADR-0009); `script run` widens
-this surface without adding a new trust axis.
+All stay within the `Trusted project` assumption (ADR-0009); `script run` and the
+loaded-value assignment (ADR-0033) widen this surface without adding a new trust axis.
 _Avoid_: attack surface, code-execution risk
 
 **Concurrent external editor**:
--- a/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
+++ b/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
@@ -1,0 +1,93 @@
+---
+status: accepted
+---
+
+# Project-local `class_name` resolution: a gda-owned static scan as a cache-independent fallback
+
+#342 gave `node add --type <class_name>` and `resource create --type <class_name>` the ability to
+resolve a **project-local GDScript `class_name`**, resolving it through
+`ProjectSettings.get_global_class_list()`. That list is populated **only** by the Godot editor scan,
+written into `.godot/global_script_class_cache.cfg`. But gda's core target is a
+[Headless operation](../../CONTEXT.md) on a headless, **editor-never-opened** [Trusted project](../../CONTEXT.md)
+— which has no `.godot/`, so the global class list is **empty** and #342's support is unreachable in
+exactly the workflow gda positions itself for (surfaced dogfooding Panda Adventure S1, #359/#360).
+Three call sites share this single dependency: node instantiation (`_instantiate_node_type`), resource
+instantiation (`_instantiate_resource_type`), and `find-references`' `class_name → script path`
+resolution (#116). The one-shot editor-scan workaround (`godot --headless --editor --quit`) confirms
+the mechanism but is not a workflow gda can adopt on the agent's behalf.
+
+## Decision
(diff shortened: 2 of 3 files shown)


qinhaihong-red:
Review against `origin/main...HEAD` for PR #366 (docs-only ADR change). I used #360/#361/#363/#364/#365 plus the existing ADR/glossary context as the spec and standards surface.

## Standards

- [P1] `docs/adr/0033-object-typed-property-assignment-via-resource-reference.md:71-73` understates the trust-model change. The ADR says loading a `.tres` for `node set` / `resource set` is "not newly widened", but the current `Project-code execution surface` glossary only covers autoload startup, `_init` from instantiating node/resource scripts, and `script run` (`CONTEXT.md:137-145`). ADR-0009 also requires this surface to be classified honestly by mechanism, and ADR-0031 sets the precedent of saying the surface widens while still staying inside the Trusted-project assumption. If this decision intentionally makes `node set` / `resource set` load an arbitrary external Resource value, the ADR should say the surface widens but adds no new trust axis, and `CONTEXT.md` should be aligned in the same patch.

- [P2] `docs/adr/0033-object-typed-property-assignment-via-resource-reference.md:47-49` requires distinct structured codes for non-`res://`, not-a-Resource, and incompatible-resource failures, but it neither names those public `GdaError.code`s nor explicitly defers them to the implementation slice. Since ADR-0002 treats `GdaError.code` as public ABI whose authoritative source is the registry/table, please either name the new codes here or use the existing ADR pattern: the implementation issue will mint them and update ADR-0002/error registry when it lands.

## Spec

- [P1] `docs/adr/0033-object-typed-property-assignment-via-resource-reference.md:69-75` overstates the supported property scope. The ADR correctly says at lines 44-46 that script-`class_name`-typed Object properties are deferred, matching #363's scope of **engine-class-typed** Object properties. But the consequences then say agents may rely on `res://` values for "Resource-typed properties" generally, and that "every other Resource-typed property" goes through `node set` / `resource set`. That reads as including script-`class_name` Resource properties, which are explicitly out of scope. Please tighten this to "engine-class-typed Resource/Object properties" and keep script-`class_name` property validation explicitly deferred to the ADR-0032-backed follow-up.

No findings on ADR-0032. Its cache-first static fallback for project-local `class_name` resolution matches #360's clarified contract and is consistent with the existing headless/trusted-project ADRs.

qinhaihong-red:
Thanks — all three findings were valid; addressed in `2070b81` (docs only, same branch). No findings on ADR-0032, left as-is.

| Finding | Verdict | Resolution @ `2070b81` |
|---|---|---|
| **[P1 standards]** trust-surface understated (`0033:71-73`) | Accepted | Reworded the consequence: the **Project-code execution surface widens** — `node set` / `resource set` gain a new entry point that **loads** an external Resource value, so a **script-backed** Resource's `_init` runs on load, from a point the surface did not previously enumerate. Framed as "widens, **no new trust axis**" per the ADR-0031 precedent. **`CONTEXT.md`'s `Project-code execution surface` glossary entry aligned in the same commit.** |
| **[P2 standards]** structured codes neither named nor deferred (`0033:47-49`) | Accepted | Took the ADR-0031 pattern: the concrete `GdaError` codes are **not fixed in the ADR**; since `GdaError.code` is public ABI sourced from the registry (ADR-0002), the implementation slice **#363 mints and registers them** when it lands. Added that sentence to the contract edge. |
| **[P1 spec]** property scope overstated (`0033:69-75`) | Accepted | Tightened the Consequences from "Resource-typed properties" → **engine-class-typed** Resource/Object properties, and the asymmetry line to "every other **engine-class-typed** Resource property", keeping script-`class_name`-typed properties **explicitly deferred** (consistent with the contract-edge scope at 44-46). |

Net: ADR-0033 now states the surface widening honestly and aligns the glossary, defers the error-code ABI to #363, and scopes the value-form ABI to engine-class-typed properties.

Issue #362, opened 2026-07-01
Harness gda_log() silently drops records when the harness is present but not daemon-launched (no public predicate to detect it)


qinhaihong-red:
## Summary

`GdaHarness.gda_log()` — the ADR-0026 opt-in entry a game calls to emit `<<<GDA:LOG>>>` records that `gda logger tail` parses — is a **silent no-op** when the run was not launched by gda-daemon:

```gdscript
# addons/gda_harness/gda_harness.gd
func gda_log(level, message, fields) -> void:
    if not _daemon_launched:
        return                     # <-- silently drops the record
    print(LOG_MARKER + JSON.stringify({...}))
```

The condition is only observable through the **private** `var _daemon_launched` — there is **no public predicate** for "did the daemon launch this run / will gda_log emit?".

## Why it's a footgun

gda positions the harness + `logger tail` as the agent-observability path, and the game AGENTS.md guidance is to "log in a way `gda logger tail` can parse" — i.e. route structured logging through `gda_log`. The natural integration is:

```gdscript
var harness := get_node_or_null("/root/GdaHarness")
if harness != null:
    harness.gda_log(level, msg, fields)   # assumes present ⇒ will emit
    return
print(...)                                 # fallback only when ABSENT
```

But **"present" ≠ "will emit"**: whenever the harness is present-but-dormant — a plain `godot`/`--headless` run, the editor, or (as in this demo) after the harness is *committed* so it's present in every run — every record is silently swallowed, with no console output and nothing for `logger tail` to read. The author gets no signal that logging is dead.

## Where it surfaced (dogfooding)

Panda Adventure (PR #359) commits the harness (to avoid per-session `project.godot` revert churn). That made `/root/GdaHarness` present in *every* run, so a plain `godot --headless` run of the source project logged **nothing** — `player_ready`/`boot`/`player_land` never appeared. The **game-side** fix was to gate on the private flag:

```gdscript
if harness != null and harness.get("_daemon_launched"):
    harness.gda_log(...)
    return
print(...)
```

That works, but it depends on reaching into `_daemon_launched` (a private var, fragile across gda versions) because no public predicate exists. No gda source was changed.

## Suggested directions (for triage)

1. **Expose a public predicate** on the harness — e.g. `is_daemon_launched()` / `will_log()` — so game code can branch correctly without touching a private var. (Most targeted, lowest risk.)
[...]

PR #367, opened 2026-07-01
feat(harness): public is_daemon_launched() predicate for log gating (#362)

qinhaihong-red:
Refs #362

## What

Adds a **public, pure-read predicate** `GdaHarness.is_daemon_launched()` on the harness autoload, returning `true` iff `gda-daemon` launched this run — the *same* condition `gda_log()` already gates on (`_daemon_launched`). Game code can now gate structured logging on the **daemon-launched state**, not on harness *presence*, and fall back to `print()` when dormant instead of silently dropping every record.

Per the Agent Brief, direction 3 from the issue (degrading `gda_log()` to `print()` when dormant) is **rejected** — a print is a non-inert side effect that breaks ADR-0018. This change fixes *discoverability* of the condition, not the condition's behavior. No error codes added.

## Changes

- **`src/gda/harness/gda_harness.gd`** — new `func is_daemon_launched() -> bool` returning the existing private `_daemon_launched` (deep-module reuse — no duplicated launch-marker / inert-gating logic). A pure read: no connection, no print, no state change. Returns `false` in a human editor run, a plain run, and an exported build (where `_ready` returns at the `template` gate before `_daemon_launched` is set), preserving the ADR-0018 inert-when-dormant guarantee.
- **`src/gda/harness/install.py`** — `HARNESS_VERSION` `3` → `4` so already-installed copies self-sync the new body (the version header stays honest with the changed harness).
- **`src/gda/skill/SKILL.md`** — new *Structured logging from game code* subsection in the **logging area** documenting the "gate on `is_daemon_launched()`, not presence" pattern with a snippet.
- **`tests/test_harness_install.py`** — fast test: the **installed** copy carries the predicate.
- **`tests/test_e2e_logger.py`** — e2e: (a) a daemon-launched session → a `gda_log` gated on the predicate emits and `gda logger tail` reads it back (predicate true); (b) a plain run → predicate false, `gda_log()` stays a no-op (no `<<<GDA:LOG>>>` leak), and the fallback path fires.

## Tests

- Fast tier: `uv run pytest -m "not e2e"` → **1025 passed**.
- e2e: `pytest tests/test_e2e_logger.py -m e2e` → **2 passed**; `tests/test_e2e_harness_install.py -m e2e` → **3 passed** (installed harness with the new body still boots inert).
- Lint/format/type: `ruff check` clean, `ruff format --check` clean, `pyright` 0 errors.

### DIFF
--- a/src/gda/harness/gda_harness.gd
+++ b/src/gda/harness/gda_harness.gd
@@ -953,14 +953,18 @@ func _error(code: String, message: String) -> String:
 
 # The PUBLIC, stable predicate reporting whether gda-daemon launched this run (#362)
 # — the SAME condition `gda_log()` gates on (`_daemon_launched`). Game code should
-# branch on THIS, not on harness *presence*: once installed, `/root/GdaHarness` exists
-# in every run, but the harness only captures logs when the daemon launched the
-# session, so gating on presence silently drops every record in a plain/editor/exported
-# run. A game-side logging helper instead gates on this predicate and falls back to
-# `print()` when it is false. It is a PURE READ — no connection, no output, no state
-# change — so the ADR-0018 inert-when-dormant guarantee holds: it returns false in a
-# human editor run, a plain run, and an exported build (where `_ready` returns at the
-# `template` gate before `_daemon_launched` is ever set), with no side effect in any.
+# branch on THIS, not on harness *presence*: where the harness IS installed it only
+# captures logs when the daemon launched the session, so gating on presence silently
+# drops every record in a plain/editor run. (A supported `gda export run` artifact
+# OMITS the harness entirely per ADR-0028, so this predicate is not even present to
+# call there — game code must resolve the autoload by node path and null-check it, as
+# the `GdaHarness` global does not parse when absent.) A game-side logging helper gates
+# on this predicate and falls back to `print()` when the node is absent or the predicate
+# is false. It is a PURE READ — no connection, no output, no state change — so the
+# ADR-0018 inert-when-dormant guarantee holds: it returns false in a human editor run, a
+# plain run, and an export that BYPASSED `gda export run` (the harness physically present
+# but inert — `_ready` returns at the `template` gate before `_daemon_launched` is ever
+# set), with no side effect in any.
 func is_daemon_launched() -> bool:
 	return _daemon_launched
 
--- a/src/gda/harness/gda_harness.gd
+++ b/src/gda/harness/gda_harness.gd
@@ -951,6 +951,20 @@ func _error(code: String, message: String) -> String:
 	return RESULT_BEGIN + JSON.stringify({"error": {"code": code, "message": message}}) + RESULT_END
 
 
+# The PUBLIC, stable predicate reporting whether gda-daemon launched this run (#362)
+# — the SAME condition `gda_log()` gates on (`_daemon_launched`). Game code should
+# branch on THIS, not on harness *presence*: once installed, `/root/GdaHarness` exists
+# in every run, but the harness only captures logs when the daemon launched the
+# session, so gating on presence silently drops every record in a plain/editor/exported
+# run. A game-side logging helper instead gates on this predicate and falls back to
+# `print()` when it is false. It is a PURE READ — no connection, no output, no state
+# change — so the ADR-0018 inert-when-dormant guarantee holds: it returns false in a
+# human editor run, a plain run, and an exported build (where `_ready` returns at the
+# `template` gate before `_daemon_launched` is ever set), with no side effect in any.
+func is_daemon_launched() -> bool:
+	return _daemon_launched
+
+
 # The active opt-in rich-log protocol (#282, ADR-0026). Project code in a live
 # session calls `GdaHarness.gda_log(level, message, fields)` to emit ONE fully
 # structured, field-carrying log record. It prints a single `<<<GDA:LOG>>>{json}`
--- a/src/gda/harness/install.py
+++ b/src/gda/harness/install.py
@@ -41,7 +41,7 @@ HARNESS_RES_PATH = f"res://{HARNESS_RES_DIR}/{HARNESS_FILE}"
 # copy to it (#225). The installed copy declares its version in a leading header
 # (`# gda-harness-version: <N>`); a mismatch re-materializes via the content
 # compare. NOT the package version — the harness changes far less often.
-HARNESS_VERSION = "3"
+HARNESS_VERSION = "4"
 
 _VERSION_HEADER_PREFIX = "# gda-harness-version:"
 _AUTOLOAD_HEADER = "[autoload]"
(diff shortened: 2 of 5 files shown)


qinhaihong-red:
## `$review` pass: Standards / Spec

Reviewed PR #367 against `origin/main...HEAD` (`c414a5e` -> `f801c4e`).

Validation run locally in an isolated `/tmp` checkout:

- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_harness_install.py -q` -> 15 passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_e2e_logger.py -m e2e -q` -> 2 passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run ruff check src/gda/harness/install.py tests/test_harness_install.py tests/test_e2e_logger.py` -> passed
- `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run ruff format --check src/gda/harness/install.py tests/test_harness_install.py tests/test_e2e_logger.py` -> passed

## Standards

Required: `src/gda/skill/SKILL.md` now recommends calling `GdaHarness.is_daemon_launched()` directly and says `/root/GdaHarness` exists in every run, including an exported build. That conflicts with `docs/adr/0028-harness-export-cleanliness.md`: `gda export run` must strip the harness before export, and its artifacts are harness-free by construction; only export routes that bypass `gda export run` may leave a physically present but inert harness.

The direct singleton sample is therefore unsafe for a physically clean export artifact, for a project before the daemon has ever installed the harness, or for any project where the autoload was explicitly removed. I verified this with a minimal Godot project that had no `GdaHarness` autoload: a script using the new sample shape fails to load with `Parse Error: Identifier "GdaHarness" not declared in the current scope.`

Please make the skill guidance absence-safe and scope the exported-build wording to the resident/non-stripped harness case, for example:

```gdscript
var harness := get_node_or_null("/root/GdaHarness")
if harness != null and harness.is_daemon_launched():
    harness.gda_log("info", "player spawned", {"hp": 100})
else:
    print("player spawned")
```

The same wording issue appears in `src/gda/harness/gda_harness.gd`: the new comment says the predicate returns false in "an exported build", but under the supported `gda export run` path the harness is physically absent. That comment should be narrowed to exported builds where the harness is present because the export bypassed `gda export run`, or otherwise mention that supported `gda export run` artifacts omit the harness entirely.

## Spec

No spec findings.

The diff satisfies issue #362's core requirements: it exposes `GdaHarness.is_daemon_launched()` as a public pure read over the same `_daemon_launched` condition that gates `gda_log()`, leaves dormant behavior unchanged, bumps the harness install version so installed copies resync, and updates the gda skill guidance to gate on daemon-launched state with a fallback path.

Residual spec coverage risk: the new tests directly exercise daemon-launched `true`, plain-run `false`, and installed-copy presence. Human-editor and exported-build false behavior is inferred from the unchanged `_daemon_launched := false` default and the existing ADR-0018/ADR-0028 inert/template-gate coverage, rather than newly asserting `is_daemon_launched()` in those exact runtime modes.

Summary: Standards has 1 required finding; Spec has 0 findings. Worst issue: the new public guidance can fail outright when `GdaHarness` is absent, which conflicts with ADR-0028's physically clean `gda export run` artifact contract.

qinhaihong-red:
## `$agent-skills:code-review-and-quality` pass

### Findings

Required: Make the new logging guidance absence-safe before publishing it through `gda skill`.

The public predicate implementation itself is straightforward, but the new `src/gda/skill/SKILL.md` snippet calls `GdaHarness.is_daemon_launched()` directly. That only works when the autoload singleton is physically present. It is not guaranteed to be present: `gda export run` intentionally strips the harness from release artifacts (ADR-0028), users can run before `gda daemon start` has installed it, and `gda daemon uninstall` can remove it.

This is not just wording. I verified a minimal project with no `GdaHarness` autoload and the new sample shape; Godot fails to load the script with:

```text
SCRIPT ERROR: Parse Error: Identifier "GdaHarness" not declared in the current scope.
```

So the recommended fallback path will not run in the absence case. Please change the guidance to first resolve the autoload, then gate on the predicate:

```gdscript
var harness := get_node_or_null("/root/GdaHarness")
if harness != null and harness.is_daemon_launched():
    harness.gda_log("info", "player spawned", {"hp": 100})
else:
    print("player spawned")
```

The adjacent `src/gda/harness/gda_harness.gd` comment should also avoid saying the predicate returns false in a generic "exported build" without noting the supported `gda export run` path removes the harness entirely.

### Five-axis review

- Correctness: one required issue above; the public sample can fail before fallback when the singleton is absent.
- Readability/simplicity: no additional blockers. The new predicate is easy to follow, though the long inline comment should be tightened while fixing the export wording.
- Architecture: the predicate reuses the existing `_daemon_launched` state rather than duplicating launch-marker logic, which fits the harness design.
- Security: no new input boundary, secret, auth, or injection concern found.
- Performance: no concern; the predicate is a boolean read and the tests do not add hot-path cost.

### Verification

Local isolated checkout:

- `pytest tests/test_harness_install.py -q` -> 15 passed
- `pytest tests/test_e2e_logger.py -m e2e -q` -> 2 passed
- `pytest tests/test_skill_command.py tests/test_skill_surface_sync.py -q` -> 35 passed
- `pytest tests/test_e2e_harness_install.py -m e2e -q` -> 3 passed
- `ruff check ...` and `ruff format --check ...` on changed Python test files -> passed
- `git diff --check origin/main...HEAD` -> passed

GitHub checks are green for ruff, unit/package build, and pyright; the Godot e2e workflow is currently skipped.

Verdict: Request changes for the absence-safe guidance issue.

qinhaihong-red:
## Review fix pushed @ 51ded89

**Finding (both passes, REQUIRED):** the logging snippet referenced the `GdaHarness` global directly, which fails to *parse* when the autoload is absent (a `gda export run` stripped build per ADR-0028, a project before `gda daemon start`, or after `gda daemon uninstall`) — taking the whole script down so even the `print()` fallback never runs.

**Resolution (51ded89):**
- `src/gda/skill/SKILL.md` — snippet now resolves the autoload first and null-checks it before gating:
  ```gdscript
  var harness := get_node_or_null("/root/GdaHarness")
  if harness != null and harness.is_daemon_launched():
      harness.gda_log("info", "player spawned", {"hp": 100})
  else:
      print("player spawned")
  ```
  Prose corrected: the harness is present only where installed, a supported `gda export run` artifact omits it entirely (ADR-0028), and the `GdaHarness` global must not be referenced directly (parse failure when absent).
- `src/gda/harness/gda_harness.gd` — narrowed the predicate's comment: it returns false in a human editor run, a plain run, and an export that **bypassed** `gda export run` (harness physically present but inert); a supported `gda export run` artifact omits the harness entirely, so the predicate isn't present to call there.

**Verification:** `uv run gda skill` emits the absence-safe form; `ruff check` + `ruff format --check .` clean; `pyright` 0 errors; `pytest tests/test_harness_install.py` 15 passed; `pytest tests/test_e2e_logger.py -m e2e` 2 passed.

Issue #381, opened 2026-07-02
project get: Dictionary/Object-valued settings render as a Variant str() debug string, not structured JSON


qinhaihong-red:
## Summary

Extend the shared read-side value projection (`_jsonify`) so compound setting/property values — `Dictionary`, `Array`/packed arrays, and inline value `Object`s — return structured, indexable JSON instead of a Variant `str()` debug string. Design pinned in **ADR-0035** (read-side companion to ADR-0033).

## Current behavior

`project get` (and `node get` / `resource get`, `project list`, the `node set`/`resource set` echo, and the live `game get` read) project a value through the shared `_jsonify` helper (#55). Scalars and a few fixed-shape value types (`Vector2`/`Vector2i`/`Color`) pass through, but every `Dictionary`, `Array`, and `Object` degrades to the Variant's `str()` form. Reproduced on Panda Adventure (S2's `fire` InputMap action):

```
$ gda project get input/fire --json
{"setting":"input/fire","type":"Dictionary",
 "value":"{ \"deadzone\": 0.5, \"events\": [InputEventKey: keycode=74 (J), ..., InputEventKey: keycode=70 (F), ...] }"}
```

An agent cannot index `value.deadzone` or `value.events[0].keycode`, and the embedded `InputEventKey` list is not valid JSON.

## Desired behavior (per ADR-0035)

`_jsonify` becomes the single read-side **value projection**, applied uniformly to **every value gda emits through it** — `project`/`node`/`resource get`, the `node set`/`resource set` echo, the per-entry `value` of `project list` and `scene get-exports`, and the live `game get` read:

- `Dictionary` → JSON object (keys coerced to strings); `Array` and the packed-array family (`PackedInt32Array`, `PackedStringArray`, `PackedVector2Array`, …) → JSON array; elements/values re-enter the projection.
- An `Object` renders as one of three **projection kinds**:
  - **reference projection** — a `Resource` with a non-empty `res://` path → `{"type": "<Class>", "resource_path": "res://…"}` (not inlined; read-side mirror of ADR-0033);
  - **inline value projection** — a whitelisted path-less value Object (`InputEvent` subclasses initially) → `{"type": "<Class>", <its own storage properties, each re-projected>}`, using the storage-property model `node get` exposes;
  - **string fallback** — any other Object (not whitelisted, no `res://` path — e.g. a live `Node`) → the existing `str()` form.
[...]

Issue #380, opened 2026-07-02
project group: structured InputMap action authoring (add-input-action / remove-input-action)


qinhaihong-red:
## Summary

Add `gda project add-input-action` / `remove-input-action` — structured InputMap action authoring in `project.godot`, mirroring the `add-autoload` / `remove-autoload` precedent.

```
gda project add-input-action fire --key J --key F [--deadzone 0.5]
gda project remove-input-action fire
```

## Current behavior

An InputMap action is a compound `project.godot` entry that `project set` cannot express: its value is a `Dictionary` of `{deadzone, events: [InputEventKey, …]}` where the events are `InputEventKey` **Objects**. `project set --value TEXT` only coerces a string to a declared scalar/typed value; there is no string form for an event list. Today an agent adding an action must hand-edit `project.godot` with reverse-engineered `Object(InputEventKey, …)` literals (the trick Panda Adventure S1 discovered and S2 copy-pasted for its `fire` action, PR #379) — exactly the fragile hand-serialization ADR-0033 rejected for scenes.

## Why (ADR-0025: agent value + structured-op fit)

- InputMap actions are how a game becomes *driveable* — by the player, and by gda's own `input action` / `input sequence` live ops, which address actions by name. Every gameplay slice an agent builds adds at least one action.
- Clean structured-op fit: typed params (action name, key list, optional deadzone) → a canonical `ProjectSettings`-mediated write, exactly like `add-autoload`.

## Desired behavior

Two new `project`-group commands following the `add-autoload`/`remove-autoload` shape (typed params → `ProjectSettings.set_setting()` → `ProjectSettings.save()`):

- `gda project add-input-action <name> --key <k>... [--deadzone <f>] [--physical]` — registers `input/<name>` with the given keys, default deadzone `0.5`. `--key` accepts a key *name* (`J`, `F`, `space`) and/or a raw keycode. The written serialization must match what Godot itself emits (equal to the `var_to_str` form of the `Dictionary`/`InputEventKey` values), so the engine and editor load it identically to a hand-authored entry.
- `gda project remove-input-action <name>` — unregisters it and persists `project.godot`.

**Existing-action / missing-action semantics (decided):** adding an action whose name already exists is a structured `already_exists` error — never a silent clobber — mirroring `add-autoload` (replace = remove + add). Removing a missing action is a structured `unknown_setting` error, mirroring `remove-autoload`.

## Key interfaces

[...]

Issue #312, opened 2026-06-27
Add `gda project list`: enumerate ProjectSettings keys for agent discovery


qinhaihong-red:
## What to build

Add a `gda project list` headless command that enumerates a project's `ProjectSettings` keys, so an agent can **discover** which setting keys exist without an editor UI — closing the gap where `project get`/`set` both require you to already know the `section/key`, and `project info` reports only a fixed handful.

Today the runtime already exposes the full set via `ProjectSettings.get_property_list()` (gda uses it internally — e.g. the autoload enumeration), but there is no outward command. `project list` makes the `list → get → set` workflow self-consistent: list to discover, get to read one typed value, set to write it.

**Default scope (decided — option A).** A bare `gda project list` lists only the project's **customized (non-default)** settings — the keys actually written in `project.godot` — keeping default output small and useful for agents. Flags widen scope:

- `--all` — also include the engine's built-in default keys (the hundreds of registered settings), not just customized ones.
- `--section <prefix>` — restrict to keys under a `section/` prefix (e.g. `application/`, `display/`).

Each listed entry reuses the **same projection** `project get` reports — `{setting, type, value}` — plus an `is_default` boolean indicating whether the key is at its engine default (vs customized in `project.godot`). Internal/non-setting properties returned by `get_property_list()` must be filtered out so only real project settings appear.

Behaviour stays consistent with the rest of the `project` group: requires a resolved project (`project_not_found` otherwise), is a state-read that never instantiates a scene, and self-describes via `--schema` like every other command descriptor (per ADR-0023 — the render map / dispatch / schema are projections of the one `HeadlessCommand` descriptor, never parallel registries).

## Acceptance criteria

- [ ] `gda project list --json` on a project returns its customized (non-default) settings, each as `{setting, type, value, is_default}`, with a stable schema.
- [ ] `--all` additionally includes built-in default keys; `--section <prefix>` filters to keys under that prefix; the two compose.
- [ ] Internal / non-setting entries from `get_property_list()` are excluded — only real `ProjectSettings` keys appear.
- [ ] `value`/`type` match the projection `project get` reports for the same key (a listed entry round-trips through `project get`).
- [ ] No resolved project → `project_not_found` (exit 4), consistent with the rest of the `project` group; the command never instantiates a scene.
- [ ] The command self-describes via `--schema` and has a human renderer, registered through the single command descriptor (no parallel registry; ADR-0023).
[...]

PR #382, opened 2026-07-02
docs(adr): ADR-0035 read-side value projection for compound Variants

qinhaihong-red:
## What

Records **ADR-0035 — read-side value projection for compound Variants**, and adds the `Value projection` glossary term to `CONTEXT.md`.

This is the decision record produced while triaging #381 (found dogfooding Panda Adventure S2's `fire` InputMap action). It does **not** implement #381 — it lands the ADR the implementation depends on, so the `ready-for-agent` brief on #381 has an authoritative contract to build against.

## Decision (summary)

The shared, byte-identical-mirrored `_jsonify` becomes the one read-side value projection across `project`/`node`/`resource get` and the live `game` reads:

- `Dictionary` / `Array` / the packed-array family project recursively (dict keys stringified).
- An `Object` renders as one of three **descriptor kinds**: a **reference descriptor** (a `Resource` with a `res://` path — referenced, not inlined; the read-side mirror of ADR-0033's write-side reference), an **inline value descriptor** (a whitelisted path-less value Object like `InputEventKey`, via the existing storage-property model), or a **string fallback**.
- Bounded by a hard recursion depth cap; **no visited-set**. Output is always JSON-encodable.
- The inline-Object **whitelist is the risk-isolation boundary** — a non-whitelisted live Object (e.g. a `Node` from `game get`) falls back to `str()`, so the shared headless/live projector never blows up on a runtime scene tree or cycle.

New **public output ABI**, which is why it is recorded. No new `Gda error code`.

## Scope

- `docs/adr/0035-value-projection-compound-variants-to-structured-json.md` (new)
- `CONTEXT.md` — new `Value projection` term under a `Structured output` subsection

Implementation is tracked by #381 (`ready-for-agent`); the independent write-side companion is #380.

Refs #381, #380. Read-side companion to ADR-0033.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -104,6 +104,23 @@ launch), survives the session process so a crash stays diagnosable until relaunc
 (ADR-0022).
 _Avoid_: console output, stdout dump
 
+### Structured output
+
+**Value projection**:
+The read-side contract for rendering a Godot value into the structured JSON a
+successful result carries. Scalars and the small fixed-shape value types pass
+through directly; a `Dictionary`/`Array` (and the packed-array family) projects
+recursively; and an `Object` is rendered by one of three **descriptor kinds** — a
+**reference descriptor** for a `Resource` that has a `res://` path (named by type
+and path, never inlined — the read-side mirror of ADR-0033's write-side
+reference), an **inline value descriptor** for a whitelisted path-less value
+`Object` (named by type plus its projected fields), or a plain **string
+fallback** for anything else. One projection shared across every read
+(`project`/`node`/`resource get`, and the live `game` reads) so a value reads the
+same everywhere; the whitelist is the boundary that keeps the shared projection
+safe on the live side (ADR-0035).
+_Avoid_: value rendering, str dump, serialization
+
 ### Failure reporting
 
 **Gda error code**:
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -110,16 +110,18 @@ _Avoid_: console output, stdout dump
 The read-side contract for rendering a Godot value into the structured JSON a
 successful result carries. Scalars and the small fixed-shape value types pass
 through directly; a `Dictionary`/`Array` (and the packed-array family) projects
-recursively; and an `Object` is rendered by one of three **descriptor kinds** — a
-**reference descriptor** for a `Resource` that has a `res://` path (named by type
+recursively; and an `Object` is rendered by one of three **projection kinds** — a
+**reference projection** for a `Resource` that has a `res://` path (named by type
 and path, never inlined — the read-side mirror of ADR-0033's write-side
-reference), an **inline value descriptor** for a whitelisted path-less value
+reference), an **inline value projection** for a whitelisted path-less value
 `Object` (named by type plus its projected fields), or a plain **string
-fallback** for anything else. One projection shared across every read
-(`project`/`node`/`resource get`, and the live `game` reads) so a value reads the
-same everywhere; the whitelist is the boundary that keeps the shared projection
-safe on the live side (ADR-0035).
-_Avoid_: value rendering, str dump, serialization
+fallback** for anything else. One projection shared across **every value gda
+emits** — the `get` reads (`project`/`node`/`resource get`), the value echoed by
+`node set`/`resource set`, the per-entry value of `project list` and `scene
+get-exports`, and the live `game get` read — so a value reads the same
+everywhere; the whitelist is the boundary that keeps the shared projection safe
+on the live side (ADR-0035).
+_Avoid_: value rendering, str dump, serialization, descriptor
 
 ### Failure reporting
 
--- a/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
+++ b/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
@@ -1,0 +1,85 @@
+---
+status: accepted
+---
+
+# Value projection: compound Variants (`Dictionary`/`Array`/`Object`) to structured JSON on the read side
+
+`project get` / `node get` / `resource get` (and the live `game` reads) project a value into the
+result's `value` field through the shared `_jsonify` helper (#55): scalars and a few fixed-shape value
+types (`Vector2`, `Vector2i`, `Color`) pass through, but **every other type — `Dictionary`, `Array`,
+and any `Object` — degrades to the Variant's `str()` debug string**. So a compound-valued setting
+(surfaced dogfooding Panda Adventure S2's `fire` InputMap action, #381) reads back as
+`"{ \"deadzone\": 0.5, \"events\": [InputEventKey: keycode=74 (J), …] }"` — an agent cannot index the
+deadzone or enumerate the bound keys, and the event list is not even valid JSON. For a tool whose
+positioning is structured, machine-readable output, compound values are exactly where the structure is
+missing. `_jsonify` is **byte-identical mirrored** into the [gda harness](../../CONTEXT.md) (the
+`--- shared coercion ---` block, enforced by `tests/test_harness_coercion_mirror.py`), so it is also the
+projector for the live `game get` / `game call` reads — where a value can be an **arbitrary runtime
+Object (a live `Node`) or a cyclic graph**.
+
+## Decision
+
+Extend the shared, mirrored `_jsonify` into a **`Value projection`** (see `CONTEXT.md`): the one
+read-side contract for turning a Godot Variant into structured JSON, applied uniformly across every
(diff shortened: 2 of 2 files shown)


qinhaihong-red:
## Review pass: Standards / Spec

Fixed point: `origin/main` (`d88c7d4`). Head: `90f64da`. Diff reviewed: `CONTEXT.md` and `docs/adr/0035-value-projection-compound-variants-to-structured-json.md`.

## Standards

Required:
- `docs/adr/0035-value-projection-compound-variants-to-structured-json.md:72-73` declares a new public output ABI whose nested shapes agents may rely on (`value.deadzone`, `value.events[0].keycode`, descriptor fields). ADR-0004 requires command output contracts to be model-driven and surfaced mechanically through `--schema` / gda-mcp, not maintained only as prose (`docs/adr/0004-schema-flag-self-description.md:47-57`, hard gate at `:104-112`). Please specify how the implementation must carry this new projection into the success result models/schema/docs/tests for each affected output surface, or explicitly record this as a deliberate exception to ADR-0004.
- `docs/adr/0035-value-projection-compound-variants-to-structured-json.md:17` and `:76` name `game call`, but I do not see an accepted or implemented `game call` command in the current surface. ADR-0025 requires new command surface to pass the agent-value + structured-operation inclusion test before being committed. Please either link the accepted decision for `game call`, or phrase this as future live value-returning operations rather than a committed command.

Judgment call:
- The new terms `reference descriptor` and `inline value descriptor` sit close to the existing glossary term `Command descriptor` (`CONTEXT.md:210-218`). Given `docs/agents/domain.md:26-30` asks us to avoid vocabulary drift, consider `reference projection` / `inline value projection` or explicitly define why these are descriptor kinds.

## Spec

Required:
- The projection surface is not reconciled across the spec and the new docs. ADR-0035 says the projection applies to `project`/`node`/`resource get` + set-echo + `project list` and live reads (`docs/adr/0035...md:22-25`), while the #381 brief scopes the shared `_jsonify` contract to get/read surfaces and says `project list` is unaffected. Current repo docs/code from #312 show `project list` already includes `{setting,type,value}` and reuses the same projection (`docs/command-catalog.md:425-430`, `src/gda/ops/operations.gd:2195-2250`), so the right fix is not necessarily to remove `project list`; it is to make the ADR, CONTEXT term, and #381 implementation brief agree on one canonical scope.
- `CONTEXT.md:118-120` says the projection is shared across every read as `project`/`node`/`resource get` plus live `game` reads, but omits the `set-echo` and `project list` surfaces that ADR-0035 includes. Please align the glossary with the ADR or narrow the ADR, otherwise an implementation can update one set of result surfaces and still appear compliant.

Summary: Standards found 2 required issues and 1 terminology consideration. Spec found 2 required scope/alignment issues. Worst issue: the new public output ABI is not tied back to the model/`--schema` contract chain that gda-mcp depends on.

qinhaihong-red:
## Review pass: Code Review and Quality

Verdict: Request changes. This is a good direction for #381, but the ADR is defining a public output ABI and needs to be tighter before it becomes the implementation contract.

Required - Architecture / contract:
- Make the value projection ABI executable through the model/schema chain, not prose-only. ADR-0035 says agents may rely on `value.deadzone`, `value.events[0].keycode`, and descriptor fields (`docs/adr/0035-value-projection-compound-variants-to-structured-json.md:72-73`), while current result model descriptions still describe only scalars and packed Vector/Color lists for `ProjectGetResult` / `ListedProjectSetting` (`src/gda/models.py:2427-2444`, `:2477-2494`) and live `GameGetResult` delegates through `NodeProperty` (`src/gda/models.py:2681-2693`). The implementation brief should name the model/schema/docs/tests that must change, or explicitly say why `value: Any` plus ADR prose is the intended stable contract. Otherwise gda-mcp and schema consumers get a weaker contract than the ADR promises.

Required - Correctness / scope:
- Single-source the projection surface before implementation starts. ADR-0035 includes get surfaces, set echoes, `project list`, and live reads (`docs/adr/0035...md:22-25`), while the new CONTEXT glossary only names get/live reads (`CONTEXT.md:118-121`). Current code also shows `project list` already carries projected values, not just keys (`src/gda/ops/operations.gd:2195-2250`). This needs one self-consistent statement so future work updates every affected result shape and does not accidentally leave list/set echoes on the old projector.

Required - Correctness / future evolution:
- Define collision behavior for the generalized projection. The ADR says `Dictionary` keys are coerced to strings and emitted as a JSON object (`docs/adr/0035...md:27-29`), and inline Object descriptors put `type` plus storage properties at the same top level (`:35-38`). That is fine for the immediate InputMap case, but the ADR generalizes to every read surface and declares the shape public ABI. Please record what happens when two non-string Dictionary keys stringify to the same key, or when a whitelisted Object has a storage property named `type` or `resource_path`. A deterministic rule, a reserved-field rule, or a nested `properties` shape would prevent accidental last-write/clobber semantics from becoming ABI.

Consider - Readability / terminology:
- `descriptor` is already a loaded term in this repo (`Command descriptor` in `CONTEXT.md:210-218`). If these are output value shapes, `reference projection` / `inline value projection` may be easier for agents to distinguish from command registration descriptors.

Security / performance:
- No new runtime code lands in this PR. The main performance/security concern is deferred to the implementation: the whitelist and depth cap must be covered by tests that prove arbitrary live Objects still fall back to string and pathological containers remain bounded.

Verification reviewed: PR diff, #381/#380 specs, relevant ADRs, current `_jsonify` mirror, project-list implementation, and PR CI status. I did not rerun local tests for this docs-only review; GitHub reports Ruff, Pyright, unit/package CI green and Godot e2e skipped.

qinhaihong-red:
Thanks — this was a sharp review; all four **required** findings were valid (verified against the code/surface). Addressed in `8410a7a`. `ProjectGetResult`/`ListedProjectSetting` `value` was already `Any`; `project list` already carries a projected `value` (`operations.gd` `_jsonify`); `game call` is absent from the surface; and `InputEvent` inherits `Resource`, so the `resource_path` collision is real, not hypothetical.

### Standards / Spec

- **[Required] Output ABI not tied to the ADR-0004 `--schema`/model chain** — **fixed.** Added a *"Carrying the ABI into the `--schema` / model chain (ADR-0004)"* consequence: the top-level `value` stays `Any` and is recorded as a **deliberate, bounded ADR-0004 exception** (a value's shape isn't statically knowable), while the *stable* reference / inline value projection shapes must be **named result models** and the `value` descriptions on `ProjectGetResult` / `ListedProjectSetting` / `NodeProperty` / `GameGetResult` updated + schema/round-trip tested. The #381 body now names these surfaces as ACs.
- **[Required] `game call` not an accepted surface (ADR-0025)** — **fixed.** Verified no `game call` exists in `cli.py` / `command-catalog.md`; dropped it → `game get` / "any future live value-returning op" (phrased as future, not committed surface).
- **[Required] Projection surface not reconciled** — **fixed.** ADR-0035, the CONTEXT `Value projection` term, and the #381 body now state **one canonical scope**: every value emitted through `_jsonify` — `project`/`node`/`resource get`, the `set` echo, `project list` per-entry value, `scene get-exports`, and live `game get`. Corrected the #381 body's wrong *"project list unaffected (keys only)"* (it already projects each entry's value).
- **[Required] CONTEXT term omits set-echo / project list** — **fixed.** The glossary term now enumerates the full surface, matching the ADR.
- **[Consider] `descriptor` collides with the `Command descriptor` glossary term** — **adopted.** Renamed to `reference projection` / `inline value projection` (+ `string fallback`) across the ADR and the CONTEXT term; added `descriptor` to its `_Avoid_` list.

### Code Review / Quality

- **[Required] Executable contract, not prose-only** — same fix as the ADR-0004 item above.
- **[Required] Single-source the projection surface** — same reconciliation as above.
- **[Required] Define collision behavior** — **fixed.** Recorded reserved-key + collision rules: `type` is the discriminator, a reference projection additionally carries `resource_path` and that's how the two kinds are told apart; the inline value projection **excludes the `Object`/`Resource` base bookkeeping** (`resource_path`/`resource_name`/`resource_local_to_scene`/`script`) so a path-less `InputEvent` (which *is* a `Resource`) can't masquerade as a reference; a class's own `type` storage prop is shadowed (documented); non-string `Dictionary` keys colliding after stringification resolve last-in-iteration-order (deterministic). Kept the flat `value.events[0].keycode` shape and recorded the nested-`properties` alternative as rejected, since the exclusion already removes the ambiguity.
- **[Security/perf]** the whitelist-fallback and depth-cap bounds are covered by the #381 implementation ACs (non-whitelisted Object → `str()`; mirror test; bounded containers).

No runtime code in this PR; it stays docs-only (`ADR-0035` + the `Value projection` CONTEXT term). Implementation remains tracked by #381 (`ready-for-agent`), whose body was updated to match.

PR #384, opened 2026-07-02
fix(ops): gate reference-scan decode on extension before reading

### LINKED ISSUE #378: project find-references / find-unused-resources: reference scan UTF-8-decodes binary files, spamming Unicode parse errors
## Summary

Make the project reference scan skip non-reference-bearing files instead of UTF-8-decoding every file under `res://`, so binary artifacts (e.g. an exported `.app`/`.pck` under `build/`) no longer spam engine `Unicode parsing error` lines.

## Current behavior

`gda project find-references` (and `find-unused-resources`, which shares the same reference graph) walks the full `res://` tree — skipping only `.godot/` — collecting every leaf file with a non-empty extension (excluding `.import`/`.godot`/`.cfg`/`.uid`). For **each** collected path it reads the whole file as text via `FileAccess.get_file_as_string()`, and only *then* branches on extension to parse `.tscn`/`.tres` (`[ext_resource]`) and `.gd` (preload/load/extends/class-token) references. Binary artifacts inside the project directory — e.g. `build/PandaAdventure.app` (dylibs, `.pck`, `.icns`, the executable) — are decoded as UTF-8, producing one engine line per binary file:

```
Unicode parsing error: Byte d is not a correct continuation byte after f3
```

Reproduced on Panda Adventure: `build/PandaAdventure.app` contains 7 files → 7 identical `Unicode parsing error` lines per run. The stdout JSON result is correct; the cost is stderr noise plus wasted IO reading multi-MB binaries that can never carry a reference the parser understands.

## Desired behavior

The scan reads/parses only reference-bearing text formats and never lets the engine emit a per-file decode error for a binary artifact. Either approach satisfies this:
- restrict the *decode* to reference-bearing text extensions (`.gd`, `.tscn`, `.tres`, `.gdshader`, `project.godot`, …), or
- skip files whose UTF-8 decode fails, without the engine logging a per-file error.

Reference results over the text formats must be byte-for-byte unchanged.

## Key interfaces

- The per-file reader on the reference-scan path — the helper that calls `FileAccess.get_file_as_string()` for every walked path *before* dispatching on extension (`_collect_references_from()`), plus the shared graph builder behind `find-unused-resources` / `dependencies`. The extension gate (or a decode-safe read) must happen **before** the decode, not after.
- The tree walker / graph-eligibility predicate (`_collect_resource_paths()` / `_is_graph_resource_path()`): if the fix narrows what gets *decoded*, confirm `find-unused-resources` still reports leaf *asset* resources (an unreferenced `.png`) as unused candidates — those are enumerated as graph nodes but never need to be decoded. Skipping the decode must not drop an asset from the unused-candidate set.

## Acceptance criteria

- [ ] `gda project find-references <target>` on a project containing binary files under `build/` (e.g. an exported `.app`/`.pck`) emits no `Unicode parsing error` line to stderr.
- [ ] Reference results over text formats (`ext_resource`, `preload`, `load`, `class_reference`, …) are unchanged — covered by a test that runs the scan with a binary file present in the tree and asserts both clean stderr and unchanged references.
- [ ] `find-unused-resources` (same graph) emits no `Unicode parsing error` and still reports unreferenced asset resources correctly.

## Out of scope

- Adding new reference *kinds* or new parseable formats.
- Changing which files count as "unused" resources (the graph semantics).
- Reading inside binary containers (`.pck`, `.app`) for references.

## Blocked by

None — can start immediately.

---
_Found re-verifying #360's fix on Panda Adventure (gda main, editable install). Consolidated during triage (AI-assisted) into this single authoritative spec._

qinhaihong-red:
## Summary

`project find-references` UTF-8-decoded **every** file the tree walk collected before dispatching on extension, so binary artifacts inside the project (e.g. an exported `.app`/`.pck` under `build/`) produced one engine `Unicode parsing error` line each per run. The scan now dispatches on extension first: only the reference-bearing text formats (`.tscn`/`.tres`/`.gd`) are ever decoded.

## Changes

- `operations.gd`: `_collect_references_from` returns early via the existing `_has_outgoing_references` predicate (the `.tscn/.tres/.gd` set) **before** `FileAccess.get_file_as_string`, mirroring the ext-first shape `_outgoing_references_of` already had. No parser logic duplicated; parse branches unchanged byte-for-byte.
- The walker (`_collect_resource_paths`) and graph-eligibility predicate (`_is_graph_resource_path`) are untouched — assets remain graph nodes and unused-candidates; only the decode narrows.
- Tests (`test_e2e_project_analysis.py`): the `refgraph_project` fixture now plants a binary `build/fake.pck` (the issue's exact invalid-UTF-8 byte shape) and an unreferenced `orphan.png`, so **all** existing reference-result tests run with binaries present; two new tests assert clean stderr + exact unchanged reference set for `find-references`, and clean stderr + intact unused-candidate reporting (both the orphan asset and the binary) for `find-unused-resources`. Red-verified: the pre-fix code reproduced the exact `Unicode parsing error: Byte d is not a correct continuation byte after f3` spam.

One honest note vs the issue text: pre-fix, `find-unused-resources` did **not** reproduce the spam — its graph path already went through the ext-gated `_outgoing_references_of`. Its new test therefore serves as the clean-stderr regression plus the graph-universe guard.

## Verification

- Full suite in an isolated worktree: fast tier **1025 passed**; full e2e tier **326 passed, 1 skipped** (the documented Linux-only export-template skip), real Godot engine.
- `uv run ruff check` / `uv run pyright` clean.

Closes #378

### DIFF
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -2905,7 +2905,15 @@ func _resolve_ref_path(ref: String, base_dir: String) -> String:
 # a preload/load/extends names one of those paths, or — when the target is a
 # class_name — when the file uses the class token as an identifier. The context is
 # the matched line, trimmed, so an agent locates the reference without re-reading.
+# Dispatches on extension BEFORE reading (issue #378): only the reference-bearing
+# text formats (_has_outgoing_references' .tscn/.tres/.gd set) are ever decoded,
+# so a binary artifact in the walked tree (an exported .pck/.app under build/)
+# never hits the engine's UTF-8 decode and never spams a per-file "Unicode
+# parsing error" to stderr. The graph universe is unchanged — only the decode
+# narrows, mirroring the ext-first shape of _outgoing_references_of.
 func _collect_references_from(path: String, target_paths: Dictionary, target_class: String, references: Array) -> void:
+	if not _has_outgoing_references(path):
+		return
 	var ext := path.get_extension().to_lower()
 	var text := FileAccess.get_file_as_string(path)
 	if text.is_empty():
--- a/tests/test_e2e_project_analysis.py
+++ b/tests/test_e2e_project_analysis.py
@@ -106,7 +106,13 @@ script="plugin.gd"
 
 @pytest.fixture
 def refgraph_project(tmp_path):
-    """A fixture project with real cross-references for the analysis commands."""
+    """A fixture project with real cross-references for the analysis commands.
+
+    Also plants binary artifacts (issue #378): an exported-build stand-in with
+    invalid-UTF-8 bytes under ``build/`` and an unreferenced binary asset. The
+    scan must never UTF-8-decode them (no engine ``Unicode parsing error`` on
+    stderr) while still enumerating them as graph nodes.
+    """
     (tmp_path / "project.godot").write_text(PROJECT_GODOT, encoding="utf-8")
     (tmp_path / "main.tscn").write_text(MAIN_TSCN, encoding="utf-8")
     (tmp_path / "hero.tscn").write_text(HERO_TSCN, encoding="utf-8")
@@ -115,6 +121,16 @@ def refgraph_project(tmp_path):
     (tmp_path / "game_state.gd").write_text(GAME_STATE_GD, encoding="utf-8")
     (tmp_path / "icon.png").write_bytes(b"\x89PNG\r\n\x1a\n" + b"\x00" * 16)
     (tmp_path / "orphan.tres").write_text(ORPHAN_TRES, encoding="utf-8")
+    # An unreferenced binary ASSET: skipped decode must not drop it from the
+    # graph universe — it stays a find-unused candidate (issue #378).
+    (tmp_path / "orphan.png").write_bytes(b"\x89PNG\r\n\x1a\n" + b"\x00" * 16)
+    # A binary build artifact whose bytes are invalid UTF-8 (0xf3 lead byte
+    # followed by a non-continuation byte — the exact shape the issue-#378 repro
+    # decoded into "Unicode parsing error: Byte d is not a correct continuation
+    # byte after f3").
+    build = tmp_path / "build"
+    build.mkdir()
+    (build / "fake.pck").write_bytes(b"\xf3\x0d\x00\xff" * 8)
     addons = tmp_path / "addons" / "widget"
     addons.mkdir(parents=True)
     (addons / "plugin.cfg").write_text(PLUGIN_CFG, encoding="utf-8")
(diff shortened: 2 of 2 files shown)


qinhaihong-red:
## Review pass 1/2: Standards / Spec

### Standards

No findings. I found no hard documented-standard violations and no judgement-call standards concerns in either changed hunk.

- `src/gda/ops/operations.gd`: the extension-before-read guard uses the existing `_has_outgoing_references` boundary and preserves the reference graph universe, aligning with the project static-analysis contract in `CONTEXT.md`/ADR-0032: raw-source/static scans for `find-references`, no project-code execution, and shared behavior with related analysis paths.
- `tests/test_e2e_project_analysis.py`: the new tests stay in the existing `@pytest.mark.e2e` suite and directly cover the graph-consistency/real-engine behavior described by the file and ADR-0000's vertical-slice/TDD convention.

Residual caveat: I did not rerun full CI/e2e. I skipped tooling-enforced issues per the review-skill instruction; ADR-0029/ADR-0030 make Ruff/Pyright CI gates authoritative, and the PR's CI currently reports Ruff, Python tests/package build, and Pyright passing, with the Godot e2e job skipped.

### Spec

No findings.

The diff satisfies issue #378's acceptance criteria: `find-references` now gates on `_has_outgoing_references()` before `FileAccess.get_file_as_string()`, preserving the existing `.tscn`/`.tres`/`.gd` reference surface while avoiding binary decode noise. The added tests cover a binary file under `build/`, clean stderr, unchanged `preload` references, and `find-unused-resources` still reporting unreferenced binary assets.

I did not see unrequested scope creep in production behavior. Residual caveat: the test also asserts `res://build/fake.pck` remains an unused candidate; that matches the current graph semantics, so it intentionally preserves broad existing behavior rather than redefining what "unused resource" means.

Verification run: `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_e2e_project_analysis.py::test_find_references_skips_decoding_binary_artifacts tests/test_e2e_project_analysis.py::test_find_unused_resources_skips_decoding_but_keeps_binary_graph_nodes` passed locally in an isolated `/tmp` checkout (`2 passed`).

Summary: Standards findings: 0. Spec findings: 0. Worst issue: none.

qinhaihong-red:
## Review pass 2/2: Code Review and Quality

I did not find any issues requiring changes.

### Correctness

- The production change is narrowly placed before the only risky operation in `find-references`: `_collect_references_from()` now checks `_has_outgoing_references(path)` before `FileAccess.get_file_as_string(path)`, so binary leaf resources are never UTF-8 decoded on that path.
- The graph universe is still built by `_collect_resource_paths()` / `_is_graph_resource_path()`, and `find-unused-resources` still builds reachability from `_outgoing_references_of()`. That preserves the existing unused-resource semantics while removing the decode noise.
- Test coverage is well targeted: the shared `refgraph_project` fixture now always contains invalid UTF-8 under `build/`, so the existing analysis tests continue exercising the reference graph with binary artifacts present; the new tests assert clean stderr, unchanged `find-references` output for the covered text reference, and unchanged unused-candidate behavior for binary assets.

### Readability / Simplicity

- The implementation reuses the existing `_has_outgoing_references()` predicate instead of adding a parallel extension list.
- The added comments are longer than strictly necessary, but they document the exact behavior boundary that caused the bug: graph membership is broad, decode/parsing is narrow. That distinction is important enough here.

### Architecture

- The change fits the current static-analysis architecture: graph enumeration remains inclusive, while outgoing-reference parsing remains limited to formats the tool actually understands.
- No new abstraction, dependency, or command-surface behavior was introduced.

### Security / Performance

- No new trust boundary is introduced. The command already operates inside the trusted-project model, and this change reduces unnecessary reads of unparseable binary artifacts.
- Performance impact is positive for projects with exported builds or other large binary assets in-tree, because those files are skipped before decode.

### Verification

- Local isolated checkout: `env UV_CACHE_DIR=/tmp/caw-uv-cache uv run pytest tests/test_e2e_project_analysis.py` passed (`14 passed`).
- Focused local regression run for the two new tests passed (`2 passed`).
- `git diff --check main...HEAD` was clean.
- GitHub checks show Ruff, Pyright, and Python unit/package build passing; the Godot e2e CI job is skipped, so the local real-engine e2e run above is the relevant added signal.

Verdict: no requested changes from this quality pass.

PR #386, opened 2026-07-02
feat(ops): ADR-0035 value projection — compound Variants to structured JSON

### LINKED ISSUE #381: project get: Dictionary/Object-valued settings render as a Variant str() debug string, not structured JSON

qinhaihong-red:
## Summary

Implements ADR-0035: the shared, mirrored `_jsonify` grows into the read-side **Value projection**, so compound values — `Dictionary`, `Array` + the full packed-array family, and whitelisted value `Object`s — return structured, indexable JSON instead of a Variant `str()` debug string, uniformly across every value gda emits through it (`project`/`node`/`resource get`, the `set` echoes, `project list` / `scene get-exports` per-entry values, and the live `game get` read).

```
$ gda project get input/fire --json
… "value": {"deadzone": 0.5, "events": [{"type": "InputEventKey", "keycode": 74, …}]}
```

## Design (as pinned by the ADR)

- **Containers** project recursively (Dictionary → object with stringified keys, last-wins collision; arrays/packed arrays → JSON arrays).
- **Objects** render as one of three kinds: **reference projection** (`{type, resource_path}` for a `res://`-pathed Resource — the read-side mirror of ADR-0033, never inlined), **inline value projection** (whitelisted path-less value Objects — `InputEvent` subclasses initially — via the existing storage-property model, with the `Resource` base bookkeeping excluded and the `type` discriminator assigned last so it shadows), or the existing **`str()` fallback** (e.g. a live `Node`; plus an `is_instance_valid` guard for freed live Objects).
- Bounded by `JSONIFY_MAX_DEPTH := 16` on the compound arms only (scalars are never depth-stringified); no visited-set; always JSON-encodable. Scalars and the fixed-shape types are byte-for-byte unchanged.
- Both `.gd` files changed identically inside the `--- shared coercion ---` block; the mirror gate stays green. `HARNESS_VERSION` bumped 4 → 5 per repo convention so the daemon resyncs installed harness copies.
- Python side: named `ReferenceProjection` / `InlineValueProjection` models; `value` stays `Any` (the ADR's recorded bounded exception — a union type would materialize dicts as models and break rendering) with projection-aware descriptions on every projected value field. **Every emitted command schema now carries the two named projection shapes as the value field's `$defs`** (shared `json_schema_extra` hook), so `--schema` / gda-mcp consumers get them as named, machine-readable models — not prose.
- **The ADR-0033 Object-typed `node set` / `resource set` echo flows through the same projection**: setting a Resource by `res://` path echoes `{type, resource_path}` — the identical shape a subsequent `get` reads back. One shape for one stored value, whichever way it is observed.

## Review response (Pass 1 + Pass 2, all three findings addressed in 26e92cd)

1. **Schema exposure gap** → fixed: all 8 projected value fields (`NodeProperty`, `SceneExport`, `ProjectGetResult`, `ListedProjectSetting`, `NodeSetResult`, `ResourceSetResult`, `ProjectSetResult`, `GameSetResult`) attach the named projection schemas as `$defs`; `gda project get --schema` now emits them, with model-level and command-schema-level test coverage.
2. **Object set-echo asymmetry** → fixed: the echo now routes through `_jsonify` (reference projection), the former asymmetry note is gone from models/tests/catalog, and the e2e pins the projected echo shape on both `node set` and `resource set`.
3. **Stale `SceneExport.value` description** → fixed: description now names the recursive ADR-0035 projection; a schema test on `scene get-exports --schema` fails on stale prose and asserts the `$defs`.

## Verification

- Fast tier **1032 passed** (includes the byte-identity mirror gate and schema gates); both `.gd` files additionally engine-syntax-checked (`--check-only`).
- Targeted e2e **149 passed** across `test_e2e_project.py` / `test_e2e_resource.py` / `test_e2e_object_property.py` / `test_e2e_node.py` / `test_e2e_daemon.py`, with independent orchestrator re-runs (70 + 26 + 28 passed) — covering: InputMap action → indexable `deadzone`/`events[0].keycode` with bookkeeping keys absent, packed arrays → lists, Gradient `offsets`/`colors` structured, reference projection on ADR-0033 set echo AND read-back, live `game get` structured Dictionary + `Node`-value `str()` fallback, and scalar no-regression.
- Full e2e suite passed on Linux CI (workflow-dispatched with `run-e2e=true`).
- `uv run ruff check` / `ruff format --check` / `pyright` clean.

Closes #381

### DIFF
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -153,7 +153,8 @@ node set` and `gda resource set` accept a **`res://….tres` resource path** as 
 create` and `resource set` this completes the external sub-resource workflow with no new command
 (`resource create res://box.tres --type RectangleShape2D` → `resource set … --property size --value
 32,64` → `node set scene.tscn --node Col --property shape --value res://box.tres`). The result echoes
-`type` `"Object"` and `value` the assigned `res://` path (pass `--project` so `res://` resolves). This
+`type` `"Object"` and `value` the ADR-0035 **reference projection** (`{type, resource_path}`) of the assigned
+resource — the same shape a subsequent `get` reads back (pass `--project` so `res://` resolves). This
 is a **separate, headless-only** step from the shared coercion above — value coercion keys only off
 the Variant type, but resolving an Object needs the expected-class hint on the property-list entry — so
 assigning a Resource on the live `gda game set` is **out of scope** and the coercion mirror is
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -139,8 +139,11 @@ mutation integrity boundary above. The supported target types and the string for
 | `Color` | `#rrggbb` / `#rrggbbaa`, or 3–4 comma-separated floats in 0..1 (`r,g,b[,a]`) | `[r, g, b, a]` |
 
 Whitespace around a value or a component is tolerated. A property of any other type is still
-reported by `node get` (its value degrades to a string projection), but `node set` cannot coerce
-to it yet and refuses with `uncoercible_value` — the coercible set grows as later slices need it.
+reported by `node get` — compound values arrive structured through the shared value projection
+(ADR-0035): a `Dictionary` as a JSON object, an `Array`/packed array as a JSON array, an `Object` as
+a reference / inline value projection or the `str()` fallback (see [`project`](#project)) — but
+`node set` cannot coerce to it yet and refuses with `uncoercible_value` — the coercible set grows as
+later slices need it.
 
 **Object-typed property assignment by `res://` reference** (ADR-0033, #363): for an **Object-typed**
 property that expects a Resource (sub)class — e.g. a `CollisionShape2D`'s `shape` (`Shape2D`) — `gda
@@ -410,9 +413,14 @@ valid result.
 **Reading and writing settings** (established by #111): `gda project get SECTION/KEY` reads one
 setting by its full `section/key` name (e.g. `application/config/name`) and reports it as a
 `{setting, type, value}` triple — `type` is the setting's declared Godot type name and `value` its
-JSON projection, the **same projection** `node get` reports for a node property. A setting that does
-not exist is a clean `unknown_setting` error (exit 4), distinguishing a typo'd key from a setting
-genuinely holding null. `gda project set SECTION/KEY --value <string>` writes a setting, **coercing**
+JSON projection, the **same projection** `node get` reports for a node property. Compound values
+arrive **structured** (ADR-0035): a `Dictionary` projects to a JSON object, an `Array` or packed
+array to a JSON array, and an embedded `Object` renders as a **reference projection**
+(`{type, resource_path}` for a Resource with a `res://` path), an **inline value projection**
+(`{type, …storage properties}` for a whitelisted path-less value Object — `InputEvent` subclasses,
+e.g. the `InputEventKey`s of an `input/*` action), or the `str()` fallback for any other Object. A
+setting that does not exist is a clean `unknown_setting` error (exit 4), distinguishing a typo'd key
+from a setting genuinely holding null. `gda project set SECTION/KEY --value <string>` writes a setting, **coercing**
 the CLI value to the setting's **declared type** — read off the setting's current value, exactly as
 `node set` reads it off the node's property list — using the **same coercion rules** the node group
 established (#55; see "Property value coercion" under [`node`](#node)). It then persists
--- a/src/gda/harness/gda_harness.gd
+++ b/src/gda/harness/gda_harness.gd
@@ -1036,14 +1036,34 @@ func _type_name(type: int) -> String:
 	return type_string(type)
 
 
-# Project a Godot property value into JSON-safe form for the result payload
-# (issue #55). Scalars pass through; the packed value types node set supports
-# become flat number arrays so node get's output is exactly the projection node
-# set accepts back: Vector2 → [x, y], Vector2i likewise, Color → [r, g, b, a].
-# Any other type degrades to its string form rather than crashing JSON.stringify
-# on an unencodable Variant — node get reports the whole storage surface, but
-# only the coercible types claim a structured projection.
-func _jsonify(value: Variant) -> Variant:
+# The value projection's hard recursion depth cap (ADR-0035): a compound value
+# nested deeper than this degrades to its string form instead of recursing on.
+# Deliberately NO visited-set — references are not descended and non-whitelisted
+# Objects stop at str(), so on-disk stored values are acyclic trees; the cap is
+# the backstop against a pathological self-referential Dictionary live-side.
+const JSONIFY_MAX_DEPTH := 16
+
+# The Object/Resource base bookkeeping properties an inline value projection
+# excludes (ADR-0035): every InputEvent IS a Resource, so without the exclusion
+# a path-less value Object would emit an empty resource_path and masquerade as
+# a reference projection; the exclusion also drops noise.
+const JSONIFY_BOOKKEEPING_PROPS: Array[String] = [
+	"resource_path", "resource_name", "resource_local_to_scene", "script",
+]
+
+# The read-side Value projection (ADR-0035, grown from issue #55): render a
+# Godot Variant into the structured JSON a result's value field carries.
+# Scalars pass through; the fixed-shape value types node set supports become
+# flat number arrays so node get's output is exactly the projection node set
+# accepts back: Vector2 → [x, y], Vector2i likewise, Color → [r, g, b, a].
+# A Dictionary projects to a JSON object (keys stringified), an Array and the
+# packed-array family to a JSON array, each value re-entering the projection;
+# an Object renders as a reference projection, an inline value projection, or
+# the str() fallback (the TYPE_OBJECT arm below). Any other type degrades to
+# its string form rather than crashing JSON.stringify on an unencodable
+# Variant, and the depth cap bounds the recursion on the compound arms — so
+# the projection is always JSON-encodable.
+func _jsonify(value: Variant, depth: int = 0) -> Variant:
 	match typeof(value):
 		TYPE_NIL, TYPE_BOOL, TYPE_INT, TYPE_FLOAT, TYPE_STRING, TYPE_STRING_NAME:
 			return value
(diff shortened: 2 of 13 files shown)


qinhaihong-red:
## Review Pass 1: Standards / Spec

### Standards

Hard violations:

1. `src/gda/models.py` adds `ReferenceProjection` and `InlineValueProjection` as standalone models, but the affected command output schemas still expose `value: Any` with prose only. I verified `uv run gda project get --schema`; neither projection model appears as `$defs` / `$ref` in the output schema. That violates ADR-0035's requirement to carry the ABI into the `--schema` / model chain with named result models that are "documented and consumable rather than prose-only" (`docs/adr/0035-value-projection-compound-variants-to-structured-json.md`, lines 94-102), and it weakens ADR-0004's model-driven self-description contract. The current tests validate `ReferenceProjection.model_json_schema()` / `InlineValueProjection.model_json_schema()` in isolation, but they do not prove any command schema exposes them.

2. The PR documents and tests an Object-property set echo exception: `node set` / `resource set` return the raw `res://` string while a later `get` returns the ADR-0035 `ReferenceProjection`. That contradicts ADR-0035's accepted surface, which explicitly includes "the value echoed back by `node set` / `resource set`" in the uniform value projection (`docs/adr/0035-value-projection-compound-variants-to-structured-json.md`, lines 22-26). Either the set echo should use the same projection, or ADR-0035 / #381 need to be amended before merging this exception.

### Spec

1. Same schema exposure gap as above: issue #381 asks for the reference / inline projection shapes to be named result models exposed through `--schema` / gda-mcp rather than prose-only. The models exist, but affected output schemas still only describe the shapes in text.

2. Same Object set echo drift as above: issue #381 and ADR-0035 include `node set` / `resource set` echoed values in the shared projection surface, but the implementation keeps the ADR-0033 Object branch as a raw path string and codifies that asymmetry in tests.

3. `scene get-exports` is only partially updated. `SceneExport.value` still has the old schema text ("scalar ... Vector2 ... Color"), and `uv run gda scene get-exports --schema` emits that stale description even though issue #381 / ADR-0035 explicitly include the per-entry `value` of `scene get-exports` in the uniform projection surface. Please update the `SceneExport` value description and add schema coverage that would catch this.

Summary: Standards found 2 hard violations; Spec found 3 gaps. Worst issue: the public schema/model chain still leaves the new projection ABI prose-only, so schema-driven agents and gda-mcp cannot consume the new shapes as named models.

qinhaihong-red:
## Review Pass 2: Code Review and Quality

Verdict: Request changes.

Findings:

1. `src/gda/models.py`: the projection models are not reachable from the affected command schemas. This is a correctness / architecture issue for the public ABI: `ReferenceProjection` and `InlineValueProjection` validate on their own, but `gda project get --schema` still emits `value` as untyped `Any` with only prose and no projection `$defs` / refs. Because gda-mcp mirrors these schemas, schema-driven callers cannot discover or validate the new shapes. Please wire the stable projection shapes into the emitted command schemas, or explicitly update the accepted ADR/spec if the intended contract is prose-only.

2. `src/gda/ops/operations.gd:647` and `src/gda/ops/operations.gd:1666`: Object-typed `node set` / `resource set` still echo the raw `res://` path instead of the ADR-0035 reference projection. The PR documents this as an asymmetry, but the accepted ADR and issue include set echoes in the uniform value projection surface. This creates two output shapes for the same stored Resource value depending on whether the caller observes the set result or performs a follow-up get. Please either return `{type, resource_path}` from these set echoes or move the exception through the ADR/spec before merging.

3. `src/gda/models.py:407`: `SceneExport.value` still advertises the old scalar / Vector / Color projection in `scene get-exports --schema`. The implementation likely benefits from the shared `_jsonify`, but the public schema text is stale for one of the explicitly listed read surfaces. Please update the field description and add a schema test so this does not regress.

What I checked:

- Reviewed the new tests before the implementation.
- `git diff --check origin/main...pr-386`
- `uv run pytest tests/test_models.py tests/test_project_commands.py tests/test_render.py tests/test_harness_coercion_mirror.py` -> 112 passed
- `uv run pytest` on the six PR-critical e2e tests for project/resource/object-property/live projection -> 6 passed
- `uv run ruff check`, `uv run ruff format --check .`, and `uv run pyright` all passed

No additional security or performance concerns stood out. The `_jsonify` recursion cap, Resource-reference stop, and InputEvent whitelist are the right safety boundaries; the remaining blockers are public-contract consistency and schema coverage.

qinhaihong-red:
Review handled — all three findings confirmed valid and fixed in 26e92cd (finding → resolution):

1. **Projection models not reachable from command schemas** (Pass 1 Standards #1 / Spec #1, Pass 2 #1) → **fixed@26e92cd**. Every projected `value` field — `NodeProperty`, `SceneExport`, `ProjectGetResult`, `ListedProjectSetting`, `NodeSetResult`, `ResourceSetResult`, `ProjectSetResult`, `GameSetResult` — now attaches `ReferenceProjection` / `InlineValueProjection` schemas as the field's `$defs` through a shared `json_schema_extra` hook (`_projected_value_schema_extra`). `value` itself stays `Any` per the ADR's recorded exception (a union annotation would materialize matching dicts into model instances and change the runtime payload / break the human renderer), so the named shapes ride the emitted schema instead. Verified with your exact probe: `uv run gda project get --schema` → `output.properties.value.$defs` = `{ReferenceProjection, InlineValueProjection}`. Covered by `test_every_projected_value_field_exposes_the_named_projection_defs` (model level) plus command-schema asserts in `test_project_commands.py` / `test_schema_command.py`.

2. **Object set-echo asymmetry vs ADR-0035's uniform surface** (Pass 1 Standards #2 / Spec #2, Pass 2 #2) → **fixed@26e92cd**. Agreed: ADR-0035 post-dates ADR-0033 and explicitly lists the `node set` / `resource set` echoes in the uniform projection surface, and ADR-0033's text never pinned the raw-string echo (it was pinned only by tests). Both Object-branch echoes now route through `_jsonify(resolved)` and return the reference projection `{type, resource_path}` — the identical shape a subsequent `get` reads back. The asymmetry prose is removed from `models.py`, `test_e2e_object_property.py`, and the command catalog; the e2e now pins the projected echo on both commands. (Both edits sit outside the mirrored shared-coercion block, so the mirror gate is unaffected.)

3. **Stale `SceneExport.value` schema text** (Pass 1 Spec #3, Pass 2 #3) → **fixed@26e92cd**. The description now uses the shared recursive-projection text (`_VALUE_PROJECTION_DESC`), and `test_scene_get_exports_schema_emits_model_derived_contract_without_other_args` asserts the emitted `scene get-exports --schema` names the value projection and carries the `$defs` — a stale scalar-only doc fails the test.

Gates after the fix: fast tier 1032 passed (mirror + schema gates included), targeted e2e re-run green (`test_e2e_object_property.py` + `test_e2e_project.py` 28 passed), ruff/pyright clean. Full-suite Linux e2e re-dispatch is queued on the updated head.

PR #474, opened 2026-07-08
fix(game): report live set verification

### LINKED ISSUE #473: gda game set — live-set write-verify false-rejects edge-triggered/self-consuming script variables
## Context

Dogfooding from **Panda Adventure #441** (the editor debug palette). Filed as a **gda-tool** issue (unprefixed) per the demo's gda-feedback convention. Category: quality defect + small contract remedy.

## Problem

`gda game set`'s live write-verify path currently treats a script-variable set as failed when the value read back after `node.set(...)` still equals the value observed before the set.

The intent is sound: catch script variables that silently no-op, such as getter-only or computed variables. But the same observation also happens for a valid edge-triggered, momentary, or self-consuming script variable. Example: a `spawn` trigger whose setter fires a side effect and then reads back as its default value. The set did take effect, but `gda game set` reports `live_uncoercible_value` because the value did not stick.

This blocks a natural live-driving pattern: debug palette commands, input pulses, one-shot triggers, and other momentary runtime controls.

## Reproduction

Panda Adventure #441's debug palette `spawn` trigger hit this behavior. The workaround shipped there in commit `a21d24f`: latch the set value instead of self-resetting, so the read-back matches the verifier. It works, but it forces every edge-triggered live control to be reshaped just to satisfy `gda game set`'s verifier.

A minimal repro is a script variable whose setter increments a separate counter while the getter always returns the default value. `gda game set` returns `live_uncoercible_value`, but a follow-up `gda game get` in the same live session can observe that the counter changed, proving the setter side effect happened.

## Decision

Adopt **success + explicit verification signal**.

`gda game set` should not guess whether an unchanged read-back means "no-op" or "valid edge-triggered consumption." The harness cannot know that from the read-back alone. Instead, the success result should report the observation directly and let the caller interpret it.

Add `verified: bool` to the `game set` success result:

- `verified: true` when the value read back after the live set equals the coerced value requested by this command.
- `verified: false` when the value read back after the live set does not equal the coerced value requested by this command.

Keep `value` as the actual observed read-back value, not the requested value.

This means:

- Normal storage-property and script-variable sets return success with `verified: true`.
- Getter-only or no-op script variables return success with `verified: false`.
- Edge-triggered or self-consuming script variables return success with `verified: false`; callers can use domain knowledge or follow-up reads to interpret the side effect.

## Key interfaces

- `GameSetResult` schema: add the `verified` boolean and document that `value` remains the observed read-back value.
- Live `game set` harness behavior: downgrade the script-variable read-back-unchanged hard error to success with `verified: false` once coercion and the set call have completed.
- CLI rendering and generated `--schema` surface: include the new result field consistently.
- Packaged gda skill guidance: tell agents to inspect `verified`, especially for script variables used as triggers or live controls.
- ADR-0020 live-state consistency: add an Outcome note clarifying that `game set` exposes read-after-write verification as structured data rather than guessing whether unchanged read-back means failure.

## Acceptance criteria

- [ ] `GameSetResult` includes `verified: bool` in JSON output and in `gda game set --schema`.
- [ ] A normal live `game set` whose read-back equals the coerced requested value returns success with `verified: true` and the observed `value`.
- [ ] A script-variable set whose read-back does not equal the coerced requested value returns success with `verified: false` and the observed `value`; it must not return `live_uncoercible_value` solely because the value did not stick.
- [ ] The existing getter-only or no-op script-variable e2e expectation is updated from hard error to success with `verified: false`, preserving the diagnostic through the result field.
- [ ] A new e2e regression covers an edge-triggered/self-consuming script variable: `game set` returns success with `verified: false`, and a follow-up live `game get` proves the setter side effect occurred in the same engine session.
- [ ] CLI docs/help, the gda skill guidance, and ADR-0020 document the new contract.

## Out of scope

- Adding a `--fire`, `--no-verify`, or `--allow-unverified` flag.
- Adding the same `verified` field to headless `node set`, `resource set`, or `project set`; that can be a follow-up if symmetry becomes valuable.
- Trying to infer whether `verified: false` means a getter-only/no-op variable or a valid edge-triggered control. The harness should report the observation, not guess intent.

## Blocked by

None — can start immediately.

qinhaihong-red:
## Summary

- add `verified` to live `game set` success results while keeping `value` as the observed read-back value
- return success with `verified:false` for script variables whose read-back differs after a completed set, including getter-only/no-op and edge-triggered controls
- document the contract across README, command catalog, ADR-0020, CLI help, and the bundled gda skill
- bump the bundled harness version so installed projects resync to the new live-set behavior

Closes #473

## Verification

- `uv run --frozen pytest -m e2e tests/test_e2e_daemon.py::test_daemon_serves_game_get_set_round_trip tests/test_e2e_daemon.py::test_daemon_game_set_readonly_script_variable_reports_unverified_observed_value tests/test_e2e_daemon.py::test_daemon_game_set_edge_trigger_reports_unverified_then_side_effect_is_readable tests/test_e2e_daemon.py::test_daemon_start_re_syncs_harness_after_a_version_bump -rs`
- `uv run --frozen ruff check .`
- `uv run --frozen ruff format --check .`
- `uv run --frozen pyright`
- `uv run --frozen pytest tests/test_harness_install.py tests/test_readme_i18n_sync.py tests/test_game_commands.py tests/test_models.py::test_game_set_result_round_trips_observed_value_and_verification_signal tests/test_render.py::test_render_game_set_renders_the_set_runtime_property tests/test_skill_command.py::test_skill_documents_game_set_verified_signal tests/test_schema_command.py::test_game_get_rect_set_schemas_report_kind_live_and_are_model_derived`
- `uv run --frozen pytest -m "not e2e"`
- `uv run --frozen pytest examples/platformer/panda-adventure/tests -m "not e2e and not engine"`
- `uv build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,7 +485,11 @@ a missing file still use the normal Error envelope.
 | `game set` | Set a runtime node property, or an explicitly named attached-script variable, on the running game. |
 
 Live `game set --property position` follows the same `Control` policy as
-`node set`; `game rect` remains a read-only rendered-geometry query.
+`node set`; `game rect` remains a read-only rendered-geometry query. `game set`
+success results include `verified`: `true` when the observed read-back value
+matches the coerced requested value, and `false` when the set completed but the
+observed value differs, such as getter-only/no-op script variables or
+edge-triggered controls.
 
 **`diag`** — runtime diagnostics
 
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=982e7a5aa9ff2150490487576a0d22ee2a3348988af1e56d1adc0a3d00f54264 -->
+<!-- gda-readme-i18n: source=README.md sha256=d0624578d2e00805a61ea7e983ce98c850dcafa07848468c961c82d4acec270e -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -496,7 +496,10 @@ normal.
 
 `game set --property position` en vivo sigue la misma política de `Control` que
 `node set`; `game rect` sigue siendo una consulta de geometría renderizada de solo
-lectura.
+lectura. Los resultados exitosos de `game set` incluyen `verified`: `true` cuando
+el valor observado al leer de vuelta coincide con el valor coercionado solicitado,
+y `false` cuando el set se completó pero el valor observado difiere, como en
+variables de script getter-only/no-op o controles edge-triggered.
 
 **`diag`** — diagnósticos de runtime
 
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=982e7a5aa9ff2150490487576a0d22ee2a3348988af1e56d1adc0a3d00f54264 -->
+<!-- gda-readme-i18n: source=README.md sha256=d0624578d2e00805a61ea7e983ce98c850dcafa07848468c961c82d4acec270e -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -501,6 +501,9 @@ offset プロパティを明示的に設定してください。
 
 Live の `game set --property position` は `node set` と同じ `Control` ポリシーに
 従います。`game rect` は引き続き読み取り専用のレンダリングジオメトリ問い合わせです。
+`game set` の成功結果には `verified` が含まれます。観測された読み戻し値が
+要求された変換済み値と一致する場合は `true`、set は完了したが観測値が異なる場合
+（getter-only/no-op のスクリプト変数やエッジトリガー制御など）は `false` です。
 
 **`diag`** — ランタイム診断
 
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=982e7a5aa9ff2150490487576a0d22ee2a3348988af1e56d1adc0a3d00f54264 -->
+<!-- gda-readme-i18n: source=README.md sha256=d0624578d2e00805a61ea7e983ce98c850dcafa07848468c961c82d4acec270e -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -481,7 +481,10 @@ Cursor 没有 `mcp add` 命令——请通过上面的 JSON 或 Settings → MCP
 | `game set` | 在正在运行的游戏上设置运行时节点属性，或显式命名的附加脚本变量。 |
 
 Live `game set --property position` 遵循与 `node set` 相同的 `Control` 策略；
-`game rect` 仍然是只读的渲染几何查询。
+`game rect` 仍然是只读的渲染几何查询。`game set` 的成功结果包含
+`verified`：当观测到的读回值匹配本次请求的已转换值时为 `true`；当 set
+已完成但观测值不同（例如 getter-only/no-op 脚本变量或边沿触发控制）时为
+`false`。
 
 **`diag`** — 运行时诊断
 
--- a/docs/adr/0020-live-state-consistency.md
+++ b/docs/adr/0020-live-state-consistency.md
@@ -28,6 +28,14 @@ and writes guarantee".
 > a window exactly as it holds for a single-frame op; the window only sequences N such
 > coherent snapshots.
 
+> **Outcome (2026-07-08, #473) — `game set` exposes read-after-write verification
+> as structured data.** A completed live `game set` returns the observed read-back
+> `value` plus `verified`, true only when that value equals the coerced requested
+> value. A mismatch is no longer guessed as failure by the harness: getter-only/no-op
+> script variables and edge-triggered/self-consuming controls both return success with
+> `verified:false`, leaving interpretation to the caller and preserving same-session
+> follow-up reads for domain-specific side effects.
+
 ## Decision
 
 `state consistency` is the property [gda-daemon](../../CONTEXT.md) provides over one
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -245,9 +245,13 @@ as `node set` round-trips through `node get`; here `unknown_property` names a pr
 **running** node's runtime property (the gda harness carries a verbatim copy of the coercion helpers,
 kept in sync by a drift test). When a `game get` / `game set` property name is explicit, the harness
 checks storage properties first, then attached-script variables; unfiltered `game get` keeps the
-storage-property listing and does not dump plain script variables. Live set round-trips through
-`game get`; its failures are the LIVE-category `live_unknown_property` / `live_uncoercible_value`
-(the harness reports them in-band, mapped by the live classifier — see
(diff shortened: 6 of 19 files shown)


qinhaihong-red:

Fixed point: `github/main` (`0e8ba56278fc87451ceb90d52b819a393a9bf8c1`)
Head reviewed: `724a262e345a941a7b0a22c41abe1d4c5797d125`
Diff command: `git diff github/main...pr-474`

### Standards

Required: `examples/platformer/panda-adventure/addons/gda_harness/gda_harness.gd` now declares `# gda-harness-version: 6`, but its body is not the bundled v6 harness from `src/gda/harness/gda_harness.gd`. It includes the new `verified` field but still lacks current bundled behavior such as the `Control.position` live-set policy branch/helpers and the newer mouse sequence support. ADR-0018 / the harness installer define the installed harness content as the version header plus the bundled body, with content comparison used for self-sync. A checked-in example harness that claims v6 while carrying a stale body creates exactly the kind of version/body drift that the harness version is meant to avoid. Please regenerate this committed installed copy from the bundled harness plus the header, or leave it intentionally stale without claiming the new harness version.

### Spec

Required: `GameSetResult.value` still has schema documentation drift. Issue #473 says to keep `value` as the actual observed read-back value, not the requested/coerced value. `src/gda/models.py` starts the field description with the new observed-readback wording, but still appends `_SET_ECHO_VALUE_DESC`, which says "The coerced value as JSON". As a result, `gda game set --schema` currently documents `value` as both observed read-back and coerced value.

Required: the Panda Adventure harness drift above also affects the #473 dogfood surface. The issue came from Panda Adventure's live debug palette, and this PR touches that checked-in harness copy, but the file now claims v6 without matching the v6 source harness. The #473 `spawn` behavior itself appears covered, but the changed dogfood harness surface should not drift from the bundled harness it claims to represent.

Verification I ran:
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pytest tests/test_harness_install.py tests/test_readme_i18n_sync.py tests/test_game_commands.py tests/test_models.py::test_game_set_result_round_trips_observed_value_and_verification_signal tests/test_render.py::test_render_game_set_renders_the_set_runtime_property tests/test_skill_command.py::test_skill_documents_game_set_verified_signal tests/test_schema_command.py::test_game_get_rect_set_schemas_report_kind_live_and_are_model_derived`
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pytest -m e2e tests/test_e2e_daemon.py::test_daemon_serves_game_get_set_round_trip tests/test_e2e_daemon.py::test_daemon_game_set_readonly_script_variable_reports_unverified_observed_value tests/test_e2e_daemon.py::test_daemon_game_set_edge_trigger_reports_unverified_then_side_effect_is_readable tests/test_e2e_daemon.py::test_daemon_start_re_syncs_harness_after_a_version_bump -rs`

Summary: Standards found 1 issue; Spec found 2 issues. Worst issue: the checked-in Panda Adventure harness now claims harness v6 while its body is not the bundled v6 harness.

qinhaihong-red:

Verdict: Request changes.

Required: correctness / public schema documentation drift. `GameSetResult.value` is meant to be the observed read-back value after a live set, and the implementation/docs mostly say that. The model field still appends `_SET_ECHO_VALUE_DESC`, whose text says "The coerced value as JSON", so the generated `gda game set --schema` output currently contradicts the new contract by describing the same field as both observed read-back and coerced/requested value. This is a public agent-facing schema surface, so please split or reword the shared description instead of reusing the set-echo helper for `GameSetResult.value`.

Required: architecture / drift control. The checked-in Panda Adventure harness copy was bumped to `# gda-harness-version: 6`, but it is not the materialized v6 bundled harness. A diff against `src/gda/harness/gda_harness.gd` shows it still lacks current bundled behavior such as the `Control.position` live-set branch/helpers and newer mouse sequence support. Since the installer defines the installed harness as the version header plus the bundled body, this creates a misleading version/body state in the dogfood example. Please regenerate that file from the bundled harness, or avoid changing the example copy to claim v6.

Consider: add drift coverage for the two consistency surfaces above. The current test suite passes while `GameSetResult.value` has contradictory generated schema text and while the Panda Adventure harness can claim the current harness version with a stale body. A focused schema-description assertion and/or an example-harness materialization check would make this class of drift harder to reintroduce.

No security or performance issues found in this pass. The core live behavior looks aligned with #473: normal sets return `verified:true`, getter-only/no-op script variables return success with `verified:false`, and the edge-triggered e2e proves the side effect through a same-session follow-up `game get`.

Verification I ran:
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen ruff check .`
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen ruff format --check .`
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pyright`
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pytest -m "not e2e"` (1089 passed)
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pytest examples/platformer/panda-adventure/tests -m "not e2e and not engine"` (314 passed)
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pytest -m e2e tests/test_e2e_daemon.py::test_daemon_serves_game_get_set_round_trip tests/test_e2e_daemon.py::test_daemon_game_set_readonly_script_variable_reports_unverified_observed_value tests/test_e2e_daemon.py::test_daemon_game_set_edge_trigger_reports_unverified_then_side_effect_is_readable tests/test_e2e_daemon.py::test_daemon_start_re_syncs_harness_after_a_version_bump -rs`
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv run --frozen pytest -m e2e tests/test_e2e_daemon.py::test_daemon_game_set_mutates_explicit_plain_script_dictionary_variable tests/test_e2e_daemon.py::test_daemon_game_set_preserves_json_container_integer_and_float_types -rs`
- `env UV_CACHE_DIR=/tmp/gda-pr474.9UTcvC/uv-cache uv build`

qinhaihong-red:
Addressed the requested changes in `972ba99`.

- Removed `examples/platformer/panda-adventure/addons/gda_harness/gda_harness.gd` from the PR diff. The example installed harness is no longer modified by #474; the bundled harness version bump remains in `src/gda/harness/install.py` so real `gda daemon start` installs/resyncs the packaged v6 harness.
- Split the live `GameSetResult.value` schema description away from the shared set-echo helper, so `gda game set --schema` now describes `value` only as the observed read-back value. Added schema assertions to guard this.

Verification:
- `uv run --frozen pytest tests/test_harness_install.py tests/test_readme_i18n_sync.py tests/test_game_commands.py tests/test_models.py::test_game_set_result_round_trips_observed_value_and_verification_signal tests/test_render.py::test_render_game_set_renders_the_set_runtime_property tests/test_skill_command.py::test_skill_documents_game_set_verified_signal tests/test_schema_command.py::test_game_get_rect_set_schemas_report_kind_live_and_are_model_derived` (44 passed)
- `uv run --frozen pytest -m e2e tests/test_e2e_daemon.py::test_daemon_serves_game_get_set_round_trip tests/test_e2e_daemon.py::test_daemon_game_set_readonly_script_variable_reports_unverified_observed_value tests/test_e2e_daemon.py::test_daemon_game_set_edge_trigger_reports_unverified_then_side_effect_is_readable tests/test_e2e_daemon.py::test_daemon_start_re_syncs_harness_after_a_version_bump -rs` (4 passed)
- `uv run --frozen ruff check .`
- `uv run --frozen ruff format --check .`
- `uv run --frozen pyright`

I also confirmed the final PR diff no longer includes the Panda Adventure installed harness file.

PR #482, opened 2026-07-08
fix(panda-editor): revert #473 spawn-latch workaround now gda 0.8.0 ships game-set verified

qinhaihong-red:
## Summary

Dogfooding follow-up closing the loop on **#473** (fixed by **#474**, shipped in **gda 0.8.0**).

`gda game set` now reports a live set whose read-back does not stick as **success with `verified:false`** (keeping `value` as the observed read-back), rather than the old hard `live_uncoercible_value` error. That removes the reason the palette's momentary `spawn` trigger was reshaped into a latch in `a21d24f`.

- **Revert the workaround.** `DebugPalette.spawn` returns to its natural momentary self-consuming pulse: setting it `true` spawns one enemy and self-resets to `false`, so `gda game get … spawn` honestly reads `false` (nothing stays armed). The overlay's "Spawn Enemy" is a momentary `Button` (no persistent checkbox binds `spawn`'s read-back), so the latch bought no UX — it was purely the old-gda compromise. The self-reset never recurses (an inline setter's assignment to its own backing var does not re-invoke the setter).
- **Pin both arms of the #473 contract** in the palette e2e: the momentary `spawn` set now asserts **`verified:false`** with the side effect still observable (`DebugSpawn0` in the tree + a `debug_spawn` record), while a latched state var (`god_mode`) asserts **`verified:true`**.

## Verification (gda 0.8.0 · Godot 4.6.3-stable, local)

- `gda game set --schema` → `GameSetResult` carries the required `verified: boolean`; `value` documented as the observed read-back.
- `gda script validate src/editor/debug_palette.gd` → `valid: true`
- `uv run --frozen ruff check …/test_editor_palette_e2e.py` → clean
- Palette e2e (the live #473 reproduction) → **1 passed** (`spawn` set: success + `verified:false` + `value:false` + spawn landed; `god_mode` set: `verified:true`)
- `uv run --frozen pytest examples/platformer/panda-adventure/tests -m "not e2e and not engine"` → **381 passed, 1 skipped**

## Notes

No `Closes` — #473 is already closed by #474; this is a game-side cleanup enabled by that fix, referencing it for traceability.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #484, opened 2026-07-08
docs: ADR-0036 gda authors by res:// path not uid; GUI uid-ization note in game AGENTS.md

qinhaihong-red:
## Summary

Formalize a previously **unstated, emergent property** of gda into an explicit decision, and warn about the human-GUI coexistence boundary where the demo actually meets it. Surfaced dogfooding Panda Adventure during the gda 0.8.0 verification session.

**`docs/adr/0036-authored-scenes-reference-by-path-not-uid.md` (new)** — gda-authored `.tscn`/`.tres` reference dependencies by `res://` path with **no `uid://`**, ratified as a deliberate **read-uid / write-path asymmetry**:
- **Write side** — path-only, no uid. It's path-only *by construction* (UID assignment is editor-only — `ResourceSaver` in a headless process emits no uid) *and* by intent (byte-determinism; a self-contained, cache-independent committed source). Same editor-only-state gap ADR-0032 hit with the class cache; extends ADR-0033's write-side "reference by `res://` path".
- **Read side** — gda stays a full uid *consumer* (`invalid_uid`/`unknown_uid`/`no_uid_assigned` per ADR-0002; `uid://` path scheme per ADR-0006; `--scene <path|UID>` per ADR-0017). Consuming a uid ≠ emitting one.
- Path-only files **load/run/export identically** (the engine resolves `ext_resource` by path and falls back to path for a missing/stale uid). No correctness cost; the only trade-off is no editor rename-tracking, which a gda workflow doesn't rely on.
- Forward-looking note: watch Godot's growing uid investment; if path-only ever degrades, revisit via deterministic path-derived uids.

**`examples/platformer/panda-adventure/AGENTS.md` (convention note)** — opening the game/editor scenes in the desktop Godot editor to **view/run/inspect** is fine (touches only gitignored `.godot/`, `*.uid` → no tracked drift), but **saving** in the GUI uid-ifies the tracked files. Handling is **awareness + recovery** (`git checkout` the churn; the gda-authored form is the source of truth), **not a guard hook** — using the GUI is the user's autonomous choice and is deliberately **not forbidden** (ADR-0018 sole-driver stance: undefended, not prohibited).

## Notes

Docs only — no code, no gates affected. Separate from the #473 spawn-latch revert (PR #482); no overlap in touched files.

### DIFF
--- a/docs/adr/0036-authored-scenes-reference-by-path-not-uid.md
+++ b/docs/adr/0036-authored-scenes-reference-by-path-not-uid.md
@@ -16,9 +16,9 @@ editor-generated, machine-local `.godot/uid_cache.bin`. The Godot editor writes 
 gda-authored files carry **none of this**: bare `[gd_scene format=3]` headers,
 `ext_resource` referenced by `path` + a local `id`, no `uid://` anywhere. This was an
 **unstated, emergent property** — never a recorded decision — and the question surfaced
-dogfooding Panda Adventure (the human-GUI-coexistence question raised during the v0.8.0
-`game set --verified` verification session): should gda "align" its output to the
-editor's uid-ized form?
+dogfooding Panda Adventure (the human-GUI-coexistence question raised while verifying gda
+0.8.0's `game set` read-back `verified` result field, #473): should gda "align" its output
+to the editor's uid-ized form?
 
 Two facts settle it:
 
--- a/docs/adr/0036-authored-scenes-reference-by-path-not-uid.md
+++ b/docs/adr/0036-authored-scenes-reference-by-path-not-uid.md
@@ -1,0 +1,106 @@
+---
+status: accepted
+---
+
+# gda-authored scenes and resources reference by `res://` path, not `uid://` (read-uid / write-path asymmetry)
+
+gda authors `.tscn`/`.tres` files by driving Godot's own `ResourceSaver` inside a
+`godot --headless` process — a [Headless operation](../../CONTEXT.md) on an
+editor-never-opened [Trusted project](../../CONTEXT.md). Godot 4's **Resource UID**
+system (`uid://…`) exists to *"maintain references between resources, even when files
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Standards

No findings.

## Spec

- [P2] `docs/adr/0036-authored-scenes-reference-by-path-not-uid.md` describes the provenance as the v0.8.0 ``game set --verified`` verification session, but `verified` is not a `game set` option. The PR body only says this surfaced during the "gda 0.8.0 verification session"; current local help exposes `game set` options as `--property`, `--value`, `--json`, `--schema`, `--params-json`, `--godot`, and `--project`, while `gda game set --schema` exposes `verified` as an output field on `GameSetResult`. This command-like wording can send readers toward a nonexistent flag. Please reword it to something like "`game set` read-back verification" or "`game set`'s `verified` result-field verification" so the ADR does not drift from the live surface.

Validation notes:
- Reviewed `git diff origin/main...HEAD` at `6f2545ee0570316b00178b7b7a7918bd8866ed69`.
- `git diff --check origin/main...HEAD` passed.
- Checked the UID claims against Godot 4.6 docs via Context7 and a local Godot 4.6.3 headless `ResourceSaver.save` probe; the generated `.tscn` had path-only `ext_resource` entries and loaded successfully.

qinhaihong-red:
Thanks — handled the [P2] Spec finding.

- **[P2] `game set --verified` reads like a nonexistent flag** → **Fixed @76ef92f.** Reworded the provenance to *"gda 0.8.0's `game set` read-back `verified` result field, #473"*, so it names the `GameSetResult` output field (surfaced via `game set --schema`) rather than a `--verified` option. Confirmed it was the only command-like `--verified` occurrence across the ADR and the AGENTS.md note.

Appreciate the independent validation of the UID claims (Context7 Godot 4.6 docs + the local 4.6.3 headless `ResourceSaver.save` path-only probe) — that matches the decision's core premise. **Standards: no findings** noted.

PR #498, opened 2026-07-15
feat(skills): state skill — cross-session Backlog with filtered inheritance

qinhaihong-red:
## Why

Feedback loops insert new tasks/issues mid-flight; the preempted "Next up" items are unfinished work that spans sessions, and the `state` skill had no mechanism guaranteeing they survive — in practice they squatted in Next up's tail ("still pending from before: …") or silently vanished on an overwrite (one unverifiable carry-over item was dropped exactly this way last session).

## What

- **Template gains a `Backlog` section**: unfinished cross-session carry-over NOT selected for "Next up" — interrupted or parked items, anchored to tracker refs (`#N`) where they exist. Capped at 5 items; overflow means the item belongs in the tracker.
- **Positioning sharpened** (the agile division of labor): the issue tracker / milestones remain the unquestioned planning authority — they own *what* the work is, but do not prescribe execution order/manner (parallel issues, one issue across sessions, preemption). STATE.md is the *worker's daily report on execution*: results plus the current orchestration of tracked work. Never a second plan.
- **`overwrite, never append` kept**: the Backlog is re-emitted on each rewrite from its **filtered inheritance** — an item is carried forward unless there is *evidence* it is done or obsolete (the deliberate inverse of "omit rather than mislead", so unfinished work cannot silently vanish); planned work displaced by inserted tasks demotes into it; an item lives in "Next up" OR the Backlog, never both.
- **Line budget relaxed** from ≤10 to ≤15 (Backlog ≤5 one-line items).

## Notes

- Tracked home is `.agents/skills/state/SKILL.md` (`.claude/skills` is a symlink into it).
- The root `STATE.md` instance is gitignored (a local working aid) — migrated locally to the new template as the first instance; not part of this diff.
- Doc-only change: no code or test surface.

### DIFF
--- a/.agents/skills/state/SKILL.md
+++ b/.agents/skills/state/SKILL.md
@@ -1,18 +1,26 @@
 ---
 name: state
-description: Update STATE.md — the lightweight cross-session "daily report" of project progress. Rewrite (never append) the current milestone/phase, what this session completed or changed, pitfalls worth reusing, and the recommended next issues/tasks. Use at the end of a working session, when wrapping up, or when invoked as /state.
+description: Update STATE.md — the lightweight cross-session "daily report" of project progress. Rewrite (never append) the current milestone/phase, what this session completed or changed, pitfalls worth reusing, the recommended next issues/tasks, and the filtered-and-inherited backlog of unfinished cross-session items. Use at the end of a working session, when wrapping up, or when invoked as /state.
 ---
 
 # State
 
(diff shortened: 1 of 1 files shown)


qinhaihong-red:
Superseded by #499: re-typed `feat` → `fix` per the git-conventional-commits convention (branch `fix/state-skill-backlog`, commit re-authored accordingly); content identical.

PR #499, opened 2026-07-15
fix(skills): state skill — cross-session Backlog with filtered inheritance

qinhaihong-red:
Supersedes #498 (originally the same content re-typed `feat` → `fix` per the git-conventional-commits convention — the change repairs a defect in the skill's mechanism: unfinished cross-session work could silently vanish on rewrite. Review-driven commits since then have evolved the contract beyond #498; this body describes the current head).

## Why

Feedback loops insert new tasks/issues mid-flight; the preempted "Next up" items are unfinished work that spans sessions, and the `state` skill had no mechanism guaranteeing they survive — in practice they squatted in Next up's tail ("still pending from before: …") or silently vanished on an overwrite.

## What (contract at current head)

- **Template gains a `Backlog` section**: unfinished cross-session carry-over NOT selected for "Next up", shown as nested one-line items, each stamped `since YYYY-MM-DD` (the snapshot's only observable age) and anchored to tracker refs (`#N`) where they exist.
- **Positioning sharpened** (the agile division of labor): the issue tracker / milestones remain the unquestioned planning authority — they own *what* the work is, but do not prescribe execution order/manner (parallel issues, one issue across sessions, preemption). STATE.md is the *worker's daily report on execution*: results plus the current orchestration of tracked work. Never a second plan. `AGENTS.md` no longer duplicates the schema/line budget — it delegates the format contract to this skill.
- **Item sources and flow**: inherit-by-filtering from the previous Backlog (+ never-run Next up items); *demote* only unfinished work NOT selected for "Next up" (immediately resumable work stays in "Next up"); *park* newly accepted, never-started items deferred to a later session; one home per item — "Next up" or Backlog, never both.
- **Removal needs evidence** (the deliberate inverse of "omit rather than mislead"): an item leaves the Backlog only when done, obsolete, promoted into "Next up", or newly homed in the tracker.
- **Budgets that yield, never truncate**: the Backlog carries a 5-item budget resolved by promoting the OLDEST items into the tracker (the filed reference is the removal evidence). With no tracker or no write authorization, the item budget yields — every item is kept, visibly over budget — and the global ≤15-line budget yields with it by exactly those Backlog lines. Losing work is never a valid way to satisfy a budget.

## Notes

- Tracked home is `.agents/skills/state/SKILL.md` (`.claude/skills` is a symlink into it).
- The root `STATE.md` instance is gitignored (a local working aid) — migrated locally to the new template as the first instance; not part of this diff.
- Doc-only change: no code or test surface.

### DIFF
--- a/.agents/skills/state/SKILL.md
+++ b/.agents/skills/state/SKILL.md
@@ -1,18 +1,26 @@
 ---
 name: state
-description: Update STATE.md — the lightweight cross-session "daily report" of project progress. Rewrite (never append) the current milestone/phase, what this session completed or changed, pitfalls worth reusing, and the recommended next issues/tasks. Use at the end of a working session, when wrapping up, or when invoked as /state.
+description: Update STATE.md — the lightweight cross-session "daily report" of project progress. Rewrite (never append) the current milestone/phase, what this session completed or changed, pitfalls worth reusing, the recommended next issues/tasks, and the filtered-and-inherited backlog of unfinished cross-session items. Use at the end of a working session, when wrapping up, or when invoked as /state.
 ---
 
 # State
 
 Maintain `STATE.md` at the repo root: a lightweight **cross-session daily report** so the next
-worker learns in ≤10 lines where the project is, what the last session did, which pitfalls to
-reuse, and what to pick up next. This is the feedback loop that makes development experience
-**compound** instead of re-discovering project status every session.
+worker learns in ≤15 lines where the project is, what the last session did, which pitfalls to
+reuse, what to pick up next, and which unfinished items must not be forgotten. This is the
+feedback loop that makes development experience **compound** instead of re-discovering project
+status every session.
 
 `STATE.md` is **transient** ("where are we now / what next", overwritten each session) and
 **self-contained** — it is not the place for durable decisions, architecture, or a glossary, and
 must not duplicate them. It is **lazily created**: if it doesn't exist yet, the first run creates it.
+
+Where the project keeps a planning authority (an issue tracker, milestones), that authority owns
+**what** the work is; it does not prescribe the **order and manner** of execution — issues run in
+parallel, one issue spans sessions, inserted work preempts planned work. STATE.md is the
+**worker's report on execution**: results, plus the current orchestration of tracked work (what
+to start next, what was interrupted and parked). It records execution state; it never becomes a
+second planning authority.
 
 This skill is **self-sufficient and orthogonal** to whatever else a project happens to use (version
 control, an issue/task tracker, design docs). It assumes such tools *may* exist but depends on no
@@ -21,7 +29,8 @@ particular one — read whatever sources the project has, skip the rest.
 **Write sparingly — omit rather than mislead.** Every field in the template is **optional**. Only
 write a line when you have something **accurate** and **useful to the next session**. If a value is
 missing, stale, unverified, or you can't judge whether it helps, **leave it out** — a wrong or
-filler line is worse than a missing one. STATE.md must orient the next worker, never misdirect them.
+filler line is worse than a missing one. STATE.md must orient the next worker, never misdirect
+them. (The Backlog inverts this default for *removal* — see the workflow.)
 
 ## When to run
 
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Review — state skill Backlog semantics

Reviewed `main...f1b86ee` as a docs-only change, with emphasis on logical consistency, completeness, and workflow executability.

**Verdict: Request changes.**

### Standards

1. **Required — synchronize the STATE.md contract in `AGENTS.md`.**

   `.agents/skills/state/SKILL.md:8-10,84-99` changes the contract to ≤15 lines and adds Backlog, while `AGENTS.md:7` still specifies ≤10 lines and lists only the previous fields. This leaves the startup guidance and the writer skill with conflicting contracts. Please either update `AGENTS.md` or remove the duplicated schema/budget there and make the state skill the clear authority.

2. **Required — make tracker promotion observable and executable.**

   `SKILL.md:66-68` asks the worker to detect whether an item has remained in Backlog for "more than a couple of sessions", but the transient overwritten snapshot records no age. If issue creation is not authorized, a draft also has no tracker reference that can be "kept only". Define an observable promotion criterion and the representation before an issue actually exists.

### Spec

1. **Required — resolve the contradiction between filtered inheritance and the hard five-item cap.**

   `SKILL.md:59-68,102-106` requires every uncertain unfinished item to survive unless completion or obsolescence is evidenced, while also requiring Backlog to remain at ≤5 items. With five inherited entries plus one unexecuted old Next up, the rules have no valid result when no tracker exists or tracker writes are unavailable. Define transfer to the tracker as an explicit removal transition, including a deterministic selection rule and a no-tracker/no-authorization fallback; do not allow truncation to silently lose work.

2. **Required — do not demote immediately resumable work.**

   `:56-57` places the right immediate continuation in Next up, but `:63-65` says any work "started but left unfinished" moves to Backlog. An unfinished task that should resume first next session therefore receives contradictory destinations. Demotion should apply only when the item is interrupted/parked **and not selected for Next up**.

3. **Required — add an insertion path for newly parked, never-started work.**

   `:59-64` sources Backlog only from the previous Backlog, unexecuted previous Next up items, and work started this session but left unfinished. A newly discovered/accepted item that is deliberately parked for a later session, but was neither started nor selected for Next up, has no destination and can still disappear.

### Consider

- The template at `:90` shows one Backlog bullet, while the guardrail at `:105` permits up to five items and requires one line per item. Show a canonical multi-item shape (nested list or repeated field) so valid output is unambiguous.

### Verification

- `git diff --check d0d5e97...f1b86ee` passed.
- Ruff, Pyright, unit tests, and package build are green.
- Godot e2e is skipped, which is appropriate for this docs-only change.
- No security, performance, or dependency concerns.

**Summary:** Standards: 2 required findings. Spec: 3 required findings. The worst issue is that the core "unfinished work cannot silently vanish" guarantee is currently incompatible with the hard five-item cap.

qinhaihong-red:
Response to the review — all 5 Required + the Consider addressed @ 3186235:

**Standards 1 (AGENTS.md contract sync) → fixed by de-duplication.** AGENTS.md no longer restates the schema or line budget; it now points at the `state` skill as the single authority for the format contract — so the two can't drift again (the stronger of the two options offered).

**Standards 2 (observable, executable promotion) → fixed.** Every Backlog item is stamped `since YYYY-MM-DD` — the snapshot's only observable age — replacing the unexecutable "more than a couple of sessions" wording. Promotion is triggered by budget overflow (or clear substance) and selects **oldest-first** by that stamp. The "draft with no reference" case is gone: without authorization the item simply STAYS in the Backlog (see Spec 1).

**Spec 1 (filtered inheritance vs hard 5-item cap) → fixed.** The cap is now a *budget* with a defined removal transition: overflow promotes the oldest items into the tracker, and the filed reference is the evidence that lets the entry leave. With no tracker or no write authorization, **the budget yields** — every item is kept and the list stays visibly over budget (itself the signal to get items homed). Truncation is explicitly forbidden in both the workflow step and the guardrail: the no-silent-loss guarantee outranks the budget.

**Spec 2 (immediately resumable work) → fixed.** Demotion now applies only to unfinished work **not selected for "Next up"**; interrupted work that should resume next session goes to "Next up" (stated in both step 3's promote clause and step 4's demote bullet).

**Spec 3 (newly parked, never-started work) → fixed.** A third source, **Park**, covers a newly discovered/accepted item deliberately deferred without being started or selected — it now has a defined destination instead of falling between the lists.

**Consider (canonical multi-item shape) → done.** The template shows the Backlog as a head line plus nested one-line items, each with a since-stamp; the guardrail's "one line per item" now has an unambiguous referent.

The local (gitignored) root STATE.md instance was migrated to the final shape as the first example.

qinhaihong-red:
## Follow-up recheck — two items remain

Rechecked the author's response and the exact fix diff `f1b86ee...3186235`. Four of the five required findings are closed, the remaining finding is partially closed, and the previous template-format Consider is closed.

**Verdict: Request changes remains.**

### Remaining findings

1. **Required — make the global ≤15-line budget yield with the Backlog budget.**

   The fix correctly makes the five-item Backlog budget yield when there is no tracker or no authorization: `SKILL.md:71-76,115-119` requires keeping every item, visibly over budget, with one line per item. However, `:81-83` and `:109` still impose an unconditional ≤15-line limit on the whole STATE.md. Once enough items accumulate, the worker again cannot satisfy both rules without losing an item. Please explicitly make the global line budget yield in the same fallback, or define another lossless representation.

2. **Required — synchronize the PR body with `3186235`.**

   The PR description still says the content is identical to #498, the Backlog is "Capped at 5 items", and inheritance removes items only with completion/obsolescence evidence. The current head instead permits visible overflow, adds promotion/tracker-home removal transitions, and contains the review-driven changes absent from #498. Please update the PR body so the stated contract matches the implementation and history.

### Confirmed closed

- `AGENTS.md` now delegates the format contract to the state skill.
- Backlog age is observable through `since YYYY-MM-DD`, with oldest-first promotion.
- Immediately resumable work remains in Next up instead of being demoted.
- Newly parked, never-started work has an explicit insertion path.
- The template now shows a canonical multi-item Backlog shape.

### Verification

- Latest head remained `3186235b0c9f91ca5e8eba867ca33f53b9667ae1` throughout the recheck.
- The cumulative PR diff passes `git apply --check --whitespace=error-all`.
- Ruff, Pyright, unit tests, and package build are green; Godot e2e is appropriately skipped for this docs-only change.

No other prior finding remains open.

qinhaihong-red:
Response to the follow-up recheck — both remaining findings addressed:

**1 (global ≤15-line budget) → fixed @ 821e784.** The global line budget now yields in the same no-tracker/no-authorization fallback as the item budget: the file grows by exactly the over-budget Backlog lines, and a line-budget breach is never resolved by dropping a Backlog item. Stated at the rewrite step (workflow step 6) and the line-budget guardrail; the intro and the emitted template header now read "~15 lines" so the header is never false while over budget.

**2 (stale PR body) → fixed.** The body has been rewritten to describe the contract at the current head: the since-stamps, the demote/park/one-home flow, the evidence set for removal (done / obsolete / promoted to Next up / tracker home), and the yield-never-truncate budget semantics — including that #498-supersession note no longer claiming identical content, and no "capped at 5" wording.

Issue #518, opened 2026-07-15
[gda-balancing] CLI interface contract — research, bADR & review


qinhaihong-red:
## Parent

- PRD #501 (2026-07-15 addendum, item 2: the CLI interface style stays consistent with `gda`).

## What to build

The research → design → review pass that makes the family CLI-consistency requirement concrete — **the single authority every `[gda-balancing]` CLI-surface issue's style clause resolves to**. Study `gda`'s CLI-contract ADRs as reference input (structured output contract, command taxonomy and grouping, schema self-description, structured params input) and record the toolkit's own binding decisions as bADR(s):

- Command taxonomy: top-level groups/verbs covering the design/template, validation, simulation, and tuning surfaces.
- The success contract: one structured JSON result per invocation.
- The error contract: a machine-readable error envelope with stable error codes; the usage/refusal exit code distinct from pass/fail verdict exit codes.
- Config/logic separation conventions: config-file arguments, output paths, never mutating the input authority.
- Self-description: whether and how the surface describes itself (schema/help) for agent consumption.
- Determinism conventions: how seeds surface on the CLI.
- The registration/dispatch mechanism that makes the contract **structurally self-enforcing**: a single seam every command plugs into (so a conforming surface is the only easy path), plus a conformance test that walks the registered surface asserting envelope and exit-code behavior — the contract is enforced by architecture and tests, never by per-issue prose.

Deviations from `gda`'s conventions are allowed where the toolkit's domain genuinely differs — but each deviation is a recorded bADR decision, never drift.

## Acceptance criteria

- [ ] bADR(s) record the command taxonomy, result and error contracts, exit-code semantics, and the self-description decision.
- [ ] The single-seam registration/dispatch mechanism and its conformance-test approach are part of the recorded design.
- [ ] Each convention is marked adopted-from-gda or a recorded deviation with rationale.
- [ ] The glossary is updated for new terms; maintainer review approval is recorded on this issue.
- [ ] No implementation in this slice.

## Blocked by

None - can start immediately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

qinhaihong-red:
## Research provenance — CLI interface contract (non-normative)

This comment is provenance, not authority — decisions land in the bADRs; this documents what the research found, with every precedent provenance-labeled by ecosystem.

[...]

Issue #504, opened 2026-07-15
[gda-balancing] Schema core & validation end to end


qinhaihong-red:
## Parent

- PRD #501.

## What to build

Implement the Standard Schema home per the reviewed design (#503): load a Schema document into a typed model, refuse element-level bad input with typed machine-readable errors, and round-trip a validated document back out as Standard Schema JSON — all driven through the CLI. Ship the minimal golden fixture (the smallest valid Schema document) in the test suite.

## Acceptance criteria

- [ ] The **definition-time formula evaluators** (named-form interpreter, expression-tree walker; bADR-0003) are implemented behind an explicit public formula seam, and the definition-time evaluation vectors V2, V3, V16 of `libs/gda-balancing/docs/badr/0004a-normative-vectors.md` pass as executable tests (runtime evaluation stays with #510, which reuses this seam — bADR-0004 ownership split).
- [ ] A valid Schema document loads to a typed model and round-trips out as Standard Schema JSON via the CLI.
- [ ] Element-level bad input (a wrong-typed collection element, a formula referencing an undeclared attribute, a malformed formula) is refused with a typed error naming the offending element; the refusal exit code is distinct from the verdict.
- [ ] The CLI can emit the Standard Schema's machine-readable self-description, and document validation is mechanical against it.
- [ ] The minimal golden fixture lives in the test suite; tests assert on external behavior only — the CLI/JSON surfaces and the **public formula seam** (bADR-0003, the authorized external test boundary for the definition-time evaluation vectors V2/V3/V16) — never internals; seeded and deterministic.

## Blocked by

- #502
- #503
- #518 — the CLI interface contract governs this command surface.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

qinhaihong-red:
Removed `ready-for-agent`: this issue is blocked by the open design gates #503 and #518; the signal returns when both gates close (per PR #521 review).

qinhaihong-red:
Vector-ownership note (PR #521, bADR-0004): this issue's executable-test mandate covers the **validation vectors** V1–V6, V11, V12 of `docs/badr/normative-vectors.md` (funnel boundary: structural/semantic refusals, defaults equality, version dispatch) plus per-rule catalog fixtures. The **runtime evaluation vectors** (V7–V10, V13, V14) are owned by #510, the first evaluator/simulation slice — this issue does not implement an evaluator.

qinhaihong-red:
[...]

Issue #501, opened 2026-07-15
PRD: Balancing pipeline — standalone game numeric design & balancing toolkit


qinhaihong-red:
## Refactor supersession — 2026-09-06

The approved refactor in #865 supersedes any assumption that internal package, Operation, Template, or toolkit versions are formally released and permanently immutable. Before the toolkit reaches v1.0, those internal definitions may be changed or withdrawn. v1.0 is the earliest point at which a compatibility commitment may be adopted; it does not automatically establish one. Content integrity, admitted-request/session consistency, exact semantic dependencies, and honest provenance remain required.

The complete product scope remains owned by this PRD and #534. Existing gameplay, simulation, calibration, documentation, and Panda cutover issues are reused; #865 owns the refactor path and mandatory deletion gates rather than replacing this product PRD.

This dated update changes only the identified contract or scheduling clauses. The original requirements, issue identifiers, historical evidence, and still-valid acceptance scope remain below. It does not claim implementation or conformance completion.

---

> **Vocabulary supersession note (2026-07-18):** the Standard Schema design (PR #521) realizes this PRD's layered-attribute requirement as **named facet compositions** over orthogonal attribute facets (bADR-0002), and renames the modifier concept: a **Modifier** is a numeric operation inside a first-class **Effect** (bADR-0006) — "modifier pools" below reads as *effect pools*. The prose below is preserved as the point-in-time requirements record; the bADRs under `libs/gda-balancing/docs/badr/` are the design authority for these terms (see the dated addendum comment).

## Problem Statement

Designing and balancing a game's numbers is done ad hoc inside each game project. The existing balancing pipeline — embedded in the Panda Adventure demo — proved the concept (config-driven `validate`/`predict` over a Monte-Carlo encounter engine and a system-dynamics long-horizon engine), but it cannot carry the next phase:

- Its surface is two commands; there is no numeric *design* support (attribute systems, builds, growth, economy), no metrics beyond TTK/TTD-vs-targets, no scoring, and no tuning.
- Its "generic" combat model hardcodes one game's mechanic fields (the warp kit, time fields) — generalizing from a single consumer produced mediocre generality.
- As demo-embedded tooling it has no product identity: no own versioning, tests-as-contract, docs, or decision records, and its evolution is tied to the demo's cadence.

The next phase needs numeric design **before** game development, and quantitative balance validation **during** it, for more games than one demo.

## Solution

[...]

PR #526, opened 2026-07-19
chore(gda-balancing): bootstrap the workspace package

### LINKED ISSUE #502: [gda-balancing] Bootstrap the workspace package
## Parent

- PRD #501 (see also the 2026-07-15 addendum comment: naming, CLI style, skeleton location).

## What to build

Wire the pre-created gda-balancing package skeleton into the repo's uv workspace and CI as an independently versioned, tested package. End-to-end tracer: a minimal CLI entry (module invocation) whose version query returns one structured JSON result on stdout, with malformed usage refused via a machine-readable error envelope on stderr and a usage exit code distinct from success — proving packaging, tests, and CLI plumbing end to end. Adopt the repo's lint/format/type gates for the package. Add the game/engine-agnosticism isolation gate — no game identity vocabulary, no game/engine/`gda` imports, no stray per-game config — at the hardened standard (recursive, AST-level import walking).

## Acceptance criteria

- [ ] The package builds and its tests run in CI as a uv workspace member; the root `gda` package's build and tests are unaffected.
- [ ] A version query via the CLI returns one structured JSON result; malformed usage yields a machine-readable usage-error envelope (bADR-0008) with a distinct exit code.
- [ ] Lint/format/type gates cover the package.
- [ ] The package's version line is governed: registered with the repo's release automation, extending ADR-0008's single-authority rule to this workspace member, with the decision recorded; until then the committed version stays a `0.0.0` placeholder.
- [ ] The isolation gate fails on a game-identity term, a game/engine/`gda` import (recursive, AST-level), or a stray per-game config file.

## Blocked by

- #518 — the version tracer's result/usage-error envelope follows the CLI interface contract.

qinhaihong-red:
Closes #502

## What

Wires the pre-created `libs/gda-balancing` skeleton into the repo as an independently versioned, tested uv-workspace member, with the end-to-end version tracer implemented strictly against **bADR-0007…0011** (the CLI interface contract, #518) and the design adjudications recorded on this run of #502.

- **Workspace**: root gains `[tool.uv.workspace] members = ["libs/gda-balancing"]` only — the member is *not* a `gda` dependency, so gda's closure and built wheel are unchanged (`uv build` at root still emits only gda dists; verified). `uv.lock` diff is the member node alone (+17 lines).
- **CLI spine (bADR-0011)**: one frozen `CommandDescriptor` per command in a single registry; dispatch, per-command `--schema`, and the conformance harness are all projections of it. Hand-rolled descriptor-driven argv binder (no Typer — the contract inverts Click's defaults: bare invocation is `missing_command` exit 3 on stderr, never an exit-0 help dump). Handlers are pure; the dispatch tail owns emission; an unexpected exception is the sole path to `internal`/exit 4.
- **Tracer**: `gda-balancing version` / `python -m gda_balancing version` (both entry points registered, byte-identical) → `{"supported_schema_line": null, "toolkit_version": "0.0.0"}` canonically emitted (bADR-0005). `supported_schema_line` stays `null` until #504 lands the first validatable Standard Schema implementation; field names adjudicated (`toolkit_version` reuses bADR-0008's `reproduction.toolkit_version`).
- **Conformance harness**: walks the registry and asserts every applicable bADR-0008 row per command — success/canonical, usage→3, fault-injected internal→4 (empty stdout + exactly one schema-valid stderr envelope, `diagnostics` only under `--debug`), `--schema` with `error` byte-identical across the walk, deterministic-refuses-`--seed`, reserved names unoccupied — plus subprocess smoke for both entry points and stream separation.
- **Isolation gate** (hardened standard): recursive AST-level import gate (`gda`/engine/game roots), word-boundary game-identity vocabulary gate (the gate module is the package's sole sanctioned appearance of the terms — adjudicated), stray per-game-config gate. Red-tested: planting `import gda`, a game term, and a stray `.json` fails all three sub-gates.
- **Gates**: repo-wide ruff already covers the member; member-local `[tool.pyright]`/`[tool.pytest.ini_options]`; CI type-check/python jobs gain member steps; member build smoke uses its own out-dir so the `gda-distributions` artifact is untouched.
- **Release automation (new ADR-0037)**: member registered with release-please under ADR-0008's single-authority rule — own component + tag scheme (`gda-balancing-vX.Y.Z`), `separate-pull-requests`, and the release-PR lock-sync generalized to iterate all Release PR branches (`prs` output; dry-run verified). Root `.` outputs stay unprefixed per the action docs, so the gda build→publish tail is untouched. Publishing gda-balancing is explicitly deferred; committed version stays `0.0.0`.

## Notes for review

- Verified against release-please docs: the root `"."` package absorbs commits from anywhere in the repo, so the **non-releasing title discipline for gda-balancing PRs continues after registration** (this PR itself is `chore`); recorded in ADR-0037.
- bADR scope tension resolved (adjudicated): 0009 assigns `--schema` delivery to #504, but 0007 marks `version` "`--schema`-bearing" and the 0011 harness asserts it from the first command → shipped here.

## Verification

- `uv run --frozen pytest libs/gda-balancing/tests` — 25 passed (harness + envelope law + tracer + isolation gate).
- Root suite `-m "not e2e"` — 1089 passed; panda pure-Python tier — 475 passed.
- `ruff check .` / `ruff format --check .` / `pyright` / `pyright --project libs/gda-balancing` — all clean.
- `uv build --package gda-balancing` and root `uv build` — both succeed; root emits only gda dists.
- Real-run tracer: both invocation forms exit 0 with identical canonical stdout; bare/`bogus`/`--seed 1` → exits 3 with `missing_command`/`unknown_command`/`unknown_argument` envelopes on stderr, stdout empty; `--schema` and `--help` exit 0.
- Isolation gate red-test: 3 failed with planted violations, green after removal.


qinhaihong-red:
## Review pass 1 — Standards / Spec

Reviewed head: `49af258012d522f2195aafe9331afef1ac788925`  
Fixed point: `258f43d9d8f62cfde3c171505d0493f5f120fdb9`

## Standards

1. **Hard violation — refusal bound is missing from the published error schema.**  
   [`envelope.py:72-85`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/envelope.py#L72-L85) defines `refusals` with `minItems: 1` but no `maxItems: 1000`. Consequently `version --schema` advertises an error schema that accepts 1001 refusals; this was reproduced with `jsonschema.validate(...)`. This contradicts bADR-0004, **“Report-all, deterministic, bounded”** (“At most 1000 refusals”), and bADR-0008, which says the carrier preserves that ≤1000 report semantics. Add `maxItems: 1000` and a schema regression test.

2. **Hard forward-compatibility violation — the canonical emission seam can emit non-JSON numeric tokens.**  
   [`emit.py:17-19`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/emit.py#L17-L19) uses `json.dumps` with Python's default `allow_nan=True`; `canonical_json({"value": float("nan")})` emits `{"value": NaN}\n` (likewise infinities), which is not standards-compliant JSON. This conflicts with bADR-0005's JSON-first canonical emission contract and bADR-0004's requirement that non-finite evaluation outcomes become typed Evaluation refusals. Set `allow_nan=False` and pin NaN/±Infinity tests at this shared seam.

## Spec

1. **[P1] The #502 descriptor seam cannot represent the required handler outcome.**  
   bADR-0011 requires the handler to “return the output model … or a typed-refusal report” and the dispatch tail to map the latter to a refusal envelope / exit 2; this outcome field is what makes the descriptor “sufficient to dispatch.” Instead, [`handler` is restricted to `BaseModel`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/descriptors.py#L47-L52), and [`dispatch` unconditionally serializes every handler return as success / exit 0](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/dispatch.py#L122-L124); `EXIT_REFUSAL` is unused. Although bADR-0008 defers an end-to-end exit-2 command to #504, bADR-0011 says #502 lands the descriptor/registry skeleton and #504 extends it. Add an explicit success-or-refusal outcome type and dispatch branch now; its refusal fixture can remain deferred.

2. **[P1] The isolation gate does not cover the game-identity vocabulary named by the requirements.**  
   #502 requires the gate to “fail on a game-identity term”; PRD #501 names the concrete prior leak as “one game's mechanic fields (the warp kit, time fields).” The regex only rejects [`godot|gdscript|panda|gadr`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/tests/test_isolation.py#L38-L42), so source containing `warp`, `time_field`, or other known identity nouns passes. This is narrower than the hardened predecessor's item/actor vocabulary. Expand the denylist with the known identity vocabulary and add parameterized red tests proving each term fails.

No scope creep was identified.

**Summary:** Standards: 2 findings. Spec: 2 findings. Worst issue: the foundational descriptor/dispatch seam cannot yet express one of its two normative handler outcomes. **Verdict: Request changes.**

Verification: `git diff --check`; 25 package tests passed; member Pyright and Ruff passed; both member and root wheels built; CI is green at the reviewed head. The contract probes above expose gaps not covered by the current suite.

qinhaihong-red:
## Review pass 2 — Code quality and design

Reviewed head: `49af258012d522f2195aafe9331afef1ac788925`

### Required findings

1. **Correctness / consistency — canonical emission accepts non-JSON numbers.**  
   [`canonical_json`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/emit.py#L17-L19) leaves `allow_nan=True`; the real probe `canonical_json({"value": float("nan")})` returned `{"value": NaN}\n`. `NaN` and ±`Infinity` are not JSON, contradicting the JSON-first canonical seam and bypassing the one sanctioned non-finite Evaluation-refusal path. Use `allow_nan=False` and add seam-level probes for all three non-finite values.

2. **Correctness / completeness — the published closed Error-envelope schema is wider than the normative contract.**  
   In [`ERROR_ENVELOPE_SCHEMA`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/envelope.py#L60-L120), `refusals` has no `maxItems: 1000`, `path` accepts strings that are not RFC 6901 JSON Pointers, and the `usage` branch permits `reproduction` even though usage failures happen before stochastic execution starts. A probe validated both a 1001-entry refusal list with `path: "not-a-pointer"` and a usage envelope carrying a reproduction key. Encode the bound/pointer shape and remove `reproduction` from usage; pin negative schema tests.

3. **Architecture / orthogonality / extensibility — command identity is not actually single-authority.**  
   [`REGISTRY`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/commands/__init__.py#L9-L12) has no uniqueness validation, so duplicate `(group, command)` paths are accepted; dispatch silently chooses the first while help renders both. Separately, [`valid_argv`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/descriptors.py#L26-L35) repeats the descriptor's command path, and the success/usage/internal/seed harness rows invoke that copy rather than deriving the path from the descriptor. A drift probe registered an `other` descriptor with a raising handler but `valid_argv=("version",)`; its “valid” row still exited 0 without addressing `other`. Validate path uniqueness and store fixture argument tails only; every harness row should prepend the descriptor-derived path.

4. **Architecture / extensibility — the descriptor outcome seam is success-only.**  
   bADR-0011's sufficient handler contract is output-model **or typed-refusal report**, yet [`handler`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/descriptors.py#L47-L52) is typed only as `BaseModel`, and [`dispatch`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/src/gda_balancing/dispatch.py#L116-L124) unconditionally emits exit 0. The first #504 document command therefore must redesign the core seam instead of merely extending the registry. Land an explicit success/refusal outcome and exit-2 dispatch branch now; the first actual refusal fixture can remain with #504.

5. **Correctness / completeness — the isolation gate misses the known leak class it exists to prevent.**  
   PRD #501 names the previous generic-model failure as one game's warp/time-field mechanics, but the denylist only covers [`godot|gdscript|panda|gadr`](https://github.com/aigengame/godot-agent/blob/49af258012d522f2195aafe9331afef1ac788925/libs/gda-balancing/tests/test_isolation.py#L38-L42). A planted `SPACESUIT_DEFENSE` source passed all three isolation tests. Carry over the hardened predecessor's known item/actor/mechanic vocabulary—including the PRD's `warp`/time-field terms—and parameterize the red tests so each known identity term proves the gate.

### Axis assessment

| Axis | Result |
|---|---|
| Correctness | **Fail** — malformed JSON is possible; published envelope and isolation contracts are under-constrained. |
| Readability & simplicity | Pass — modules are cohesive and control flow is understandable. |
| Architecture | **Fail** — refusal is absent from the claimed central seam; registry identity can drift. |
| Security | Pass — default internal errors are sanitized; debug disclosure is explicit. |
| Performance | Pass — no material hot-path or unbounded-runtime concern in this bootstrap. |
| Consistency | **Fail** — implementation disagrees with bADR-0004/0005/0008/0011. |
| Completeness | **Fail** — normative outcome, envelope, and isolation cases are missing. |
| Orthogonality | **Fail** — command identity is duplicated between descriptor and fixture. |
| Extensibility | **Fail** — the next registered document command must modify foundational dispatch behavior. |

### Verification

- 25 member tests passed; member Pyright and Ruff passed.
- Member and root wheels built; the root wheel contains no `gda-balancing` files.
- `git diff --check` passed; CI is green at this head.
- Release-please root/path output handling and the multi-branch lock-sync loop were inspected; no additional blocking release-workflow finding.

**Verdict: Request changes.**

qinhaihong-red:
## Resolution map — review pass 1–2, addressed @ b42b023

All findings accepted; every reviewer probe was re-run against the new head and now behaves per contract.

| # | Finding (pass/axis) | Resolution |
|---|---|---|
| 1 | P1-Std-1 / P2-2 — `refusals` unbounded in the published schema | **Fixed @ b42b023** — `maxItems: 1000` via `REFUSAL_BOUND` (bADR-0004's bound now lives in the contract); boundary tests: exactly-1000 valid, 1001 rejected. |
| 2 | P1-Std-2 / P2-1 — canonical seam emits `NaN`/±`Infinity` | **Fixed @ b42b023** — `allow_nan=False` at the single seam; a non-finite value now raises → `internal`/exit 4 (the sanctioned non-finite path stays the upstream Evaluation refusal, bADR-0003/0004). New `test_emit.py` pins all three non-finite probes + canonical properties. |
| 3 | P1-Spec-1 / P2-4 — descriptor seam success-only, `EXIT_REFUSAL` unreachable | **Fixed @ b42b023** — handler outcome is now the full bADR-0011 contract: `BaseModel \| RefusalReport`; the dispatch tail maps a report to the `refusal` envelope on **stdout** / exit 2 (`refusal_envelope` builder carries refusals verbatim + `truncated`). Proven by injection in the harness (`test_refusal_outcome_maps_to_refusal_envelope_exit_2`), same seam as the internal row; the first real producer remains #504's funnel, as the review allowed. |
| 4 | P2-2 (extras) — `path` not pointer-constrained; `usage` permitted `reproduction` | **Fixed @ b42b023** — refusal `path` now carries the RFC 6901 pattern (`""` = whole document stays valid); `reproduction` removed from the usage branch (usage errors resolve at binding, before execution — a seed can never have been drawn; it stays legal on `refusal`/`internal`). Negative tests pin both. |
| 5 | P2-3 — registry identity not single-authority; fixture argv duplicates the path | **Fixed @ b42b023** — `build_registry()` rejects duplicate `(group, command)` paths (tested); `ConformanceFixtures.valid_argv` → `valid_args` (argument **tail** only) and every harness row derives the command path from the descriptor (`_valid_invocation`), so your drift probe is now unrepresentable by construction. |
| 6 | P1-Spec-2 / P2-5 — isolation vocabulary misses the named leak class | **Fixed @ b42b023** — two changes: (a) **identifier-aware boundaries** replace `\b` (which treats `_` as a word char — the reason `SPACESUIT_DEFENSE` passed; it also would have passed the predecessor gate): terms are now bounded by non-alphanumerics, catching bare/snake_case/SCREAMING_CASE forms. (b) Denylist extended with the demo's item/actor identity nouns (`spacesuit`, `laser`, `bun`, `wine`) and PRD #501's named leak fields (`warp`, `time_field`, `time_dilation` — the demo's actual field roots), plus `unity`/`unreal` per the #502 adjudication. Per-term parameterized red tests prove all four spellings of every term trip the gate. |

**Deliberate narrowing (flagged, not silent):** the predecessor gate's genre-generic mechanics words (`gravity`, `hud`, `obstacle`, `crate`, `pickup`, `boss`) are **excluded** — they are genre vocabulary the toolkit's genre templates and fixtures may legitimately speak (e.g. a survivors-like `pickup_radius`, `boss_encounter`), not one game's identity. Negative controls pin them as allowed. If you read PRD #501's leak class as requiring these too, say so and I'll revisit rather than let the disagreement sit silent.

Verification @ b42b023: 55 member tests (was 25; +30 covering every finding), ruff/format/pyright (root + member) clean, both entry points re-smoked; all six reviewer probes re-run and rejected/handled as required. CI running on the new head.

qinhaihong-red:
## Recheck — one blocking residual

Rechecked head: `b42b0230dc6f64986ddb2348dd2d3baaa31d2042`

The previous fixes close the original canonical-JSON, published-schema, registry-authority, fixture-orthogonality, and isolation findings. The refusal outcome seam is present, but its typed-outcome-to-envelope invariant is still incomplete.

### [P1] `RefusalReport` can produce schema-invalid stdout

[`Refusal.path`](https://github.com/aigengame/godot-agent/blob/b42b0230dc6f64986ddb2348dd2d3baaa31d2042/libs/gda-balancing/src/gda_balancing/envelope.py#L56-L64) is an unconstrained `str`, and [`RefusalReport.refusals`](https://github.com/aigengame/godot-agent/blob/b42b0230dc6f64986ddb2348dd2d3baaa31d2042/libs/gda-balancing/src/gda_balancing/envelope.py#L67-L76) has no 1–1000 model-level bound. [`refusal_envelope()`](https://github.com/aigengame/godot-agent/blob/b42b0230dc6f64986ddb2348dd2d3baaa31d2042/libs/gda-balancing/src/gda_balancing/envelope.py#L162-L170) carries either value verbatim, and dispatch emits it without validating the advertised error schema.

Direct probes:

```text
empty       constructed=true schema_valid=false
oversized   constructed=true schema_valid=false
bad_pointer constructed=true schema_valid=false

dispatch(empty report):
exit=2 stderr_empty=true schema_valid=false
payload={"error":{"category":"refusal",...,"refusals":[],"truncated":false}}
```

Therefore a handler can return a successfully constructed typed outcome while dispatch writes a payload that violates `ERROR_ENVELOPE_SCHEMA`. The current conformance test covers only one valid singleton report, so it does not prove the bADR-0011 outcome-to-invocation-result invariant.

Please enforce the RFC 6901 path constraint and the 1–1000 refusal bound in the Pydantic outcome models (sharing the existing constants), then add negative construction tests plus a dispatch invariant test. Validating only the manually assembled schema is insufficient because handlers consume the model types, not the schema.

Verification at this head: 55 member tests passed; member Ruff and Pyright passed; `git diff --check` passed; current CI is green. The isolated review clone remained clean.

**Verdict: Request changes.**

qinhaihong-red:
## Resolution map — recheck residual, addressed @ HEAD

| # | Finding | Resolution |
|---|---|---|
| 1 | [P1] `RefusalReport` can produce schema-invalid stdout (unconstrained `Refusal.path`, unbounded `refusals`, dispatch emits verbatim) | **Fixed** — the constraints moved onto the outcome models themselves, exactly as suggested: `Refusal.path` carries the RFC 6901 pattern and `RefusalReport.refusals` is bounded `1..REFUSAL_BOUND`, both **sharing the published schema's constants** so model and contract cannot drift. An invalid report is now unconstructible; since handlers consume the model types, whatever constructs emits schema-valid stdout. |

Evidence at this head:

- Your three construction probes now raise `pydantic.ValidationError`: `empty`, `oversized` (1001), `bad_pointer` — all `constructed=false`.
- New negative construction tests pin all three, plus at-bound (exactly 1000) constructs and validates.
- New dispatch invariant coverage: the conformance refusal row now drives **both constructible edges** — a singleton report with the whole-document pointer `""` and a 1000-refusal truncated report — asserting exit 2, empty stderr, and `ERROR_ENVELOPE_SCHEMA`-valid stdout each time.

Verification: 56 member tests passed; ruff/format + pyright (member) clean; probes re-run as above. CI running on the new head.

qinhaihong-red:
## Recheck 2 — the previous P1 is closed; one contract-consistency issue remains

Rechecked head: `0e30098ec1a17cb829a148d5593b3223a30b5ae6`

The previous residual is fixed: empty and 1001-entry reports, plus an ordinary non-pointer path, are now unconstructible; the 1000-entry boundary constructs and dispatches as schema-valid stdout. The focused Standards and Spec passes found no remaining issue with that typed-outcome-to-dispatch invariant.

### Required — sharing the regex text does not give Pydantic and JSON Schema identical pointer semantics

[`_JSON_POINTER_PATTERN`](https://github.com/aigengame/godot-agent/blob/0e30098ec1a17cb829a148d5593b3223a30b5ae6/libs/gda-balancing/src/gda_balancing/envelope.py#L52-L68) is reused by both the Pydantic model and the [published JSON Schema](https://github.com/aigengame/godot-agent/blob/0e30098ec1a17cb829a148d5593b3223a30b5ae6/libs/gda-balancing/src/gda_balancing/envelope.py#L118-L122), but the two regex engines do not give the terminal `$` anchor the same behavior.

Minimal probe:

```text
path = "\n"
Pydantic Refusal construction: rejected
ERROR_ENVELOPE_SCHEMA validation: accepted
```

Reproduction:

```python
payload = {
    "error": {
        "category": "refusal",
        "message": "m",
        "refusals": [{"code": "x", "path": "\n", "detail": "x"}],
        "truncated": False,
    }
}
assert Draft202012Validator(ERROR_ENVELOPE_SCHEMA).is_valid(payload)
```

A lone newline is not an RFC 6901 JSON Pointer: a non-empty pointer must start with `/`. Therefore the public schema is still wider than the normative bADR-0004/0008 contract, and the claim that the shared pattern prevents model/schema drift is not true at the semantic level.

Please encode the empty-string-or-leading-`/` shape explicitly in the published schema, alongside the existing token-escape constraint (or use another engine-stable formulation), and add this lone-newline case as a regression test for both model and schema behavior.

Verification at this head: 56 member tests passed; member Ruff and Pyright passed; `git diff --check` passed; current CI is green. The review clone remained clean.

**Verdict: Request changes.**

qinhaihong-red:
## Resolution map — recheck 2, addressed @ HEAD

| # | Finding | Resolution |
|---|---|---|
| 1 | Shared regex text ≠ shared semantics: Python `re`'s terminal-`$` trailing-newline leniency let a lone `"\n"` path pass the published schema while the model's Rust engine rejected it | **Fixed** — the published schema now encodes RFC 6901's empty-or-leading-`/` shape explicitly (`anyOf: [{"const": ""}, {"pattern": "^/"}]`) alongside the existing token-escape `pattern`, exactly the formulation you suggested — no reliance on `$` end-anchor semantics in any engine. |

Evidence at this head:

- Your minimal probe re-run: `Draft202012Validator(ERROR_ENVELOPE_SCHEMA).is_valid(<lone-"\n" path>)` → **False**; `Refusal(path="\n")` → rejected. Model and schema now agree on the divergent input.
- New regression test pins both sides on `"\n"`, and pins the sound counter-case: a newline **embedded inside a reference token** (`"/a\nb"`) stays legal on both sides — RFC 6901 permits any non-`/`/`~` character in a token, so the fix narrows exactly the illegal shape and nothing more.
- I also withdraw the earlier phrasing that the shared constant "prevents drift": it prevented *textual* drift only. The shape constraint plus the divergent-input regression is what now holds the two engines to the same semantics.

Verification: 57 member tests passed; ruff/format + pyright (member) clean. CI running on the new head.

qinhaihong-red:
## Architecture review — consistency, completeness, orthogonality, extensibility, and verifiability

Reviewed head: `ccfa8f07a19c65ff017fe79db8ca1228bb3c8a06`

The overall bootstrap architecture is strong: the workspace member is dependency-isolated from `gda`; handlers are pure; dispatch owns binding and emission; command identity is registry-derived; and the registry-driven conformance walk gives future commands a useful safety net. The recheck-2 pointer issue is not repeated here because this head addresses it separately.

| Dimension | Assessment |
|---|---|
| Consistency | Good direction, but the declared success model is not enforced by dispatch. |
| Completeness | Strong for the #502 bootstrap slice; real document producers and publishing are explicitly deferred. |
| Orthogonality | Package/handler/dispatch boundaries are clean; success schema and runtime outcome can still diverge. |
| Extensibility | Adding a descriptor is straightforward, but a future handler can silently violate its descriptor's output contract. |
| Verifiability | The conformance walk is a major strength; it needs one more invariant row for mismatched success outcomes. |

### [P1] `output_model` is descriptive, not authoritative at runtime

[`CommandDescriptor`](https://github.com/aigengame/godot-agent/blob/ccfa8f07a19c65ff017fe79db8ca1228bb3c8a06/libs/gda-balancing/src/gda_balancing/descriptors.py#L42-L58) declares both `output_model` and a handler whose success side is any `BaseModel`. [`dispatch`](https://github.com/aigengame/godot-agent/blob/ccfa8f07a19c65ff017fe79db8ca1228bb3c8a06/libs/gda-balancing/src/gda_balancing/dispatch.py#L127-L132) distinguishes only `RefusalReport`; every other `BaseModel` is emitted as success / exit 0 without checking it against `descriptor.output_model`.

Direct probe against this head, replacing the `version` handler with one returning an unrelated `Bogus(BaseModel)`:

```text
exit=0
stderr_empty=true
payload={"unexpected": 7}
advertised_output_valid=false
```

The handler return is accepted by the current annotation and dispatch, yet the result violates the same descriptor's published `output` schema. This leaves the bADR-0011 "handler returns the output model" law dependent on one representative success fixture rather than making it an invariant of the central seam. A conditional wrong-model branch in a future command can therefore emit schema-invalid success stdout even when registration itself is valid.

Please make the success branch reject outcomes that are not instances of `descriptor.output_model` before emission. The mismatch should follow the existing unexpected-exception path (`internal` / exit 4, empty stdout), and the conformance harness should inject a wrong success model to pin that behavior. A generic descriptor can improve static coupling later, but the inexpensive runtime invariant is the required closure now.

### Non-blocking, documented design debts

- **Model/schema semantic duplication:** the pointer fixes demonstrate that sharing constants does not by itself make Pydantic validation and the hand-written public schema one semantic authority. Consider deriving more of the schema from typed models or adding focused parity/property tests as new envelope fields arrive.
- **Release-train coupling:** ADR-0037 explicitly records that the root `"."` release-please package absorbs member commits, so member PRs rely on non-releasing title discipline while publishing is disabled. This is acceptable for #502, but the first member-release issue must mechanically isolate the tag gate/publish path before releasing commit types are enabled.

**Verdict: Request changes for the success-output invariant; retain the rest of the architecture.**

qinhaihong-red:
## Resolution map — architecture review, addressed @ HEAD

| # | Finding | Resolution |
|---|---|---|
| 1 | [P1] `output_model` descriptive, not authoritative: dispatch emitted any `BaseModel` as success / exit 0, so a wrong-model handler return could ship stdout violating the descriptor's own `--schema` | **Fixed** — the success branch now enforces `isinstance(outcome, descriptor.output_model)` before emission; a mismatch raises and takes the existing unexpected-exception path exactly as requested: `internal` / exit 4, stdout empty, sanitized envelope (the mismatch detail appears in `diagnostics` only under `--debug`). New per-descriptor harness row (`test_wrong_success_model_row`) injects a wrong success model and pins exit 4 + empty stdout + schema-valid `internal` envelope for every registered command. |

Evidence at this head — your probe re-run: replacing the `version` handler with a `Bogus(BaseModel)` return now yields `exit=4`, `stdout` empty, `internal`/`internal_error` envelope on stderr (was `exit=0` with `{"unexpected": 7}`).

**Non-blocking debts, acknowledged:**

- *Model/schema semantic duplication* — agreed; the recheck-2 lesson generalizes. Logged on #504 (the next schema producer): as envelope/schema surface grows, prefer deriving published schemas from the typed models or add engine-parity tests on divergence-prone inputs, rather than shared constants alone.
- *Release-train coupling* — already recorded as a hard precondition in ADR-0037: the member's first-release issue must wire its publish tail and extend the tag gate to its component **before** releasing commit types are enabled for `libs/` PRs. No change needed at this head.

Verification: 58 member tests passed; ruff/format + pyright (member) clean; CI running on the new head.

qinhaihong-red:
## Architecture recheck — success-output invariant remains partial

Rechecked head: `e852d47a37767b2a9a7253b470dadaa5cd1dbbd7`

The unrelated-`BaseModel` case is fixed: a `Bogus(BaseModel)` return now follows the existing `internal` / exit 4 path, with empty stdout, a sanitized non-debug envelope, and mismatch diagnostics only under `--debug`.

### [P1] An output-model subclass can still emit success stdout that violates the declared schema

[`dispatch`](https://github.com/aigengame/godot-agent/blob/e852d47a37767b2a9a7253b470dadaa5cd1dbbd7/libs/gda-balancing/src/gda_balancing/dispatch.py#L127-L141) checks:

```python
isinstance(outcome, descriptor.output_model)
```

That accepts subclasses of the declared model. Pydantic serializes fields declared by the runtime subclass, while the descriptor's published output schema is closed with `additionalProperties: false`.

Direct probe using `ExtendedVersionResult(VersionResult)` with one added `unexpected` field:

```text
isinstance=true
exit=0
stderr_empty=true
payload={
  "supported_schema_line": null,
  "toolkit_version": "0.0.0",
  "unexpected": 7
}
advertised_output_valid=false
```

The handler therefore still has a constructible path to exit-0 stdout that contradicts its descriptor's `output` schema. The [new conformance row](https://github.com/aigengame/godot-agent/blob/e852d47a37767b2a9a7253b470dadaa5cd1dbbd7/libs/gda-balancing/tests/test_cli_conformance.py#L89-L101) injects only an unrelated `Bogus(BaseModel)`, so it cannot catch this subtype loophole.

Please either:

1. require exact runtime model identity (`type(outcome) is descriptor.output_model`), or
2. strictly revalidate and normalize through `descriptor.output_model` before emission, rejecting extra subclass fields.

Add an extra-field subclass regression proving the mismatch yields `internal` / exit 4 with empty stdout. This is the remaining closure needed for bADR-0011's “handler returns the output model” law and the descriptor's closed output schema to be the runtime authority.

Verification at this head: 58 member tests passed; member Ruff and Pyright passed; `git diff --check` passed; current CI is green. The ordinary mismatch non-debug/debug paths behave correctly.

**Verdict: Request changes.**

qinhaihong-red:
## Resolution map — architecture recheck, addressed @ HEAD

| # | Finding | Resolution |
|---|---|---|
| 1 | [P1] `isinstance` admits output-model subclasses; an extra-field subclass emits exit-0 stdout violating the closed output schema | **Fixed** — took your option 1: the success branch now requires **exact runtime identity** (`type(outcome) is descriptor.output_model`); any other return — unrelated model or subclass alike — takes the `internal` / exit 4 path with empty stdout. New per-descriptor harness row (`test_subclass_success_model_row`) builds an extra-field subclass of each descriptor's own output model via `create_model` and pins exit 4 + empty stdout + schema-valid `internal` envelope. |

Evidence at this head — your probe re-run: `ExtendedVersionResult(VersionResult)` with an `unexpected` field now yields `exit=4`, empty stdout, `internal` envelope (was `exit=0` with the extra field in the payload).

**Same loophole class, closed proactively on the other lawful outcome:** `RefusalReport.refusals` items are model-typed, so a `Refusal` subclass's extra fields could ride `model_dump` into the refusal envelope past its closed item schema. `refusal_envelope()` now projects `code`/`path`/`detail` field-by-field — the envelope's shape is owned by the builder, never by the runtime item type — pinned by `test_refusal_envelope_projects_items_exactly` (subclass in, exact closed item out, schema-valid).

Verification: 60 member tests passed; ruff/format + pyright (member) clean; CI running on the new head.

Issue #510, opened 2026-07-15
[gda-balancing] Monte-Carlo Experiments and core Metric datasets


qinhaihong-red:
## Refactor supersession — 2026-09-06

Reuse this issue for Monte-Carlo delivery under #865; do not create a second estimator issue. The historical 1.x seed/toolkit-version and fixed v1 phase comments are historical context, superseded where they conflict with the current language and execution contract. After #875, unrelated toolkit/build labels or whole-LDB changes cannot make an otherwise identical executable Experiment ineligible. Actual seed derivation, numeric/RNG/runtime policy, method inputs, censoring/correlation, and observation data remain bound.

Independently recomputable Metrics remain required. Authenticated Verifier issuance is deferred under #542–#544; it must not be made a prerequisite for this functional estimator slice.

This dated update changes only the identified contract or scheduling clauses. The original requirements, issue identifiers, historical evidence, and still-valid acceptance scope remain below. It does not claim implementation or conformance completion.

---

## Parent

- #501
- #534

## What to build

Implement seeded Monte-Carlo encounter estimation as an admitted Evaluation method over an exact
Resolved Model and Experiment. Publish round-trippable replicated observations and Metrics for
normalized TTK, TTD, survival, and the declared encounter outcomes. The evaluator is a public
operation; it may not consume an unbound config, invent host-only combat semantics, or collapse the
replicate dataset into an unverifiable score.

## Acceptance criteria

- [ ] The Experiment content-binds the Resolved Model, Runtime profile, replication unit, sample count, seed derivation, censoring, and correlation/common-random-number policy.
- [ ] Repeating the exact inputs produces the same observations, Metrics, dispositions, and artifact identities.
- [ ] TTK, TTD, survival, uncertainty, and sample disposition are computed from preserved replicate-level observations and round-trip through the public Metrics schema.
- [ ] Positive, boundary, censored, correlated, malformed-input, and mutation cases execute through the public evaluator boundary.
- [ ] Missing or invalid prerequisites atomically publish only a typed terminal-audit artifact set; no partial Metrics, Verdict, or Evidence success artifact is published.
- [ ] An authenticated Verifier can independently check the exact Experiment, dataset, aggregation policy, and reported Metrics.

## Blocked by

- #540
- #552
- #576

qinhaihong-red:
[...]

PR #527, opened 2026-07-19
chore(gda-balancing): Standard Schema core & validation end to end (#504)

### LINKED ISSUE #504: [gda-balancing] Schema core & validation end to end

qinhaihong-red:
Closes #504 — the complete Standard Schema core, end to end. Originally opened as the attributes slice of a 2-PR split; per review round 1, the `effects` section was folded in so the exposed `1.0.0` contract is whole, and every review finding is addressed on this branch (9 follow-up commits, mapped in the review reply).

## What this delivers

- **Public formula seam** (`gda_balancing.formula`, bADR-0003): both definition-time evaluators behind the one seam #510 reuses; strict left-to-right folds, IEEE 754-2019 signed-zero min/max, half-away-from-zero `round`, exact bounded integer `power` with per-step finiteness (incl. the reciprocal path), `EvaluationRefusal` (`non_finite_evaluation`); `evaluate_bases` = topological, bounds-clamped definition-time finals.
- **Complete 1.0 document model**: closed envelope, `meta`, `$schema` mirror, `parameters`, `attributes` (facets/tiers/bounds), **`effects`** (stacking types, modifiers, durations, periods); reserved sections permissive + excluded from emission, refused semantically with their dedicated code.
- **Three-phase boundary funnel** (bADR-0004): preflight (caps incl. numeric/Unicode ingress domain, duplicate-key refusal **with element pointers**, depth pre-scan, version dispatch), structural (jsonschema 2020-12 against the generated schema, **linear-time**), semantic (**30-rule frozen registry**; catalog + conformance walks are projections). RFC 6901 element-level refusals, report-all/dedup/order/1000+`truncated`.
- **Versioned dispatch**: a resolved immutable `VersionBundle` (schema/model/rules/`$id`) threaded through every phase — a declared `major.minor` validates under its own definition; proven with a synthetic 1.1 line.
- **CLI surface** (bADR-0007..0011): `design validate`, `design format` (canonical round-trip; optional members absent-or-typed, never `null`), `schema get structural|catalog`, `--out` artifact sink (atomic, receipt, aliasing guard), `version.supported_schema_line = "1.0"`; exit 2 refusal envelope end to end.
- **Self-description artifacts** (bADR-0005): generated structural schema (ECMA-portable patterns + engine-parity guards, linear `if/then` arity shape) and semantic rule catalog (scope = array of pointer templates), pinned by committed goldens.

## Vectors

V1–V6, V11, V12, V16 — the full #504-owned set — as executable tests at the CLI/JSON surfaces and the formula seam, plus the per-rule catalog walk end to end (30 rules; funnel-reachability is an executable invariant).

## Review-round-1 resolution highlights

- 1.0 completeness: effects landed (`cd3cc0f`) — the split's PR2 is dissolved into this PR.
- Exponential validation: linearized on both engines (`20a6804`); review's depth-12 probe 10.9s → ~1ms; depth-32 accepted under a latency regression, depth-33 refuses semantically (cap 64→96, bADR-0004 dated amendment).
- Ingress: big-int/digit-limit/lone-surrogate inputs now typed refusals (`b8c86f3`, new `string_not_unicode`); dup-key refusals carry real pointers.
- Facet validity: `bounds_empty|inverted|outside_domain`, `base_outside_domain`, unique `accepts` (`3b7b720`).
- Optional ≠ nullable (`a15adef` + bADR-0005 dated amendment `ed248b2`).
- Version bundle threading (`e6cc71a`); `_power` refusal-path fix (`7503465`); test hygiene + committed minimal fixture + scope arrays (`1e9ccdf`).

## Notes for review

- `number_not_finite` / `string_not_unicode` are preflight-family additions beyond bADR-0004's listed caps — crash-class guards (canonical emission must stay encodable), justified in `preflight.py`.
- Surrogate-bearing KEYS refuse at the deepest safely-encodable pointer (the enclosing element) — embedding the surrogate in the pointer would recreate the crash the guard prevents.
- Known corner (accepted): the `--out` aliasing guard gates on the positional naming an existing file; a descriptor-level document-input marking is the clean fix once a second document+sink command exists.
- No accepts-channel gating rule for modifier targets: bADR-0006 scopes #504 to target *declaredness*; channel gating is pipeline behavior (#510). Flagged as a possible follow-up.
- `jsonschema` is now a runtime dependency; root `uv.lock` relocked.

## Verification

`pytest libs/gda-balancing/tests`: **313 passed, 17 skipped**. pyright/ruff clean. e2e tier (`test_e2e_cli.py`) drives the key user path through the installed entry point. Real-CLI smoke on every review probe: all typed refusals with precise paths.


qinhaihong-red:
## `$review` — Standards and Spec

Reviewed head: `98c192eb49577623c86b5bb2206265bbda23b0a6`  
Merge base: `99a4b7fa9214606fca00140df2c8377e0be6af03`

### Standards

1. **Critical — hard violation:** `schema/model/document.py:9-11,64-77` publishes `SCHEMA_VERSION = 1.0.0` while deliberately omitting `effects`; valid v1 effect documents are therefore structural failures and the catalog lacks all effect rules. bADR-0001:22-30 and bADR-0006:18-42 make `effects` and its validation normative v1. Do not expose 1.0 until the full v1 core lands, or version this partial contract separately.

2. **Critical — hard violation:** recursive unions in `schema/model/formula.py:131-189` make structural validation effectively exponential. A read-only probe over valid unary trees measured depth 4 = 0.008s, depth 8 = 0.133s, depth 10 = 1.187s, and depth 12 = 10.773s. This defeats bADR-0003:145-152's bounded-cost claim and bADR-0004:60-63's resource-exhaustion contract. Replace recursive `anyOf`/`oneOf` branching with a discriminator/conditional shape and add latency tests near the accepted boundary.

3. **Hard violation:** `preflight.py:36-39,89-97` makes normative formula depths 31-32 unreachable: the enclosing document raises total JSON depth above 64, so both are refused as `nesting_too_deep`. bADR-0003 permits formula depth <= 32; the caps must compose orthogonally. Add acceptance tests at exactly 32.

4. **Hard violation:** `attributes.py:36-47` plus `rules.py:228-244` accepts `probability` attributes with `bounds: {}`, bounds outside `[0,1]`, and inverted `floor > cap`. bADR-0002:22-42 requires probability's `[0,1]` value space and mandatory narrowing design bounds. Add structural/semantic rules and negative vectors.

5. **Hard violation:** duplicate-key refusals at `preflight.py:99-115` always use root path `""`; bADR-0004:65-74 requires the JSON Pointer to the offending element, never merely the enclosing collection. Preserve parser location and report `/meta` or the nested key path.

6. **Judgment/design gap:** preflight validates a line but discards it (`preflight.py:209-250`); structural validation then always uses one global current validator (`structural.py:41-64`). This cannot satisfy bADR-0001:64-76 / bADR-0005:109-117 when 1.1 lands and 1.0 must retain its own envelope. Carry a selected version bundle through all funnel phases now.

**Standards finding count: 6 (5 hard, 1 judgment). Worst issue: the advertised 1.0 contract rejects the normative v1 `effects` section.**

### Spec

1. **Critical — this cannot safely merge as a standalone Schema release.** Issue #504 requires the entire "Schema core & validation end to end"; bADR-0001:22-30 and V1 (`0004a`:24-25) define `effects` as designed v1, while bADR-0004:118-124 assigns V1-V6 to #504. The PR nevertheless publishes `SCHEMA_VERSION = "1.0.0"` while omitting `effects`; even `effects: {}` is refused at `/effects`. Either land both PRs atomically or do not expose/version/freeze this artifact as 1.0 before effects, V5, and V6 exist.

2. **Required — integer negative `power` violates the evaluator contract.** bADR-0003:52-57,145-149 requires individually evaluated operations and typed non-finite refusal. `_power()` (`formula.py:264-279`) neither checks intermediate multiplications nor uses the refusal-aware division. Probe: `1e200^-2` returns `0.0` after hidden overflow; `1e-200^-2` leaks `ZeroDivisionError`, violating the one sanctioned downstream refusal class.

3. **Required — probability/bounds validity is incomplete.** bADR-0002:22-42 says probability is `[0,1]` and bounds narrow within it. The sole rule (`rules.py:228-244`) checks only that a `bounds` object exists. Probes show `bounds: {}`, `floor: -1, cap: 2`, and even reversed `floor: 100, cap: 0` all validate.

4. **Required — version "dispatch" has no versioned target.** bADR-0001:64-76 and bADR-0005:109-117 require validation against the declared minor's immutable envelope. Preflight only boolean-accepts a line (`preflight.py:209-251`); structural/model/semantic phases always use the single current schema/model (`funnel/__init__.py:71-89`). The first additive minor would therefore validate older documents under newer rules.

5. **Required — duplicate-key refusals lose element location.** bADR-0004:65-74 requires a JSON Pointer to the offending element. `preflight.py:99-115` always emits `path=""`; multiple duplicates at different locations also collapse under `(code, path)` deduplication.

No material scope creep found.

**Spec finding count: 5. Worst issue: publishing an incomplete but immutable-looking Standard Schema 1.0.**

Overall: **6 Standards findings, 5 Spec findings; request changes.** The strongest blocker is the standalone publication of an incomplete 1.0 contract, followed by exponential validation cost on inputs far below the declared depth cap.

Verification at this head: member suite `210 passed, 18 skipped`; Ruff and Pyright passed; `git diff --check` passed; current CI checks are green. Green checks do not exercise the probes above.

qinhaihong-red:
## `agent-skills:code-review-and-quality` — Multi-axis review

**Reviewed head:** `98c192eb49577623c86b5bb2206265bbda23b0a6`  
**Verdict:** **Request changes**

I treated the generated artifacts and the passing suite as evidence, then exercised the public funnel, CLI, and formula seam with adversarial boundary values. The central registry/descriptor approach is promising, but the current core still has crash-class ingress gaps, an algorithmic resource-bound failure, and several places where the published contract is not the runtime contract.

### Findings

1. **[P1][Performance / security boundary] Valid expression validation grows exponentially before the advertised depth cap.**  
   [`Node = Union[LiteralNode, AttrRef, ParamRef, OpNode]`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/model/formula.py#L175-L189) leaves every recursive node to be tried against several union branches. On the same machine, otherwise-valid unary trees took about `0.008s` at depth 4, `0.133s` at 8, `1.187s` at 10, and `10.773s` at 12. The semantic limit is 32, so a sub-1 MiB untrusted document can consume excessive CPU long before a limit is reported. Give every node kind a single discriminated parse path (or an equivalent linear validator), and add a near-limit latency regression test.

2. **[P1][Correctness / boundary hardening] User-controlled JSON can still escape the refusal path as `internal` / exit 4.**  
   [`json.loads` catches only `JSONDecodeError`, and `_walk_caps` inspects only `float`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/funnel/preflight.py#L99-L119). Reproductions:
   - a 400-digit integer passes structural validation, then triggers the funnel's “engine-parity bug” `RuntimeError`;
   - a 5,000-digit integer leaks Python's integer-conversion `ValueError`;
   - an escaped lone surrogate (`"\ud800"`) is accepted, then `design format` fails during UTF-8 output with `UnicodeEncodeError`.

   These are invalid inputs, not engine failures. Normalize/cap the complete numeric domain and reject invalid Unicode scalar values at the input boundary; catch all documented decoder failures and add CLI tests that require a typed refusal / exit 2 with a precise path.

3. **[P1][Correctness] Negative exact-integer powers bypass per-step finite and division semantics.**  
   [`_power`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/formula.py#L264-L285) multiplies without `_finite` and computes the reciprocal with raw `/`. Consequently `1e200^-2` silently returns `0.0` after an overflowed intermediate, while `1e-200^-2` leaks a plain `ZeroDivisionError`. Apply the same finite check after every multiplication and route the reciprocal through the evaluator's checked division/refusal path.

4. **[P1][Consistency / canonical form] “Optional `$schema`” has become nullable and is materialized as `null`.**  
   [`schema_ref: str | None = None`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/model/document.py#L62-L68), [`raw.get("$schema")`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/funnel/semantic/rules.py#L421-L432), and default-inclusive model emission together accept an explicit `"$schema": null` and emit it for a minimal formatted document. The bADR contract says the member is optional and, when present, must identify the matching schema; `null` does neither. Make presence distinct from nullability, omit absent optional-without-domain-default members, and keep only actual domain defaults (for example, `accepts: []`) materialized.

5. **[P1][Completeness / orthogonality] The attribute facets admit states their own domain vocabulary says are invalid.**  
   [`Bounds` permits either side to be absent and `accepts` is an unconstrained tuple](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/model/attributes.py#L36-L47). The public funnel currently accepts probability attributes with `base: 2`, `{bounds: {}}`, bounds outside `[0,1]`, reversed `floor > cap`, and duplicate channels such as `["effects", "effects"]`. Define the local set constraint (`uniqueItems`) structurally, then add semantic rules for ordered/complete bounds and probability/percentage domain compatibility, with one negative vector per invariant.

6. **[P2][Architecture / extensibility] Version dispatch validates a line but does not select a versioned implementation bundle.**  
   [`_version_dispatch` resolves `line` and then returns only refusals](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/funnel/preflight.py#L209-L251); [`validate` then always uses the global current schema, model, and rules](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/src/gda_balancing/schema/funnel/__init__.py#L66-L93). This works only while one line exists and cannot implement the stated “serve older minors with their own definition” policy. Return a resolved immutable version bundle (schema, model/parser, rules/catalog, canonicalization policy) and thread it through every phase.

7. **[P2][Verification] The new tests do not fully honor #504's external-contract-only acceptance criterion.**  
   [`test_dependency_graph.py`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/tests/test_dependency_graph.py#L1-L21) constructs internal models and calls internal graph helpers, while [`test_semantic_catalog.py`](https://github.com/aigengame/godot-agent/blob/98c192eb49577623c86b5bb2206265bbda23b0a6/libs/gda-balancing/tests/test_semantic_catalog.py#L21-L48) directly executes registry rule functions. Those unit tests can remain as supplemental diagnostics, but the acceptance evidence needs to live at the CLI/JSON and public formula seams, including a committed minimal golden fixture rather than repeated inline minima. The missing boundary cases above are exactly what the internal tests did not expose.

### Design-quality assessment

- **Consistency:** blocked by the partial `1.0.0` claim noted in the preceding `$review` comment and by optional-vs-null canonicalization drift.
- **Completeness:** blocked by missing input-domain, bounds, and evaluator failure semantics.
- **Orthogonality:** blocked because the global JSON depth cap makes valid formula depths 31–32 unreachable, while semantic formula depth is specified independently; version acceptance is likewise decoupled from artifact selection.
- **Extensibility:** the registries and generated artifacts are a good base, but boolean version acceptance and recursive union validation will not scale to additional schema lines or larger formula trees.

### Verification

- Member suite: `210 passed, 18 skipped`
- Ruff check + format check: pass
- Pyright: `0 errors`
- `git diff --check`: pass
- GitHub checks at the reviewed head: Ruff, Pyright, and Python unit/build all successful; Godot e2e skipped

Passing checks do not cover the adversarial inputs above, so I would not merge this schema core until the P1 findings and the incomplete-v1 contract identified in the first review are closed.

qinhaihong-red:
## Review round 1 — resolution map

All 6 Standards, 5 Spec, and 7 multi-axis findings are addressed on this branch (head `1e9ccdf`, 9 commits since the reviewed head `98c192e`). Suite: 313 passed / 17 skipped; pyright/ruff clean; CI green.

| Finding | Resolution |
|---|---|
| **Standards#1 / Spec#1** — 1.0 exposed without `effects` | **Fixed @ `cd3cc0f`** — full `effects` section (model + 11 semantic rules + V5/V6) folded into this PR; the 2-PR split is dissolved, this PR now `Closes #504`. Catalog: 30 rules. |
| **Standards#2 / multi#1** — exponential union validation | **Fixed @ `20a6804`** — callable-discriminator single-dispatch unions (pydantic) + `if/then` arity reshape of the published OpNode schema (failed branches never recurse). Review's depth-12 probe: 10.9s → ~1ms; latency regressions at depth 32 and 256 nodes; accept⇔construct equivalence battery. |
| **Standards#3** — formula depths 31–32 unreachable | **Fixed @ `20a6804`** — `MAX_NESTING_DEPTH` 64→96 with derivation + dated bADR-0004 amendment; depth-32 accepts (exit 0), depth-33 refuses `expression_tree_too_deep` end to end (un-shadowed). Deliberately gated on the linearization: under the exponential validator the old cap was a load-bearing shadow (Stage-G root cause). |
| **Standards#4 / Spec#3 / multi#5** — probability/bounds validity | **Fixed @ `3b7b720`** — new rules `bounds_empty`, `bounds_inverted`, `bounds_outside_domain`, `base_outside_domain` (+ boundary-inclusive [0,1] per bADR-0002) and structural `uniqueItems` on `accepts` (both engines). Every review probe now exits 2 with the expected code/path. |
| **Standards#5 / Spec#5** — duplicate-key refusals at `""` | **Fixed @ `b8c86f3`** — id-tracked location recovery in the parse hook + path-tracked walk: one refusal per duplicated key at its element (`/meta`, `/parameters/power`); distinct locations no longer collapse under dedup. |
| **Standards#6 / Spec#4 / multi#6** — no versioned dispatch target | **Fixed @ `e6cc71a`** — frozen `VersionBundle` (line, structural schema, model, rules, `$id`) resolved at preflight and threaded through every phase; `refusal_code_namespace` unions all bundles. Proven with a synthetic 1.1 line: a 1.0 document never sees 1.1 rules. |
| **Spec#2 / multi#3** — `_power` bypasses refusal semantics | **Fixed @ `7503465`** — per-step finiteness in the multiplication loop; reciprocal through the checked division. `1e200^-2` and `1e-200^-2` both raise the typed `non_finite_evaluation` refusal. |
| **multi#2** — ingress escapes to exit 4 (big ints, digit-limit, lone surrogates) | **Fixed @ `b8c86f3`** — ints with no finite double → `number_not_finite` at the value's path; `ValueError` (digit limit) caught → `malformed_json`; new preflight code `string_not_unicode` for unpaired surrogates in values AND keys. One recorded deviation: a surrogate-bearing KEY refuses at the deepest safely-encodable pointer (its enclosing element) — embedding the surrogate in the emitted pointer would recreate the very UnicodeEncodeError the guard prevents. |
| **multi#4** — optional `$schema` became nullable / null materialized | **Fixed @ `a15adef` + `ed248b2`** — published schema drops `X \| None` null arms and `default: null` (explicit `null` refuses structurally, incl. `/$schema`, `/meta/description`); emission omits absent optionals (`exclude_none`) while domain defaults (`accepts: []`, empty sections) stay materialized; `VersionResult.supported_schema_line` tightened to required `str`. Contract recorded as a dated bADR-0005 amendment (point-in-time prose preserved). |
| **multi#7** — internal tests as acceptance evidence; no committed minimal fixture | **Fixed @ `1e9ccdf`** — committed `tests/fixtures/minimal_design.json` (drives vectors, round-trip, and the subprocess e2e key path); internal diagnostics (`test_dependency_graph`, semantic-layer walk, bundle unit tests) labeled supplemental; catalog `scope` normalized to an array of pointer templates (the two dual-site reference rules carry one template per site). |

Deferred (recorded, not silently dropped): the `--out` aliasing guard's descriptor-level document-input marking (PR body "Notes"), and the accepts-channel gating non-rule (bADR-0006 scopes #504 to target declaredness; channel gating is #510 pipeline behavior).

qinhaihong-red:
## Recheck — remaining findings

**Rechecked head:** `1e9ccdff8dfb8b404f8386136f559468d725ae77`  
**Verdict:** **Request changes**

The resolution map is substantially supported: negative integer power, linear expression validation, depth-cap composition, ordinary duplicate-key locations, optional-vs-null emission, version-bundle threading, the minimal fixture, and the original bounds probes are fixed. The following contract gaps remain.

### 1. **Critical — effect magnitudes bypass the shared formula semantic rules**

The form-field and expression-tree rules still iterate only attribute base formulas through [`_iter_formula_bases`](https://github.com/aigengame/godot-agent/blob/1e9ccdff8dfb8b404f8386136f559468d725ae77/libs/gda-balancing/src/gda_balancing/schema/funnel/semantic/rules.py#L440-L575). Effect magnitudes receive reference-integrity checks, but not the named-form constraints or the depth/node caps required by bADR-0003 for every formula consumer.

Public-boundary reproductions:

- an instant effect whose magnitude is `{"form":"lookup_table", ..., "table":[]}` exits 0 as valid, although lookup tables require at least one entry;
- an effect magnitude containing a depth-33 expression tree is accepted as a `DesignDocument`, although the v1 limit is 32.

Please drive all formula rules from one iterator yielding `(formula, pointer_tokens)` for both attribute bases and effect magnitudes, then add CLI vectors and catalog scopes for every consumer/site. Until then the newly added `effects` section does not close the complete-1.0 finding.

### 2. **Critical — a surrogate key combined with a descendant violation still exits 4**

[`_walk_caps`](https://github.com/aigengame/godot-agent/blob/1e9ccdff8dfb8b404f8386136f559468d725ae77/libs/gda-balancing/src/gda_balancing/schema/funnel/preflight.py#L236-L280) emits a safe refusal for an unpaired-surrogate key, but then recurses with that unsafe key in `tokens`; duplicate-key handling also builds the key-bearing pointer before validating the key.

This input:

```json
{"schema_version":"1.0.0","meta":{"name":"x"},"parameters":{"\ud800":1e999}}
```

returns exit 4 / `internal_error`. The underlying failure is a Pydantic `ValidationError` while constructing `number_not_finite` at `/parameters/\ud800`. A duplicated surrogate key has the same class of failure. Once a key cannot be represented safely, descendant/duplicate refusals must retain the deepest safe pointer (or stop walking that subtree); no input-controlled composition should escape the typed refusal path.

### 3. **Required — `probability` does not remain inside its `[0,1]` value domain**

The funnel accepts:

```json
{
  "domain": "probability",
  "base": {"formula": {"literal": 2}},
  "bounds": {"floor": 0}
}
```

and the public `evaluate_bases` seam returns `2.0`. bADR-0002 says probability's value space is `[0,1]`, while design bounds merely narrow within that space. The evaluator currently [clamps only to explicitly present floor/cap members](https://github.com/aigengame/godot-agent/blob/1e9ccdff8dfb8b404f8386136f559468d725ae77/libs/gda-balancing/src/gda_balancing/formula.py#L168-L215).

Please make the domain range an independent invariant (for example, an intrinsic probability clamp composed with design bounds), or require bounds that fully enforce the domain. Cover formula-derived values and future effect contributions, not only direct bases.

### 4. **Required — the semantic catalog's scope contract is still incomplete and drifts from bADR-0005**

bADR-0005 describes `scope` as one JSON Pointer template, while the implementation now publishes an array without a normative amendment. Even under the new array shape, the entries do not enumerate every site the corresponding checks can report:

- `form_points_not_increasing` and `form_points_insufficient` can report both `.../points` and `.../table`, but catalog only `.../points`;
- `temporal_value_not_positive` can report `.../duration` and `.../period`, but catalog only `.../duration`;
- closing finding 1 will also require effect-magnitude sites for the shared formula rules.

Synchronize the bADR and add a behavioral anti-drift test requiring every emitted refusal path to match one of its catalog scope templates. The current catalog↔registry identity assertion only repeats the same incomplete metadata.

### Verification

- Package suite: `313 passed, 17 skipped`
- Ruff, format check, Pyright, and `git diff --check`: pass
- Depth 31/32 attribute formulas validate quickly; depth 33 reaches `expression_tree_too_deep`
- Both negative-integer-power probes now raise `EvaluationRefusal/non_finite_evaluation`
- GitHub Ruff, Pyright, and Python unit/build checks: success; Godot e2e: skipped

Green checks do not contain the adversarial combinations above, so the PR is not ready to merge at this head.

qinhaihong-red:
## Recheck (round 2) — resolution map

All 4 remaining findings addressed at head `4087563` (2 commits since the rechecked head `1e9ccdf`). Suite: **337 passed / 17 skipped**; pyright/ruff clean; CI green.

| Finding | Resolution |
|---|---|
| **1 (Critical)** — effect magnitudes bypass the shared formula rules | **Fixed @ `f37691f`** — one `_iter_formula_sites` authority yields `(formula, tokens)` over every attribute base AND every formula-valued magnitude; all four form-field rules and both tree caps drive from it (reference rules unified onto it too). Both repros refuse: empty-table magnitude → `form_points_insufficient` at `/effects/items/<id>/modifiers/0/magnitude/table`; depth-33 magnitude → `expression_tree_too_deep` at the magnitude root. +5 CLI magnitude vectors. |
| **2 (Critical)** — surrogate key + descendant violation exits 4 | **Fixed @ `4087563`** — an unpaired-surrogate KEY refuses at the deepest safely-encodable pointer and its subtree is NOT walked (reviewer-sanctioned option); duplicate-key recovery validates encodability before minting the pointer; defense-in-depth `_safe_pointer` on every preflight `Refusal` site. The exact repro (`{"\ud800": 1e999}`) and the duplicated-surrogate/deep-subtree variants all exit 2 with UTF-8-encodable envelopes. |
| **3 (Required)** — probability escapes `[0,1]` | **Fixed @ `4087563`** — the definition-time clamp is domain-composed: new public seam helper `clamp_to_attribute` (effective floor `max(0, floor?)`, cap `min(1, cap?)` for probability; declared-bounds-only for percentage/number), used by `evaluate_bases` and documented for #510's runtime-pipeline reuse. The repro now yields `1.0` at the seam; the document itself stays legal (clamping is evaluation-time), and the Stage-I static rules keep refusing statically-decidable violations. Seam + unit rows added. |
| **4 (Required)** — catalog scope incomplete + bADR drift | **Fixed @ `f37691f`** — dated bADR-0005 amendment records scope as one-or-more templates (array, one per site); every rule's array swept to enumerate all emitted sites (`points`+`table` × base+magnitude; `duration`+`period`; magnitude sites for all shared formula rules; `tier_reference_undefined` corrected to `…/tier`). NEW behavioral anti-drift test: every emitted refusal path must match one of its rule's templates under a precisely-stated matching contract (wildcard `{…}` tokens; prefix-matching only for formula-root scopes). |

qinhaihong-red:
## Recheck round 3 — remaining contract findings

**Rechecked head:** `4087563c919b0683ab968608addfd36ca420de9b`  
**Verdict:** **Request changes**

The runtime repairs are substantially effective: formula rules now reach effect magnitudes, the surrogate-key reproductions return an encodable exit-2 refusal rather than exit 4, probability formula results clamp into `[0,1]`, and catalog scopes enumerate the newly covered sites. No new runtime crash or performance regression was found. Four contract gaps remain.

### 1. **Required — surrogate-key handling silently weakens the report-all contract**

[`_walk_caps`](https://github.com/aigengame/godot-agent/blob/4087563c919b0683ab968608addfd36ca420de9b/libs/gda-balancing/src/gda_balancing/schema/funnel/preflight.py#L241-L277) now stops at a surrogate-bearing key and skips the key's `duplicate_object_key` refusal. This safely closes the exit-4 bug, but bADR-0004 still requires duplicate-key detection, all violations in an executed phase, and a pointer to the offending element rather than only its enclosing collection.

Either preserve the relevant typed refusals at a safely representable location, or explicitly amend bADR-0004 to define the unavoidable unaddressable-key exception. The implementation and normative contract must not disagree silently.

### 2. **Required — the probability effective-bounds law is not recorded in the normative design**

[`clamp_to_attribute`](https://github.com/aigengame/godot-agent/blob/4087563c919b0683ab968608addfd36ca420de9b/libs/gda-balancing/src/gda_balancing/formula.py#L211-L251) correctly composes the probability domain's intrinsic `[0,1]` space with declared narrowing bounds. However, bADR-0002 still specifies `clamp(..., bounds)` and says an absent bound side is unbounded. For a valid probability with only `{floor: 0}`, those statements conflict with the new intrinsic cap of `1`.

Record `effective bounds = intrinsic domain bounds ∩ declared bounds` normatively in bADR-0002, including the definition-time and #510 runtime-pipeline consequences.

### 3. **Required — two published catalog descriptions still exclude effect magnitudes**

The `expression_tree_too_deep` and `expression_tree_too_large` rules now correctly include magnitude scopes, but their [public descriptions](https://github.com/aigengame/godot-agent/blob/4087563c919b0683ab968608addfd36ca420de9b/libs/gda-balancing/src/gda_balancing/schema/funnel/semantic/rules.py#L1273-L1296) still say “base-formula expression tree.” The catalog is therefore self-contradictory. Rename these to describe any formula expression tree and regenerate the golden.

### 4. **Required — the scope anti-drift test proves soundness, not completeness**

[`test_every_emitted_refusal_path_matches_a_scope_template`](https://github.com/aigengame/godot-agent/blob/4087563c919b0683ab968608addfd36ca420de9b/libs/gda-balancing/tests/test_semantic_catalog.py#L135-L296) verifies that paths emitted by a manually maintained corpus match some declared template. It does not verify that every declared template is reachable; golden/metadata identity only snapshots the same declaration. For example, `temporal_value_not_positive` declares both `duration` and `period` scopes, while its behavioral fixture exercises only `duration`.

Make the behavioral evidence authoritative per `(rule, scope template)` and assert both directions: every emitted path is covered, and every declared template is exercised. That will also make a future third formula consumer fail the test if it is omitted from the corpus/catalog.

### Verification

- Full package suite: `337 passed, 17 skipped`
- Focused repair suites: `177 passed`
- Ruff check and format check: pass
- Pyright: `0 errors`
- `git diff --check`: pass
- GitHub Ruff, Pyright, and Python unit/build checks: successful; Godot e2e skipped

The fixes improve orthogonality and extensibility, especially through the shared formula-site iterator and composed clamp seam. Consistency and completeness remain blocked by the contract gaps above.

qinhaihong-red:
## Recheck round 3 — resolution map

All 4 findings addressed at head `0d1e983` (1 commit since `4087563`). Suite: **363 passed / 17 skipped**; pyright/ruff clean; CI green.

| Finding | Resolution |
|---|---|
| **1** — surrogate-key handling weakened report-all | **Fixed** — a duplicated unrepresentable key now DOES emit `duplicate_object_key`, anchored at the deepest safely-encodable ancestor with the `ascii()`-escaped key name in the detail (alongside the `string_not_unicode` for the same key). AND the exception is normative: dated bADR-0004 amendment defines the single sanctioned unaddressable-key exception to element precision (anchor at safe ancestor, escaped key in detail, subtree unwalked). Implementation and contract no longer disagree. |
| **2** — effective-bounds law unrecorded | **Fixed** — dated bADR-0002 amendment after the value-pipeline bullet: **effective bounds = intrinsic domain bounds ∩ declared bounds** (probability `[0,1]`; percentage/number no intrinsic sides), binding #504's `clamp_to_attribute` and the #510 runtime composition (both `P`'s clamp and the ledger equation). |
| **3** — two catalog descriptions excluded magnitudes | **Fixed** — `expression_tree_too_deep`/`expression_tree_too_large` descriptions reworded to "A formula expression tree …"; the other shared formula rules swept (no further base-centric wording); catalog golden regenerated (2 description lines, scopes unchanged). |
| **4** — anti-drift proved soundness, not completeness | **Fixed** — `SemanticRule.violation_fixtures` is now a tuple positionally aligned with `scope` (`__post_init__`-enforced), 43 fixtures across 30 rules (22 new across the 9 multi-site rules, incl. `temporal_value_not_positive`'s `period` fixture). The walk parametrizes over `(rule, template_index)`: each fixture refuses end to end with exactly its code AND emits a path matching **its aligned template**; the global soundness walk remains. Both directions executable: emitted ⊆ declared and declared ⊆ exercised — a future formula consumer omitted from corpus/catalog fails the suite. |

qinhaihong-red:
## Recheck round 4 — one required finding remains

**Rechecked head:** `0d1e983d262a731fae43384382d39d4e27ac898d`  
**Verdict:** **Request changes**

Three round-3 findings are closed:

- bADR-0002 now records the effective-bounds law and binds both #504 and #510;
- the two tree-limit catalog descriptions now cover every formula consumer;
- every currently declared v1 scope template has aligned semantic-layer and real-CLI behavioral evidence.

The unaddressable-key contract remains partial.

### **Required — multiple unrepresentable sibling keys collapse into one generic refusal**

The new [bADR-0004 amendment](https://github.com/aigengame/godot-agent/blob/0d1e983d262a731fae43384382d39d4e27ac898d/libs/gda-balancing/docs/badr/0004-boundary-funnel-validation-semantics.md#L67-L86) says an unaddressable-key refusal anchors at the deepest safely encodable ancestor, carries the escaped key name in its detail, and otherwise preserves report-all.

[`_walk_caps`](https://github.com/aigengame/godot-agent/blob/0d1e983d262a731fae43384382d39d4e27ac898d/libs/gda-balancing/src/gda_balancing/schema/funnel/preflight.py#L251-L289) includes the escaped key only for `duplicate_object_key`; each plain `string_not_unicode` still has the same generic code, path, and detail. [`assemble`](https://github.com/aigengame/godot-agent/blob/0d1e983d262a731fae43384382d39d4e27ac898d/libs/gda-balancing/src/gda_balancing/schema/funnel/report.py#L17-L27) then deduplicates on `(code, path)`, so distinct unrepresentable keys at the same safe ancestor disappear.

Public-boundary probes:

```json
{"schema_version":"1.0.0","meta":{"name":"x"},"parameters":{"\ud800":1,"\ud801":2}}
```

returns only:

```text
string_not_unicode @ /parameters
detail: object key is not valid UTF-8 text (contains an unpaired surrogate)
```

Neither escaped key is named. When both keys are also duplicated, the result contains only the `duplicate_object_key` for `\ud800` plus one generic Unicode refusal; every `\ud801` violation is lost.

Define a stable aggregation or deduplication exception for unaddressable keys—for example, aggregate all escaped offending key names per `(safe ancestor, refusal code)`—and record exactly that behavior in bADR-0004. The current implementation rejects safely and no longer exits 4, but it does not yet satisfy the new diagnostic-completeness contract.

### **Consider — the future formula-consumer anti-drift claim is still stronger than the mechanism**

[`rules.py`](https://github.com/aigengame/godot-agent/blob/0d1e983d262a731fae43384382d39d4e27ac898d/libs/gda-balancing/src/gda_balancing/schema/funnel/semantic/rules.py#L27-L34) says a future third formula consumer omitted from the corpus will fail the test. However, both [`_RULE_TEMPLATE_CASES` and the soundness corpus](https://github.com/aigengame/godot-agent/blob/0d1e983d262a731fae43384382d39d4e27ac898d/libs/gda-balancing/tests/test_semantic_catalog.py#L42-L50) are derived from the manually declared `scope` and `violation_fixtures`. If a new `_iter_formula_sites` branch is omitted from both declarations, no corpus document exercises it and the suite still passes.

Current v1 coverage is complete, so this is not a merge blocker. Either narrow the future-proofing claim or make formula sites one descriptor authority that projects the walker tokens, scope templates, and fixture factories.

### Verification

- Full package suite: `363 passed, 17 skipped`
- Focused recheck suites: `203 passed`
- Ruff check and format check: pass
- Pyright: `0 errors`
- `git diff --check`: pass
- GitHub Ruff, Pyright, and Python unit/build checks: successful; Godot e2e skipped

qinhaihong-red:
## Recheck round 4 — resolution

Both findings addressed at head `c09ba2c`. Suite: **366 passed / 17 skipped**; pyright/ruff clean; CI green.

| Finding | Resolution |
|---|---|
| **Required** — unrepresentable sibling keys collapse under `(code, path)` dedup | **Fixed** — key-caused refusals now aggregate per `(deepest safely-encodable ancestor, code)`: one `string_not_unicode` (and one `duplicate_object_key` where applicable) per ancestor, the detail listing ALL offending keys `ascii()`-escaped, deterministically sorted — dedup can no longer drop a distinct key. The probe now yields `object keys are not valid Unicode scalar strings: '\ud800', '\ud801'` in one refusal; the both-duplicated variant names both keys in both codes; different ancestors stay separate refusals. The bADR-0004 amendment (added on this branch, refined in place) records exactly this aggregation. |
| **Consider** — future-consumer anti-drift claim stronger than mechanism | **Narrowed** (reviewer-sanctioned option) — the claim now states the suite enforces mutual consistency of the three manual declarations (`_iter_formula_sites`, scope arrays, fixture factories) over the declared scope set; a wholly undeclared site is not auto-discovered. A single formula-site descriptor authority projecting walker/templates/fixtures remains a possible follow-up if a third consumer lands. |

Issue #528, opened 2026-07-19
[gda-balancing] Enable releasing PR titles: root exclude-paths + first-release gate (ADR-0037)


qinhaihong-red:
## What to build

Retire the release-title debt: today every gda-balancing PR must carry a non-releasing (`chore`) title because the root release-please package `"."` absorbs commits from **all** repo paths (verified during #502/PR #526: a `feat(gda-balancing)` on main would bump the `gda` package and open a spurious Release PR). The cost is recorded mislabeling — feature work lands in main history as `chore` (e.g. PR #527), and gda-balancing's future first-release changelog accrues nothing.

Three sequenced steps, one slice:

1. **Stop the absorption** — configure the root release-please package with `exclude-paths` covering `libs/gda-balancing/**` (verify the exact mechanism against current release-please documentation before implementing; do not trust memory). Prove it live on main.
2. **Gate the first release** (ADR-0037: gda-balancing publishing is deferred) — once releasing titles are enabled, release-please will start opening gda-balancing Release PRs; design and land the guard that makes an accidental tag/publish impossible before the publish tail exists (unmerged-Release-PR policy, CI guard, or equivalent — the chosen form is adjudicated in PR review).
3. **Flip the convention** — after #504's PR pair is merged, gda-balancing PRs use truthful Conventional-Commit types (`feat`/`fix`/…); update every place the old rule is recorded, and record the first-release changelog approach (the pre-flip `chore`-titled foundation work gets hand-authored initial release notes).

## Acceptance criteria

- [ ] A releasing-type commit touching only `libs/gda-balancing/**` on main produces **no** `gda` version bump and no `gda` Release PR (verified live, evidence linked on this issue — mechanism verified, not inferred).
- [ ] The gda-balancing release gate is landed and exercised: with releasing titles enabled, no tag or publish can happen before the ADR-0037 publish tail lands.
- [ ] The PR-title convention flip is recorded in its authoritative home(s), replacing the non-releasing-titles rule.
- [ ] The first-release changelog approach is recorded.
- [ ] ADR-0037 carries a dated Outcome note if its recorded behavior changes.

## Blocked by

None — the config work can start immediately; the convention flip itself waits until both #504 PRs are merged (no mid-stream type switch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

qinhaihong-red:
[...]

PR #529, opened 2026-07-20
chore: gda-balancing gets its own project + release boundary (#528)

qinhaihong-red:
Part of #528. Gives gda-balancing a release boundary that actually holds, so member PRs can carry truthful `feat`/`fix` titles. The title flip itself and its live proof stay on #528.

## Why the shape changed mid-PR

This PR started as "add `exclude-paths` to the root package". Review proved that insufficient, and the second attempt (also excluding `uv.lock`) provably ineffective: release-please matches `exclude-paths` entries as **directory prefixes** (`file.indexOf(path + "/") === 0`) and drops a commit from a package only when **every** changed file is excluded. A root-level file cannot be excluded at all — so a member dependency change (member `pyproject` **+** the shared workspace `uv.lock`; #527 is exactly that shape) kept counting for the root `gda` package.

Three alternatives were rejected on their merits:
- **Force such PRs to `chore`** — a member runtime-dependency change may itself warrant a release, and if the follow-up `feat` slips, dependency metadata sits in main with no member release.
- **Split pyproject and lock into two PRs** — not implementable: the intermediate state fails frozen-lock CI.
- **A "root-owned paths" allowlist in CI** — a second authority that drifts the moment a directory or workflow is added; the same class of finding this review already raised about the tag map.

**So the boundary moves instead: gda-balancing leaves the uv workspace and becomes an independent uv project with its own lock inside its own directory.** Every file a member change touches now lives under `libs/gda-balancing/`, which the single existing `exclude-paths` entry covers completely. Recorded as **ADR-0038**; ADR-0037's single-authority release model is unchanged — only its workspace premise is superseded, marked in place.

## What is in here

- **Project split** — root drops `[tool.uv.workspace]`; the member gains its own dev group (`pytest`, `pyright`), its own `uv.lock`, and a pinned `.python-version` (see below).
- **CI environments split** — `setup-python-env` takes a `sync` input (`root` | `member` | `both` | `none`); each job pays only for what it needs, cache keys cover both locks.
- **Release pipeline** — the member's build/test/tag-validate run against its own project; the Release-PR lock sync now refreshes **both** locks on every Release PR branch and commits whichever changed (no branch-name parsing, so a future package needs no edit).
- **Tag gate** — derives all four inputs release-please uses (`component`, `include-component-in-tag`, `include-v-in-tag`, `tag-separator`), per package with the top-level value as default. Verified it reproduces today's tags and follows a hypothetical `include-v-in-tag: false` / `tag-separator: "/"` config.
- **Scope guard** (new CI job) — a PR touching the member directory *and* anything outside it fails when its title is releasing-typed, telling the author to split. The member directory is **derived from `exclude-paths`**, not restated. Belt-and-braces today; load-bearing once #528 flips the titles.
- **Co-install smoke** (new CI job) — builds both wheels, installs them together into one clean venv, runs a command from each. This deliberately replaces the feedback the shared lock used to give for free: two independent locks each resolve alone and could silently drift into wheels that cannot sit side by side.

## Notes for review

- **`.python-version` was a latent trap the split exposed**: without it the member project resolved to CPython 3.14 locally while the root stayed on 3.13 — CI is unaffected (`setup-uv` pins the interpreter job-wide) but a local run would silently differ from the gate. Pinned, lock byte-unchanged, suite and pyright re-verified on 3.13.
- **`uv build --package` is workspace-only**; the member now builds via `--project`. A relative `--out-dir` resolves against **cwd, not the project dir**, so the release artifact still lands where `upload-artifact` expects it.
- **`--frozen` kept, not switched to `--locked`** — repo-wide `--locked` is worthwhile but orthogonal hardening; happy to fold it in if you want it here.
- Guard trigger is "the PR touches the member directory", not a parsed title scope — it needs no scope-naming convention and matches the actual failure mode.

## Verification

Root suite `1089 passed`; member suite `366 passed / 17 skipped` in its own environment (including the subprocess entry-point tier — `uv run --project` puts the console script on PATH); both pyright runs clean; ruff clean over 282 files; both builds produce sdist+wheel; dual-install smoke runs both commands from one venv; neither lock references the other package. The guard's real heredoc was driven through 7 cases including the `libs/gda-balancing-extra/` prefix trap. CI green, including both new jobs.


qinhaihong-red:
Reviewed `c4376957f7a00d2c711238926824b737314f1cf9` against `main`.

## Standards

1. **Hard violation — `.github/workflows/release.yml:447-459`: tag identity now has two authorities.** `PREFIXES` duplicates the package/component/tag rules already owned by `release-please-config.json`. Adding or renaming a manifest package without updating this hidden map raises `KeyError` and blocks release-PR maintenance for every train. This violates the repository rule that duplicated facts must be derived from their authority or mechanically guarded. Please derive the required tags from the release-please config, or add a parity check that makes the duplication safe.

2. **Hard violation — affected documentation was not fully synchronized.** `.github/workflows/ci.yml:110-112` still says “gda-balancing publishing is deferred,” although this PR wires publishing. ADR-0037's new Outcome says the two-Release-PR slip mode is retired, while its current Consequences (`:88-91`) still says a member release commit proposes two Release PRs. Please update or explicitly mark those statements as historical/superseded so the current contract is unambiguous.

3. **Judgement call — `release-please-config.json:22-24` handles only changes whose files all stay under the member directory.** The shared-lock case below remains absorbed by the root package. This weakens ADR-0037's separate-train design and the repository's adjacent-layer/extensibility rules; the design needs evidence or a policy for member changes that legitimately update root-owned workspace artifacts.

No further Standards breach was found in the member tail's recovery chain or least-privilege split: build is read-only, only the PyPI job has OIDC, GitHub publish has `contents: write` only, and failed upstreams become explicit failed tail jobs.

## Spec

1. **Blocking — root absorption remains for normal member dependency changes.** Issue #528 aims to retire the title debt so gda-balancing PRs can use truthful `feat`/`fix` types; that goal is broader than AC1's synthetic member-directory-only case. release-please excludes a commit only when **every** relevant file is under an excluded path ([current implementation](https://github.com/googleapis/release-please/blob/5b3157824739ed7fe5d2e0fd5d1c1c8f0c4fa204/src/util/commit-exclude.ts#L47-L57)). A normal member dependency change touches both `libs/gda-balancing/pyproject.toml` and the shared root `uv.lock`; PR #527 is a concrete recent example. Because this PR excludes only `libs/gda-balancing`, `uv.lock` keeps such a squash commit in the root release history, so a truthful member `feat`/`fix` can still bump `gda` and open the spurious root Release PR. The exclusion design and its verification need to cover realistic member-only workspace changes involving shared root artifacts without hiding genuine root changes.

No additional scope-creep or member-tail defect was found. The post-merge live proof, releasing-title exercise, authoritative convention flip, and first-release changelog approach remain intentionally deferred on #528 rather than being separate blockers in this partial PR.

**Summary:** Standards: 3 findings. Spec: 1 finding. The worst issue is the remaining root absorption for member changes that update the shared `uv.lock`.

Validation at this head: release-please config/output semantics checked against current upstream docs/source; workflow JSON/YAML and 15 shell steps parsed; `uv build --package gda-balancing` succeeded; member suite passed (`366 passed, 17 skipped`); current PR CI is green.

qinhaihong-red:
## Review resolution

All 4 findings addressed at head `1dc7e16`. CI green.

| Finding | Resolution |
|---|---|
| **Spec#1 (Blocking)** — root absorption remains for member changes touching the shared lock | **Fixed** — `uv.lock` joins the root's `exclude-paths`. Confirmed your analysis two ways before acting: release-please excludes a commit only when *every* changed file is excluded, and **#527's own squash commit is exactly that shape** (`libs/gda-balancing/pyproject.toml` + root `uv.lock`), so the member-path-only exclusion would indeed have left a truthful `feat(gda-balancing)` bumping gda. The lock exclusion is also principled, not just expedient: **`uv.lock` is not in any published distribution** — I built the sdist and listed it; it ships `pyproject.toml` dependency ranges only — so a lock change alters nothing users receive and warrants no bump. Genuine root dependency changes still bump gda, because they edit the un-excluded root `pyproject.toml`. Recorded in ADR-0037's Outcome note with both arguments. |
| **Standards#1** — tag identity had two authorities | **Fixed** — the gate no longer restates a prefix map. It **derives** each package's tag from `release-please-config.json` (`component` + `include-component-in-tag`, with the top-level value as default) — the same inputs release-please itself uses — so adding or renaming a package needs no workflow edit and cannot drift or `KeyError`. A package declaring `include-component-in-tag` without a component fails loudly with a named error rather than guessing. Verified the derivation reproduces the previous values exactly, on today's manifest and on a post-member-release manifest. |
| **Standards#2** — stale documentation | **Fixed** — `ci.yml`'s "publishing is deferred" comment now describes the actual split (CI build-smokes the member; the Release workflow owns its release build). ADR-0037's two-Release-PR consequence bullet carries an inline *superseded* marker pointing at the Outcome note, keeping the point-in-time prose intact per the repo's ADR-preservation convention. |
| **Standards#3** — member changes touching root-owned artifacts | **Addressed by the same fix**, with the policy stated explicitly in the ADR: the shared lock is excluded (it is a dev-environment artifact, not a shipped one), while a member change touching genuinely root-owned code — workflows, root `pyproject.toml`, `src/gda/**` — is *not* member-only and correctly keeps bumping the root. |

The post-merge live proof and the title flip remain on #528 as before.

qinhaihong-red:
Rechecked the fixes and the author's resolution at exact head `1dc7e169dd1fa0a49e4b60f3b71bbc1a4c6414c0`.

## Standards

1. **Partially resolved — tag identity still has a second implementation.** The hard-coded package map is gone, and the gate now derives `component` and `include-component-in-tag` from `release-please-config.json`. However, `.github/workflows/release.yml:447-474` still hard-codes `-v` and does not derive release-please's `include-v-in-tag` or `tag-separator`. release-please inherits all four inputs when constructing a tag ([manifest config](https://github.com/googleapis/release-please/blob/5b3157824739ed7fe5d2e0fd5d1c1c8f0c4fa204/src/manifest.ts#L1771-L1780), [tag construction](https://github.com/googleapis/release-please/blob/5b3157824739ed7fe5d2e0fd5d1c1c8f0c4fa204/src/util/tag-name.ts#L27-L36)). The current configuration happens to produce the expected tags, but a supported config change such as `"include-v-in-tag": false` or `"tag-separator": "/"` would make the workflow look for a different tag from the one release-please creates. The single-authority/drift finding therefore remains partially open.

2. **Resolved — the stale CI comment and historical ADR consequence are fixed.** `.github/workflows/ci.yml:109-115` now describes the CI/release split accurately, and ADR-0037 explicitly marks the old two-PR consequence as superseded.

3. **Unresolved — the new `uv.lock` exclusion and the ADR claims built on it are ineffective.** release-please's matcher treats every `exclude-paths` entry as a directory prefix: `file.indexOf(`${path}/`) === 0` ([implementation](https://github.com/googleapis/release-please/blob/5b3157824739ed7fe5d2e0fd5d1c1c8f0c4fa204/src/util/commit-exclude.ts#L47-L62)). Consequently, the literal `"uv.lock"` entry matches paths under `uv.lock/`; it does **not** match the root file `uv.lock`. The exact release-please-action v4 bundle used by this workflow contains the same matcher. ADR-0037's new statements that the shared lock is excluded and that this retires the absorption mode are therefore currently false.

## Spec

1. **Blocking — root absorption remains for realistic member dependency changes.** I exercised the #527-shaped commit directly:

   ```text
   changed files:
     libs/gda-balancing/pyproject.toml
     uv.lock

   root exclude-paths:
     libs/gda-balancing
     uv.lock

   release-please result: commit_excluded = false
   ```

   The member file is excluded by the directory prefix, but the root `uv.lock` matches neither entry, so the commit remains in the root package's release history. A truthful member `feat`/`fix` can therefore still bump `gda` and regenerate the unwanted root Release PR. The original Spec blocker is not closed, and the post-merge live proof/title flip on #528 cannot safely proceed on this implementation.

Confirming that `uv.lock` is absent from the built root sdist establishes that it is not a shipped artifact, but it does not change release-please's path-matching semantics.

**Summary:** Standards: 1 resolved, 1 partially resolved, 1 unresolved. Spec: 1 unresolved blocker. The worst issue remains root absorption for member changes that update the shared `uv.lock`.

Validation at this head: current PR CI is green; workflow YAML and all 15 shell steps parsed; `git diff --check` passed; the member suite passed (`366 passed, 17 skipped`); both root and member distributions built successfully.

qinhaihong-red:
## Recheck resolution — head `fb97a7b`

Both open findings are closed, one of them by abandoning the approach rather than patching it.

| Finding | Resolution |
|---|---|
| **Spec#1 (Blocking)** + **Standards#3** — the `uv.lock` exclusion is ineffective; root absorption remains | **Closed by removing the cause.** You were right on the mechanism and I verified it independently against the matcher: entries are directory prefixes, so `"uv.lock"` matched nothing and the ADR claims I built on it were false. Both were corrected in `3c7f13f` before anything else. The real fix is structural: **gda-balancing leaves the uv workspace and becomes an independent uv project with its own lock under its own directory** (`fb97a7b`), so a member dependency change touches only `libs/gda-balancing/**` and the single `exclude-paths` entry covers it. Recorded as ADR-0038 with the mechanism fact and the three rejected alternatives (forced-`chore`, two-PR split — not implementable under frozen-lock CI, and a root-path allowlist — a second authority prone to drift). |
| **Standards#1 (partial)** — tag identity still hard-coded `-v` | **Closed.** The gate now derives all four inputs release-please composes a tag from — `component`, `include-component-in-tag`, `include-v-in-tag`, `tag-separator` — each per-package with the top-level value as default. Verified it reproduces today's tags and tracks a hypothetical `include-v-in-tag: false` + `tag-separator: "/"` config. |
| **Standards#2** | Was already resolved; unchanged. |

Two things the migration surfaced that were not in anyone's plan, both fixed and recorded: `uv build --package` is workspace-only (member builds via `--project`, and a relative `--out-dir` resolves against cwd — the artifact path survives), and `.python-version` is not inherited across the split (the member resolved to 3.14 locally against a 3.13 gate until pinned).

Also added, to compensate what the shared lock gave for free: a **co-install smoke** job (both wheels into one clean venv, a command from each), and a **scope guard** job that fails a releasing-typed PR spanning the member and anything outside it — with the member directory **derived from `exclude-paths`**, so it does not become the drifting second authority your earlier finding warned about.

#528 still owns the title flip and the post-merge live proof.

qinhaihong-red:
Re-reviewed Solution 1 at exact head `fb97a7bd499ada1562cbfce9bb374bd2b2292e03` against `main`.

The structural change is sound: `gda-balancing` now owns its project and lock under `libs/gda-balancing/**`, so the original shared-root-`uv.lock` absorption blocker is closed. The remaining issues are in the release-boundary guards.

## Standards

1. **Hard violation — `.github/workflows/ci.yml:220-228` does not derive the releasing types from the release configuration.** The guard hard-codes only `feat`, `fix`, and a breaking `!`, while `release-please-config.json:7-11` also makes `deps` and `revert` visible release sections. release-please's default strategy patch-bumps non-feature significant commits ([upstream implementation](https://github.com/googleapis/release-please/blob/main/src/versioning-strategies/default.ts#L30-L92)). Therefore a mixed-path PR titled `deps(gda-balancing): ...` or `revert(gda-balancing): ...` passes this guard even though the resulting squash commit can produce releases. The comment at `:223-224` claims the set matches this repo's release configuration, but it does not. This violates the repository's single-authority and completeness rules. Please derive the releasing types from `release-please-config.json`, or mechanically cover every visible release type plus the breaking marker.

2. **Judgement call — `.github/workflows/release.yml:459-495` still implements only part of release-please's effective component resolution.** The gate derives the boolean/tag-format settings, but obtains the component only with `package.get("component")`. release-please can obtain the effective component from package/strategy metadata when an explicit `component` is absent. Today's configuration is safe because the only component-bearing member declares it explicitly, but the workflow comment claims equivalence for future packages more broadly than the implementation provides. Reuse release-please-equivalent component resolution, or narrow and mechanically guard the supported configuration contract.

## Spec

1. **Required — the mixed-path guard still has a truthful-title bypass.** ADR-0038 and #528 rely on this job becoming load-bearing when member PRs switch to real Conventional Commit titles. A PR spanning `libs/gda-balancing/**` and root-owned paths can use the truthful `deps(gda-balancing): ...` or `revert(gda-balancing): ...` title, pass CI, and then enter both release histories. That contradicts the intended “member-only changes release only the member; mixed ownership fails loudly” boundary. Add committed regression cases covering `feat`, `fix`, `deps`, `revert`, breaking forms, and member-only/root-only/mixed path sets.

**Summary:** Standards: 2 findings. Spec: 1 required finding. Verdict: **request changes**.

Validation at this head: root tests `1089 passed`; member tests `366 passed, 17 skipped`; both pyright runs clean; Ruff clean; both frozen lock checks and builds pass; clean co-install smoke passes; current GitHub CI is green.

qinhaihong-red:
## Code review and quality

Reviewed exact head `fb97a7bd499ada1562cbfce9bb374bd2b2292e03`. Verdict: **request changes**.

1. **Critical correctness — title edits bypass the load-bearing scope guard.** `.github/workflows/ci.yml:3-5` uses a bare `pull_request:` trigger. GitHub's default activity types are `opened`, `synchronize`, and `reopened`; `edited` is not included ([official event documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)). A mixed-path PR can therefore pass as `chore: ...`, then change only its title to `feat(gda-balancing): ...`; the head SHA and existing green check remain unchanged, and the guard is not rerun. Prefer a small dedicated workflow triggered by:

   ```yaml
   on:
     pull_request:
       types: [opened, synchronize, reopened, edited]
   ```

   Before #528 flips the title convention, make that check required. The repository's active default-branch ruleset currently has no required-status-check rule, so even a correctly triggered failure is advisory rather than load-bearing.

2. **Required completeness/testability — the release classifier omits configured release types and has no committed regression matrix.** `.github/workflows/ci.yml:220-228` recognizes only `feat`, `fix`, and `!`, but `release-please-config.json` exposes `deps` and `revert` as release-producing sections. The current green job exercised this PR's `chore` no-op path; it does not prove the rejection path. Move the classifier into a testable script or otherwise add table-driven cases for `feat`/`fix`/`deps`/`revert`, arbitrary breaking types, member-only/root-only/mixed files, and the title-edit trigger.

3. **Required maintainability — the member release validator reintroduces a second tag implementation.** `.github/workflows/release.yml:310-325` hard-codes `prefix = "gda-balancing-v"`, while the maintenance gate at `:459-495` now derives `component`, `include-component-in-tag`, `include-v-in-tag`, and `tag-separator` from `release-please-config.json`. A supported tag-config change can make release-please create a valid tag that this build job rejects. Use one shared, tested tag-derivation implementation for both paths.

No additional security, performance, or readability blockers found. The package/environment separation, co-install smoke, and release permission split are otherwise sound.

qinhaihong-red:
## Resolution — head `b9db339`

All four findings fixed; both reviews' items map onto the same four. Root suite **1175 passed** (baseline 1089 + 86 new), member suite unaffected at 366/17, CI green.

| Finding | Resolution |
|---|---|
| **B1 (Critical)** — a title edit bypasses the guard (`pull_request` default types exclude `edited`) | **Fixed** — the guard moved into its own workflow `.github/workflows/release-scope-guard.yml` with `types: [opened, synchronize, reopened, edited]`, so a title change re-evaluates it at the same head SHA. `ci.yml` deliberately keeps the default types, with a comment on why a title edit must not re-run the whole suite. |
| **A1 / B2 / Spec#1** — releasing types hard-coded and incomplete | **Fixed** — `scripts/release_scope_guard.py` derives them from `changelog-sections` (every non-hidden section). **Derived set: `feat, fix, deps, revert`** — confirming your finding that `deps` and `revert` were missing. Verified end to end: `deps(gda-balancing): …` spanning member + root now exits 1 (previously passed), `chore` mixed stays a no-op, and a breaking `!` on *any* type is treated as releasing. |
| **B3** — a second tag implementation in the member build | **Fixed** — `scripts/release_tags.py` is the one derivation, now serving all three call sites (root build validation, member build validation, release-PR tag gate). No tag prefix is restated anywhere in YAML. |
| **A2 (judgement)** — component resolution narrower than claimed | **Narrowed and guarded**, not reimplemented: the module documents the explicit-`component` contract and raises a named `TagDerivationError` on any config that would need inference, plus a drift alarm asserting the real config derives cleanly. Reimplementing release-please's full resolution would recreate the duplication the other findings are about. |

**Committed regression matrix** (your Spec#1 ask): `tests/test_release_tags.py` (13) and `tests/test_release_scope_guard.py` (73) — titles × path sets over `feat`/`fix`/`deps`/`revert`/`chore`/`docs`/arbitrary-`type!`, scoped and unscoped, malformed titles, member-only / root-only / mixed / empty, the `libs/gda-balancing-extra/` prefix trap, per-package overrides beating top-level defaults, placeholder skipping, and the missing-component error.

**One thing manual driving caught that the tests could not:** both scripts crashed under macOS's system Python 3.9 (`list[str] | None` evaluated at import). The release-PR tag gate and the guard workflow both invoke bare `python3` with no venv, so "stdlib only, runs anywhere" required `from __future__ import annotations`; re-verified on 3.9.6.

**Still needs a maintainer action, not code:** the guard blocks nothing until **"Member releasing-PR scope guard" is a required status check** on `main` — I verified the active ruleset currently carries only `deletion`, `non_fast_forward`, `pull_request`, so every check is advisory today. Noted in the workflow comment and ADR-0038, and it is a precondition for #528's title flip rather than for this PR.

qinhaihong-red:
Rechecked the author's resolution at exact head `b9db3390d894156615dfd132a7c08714da94d3e1`.

## Resolved

- The title-edit bypass is closed: `.github/workflows/release-scope-guard.yml` is a dedicated workflow triggered by `opened`, `synchronize`, `reopened`, and `edited`.
- The current configured releasing types are covered: `deps` and `revert`, arbitrary breaking `!` types, member/root/mixed paths, and the sibling-prefix trap all have committed regression coverage.
- Tag construction now has one implementation in `scripts/release_tags.py`, shared by both build validators and the release-PR tag gate.
- The component-resolution concern is closed by an explicit, fail-loud supported contract plus a real-config drift test.

## Still open — Standards / request changes

**`scripts/release_scope_guard.py:48-55` reads only the top-level `changelog-sections`, so package-level overrides can still make the load-bearing guard silently pass a two-release PR.**

release-please's [configuration schema](https://github.com/googleapis/release-please/blob/main/schemas/config.json) permits `changelog-sections` under each `packages.<path>` entry, and its [merge logic](https://github.com/googleapis/release-please/blob/main/src/manifest.ts#L1747-L1757) gives the package value precedence over the top-level default.

Concrete valid repro:

```text
top-level visible types: feat
root package visible types: security
gda-balancing visible types: security

title: security(gda-balancing): harden dependency handling
files:
  libs/gda-balancing/pyproject.toml
  README.md

current guard result:
  releasing = false
  ok = true
```

Both effective package configurations make `security` release-producing, so the mixed commit can produce both Release PRs, but `releasing_types()` sees only the top-level `feat` entry. This leaves the earlier single-authority/completeness finding partially resolved and contradicts ADR-0038's claim that the releasing types are derived from the release-please configuration.

Please either:

1. derive each package's effective `changelog-sections` and detect types that release both the root and the touched member; or
2. explicitly reject package-level `changelog-sections` and add a shipped-config drift test, matching the fail-loud approach used for component inference.

## External precondition

The code-level title-edit fix is correct, but the active `main` ruleset still does not require `Member releasing-PR scope guard`. That does not block merging this partial PR while member titles remain non-releasing, but it **must** block #528's convention flip and live proof.

**Summary:** the previous Spec finding is resolved; tag derivation and title-edit handling are resolved; one Standards finding remains partially open. Verdict: **request changes**.

Validation: 86 targeted tests passed; targeted Ruff and Pyright passed; the mixed `deps` case was rejected under system Python 3.9.6; `git diff --check` passed; exact-head GitHub CI and the dedicated scope-guard workflow are green.

qinhaihong-red:
## Resolution — head `23aa53f`

The last open Standards finding is closed. Root suite **1179 passed**; member suite unaffected (366/17); CI green including the dedicated guard workflow.

**Package-level `changelog-sections` are now resolved per package.** I took your option 1 (derive) rather than option 2 (reject): rejecting a legitimate upstream feature would trade one incompleteness for a different restriction, while deriving keeps the "derive, don't restate" line the rest of this PR follows.

- `releasing_types(config, path)` resolves ONE package's effective sections — its own value → the top-level value → release-please's built-in defaults.
- New `touched_packages(config, changed_files)` — the root `"."` always in scope, plus any declared package whose directory the PR touches.
- `verdict(...)` unions those packages' releasing sets.

**One deliberate deviation from your phrasing, flagged rather than buried:** you described detecting types that release *both* the root and the touched member; I used the **union** (releasing for ANY touched package). A type that releases only the root still reproduces the original harm — a member-scoped change bumping `gda` — and a guard must have no false pass. Under today's config (no package-level overrides) the two predicates coincide exactly; the difference only appears in the hypothetical configs your finding is about. Justified in a comment at the predicate.

**Verification of your exact repro** (top-level `feat` visible; root and member each overriding with `security`; title `security(gda-balancing): …`; files `libs/gda-balancing/pyproject.toml` + `README.md`): now **exit 1**, message naming `README.md`. The same title/files against the real shipped config still exits 0 (this repo declares no `security` section), and `deps` straddling still exits 1 — existing behaviour unchanged.

**The fallback set was checked, not invented:** with no `changelog-sections` at either level the guard falls back to release-please's own `DEFAULT_CHANGELOG_SECTIONS` non-hidden entries — `feat, fix, perf, revert`. Note `deps` is *not* among them (it is this repo's own addition), so a drift test pins the shipped config's `feat, fix, deps, revert` for both packages rather than trusting the fallback to cover it.

**Test honesty:** two discriminator probes were run — against the pre-change guard all 6 new cases fail, and swapping the union for an intersection fails exactly the two cases that pin the conservative predicate. The new tests are not vacuous.

ADR-0038, the module docstring, and the workflow comment were all updated to describe the per-package derivation and the conservative union, so the record matches the implementation.

**External precondition acknowledged and unchanged:** the guard is advisory until "Member releasing-PR scope guard" is a required status check on `main` — I verified the active ruleset carries only `deletion`, `non_fast_forward`, `pull_request`. Agreed it does not block this PR, and it is recorded as a hard precondition for #528's title flip and live proof.

qinhaihong-red:
Rechecked the author's latest resolution at exact head `23aa53f6fb4dc9c45f223d0efa01907134d08b7c`.

## Resolved

- The previous package-level `changelog-sections` bypass is closed for explicit section lists. The exact repro — top-level `feat`, root/member `security` overrides, a `security(gda-balancing): ...` title, and member + root paths — now yields `releasing=True, ok=False`.
- Resolving sections per touched package and taking the conservative union is an acceptable fail-closed ownership rule. It can reject a mixed PR when only one train would release, but it cannot silently pass the original root-absorption failure.
- The previous tag-derivation, title-edit trigger, and explicit-component-contract findings remain resolved.

## Still open — Standards / request changes

### 1. The no-sections fallback uses the wrong defaults for this repository's `python` release strategy

`scripts/release_scope_guard.py:54-85` copies the generic visible `DEFAULT_CHANGELOG_SECTIONS` as:

```text
feat, fix, perf, revert
```

But this repository declares `"release-type": "python"`. release-please's Python strategy installs its own defaults, whose visible types additionally include `deps` and `docs` ([pinned release-please v17.3.0 source](https://github.com/googleapis/release-please/blob/v17.3.0/src/strategies/python.ts#L31-L49)).

Concrete valid repro: remove `changelog-sections` at both the package and top levels, retain `release-type: python`, and use mixed member/root paths. The current guard reports:

```text
deps: bump pydantic  -> releasing=False, ok=True
docs: update guide   -> releasing=False, ok=True
```

Both types survive the Python strategy's default filtering and can patch-bump the touched Python packages, so this is a permissive false negative in the load-bearing guard. `tests/test_release_scope_guard.py:115-124` currently codifies the generic fallback rather than the effective Python fallback.

The shipped configuration explicitly declares sections, so this path is not active today; however the helper, ADR, workflow comment, and test claim that the upstream fallback is supported. Please either:

1. fail loudly when no effective explicit `changelog-sections` exists; or
2. derive defaults from each package's effective `release-type`.

While fixing inheritance, preserve explicit empty arrays by key presence rather than `package.get(...) or ...`: release-please uses nullish precedence, so `[]` overrides the top-level value. The current divergence is conservative-only, but it is still not the claimed upstream semantics.

### 2. The load-bearing scripts still run under an unpinned system Python

The repository's products and uv environments are explicitly Python 3.13, but two production paths still bypass them:

- `.github/workflows/release-scope-guard.yml:62` runs `python3 scripts/release_scope_guard.py`
- `.github/workflows/release.yml:450` runs `python3 scripts/release_tags.py`

Neither job establishes Python 3.13 before those calls. `.python-version` and an existing `.venv` do not redirect a bare `python3`; uv's project environment is used through `uv run` or activation ([uv project documentation](https://github.com/astral-sh/uv/blob/main/docs/guides/projects.md)).

Testing macOS system Python 3.9.6 was a deliberate probe of those real call sites, and `from __future__ import annotations` fixes the known import crash. However, 3.9 itself is incidental rather than a repository support target, and that compatibility currently has no committed CI coverage. The green pytest/Ruff/Pyright jobs run under uv-managed Python 3.13; they do not prove the bare-system-Python contract.

Prefer explicitly setting up Python 3.13 in both jobs, then running the stdlib-only scripts without syncing project dependencies. If bare system Python is intentionally retained, define a minimum version and add a corresponding CLI smoke job.

## External precondition

The active `main` ruleset still has no required-status-check rule. `Member releasing-PR scope guard` remains advisory even though its exact-head run is green. This does not block the current partial PR while releasing member titles remain disabled, but it must block #528's convention flip and live proof.

## Verdict

The exact prior package-override repro is resolved, but the generalized guard still has one false-pass configuration and its runtime interpreter is not reproducibly fixed. Verdict: **request changes**.

Validation at this head: 90 focused tests passed; targeted Ruff, format, and Pyright passed; `git diff --check` passed; the exact override repro is rejected; the Python-default `deps`/`docs` repro passes incorrectly as described above; system Python 3.9.6 runs both current CLIs; exact-head GitHub checks are green.

qinhaihong-red:
## Resolution — head `6aeddc2`

Both findings closed. Root suite **1181 passed**; member suite 366/17 unaffected; CI green including the dedicated guard workflow.

### 1. Sections-less fallback → fail loud, and inheritance now keys on presence

You were right that the generic `DEFAULT_CHANGELOG_SECTIONS` are the wrong table for a `release-type: python` repo. Rather than reproduce upstream's per-strategy default tables — which would plant exactly the second authority your earlier findings on this PR objected to, and would need tracking for every release type — the fallback is **removed**: a package with no effective explicit `changelog-sections` (neither its own nor top-level) now raises `ScopeGuardConfigError` naming the package and the reason (defaults vary by `release-type`; the python strategy makes `deps` and `docs` visible).

The false pass was **demonstrated, not asserted** — the pre-change script against a sections-less config with a straddling `deps: bump pydantic`:

```
OLD → "declares no releasing type; guard is a no-op."   exit=0
NEW → "declares no changelog-sections for package '.' … would let a
       releasing title pass unguarded."                  exit=1
```

Inheritance now resolves by **key presence** via a sentinel, not `package.get(...) or top_level`, so an explicit `[]` overrides the top-level list per release-please's nullish precedence. An explicit `[]` **is** explicit: it yields the empty set and does not raise (documented in the docstring, the test, and ADR-0038). That test genuinely discriminates — under the old `or` expression the same package resolved to `{feat}`.

The drift test pinning the shipped config's `feat, fix, deps, revert` for both packages is untouched, which is what keeps the live path honest.

### 2. Pinned interpreter

Both jobs now run `setup-python-env` with `sync: none` and invoke the scripts as `uv run --no-project --python 3.13 python scripts/…` — explicit interpreter, still no project resolution (the scripts stay stdlib-only by design). The exact invocation was executed locally on both scripts before committing.

**Three call sites converted, not two:** `release.yml`'s inline `python3 -c` JSON parser for `RELEASE_PRS` sits in the same `release-pr` job and was equally unpinned; converted and verified against real branch JSON that the multi-line quoting survives. The job's later conditional `astral-sh/setup-uv` step was **removed** as redundant under the new job-level setup — I verified the ordering leaves uv available for the tag gate, the JSON parse, and both `uv lock` calls.

All "3.9 support" framing is gone (it was an incidental probe environment, not a repo target); `from __future__ import annotations` stays as hygiene. ADR-0038 and the workflow comments now state the actual contract: stdlib-only scripts on the repo's pinned Python, no project sync required.

### External precondition — unchanged and acknowledged

`Member releasing-PR scope guard` is still advisory: I re-verified the active `main` ruleset carries only `deletion`, `non_fast_forward`, `pull_request`. Agreed it does not block this PR while member titles stay non-releasing, and it is recorded as a hard precondition for #528's flip and live proof.

PR #532, opened 2026-07-20
chore: lift the non-releasing title discipline for gda-balancing (#528)

qinhaihong-red:
Part of #528 — step 3's convention flip. The live proof (this issue's remaining AC) arrives with the first genuinely-releasing member PR; see "What happens next".

## What

Every precondition ADR-0037 set for the flip is now met — publish tail wired, tag gate extended to all manifest packages (PR #529), and **`Member releasing-PR scope guard` made a required status check** (repo settings, done). So the discipline is lifted: **gda-balancing PRs now use truthful `feat`/`fix` types** and release on their own train.

Records updated: ADR-0037 (dated flip note in its existing Outcome block), ADR-0038 (the guard is load-bearing now, not belt-and-braces), and `libs/gda-balancing/AGENTS.md` — the last of these matters most in practice, since that is what an agent working in the package actually reads; the ADRs are the decision record.

## Two limits the flip deliberately does NOT touch

- **It follows the exclusion, not the "is it gda?" intuition.** `examples/**` (panda) and every other un-excluded path is still absorbed by the root package, so work there keeps a non-releasing type or it bumps `gda`. Flipping those too would be the easy over-generalization.
- **A PR spanning the member and anything outside it is still absorbed.** The guard refuses it when releasing-typed, and the documented answer is to **split** — not to downgrade the type, because a member dependency change that warrants a release must not be recorded as a chore (the reasoning that killed the earlier chore-discipline proposal).

**This PR demonstrates its own rule:** it touches root ADRs *and* the member's AGENTS.md, so it is mixed-path and correctly titled `chore`. Verified against the shipped guard — the same file set under `feat(gda-balancing): …` is blocked.

## First-release changelog

Recorded in ADR-0037: the member's foundation (#502, #504) landed under `chore` titles by the very constraint this issue retires, so those commits are invisible to release-please's changelog. The first member release gets **hand-authored release notes** describing the schema core — a one-time cost of the old discipline; every release after it is fully generated.

## What happens next

The next member-only PR carrying a real `feat`/`fix` will make release-please propose a `gda-balancing` Release PR (0.0.0 → 0.1.0). Merging that Release PR publishes to PyPI through the tail wired in #529 — so whether to cut the first release, and when, stays a deliberate decision rather than a side effect. That merge is also what produces this issue's live evidence (no `gda` bump, no spurious root Release PR).

### DIFF
--- a/.github/workflows/release-scope-guard.yml
+++ b/.github/workflows/release-scope-guard.yml
@@ -31,15 +31,18 @@ jobs:
       contents: read
       pull-requests: read
 
-    # NOTE: a guard only blocks a merge once it is a REQUIRED status check. Add
-    # "Member releasing-PR scope guard" to the branch protection rules for
-    # `main` — a repo-settings action a maintainer must take, which no workflow
-    # can do for itself. Until then this reports but does not enforce.
+    # LOAD-BEARING (#528): "Member releasing-PR scope guard" is a REQUIRED
+    # status check on `main`, and gda-balancing PRs now carry truthful
+    # releasing types (`feat`/`fix`/...) — so this job is what actually stops a
+    # releasing-typed PR that spans `libs/gda-balancing/**` AND anything
+    # outside it from merging into both release histories (ADR-0037's flip
+    # note, ADR-0038).
     #
-    # Currently belt-and-braces regardless: member PRs are still held to
-    # non-releasing types by discipline. It becomes truly load-bearing when #528
-    # flips that discipline and member PRs start carrying truthful feat/fix
-    # titles.
+    # The required-check configuration is part of the guarantee, not a
+    # convenience: a guard only blocks a merge while it is required, and no
+    # workflow can grant itself that status. If the rule is ever removed from
+    # the ruleset, this job silently degrades to advisory and the flipped title
+    # discipline loses its enforcement — keep it required.
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -197,9 +197,9 @@ jobs:
       # Idempotent via skip-existing: PyPI files are immutable, so a re-run from
       # the same tagged commit skips whatever already landed and uploads the rest,
       # mirroring the --clobber idempotency of the gh release upload downstream.
-      # Configuring the PyPI pending publisher is the one-time human step on #207;
-      # until then this step fails the release — intended, it gates tag creation on
-      # a working publisher.
+      # The trusted publisher for `gda` is configured (#207); this step failing
+      # is therefore a real publish failure, and it gates tag creation on a
+      # working publisher — no tag is minted unless PyPI accepted the upload.
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
--- a/docs/adr/0037-workspace-members-join-the-single-authority-release-model.md
+++ b/docs/adr/0037-workspace-members-join-the-single-authority-release-model.md
@@ -59,7 +59,9 @@ separate release trains, and no hand-edited version anywhere.**
   `libs/gda-balancing/**` keep non-bumping conventional-commit types
   (`chore`/`docs`/`refactor`/...) until a member release is deliberately
   wanted. Registration governs the version line; it does not open the release
-  train.
+  train. *(Historical — superseded by the dated **Flip** note in Consequences:
+  the member's path is excluded and the discipline is lifted for it. Still
+  current for every un-excluded path, e.g. `examples/**`.)*
 - **Publishing gda-balancing is deferred.** No PyPI publish tail, no trusted
   publisher, no artifact upload. Known dormant consequence, accepted: if a
   member Release PR were merged today, the cut job would leave a tag-less
@@ -67,6 +69,9 @@ separate release trains, and no hand-edited version anywhere.**
   maintenance tag gate (#79/#82) only guards the root package's tag. The
   member's first-release issue must wire its publish tail, extend the tag
   gate to its component, and only then may member PRs adopt releasing types.
+  *(Historical — all three preconditions were met by #528/#529; see the dated
+  **Flip** note in Consequences. Publishing is wired but no member release has
+  been cut yet.)*
 
 ## Consequences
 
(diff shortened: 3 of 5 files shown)


qinhaihong-red:
Review at exact head `6faf9d2a875d1b91904ec3997aab039711b74acf`.

## Standards

- **[P2][Hard violation] The title-discipline flip did not fully update the old operational guidance.** [`.github/workflows/release-scope-guard.yml:34-42`](https://github.com/aigengame/godot-agent/blob/6faf9d2a875d1b91904ec3997aab039711b74acf/.github/workflows/release-scope-guard.yml#L34-L42) still tells maintainers to add the check as required, says it only reports rather than enforces, and says member PRs are still held to non-releasing titles. That now conflicts with the active ruleset and with the flip recorded in ADR-0037, ADR-0038, and `libs/gda-balancing/AGENTS.md`. Rewrite the comment to say that `Member releasing-PR scope guard` is required on `main`, is now load-bearing, member-only PRs may use truthful releasing types, and mixed-path releasing PRs are rejected. The comment should also state that required-check configuration must remain in place. The older present-tense ADR statements about the former discipline should be marked clearly as historical or superseded by the dated Flip outcome.

## Spec

- **[P2] #528 requires every recorded copy of the old rule to be updated, but the workflow copy was missed.** This does not change runtime behavior today, but it leaves an authoritative operational surface describing the opposite enforcement and title policy, which can mislead future workflow or ruleset maintenance.

Validation performed:

- The default-branch active ruleset includes the exact required context `Member releasing-PR scope guard` and has no bypass actors.
- The current mixed-path file set passes under the actual `chore` title and is rejected under a hypothetical `feat(gda-balancing)` title.
- `79` focused scope-guard tests pass; `git diff --check` is clean.
- Required PR checks are green and the PR is mergeable.
- The remaining live-proof AC is still tracked by open issue #528; its latest maintainer progress comment explicitly orders required check → flip → live evidence, so that deferral is not an additional finding.

**Summary:** 1 Standards finding, 1 Spec finding; highest severity P2.

qinhaihong-red:
## Resolution — head `6c5a84a`

Both findings are the same miss, and it was a real one: I swept the ADRs and `AGENTS.md` but not the guard's **own workflow comment** — the surface a maintainer touching the ruleset actually reads. CI green.

**`.github/workflows/release-scope-guard.yml`** — the comment now states that `Member releasing-PR scope guard` **is** a required status check on `main`, that the job is therefore **load-bearing** (it stops a releasing-typed PR spanning the member and anything outside it from merging), and that member-only PRs carry truthful releasing types. It also makes the point you asked for explicitly: keeping the rule in the ruleset is **part of the guarantee, not a convenience** — if it is removed the job silently degrades to advisory and the flipped discipline loses its enforcement.

**ADR-0037's present-tense statements about the former discipline** now carry inline historical markers pointing at the dated **Flip** note, in the same style this file already uses for its superseded bullets:
- the "non-releasing title discipline continues after registration" bullet — marked historical for the member's path, and explicitly **still current for every un-excluded path** (`examples/**`), so the marker cannot be read as a blanket lift;
- the "must wire its publish tail … and only then may member PRs adopt releasing types" bullet — marked as satisfied by #528/#529, noting publishing is wired but no member release has been cut;
- the changelog bullet's "(its history is non-releasing by discipline)" parenthetical — marked historical, with the reason the changelog still starts empty (pre-flip history is `chore`, hence the hand-authored first release notes).

**One unrelated stale comment the re-sweep surfaced, fixed in the same pass and flagged rather than folded in silently:** `release.yml`'s PyPI step still said configuring the `gda` trusted publisher was a pending one-time human step and that the step fails "until then" (#207). That publisher has been configured since `gda` started publishing, so the comment now says a failure there is a real publish failure that gates tag creation. Out of this PR's scope, but leaving a second contradictory operational comment in place while fixing the first would have been the same defect.

I re-ran the sweep across `.github/`, `scripts/`, `docs/` and the package docs; the remaining "non-releasing" hits are `scripts/release_scope_guard.py` describing commit *types* (not the discipline), which is correct usage.

qinhaihong-red:
Recheck at exact head `6c5a84a13b4d2392ba712107189aeee8c6e6c2d6`.

The previous P2 is fully resolved: the scope-guard workflow now accurately documents the required/load-bearing check, ADR-0037 marks the former rules as historical, and the author response matches the diff. The small `release.yml` comment cleanup is accurate and does not change behavior.

One P2 remains. This was present in the original PR head and was missed in the first review; it was not introduced by the fix commit.

## Standards

- **[P2] Do not require non-releasing titles for genuine root-package releases.** [ADR-0038 lines 125–127](https://github.com/aigengame/godot-agent/blob/6c5a84a13b4d2392ba712107189aeee8c6e6c2d6/docs/adr/0038-gda-balancing-leaves-the-uv-workspace.md#L125-L127) say that every un-excluded path, including `root`, still needs a non-releasing type. That conflicts with ADR-0007 and ADR-0034: a genuine root `feat` or `fix` should use its truthful releasing type and produce a `gda` version bump. The accurate rule is already present in ADR-0037: an un-excluded-path change either uses a non-releasing type when it should not release `gda`, **or** uses a truthful releasing type and bumps `gda`.

## Spec

- **[P2] The convention-flip scope is stated too broadly.** #528 lifts the old title constraint only for `libs/gda-balancing/**`; it does not require genuine root features and fixes to become non-releasing. Grouping `examples/**` and `root` under “still needs a non-releasing type” would suppress legitimate `gda` releases. Please change this sentence to say that un-excluded paths remain attributed to the root release train: changes that should not release `gda` use a non-releasing type, while genuine root features/fixes keep truthful releasing types and bump `gda`.

Validation:

- `79` focused scope-guard tests pass.
- `git diff --check` is clean.
- Required CI checks are green and the PR is mergeable.

**Summary:** the previous finding is resolved; one remaining P2 appears on both review axes.

qinhaihong-red:
## Resolution — head `9d64227`

Finding accepted; it was a real logic error I introduced, not a wording nit. Writing "every un-excluded path still needs a non-releasing type" grouped root `gda` work with `examples/**` and would have suppressed legitimate `gda` releases — directly contradicting ADR-0007/ADR-0034, where a genuine feature or fix bumping `gda` *is* the normal flow.

**The rule as now stated in both flip notes:** un-excluded paths stay attributed to the root release train, so a title there must honestly describe the change's effect on `gda` —

- a genuine `gda` feature or fix keeps its **truthful releasing type** and bumps `gda`;
- work that merely *lives* in an un-excluded path without being a `gda` change (`examples/**`, panda — the standing case) takes a **non-releasing type** so it does not bump `gda` falsely.

The closing line makes the distinction explicit: the flip changes **which paths are attributed to the root**, not the requirement that the type be truthful.

**Corrected in three places**, since I had written the same over-broad sentence more than once:
- ADR-0038's Outcome note (the lines you cited);
- ADR-0037's Flip note, which had the same grouping in its "scoped to the excluded path only" bullet;
- ADR-0037's historical marker on the old discipline bullet, which said the discipline was "still current for every un-excluded path" — now it says the underlying *truthfulness* rule still holds there, which is the accurate residue.

`libs/gda-balancing/AGENTS.md` was checked and needed no change: it only claims the lift is "for this directory only" and never characterizes what other paths must do.

I also corrected the same error in my own working notes so the wrong formulation does not get re-applied from memory in a later session.

Issue #505, opened 2026-07-15
[gda-balancing] RPG genre template + CRPG/JRPG/ARPG reference fixtures


qinhaihong-red:
## Parent

- PRD #501.

## What to build

Ship the RPG family Genre template as an instance of the Standard Schema — data, never code paths keyed on genre: directly allocatable primary stats; derived resource/offense/defense/mobility stats with sensible default formulas; bounded percentage/probability attributes with mandatory caps — the tier vocabulary declared as named facet compositions (bADR-0002). Ship three paper Reference fixtures — **CRPG, JRPG, and ARPG** — each exercising the family template end to end through the CLI (instantiate → validate → emit) as executable consumers and golden examples. The three must differ meaningfully along the subtypes' emphases (not copies with renamed ids): e.g. the ARPG fixture exercises the combat-pacing-relevant attributes (attack/cast speed, mobility, hit/dodge modifiers), the CRPG fixture the allocation-heavy primary-stat spread, the JRPG fixture the formula-driven derived tiers.

## Acceptance criteria

- [ ] The RPG template validates as a plain Schema instance; no genre-keyed code paths exist.
- [ ] All three fixtures (CRPG, JRPG, ARPG) run end to end (instantiate → validate → emit) in the test suite as golden consumers.
- [ ] The three fixtures differ meaningfully along their subtype emphases, exercising distinct template paths.
- [ ] Template defaults and formulas use industry-standard terminology and are documented alongside the template.

## Blocked by

- #504

🤖 Generated with [Claude Code](https://claude.com/claude-code)

qinhaihong-red:
Terminology synced with the reworked Standard Schema design (PR #521): the tier vocabulary is a named facet composition (bADR-0002); "modifier tier" wording replaced by bounded percentage/probability attributes with mandatory caps.

qinhaihong-red:
Removed `ready-for-agent`: blocked by #504, whose own design gates (#503, #518) are open — the label returns when this becomes truly pick-up-ready (per PR #521 follow-up re-review).

qinhaihong-red:
Schema 2.0 supersedes this coarse template issue with independently verifiable slices: #553 owns minimal template instantiation, #546-#552 own the RPG tracer bullet, #555-#565 own the remaining RPG contracts, and #566 owns the complete RPG template release plus CRPG/JRPG/ARPG reference sources. No separate 1.x-shaped work remains here.

Issue #503, opened 2026-07-15
[gda-balancing] Standard Schema design — research, draft & review (bADR)


qinhaihong-red:
> **Vocabulary supersession note (2026-07-18):** "attribute tiers (primary / derived / modifier) with defaults" is realized as **named facet compositions** over orthogonal attribute facets (bADR-0002 — defaults are template-shipped `base` values), and the modifier half is designed as first-class **Effects** (bADR-0006). The text below is the point-in-time gate definition; the bADRs on PR #521 are the design authority.

## Parent

- PRD #501.

## What to build

The research → design → review pass for the Standard Schema — the toolkit's sole spec and single authority source. Research industry practice for stat-system and formula-as-data representation (industry-standard terminology only; no invented terms). Draft the Schema core: document structure and versioning; attribute tiers (primary / derived / modifier) with defaults; formulas-as-data with caps and clamps; the subsystem sections (combat, builds/modifier pools, encounters, growth, economy) at least as reserved shapes; validation semantics (element-level typed refusals at the boundary funnel); emission format (JSON first, format-extensible); and the Standard Schema's **machine-readable self-description** — how the Standard Schema publishes/introspects itself so documents are mechanically validatable against it (US8; distinct from #518's CLI-surface self-description). Record the decisions as bADR(s); update the glossary (`BALANCING-CONTEXT.md`) for new terms. HITL: maintainer review gates implementation.

## Acceptance criteria

- [ ] bADR(s) record the schema document structure, attribute-tier model, formula representation, validation semantics, and versioning approach.
- [ ] The Standard Schema's machine-readable self-description mechanism is part of the recorded design.
- [ ] The glossary is updated for every new term the design introduces.
- [ ] Maintainer review approval is recorded on this issue before any implementation slice starts.
- [ ] No implementation in this slice.

## Blocked by

None - can start immediately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

qinhaihong-red:
## Research provenance — Standard Schema design (non-normative)

Industry-practice research backing the Standard Schema design pass (bADR-0001..0005, PR to follow). **This comment is provenance, not authority** — decisions are recorded in the bADRs; this documents what the research found, with every precedent provenance-labeled by ecosystem.

[...]

PR #533, opened 2026-07-20
feat(gda-balancing): RPG genre template behind template list/get

qinhaihong-red:
Part of #505 (first of two slices; the fixtures slice will close it). Also produces #528's live evidence once merged: this is a releasing-typed PR touching only `libs/gda-balancing/**` — after merge, **no** root `gda` Release PR may appear, while a `gda-balancing` 0.1.0 Release PR should auto-open (to be left unmerged; cutting the first release is a deliberate later decision).

## What ships

- **`meta.genre` lineage** — the optional, purely descriptive `GenreLineage {family, variant?}` bADR-0001 deferred to #505. Structural golden regenerated; the semantic catalog is byte-unchanged (nothing branches on genre — AC1's no-genre-keyed-code-paths).
- **The RPG family Genre template** — `src/gda_balancing/templates/rpg.json`, a packaged canonical Standard Schema instance declaring the bADR-0002 tier vocabulary (`primary`/`derived`/`tertiary` as named facet compositions), allocatable primaries, formula-derived resource/offense/defense/mobility stats (named-form-heavy, all knobs as `parameters`), bounded tertiary rates with mandatory caps, and one legal worked-example effect.
- **`template list` / `template get <id> [--out]`** — bADR-0007's declared group; **get is instantiate** (bADR-0012): a Genre template already is a valid Design document. Descriptors mirror `schema get` (Literal input → automatic usage on unknown id; artifact-sink; conformance harness sweeps both automatically).
- **bADR-0012** — template representation & delivery, the instantiation shape, `meta.genre`'s in-place 1.0 landing (pre-publication only; 1.1 bump considered-and-rejected), the isolation-gate carve-out, and the fixture lineage relation.
- **Isolation-gate amendment** — `templates/` exempt from the stray per-game-config JSON scan (a Genre template is genre-generic, not per-game config); template *and* fixture JSON now join the game-identity vocabulary scan (closing a prior gap).
- **`docs/templates/rpg.md`** — every default and formula documented with multi-source provenance (AC4); judgment calls and genuinely contested conventions (crit-damage +50% vs +100%, CDR cap philosophies, the toughness-slot naming) are flagged, not averaged.

## Research-forced corrections (vs the naive draft)

- Dexterity (9/14 surveyed systems) over Agility; Intelligence (13/14) over Intellect; Vitality kept as a documented judgment call.
- The XP curve is **polynomial** (triangular family) — the attested shape per Schreiber's *Game Balance Concepts* and the D&D/Pathfinder XP tables; true constant-ratio exponential is rare in balanced games and left out of the template.
- `damage_reduction` ships the cross-franchise rational curve `armor/(armor + K)` capped at the 75% convention — the one attested case where the expression-tree fallback (not a named form) is the honest representation.
- Movement speed is a flat 100% baseline (no surveyed system derives it from a primary stat); `initiative` (Dexterity-derived action order — D&D initiative, JRPG turn order) carries the derived-mobility slot.

## Verification

- 400 passed / 28 skipped; pyright and ruff clean; conformance harness auto-covers the two new descriptors.
- `template get rpg` ≡ `design format rpg.json` ≡ committed bytes (three-way identity); `design validate` exit 0 (AC1).
- e2e key path through the installed console script: instantiate (`--out`) → validate → emit, plus `template list`.
- **Wheel/sdist packaging verified**: `uv build` shows `gda_balancing/templates/rpg.json` in both artifacts (first non-`.py` payload under `src/`).


qinhaihong-red:
## `$review`: Standards / Spec

Reviewed exact head `c78c218ed0661deb16e2ddeb1563c53fdf2b464d` against base `370a5121d1f5c0b63a4980df6009c96e08ee7427`.

**Verdict: Request changes.**

## Standards

1. **Hard — Reference fixtures are missing.** Accepted [bADR-0012](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/docs/badr/0012-genre-template-representation-and-delivery.md) requires them under `tests/fixtures/reference/`, while [`docs/templates/rpg.md`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/docs/templates/rpg.md) claims they exercise omitted paths. Only `minimal_design.json` exists. This violates the repository's completeness and artifact-synchronization rules.

2. **Hard — `tertiary` does not encode “bounded rates.”** [`rpg.json`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/templates/rpg.json) constrains only `accepts == {effects}`. An unbounded numeric tertiary attribute validates. Root cause is the Standard Schema: [`Tier`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/attributes.py) cannot express domain unions or required bounds, so it cannot represent `(percentage | probability) × (direct | formula) × bounded`.

3. **Hard — `crit_chance` bounds merely restate `[0,1]`.** bADR-0002 requires probability design bounds to narrow the intrinsic domain, never merely restate it. The funnel accepts this template's `{floor: 0, cap: 1}`, revealing a Standard Schema semantic-rule gap in addition to the template conflict.

4. **Hard — movement-speed semantics drift.** Data encodes `domain: number`, `base: 100`; the documentation calls it a percentage-of-baseline. bADR-0002 requires fractional representation for percentages and says the 0–100 convention never appears in a Design document.

5. **Hard — template identity consistency is only checked one-way.** [`test_template_ids_have_one_authority`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/tests/test_template_command.py) ensures registered IDs have resources, but does not reject orphan packaged `*.json` resources. That falls short of bADR-0012's single-authority rule.

**Standards count: 5 hard findings, 0 judgment calls.**

## Spec

1. **Required — Standard Schema root defect: mandatory caps are not enforceable.** Issue #505 requires “bounded percentage/probability attributes with mandatory caps — the tier vocabulary declared as named facet compositions.” `Tier` can constrain only `domain`, `base`, and `accepts`, while `bounds_required_for_domain` accepts any non-empty bounds. Removing only `cooldown_reduction.bounds.cap` still yields `{"valid": true}`. Fix bADR-0002/model/semantic rules, not merely `rpg.json`.

2. **Required — Standard Schema root defect: derived resources lack capacity/current-state semantics.** PRD #501 asks for “derived survival/offense/defense/mobility stats” that are “inspectable and simulatable.” `health` is formula-derived capacity, but the shipped periodic `regeneration` modifies that same attribute. bADR-0006 clamps deltas only to the target's bounds; `health` has none, and `Bounds` accepts literals rather than formula/attribute references. Regeneration can therefore exceed derived maximum health. The Schema needs explicit resource capacity/current semantics or dynamic bounds; the template alone cannot repair this safely.

3. **Required — template-local mobility/unit inconsistency.** Issue #505 requires “derived … mobility stats with sensible default formulas.” `movement_speed` is untiered, direct, and encoded as `domain: number`, `base: 100`, while its documentation calls it a “percentage-of-baseline convention”; this contradicts bADR-0002's fraction convention. `initiative` is then labeled mobility to fill the derived slot although it models action order, not movement. Use consistent units/category and provide a genuine derived mobility path.

4. **Required — AC4 provenance remains unauditable.** Issue #505 says defaults/formulas must “use industry-standard terminology and [be] documented alongside the template.” The page names systems and survey counts but provides no bibliography, editions, rule sections, or URLs; even the “14 systems surveyed” list ends with an ellipsis. The claimed consensus cannot be reproduced.

5. **Deliberately deferred, but issue #505 remains incomplete.** Its exact AC says all CRPG/JRPG/ARPG fixtures must run end-to-end and differ meaningfully. This PR explicitly defers them to slice two; no such fixtures exist yet.

No other material scope creep.

**Spec count: 4 required findings and 1 deliberately deferred issue-level gap.**

**Summary:** Standards 5; Spec 5. The worst issue is the Standard Schema's inability to model and enforce a derived resource capacity separately from its mutable runtime value, which makes the shipped `health`/`regeneration` pair semantically unsafe even though structural validation and current CI pass.

qinhaihong-red:
## `$agent-skills:code-review-and-quality`

Reviewed exact head `c78c218ed0661deb16e2ddeb1563c53fdf2b464d`.

**Verdict: Request changes.** The CLI implementation is simple, safe, and well-tested, but the first concrete RPG consumer exposes several blocking model defects. Three are rooted in the Standard Schema and should be fixed there rather than patched around in `rpg.json`.

### Required findings

1. **Correctness / Architecture — the Standard Schema cannot represent derived resource capacity separately from mutable current state.**

   `health` evaluates to a derived capacity of `200`, but it has no bounds and the shipped `regeneration` effect mutates that same attribute. A direct seam probe gives `derived_health=200`, then `clamp_to_attribute(200 + 5×2, health)=210`. This is not merely a missing literal cap: [`Bounds`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/attributes.py#L57-L68) accepts only literal floats, so it cannot say `current_health <= formula-derived max_health`; a parameter-valued cap is structurally refused. A conventional RPG resource explicitly distinguishes maximum and current values—e.g. the [D&D Basic Rules](https://www.dndbeyond.com/sources/dnd/br-2024/playing-the-game) define current HP separately and clamp healing to HP maximum.

   **Root fix:** reopen bADR-0002/bADR-0006 and the #503/#504 Standard Schema contract to add explicit resource `capacity/current` semantics or formula/reference-valued dynamic bounds. Adding a fixed cap to this template would stop following `vitality` and would not repair the model.

2. **Architecture / Orthogonality — `domain`, bounds obligation, and tier membership are coupled in contradictory ways.**

   The `tertiary` pattern constrains only `accepts`, so an unbounded `number/direct` tertiary attribute validates. Removing only `cooldown_reduction.bounds.cap` also validates because the domain rule accepts a floor-only object. Conversely, the global “every percentage/probability must declare bounds” rule forces bounds where the genre intentionally has none: the RPG page admits `crit_damage.cap=3.0` is schema-mandated rather than attested, while `movement_speed` evades the rule by encoding a documented percentage as `number: 100`. `crit_chance [0,1]` then merely restates its intrinsic probability domain.

   **Root fix:** make numeric scale/unit independent from constraint policy, and let tier predicates express domain/base sets plus required floor/cap semantics. The RPG tier can then require capped rates without forcing arbitrary caps onto every percentage in every genre.

3. **Correctness / Extensibility — the advertised tuning surface excludes important RPG knobs and accepts values that immediately fail evaluation.**

   The Standard Schema stores `parameters` as `id -> float`; direct bases, bounds, and polynomial coefficients are literal-only. Therefore crit base/caps, primary bases, and the XP coefficients cannot participate in the Phase-2 parameter inventory even though bADR-0003 calls that inventory the design's tuning surface. [`test_rpg_template_knobs_are_parameters_not_literals`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/tests/test_template_command.py#L174-L193) proves only that no declared parameter is dead; it does not prove that all tuning knobs are parameters.

   The absence of parameter constraints is already observable: setting the accepted `armor_base` parameter to `-120` still makes `design validate` return `{"valid": true}`, but `evaluate_bases` immediately refuses because `armor / (armor + armor_pivot)` divides by zero.

   **Root fix:** give parameter declarations machine-readable constraints/units/descriptions and decide which direct values, bounds, and curve coefficients may reference them. Add a regression proving every shipped parameter range preserves definition-time evaluatability.

4. **Correctness — `experience_to_next_level` applies a cumulative triangular curve to an incremental field.**

   The current `50·L + 50·L²` produces next-level costs `100, 300, 600, ...`; summing those costs makes cumulative XP cubic/tetrahedral. The cited [Game Balance Concepts example](https://gamebalanceconcepts.wordpress.com/2010/07/) instead says level transitions cost `1000, 2000, 3000, ...`—linear increments whose **cumulative** XP-to-level relation is triangular. The referenced [D&D](https://www.dndbeyond.com/sources/dnd/br-2024/creating-a-character) and [Pathfinder](https://legacy.aonprd.com/coreRulebook/classes.html) tables likewise publish cumulative totals, not evidence that each incremental cost is triangular.

   Rename the attribute to cumulative XP and use the appropriate inverse/lookup semantics, or keep `experience_to_next_level` and use a linear default. This is a template/research interpretation defect, not a Standard Schema limitation.

5. **Consistency / RPG coverage — the template conflates defense, mitigation, initiative, and movement to fill required slots.**

   `armor` is derived from Dexterity using Diablo II “Defense Rating” as provenance, but Blizzard's [own terminology](https://classic.battle.net/diablo2exp/basics/characters.shtml) says Defense Rating reflects the ability to **dodge** attacks; the template then feeds it into an armor damage-reduction curve. `initiative` is categorized as mobility although it models action order, while actual movement is an untiered percentage encoded as a number.

   Split these orthogonal mechanics (`armor`/mitigation vs `defense` or `evasion`; `initiative`/action order vs movement), use consistent units, and either provide a genuine derived mobility stat or correct issue #505's assumption that every RPG family baseline needs one.

### Five-axis summary

- **Correctness:** blocked by findings 1, 3, 4, and 5.
- **Readability & simplicity:** command code is straightforward; the numeric documentation is readable but its source trail and several labels do not support the claims they carry.
- **Architecture:** blocked by findings 1–3; attribute maps are extensible, but the underlying facets cannot express the first real consumer's invariants.
- **Security:** pass. The `Literal`-bound template id prevents path traversal; no secrets, untrusted parsing path, or new dependency was introduced.
- **Performance:** pass. Registry/resource work is bounded and linear in the small packaged template; no hot-path or unbounded operation was added.

### Verification

- `400 passed, 28 skipped`
- Pyright: clean
- Ruff: clean
- wheel and sdist build successfully; both contain `gda_balancing/templates/rpg.json`
- `git diff --check`: clean
- GitHub checks: green at the reviewed head
- Counterexample probes reproduced all Schema/model failures described above

qinhaihong-red:
## RPG production-model calibration review

Reviewed exact head `c78c218ed0661deb16e2ddeb1563c53fdf2b464d`.

**Verdict: Request changes.** The delivery mechanics remain well-factored and green, but the shipped numeric artifact is not yet a defensible **RPG-family model**. It is one opinionated four-stat, mana-pool, armor-mitigation build presented as a family baseline. The gap is deeper than adding a few missing attributes: the first concrete template exposes Standard Schema assumptions that prevent production modeling, simulation, and calibration.

This review calibrates the PR against CRPG/TTRPG, JRPG, ARPG, and MMO evidence. The stable commonality is not a fixed stat list. It is this causal chain:

```text
profile
  → stat/resource definitions
  → typed formula and contribution graph
  → resolution/actions/effects
  → bound actor/build/scenario instances
  → replayable runtime state
  → progression/economy
  → targets, simulation, and calibration evidence
```

Attribute names, dice/check systems, health models, armor semantics, action economy, resource systems, and progression are profile choices.

### External calibration

- D&D 2024 has six abilities, a score-to-modifier transform, roll-vs-AC resolution, separate current/max/temporary HP, multiple movement modes, and cumulative XP thresholds. Its own rules distinguish those concepts rather than representing them as one scalar layer ([SRD 5.2](https://media.dndbeyond.com/compendium-images/srd/5.2/SRD_CC_v5.2.pdf)).
- Pathfinder 2e uses typed modifier stacking, four outcome degrees, flexible initiative inputs, and a three-action economy. More importantly, its official creature-building guidance says NPC statistics are selected from level/role **final-value benchmarks**, with deliberate accuracy/damage and AC/HP trade-offs, rather than derived bottom-up like PCs ([Playing the Game](https://2e.aonprd.com/Rules.aspx?ID=2263&NoRedirect=1), [Bonuses](https://2e.aonprd.com/Rules.aspx?ID=2281), [Building Creatures](https://2e.aonprd.com/Rules.aspx?ID=2874)).
- Fate models capabilities as skills/aspects/stunts and harm as stress/consequences, demonstrating that classic primary attributes and continuous HP are not RPG invariants ([character sheet](https://fate-srd.com/fate-core/character-sheet), [resolving attacks](https://fate-srd.com/fate-core/resolving-attacks)).
- Pillars of Eternity resolves Accuracy against four distinct defenses and produces miss/graze/hit/crit outcome bands; it also uses per-rest, per-encounter, modal, and class-specific resources rather than a universal mana/cooldown model ([attack and resource design](https://eternity.obsidian.net/eternity/news/update-39-non-core-classes-cooldowns-attack-resolution-damage-vs-armor-and-a-tileset), [published game-data format](https://eternity.obsidian.net/game-data-formats/components)).
- Unreal's official RPG/GAS documentation separates base/current values and current/max resources, applies temporary effects to current state, and uses level-indexed curve tables for rebalanceable whole-game data ([Gameplay Attributes](https://dev.epicgames.com/documentation/en-us/unreal-engine/gameplay-attributes-and-attribute-sets-for-the-gameplay-ability-system-in-unreal-engine), [ARPG attributes](https://dev.epicgames.com/documentation/en-us/unreal-engine/attributes-and-effects-in-arpg?application_version=4.27)).
- Schreiber's cited material does not establish one universally correct curve. It says curve choice depends on desired length and flow, and its triangular example has **linear per-level costs** whose cumulative total is triangular; chained reward and XP curves must be analyzed together ([cost curves](https://gamebalanceconcepts.wordpress.com/2010/07/21/level-3-transitive-mechanics-and-cost-curves/)).

### Required findings

#### 1. The family template confuses a profile with the RPG-family kernel

[`rpg.md`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/docs/templates/rpg.md#L29-L54) promotes STR/DEX/VIT/INT at base 10 to the family baseline, then adds universal `mana`, `attack_power`, `armor`, global crit/CDR, and one scalar `elemental_resistance`. Frequency in a hand-selected survey does not make these concepts semantically interchangeable:

- D&D/PF need ability modifiers, proficiency/skills, several defenses/saves, action economy, and spell slots/focus in addition to or instead of mana.
- JRPGs commonly distinguish physical/magical offense and defense, turn/ATB speed, per-element affinities, status accuracy/resistance, and skill costs.
- ARPG/MMO models need attack/cast speed, hit/accuracy and evasion, resource generation/spending, damage types/conversion/penetration, equipment/affix sources, and often multiple class-specific resources.
- Some RPGs use skills/tracks and no classic primary-stat or HP/mana model at all.

The current map is syntactically extensible—another id can be added—but it is not a family model because the semantic contracts and subsystem bindings needed to make a new id meaningful do not exist.

**Fix:** keep the evaluator genre-agnostic, but split the data layer into a small versioned `rpg-core` capability contract plus CRPG/JRPG/ARPG profiles/modules. A profile binds arbitrary stat ids to semantic roles and supplies its own resolution, resource, progression, and timing policies. Do not claim a universal concrete vocabulary.

#### 2. Several independent mechanics are collapsed into single attributes

The template currently conflates:

- resource capacity and mutable current state (`health`/`mana`);
- avoidance/target defense, armor rating, and damage mitigation (`armor` → `damage_reduction`);
- action order and mobility (`initiative` is labeled mobility);
- distance rate, percentage display, and action cost (`movement_speed: number = 100`);
- every elemental/damage-type resistance into one scalar;
- cumulative XP-to-level and incremental `experience_to_next_level`;
- cooldown reduction, haste/action rate, and their different stacking laws.

This is why the shipped regeneration example can heal the formula-derived `health` value above its capacity, why Dexterity-derived Diablo-style defense is fed into an unrelated armor-reduction curve, and why adding a new damage type cannot be modeled as data. These are template defects, but most cannot be repaired honestly inside the current Schema.

The fixed `armor / (armor + 100)` relation is also not a family formula. For example, Path of Exile makes armor effectiveness depend on the **size of the incoming hit**, while Pillars compares attack accuracy to a defense and applies damage-vs-armor later; tabletop armor may instead change hit probability or subtract damage ([PoE armor](https://www.poewiki.net/wiki/Armour), [Pillars resolution](https://eternity.obsidian.net/eternity/news/update-39-non-core-classes-cooldowns-attack-resolution-damage-vs-armor-and-a-tileset)). A usable model must make the mitigation policy and its source/target/event inputs explicit.

At minimum, the model needs separate roles for `capacity`, `current`, `buffer`, `rating`, `probability`, `multiplier`, `rate`, `duration`, `counter`, and `threshold`; registries/maps for damage, defense, movement, and resource types; and subsystem references that bind those roles.

#### 3. Standard Schema's numeric type is too weak to preserve meaning

[`Domain`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/attributes.py#L25-L33) is only `number | percentage | probability`; bounds are literal floats and [`Attribute`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/attributes.py#L89-L104) has no unit or semantic kind. Consequently, HP, XP, level, meters/feet, seconds, rating, points-per-second, and dimensionless factors all validate as interchangeable numbers.

This root defect directly caused `movement_speed = 100` to validate despite the fractional-percentage law and prevents dimensional checks such as:

- `HP + seconds` must refuse;
- a regeneration magnitude must be `HP / tick` or `HP / second`;
- `armor + armor_pivot` must use the same rating dimension;
- `current_health <= max_health` needs a reference/formula-valued bound;
- level/count quantities may require integer/quantization semantics.

**Root fix in bADR-0002/0003:** separate `scalar_kind`, `dimension/unit`, `semantic_kind`, mathematical domain, design bounds, and representation/rounding. Type-check every formula node and modifier operation.

#### 4. Standard Schema has no entity binding or source/target formula context

The document currently mixes a global definition with what appears to be one actor's values. There is no declared `BoundInstance` for actor archetype, class, level, allocation, loadout, rolls, or encounter context, and no replayable runtime state object.

The formula AST is also explicitly limited to unscoped [`attr` and `param` references](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/docs/badr/0003-formula-as-data-representation.md#L30-L69). It therefore cannot express the basic combat relation:

```text
damage = f(source.attack_power, ability.coefficient,
           target.armor, target.resistance[damage_type], event)
```

without an out-of-schema code path or a second formula dialect. The future `combat` section cannot fix this merely by adding fields if the single formula authority has no typed context references.

**Root fix in bADR-0001/0003/0006:** separate immutable `DesignDefinition`, bound actor/build/scenario instances, and `RuntimeSnapshot + EventLog`; add typed formula contexts (`source`, `target`, `self`, `ability`, `scenario`, `previous_state`) with per-consumer allowed-reference rules. Keep the pure definition DAG acyclic; model recurrence as state transitions.

The same problem appears in effects. The current shape has only [`target`, `operation`, `application`, and `magnitude`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/effects.py#L118-L148), plus a small duration/stacking envelope. A production status/effect needs a source and target context, eligibility/immunity conditions, trigger/event, snapshot-vs-dynamic capture, stack group/cap, reapplication policy, dispel/removal semantics, and deterministic priority. Otherwise equipment, skills, procs, ailments, and encounter auras will each need private logic.

#### 5. The declared tuning surface cannot calibrate this template

[`parameters`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/document.py#L77-L91) is `id → float`. It carries no unit, hard/soft range, role, transform, precision, owner, rationale, or provenance. Direct bases, bounds, and polynomial coefficients are literal-only, so primary bases, crit bases/caps, and the XP curve shape are excluded from the advertised tuning surface.

No target or scenario is attached to the defaults. A config being valid and evaluatable says nothing about whether it is balanced. A production calibration contract needs:

```text
Parameter:
  value, quantity_type, hard_range, preferred_range,
  role = tunable | design_constant | scenario_input | observed,
  transform, step/precision, provenance

BalanceSuite:
  scenario matrix, metrics, target ranges/distributions,
  weights, hard constraints, seed set, sample/confidence budget
```

The minimum scenario lattice is:

```text
profile × role/archetype × level checkpoint × gear/build band
        × encounter shape × enemy relative power × player/AI skill × seed
```

Then run: semantic invariants → deterministic vectors → baseline simulation → local/global sensitivity → constrained multi-objective search → holdout fixtures → human/playtest review. Track TTK/TTD, hit/outcome distribution, effective DPS/EHP, resource starvation/overcap, action uptime, progression time, economy flows, build dominance, percentiles, and tail failures. Use fixed/common seeds when comparing candidates.

The reference calculations should be decomposed, not hidden in one "power" number:

```text
expected_DPS
  = attempts_per_second
  × P(hit)
  × expected_damage_per_hit
  × (1 + P(crit) × (crit_multiplier - 1))
  × defense_multiplier
  × uptime

net_resource_rate
  = regeneration
  + on_hit × hits_per_second
  + on_kill × kills_per_second
  - cost_per_use × uses_per_second
  - upkeep

expected_time_to_level
  = XP_to_next / XP_per_minute(profile, content, skill, party, penalties)
```

Each term must retain units, profile policy, and trace provenance. Calibrate burst and sustained windows, small-hit and heavy-hit defense, single-target and AoE, full and depleted resources, and P50/P90/P99 tails—not only means.

Until those exist, `rpg.md` must call these values **illustrative examples**, not a balanced or proven family baseline.

#### 6. Tier and template lineage are not sufficient extension mechanisms

The direction “tiers are data, not Schema law” is sound, but [`Tier`](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/src/gda_balancing/schema/model/attributes.py#L71-L86) can constrain only one exact domain, one base kind, and exact channels. It cannot express the documented tertiary contract `(percentage | probability) × (direct | formula) × required cap`, and it encodes neither semantic role nor unit.

bADR-0012's temporary fixture lineage rule—equal tier maps—does not prove that a fixture instantiated the template, preserves its definitions, or records which template version it used ([lines 62–65](https://github.com/aigengame/godot-agent/blob/c78c218ed0661deb16e2ddeb1563c53fdf2b464d/libs/gda-balancing/docs/badr/0012-genre-template-representation-and-delivery.md#L62-L65)). Two unrelated documents can satisfy it.

**Fix:** treat tiers as optional views/groups over typed definitions. Add an explicit, versioned template/profile reference plus declared overlay/override operations and a resolved-content hash. Test semantic inheritance, not tier-map equality. Adding a stat, damage type, resource type, profile, or contribution source must be data-only and must preserve an auditable provenance chain.

### Production target model

```text
DesignDefinition
├── quantity types / units / parameters / provenance
├── stat definitions + typed formula DAG
├── resource definitions (capacity/current/flows/policies)
├── contribution and stacking policies
├── checks/resolutions/actions/effects
├── items/affixes/equipment/builds
├── combat/growth/economy/encounters
├── profiles/views
└── scenarios/targets/constraints

BoundInstance
└── actor/archetype + level + allocation + loadout + scenario inputs

RuntimeSnapshot + EventLog
└── current resources/cooldowns/charges/effect instances + seed + trace
```

For example, health should resemble:

```text
stat health_max:
  kind: capacity
  unit: hp
  formula: health_base + health_per_vitality * vitality

resource health:
  capacity_ref: health_max
  initial: full
  floor: 0
  capacity_change: preserve_fraction | preserve_absolute | formula
  overflow: discard | buffer | allow
```

and an effect should carry source, target, application phase, typed magnitude, schedule/trigger, stacking group/rule/cap, snapshot-vs-dynamic capture, and deterministic ordering. Equipment/build entries produce contributions/effects; they do not rewrite stat definitions.

### Recommended delivery sequence

1. **Preserve the good parts:** canonical packaged data, no genre-keyed evaluator paths, map-key identity, closed formula AST, deterministic validation, and tiers as optional template vocabulary.
2. **Reopen the core design before first publication:** amend bADR-0001/0002/0003/0006/0012 for quantity semantics, definition/instance/state separation, typed formula contexts, resources, parameter metadata, profile lineage, and extension.
3. **Split transport from numeric design if necessary:** `meta.genre` and `template list/get` can ship independently; do not freeze `rpg.json` as the canonical family baseline while the core cannot express its invariants.
4. **Build data-only family/profile layers:** RPG core plus CRPG/JRPG/ARPG profiles/modules. The three reference fixtures should demonstrate genuinely different resolution, timing, resource, progression, defense, and build paths—not only renamed attributes.
5. **Land combat/build/growth/economy/encounter schemas around references to the core:** do not add private formula or modifier dialects.
6. **Add balance suites and acceptance gates:** scenario targets, sensitivity/Monte-Carlo results, holdouts, trace replay, and exact source/version citations. A default may be called calibrated only when its report is committed and reproducible.

### Current-head disposition

The previously reported correctness findings remain open at this unchanged head: resource capacity/current, mandatory-cap expressibility, movement units, incremental-vs-cumulative XP, defense/mobility conflation, incomplete tuning parameters, missing fixtures, and unauditable provenance. The new system-level conclusion is that these are not isolated cleanup items. They are evidence that the current Standard Schema abstraction boundary is below the product requirement.

If this PR remains a numeric-design PR, it should not merge until the core design is reopened and the template is re-authored against the resulting model. If the delivery mechanism needs to land now, split it into a delivery-only PR and keep the RPG numeric artifact experimental/unshipped.

qinhaihong-red:
Closing per maintainer direction. The RPG template work is superseded by a Standard Schema 2.0 redesign; the replacement PRD and balancing-domain decisions will be developed from current origin/main on a new branch.

Issue #531, opened 2026-07-20
[gda-balancing] One formula-site descriptor authority projecting walker, catalog scopes, and fixtures


qinhaihong-red:
## Parent

- #504 (schema core; raised as a non-blocking "Consider" in PR #527's round-4 review and recorded there).

## What to build

Every formula rule (named-form field constraints, expression-tree limits, reference integrity) applies to **every** formula consumer — bADR-0003's "no consumer gets a private formula dialect". The schema core delivers that for the two v1 consumers (attribute bases, effect magnitudes) through three hand-maintained declarations that must agree: the site iterator the semantic phase walks, the per-rule catalog scope templates, and the per-template violation fixtures.

The conformance suite enforces that those three **agree with each other** — every emitted refusal path matches a declared template, and every declared template is exercised by its own fixture. What it cannot do is notice a site that was never declared anywhere: a third formula consumer added to the iterator but omitted from the scope arrays and fixture factories leaves no failing evidence, because the corpus is derived from the same declarations.

Collapse the three into one authority: a formula-site descriptor that names a consumer once and **projects** the walker's pointer tokens, the catalog's scope templates, and the fixture factory input. Adding a consumer then means adding one descriptor entry, and omitting the projections becomes structurally impossible rather than test-detectable.

**Trigger for doing this work:** a third formula consumer (a reserved section gaining formula-valued fields — e.g. growth curves — is the likely first). With two consumers the duplication is small and the mutual-consistency tests cover it; the abstraction pays off when the third arrives and its shape is known.

## Acceptance criteria

- [ ] One declaration per formula consumer is the single source for: the site iterator's traversal, the catalog scope templates of every shared formula rule, and the violation-fixture construction for those templates.
- [ ] Adding a consumer to that declaration alone extends the semantic walk, the published catalog scopes, and the per-template behavioural evidence — no second edit site.
- [ ] The existing per-template conformance evidence still holds for every rule and every consumer (both directions: emitted paths ⊆ declared templates, declared templates ⊆ exercised).
- [ ] Published catalog output for the current consumers is unchanged by the refactor (golden diff is empty, or its diff is reviewed and intended).
- [ ] The narrowed anti-drift claim recorded in the rule registry is updated to state what the new mechanism actually guarantees.

## Blocked by

None mechanically — but deliberately deferred until a third formula consumer exists (see trigger above), so the abstraction is shaped by a real third case rather than guessed.

[...]

PR #581, opened 2026-07-23
feat(gda-balancing)!: build typed Quantity model into canonical RIR

### LINKED ISSUE #539: [gda-balancing] Build one typed Quantity model into Lock, RIR, and Resolved Model
## Parent

- #534
- #501

## What to build

Take one composable Model Source Package using typed Quantity symbols and one complete selected
Domain package release through `model check` and `model build`. Produce the selected Package Lock,
canonical RIR semantic payload, exact-authority Resolved Model wrapper, Debug Map and build/
resolution receipts as one recoverable public artifact path. Two independent lowerers must consume
the same permanent Kernel/LDB authority and mutually consume each other's result.

## Acceptance criteria

- [ ] The source distinguishes constant, parameter, input, state, derived, output, and random symbols with explicit representation, kind, unit/dimension, support/domain, and Numeric policy.
- [ ] Resolution closes the selected dependency, capability, type, conversion, and operation graph in Package Lock; resolver provenance appears only in a Resolution receipt.
- [ ] Equivalent source orderings produce byte-identical Lock and RIR payloads; source/debug/compiler provenance is excluded from RIR identity and bound separately.
- [ ] The Resolved Model wrapper binds the exact Kernel, whole LDB, selected Lock, and RIR payload; coherent but incorrect reidentification of selected operations, diagnostics, profiles, or reasons refuses admission.
- [ ] Two independent lowerers produce mutually consumable, byte-identical RIR for the selected program and agree on positive, negative, boundary, and mutation vectors.
- [ ] `model check` emits no build artifact; `model build` atomically publishes the complete typed artifact set with a framed manifest, resolvable receipt, caller-supplied Invocation key, and retry recovery.
- [ ] Direct or symlink input/output aliasing, partial publication, Invocation-key conflict, and pre/post-commit faults have the specified usage/internal/recovery behavior without mutating inputs.

## Blocked by

- #538

qinhaihong-red:
Closes #539.

## Summary

- compile one typed Quantity Model Source through the selected Package Lock into canonical RIR and an exact-authority Resolved Model
- make both production and independent lowerers consume the permanent Kernel Specification + LDB authority and mutually admit byte-identical results
- publish the complete build artifact set atomically with framed-manifest recovery, Invocation-key conflict handling, and input/output alias protection
- type runtime projections and output equalities from Kernel contracts, including empty semantic collections
- enforce Kernel-declared runtime projection accounting with typed static resource-exhaustion refusal

## Validation

- `599 passed, 37 skipped`
- Ruff check passed
- Ruff format check passed
- Pyright: `0 errors, 0 warnings, 0 informations`
- fixed-SHA independent review at `546b9d19dc737b0266696918aae4c22f606a9e1d`: PASS, 0 actionable findings

## Evidence boundary

This PR validates the single typed-Quantity vertical loop required by #539. It does not, by itself, claim complete Schema 2.0 or cross-genre completeness. New Kernel transform or accounting-event vocabulary remains a Schema-major review concern.


qinhaihong-red (commented):
## High-effort multi-agent review

Multi-angle finder pass + one independent adversarial verifier per finding location, against `origin/main...origin/feat/539-typed-quantity-model-build` (18 changed files). 12 candidates → 9 survived verification → 8 findings after merging duplicates (3 refuted, listed at the end). Ranked most-severe first.

---

### Correctness

#### 1. [CONFIRMED] `model build` reports a missing/malformed `GDA_BALANCING_ANCHOR_KEY` through the exit-4 `internal` envelope instead of an exit-3 usage error

`libs/gda-balancing/src/gda_balancing/schema2/model.py:2193`

Running `gda-balancing model build model.json --out out.json --invocation-key <64hex>` without exporting `GDA_BALANCING_ANCHOR_KEY` (or with a value that is not exactly 64 lowercase hex digits) runs `check_model_source` plus the full lowering and commits the invocation directory — then publish reaches `_write_anchor_exclusive` → `_anchor_authentication_key()`, which raises a bare `RuntimeError`. Nothing in the chain converts it, so the dispatch tail emits the exit-4 `internal` envelope ("the toolkit failed unexpectedly (RuntimeError)"), the channel bADR-0008 reserves for unexpected toolkit faults, with the actionable message hidden behind `--debug`. The env var is referenced only inside `model.py` and set only by the test autouse fixture, so a real CLI user hits this immediately, on every invocation, paying a full lowering + rollback each time. An environment/config precondition should surface as a usage error (exit 3). Same root cause at `model.py:2194`.

#### 2. [PLAUSIBLE] The `not owners` guard was dropped from the reference-contract closure loop

`libs/gda-balancing/src/gda_balancing/schema2/bootstrap.py:945`

`_reference_contracts_close` no longer refuses a reference contract whose declared `owners` path resolves to an empty list — the `for owner in owners` loop body never runs, so closure passes vacuously (`_path_is_declared` returns True for a declared-but-empty terminal list because `all([])` is True). An authoring regression in `kernel.json` / `language-bundle.json` that empties an owner collection while its path stays declared now boots the toolkit with that cross-authority reference constraint silently unenforced, where the old code refused at ingress. The same relaxation was mirrored into the independent consumer-B check in the conformance test, so mutual-admission does not catch it. If this relaxation is intentional, it should be recorded (bADR); otherwise restore the guard.

---

### Cleanup

#### 3. [CONFIRMED] `_bind` binding-law docstring contradicts the new `--out` behavior

`libs/gda-balancing/src/gda_balancing/dispatch.py:255`

The docstring still says "`--out` is accepted only for an `artifact_sink` command; anywhere else it is an unknown argument", but this PR makes `--out` a normal bound field for MODEL_BUILD (artifact_set) via `option_bindings` (the condition at L273 is now `if name == _OUT_FLAG and descriptor.artifact_sink:`). A maintainer trusting the docstring will misjudge how build's `--out` is routed and may "fix" that condition, breaking `model build --out`.

#### 4. [CONFIRMED] Artifact-membership expression duplicated between `_descriptor_body` and `surface_manifest`

`libs/gda-balancing/src/gda_balancing/schema2/surface.py:413`

The `"atomic-artifact-set" if descriptor.artifact_sink or descriptor.artifact_set else "stdout-only"` ternary and the member→dict comprehension are copy-pasted from `_descriptor_body` (L228–238). A future change to the artifact_set row shape edited in only one site makes the per-command `--schema` descriptor identity and the surface-manifest row disagree. Suggest extracting a shared `_artifact_membership(descriptor)` helper.

#### 5. [CONFIRMED] `_recover_publication` calls `load_authorities()` twice for the same value

`libs/gda-balancing/src/gda_balancing/schema2/model.py:2499` (and L2522)

Both calls fetch the same kernel `content_identity`; each re-reads and re-parses both authority files from disk (including the 256KB size guard and the byte-level nesting-depth scan), discarding the language bundle. Hoist it into a single local.

#### 6. [PLAUSIBLE] The "--out must not resolve to the input path" rule is implemented a second time inline

`libs/gda-balancing/src/gda_balancing/schema2/model.py:2572`

`_publish_model_artifacts_locked` re-does a bare `os.path.realpath(source) == os.path.realpath(out)` compare with the identical error message as `dispatch._reject_input_aliasing` (L377) but different guarding (dispatch exempts a non-existent positional). The same user-facing rule now enforces differently across the two artifact modes, and a fix to one guard silently leaves the other divergent. Related: #531 (formula-site descriptor authority) was deferred pending a second consumer — one now exists.

#### 7. [PLAUSIBLE] Capability-manifest payload shape duplicated between `lower_checked_model` and `_expected_capability_manifest`

`libs/gda-balancing/src/gda_balancing/schema2/model.py:2118` (vs L1992–2016)

The two sites build the identical 11-key payload independently. A future field added or renamed in one site only makes the re-derivation in `_recover_publication` (L2485) mismatch the committed artifact and raise "committed Capability manifest is not an exact projection" (exit 4) — i.e. crash-recovery of a perfectly valid published build fails. Suggest one shared payload-builder helper.

#### 8. [PLAUSIBLE] Refusal catalog rebuilt inline in three functions

`libs/gda-balancing/src/gda_balancing/schema2/model.py:142` (also L54, L164)

`_model_refusal_catalog`, `_refusal`, and `_bounded_refusal` each independently compute `BOOTSTRAP_REFUSAL_CATALOG + tuple((item["code"], item["stage"]) for item in language_bundle["diagnostics"])`. If the catalog's source set changes, a site left un-updated produces a catalog missing codes, so `_refusal`/`_bounded_refusal` raise `KeyError` on `stages[code]` (→ internal/exit 4) instead of emitting the refusal. Suggest one `_refusal_catalog(language_bundle)` helper.

---

<details>
<summary>Refuted candidates (verified not to be issues)</summary>

- `model.py:883` — "only `_RuntimeProjectionResourceExhausted` is converted to a refusal; other projection errors escape as exit-4": the predicted trigger is unreachable — the model-source wire schema pins the same symbol fields the runtime projection seeds with `const`, so a structurally-valid, resolution-passing source cannot produce the other error paths.
- `model.py:2164` — sha256-prefix check triplicated across `_store_*_path` helpers: factually true, but pure style with no observable consequence.
- `dispatch.py:57` — "`_UsageError` docstring says handlers never see usage errors, contradicting model.py": the quoted docstring text does not exist in the code.

</details>

Only finding 1 is user-visible today; finding 2 needs an intentionality call; the rest are maintainability debt that can land here or be split out to issues.

qinhaihong-red:
Addressed the review at fixed head `ab132d5baa824602313dfca866ea933b9b79970b`.

1. **Anchor configuration — fixed.** `model build` now validates and captures `GDA_BALANCING_ANCHOR_KEY` at the handler invocation boundary, before source reading, resolution/lowering preflight, lock/store creation, or publication. Missing or malformed values produce the declared `invalid_argument` usage envelope / exit 3. The same immutable key bytes are passed through publish, recovery, anchor write, and anchor verification without rereading the environment.
2. **Empty reference-contract owners — intentionally retained.** The contract universally quantifies the owner rows that exist; a declared empty owner collection is closed by construction. This is required for optional language collections such as empty Conversions. The Kernel contract does not impose `minItems`, and the existing two-consumer `test_two_consumers_type_empty_semantic_collections_from_kernel_contracts` proves the intended admission behavior. I added an explanatory implementation comment; no bADR change is needed because this is existing machine-authority semantics, not a new decision.
3. **`_bind` documentation — fixed.** The docstring now distinguishes dispatch-owned legacy `artifact_sink --out` from descriptor-bound `artifact_set --out`.
4. **Artifact membership projection — consolidated.** Descriptor identity and Surface manifest now share `_artifact_membership`.
5. **Recovery authority reload — removed.** Recovery consumes the already checked exact Kernel and uses its identity for both provenance checks; it no longer reloads the authority twice.
6. **Input/output aliasing — consolidated without weakening either mode.** Both consumers share `reject_input_aliasing`, with an explicit policy: legacy generic positionals remain existing-path-only, while Model artifact publication treats the admitted source as a known path even if its locator disappears before publication. Alias validation occurs before lock/store creation.
7. **Capability manifest — consolidated.** Lowering and recovery now share `_capability_manifest`.
8. **Refusal catalog — consolidated.** Descriptor initialization, single refusal construction, and bounded refusal construction now share `_refusal_catalog`.

Validation:

- `605 passed, 37 skipped`
- Ruff check passed
- Ruff format check passed
- Pyright: `0 errors, 0 warnings, 0 informations`
- Independent fixed-SHA re-review: PASS, 0 actionable findings; 12 focused checks passed

Commits: `61e4196`, `ab132d5`.

qinhaihong-red (commented):
## Re-review of the fix commits — PASS

Independent recheck of `61e4196` + `ab132d5` (fixed head `ab132d5baa824602313dfca866ea933b9b79970b`): two adversarial verifiers over the code, plus all four validation gates re-run in a clean detached worktree at that SHA. **All 8 findings are resolved or justified; no new defect found; review closed from my side.**

| # | Finding | Verdict |
|---|---|---|
| 1 | Anchor key → exit-4 internal envelope | **FIXED** |
| 2 | Dropped `not owners` closure guard | **RETENTION JUSTIFIED** (one framing caveat below) |
| 3 | Stale `_bind` docstring | **RESOLVED** |
| 4 | Artifact-membership duplication | **RESOLVED** (`_artifact_membership`, both sites) |
| 5 | Double `load_authorities()` in recovery | **RESOLVED** (freshness semantics preserved — see below) |
| 6 | Input-aliasing rule duplicated | **RESOLVED** (`path_contracts.reject_input_aliasing`, both modes exact) |
| 7 | Capability-manifest payload duplicated | **RESOLVED** (`_capability_manifest`, no identity drift) |
| 8 | Refusal catalog rebuilt inline ×3 | **RESOLVED** (`_refusal_catalog`, no inline rebuild remains) |

### Verification highlights

- **Finding 1, end to end:** `publication_authentication_key()` is the first statement of the build handler (`commands/model.py:110`), before `check_model_source` and any lock/store creation; it raises `UsageError("invalid_argument", …)`, which is declared in `MODEL_BUILD.usage_codes` — so the schema-2 dispatch gate (`dispatch.py:177-182`) passes it through to the usage envelope / exit 3 rather than flipping it to exit 4 (the one place this fix could have silently failed; verified). No production-path env re-read remains — the key bytes are threaded as a parameter through publish, recovery, anchor write, and verification. The new tests pin both properties: `not store.exists()` on failure proves ordering vs store/lock creation, and the non-existent-source case asserting `invalid_argument` (not `unreadable_input`) proves the key check precedes source reading.
- **Finding 5:** recovery now consumes `checked.kernel`, which is loaded fresh by the same invocation's `check_model_source`; a cross-invocation authority swap is still detected (check-time reload + `admit_resolved_model`'s independent fresh load). No detection-semantics regression — the change only removes redundant same-invocation reloads.
- **Finding 6:** legacy `artifact_sink` behavior is byte-identical (default `input_is_known_path=False` reproduces the old existence exemption); `input_is_known_path=True` exactly reproduces the model path's old unconditional compare; the check now runs before lock/store creation; `path_contracts.py` is a clean leaf module (no import cycle); `envelope.UsageError` vs the old `_UsageError` marker lands in the same `except UsageError` handlers with the same code.
- **Finding 7:** the unified payload is byte-identical to both pre-fix sites (the only textual difference was runtime-noop `cast()` wrappers), so no content-addressed identity shift for published artifacts.
- **Gates independently reproduced at `ab132d5`:** pytest **605 passed, 37 skipped, 0 failed**; ruff check clean; ruff format clean (303 files); pyright **0 errors, 0 warnings, 0 informations**. All four match the reported numbers.

### One non-blocking note on finding 2

The retention argument is sound: reference closure is a referential-integrity law, an empty owner collection is vacuously closed, and the one collection whose emptiness is load-bearing (`model_lowerings`) is still independently guarded (`bootstrap.py:739`), so a functionally broken language is still refused. The kernel contract indeed imposes no `minItems` on the 11 owner collections, and `test_two_consumers_type_empty_semantic_collections_from_kernel_contracts` pins the admission behavior through both consumers.

However, "existing machine-authority semantics, not a new decision" understates it: at the PR base, an empty declared collection was *refused* at ingress; this PR *admits* it. That is a (well-reasoned, test-pinned) behavioral change to the admission surface, uniform across all 11 reference contracts rather than scoped to the genuinely optional ones. Not blocking — but if admission-surface changes are the kind of thing bADRs record, a one-paragraph note would keep the decision trail honest.

PR #584, opened 2026-07-25
feat(gda-balancing): migrate semantics-preserving 1.x source

### LINKED ISSUE #554: [gda-balancing] Migrate semantics-preserving 1.x source and refuse deprecated constructs
## Parent

- #534
- #501

## What to build

Implement the limited clean-forward migration path through `model migrate`. Convert only 1.x
source constructs with a deterministic semantics-preserving mapping into 2.x Model Source; bind
every mapping/default/warning/refusal in an audit report and refuse unsupported constructs without
partial output. Do not add a 1.x runtime, save/replay/ruleset adapter, reverse migration, or gray
rollout.

## Acceptance criteria

- [ ] The converter consumes exact 1.x input plus exact converter/Kernel/LDB identities and produces a new 2.x Model Source identity only when every required mapping is semantics-preserving.
- [ ] The migration report binds input/output identities, mappings, explicit defaults, warnings, deprecated/unsupported constructs, and typed refusals.
- [ ] Unsupported or lossy constructs refuse the conversion with no valid-looking partial 2.x model artifact.
- [ ] Positive, negative, limit, identity, and repeatability vectors prove deterministic conversion and report behavior.
- [ ] The converted source builds through #539 and exposes no compatibility runtime, reverse path, or hidden 1.x semantic authority.
- [ ] Schema version and toolkit package version remain independent throughout migration.

## Blocked by

- #539

qinhaihong-red:
## Summary

- add the sole clean-forward Schema 1.x entry point, `gda-balancing model migrate`
- convert only the admitted semantics-preserving subset into an exact Schema 2.0 Model Source Package
- publish content-addressed migration success/refusal reports bound to the exact input, converter specification, Kernel, LDB, and output identities
- refuse unsupported, lossy, unreadable, non-regular, or oversized inputs without publishing a partial Model Source artifact
- remove the remaining production Schema 1.x design/formula surfaces instead of adding a compatibility runtime or reverse path

## Authority and reliability

- make the converter specification an LDB-owned artifact contract and wire schema
- bind both migration report schemas to the same complete converter artifact
- re-admit embedded artifact bindings in bootstrap, including wire-schema identity, content identity, and cross-report canonical equality
- bound source observation to regular files up to 16 MiB while hashing the complete admitted byte sequence
- keep Schema version and toolkit package version independent

## Acceptance evidence

- positive conversion, unsupported/lossy refusal, boundary, identity, repeatability, and no-partial-publication vectors
- converted output passes the existing Schema 2.0 model build pipeline
- independent bootstrap consumers reject unilateral converter identity, specification, and artifact-schema drift
- fixed-SHA subagent review of `15303c8`: PASS

## Validation

- `695 passed, 54 skipped`
- `ruff check .`
- `ruff format --check .`
- `pyright --project libs/gda-balancing`
- sdist and wheel build successfully

Closes #554


qinhaihong-red (commented):
## Code review — head `15303c8`

Multi-angle review (8 finder angles, every candidate adversarially verified; 2 candidates refuted, 1 dropped as defensible style). Overall verdict: **high quality** — the 1.x surface removal migrates cleanly (removed-behavior and cross-file audits found nothing left dangling; 695 tests pass) and no convention violations were found. Seven findings below, most severe first. **#1 and #2 are worth addressing before merge**; #3 has measured impact; #4–#7 are low-cost cleanups.

---

### 1. Reachable refusal misreported as `internal_error` (exit 4) — CONFIRMED, reproduced
`src/gda_balancing/commands/model.py:253`

Migration enforces `max_symbols` but never the language bundle's `max_source_bytes` (1 MiB). A valid 1.x source (~640 KB, 3000 long-named parameters — passes the 1.x funnel) expands to ~1.32 MB during conversion: `migrate_design_source` returns `MigrationSuccess`, then `check_model_source_value` refuses with `language.source_too_large`, and line 253 (`raise RuntimeError("migrated Model Source failed exact-authority admission")`) turns that input-caused refusal into `{"error": {"category": "internal", "code": "internal_error", ...}}` with exit 4 instead of a structured refusal envelope with exit 2 — the `language.source_too_large` diagnostic and the migration report are lost. Reproduced by execution against this head.

**Fix:** pre-check `max_source_bytes` in migration (emitting `migration.reason.target-limit-exceeded`, symmetric with the `max_symbols` handling), or return the refusal (widening `MIGRATION_REFUSAL_CATALOG` to the language stage).

### 2. `_bounded_refusal` duplicated with the opposite collision policy — CONFIRMED
`src/gda_balancing/schema2/migration.py:606`

migration.py's private `_bounded_refusal` near-duplicates `schema2/model.py:175-207` with a silently different collision policy: last-wins dict comprehension vs model.py's first-wins `setdefault`. Reachable via `_source_refusal` (migration.py:593-601): every 1.x funnel refusal maps to the same `migration.reason.source-invalid` code keyed by path, so two refusals at the same path with different messages collide — migration keeps the last, while the identical situation elsewhere in the 2.x surface keeps the first. The two copies will also drift on future ordering/truncation fixes.

**Fix:** align the collision policy and factor the shared dedup/sort/truncate ordering (the stage-catalog and truncated-flag differences can stay parametric).

### 3. Authority admission derived three times per invocation — CONFIRMED, measured
`src/gda_balancing/commands/model.py:198`

One `model migrate` run re-derives the identical Kernel/LDB admission three times: the direct `admit_authorities` call, `check_model_source_value`'s unconditional internal re-admission (`schema2/model.py:811-816`), and `member_is_admitted` during publish — plus a second `load_authorities()` disk read via the refusal-detail schema callable on every refusal. Measured: `admit_authorities` ≈ 122 ms per call, so ~370 ms of a 1.14 s CLI invocation (~1/3 of wall clock) is duplicate admission. No bADR mandates per-callsite re-admission of the authority pair within one process (independent admission at ingress and member re-verification are both preserved by the fix).

**Fix:** thread the already-computed `Admission` through (or memoize `admit_authorities`/`load_authorities` keyed on content identity). `_artifact_schema`/`_artifact_contract` validator compilation is likewise uncached.

### 4. Warning reports wired by positional index; `[1]` path untested — CONFIRMED
`src/gda_balancing/schema2/migration.py:262`

Converter warning reports are wired positionally (`_CONVERTER_WARNING_RULES[0]["report"]` at line 262, `[1]["report"]` at line 271) while the adjacent mapping rules are looked up by stable id. Inserting or reordering a warning rule silently swaps the `metadata.omitted` and `schema-reference.replaced` payloads; both are `dict[str, JsonValue]` so nothing fails statically, and no test asserts the `schema-reference.replaced` ($schema-present) warning, so a miswiring would ship undetected.

**Fix:** give warning rules an `id` and reuse the `_mapping_report_text` lookup pattern (migration.py:500-505); add a test that triggers the `$schema` warning.

### 5. `_reason()` re-implements model.py's `_reason_by_id` — CONFIRMED
`src/gda_balancing/schema2/migration.py:554`

The unique-reason-by-id lookup contract now has two independently-maintained implementations (they already differ in `.get()` vs `[]` access and error text). No circular import blocks reuse — nothing imports `schema2.migration` from model.py's dependency tree.

**Fix:** make `_reason` a thin wrapper over model.py's `_reason_by_id`, projecting diagnostic/stage from the returned dict.

### 6. `source_bytes_identity()` has zero production callers — CONFIRMED
`src/gda_balancing/schema2/migration.py:178`

Production always uses the separate streaming digest in `load_design_source_observation` (line 222); the helper exists in `src/` solely as the test oracle for `tests/test_schema2_migration_cli.py`. Two implementations of the domain-separated source-identity hash live in the shipped package, one dead — a future change to the identity scheme edited in only one silently invalidates the other. The test's independent recomputation is a legitimate oracle, but the helper belongs in the test module.

### 7. Migration refusal detail hardcoded across three files — note for the second consumer
`src/gda_balancing/dispatch.py:222`

Shared dispatch builds `observed_details` by hardcoding `{"migration_report": outcome.migration_report}` instead of the descriptor-driven `getattr(outcome, spec.field_name)` — completing a special-case that also spans `RefusalDetailSpec`'s `Literal["migration"]`/`Literal["migration_report"]` lock (descriptors.py:99-111) and the ambient `migration_report` field on the shared `Schema2RefusalReport` (diagnostics.py:38-44). Code comments show the closure is deliberate ("the only Schema 2.x refusal detail"), so this is low priority — but the dispatch site could already read the field name from the spec today at no cost, keeping the closure enforced solely by the descriptor Literals. The next stage-specific refusal detail otherwise requires editing shared infrastructure in three files.

---

Refuted during verification (for the record): a wire-schema lookup duplication candidate and an `O_NOFOLLOW` inconsistency candidate (the documented "regular file" contract is satisfied; the readers have genuinely different subjects).

qinhaihong-red:
Addressed all seven findings in `0cc6190`.

1. Oversized migrated targets are now canonicalized and checked against the admitted LDB `max_source_bytes` before publication. Overflow returns the typed `migration.target_limit_exceeded` refusal, with no partial output.
2. Migration and model diagnostics now share one first-wins `bound_diagnostics` implementation, so same-location collisions preserve the first source refusal consistently.
3. Migration loads its immutable packaged authority pair once per process and threads the matching `BootstrapAdmission` through repeated model admission, eliminating redundant authority reads/admission derivations.
4. Converter warnings are selected by stable report code rather than positional indices. The converter contract test now explicitly exercises and asserts both the metadata and `$schema` warning paths.
5. Exact LDB reason lookup is centralized in the shared diagnostics module.
6. The test-only source-byte identity oracle was moved out of the shipped package.
7. Dispatch now derives refusal detail values from the descriptor-declared field names instead of hard-coding `migration_report`.

The authority-bound converter and LDB identities, dependent template release identity, architecture/spec text, and regression tests were refreshed together.

Validation:

- `uv run pytest -q`: 698 passed, 54 skipped
- Schema 2 focused suite: 234 passed
- CLI/envelope conformance: 112 passed, 54 skipped
- Pyright: 0 errors
- Ruff check/format: passed
- sdist and wheel build: passed

PR #611, opened 2026-08-03
refactor(gda-balancing): route scheduler conformance through Runtime seam

### LINKED ISSUE #610: [gda-balancing] Route scheduler conformance vectors through the real Runtime seam
## Parent

- #594
- Follow-up from the non-blocking review observations on PR #609

## What to build

Give scheduler conformance execution one explicit home instead of keeping a packaged helper that is called only by tests. The production scheduler-vector consumer must exercise the same Runtime scheduling seam used by normal Experiment execution, while the independent reference consumer remains a genuinely separate implementation.

Mutation injection belongs to the conformance harness rather than the normal Runtime API. The harness derives the closed detector inventory from the exact admitted Kernel, binds every declared detector to an implementation in each consumer, and fails closed when either consumer has missing or unexpected support. This removes test-owned inventory duplication without treating host implementation code as semantic authority.

This is an AFK implementation follow-up and does not block PR #609.

## Acceptance criteria

- [ ] Normal Experiment execution and the production scheduler-vector consumer exercise the same scheduling seam; no helper shipped in the runtime package is referenced only by tests.
- [ ] The normal Runtime execution API contains no mutation parameter or test-only mutation branches.
- [ ] The conformance harness derives the allowed detector inventory from the exact admitted Kernel and verifies exhaustive, no-extra detector implementations independently for the production and reference consumers.
- [ ] Unknown, missing, or unsupported detector implementations fail closed with a deterministic test or conformance failure.
- [ ] Production and reference consumers remain independently implemented and each compares its observations with the shipped vector `expect`; agreement between the consumers alone is not accepted as evidence.
- [ ] The five semantic-mutation vectors remain pairwise orthogonal. Positive and refusal vectors may intentionally detect broader semantic breakage and are not misclassified as dedicated mutation detectors.
- [ ] A public `experiment run` regression proves that the production seam used by conformance tests is also exercised by the user-visible Runtime path.
- [ ] Public CLI behavior, artifact bytes, and exact authority identities remain unchanged unless an intentional contract change is required; any such change follows the normal identity-rotation and sealed-authority workflow.
- [ ] Relevant focused tests, the complete inventory gate, Ruff, Pyright, sealed-LDB verification, and the full gda-balancing suite pass.

## Blocked by

None - can start immediately

qinhaihong-red:
## Summary

- introduce a Kernel-bound `RuntimeScheduler` seam and route normal Experiment ordering, scheduling, cancellation, artifact validation, and recovery through it
- move scheduler-vector mutation injection and evaluation out of the shipped runtime into the conformance harness
- derive each consumer's closed detector bindings from the admitted Kernel and test missing, unexpected, and unknown cases fail closed
- compare both independently implemented consumers against every shipped vector expectation and enforce the five semantic detectors as a pairwise-orthogonal matrix
- retain public `experiment run` regression coverage for the shared scheduling seam

## TDD evidence

- RED: missing Runtime scheduler seam; missing detector-inventory harness; missing/no-extra/unknown detector cases did not fail closed
- GREEN: added the shared production seam, test-owned mutation adapters, exhaustive Kernel-derived binding checks, and the 5 x 8 vector matrix
- REFACTOR: removed the packaged `_evaluate_scheduler_vector` helper and all production mutation branches

## Validation

- `uv run pytest -q`: 1318 passed, 27 skipped
- focused scheduler/conformance/public CLI regressions: 10 passed
- Ruff format/check: passed
- Pyright: 0 errors
- logical inventory: 1345 tests / 170 vectors; no missing, overlap, uncovered, or unexpected entries
- sealed LDB rebuild check: passed
- independent `$review` Standards axis: no findings
- independent `$review` Spec axis: no findings; repeated full suite passed

Closes #610


qinhaihong-red:
## Two-axis review (Standards / Spec, vs merge-base `d066ffe`)

Overall: the PR substantially and faithfully implements #610 — the seam is genuinely shared (proven by a mutation probe, not just by reading), the runtime API carries no mutation branches, and the closed detector inventory is derived from the admitted Kernel. The actionable items below are concentrated in test-side naming and coverage details. The two axes were reviewed by independent agents and are reported separately.

## Standards axis

### Documented-standard violations

**1. HARD — test-module naming/ownership inverted.** `docs/agents/testing.md` ("Test ownership and Consumer B") fixes the convention: the `*_conformance_support.py` module is the *independent* consumer that imports no production module, while Consumer A lives in `*_production_support.py`. Both precedents obey it (`schema2_bootstrap_production_support.py`, `schema2_formula_conformance_support.py`).

The new `tests/schema2_scheduler_conformance_support.py` is named for the conformance consumer but *is* the production consumer: it holds `evaluate_runtime_scheduler_vector` ("Execute a scheduler vector through the production Runtime seam") and imports `gda_balancing.schema2.runtime_scheduler`. → Rename to `schema2_scheduler_production_support.py`. Its docstring — "Scheduler conformance harness shared by the test-side consumers" — is also untrue: only `require_complete_scheduler_detector_bindings` is shared; the reference consumer stays in the test file.

Explicitly checked and **not** violated: RULES.md DRY (the three-way detector duplication is guarded by `require_complete_scheduler_detector_bindings` against Kernel authority — the documented consistency-test pattern), bADR-0022 consumer independence, truthful conventional-commit types, `BALANCING-CONTEXT.md` / `ARCHITECTURE.md`.

### Baseline smells (judgement calls, not blockers)

- **Middle Man ×2** — `_kernel_scheduler_contract` and `_runtime_ordering_tuple` are now pure one-line delegations to `RuntimeScheduler`.
- **Incomplete seam** — 14 raw scheduler subscripts remain in `experiment.py` beside the new seam methods (e.g. `scheduler["schedule"]["child_phase"]` immediately precedes `runtime_scheduler.schedule_position_signal(...)`).
- **Lying cast** — `contract` is declared `Mapping[str, Any]` but `cast(dict[str, Any], …)` asserts a subtype the field never promises.
- **Repeated construction** — `RuntimeScheduler(scheduler)` is rebuilt per element inside two list comprehensions although a hoisted instance exists in `evaluate_experiment`.
- **Primitive Obsession / Repeated Switches** — `_RuntimeSchedulerMutation` uses magic strings (`ordering="kernel"`, `state_scope="scenario"`, …) retested at five sites; the sibling `_ReferenceSchedulerMutation` uses booleans, which read better.
- **Speculative Generality (mild)** — `_Event = TypeVar(...)` serves one method with three call sites, all concretely `list[dict[str, Any]]`.

## Spec axis (vs #610 acceptance criteria)

### Verified met (on the PR head, isolated worktree)

- **AC2** — `mutation` occurs 0 times in `src/gda_balancing/schema2/experiment.py` on the PR head (12 on main).
- **AC1/AC7 (scheduler side)** — mutation probe: flipping the sign in `RuntimeScheduler.ordering_key` fails both the conformance matrix test and the public-CLI regression `test_public_experiment_orders_same_time_root_events_and_commits_between_them` (a real `experiment run` invocation). The seam is genuinely shared.
- **AC3** — `kernel.json` `…kinds[scheduler-scenario].mutation_detectors` holds exactly the 5 names; both consumers call `require_complete_scheduler_detector_bindings` before dispatch (production even when `mutation=None`).
- **AC6** — orthogonality assertion correctly restricted to `category == "semantic-mutation"` → 5×5 identity; matrix widened 5×5 → 5×8.
- **AC8** — no authority/LDB/CLI files in the diff (4 files only).
- **AC9** — full suite 1318 passed / 27 skipped; Ruff and Pyright clean (reproduced locally).

### Partial

1. **AC1 partial** — "no helper shipped in the runtime package is referenced only by tests": `_evaluate_value_program_vector` (`src/gda_balancing/schema2/experiment.py:2719`) has exactly one `src/` occurrence (its own `def`) and is called only from tests (`test_schema2_experiment_cli.py:1080, 5841, 5875`). The exact anti-pattern this issue removes for scheduler vectors survives for the value-program kind. *Aggregator note:* this helper predates the PR and #610's "What to build" scopes to scheduler conformance — suggest a follow-up issue rather than blocking here, but the AC as written is not fully met.
2. **AC5 partial** — "agreement between the consumers alone is not accepted as evidence": in the mutated matrix, for the 3 non-`semantic-mutation` candidates × 5 mutations = 15 cells, the sole assertion is `assert production_detected is reference_detected` — no shipped-`expect` expectation is bound, which is precisely the excluded evidence shape. Suggest binding an expectation for those cells or documenting the intentional exemption in the test.

### Scope creep

None material. The three `test_runtime_scheduler_seam_*` unit tests are not requested by any criterion but are proportionate coverage of the new seam.

### Implemented but questionable

- `test_runtime_scheduler_seam_orders_events_from_the_kernel_contract` asserts `ordered_events(...) == vector["expect"]["event_order"]`, which holds only because the `omitted-key` vector happens to be single-scenario with 2 pending events; the assertion silently breaks if the vector gains a scenario or a canceled event (consumer semantics additionally prefix `scenario_order` and filter `canceled`/`completed`).
- `scheduler_detector_inventory`'s `"inventory is not closed"` branch is untested (outside AC4's listed cases).

---

**Summary:** Standards — 7 findings, worst: the `*_conformance_support.py` naming/ownership inversion (hard violation; rename + fix the docstring). Spec — 4 findings, worst: AC1 partial via the pre-existing `_evaluate_value_program_vector` (follow-up-issue material); AC5's 15 agreement-only cells are worth addressing in this PR.

qinhaihong-red:
Addressed the review findings in `d828962`.

### Changes

- Renamed the Runtime-backed adapter to `schema2_scheduler_production_support.py` and corrected its ownership docstring, matching `docs/agents/testing.md`.
- Replaced agreement-only mutation assertions with an explicit oracle for every 5 x 8 cell. Production and reference results are now asserted independently against the shipped vector `expect`; the semantic 5 x 5 matrix remains pairwise orthogonal, while the positive/refusal overlaps are declared explicitly.
- Replaced the vector-shape-dependent ordering seam test with self-contained events.
- Added closed-inventory coverage for non-list, empty, unsorted, duplicate, empty-name, and non-string detector declarations. This exposed and fixed a non-string path that raised `TypeError` before the deterministic `ValueError`.
- Removed the two pure forwarding helpers, hoisted repeated `RuntimeScheduler` construction, kept the scheduler contract typed as the real immutable `Mapping` boundary, and replaced production mutation magic-string switches with named booleans.

### Evaluated but intentionally unchanged

- `_evaluate_value_program_vector` predates this PR and belongs to the value-program conformance kind; #610's implementation scope is scheduler conformance, so moving it here would expand the PR beyond the reviewed requirement.
- Raw scheduler metadata reads remain where they own identity, projection, or admission data rather than scheduling behavior. `_Event` remains because it preserves each caller's concrete mapping type through sorting.

### Verification

- Full suite: `1324 passed, 27 skipped`
- Logical inventory: 1,351 tests / 170 vectors; no missing, overlap, uncovered, or unexpected entries
- Ruff, Pyright, sealed-LDB verification, and `git diff --check`: passed
- Follow-up subagent review: Standards 0 findings; Spec/closure 0 findings
- All required CI on exact head `d828962` is green; PR is mergeable.

PR #613, opened 2026-08-03
feat(gda-balancing): dogfood reciprocal combat events

qinhaihong-red:
## Summary

- dogfood the generic multi-Event Runtime with two reversed directional combat entrypoints at the same logical time
- expose deterministic root admission, priority/order vectors, latest-Snapshot visibility, explicit pending cancellation, reciprocal Formula feedback, and both-actor Metrics through the public example
- close review gaps by scoping lifecycle Formula sites to selected entrypoints and adding controlled byte-stability, alternative Metric, backward-scheduling, and ambiguous-root-order proofs

## Review closure

- Standards re-review: no findings
- Spec re-review: no findings
- the public backward-scheduling vector also closes a CLI surface gap by declaring runtime.schedule_backward in the Experiment run refusal catalog

## Validation

- fast: 372 passed
- authority: 173 passed
- language: 272 passed
- composition: 500 passed, 27 allowed skips
- smoke/e2e: 15 passed
- outcome gates: no unexpected skips or xfails
- inventory closure: 1,359 tests; no missing, overlapping, uncovered, or unexpected shard tests
- Pyright: 0 errors, 0 warnings
- Ruff check/format: clean
- LDB rebuild check: clean

## Human acceptance gate

Issue #595 intentionally remains open. The product/architecture owner must run the public reciprocal-combat tutorial and record accept, accept with explicit conditions, or reopen before downstream work treats the dogfooding slice as complete.

Refs #595

Issue #597, opened 2026-07-29
[gda-balancing] Reduce full-suite latency without coverage loss


qinhaihong-red:
## Refactor supersession — 2026-09-06

#865 does not undo the verified performance results recorded here. The exact 1,409-test inventory, old branch/base requirements, and unchanged-contract constraints below belong to that historical optimization experiment. They do not require retaining tests or CI inventory entries whose sole subject is an explicitly retired version, converter, sink, or identity mechanism. Replacement/current behavior, refusal, resource and packaging coverage must remain real and complete.

Keep this issue’s remaining performance follow-up: attribute the serial regression and measure the exact-source release path; use the existing CI mechanisms, and introduce no new coverage protocol or timeout increase as a substitute for a measured improvement.

This dated update changes only the identified contract or scheduling clauses. The original requirements, issue identifiers, historical evidence, and still-valid acceptance scope remain below. It does not claim implementation or conformance completion.

---

## Parent

- #501
- #534

## Goal

Reduce the wall-clock and cumulative cost of the complete `gda-balancing` test suite without changing the existing test collection or introducing a second test-governance system.

The previous implementation attempt is superseded. Its behavior-claim ledger, accepted-claim manifest, migration contracts, custom local shard runner, aggregate JUnit protocol, per-shard wall-timer protocol, and expanded release evidence pipeline added substantially more complexity than the latency problem justified. The replacement starts from the clean #596 head before those commits.

## Accepted baseline

- exact commit: `2b81e2e7428656137cfc26c884681e8e03666b72`
- collection: 1,409 tests
- outcomes: 1,382 passed / 27 skipped
- local three-run cumulative median: 1,292.539 seconds
- local three-run wall median: 1,297.274 seconds
- required-CI samples: 483 / 496 / 483 seconds; median 483 seconds

The dominant files are authority CLI, experiment CLI, real-subprocess E2E, Model CLI, and language bootstrap. Repeated packaged-authority loading/admission and repeated wheel subprocess startup are the first measured causes to address.

## Principles

- Stay within full-suite latency optimization. Do not add product semantics, new coverage policy, or speculative scheduling infrastructure.
- Keep the same 1,409 collected tests. Optimization commits do not delete, rename, split, merge, or re-parameterize tests and do not remove existing assertions.
- Use the existing green suite and the recorded latency gap as the behavioral and performance baseline. Do not add synthetic timing tests.
- Keep a change only when a focused measurement attributes a real reduction to it.
[...]

Issue #596, opened 2026-07-29
[gda-balancing] Dogfood one periodic Effect through the ordinary Runtime schedule


qinhaihong-red:
## Parent

- #534
- #501

## What to build

Dogfood one permanent periodic Effect through the same multi-Event Experiment/Runtime path delivered
by #594 and exercised by #595. A balancing designer authors one bounded damage-over-time or
regeneration Effect, binds its per-tick magnitude Formula in Model Source, applies it, inspects its
scheduled tick and expiry Events, observes each committed Snapshot and Metric, edits the Formula,
and reruns. At least one tick shares a logical time with an ordinary combat Event so the example
exercises the normative total order and latest-committed-Snapshot rule rather than a private Effect
loop.

The admitted periodic variant must have complete semantics for its declared duration, period,
source layer, snapshot/live timing, schedule, tick contribution, expiry, Numeric behavior,
outcomes/refusals, and resource bounds. Absent immunity, stacking, dispel, buildup, multi-contributor,
or same-Event request-set variants remain absent and cannot be partially interpreted. Their complete
coverage remains with #550 and #557–#559.

## Acceptance criteria

- [ ] A complete content-addressed `game.effect` Package Release admits one bounded periodic Effect
  lifecycle and exact apply/tick/expire Operations. Kernel owns only irreducible Runtime scheduling
  laws; the Effect package owns duration/period, capture/read, tick, contribution, expiry, and typed
  outcome semantics; the receiving resource owns its state and bounds.
- [ ] Model Source owns the named pure per-tick magnitude Formula and exact binding to the Effect
  Operation slot under #590. No host callback, evaluator default, Formula fallback, example-only
  arithmetic, or second Effect expression language supplies the magnitude.
- [ ] One public Experiment executes `apply → tick → tick → expire` across declared logical times
  through #594's ordinary queue. Every lifecycle Event has a stable Event and Effect-instance
  identity, reads its declared committed Snapshot context, commits atomically, and leaves an
  observable Snapshot boundary.
- [ ] At least one Effect tick and one ordinary combat Event share a logical time. Their fixed
  phase/priority/enqueue ordering and post-commit visibility are explicit in the trace; swapping only
  the declared priority produces the deterministically corresponding Formula input, contribution,
  state, trace, and Metric result.
- [ ] Snapshot capture evaluates the Formula once at application and reuses the captured value;
  live timing reevaluates at each tick against that tick Event's pre-Event committed Snapshot. Both
[...]

Issue #595, opened 2026-07-29
[gda-balancing] Dogfood same-logical-time reciprocal combat through ordered Runtime Events


qinhaihong-red:
## Parent

- #534
- #501
- #540

## What to build

Dogfood #594's generic multi-Event Runtime with one permanent reciprocal RPG combat example. The
Model exposes two directional entrypoints—player attacks enemy and enemy attacks player—and one
Experiment queues both root Events at the same logical time. `game.combat.cast` and damage remain
directional operations; the reciprocal exchange is authored by composing exact Model entrypoints
through the ordinary Experiment/Runtime path rather than introducing a bidirectional damage
primitive or combat-specific timeline.

Coincident Events at the same logical time do not use nondeterministic host parallelism or batch
state semantics. Both Events are admitted before dispatch and execute in the Runtime's stable total
order. The later Event reads the Snapshot
committed by the earlier Event. An admitted pending attack is not silently dropped because a prior
same-time Event changed actor state; cancellation or eligibility revalidation requires an explicit
package-owned Action/Combat operation and stable Event identity. This slice makes that boundary
observable without claiming the later interruption, defeat, or turn-policy coverage owned by
#548, #549, and #564.

## Acceptance criteria

- [ ] One Model Source declares distinct player and enemy state/parameter Symbols and two exact
  entrypoints that bind the same directional combat Operation with reversed actor/target operands.
  No duplicated bidirectional Operation, host-side role swap, ambient name binding, or RPG-specific
  scheduler dispatch is introduced.
- [ ] The Experiment declares both attacks as distinct `transition-invocation` members with unique
  authored `root_event_ref` values at the same logical time; their phase is derived as
  `transition`, not caller-selected. Runtime maps both references to distinct Event ids, assigns
  enqueue sequence, executes both through #594's total order, and commits one Snapshot after each
  Event.
- [ ] Public trace proves both attacks, their exact entrypoint/call-site/operand identities, logical
  time, phase, priority, enqueue sequence, RNG draws, typed outcomes, damage, resource changes,
  state before/after, and the Snapshot observed by the later Event.
- [ ] One same-time priority vector changes only the declared priority; a separate admission-order
  vector changes only canonical root-member order. Each produces the corresponding deterministic
  enqueue/dispatch/trace identity. Repeating any exact Experiment is byte-stable under the same
  complete reproduction inputs; wall-clock or thread order cannot change the result.
[...]

PR #614, opened 2026-08-04
feat(gda-balancing): dogfood a periodic Effect lifecycle

qinhaihong-red:
## Summary

- publish the closed `game.effect@1.0.0` periodic lifecycle through the ordinary multi-Event Runtime queue
- bind snapshot and live magnitudes exclusively through Model Source Formula slots and expose exact Formula-evaluation evidence
- add conformance, rollback, publication/recovery, identity-fanout, and independent-evaluator vectors
- add a runnable periodic Effect tutorial and reconcile the narrow Runtime, package, evidence, and glossary authorities

## Dependency

This is a stacked PR based on `feat/595-reciprocal-combat` and PR #613 because #596 is blocked by #595. Review the `feat/595-reciprocal-combat..feat/596-periodic-effect` range.

Test-suite latency work is tracked separately by #597 and is intentionally absent from this branch.

## Validation

- exact branch head: `2b81e2e7428656137cfc26c884681e8e03666b72`
- full member collection: 1,409 tests
- full member result: 1,382 passed / 27 accepted skips
- packaged conformance vectors: 194
- Pyright: 0 errors
- Ruff lint and format checks: passed

## Scope

The human product/architecture acceptance gate in #596 remains open. This PR does not claim immunity, stacking, dispel, buildup, multiple contributors, same-Event request precedence, Replay, Evidence, or complete Effect/RPG coverage.

Refs #596


qinhaihong-red:
Review follow-up is complete at exact head `2b81e2e7428656137cfc26c884681e8e03666b72`.

Addressed:

- `e579ba7` anchors each operation-local periodic contract and relation declaration to the independently owned `game.effect.periodic` capability policy. Partial/all relation-vector removal while retaining the exact release, coherent tick/expiry target substitution, and inconsistent `range(period, duration, period)` schedules are now refused by both consumers.
- The Kernel remains domain-generic: it owns only the relation-policy mechanism and the non-allocating `integer-range-equal` operator; concrete duration/period/schedule/role semantics remain in `game.effect@1.0.0`.
- `2b81e2e` closes the public package schema for `integer-range-equal.right_value`: `start_path`, `stop_path`, and `step_path` are all required and closed, while canonical object literals for other operators remain supported.

One provisional finding was withdrawn after architecture re-evaluation: deleting all package-owned periodic semantics and recomputing Package/vector/LDB identities constructs a different Language World, not a substitution of the original exact release. bADR-0016 explicitly permits the same logical `(id, version)` coordinate to bind different content under different LDB identities. With the original exact LDB/release identity fixed, policy deletion is refused by the existing identity chain. No Kernel-level Effect allowlist or duplicate package-local authority was added.

Validation:

- Required CI matrix: all required checks passed at exact head, including fast, authority, language, composition, wheel/subprocess, inventory closure, Ruff, Pyright, package build, co-install, and release-scope guard.
- Pytest collection: 1,409 tests.
- Relation/removal/substitution/range targeted review suite: 32 passed.
- Public-schema extended mutation matrix: 11 passed.
- Pyright: 0 errors, 0 warnings; Ruff and `git diff --check`: passed.
- Final delegated Standards + Spec review: no findings, APPROVE.

The human product/architecture acceptance gate on #596 remains intentionally open.

qinhaihong-red:
The member-owned #597 optimization is complete on this PR at `6d1ece3`; the root-owned workflow wiring is isolated in stacked PR #615.

Final evidence:

- accepted `1,409 tests / 194 vectors / 27 skips`; optimized `1,537 tests / 194 vectors / the same 27 skips`
- 19/19 current claims and 333 subjects closed; accepted 15/305 mapping closed
- local cumulative median `1292.539s → 986.854s` (**23.6% improvement**)
- local wall median `1297.274s → 1001.421s` (**22.8% improvement**)
- required CI median `483s → 293s` (**39.3% improvement**), provisional max 294s
- slowest optimized CI shard 222s, with 53.8% headroom below 480s
- independent subagent review: **APPROVE**, no P1/P2
- all checks green on #614 and #615

The full three-run tables, exact-SHA run links, measurement hygiene, and review evidence are posted on #615.

qinhaihong-red:
## Review — Standards & Spec (two-axis)

**Range reviewed:** `origin/feat/595-reciprocal-combat..2b81e2e7` (the branch head before #616 was merged in; #616's test-latency work was reviewed separately and is excluded here). Spec basis: #596 plus this PR's declared scope.

---

### Standards axis

#### Hard finding (verified)

**`src/gda_balancing/schema2/bootstrap.py:1410` reads a leaked loop variable `kind`**

```python
and set(relation["probe"]) == set(kind["probe_members"])
```

`kind` is bound inside the `for vector in vectors:` loop starting at line 1091; line 1410 sits in the policy-validation block after that loop, so it reads the **last vector's kind** instead of the authoritative `relation_kind` fetched at line 1297. Consequences:

1. Relation probe member sets are compared against an arbitrary kind's `probe_members`.
2. The `runtime-scenario` / `value-program` / `scheduler-scenario` kinds have no `probe_members` (checked against `kernel.json`) — if a package's last vector is one of those, this raises `KeyError` and escapes admission instead of producing a typed refusal, against ARCHITECTURE §2.1 ("refuse … explicitly").
3. With an empty `vectors` list, `kind` is unbound.

This also breaks the single-authority rule (RULES.md): the `operation-relation` contract's sole authority is the Kernel entry, i.e. `relation_kind`. It is currently masked only because `game.effect@1.0.0`'s last vector happens to be `operation-relation`, and no test covers the ordering. **Suggested fix:** use `relation_kind["probe_members"]` (or the resolved declaration extension) and add a vector-ordering / empty-vectors regression test.

#### Judgement calls (smell baseline — not blockers)

- **Duplicated Code** — `experiment.py:1348-1404` vs `4276-4331`: the formula-evaluation record construction (slot argument mapping, sorted arguments, nine-member dict, `evaluation_sites.get(index+1)` criterion) is repeated line-for-line. The replay assertion requires the two to stay byte-identical, yet there is no shared constructor (both already share `_instruction_evaluation_sites` / `_operation_formula_slot`, so extraction is unobstructed).
- **Middle Man / mode switch** — `_replayed_schedule_arguments` (`experiment.py:1445`) now only forwards and takes `[0]`; `_replayed_event_evidence` uses `target_schedule: … | None` to carry two jobs and returns a triple-optional shape, which hurts readability.
- **Repeated Switches** — `commands/package.py:415` adds a fourth `if kind_id == …` branch to `_package_vector_schemas` (each branch 100–170 lines, same three closing statements at 518-528 / 713-723 / 891-901); a per-kind constructor map would collapse this.
- **Duplicated Code / Data Clump** — `model.py:555` and `589` repeat the `package_versions` + `operations` + `selected` triptych; the `(package, version, id)` triple keeps travelling bare through `model.py` and `experiment.py:389/531` ("Operation coordinate" wants to become a type). Pre-existing idiom, mild.
- **Primitive Obsession** — `experiment.py:380/383`: `_INVALID_FORMULA_EVIDENCE = object()` degrades the annotation to `JsonValue | object`, forcing identity comparison on every caller.
- **DRY (glossary)** — `BALANCING-CONTEXT.md:1075` "Magnitude timing policy" restates the `snapshot`/`live` axis already owned by "Effect specification (2.x)" (line 689) and "Formula evaluation context" (line 423); line 1068 "Effect instance" defines instance identity via the apply-time Named-stream draw while line 689 says instances arise from buildup activation — the two entries don't cross-reference. Suggest one-way references to the existing entries.

Clean: bADR-0014/0016/0018 amended via dated `> Amendment` notes without rewriting decisions; ARCHITECTURE 12.9/12.10 renumbering leaves no dangling references; no game/engine identity leaks; tests drive the real CLI with real artifact mutation (no Fake/Mock-only evidence); the new refusal reasons are guarded by the exact-catalog consistency test.

---

### Spec axis (vs #596)

All 13 acceptance items land except the human product/architecture acceptance gate, which this PR explicitly leaves open. The tutorial (`examples/schema2/rpg-periodic-effect/README.md`) was executed end-to-end during review and every asserted number reproduced (snapshot capture `15`, `100→85→70`; live combat-first `85`, tick-first `75`, snapshot same-time `60`; after retuning, Formula `0`, final `100`, with `package_lock_identity` / `kernel_identity` unchanged).

**(a) Partial** — AC4: *"swapping only the declared priority produces the deterministically corresponding Formula input, contribution, state, trace, and **Metric result**"*. In the published `same-time-experiment.json`, `target_health_remaining` has `target: {minimum: 0, maximum: 1000}`, so both orderings (85 / 75) pass and the Metric cannot witness the ordering difference; only the trace assertions in `tests/test_e2e_cli.py:836-840` carry it. Consider tightening the published target (or adding a second metric) so the Metric axis witnesses AC4 too.

**(b) Scope creep** — none of substance. One unused generalization: `kernel.json` `operators` declares `canonical-equal`, but the 12 `game.effect` relation vectors use only `integer-*` and `schedule-projection-equal`; no consumer exists repo-wide.

**(c) Implementation concern** — **receiving-resource bounds are not enforced at runtime.** AC1: *"the receiving resource owns its state and bounds"*; AC6: *"Numeric bounds … are statically bounded or Runtime-budgeted"*. Reproduced live: setting `magnitude_threshold = 0` makes the snapshot capture magnitude `100`, and after two ticks the committed `target_health` is `-100` — outside the `closed-interval [0, 1000]` declared at `examples/schema2/rpg-periodic-effect/model-source.json:41-45` — yet the command exits 0 and publishes Snapshot/Metric with no runtime refusal, contradicting `BALANCING-CONTEXT.md:1161` (a Runtime Operation violating its declared domain is a Runtime refusal). The unclamped `subtract-state` already exists in `game.combat.damage-v1`, so this is **not introduced by this PR** — but #596 lists it as an acceptance item for this slice and no vector covers it. Suggest adding a domain-refusal vector, or explicitly recording the gap on #596 as a known carry-over.

Minor: the new example's `experiment.json` uses all-zero identities + jq rebinding, diverging from `rpg-combat-cast`'s concrete-identity convention (documented in the tutorial; noted for the record).

---

**Summary:** Standards — 7 findings (1 hard: the leaked `kind` at `bootstrap.py:1410`; 6 judgement calls). Spec — 3 findings (worst: declared-bounds violations are not refused at runtime; pre-existing, but an acceptance item of this slice with no coverage).

qinhaihong-red:
Review follow-up is complete at exact head `fb0b5cc2d1ed7dbc4d4608eec8e059a9b623e1c3`.

### Decision record

- **Fully adopted — leaked `kind` in operation-relation validation.** Production and the independent consumer now read the authoritative `operation-relation` kind. Reordered and empty-vector regressions prove that admission is order-independent and produces a typed refusal rather than a host exception.
- **Fully adopted — duplicated Formula-evaluation record construction.** Runtime and replay now call one local pure constructor, preserving the exact public record shape without adding a protocol or mode abstraction.
- **Partially adopted — AC4 Metric evidence.** The test now asserts the actual `target_health_remaining` Metric result for both priority orders (`85` and `75`). I did not tighten the tutorial target or add another Metric: AC4 requires the Metric result to change, not the pass/fail verdict, and changing the authored acceptance target would add unrelated tutorial policy.
- **Fully adopted — receiving-resource bounds.** State writes now enforce the receiving declaration's closed interval and reuse the existing `runtime.numeric_overflow` refusal. The public periodic path proves that the second tick is refused, only that event is rolled back, the committed apply/first-tick prefix remains intact, and the terminal audit is publishable. The independent reference evaluator covers the same within-Int64/domain-overflow case.
- **Fully adopted — checked-in Experiment identities.** Both periodic Experiment documents now bind the exact checked-in Model build. The ordinary tutorial paths run them directly; rebinding remains only for an edited Model Source.
- **Fully adopted — glossary cross-references.** `Effect instance` and `Magnitude timing policy` now point to the broader Effect/capture-timing concepts without creating another authority.
- **Not adopted — broad replay mode refactor, package schema constructor map, `OperationCoordinate` type, sentinel annotation refactor, and removal of `canonical-equal`.** These observations identify no current correctness failure in this slice. Implementing them here would widen ownership and churn disproportionate to the review problem; `canonical-equal` also remains a valid generic relation operator even though this periodic package does not consume it.

One adjacent fix was required for closure: terminal-audit validation now reads a scheduled Event's structured operation reference by its `id`. The new domain refusal exposed that old mismatch; without the local fix, the valid refusal artifact could not pass semantic admission.

### Validation

- Local required shards plus smoke: **1,386 passed / 27 existing skips** (`1,413` collected); inventory closure passed with no missing or overlapping tests.
- Targeted review suite: **14 passed**, including independent replay/evaluator and coherent evidence-mutation checks.
- Ruff lint/format, Pyright (`0 errors, 0 warnings`), JSON parsing, and `git diff --check`: passed.
- Exact-head CI: all required checks passed, including six test shards, wheel/subprocess smoke, inventory closure, package build, co-install, and release-scope guard.
- Delegated exact-head review: **no actionable findings**; the accepted fixes are closed and the explicit non-adoptions remain proportionate.

The human product/architecture acceptance gate on #596 remains intentionally open.

PR #617, opened 2026-08-05
docs(skills): add entropy review and align validation design

qinhaihong-red:
## Summary

- add a project-local `entropy-review` skill for proportionality and software-entropy review
- rename `evidence-driven-architecture-design` to `validation-driven-design`, aligning with common `*-driven-design` naming while preserving its architecture-specific triggers and workflow
- keep the skills independently installable while preserving the reasoning order: establish architecture requirements, decision support, and checks before comparing complexity with simpler options

## Validation

- both skill folders pass the official `quick_validate.py`; `git diff --check` passes
- independent Spec review found no missing requirement, scope creep, or incorrect implementation
- independent Standards review found no hard violation; forward tests confirmed proportional review preserves necessary checks and lightweight design does not expand into a full architecture program
- accepted review findings on ambiguous ordering, repeated guidance, private taxonomy, verdict vocabulary, and legalistic prose; retained `agents/openai.yaml` only for the new skill because it is recommended UI metadata, not a repository-wide convention
- branch diff is confined to `.agents/skills/**`

PR #618, opened 2026-08-05
docs(skills): add Godot modular architecture design

qinhaihong-red:
## Summary

- add a reusable Godot modular architecture design skill
- define broad library-style Add-ons alongside Systems, Content, and UI ownership
- require acyclic downward dependencies with Godot-style "call down, signal up" communication
- cover scene composition, Autoload ownership, concrete placement examples, and proportionate review workflow

## Review follow-up

- clarify that Add-on is a local broad-library convention and distinguish first-party libraries from third-party plugins
- make physical project roots optional until their responsibilities exist
- separate application coordination from authored assets within Content by owner and reason to change
- replace ambiguous terminology and remove repeated dependency guidance

## Validation

- skill-creator `quick_validate.py`: passed
- `git diff --check`: passed
- terminology sweep: passed
- initial independent subagent review completed before the PR
- PR review feedback addressed in `bb5b2e9`

### DIFF
--- a/.agents/skills/design-godot-modular-architecture/SKILL.md
+++ b/.agents/skills/design-godot-modular-architecture/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: design-godot-modular-architecture
-description: Design and review modular Godot project architectures around Add-ons, Systems, Content, and UI, with acyclic downward dependencies and indirect upward communication. Use when structuring a new Godot project, separating reusable systems from authored content, reviewing architectural coupling, deciding where scenes, scripts, and resources belong, or planning an incremental modularization.
+description: Design and review modular Godot project architectures around Add-ons, Systems, Content, and UI, with acyclic downward dependencies and indirect upward communication. Use when structuring a new Godot project, separating reusable systems from authored content, reviewing architectural coupling, deciding where scenes, scripts, and resources belong, planning an incremental modularization, or isolating experiments without weakening production boundaries.
 ---
 
 # Design Godot Modular Architecture
@@ -188,6 +188,18 @@ Do not place every artifact for one topic in the same area. Place each part acco
 
 Use this split to keep one owner for each rule and state, avoid hard-coded assets in Systems, and avoid project language in Add-ons.
 
+## Isolate Experiments When Needed
+
+When production boundaries make early exploration unnecessarily slow, use an optional development-only sandbox. Name and organize it according to project convention, such as by experiment or contributor. Do not treat it as a fifth production layer or a required project root.
+
+Allow sandbox code to depend on any production area. Do not let production code, scenes, Resources, Autoload registrations, or project settings reference sandbox artifacts.
+
+Exclude the sandbox from every release export and verify the exclusion in the build process; automate the check when CI produces releases. Do not assume that a directory name has special export behavior.
+
+Relax internal implementation standards when speed matters, but keep the isolation boundary strict. Commit shared experiments when team testing or feedback is useful; keep throwaway work local when sharing has no value.
+
+Do not promote sandbox code merely by moving files. Once an experiment is accepted, classify each responsibility under Add-ons, Systems, Content, or UI; refactor or reimplement it to production standards; verify the resulting behavior; and remove the sandbox version. Temporary duplication inside the sandbox is acceptable while the design is uncertain.
+
 ## Workflow
 
 ### 1. Pin Down the Current Need
@@ -216,6 +228,7 @@ Read the project's architecture docs, ADRs, and project conventions when they ex
 - Scene and Resource references.
 - Scene inheritance and scene instantiation.
 - `@tool` scripts and other editor-only code.
+- Experiment or prototype directories, export presets, and build exclusions.
 - Parent-child lifecycle assumptions.
 - State owners and mutation paths.
 - Signals, callbacks, groups, and event channels.
@@ -244,6 +257,7 @@ Draw the source dependencies required by the current behavior. Check for:
 - Global type names, node groups, or string identifiers that hide coupling.
 - Several modules owning the same rule or state.
 - Scene and Resource references that cross a forbidden boundary.
+- Production references to experimental sandboxes.
 
 Do not create an interface for every call. Add indirection only when it protects a real boundary, supports a needed substitution, or enables upward communication.
 
--- a/.agents/skills/design-godot-modular-architecture/SKILL.md
+++ b/.agents/skills/design-godot-modular-architecture/SKILL.md
@@ -9,14 +9,14 @@ description: Design and review modular Godot project architectures around Add-on
 
 Design the smallest architecture that gives the current project clear ownership, useful reuse boundaries, and one-way dependencies.
 
-Use four project roots:
+Default to four project roots:
 
 - `addons/`: reusable technical libraries.
-- `systems/`: reusable model and rule-driven capabilities.
+- `systems/`: reusable game rules and domain state.
 - `content/`: application flow and project-specific content.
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Review — consistency / completeness / orthogonality + terminology

**Overall:** the prose is markedly better than PR #617 — concrete, imperative, domain-grounded, and most of the content is real knowledge. Remaining issues: one core terminology problem (redefining Godot's `addons/`), two practical completeness gaps (UI scene composition, Autoload discipline), one self-contradiction, and the same internal-repetition habit as #617, in milder form. Recommend revising before merge.

Process checks: branch is cleanly based on current `origin/main`, CI is green (7 pass, 5 path-gated skips), the `docs` type is correct (non-releasing), and the content contains no demo-project references. The branch name `feat/panda-dev` does not reflect the change (a docs/skills addition, nothing panda-related) — not blocking, but future branches should be named for their actual content.

### Should fix before merge

**1. Redefining the established Godot term `addons/` (core terminology issue).**
The doc admits the overload itself: "Use **Add-on** in the broad sense of a reusable library module. It is not limited to a Godot plugin with `plugin.cfg` …". But `addons/` has an established, specific meaning in the Godot ecosystem — the install target for editor plugins and Asset Library packages, with engine-special handling (plugin scanning, enable/disable). The overload's real costs go unaddressed: first-party generic libraries and third-party downloaded plugins end up mixed in one root (the "Typical contents" list places "Third-party Godot plugins" alongside self-written helpers), which muddies exactly the ownership boundary the skill is about; and "Add-ons may depend on … explicitly declared peer Add-ons" cannot govern third-party plugins, whose dependencies are not yours to declare. Needing a paragraph to redefine an engine-reserved word is itself the signal. Recommend either a separate root for first-party reusable libraries (`lib/`, `core/`, or similar), leaving `addons/` its established meaning — or, at minimum, stating the trade-off head-on and how mixed first-/third-party ownership is handled.

**2. UI scene composition is unspecified (completeness gap — the first question a practitioner will hit).**
The rule chain: Content must not depend on UI, and "Scene and Resource references that cross a forbidden boundary" is a violation. Consequence: a level scene instancing a HUD `CanvasLayer` is forbidden — which is the single most common pattern in real Godot projects. So who instances the HUD, and where does it live in the tree? The bootstrap paragraph implies the main scene / startup script composes UI with Content, but it is never said. Add a few sentences making it explicit: scene instantiation *is* a dependency; UI scenes are composed by the bootstrap / UI root, not embedded in Content scenes.

**3. No placement rule for Autoloads (completeness gap).**
Autoloads appear three times (registration allowed in `project.godot`; on the inspection list; "Autoloads that bypass ownership boundaries" flagged as a violation), but the doc never answers: which layer may an Autoload belong to? May a System be an Autoload? Should event channels be Autoloads? And given the bootstrap already "creates and connects modules, injects resources or callbacks" — what legitimate use remains for Autoloads at all? Autoloads are the number-one coupling vector in Godot projects; a Godot modular-architecture skill that inspects them but never governs them leaves its biggest hazard unregulated.

**4. "Use all four project roots" contradicts the skill's own principles (consistency).**
The Goal says "Design the **smallest** architecture …"; the Scale section says a small project may need only "clear ownership and a few visible dependencies"; the same paragraph says "Create only the modules the project needs" — yet the four roots are stated as a mandate. A small project may have no reusable technical library at all. Change to "Default to four project roots" and explicitly allow collapsing or omitting roots for small projects.

**5. Content merges two concerns the doc itself calls different, contradicting §3's own criterion (orthogonality).**
Step 3 classifies "by reason to change and reuse boundary"; the Content section then concedes that application flow and authored assets "have different owners and reasons to change", and waves it off with "separate them in the way that best fits the project". The actual adjudication exists and should be stated: by reuse boundary both are project-specific and non-reusable, so they share a root — reuse boundary outranks reason-to-change here. Say that, instead of leaving a visible conflict with the skill's own classification rule.

**6. Internal repetition (same pattern as #617, milder).**
The one-way dependency rule has five homes: the master matrix plus a "Do not let X depend on Y" restatement in each of the four sections. Communication guidance has two homes (the paragraph in the layering section vs. step 5). The proportionality/YAGNI theme recurs in roughly seven places (Goal, step 1, step 4, step 6, step 7, Scale, Output). The closing "Keep the Design Practical" section is entirely recap — four bullets, all restating earlier content, zero new information. Recommend: the master matrix as the single authority with at most one concrete per-section reminder; fold the layering section's communication paragraph into step 5; delete the closing section.

### Terminology and style

- **"Model" is overloaded, at high frequency on both meanings**: Systems' "model behavior / model state / world-model state" (domain model) collides with Content's "Models, textures, materials" (3D models) in the same document — a genuine ambiguity in a game-dev doc. Use "game-model state" or "domain state" on one side and "3D models" or "meshes" on the other.
- **Sibling-skill house vocabulary leaks in**: "Read the project's **authority documents**" is the validation-driven-design family's private jargon; since the skills are deliberately decoupled, use plain words here ("the project's architecture docs / ADRs / conventions"). Same for "reuse pressure" (plainer: "expected reuse").
- **Empty bullet**: "Cohesive world-model state owned by a named System" is circular (Systems typically contain … state owned by a System). Delete or make concrete.
- **"Dialog cards"** is not an established term, and the same list mixes "dialog" and "dialogue" with both senses in play (dialog = UI window, dialogue = conversation). Standardize.
- **Positive suggestion**: the whole dependency-direction chapter is the established Godot community maxim **"call down, signal up"** — naming it anchors readers in one step and follows the use-established-terms principle better than any restatement.

### Minor

- **PR body accuracy**: "independent subagent review … no remaining findings" is contradicted by the findings above; update the Validation section before merge. Same pattern as #617 — this validation-claim template is worth rethinking.
- **`agents/openai.yaml` convention still partial**: now two new skills carry it while the five existing skills don't; the note from the #617 review stands.

qinhaihong-red:
Thanks for the thorough review. Addressed in `bb5b2e9`.

Adopted:

- Scene instantiation is now explicitly treated as a dependency. Content and UI are composed by the bootstrap or a UI-owned composition scene; Content scenes do not embed UI scenes.
- Autoload is defined as a registration and lifetime choice, not a layer. Each Autoload remains under its owning root and follows the same dependency matrix. The guidance also covers System Autoloads, injected alternatives, and event channels.
- Content now explains why application flow and authored assets share the project-specific reuse boundary, while requiring internal separation by owner and reason to change.
- Ambiguous terminology was replaced: domain state and game state replace the overloaded model wording, 3D models and meshes identify visual assets, and dialogue panel replaces dialog card. The workflow now uses plain names for project docs and expected reuse.
- The dependency matrix remains the single authority. Repeated per-section rules and the recap section were removed, while concrete Godot boundary reminders were retained.
- The PR body was updated so it no longer claims that no findings remained.

Partially adopted, by design:

- The broad Add-on meaning is retained because it is the architecture convention this skill is documenting. The text now labels it as a local convention, separates first-party libraries from third-party plugins by ownership, requires visible first-party Add-on dependencies, treats third-party plugins as external packages, and separates editor-only code from runtime code.
- The four roots remain the conceptual default, but empty physical roots are no longer mandatory. A small project may omit a root until that responsibility exists; present responsibilities should not be collapsed into the wrong owner.

Not changed in this PR:

- `agents/openai.yaml` remains because it is valid metadata produced by the current skill scaffold. Aligning older skills would be a separate repository-wide cleanup.
- The existing branch was not renamed after the PR was opened; future branch names should reflect the actual change.

Validation after the revision:

- skill-creator `quick_validate.py`: passed
- `git diff --check`: passed
- terminology sweep: passed
- all required CI checks: passed

PR #619, opened 2026-08-05
docs(skills): add skill-review skill

qinhaihong-red:
## Summary

- add a project-local `skill-review` skill for reviewing skill documents and skill-focused changes
- provide a concise four-step workflow for inspection, review, validation, and reporting
- review correctness and usability, consistency and completeness, orthogonality and DRY, terminology, and prose
- inspect companion resources progressively, make repository checks conditional, and require explicit authorization for remote writes
- keep the guidance domain-neutral, separate findings from proposed fixes, and require passing reviews to summarize their evidence and validation gaps

## Validation

- based on `origin/main` at `40de7e52`
- `skill-creator` `quick_validate.py`: passed
- `git diff --check` and targeted redundancy and terminology sweeps: passed
- self-review at `6df41092` exposed the evidence-free `Pass` output; the reporting rule now addresses that observed failure
- blind positive-path test against merged PR #618 at `6df41092`: **Pass**
- blind negative-path test with the current skill against historical artifact `5529b049`: **Changes required**, with six evidence-backed findings and explicit validation limits

### DIFF
--- a/.agents/skills/skill-review/SKILL.md
+++ b/.agents/skills/skill-review/SKILL.md
@@ -67,6 +67,8 @@ Start with one conclusion:
 
 Then group actionable findings under **Required changes**, **Terminology and prose**, or **Minor** as appropriate. For each finding, give the location, evidence, impact, and the smallest practical alternative that preserves the intent. Keep optional improvements separate and include them only when they materially help.
 
+If new context invalidates a proposed fix but not the underlying problem, revise the fix instead of dropping the finding. For example, if a proposed cross-reference conflicts with a self-contained skill, keep the finding that an instruction relies on missing context and propose a self-contained rewrite.
+
 If there are no substantive findings, report **Pass** with the reviewed target and a one- or two-sentence summary of the material checks performed. Mention any important validation that was not performed, then stop.
 
 Return the review in the medium the user requested. Edit files, post comments, or perform other remote writes only when the user explicitly authorizes them.
--- a/.agents/skills/skill-review/SKILL.md
+++ b/.agents/skills/skill-review/SKILL.md
@@ -1,73 +1,89 @@
 ---
 name: skill-review
-description: Review skill documents (SKILL.md and companion files) for consistency, completeness, orthogonality, and terminology accuracy — flagging coined terms, generated-prose patterns, and legalistic phrasing, each with a smaller concrete alternative. Use when reviewing a new or changed skill document, reviewing a PR that adds or renames skills, or re-checking an author's claimed fixes after a skill review.
+description: Review skill documents and skill-focused changes for correctness, consistency, completeness, orthogonality, DRY, terminology accuracy, and clear prose. Use when reviewing a new or changed skill, checking whether its workflow and companion resources are usable, reviewing a PR that adds or renames skills, or re-checking claimed fixes.
 ---
 
 # Skill Review
 
 ## Goal
 
-Review whether a skill document is **accurate** (terminology), **complete** (for its audience), and **economical** (orthogonal structure, no filler). Report only findings that can change a decision or an implementation; when nothing substantive is wrong, pass the document and stop — do not manufacture findings. Keep the review cost proportionate to the document's size. This skill holds itself to its own standards.
+Decide whether a skill can do its declared job with accurate instructions and the smallest sufficient structure. Report only actionable issues supported by the artifact or verifiable evidence. Scale the review to the skill's size, risk, and intended use.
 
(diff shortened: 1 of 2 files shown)


qinhaihong-red:
## Re-review of the refactor — conducted under `entropy-review` principles

### Verdict: **Keep** (adopt the refactor)

The refactor's direction and substance are right: entropy drops materially (content regrouped into a four-step workflow, slogans stripped, the three-state conclusion collapsed to two), and each piece of added complexity carries its own evidence. Two spots, however, were over-trimmed past essential content and need small restorations — per entropy-review's own first boundary, "Preserve essential complexity."

### Goal and minimum sufficient solution

Goal: let an agent review skills and skill-focused changes with the accumulated principles. The minimum sufficient solution is approximately what the refactor delivers — four-step workflow, dimension checks, report contract. The deviations are only the two findings below.

### Findings

**F1 — Zero examples contradicts the skill's own rule. Action: partially restore.**
The refactored document itself requires "Keep concrete domain examples when they teach a non-obvious distinction" and "Do not let a skill contradict its own guidance" — yet every example was deleted. At least three of the removed examples taught exactly such non-obvious distinctions: the anonymous cross-reference sentence (one that only parses against an unnamed sibling skill), the obligations that follow from widening a domain-reserved word, and verdict vocabulary that fails one of the declared inputs. Their generalized stand-ins ("remove hidden assumptions about sibling skills and unexplained private vocabulary") are markedly weaker; a future reviewer cannot reconstruct the detection method from the abstract sentence. Smaller alternative: restore the two or three highest-teaching-value parenthetical examples (~4 lines) — not all fifteen; full restoration would be the drive for completeness this review is meant to catch.

**F2 — The "re-derive the remedy from stated intent instead of withdrawing the finding" rule was lost. Action: restore one sentence.**
The refactor keeps only the re-review item for designs declined "by design" — an after-the-fact check, not the mid-review rule for what happens when an author's stated intent invalidates a finding's remedy. This rule changed a real outcome once: a deliberate no-cross-references model invalidated "add a cross-reference" but sharpened the finding into "delete the anonymous ordering sentence." Without it, the natural reviewer behavior after an intent explanation is to withdraw the whole finding. One sentence in the Report step suffices.

### New obligations — all justified, keep

- **Correctness and Usability axis + forward-testing**: evidence-backed (self-review exposed an evidence-free Pass; blind test against a merged PR) and fills a real gap — the original never checked whether instructions are executable.
- **Explicit authorization for remote writes**, **conditional validation**, and **"a passing validator is structural evidence, not proof"**: each carries current value.
- **"Do not infer who or what produced the prose"**: turns style review into observable text properties. Sound.

### Accepted trims (no restoration requested)

The canonical-maxim heuristic and the "needing a paragraph to redefine an established word is itself the signal" heuristic (weakly covered by the surviving terminology rules); the family-collaboration-model premise (operationally covered by the self-containment default); three-state → two-state conclusion; the "holds itself to its own standards" slogan — rightly deleted, though the property must now be earned through F1.

### Process checks

Branch rebased onto current `origin/main` (`40de7e5`); CI green (7 pass / 5 path-gated skips); merge state CLEAN; `agents/openai.yaml` consistent with the new description; the PR body matches the diff, with no scope beyond what it declares; the self-review disclosure is honest. `quick_validate.py` remains an external tool whose run could not be reproduced locally.

### Assumption to test

The blind forward test covered one case and returned **Pass** — the **Changes required** path has never been forward-tested. Cheapest check: blind-test a historical version known to have defects (for example, the initial entropy-review draft) and see whether the review reproduces its main findings.

Lean order: F2's one sentence → F1's two or three examples → (optional) the negative-path blind test.

qinhaihong-red:
Addressed the actionable parts of the latest entropy review in `96fb1b0`.

- **F2 fully adopted:** the Report step now says to revise a proposed fix when new context invalidates the fix but not the underlying problem.
- **F1 partially adopted:** added one concise, domain-neutral example that demonstrates the same finding-versus-remedy distinction. I did not restore the historical domain-reserved-word or verdict examples because their rules are already explicit and those examples would reintroduce coupling or obsolete cases.
- **Negative-path forward test completed:** a clean-context subagent used the current skill to review historical artifact `5529b049`. It returned **Changes required** with six evidence-backed findings covering correctness/usability, progressive inspection, PR-only assumptions, unauthorized remote writes, the output contract, and provenance inference. It also disclosed its validation limits.

Validation: `quick_validate.py` passed, `git diff --check` passed, the PR body matches the final scope, and all required CI checks are green.

PR #623, opened 2026-08-05
refactor(panda-adventure): adopt modular runtime architecture

### LINKED ISSUE #621: [Panda Adventure] Refactor runtime into one-way modular architecture
## Background

Panda Adventure's runtime was organized mostly by technical type. UI, application orchestration, gameplay rules, concrete content, and development tools appeared as peers, so ownership and dependency direction were difficult to see and easy to violate. A unique Player dependency was discovered through a global group, and the development editor lived beside shipped runtime scenes. Moving data or assets also required coordinated path changes across runtime code, tooling, tests, and documentation.

The refactor must preserve behavior while making the runtime architecture explicit and mechanically checkable.

## What to build

Refactor the runtime around one downward dependency chain:

`UI -> Content -> Systems -> Add-ons -> Godot`

Use the roots as ownership boundaries:

- **Systems** own gameplay state, rules, invariants, and reusable domain behavior.
- **Content** owns application orchestration, concrete scenes, configuration, data, assets, and presentation composition.
- **UI** owns screen-level presentation and the composition shell that sends intent downward and observes state changes.
- **Add-ons** are project-independent libraries in the broad sense; they are not limited to Godot editor plugins.

Move development-only editor surfaces outside the shipped runtime chain. Preserve existing data authority, asset generation, balancing workflows, and runtime behavior throughout the migration.

## Entropy-reviewed plan

The design was reviewed with the entropy-review workflow and deliberately kept smaller than a full DDD implementation:

- Treat domain-driven design as a boundary and ownership guide, not as a requirement to introduce repositories, services, interfaces, or other patterns without a concrete need.
- Allow direct calls in the downward dependency direction. Use narrow signals or callbacks for upward communication.
- Let Content select and configure Systems; do not let Systems depend on concrete Content configuration, scenes, or UI.
- Let the UI composition shell bind Content entry points, observe published state, and translate input into application intent.
- Replace only the unique Player group lookup with explicit injection. Keep capability-oriented groups where they remain useful for one-to-many discovery.
- Add a mechanical dependency guard instead of relying only on prose.
- Keep Python tools, tests, and documentation outside the runtime layer model.
- Do not add an event bus, service locator, interface hierarchy, or experimentation sandbox until a demonstrated need justifies it.

This remains one wide refactor because the directory migration, dependency direction, path updates, composition changes, and export boundary share one behavior-preservation contract and cannot be validated independently after the fact.

## Acceptance criteria

- [ ] Runtime files are organized under Add-ons, Systems, Content, and UI roots with the documented ownership boundaries.
- [ ] Production dependencies point only within the same root or downward through `UI -> Content -> Systems -> Add-ons -> Godot`.
- [ ] A mechanical test rejects upward or cross-layer runtime dependencies.
- [ ] Systems contain gameplay rules and state without depending on concrete Content configuration, scenes, or UI.
- [ ] A UI composition shell wires the runtime, sends user intent downward, and observes end-state changes indirectly.
- [ ] The unique Player dependency is injected explicitly; capability-oriented groups remain available for one-to-many discovery.
- [ ] Development editor scenes and scripts are outside shipped runtime roots and excluded from the player export.
- [ ] Data, schema, generated resources, assets, build scripts, balancing tools, tests, and documentation use the migrated paths consistently.
- [ ] The architecture decision is recorded in the Panda Adventure domain documentation.
- [ ] The full Panda Adventure test suite, lint checks, production GDScript validation, headless boot, and a real macOS export pass.

## Blocked by

None.

### Key context

- Start from the Panda Adventure game context and local architecture decisions before changing runtime ownership.
- Preserve the JSON-to-generated-resource authority chain and existing asset lifecycle behavior.
- Treat the dependency guard, headless boot, and exported pack contents as architectural evidence rather than relying on directory names alone.

### Implementation notes

- Prefer moves plus focused composition changes over behavioral rewrites.
- Update path consumers together: runtime resources, project configuration, export configuration, Python tools, tests, and domain documentation.
- Keep upward communication narrow and local to the collaboration that needs it.

### Out of scope

- Replacing useful capability groups with a general registry.
- Introducing a project-wide event bus or service locator.
- Reworking gameplay rules, balancing values, or asset generation semantics.
- Adding a permanent experimentation area without a current experiment to host.

### Risks / edge cases

- Godot resource files and imported asset metadata may retain stale paths after moves.
- A development-only editor resource can be pulled into the export transitively even when its top-level scene is excluded.
- Tests may pass while the real project boot or exported pack still references an old path.

qinhaihong-red:
## Summary

- reorganize the Panda Adventure runtime into Add-ons, Systems, Content, and UI with one-way dependencies
- add a UI composition shell, explicit Player binding, and a Systems-owned movement model while preserving capability groups
- move development editor surfaces outside the shipped runtime and synchronize data, asset, tooling, test, export, and documentation paths
- add a mechanical dependency guard and record the decision in gADR-0020
- emit a structured composition-ready event and remove the remaining migrated-path documentation drift

Closes #621

Supersedes #622, which GitHub closed when its head branch was renamed to match the repository's `feat/<issue>-<slug>` convention.

## Architecture

The runtime dependency chain is:

`UI -> Content -> Systems -> Add-ons -> Godot`

Direct calls are allowed downward. Upward communication uses narrow signals or callbacks. Content selects and configures Systems; Systems do not depend on concrete Content scenes or configuration. "Add-ons" is used in the broad library sense, not only for Godot editor plugins.

The entropy review intentionally rejected a general event bus, service locator, interface hierarchy, and permanent experimentation sandbox because the current collaborations do not require them.

## Validation

- `env UV_CACHE_DIR=/tmp/panda-modular-uv-cache uv run pytest examples/platformer/panda-adventure/tests` — 548 passed, 4 skipped
- `env UV_CACHE_DIR=/tmp/panda-modular-uv-cache uv run ruff check examples/platformer/panda-adventure` — passed
- `env UV_CACHE_DIR=/tmp/panda-modular-uv-cache uv run ruff format --check .` — passed
- modular dependency guard — passed as part of the full suite
- production headless boot and GDScript seams — passed as part of the full suite
- `examples/platformer/panda-adventure/scripts/export_macos.sh` — real macOS release export passed
- `git diff --check` — passed

## Review focus

- verify that ownership matches the documented layer responsibilities, not just the directory names
- check scene/resource paths, imported asset metadata, generated-resource authority, and export boundaries for migration drift
- check that upward communication remains indirect and local without hiding dependencies behind global infrastructure


qinhaihong-red:
Three-axis review of this PR at head 84f6736 (`origin/main...feat/621-panda-modular-architecture`): **Standards** (repo/domain conventions + smell baseline), **Spec** (issue #621 conformance), and **Entropy** (implementation-time re-check against the design-time entropy-reviewed plan). Axes are reported separately and not cross-ranked.

## Standards

### Documented-standard breaches

1. **New domain terms never entered the glossary.** `AGENTS.md` + `docs/agents/domain.md` make `GAME-CONTEXT.md` the game's shared-language authority. The PR promotes **Game Shell**, **Gameplay**, and the four module roots (Add-ons/Systems/Content/UI) to first-class terms — used in `AGENTS.md:54-58`, throughout gADR-0020, and inside three existing glossary entries (`GAME-CONTEXT.md:589`, `:599`, `:612-617`) — yet adds no glossary entry for any of them. Terms are consumed by the glossary without being defined in it.
2. **Historical gADR paths retro-edited with two mechanisms, both incomplete.** `0004`/`0010` use `> Current placement` notes (matches the gADR-0000 dated-clarification precedent) while `0014`–`0018` rewrite paths in-body. The rewrite is then unfinished: `0004:51,112` and `0010:36,121` still say `main.tscn`; `0019:62` still says `assets/**`; `0010` has both states in one file (`:105` updated, `:121` not). Pick one mechanism and finish it.
3. **`docs/project-manifest-notes.md` drops content it is the sole authority for.** Its own charter is "update this doc when the rationale changes". The rewrite deletes the collision-layer enumeration (Platform=`terrain`(1), Obstacle, Pickup=`pickup`(6) masking `player`(2)) and the Obstacle gravity-prop rationale, replacing them with "Projectiles and areas retain their established masks" — a pointer to a record that no longer exists.

### Baseline smells (judgement calls)

4. **Duplicated Code / Data Clumps** — five test files re-create the config-object API the Systems just lost: `test_boss_logic.gd:56,60,78`, `test_combat_logic.gd:55`, `test_enemies_logic.gd:76`, `test_items_logic.gd:48`, `test_player_logic.gd:37` (the last is literally the deleted `_compute_velocity(..., config: PlayerConfigScript, ...)` signature). gADR-0020's "controllers pass the scalar values a rule needs" endorses the Content→Systems seam, not five duplicated unpack adapters.
5. **Speculative Generality / inconsistent guard** — `ui/game_shell.gd:15-19`: `has_method(...)` duck-typing guards the bind, but line 18's `gameplay.run_ended.connect(...)` is unguarded — if the guard could ever be false, line 18 hard-errors. The scene is statically composed; drop the duck-typing.
6. **Duplicated Code** — `tests/test_modular_dependencies.py:31-34` vs `:48-51`: identical root walk + suffix filter; extract a `_runtime_files()` helper.
7. **Minor** — `ui/hud_controller.gd:108`: `_player == null or not is_instance_valid(_player)` is redundant. `tools/editor/debug_palette.gd:198-207`: Message Chain `get_play_instance() → get_node_or_null("Gameplay") → player_node()`.

Commit/PR typing is clean (`refactor(panda-adventure)` correctly non-releasing); no stale `res://src/` or `res://data/` references survive at head.

## Spec (vs issue #621)

Overall: substantially conformant — all 10 acceptance criteria are addressed. Findings are gaps and two design misfits, not a failed refactor.

1. **Content reloads the UI composition root, contradicting the guard's own rule.** `content/controllers/game_flow_director.gd:73` — `_level.get_tree().reload_current_scene()`. With `run/main_scene="res://ui/game_shell.tscn"` (`project.godot:14`), Retry makes a Content module tear down and reload the UI root. Spec: *"Production dependencies point only within the same root or downward"*; gADR-0020: *"lower modules do not locate or load higher modules"*. Invisible to the guard (no `res://` literal). UI-owned retry, or a shell-scoped reset, would honor the direction.
2. **Guard covers only `res://` literals.** `tests/test_modular_dependencies.py:20,28-30` — GDScript `class_name` globals (22 in `content/`, 3 in `ui/`) and `uid://` refs are invisible; nothing mechanically stops `systems/*.gd` from naming `HudController`. Verified: **no live violation exists today**, so this is coverage debt. There is also no negative self-test proving the guard fails on an injected violation (*"A mechanical test rejects upward or cross-layer runtime dependencies"*).
3. **Silent duck-typed wiring in the shell.** `ui/game_shell.gd:15-17` — a rename silently yields an unbound HUD (only a `player_bound:false` log field), while `run_ended`/`retry_requested` at `:18-19` are connected unguarded. Same site as Standards #5.
4. **No mechanical evidence for export contents.** `export_presets.cfg:11` (`exclude_filter="tools/editor/*,tests/*"`) plus the source-level check at `tests/test_modular_dependencies.py:44` cover the stated transitive-inclusion risk, but nothing asserts the built pack. Brief: *"Treat … exported pack contents as architectural evidence rather than relying on directory names alone"* — only a manual claim backs it.
5. **Minor: rationale lost in the movement extraction.** `systems/player_movement_system.gd:1-36` is a faithful extraction (formulas identical), but the gADR-0009 explanation of `gravity * time_scale * time_scale` was dropped with the old `player_controller.gd` block. Behavior preserved; the "why" is gone.

**Verified clean:** zero `res://src|scenes|assets|data` hits repo-wide; Player group fully removed with injection in place (`content/controllers/level_controller.gd:101,176,283`, `ui/hud_controller.gd:45`); capability groups `gravity_affectable`/`time_dilatable` kept; Systems take scalars with no Content/UI refs; `docs/gadr/0020-modular-runtime-architecture.md` present; `.gitattributes`, `style.json`, `targets.json`, and docs all migrated.

## Entropy

**Verdict: Keep** — with three ~5-line simplifications and one signature decision. The implementation stayed inside the design-time entropy-reviewed plan: no event bus, service locator, interface hierarchy, or sandbox; the composition root is 20 lines, the guard 65, everything else is a move plus path rewrite. Nothing the plan's hard constraints require was cut.

1. **Defensive duck-typing in the composition root** — `ui/game_shell.gd:15-19` guards a composition authored in the same scene (`ui/game_shell.tscn:3-6`); `player_bound` exists only to fill a log field. A missing method is a wiring bug that should fail loudly at boot, not degrade to a HUD that silently never updates. **Simplify:** call `hud.bind(gameplay.player_node())` directly, drop `player_bound`.
2. **`_shown` re-represents the End-state latch** — `ui/end_screen_controller.gd:37,75,89-90`. `GameFlowDirector.retry()` already gates on the pure latch (`content/controllers/game_flow_director.gd:69-70`); `tests/test_level_e2e.py:252-256` passes on the Content gate alone. **Simplify:** one state, one owner.
3. **Third guard test is half-duplicate** — `tests/test_modular_dependencies.py:60-65`: `assert "res://ui/" not in gameplay` is what test 1 already reports (content rank 2 < ui rank 3). Keep the `run/main_scene` assertion (the only mechanical statement that the shell is the entry point); drop the redundant one.
4. **Movement extraction is where scalarization cost more than it bought** — `systems/player_movement_system.gd:9-20` takes 10 positional params (4 same-typed config floats), has exactly one caller (`content/controllers/player_controller.gd:184`), and `tests/gdscript/test_player_logic.gd:37` had to re-add a config-shaped adapter. Contrast combat/AI/warp, where the same change *deleted* fixture helpers from `tests/gdscript/make_balancing_fixtures.gd` — there scalars matched the consumers and net-reduced code. **Simplify (either):** take an untyped `config` param — the project's own idiom (`game_flow_director.gd:26-28`), guard stays green — or leave both pure functions in Content, since only the Player consumes them. Worth settling before further Systems adopt the scalar style. (Same site as Standards #4 / Spec #5.)
5. **Optional** — `scripts/build_config.py:789` keeps `_ASSETS_ROOT = "content/assets"` while `tools/panda_assets/style.json:4` declares the same fact; reading `config.assets_root` (as line 1138 now does) collapses it to one authority.

**Essential complexity to keep:** the four roots and the path-scan guard; the `run_ended` level→shell relay (`content/controllers/level_controller.gd:290-291`) as the narrow upward signal; `enemy.setup(kind, _player)` injection; the single `tools/editor/*` export exclude — sound precisely because guard test 2 proves no runtime file references it; gADR-0020.

**Assumptions worth testing cheaply:** (a) no `tools/editor` resource survives transitively into the export — one macOS export, list the `.pck` contents; (b) the guard actually fails on a real violation — inject an upward preload locally, run, revert (probe, don't commit); (c) HUD rebinds after a retry reload — already covered by `tests/test_level_e2e.py:296,389`.

---

**Summary.** Standards: 7 findings, worst is the glossary gap (#1). Spec: 5 findings, worst is the Content-reloads-UI-root direction violation (#1), which the guard mechanically cannot see (#2). Entropy: Keep, with the movement-system signature (#4) as the one decision worth settling now. One site is hit by all three axes: the `ui/game_shell.gd:15-19` duck-typed wiring — fixing it once resolves Standards #5, Spec #3, and Entropy #1.

All paths are relative to `examples/platformer/panda-adventure/`; line numbers refer to PR head 84f6736.

qinhaihong-red:
Addressed the review at exact head `13f701a`.

I de-duplicated findings that pointed to the same site and handled them as follows.

### Fully adopted

- Added glossary entries for Add-ons, Systems, Content, UI, Gameplay, and Game Shell.
- Finished the gADR migration consistently. gADR-0004 and gADR-0010 now describe the current paths and ownership directly; the mixed `Current placement` notes were removed. The remaining stale `main.tscn` and `assets/**` paths were corrected.
- Restored the manifest's concrete collision topology and Obstacle rationale, including the existing `time_field` layer 7 and its `player|projectile` mask.
- Replaced duck-typed shell wiring with direct calls. The shell now verifies that Gameplay returned a valid Player before binding, connecting signals, or emitting `game_shell_ready`.
- Kept Retry eligibility and logging in Content, but moved `reload_current_scene()` to the UI-owned Game Shell. Content returns whether the intent was accepted.
- Expanded the dependency guard to cover `res://` paths, global `class_name` references, and unresolved `uid://` loads. It now has a negative fixture for every supported violation and allowed fixtures proving that class names in comments and strings do not false-positive.
- Extracted the shared runtime-file walk, removed the duplicate Gameplay/UI assertion, removed the redundant HUD null clause, and restored the time-dilation gravity rationale.

### Partially adopted

- Removed EndScreen's duplicate `_shown` state, but retained `visible` as the UI-local emission guard. Content's End-state gate remains authoritative; a hidden surface should still not emit retry intent.
- Removed the defensive `player_bound` control variable, but retained the structured `player_bound:true` log field after successful Player validation to preserve the existing logging contract.
- Verified the built pack mechanically rather than adding a permanent PCK parser to the test suite. The exact exported artifact reports:
  - `res://ui/game_shell.tscn=true`
  - `res://content/scenes/gameplay.tscn=true`
  - `res://tools/editor/editor.tscn=false`
  - `res://tools/editor/editor_controller.gd=false`

### Not adopted

- Kept the scalar Systems APIs and test-side adapters. Passing an untyped Content config object would still be a conceptual Systems -> Content dependency; moving the pure rules into Content would contradict issue #621's ownership boundary. The local test helpers avoid repeating scalar unpacking at every assertion.
- Kept the debug palette helpers unchanged. The reported chain is already encapsulated behind `_play_root()` and `_find_player()`.
- Kept `build_config.py`'s `_ASSETS_ROOT`. It is the build/manifest root, while `style.json.assets_root` is input to the separate asset-pipeline configuration; making the builder parse that tool config would couple the two pipelines.

### Verification

- Full Panda Adventure suite: `549 passed, 4 skipped` (the skips require the optional live-assets group or a macOS window-server session).
- Final targeted guard and plain-boot checks: `5 passed`; Retry E2E: `2 passed`.
- Ruff check/format and `git diff --check`: passed.
- Godot `4.6.3-stable`; release macOS export succeeded.
- Exported application booted and emitted `game_shell_ready {"player_bound":true}` and `hud_ready`.
- Independent subagent recheck: PASS, no remaining findings.
- Exact-head required CI: all passing; conditional Godot e2e jobs skipped by workflow policy.

Commit: `13f701a1a7d651afc880a1da0605e93078f390f7`

qinhaihong-red:
Recheck of the fix commit `13f701a` against the review above and the author's finding→resolution mapping. Every verdict below was verified against the source at `13f701a`; paths are relative to `examples/platformer/panda-adventure/`.

## Per-finding verdicts

| Finding | Verdict | Evidence @ 13f701a |
|---|---|---|
| S1 glossary | **verified-fixed** | `GAME-CONTEXT.md:579-615` — 6 entries under `### Runtime architecture`; wording matches `gadr/0020:28-35`; no clash with existing entries. |
| S2 gADR paths | **fixed-with-caveat** | Zero `main.tscn` / `assets/**` hits repo-wide; one mechanism now — but the mechanism chosen is body rewriting, see **N1**. |
| S3 manifest-notes | **verified-fixed** | `docs/project-manifest-notes.md:33-40,74-88`. All layer claims verified against `project.godot:82-88` and all seven scenes — including `time_field` 64/10 = layer 7 / `player\|projectile`. All correct. |
| S4 = E4 scalar Systems API | **rationale-accepted** (with note) | #621 puts rules in Systems and forbids Systems→Content config, so both proposed alternatives are indeed foreclosed. Note: a Systems-owned param type is a third option not weighed — `systems/stats_config.gd:1` is an existing precedent. |
| S5 = P3 = E1 shell wiring | **verified-fixed** | `ui/game_shell.gd:12-23` — direct `player_node()`/`bind()`, `is_instance_valid` + `push_error`; both `has_method` guards gone. |
| S6 duplicated walk | **verified-fixed** | `tests/test_modular_dependencies.py:26-32` `_runtime_files()`, used at `:74` and `:118`. |
| S7a hud null clause | **verified-fixed** | `ui/hud_controller.gd:108`. |
| S7b debug palette | **rationale-accepted** | Two named helpers with one guard each; dev-only. |
| P1 retry direction | **verified-fixed** | `game_flow_director.gd:69-74` returns `bool`, no `get_tree()`; `level_controller.gd:286-287` passthrough; `game_shell.gd:26-29` owns the reload. Upward path stays the narrow `retry_requested` signal. |
| P2 guard coverage | **fixed-with-caveat** | `test_modular_dependencies.py:36-102` — comment/string blanking, `\b` matching, correct class→layer map; real negative/allowed fixtures at `:131-166`. Caveats in **N3**/**N6**. |
| P4 export-pack evidence | **not-fixed** (rationale weak) | No permanent artifact added; the claimed pack listing is a one-off manual claim, unverifiable from the repo. `export_presets.cfg:11` remains the only mechanical statement. |
| P5 movement rationale | **fixed-with-caveat** | `systems/player_movement_system.gd:7-11` restores the v²/2g argument, but drops the `(gADR-0009)` citation and the +Y-down sign note present at `b0cb7d0:src/controllers/player_controller.gd:130-140`. |
| E2 `_shown` | **verified-fixed** | `ui/end_screen_controller.gd:36,73,88` — `_shown` gone; `visible` false in `_ready:46`, true only in `show_end`. The retry-dead-mid-run e2e contract (`tests/test_level_e2e.py:252-256`) still holds. |
| E3 duplicate guard assert | **verified-fixed** | `test_modular_dependencies.py:125-129`; the dropped assertion is subsumed by `_dependency_violations`. |
| E5 `_ASSETS_ROOT` | **rationale-disputed** | `build_config.py:1131-1138` already imports `assets.config.load_style_config` and consumes `config.assets_root`, so "would couple the two pipelines" is already the status quo in that file. Still cosmetic — not blocking. |

## New findings introduced by the fix commit

- **N1 (main)** — `gadr/0004` and `gadr/0010` had their bodies rewritten and the dated notes deleted. This domain's precedent is uniformly the opposite (`0000:37`, `0001:68`, `0003:109`, `0007:83,91`, `0014:15,23`, `0015:78,97,110` all preserve the body and add a dated Amendment/Outcome note). `gadr/0010:54-58,63-67` now attributes gADR-0020's shell/Content retry split to S9's original decision and silently drops the "non-CanvasLayer child of Main" qualifier — that is decision history being rewritten, not stale-path cleanup. S2 asked for one mechanism; this picked the one the domain has never used.
- **N2** — `gadr/0004:108-110` still cites `docs/project-manifest-notes.md` for "the sanctioned fallback for this file", but the notes' new `## ui/game_shell.tscn` section (`:58-65`) dropped the hand-edit / gda-gap paragraph (`b0cb7d0:docs/project-manifest-notes.md:87-91`). Dangling reference.
- **N3** — Guard asymmetry: `_runtime_references` (`:33`) scans raw text, so a `res://ui/...` inside a systems *comment* would fail, while the class-name check strips comments; and a `uid://` hit (`:88`) is reported under the "upward runtime dependencies" assert message (`:108`) though it isn't a direction violation.
- **N4** — `game_shell.gd:23` logs `{"player_bound": true}` unconditionally (the failure path returns at `:22`), so the retained field is a constant; `tests/test_game_log_fallback.py:59` now pins that literal. The preserved "logging contract" carries no information.
- **N5** — The early return at `game_shell.gd:20-22` also skips `run_ended`/`retry_requested` wiring, so a Player-less run now has no End screen at all (previously still wired). Defensible, but undocumented.
- **N6** — `test_runtime_does_not_depend_on_editor_tools` (`:116-123`) has no negative fixture; the reply's "a negative fixture for every supported violation" overstates.
- **N7** — `game_flow_director.gd:104-105` still justifies keeping the level processing "so the explicit retry entry point survives"; retry now enters from the never-frozen shell sibling, so the comment's reason is stale.

## Overall

Substantively the fix commit is good: every code-level finding (S5/P1/P3/E1/E2/E3/S6/S7a) is genuinely resolved in source, the collision-topology restoration is factually correct against `project.godot` and all seven scenes, the new guard is a real mechanical check with real negative fixtures, and the e2e retry contract survives the `visible` gate. The not-adopted rationales are mostly sound — the Systems/Content one is directly supported by #621's text; only E5's is factually wrong, and P4 left free evidence on the table. Required CI is green at `13f701a` and the PR is mergeable.

**Recommendation: one more doc-only round on N1/N2** — restore the two gADR bodies, surface the current state via dated `> Outcome (…, gADR-0020)` notes per domain precedent, and restore (or re-target) the manifest-notes hand-edit paragraph that `gadr/0004:108-110` still cites — **then merge. N3–N7 are follow-up material, not blockers.**

qinhaihong-red:
Addressed the latest follow-up review in `96be4f9`.

### Fully adopted

- **N1 — preserve gADR history:** restored the original bodies of gADR-0004 and gADR-0010 and added dated `Outcome (2026-08-05, gADR-0020)` notes describing the modular relocation.
- **N2 — restore the cited manifest note:** documented the sanctioned hand-authored `ext_resource` / `instance=` fallback for `Gameplay`, `Hud`, and `EndScreen`.
- **N3 — make dependency guards code-aware:** GDScript comments are stripped before `res://`, `uid://`, and editor-tool checks; the diagnostic now says `runtime dependency violations`; fixtures cover commented references.
- **N4 — remove the constant readiness field:** `game_shell_ready` now emits `{}` and the fallback test asserts the meaningful record only.
- **N5 — document invalid-Player behavior:** `GameShell` now states that a missing Player is a composition error and aborts all shell wiring instead of leaving a partial UI.
- **N6 — add the missing negative editor-tool fixture:** a real runtime preload of `res://tools/editor/...` is now rejected.
- **N7 — remove the stale retry explanation:** the comment now describes shell-owned retry submission through Content.
- **P5 — complete the movement rationale:** documented Godot's +Y-down signs and cited gADR-0009 for full slow motion.
- **E5 — eliminate the duplicate asset-root authority:** the prior explanation was incorrect. `build_config.py` already shares the Asset pipeline configuration, so the hard-coded `_ASSETS_ROOT` was a real split-brain risk. Both manifest consumers now read `StyleConfig.assets_root`; an independent re-review passed, including 54 asset-pipeline tests.

### Not adopted

- **Permanent PCK parser for P4:** the issue requires a real release export, not a new projectless pack-inspection harness. I retained the real-artifact acceptance check and reran it at this head: the macOS release export succeeded and the exported application booted headlessly on Godot 4.6.3 with exit 0 and `game_shell_ready {}`. The residual limitation is explicit: CI does not inspect the PCK file table.
- **Move the one-call application parameter type into Systems:** there is still one Content caller and no current reusable settings aggregate. Moving it now would add ownership without a demonstrated second consumer.

### Verification at exact head `96be4f9`

- Panda Adventure suite: **550 passed, 4 skipped** (only unavailable window-server/live-assets cases).
- Targeted asset/dependency/log tests: **60 passed**.
- `gda project info`: **Godot 4.6.3**, main scene `res://ui/game_shell.tscn`.
- `gda export run --mode release`: succeeded with no warnings; exported app boot: exit 0.
- GitHub CI: all required checks passed; PR remains mergeable.

PR #626, opened 2026-08-05
refactor(gda-balancing): adopt dependency-directed implementation layers

### LINKED ISSUE #624: [gda-balancing] Refactor modules into a layered, dependency-directed architecture
## Problem

`gda-balancing` has grown to roughly 40,000 lines of production Python, but its implementation modules no longer expose clear ownership or dependency boundaries. The problem is not a lack of directories: a few modules combine several unrelated reasons to change.

Current evidence includes:

- `schema2.model` combines Model Source checking, resolution, lowering, Resolved Model admission, artifact inspection, filesystem publication, and recovery.
- `schema2.bootstrap` combines authority admission, language-rule execution, package closure, runtime contracts, and conformance-vector validation.
- `schema2.experiment` combines Experiment admission, deterministic Runtime execution, replay, Evidence production, terminal audit, and artifact-set validation.
- Command modules contain Template semantics and large Standard Schema projections instead of remaining presentation adapters.
- Domain-oriented modules import command descriptors, CLI envelopes, and path-handling concerns.
- The internal import graph contains a command-registry cycle.

This makes the system harder to understand, increases change amplification, and makes the next Standard Schema, Runtime, Experiment, and Template capabilities more expensive to add safely.

The product requirements remain owned by #501. The Standard Schema authority chain, domain vocabulary, and accepted decisions remain owned by the balancing context, architecture document, and bADRs. This issue changes implementation organization only.

## What to build

Refactor the package into a dependency-directed layered architecture, using the ownership and one-way dependency principles from the project modular-architecture guidance and the core ideas of Domain-Driven Design.

Layer order from lowest to highest:

1. **Infrastructure** — domain-neutral technical mechanisms such as bounded input reading, stable JSON byte encoding, package-resource access, file locking, and atomic filesystem operations.
2. **Domain** — Standard Schema authority admission, Formula semantics, Model compilation, Runtime, Experiment and Evidence semantics, Template semantics, artifact identity, and publication rules.
3. **Application** — end-to-end use-case orchestration for the public commands.
4. **UI / Interfaces** — CLI argument binding, descriptors, registry, help, response schemas, stdout/stderr rendering, exit codes, and refusal-envelope mapping.

The allowed dependency direction is `interfaces.cli -> application -> domain -> infrastructure`. Elements may depend only on the same layer or a lower layer, and same-layer dependencies must remain acyclic. The CLI composition root may assemble lower-layer components but must contain no business rules.

Use Domain-Driven Design to align modules with the existing ubiquitous language and reasons to change. Do not introduce additional bounded contexts, generic `common` or `shared` modules, blanket Repository interfaces, a DI container, a service locator, or an architectural event bus. Typed results and refusals are sufficient for current upward observation. Runtime `Event` remains domain data, not cross-layer messaging.

Preserve these ownership boundaries:

- Infrastructure may implement generic canonical byte operations, but Standard Schema member selection, ordering, identity preimages, and normative canonicalization policy remain Domain concerns.
- Authority JSON is Domain-owned content; reading packaged resources is Infrastructure.
- CLI success/error response schemas belong to the UI layer; Standard Schema semantics, admission, and authority projections do not.
- Standard Schema 1.x is a migration input protocol under the migration application boundary; active 2.x Domain code must not depend on it.
- Artifact publication rules are separate from atomic filesystem operations. Introduce a storage port only if a real second implementation or substitution requirement appears.
- Preserve one production authority-lifecycle owner, one admitted packaged context, and the existing process-scoped admission/cache behavior.

### Internal execution order

Keep the issue open and execute the following steps in dependency order. Each step must leave the public CLI and test suite working; do not create child tracking issues for these steps.

- [x] Record the accepted layering and migration constraints in the next available bADR, characterize the public behavior baseline, and identify the current ownership of affected responsibilities.
- [x] Establish the first complete four-layer tracer through `package list`, including a small AST import-direction check for modules already moved into the new layers.
- [x] Migrate the remaining read-only authority commands: `package get`, `schema get`, `version`, and `manifest`.
- [x] Migrate `formula render` and `formula parse`, preserving the canonical `body`/`expression` round trip.
- [x] Migrate `model check`, then separate Model resolution/lowering and `model build` orchestration.
- [x] Separate artifact publication/recovery from Model semantics without introducing speculative storage abstractions.
- [x] Migrate `model inspect` and explanation/artifact validation.
- [x] Migrate `experiment check` and Experiment admission.
- [x] Separate deterministic Runtime scheduling and program execution while keeping `experiment run` complete.
- [x] Separate replay, Evidence, terminal-audit, and experiment artifact-set validation.
- [x] Migrate Template catalog retrieval, then Template admission and instantiation.
- [x] Isolate Standard Schema 1.x migration and remove its dependency on CLI envelopes.
- [x] Move the remaining CLI composition, descriptor, dispatch, and rendering responsibilities into the UI boundary; eliminate residual cycles and remove obsolete internal entry points.
- [x] Update the architecture document to describe the implemented structure and complete the final behavior, type, import-direction, packaging, and source/wheel verification.

Do not predeclare a permanent dependency graph between candidate Domain modules such as authority, formula, model, runtime, experiment/evidence, and template. Extract cohesive clusters according to actual ownership, then verify the resulting import graph after each tracer. Avoid catch-all artifact or diagnostics modules.

### Validation cadence

- Run focused tests after each small change.
- Run the complete package test suite and pyright before completing each coherent migration stage.
- Run source/wheel parity only when packaged resources, entry points, or artifact publication are affected.
- Compare public behavior: stdout/stderr, exit codes, diagnostic content and ordering, canonical bytes and identities, publication receipts and recovery, Formula round trips, and deterministic Runtime results.
- Preserve existing tests without unexplained coverage reduction; do not turn the current collected-test count into a permanent numeric gate.
- Treat any behavior difference as a separate requirement change, not as part of this refactor.

## Acceptance criteria

- [x] Production modules follow the four-layer dependency direction with no upward imports or cross-layer strongly connected components.
- [x] Domain code does not import command handlers, CLI descriptors, CLI envelopes, or presentation schema code.
- [x] CLI and command modules contain no Formula, Model, Runtime, Experiment, Evidence, or Template business semantics.
- [x] The command-registry cycle is removed and one UI-owned descriptor registry remains the source for dispatch and manifest projection.
- [x] Kernel/LDB authority, structural-versus-semantic validation, Formula pairing, artifact identities, atomic publication, Runtime determinism, and all public CLI contracts remain unchanged.
- [x] Packaged authority loading preserves its single production owner and existing admission/cache lifecycle.
- [x] Standard Schema 1.x remains available only for the accepted source-migration contract and is isolated from the active 2.x Domain.
- [x] No global event bus, DI container, service locator, blanket Repository layer, catch-all shared module, duplicate authority, or permanent dual module layout is introduced.
- [x] Existing behavioral tests pass with no unexplained coverage reduction; pyright passes; relevant packaged-resource and source/wheel parity checks pass.
- [x] The bADR records the durable architecture decision and the architecture document describes the final implementation; no third permanent module-map authority is created.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Summary

- record bADR-0025 and the implemented four-layer architecture
- migrate public command flows through Infrastructure, Domain, Application, and CLI boundaries while removing obsolete cycles and entry points
- isolate Standard Schema 1.x to source migration and preserve all public Schema 2.x behavior
- add AST dependency checks and bounded source ingress without speculative storage, DI, event-bus, or shared-module mechanisms

## Validation

- inventory closure: 1,428 tests and 194 package vectors; no missing, overlap, or uncovered entries
- full seven-shard suite: 1,401 passed and 27 declared skips
- all shard outcome-closure checks passed with no unexpected skips or xfails
- Ruff check and format, Pyright, source/wheel smoke, sdist, and wheel build passed
- independent Spec, Standards, and architecture reviews completed; the latest review found no documented-standard violations or missing acceptance criteria

Closes #624


qinhaihong-red:
Implementation review is complete at exact head 2201e2c.

- Spec review: PASS. The only finding was an oversized Experiment identity regression; 2201e2c restores byte-identical diagnostic output while keeping memory bounded through streaming SHA-256.
- Standards review: PASS. The layered boundaries, bounded ingress, CLI schema ownership, Schema 1.x isolation, and Model dependency ownership are closed without additional framework mechanisms.
- Local final validation: all 7 shards passed (1,401 passed, 27 declared skips), inventory closure covers 1,428 tests and 194 package vectors, all outcome checks passed, and Ruff, Pyright, source/wheel smoke, sdist, and wheel build passed.
- GitHub required checks are green.

qinhaihong-red:
## Review — two-axis (Spec / Standards) + entropy review @ head `2201e2c`

Three independent review passes over `origin/main...2201e2c` (144 files, 41 commits): a Spec axis against #624, a Standards axis against bADR-0025 / ARCHITECTURE.md / repo conventions, and an entropy review. All verification below is structural (source/diff/AST inspection); the test suite and pyright were not re-run by the reviewers.

**Overall**: the refactor's premise is evidenced, not asserted — the `commands` registry cycle was reproduced on `origin/main` (Tarjan over the AST import graph), the upward `schema2/model.py` → envelope/descriptors edges are literal, and the declined mechanisms (DI, event bus, Repository, `common`/`shared`, compatibility shims) are genuinely absent. Entropy verdict: **Keep**. However, the Spec axis found three acceptance criteria that are not fully met as checked, and the Standards axis found a violation of this PR's own bADR.

### Spec findings (vs #624)

1. **Schema 1.x isolation landed in the wrong layer.** Spec: *"Standard Schema 1.x is a migration input protocol under the migration **application** boundary; active 2.x Domain code must not depend on it."* The consumer is `domain/migration.py` (Domain), and the gate licenses it: `_SCHEMA1_MIGRATION_CONSUMERS = frozenset({"gda_balancing.domain.migration"})` in `tests/test_layer_dependencies.py`. The leak is also bidirectional — `schema/funnel/__init__.py` imports `gda_balancing.domain.errors`, and `_architectural_modules()` never scans `schema/`, so no upward-import rule applies to that tree. The checked acceptance item *"isolated from the active 2.x Domain"* does not hold as stated; either complete the migration-boundary move or explicitly re-scope the criterion on #624.

2. **Experiment semantics remain in the UI layer.** Spec: *"CLI and command modules contain no … Experiment … business semantics."* `interfaces/cli/experiment_fixtures.py` (169 lines, shipped in `src/`) authors a full Experiment Specification — walks RIR entrypoints/operations, calls `domain.experiment.derive_scenario_program_requirements`, composes scenarios/metrics/acceptance. It was moved from `commands/experiment.py`, not migrated.

3. **Model resolution/lowering not actually separated.** Spec step: *"separate Model resolution/lowering and `model build` orchestration."* `domain/model/resolution.py` is still 7,258 lines (was 9,560); `domain/model/checking.py` is a 356-line facade importing 17 private names (`_model_lowering`, `_model_check_diagnostics`, `_resolution_profile`, …) from it. Publication/inspection/compilation did split out; resolution vs lowering did not — this is the "split only the largest files" shape that bADR-0025's own rejected-options section warns against.

4. **A dispatch contract was deleted inside a behavior-preserving refactor.** Spec: *"Treat any behavior difference as a separate requirement change."* `dispatch._invoke_descriptor` no longer renders a 1.x `RefusalReport` as a refusal envelope / exit 2; the replacement test asserts exit 4 `internal_error`. Unreachable in production (all 16 descriptors are major 2), so no user-visible change — but it should be recorded as its own requirement change, not absorbed silently.

### Standards findings

1. **`domain/authority/validation_support.py` (new, 318 lines) violates bADR-0025.** The bADR prohibits generic catch-all modules ("Generic `common`, `shared`, catch-all … modules are prohibited because they hide ownership"); `validation_support` is that module under a synonym — four siblings import 11 private helpers from it, and "support" names no owner.

2. **CLI response-schema projection lives in Domain.** bADR-0025 assigns response-schema projection to UI/Interfaces, yet `domain/authority/package_projection.py` defines `package_release_success_schema()` / `package_vector_set_success_schema()`, consumed as `CommandDescriptor.success_schema` by `interfaces/cli/package.py` and `schema.py`.

3. **Infrastructure doesn't own what ARCHITECTURE §4.1 claims.** "`infrastructure` owns … package-resource access", but `domain/authority/context.py` still imports `importlib.resources.files` and injects it; the infra function is a one-line pass-through.

Judgement-call smells (not hard violations): 115 private-name imports across module boundaries (worst: `evidence.py` ← `runtime.execution` ×32, `model/checking.py` ← `model.resolution` ×24 — no narrowed interface formed); the four-line authority prelude duplicated verbatim across `application/{package_list,package_get,schema_get,version}.py`; the `publish_artifact_set(…)` parameter clump repeated 3×; `infrastructure/distribution.py` as a three-line delegate; `tests/_legacy_design_adapters.run_legacy_cli` re-implementing envelopes/exit codes instead of driving production `dispatch()` (defensible since those commands are retired, but it is a test-authored fake as sole evidence).

### Gate brittleness (Spec + entropy, independently converged)

`tests/test_layer_dependencies.py` contains two snapshot assertions posing as architecture rules:

- `test_cross_domain_refusal_catalogs_are_not_model_owned` hand-lists four CLI file paths — a **new** `interfaces/cli/*.py` importing `domain.model.resolution` passes silently, renaming any listed file raises `FileNotFoundError`, and meanwhile `template_catalog.py` / `template_instantiation.py` already import `MODEL_REFUSAL_CATALOG` from `domain.model.resolution`, the very coupling the test names as prohibited. It passes without enforcing its own invariant. → Restate as a universal rule over `interfaces/cli/*`.
- `test_model_checking_does_not_depend_on_model_compilation` pins one internal edge that #624 never names; the acyclicity gate already prevents the cycle it fears. → Delete.

The four generic gates (`…do_not_import_upward`, `…are_acyclic`, `…one_declared_owner`, `…schema1…migration_boundary`) are the keepers — each states a bADR rule and generalizes to unseen files.

### Entropy: simplify suggestions (none block merging)

1. Generalize/delete the two snapshot gates above.
2. The oversized-input identity repair (`2201e2c`) plumbs the digest through an optional `InputTooLargeError.sha256` plus a `RuntimeError` guarding a state the same function makes impossible — return the digest in a typed result and drop the guard; record the `eb0540d` ingress-bounding move as its own change note on #624 (the bound itself is LDB-authored and pre-existing, not speculation).
3. Collapse the duplicated authority prelude into one `require_admitted_context(provider)`; call `importlib.metadata.version` directly instead of via `infrastructure/distribution.py`; drop the redundant `test_package_list_cli_adapter_cold_imports…` subprocess test (subsumed by the composition-root one; both `assert stderr == ""`, so any unrelated warning fails CI).

### Confirmed correct

Entry points moved (`pyproject` → `interfaces.cli.main:main`, `__main__.py`); old `commands/` and root `cli.py`/`envelope.py`/`emit.py`/`descriptors.py`/`dispatch.py`/`path_contracts.py` deleted; registry cycle broken via lazy `_live_registry`; no dynamic imports evading the AST gate; bounded ingress preserves the full-stream SHA-256 identity at head; identical 16-descriptor surface and order; `schema2/` reduced to packaged authority JSON; conventional-commit types and the in-directory `refactor(gda-balancing)` title comply with `libs/gda-balancing/AGENTS.md`; `BALANCING-CONTEXT.md` untouched. The per-command `application/*` + `interfaces/cli/*` pairs are *not* flagged as indirection: adding a command now touches 3 small files instead of appending to 1,380–2,106-line modules — change amplification fell.

---
*Reviews ran as three independent sub-agents (Spec, Standards, entropy) over the full diff; findings above are deduplicated across them.*

qinhaihong-red:
Addressed the latest review at exact head `2c7538c` (commits `c8646be`, `2c7538c`).

### Fully adopted

- Moved the Standard Schema 1.x converter from Domain to the migration Application boundary. The legacy `schema/` tree no longer imports active layers; the architecture gate now enforces both the sole consumer and the reverse isolation rule.
- Moved Experiment Specification fixture authoring into Application. UI now only binds the public Model-build descriptor identity.
- Split Model responsibilities into `resolution.py`, `lowering.py`, and `admission.py`, with build orchestration remaining in Application. No compatibility facade was added.
- Made package-resource access truthfully Infrastructure-owned, including the `importlib.resources.files` call.
- Moved the migration's regular-file open/fstat/bounded-read/hash mechanism into Infrastructure while retaining migration limits, identity prefix, and error mapping in Application.
- Consolidated the repeated Application authority-admission prelude and renamed the cohesive Authority contract validator from `validation_support` to `contract_validation`.
- Removed the two implementation-snapshot architecture tests and the redundant package-list cold-import test; retained/generalized only contract-derived ownership, direction, cycle, composition-root, and Schema 1.x isolation gates.

### Partially adopted

- Package Release/vector schemas were renamed to remove the misleading `success` suffix, but remain Domain-owned authority-member projections. UI still owns the enclosing success-response schemas, as required by #624.
- Hashed bounded input now returns a typed observation (`data | None` plus the complete SHA-256) instead of optional exception state and an impossible guard. No separate protocol/requirement note was added because the existing ingress bound and public refusal remain unchanged.
- Cross-module private imports were treated as a judgement smell, not a numeric architecture invariant. The Model ownership split reduced the mixed responsibility; no private-import gate or speculative interface layer was introduced.

### Not adopted

- Did not restore the synthetic Standard Schema 1.x `RefusalReport` branch in schema-major-2 dispatch or record it as a public behavior change. All production descriptors are schema-major 2 and that injected value was an unreachable internal seam removed with the obsolete entry point.
- Did not move authority projections into UI, bypass the Infrastructure distribution boundary, or introduce broad parameter-clump/private-import abstractions without a concrete current requirement.

### Verification

- Local full suite after the main fix: **1399 passed, 27 skipped** (1426 collected).
- Final focused follow-up: **35 migration/layer tests** and **77 authority tests** passed.
- Ruff lint/format and pyright passed.
- All exact-head PR checks are green, including all gda-balancing shards and wheel/subprocess smoke.
- Independent two-axis review: Spec found no actionable issue; Standards found the migration I/O ownership gap above, which was fixed, then passed the exact-head recheck with no residual findings.

qinhaihong-red:
## Re-review @ head `2c7538c` — resolutions verified

Re-checked every claimed resolution against the actual source at exact head `2c7538c` (independent source/AST verification of the `c8646be` + `2c7538c` delta), plus CI status. **Outcome: the review passes — all hard violations and acceptance gaps from the previous review are genuinely fixed, no regressions found.** CI is green at exact head (only the schedule/dispatch-only Godot e2e and nightly suites skipped, as expected).

### Per-claim verdicts

- **1.x isolation — VERIFIED.** Sole importer of `gda_balancing.schema.*` is now `application/migration.py`; the `schema/` tree imports nothing outside itself (`funnel/__init__.py` defines its own `SchemaInputReadError` instead of importing `domain.errors`); the new gate rglobs `schema/` for reverse imports and the sole-consumer whitelist moved to the Application boundary.
- **Experiment fixtures — VERIFIED.** `interfaces/cli/experiment_fixtures.py` is a 30-line binding; the 185-line authoring lives in `application/experiment_fixtures.py`, the sole caller of `derive_scenario_program_requirements`.
- **Model split — VERIFIED.** `resolution.py` 7,258 → 1,071 lines with `lowering.py` (4,369) and `admission.py` (1,914) as real owners; consumers repointed; no re-export facade; `checking.py` is genuine orchestration.
- **Package-resource access — VERIFIED.** `importlib.resources` exists in exactly one module: `infrastructure/package_resources.py`; the injection parameter is gone.
- **Migration input IO — VERIFIED.** Mechanism (open/fstat/S_ISREG/bounded read/hash) in Infrastructure; policy (identity prefix, limits, `MigrationInputError` mapping) stays in `application/migration.py`; no Standard Schema policy leaked downward.
- **Prelude + rename — VERIFIED / PARTIAL.** One `admit_command_authority` in `application/authority.py` used by all four read-only use cases. The `validation_support → contract_validation` rename is byte-identical content-wise — same four importers, same private-name coupling — but the new name does state the owner (closed Authority contract validation), which resolves the "catch-all hides ownership" finding at minimal cost. Acceptable.
- **Gate tests — VERIFIED, one residual.** Both snapshot tests and the redundant cold-import subprocess test are gone; the remaining 10 gates are contract-derived and generalize to unseen files. Residual: the coupling the deleted snapshot test named as prohibited (`interfaces/cli/template_catalog.py` / `template_instantiation.py` importing `MODEL_REFUSAL_CATALOG` from `domain.model.resolution`) still exists and is now simply unenforced. It is legal under the direction rule, so this is an observation, not a blocker — noting it so the "delete" (vs "generalize") disposition is on record.
- **Domain schema projections — PARTIAL, as declared.** The misleading `success` suffix is gone and UI owns the enclosing composition, but for `package get` the Domain projection is still effectively the whole success payload. The "authority-member projection" framing is defensible; accepted as a judgment call.
- **Bounded-input identity — VERIFIED.** Typed `BoundedInputObservation(data: bytes | None, sha256: str)`; the optional exception attribute and the impossible-state `RuntimeError` guard are deleted.

The three declared **Not adopted** items (synthetic `RefusalReport` dispatch branch, `infrastructure/distribution.py` wrapper, private-import smell) are judgment calls consistent with the code as it stands; no further objection.

### Minor observations (non-blocking, for the record)

1. `schema/funnel.load()` + `SchemaInputReadError` now have no production consumer — only `tests/_legacy_design_adapters.py` — i.e. the isolation fix left a test-only escape hatch in the legacy tree.
2. `infrastructure/input_bytes.py` now carries three overlapping read functions and four exception types; worth watching for drift if a fourth variant appears.
3. `checking.py`'s private-name cross-module imports went 17 → 24 (now spread across `resolution`/`lowering`), and `admission.py` pulls 23 private names from `lowering.py` — consistent with the declared "no speculative interface layer" stance, recorded here as the accepted trade.

No regressions in the delta; test edits are import renames only, with no weakened assertions. From this reviewer's side the PR is ready for the merge decision.

qinhaihong-red:
Handled the non-blocking observations from the latest re-review at exact head `bc79bf6`.

### Fully adopted

- Removed `schema.funnel.load()` and `SchemaInputReadError`. They had no production consumer after migration ingress moved to Application/Infrastructure, but still exposed a shipped test-only input path. The retired CLI regression driver now owns its bounded read locally and preserves the former `unreadable_input` result.

### Not adopted

- Did not merge the three `infrastructure.input_bytes` operations. They have different current contracts: bounded bytes, complete-stream identity for an oversized observation, and regular-file identity with a bounded parse prefix. A unified parameterized reader would add complexity without removing a demonstrated defect.
- Did not add an interface layer or numeric gate for private cross-module imports. The imports remain a known package-internal trade-off; there is no new dependency-direction, behavior, or change-amplification failure to justify a broad follow-up in this refactor.
- Retained the Template UI dependency on the Domain-owned Model refusal catalog. It follows the allowed dependency direction and does not move business semantics into UI.

### Verification

- Local focused regression set: **121 passed**.
- Local Ruff lint/format and Pyright passed.
- All exact-head PR checks passed, including all six gda-balancing test shards, inventory closure, and wheel/subprocess smoke.

Commit: `bc79bf6`.

qinhaihong-red:
## Architecture review: layered refactor vs DDD principles (#624)

Reviewed at head `bc79bf6` (merge-base `c594295`, 40 commits, 149 files). Head is now `07a1abf`; the delta since the reviewed head is docs-only (7 docs commits + a main merge — no `src/`/`tests/`/`tools/` changes), so every finding below holds at the current head.

Verification performed: the AST layer-dependency gate (`tests/test_layer_dependencies.py`) passes 10/10 at head; sampled behavior tests pass 79/79 (version bundle + command, formula CLI round-trip). The full 1,428-test suite, pyright, and source/wheel parity were not re-run independently.

### Overall DDD verdict

The refactor conforms to the layering defined by #624 and bADR-0025: one-way dependencies `interfaces.cli -> application -> domain -> infrastructure`, no upward imports or cross-layer cycles (AST gate), domain is free of CLI descriptors/envelopes/presentation schema, no speculative scaffolding (no DI container, service locator, event bus, blanket Repository, or `common`/`shared` modules), and the ubiquitous-language clusters (authority / formula / model / runtime / experiment+evidence / template) have clear owners. The main caveat is intra-layer seams (S1 below).

### Standards

- **[Medium, judgement] Intra-layer private-symbol coupling** — module boundaries are partly nominal. `domain/evidence.py:22` imports ~30 private symbols from `domain/runtime/execution.py` (`_NamedRng`, `_evaluate_initialization_programs`, `_runtime_journal_contract`, ...); `domain/model/lowering.py:29` imports 12+ private symbols from `domain/model/resolution.py` (`_model_lowering`, `_formula_policy`, `_selected_values`, ...); `model/admission.py` and `authority/vector_validation.py` follow the same pattern. The gate checks cross-layer direction and cycles but not these same-layer seams. Suggestion: promote the referenced internals to explicit public module seams, or merge the modules, and track as follow-up debt.
- **[Low-Medium, judgement] Duplicated Kernel vector-contract knowledge** — the same closure-vector member laws (`diagnostic_members`, `fixture_modes`, `lock_oracle_members`, `relation_kinds`) are re-asserted in `domain/authority/package_projection.py:940` (~10 sites), `vector_validation.py:629` (~9), and `admission.py:359`. Schema generation and both validation paths each restate the same contract shape; consider one shared contract module (DRY / single source of truth).
- **[Low, judgement] Duplicated refusal helper** — `_bounded_refusal` exists in both `application/migration.py:631` and `domain/model/resolution.py:164`, differing only in stage handling; it could live once in `domain/diagnostics.py`.
- **[Low] Import ordering** — `interfaces/cli/surface.py:6` places `from copy import deepcopy` between first-party imports; ruff (E4/E7/E9/F) does not enforce it. Non-blocking.

### Spec

- **[Low-Medium] Stale review claim in the PR body** — the body says "independent Spec and Standards reviews passed at exact head 2201e2c"; the head is now `07a1abf` and the branch advanced materially since `2201e2c` (including `c8646be`, which split `domain/model/resolution.py` into admission/lowering, +1,914/+4,369 lines). Please update the claim to a current SHA or drop the exact-head wording.
- **[Low, judgement] Two "fix" commits change observable diagnostics** — `eb0540d` (bound source reads at ingress) and `2201e2c` (preserve oversized input identities) move the oversized-input refusal `content_identity` from the intermediate `"unidentified"` back to `sha256:<hash>`. The pre-refactor base (`c594295`) already reported the sha256 identity, so this restores prior public behavior rather than changing it — corrective and in scope, though the migration briefly violated the issue's "behavior differences are separate requirements" rule.
- No missing acceptance criteria and no out-of-scope production changes found (the delta since the review head is docs-only).

### Summary

Standards: 4 judgement-call findings, 0 documented-standard violations — worst: intra-layer private seams (S1). Spec: 3 findings — worst: stale PR-body review claim; no missing ACs.

qinhaihong-red:
Handled against current head `5777ebf`.

| Claim | Decision | Resolution and evidence |
| --- | --- | --- |
| Intra-layer private-symbol coupling | Partially Adopt | The coupling is a valid maintainability observation, but it does not violate the allowed same-layer, acyclic dependency contract. This PR does not mechanically publish more than 100 internal names or merge the separated large modules because neither action narrows the dependency. A narrower seam should be extracted only when repeated changes demonstrate concrete amplification. |
| Duplicated Kernel vector-contract knowledge | Do Not Adopt | `package_projection.py` and `vector_validation.py` consume the same Kernel-owned `model_program_vector` for different purposes: schema projection and vector admission. `admission.py` validates the separate LDB Diagnostic contract. A new shared contract module would couple distinct consumers without removing a peer authority. |
| Duplicated `_bounded_refusal` | Do Not Adopt | The migration wrapper always returns a migration-stage report and merges the caller's truncation state. The Model wrapper derives one stage, returns `None` for no diagnostics, and rejects mixed stages. Their common bounding mechanism already lives in `domain/diagnostics.py`; combining the wrappers would add policy switches. |
| Import ordering | Fully Adopt | `5777ebf` groups standard-library and first-party imports correctly in `interfaces/cli/surface.py`. |
| Stale exact-head statement in the PR body | Fully Adopt | The obsolete `2201e2c` claim was removed. The body now records the durable review result without presenting an old SHA as proof for the current head. |
| Oversized-input identity repair | Do Not Adopt (no current defect) | The observation accurately describes an intermediate Experiment regression and its repair. The final branch restores the pre-refactor SHA-256 identity with bounded memory, so no additional behavior change or requirement is needed. |

Validation at `5777ebf`:

- local Ruff check and format check passed;
- focused architecture and CLI conformance tests: `265 passed, 27 skipped`;
- all required GitHub checks passed, including all six gda-balancing shards, inventory closure, Pyright, and wheel/subprocess smoke;
- an independent recheck reached the same handling decisions and found no remaining contract defect in these claims.

qinhaihong-red:
Review conclusion: Changes required for the final form of Section 3.4, "Symbolic architecture".

Scope: the two latest commits in PR #626, from `70aa00b` through `e4eeb0f`.

The following issues remain before merge:

1. **High ? The symbolic model does not make the two RIR identities explicit**

Section 3.4 defines `Id(x)` as the exact content identity, but represents the Resolved Model as:

```
RM = Bind(RIR, Id(K), Id(L), Id(Lock), ...)
```

The established architecture requires the Resolved Model to bind both `RIR.semantic_identity` and `RIR.content_identity`. See the [RIR and Resolved Model identity rules](https://github.com/aigengame/godot-agent/blob/e4eeb0f64bfa04afa9e8a185e4455629e4b481f3/libs/gda-balancing/docs/ARCHITECTURE.md#L487).

The current notation can lead an implementer to treat RIR as having one identity. It does not reliably distinguish semantic equivalence from exact wire-content equality.

Please either introduce the two components explicitly:

```
RIR_s = semantic_identity(RIR)
RIR_c = content_identity(RIR)

RM = Bind(Id_c(K), Id_c(L), Id_c(Lock), RIR_s, RIR_c, ...)
```

or state that `Bind(RIR, ...)` expands to both identities.

2. **High ? The Evidence chain is too compressed, and the publication/governance boundary is unclear**

The current model is approximately:

```
Run / Dataset -> ValidateAndIssueEvidence -> Govern -> Publish
```

The established architecture requires a Replay comparison or Cross-evaluator comparison before Evidence issuance. A single Evaluation Run cannot issue Evidence by itself. See the [Evidence chain](https://github.com/aigengame/godot-agent/blob/e4eeb0f64bfa04afa9e8a185e4455629e4b481f3/libs/gda-balancing/docs/ARCHITECTURE.md#L968) and the [bADR-0018 Evidence-issuance rules](https://github.com/aigengame/godot-agent/blob/e4eeb0f64bfa04afa9e8a185e4455629e4b481f3/libs/gda-balancing/docs/badr/0018-unified-metrics-calibration-and-evidence-chain.md#L131).

`VerifiedPrerequisites` may remain an abstraction, but the detailed model should state:

- where Replay and Cross-evaluator comparisons are produced;
- that comparison artifacts are Evidence inputs, not Evidence assertions;
- how claim-specific prerequisites such as calibration, holdout, and drift artifacts enter validation;
- whether `Govern` is an external step over already published artifacts or part of a separate atomic publication transaction.

Without this clarification, an implementation could issue Evidence directly from one Run.

3. **Medium ? Required build companion artifacts are missing from the symbolic model**

`Compile_A` returns `AST, HIR, RIR, DebugMap, RM`, but the architecture requires every successful build to produce and atomically publish the Model explanation, Build receipt, Debug Map, and Resolved Model. See the [Model explanation and build-publication rules](https://github.com/aigengame/godot-agent/blob/e4eeb0f64bfa04afa9e8a185e4455629e4b481f3/libs/gda-balancing/docs/ARCHITECTURE.md#L502).

The model also produces a `ResolutionReceipt` without specifying how it enters the Build receipt or the complete artifact set.

Please add `ModelExplanation` and `BuildReceipt` to the build result, or state clearly that a later build/publication stage generates, validates, and atomically commits them with the Resolved Model. They should remain outside the Resolved Model identity.

4. **Medium ? The notation is still not fully self-contained**

The second commit correctly moves the notation before the symbolic models, but several symbols used by the models are still undefined in Section 3.4, including `RootManifest`, `Manifest_i`, `Requirements(M)`, `RuntimeProfileRef(E)`, `ClaimKind`, `CompleteArtifactSet`, and `AtomicFilesystem`.

`Seal` is also used as a constructor without defining whether it means canonicalization, closure binding, content-identity generation, or a cryptographic seal.

Please add a complete symbol table grouped by artifact, input, judgment, and host mechanism, and define `Seal` explicitly. The heading "Symbolic architecture" should also state that this is a non-normative symbolic view of the existing architecture, so readers do not interpret it as a new runtime architecture.

These issues affect the implementability and auditability of identity, Evidence, and artifact publication. After revision, please recheck Section 3.4 against Sections 6, 9, and 10 and bADR-0012/0013/0018.

qinhaihong-red:
Handled the latest symbolic-architecture review at exact head `128496e6`.

| --- | --- | --- |
| Resolved Model does not expose both RIR identities | Fully Adopt | Added `RIR_s = SemId(RIR)` and `RIR_c = Id(RIR)`. The `RM` binding now includes both identities explicitly, preserving the distinction between semantic equivalence and exact canonical JSON identity. |
| Evidence chain and governance/publication boundary are too compressed | Partially Adopt | The existing prose already prohibited one Evaluation run or Metric dataset from issuing Evidence, so that premise did not require a semantic change. The detailed model now shows Replay and Cross-evaluator comparison production, claim-specific comparison/calibration/holdout/drift prerequisites, comparison artifacts as published Evidence inputs rather than assertions, Evidence publication, external governance over published artifacts, and a separate Approval Record publication transaction. |
| Required build companions are absent | Fully Adopt | The model now includes the Model explanation, generated Capability manifest, Resolution receipt, Build receipt, and the complete atomic build set. It shows the Resolution receipt entering Build provenance and the complete set while keeping every companion outside Resolved Model identity. |
| Notation is not self-contained | Partially Adopt | Added grouped definitions for the non-obvious authority, input, judgment, derived-artifact, Evidence, publication, and host symbols; defined `Seal` as canonical closure binding rather than cryptographic signing; and renamed the section to “Non-normative symbolic architecture.” I did not duplicate the full domain glossary as a second terminology authority. |

The compact mental model also now has labeled sections for authority admission, model build, Experiment/Runtime execution, comparison and Evidence, external governance, host call/result flow, and Domain use of Infrastructure mechanisms. This keeps the two final arrows from appearing as an unexplained continuation of the Evidence chain.

Validation at `128496e6`:

- rechecked Section 3.4 against Sections 6, 9, and 10 and bADR-0012/0013/0018;
- `git diff --check` passed;
- `tests/test_layer_dependencies.py`: **10 passed**;
- GitHub checks for this exact head are currently running; no CI-success claim is made yet.

Issue #585, opened 2026-07-25
[gda-balancing] Dogfood one seeded Roguelike reward/build tuning loop


qinhaihong-red:
## Parent

- #534
- #501

## What to build

Before proof hardening and general RPG/Roguelike coverage closure, consume #540's public
Model/Experiment/runtime/artifact path plus #590's Formula-authoring and Model-explanation contract
with one structurally different Roguelike tuning example.
A balancing designer authors a seeded reward pool, one declared rarity policy, one build slot with
an explicit conflict/replacement rule, and the reward disposition in Model Source and an Experiment
Specification. The designer runs the exact seed sequence, inspects the public trace and Metrics,
changes one reward weight or policy/build constraint, and reruns without changing host code.

This is a permanent product-feedback slice, not a disposable prototype, coverage-row closure, or
template support claim. Its observations challenge product and architecture assumptions for #534
and inform the later formal closure in #567–#574; they do not evaluate or close Gate 5's
cross-genre claim.

## Acceptance criteria

- [ ] The example publishes exact Model Source and Experiment artifacts and executes only through the public `model` and `experiment` commands introduced by #538–#540; no private fixture path, host-owned generator, evaluator default, or second language supplies semantics.
- [ ] The Roguelike source declares and statically binds at least one game-owned Formula to an admitted Operation slot, and its build publishes the mandatory Model explanation. The example inspects the exact Formula/Operation boundary and changes that Formula or its binding without package/evaluator fallback or genre-specific dispatch.
- [ ] One e2e designer loop performs configure → build/check → run → inspect → edit one authored reward/policy/build value → rerun, with exact input identities and an observable, explainable change in the reward/build result or Metric.
- [ ] The example exercises an ordered eligible reward pool, a named RNG stream, one closed rarity-policy variant, one build conflict or atomic replacement, and one exhaustive typed reward disposition.
- [ ] Public artifacts expose the seed/stream and draw indexes, ordered candidates, selected policy state before/after, selected reward, build decision/replacement, disposition, and the Metrics needed to compare the two authored configurations.
- [ ] Unsatisfiable selection without a declared fallback is a typed refusal; a declared no-reward or relaxed-pool fallback is a gameplay outcome; build conflict and invalid configuration are not mislabeled as internal failures.
[...]

PR #637, opened 2026-08-11
feat(gda-balancing): execute structured values end to end

### LINKED ISSUE #636: [gda-balancing] Execute structured values through Model and Experiment
## Parent

- #585
- #534
- #501

## Background

Issue #585 must author and execute an ordered reward pool, a closed rarity policy, a typed reward
disposition, and an atomic build replacement through the public Model and Experiment path. These
concepts require ordered, nominal, and structured values. Encoding them as integers, loose JSON, or
host-owned objects would lose their type and identity contracts and would bypass the accepted
Standard Schema authority boundaries.

At issue creation, the admitted `standard.schema@2.2.0` Model Source contract accepts only scalar
`Int` symbols. Formula operation-call results add only scalar `Int` and `Bool`. The macro
architecture and bADR-0016 already define a closed constructor set that includes `Enum`, `Record`,
`List`, and `Ref<T>`, but the current machine path does not execute those constructors end to end.
The current Language Definition Bundle also has no `game.generation`, `game.build`, or
`game.collection` release. Those Domain packages belong to #585; this issue must not pre-implement
their semantics.

This issue closes that implementation gap before #585. It does not add a constructor to the
accepted set, claim Roguelike support, or close a Gate 5 coverage row.

## Human decisions (resolved 2026-08-11)

The human architecture owner approved these boundaries before AFK implementation:

- `standard.schema@2.3.0` owns the `Enum`, `Record`, `List`, and `Ref<T>` constructor definitions,
  their type/value/literal/lowering rules, and the minimum generic operations for Record field
  access, Enum equality, reference equality, and bounded List index access. It does not own
  scenario-specific nominal types or higher-order collection algorithms.
- `standard.conformance.structured@1.0.0` owns the maintained neutral scenario's nominal
  `CandidateKind`, `Candidate`, `SelectionState`, and `SelectionResult` types and its normative
  vectors. It is a formal, versioned LDB member for public structured-value conformance, not a
  private fixture or a source of generic language semantics.
- Source identifies a nominal type by package coordinate and exported type id. Resolution binds it
  to the exact LDB identity, Package Release identity, and exported type id; structural equality
  never substitutes for that identity. Enum values contain one declared member id. Record values
  contain all and only the declared fields. Record object-member order is semantically neutral.
  `List<E>` is invariant and preserves length, duplicates, element values, and authored order.
  `Ref<T>` contains the exact nominal target and one package-defined canonical key; equality
  requires both to match. This scenario uses lowercase ASCII keys matching
  `^[a-z][a-z0-9_]*$`, unique within the target nominal type. Public artifact boundaries expose one
  `{type, value}` envelope without repeating types that the enclosing type uniquely determines.
- The current machine Kernel is not sufficient as written: its Runtime program has no lookup or
  equality node and its literal contract is still integer-only. Before the rest of the slice, amend
  bADR-0022 and the exact Kernel Specification and vectors to generalize the existing typed-literal
  contract, add the already accepted bounded `lookup` primitive for static Record-field and bounded
  List-index access, and add one exact-type canonical `equal` primitive used here for Enum and
  `Ref<T>`. This is the approved Schema-major authority completion for this issue.

Model Source owns only typed symbols and values that use package-owned types. It does not define or
export nominal types. `standard.compiler` and `standard.runtime` remain conforming execution
owners, not value-semantic authorities.

## What to build

Deliver one maintained, engine- and game-agnostic structured-value scenario through the existing
public commands. Its Model Source declares typed symbols and values for the package-owned closed
Enum, nominal Records, an ordered List of Records, and typed references selected above. An
Experiment uses the existing named RNG contract to choose an index, performs a bounded generic
lookup, compares typed values, atomically replaces structured state, returns a
structured result, and publishes an observable numeric Metric derived from that result. After the
scenario is delivered, editing one authored candidate value or order and rerunning the same
Experiment must produce an explainable public change without changing host code or dispatch.

Implement the smallest generic semantics that this scenario requires. The initial scope is:

- the approved bADR-0022 and Kernel Specification completion, with exact positive, negative, and
  boundary vectors before host execution support;
- `Enum`, `Record`, `List`, and `Ref<T>` admission and type relations;
- Record field access, Enum and reference equality, and bounded List index access;
- canonical encoding, identity, type checking, HIR/RIR lowering, and public explanation of nested
  values;
- structured Operation operands and results, structured Runtime state, atomic commit or rollback,
  and structured Snapshot and trace projections; and
- the existing independent-consumer conformance path for the exact admitted behavior.

The Language Definition Bundle owns constructor definitions, type relations, literal rules,
lowering rules, and reducible operations through the exact Package Releases selected above. Domain
packages only instantiate and compose those contracts. Host implementations conform to the
admitted authority and must not add a private value matcher, serializer, evaluator table, or
fallback. The approved typed-literal, lookup, and equality changes do not authorize aggregation,
user-defined traversal, reflection, automatic dereference, host callbacks, or a new Runtime phase.

`Fixed`, `Decimal`, `Float`, `Vector`, `Set`, `Map`, and `Distribution` are initial non-goals, not
permanent exclusions. Add another already accepted constructor only after a concrete #585 path
fails and evidence shows that the initial constructors cannot compose the required behavior. Obtain
maintainer confirmation before expanding the issue.

## Acceptance criteria

- [ ] bADR-0022, the exact Kernel Specification, its identity, Kernel vectors, the macro
  architecture, and the LDB projection agree on the approved typed-literal, Record/List lookup, and
  exact-type equality laws. No prose-only or host-only primitive remains.
- [ ] One maintained neutral scenario runs through public `model check`, `model build`, `model
  inspect`, `experiment check`, and `experiment run` commands. It contains nested
  `List<Record<...>>`, closed Enum values, and nominal `Ref<T>` values; no private fixture or helper
  path supplies semantics.
- [ ] The exact LDB authority defines the four constructors, their recursive type relations,
  literals, generic operations, Diagnostics, lowering rules, resource bounds, and normative
  vectors. Derived schemas, catalogs, Model explanation, and Runtime projections consume or
  reverse-check that authority instead of becoming peer definitions.
- [ ] Model checking and lowering preserve nominal Record and reference identities, closed Enum
  membership, invariant List element type, and List order through Typed HIR and RIR. Reordering
  object members alone is semantically neutral; changing List order or a typed value changes the
  applicable semantic identity or observable result.
- [ ] Runtime performs one statically bounded List index lookup using the existing named RNG
  result, reads typed Record fields, exercises both Enum and `Ref<T>` equality, and atomically
  commits a structured state replacement and structured result through the existing Event
  transaction.
- [ ] One declared failure occurs after a structured write is buffered. The applicable existing
  contract classifies it as either a discriminated gameplay outcome or a Runtime refusal, rolls
  back that Event, and leaves the committed Snapshot unchanged. It publishes only the artifacts
  required by that classification, including a terminal-audit artifact set only when applicable,
  without a second transaction mechanism.
- [ ] Public Model explanation, Snapshot, Event trace, Evaluation run, and Metric dataset expose
  enough typed identities and values to explain the selected element and the state before and after
  the replacement. No integer code, expected-output record, fixture name, or host-only field stands
  in for a structured value.
- [ ] With the seed and draw index fixed, editing the selected candidate or reordering the List so
  that the same index selects a different candidate changes the structured result or derived Metric
  while leaving unrelated named RNG streams and existing Runtime ordering unchanged.
- [ ] Production and the existing independent consumer agree on positive, negative, and boundary
  vectors. At minimum, vectors cover an unknown Enum value, a missing or extra Record field, a wrong
  List element type, a wrong `Ref<T>` target, an empty or out-of-range lookup, and the declared
  lookup/resource limit.
- [ ] Existing scalar Model/Experiment examples and their public artifacts remain valid. The full
  relevant test inventory passes without a second type authority, a new conformance framework, or
  duplicate host constants.
- [ ] The implementation adds no reward, rarity, disposition, build, or collection policy; no
  `game.generation`, `game.build`, or `game.collection` release; and no genre-specific compiler or
  evaluator dispatch. #585 owns those Domain packages and the Roguelike designer loop.
- [ ] Initial implementation can add generic, authority-driven compiler and evaluator support for
  the accepted constructors and the explicitly approved typed-literal, lookup, and equality
  Kernel completion. The implementation stops and records the failing scenario and the narrowest
  affected authority if it requires package-, fixture-, or genre-specific dispatch, any other
  irreducible Kernel primitive, a constructor-set change, or a Runtime-phase change.
- [ ] The maintained scenario exercises each generic structured-value capability that #585
  currently requires without package- or genre-specific behavior. This issue alone does not satisfy
  any #585 or Gate 5 acceptance claim.

## Blocked by

None - can start immediately.

qinhaihong-red:
Closes #636.

## Summary

- complete the approved Kernel contract for generalized typed literals, statically bounded Record/List lookup, and exact-type equality
- admit authority-defined `Enum`, `Record`, `List`, and `Ref<T>` values, preserve their selected semantics through HIR/RIR, and execute them through the existing Experiment transaction
- publish `standard.schema@2.3.0` and the neutral `standard.conformance.structured@1.0.0` scenario with conformance vectors, public examples, and updated architecture/domain documentation
- preserve existing scalar examples while exposing typed structured state, results, traces, evaluation output, and derived metrics

## Review handling

The delegated Standards/Spec review initially found seven issues. After those fixes, a narrow re-review identified three remaining authority-boundary gaps; all three were accepted because they directly enforced #636 rather than expanding its scope:

- List lookup now requires a visible Kernel-compatible `Int` local during static operation admission in both production and the independent consumer.
- Quantity lookup result contracts are projected from the exact selected literal profile; no package-version host constant remains.
- constructor `value_rule` is a Kernel-owned closed discriminated union, and both consumers reject unknown, incomplete, or ill-typed laws.

The final delegated re-review reported no actionable findings. Its mutation/regression slice passed (`11 passed`).

## Validation

- grouped relevant inventory: 1,416 passing tests and 27 intentional skips; the one overlapping-profile regression exposed during grouping was fixed and its affected slice rerun successfully
- existing combat, periodic-effect, and new structured-selection examples: `model build` / `experiment check` pass with exact regenerated identities
- `pyright`: 0 errors, 0 warnings
- `ruff check` and `ruff format --check`: pass
- LDB rebuild `--check`: pass
- `git diff --check`: pass

## Scope boundary

This PR adds only generic authority-driven structured-value execution. It does not add reward, rarity, disposition, build, collection, or genre-specific policy; those remain owned by #585.


qinhaihong-red:
## Design review — domain modular architecture (completeness / orthogonality / DRY / extensibility)

Reviewed at head 5135d0ed against issue #636 and the governing artifacts (bADR-0016/0022/0025, BALANCING-CONTEXT.md, ARCHITECTURE.md). Method: four delegated digests (authority JSON, model-side Python, runtime-side Python, conformance/tests/examples), with every load-bearing claim below verified directly (functions read in full, greps re-run) before reporting.

**Verdict**: the core design is sound. Three deviations remain; findings F1 and F2 touch #636 acceptance criteria directly and should be fixed before merge.

### What holds up

- The constructor `value_rule` closed discriminated union is defined once in kernel.json; the RIR wire schema deliberately abstains (`value_rule: {}`). Single authority, no parallel registry.
- `structured_values.py` evaluates through the authority's own indirection members (`fields_member`, `element_member`, …); unknown operators refuse. One shared evaluator serves admission, lookup, equality, and vector execution.
- The existing Event transaction is reused as-is; the structured lookup refusal joins the existing rollback path. No second transaction mechanism.
- Vector coverage meets the #636 minimum list case by case, pinned by a three-way production / independent-consumer / expect agreement test (`test_schema2_authority_cli.py:722`).
- The maintained scenario is neutral, runs only through public CLI commands, and the "edit a candidate / reorder the list → explainable change, unrelated RNG streams unchanged, rollback intact" property is pinned by `test_schema2_experiment_cli.py:1566`.
- Terminology stayed synchronized: BALANCING-CONTEXT.md adds **Typed value envelope** and generalizes **Literal Typing Profile**; bADR-0022 is amended with a dated note.

### F1 (high) — private static value navigator in `admission.py`

`_operation_contract_for_structured_type` / `_declared_lookup_operation_contract` (`domain/authority/admission.py:3232-3320`) hardcode the constructor kinds `"record"` / `"list"` / `"ref"`, the member names `"fields"` / `"name"` / `"type"` / `"element"`, and the `core.quantity`-owned representation name `"Int"` (admission.py:3314). The runtime path (`structured_values._validate`) dereferences the same knowledge through the constructor's declared `value_rule` indirection members. The same rule — how to read a Record field or a List element from a type definition — now has two owners, and only one is authority-driven. It is correct today only because `standard.schema@2.3.0` happened to choose member names identical to its own indirection targets; renaming any `value_rule` member would silently desynchronize static typing from runtime evaluation. #636 states hosts "must not add a private value matcher … no duplicate host constants". The scalar half of this function does project from the selected literal profile (per the re-review fixes), but the record/list navigation and the `"Int"` literal do not.

Suggested fix: consume the `value_rule` indirection here as well, or delegate the static lookup-result contract to a shared `structured_values` helper.

### F2 (high) — declared-but-unconsumed generic operations, with dangling kernel coordinates

The four `standard.schema` `structured_operations` (`enum-equal-v1`, `list-at-v1`, `record-field-v1`, `ref-equal-v1`) have zero consumers under `src/` (`git grep structured_operations` — no hits); their kernel-node bodies are never executed or type-checked, and `select-v1` inlines `lookup` / `equal` nodes directly. This is the mirror image of #636's "No prose-only or host-only primitive remains": an authority-only primitive. Direct consequence: their bodies reference `kernel/2.0.0/Int64` and `kernel/2.0.0/String`, and neither type id exists in kernel.json (its declared ids are `Boolean`, `Unit`, `EventReference`) — no closure check catches this, which also shows kernel-coordinate references sit outside the current reference law.

Suggested fix: give the four operations an execution or closure-check path (vectors, or reference-law coverage of `structured_operations` bodies) — or, with maintainer confirmation per #636, narrow them out of this slice. The dangling `Int64` / `String` coordinates need repair either way.

### F3 (medium) — type-expression grammar is host-only

The `kind: "nominal"` discriminator and the constructor-application shapes (`{"kind":"record","fields":[…]}` etc.) are declared in no JSON authority: `nominal_types.definition` and `structured_operations.{inputs,result,body}` are unconstrained `canonical-value`, and the wire schema abstains. Only `structured_values._nominal_key` knows this grammar. This also falls under "no host-only primitive"; at minimum record it as a known gap, preferably declare the grammar in the kernel meta-format.

### F4 (medium) — typed-envelope admission law duplicated inside production

The law dict `{"envelope_members":["type","value"],"operator":"recursive-typed-envelope","resource_charge_per_node":1,"type_relation":"exact-selected-type"}` is hand-typed in both `admission.py:3010` and `structured_values.py:71` with no cross-check. The kernel↔package copy is guarded (drift refuses at admission), and the production↔independent-consumer duplication is by design; the production-internal pair is not — if kernel.json's law changes and only one copy follows, the failure mode is a raw `ValueError` at model-build/run time instead of an admission diagnostic. One shared production constant (or one module reading the other) closes it.

### Low / watch

- The value_kind compatibility-collapse rule appears near-verbatim three times (`runtime_validation.py:16`, `_lowering.py:3433`, `_lowering.py:1830`).
- The `{type, value}` shape guard is inlined twice in `execution.py`; `structured_values` exports no predicate.
- The `value_rule` mutation-refusal test (`test_schema2_bootstrap_authority.py:43`) covers only `standard.schema.list`; Enum/Record/Ref laws have no mutate-the-law coverage.
- Kernel `lookup` node declares refusal `lookup-out-of-range`, but the wire signal is `structured-lookup-out-of-range`, and nothing consumes the node's `refusals` list.
- `standard.schema@2.3.0`'s own conformance vectors embed `standard.conformance.structured` coordinates (4 occurrences) while its `dependencies.required` is empty — a downstream reference from the generic package; a package-neutral fixture type would remove it.
- `_structured_model_source()` in `test_schema2_model_cli.py` rebuilds the checked-in structured-selection Model Source field by field — fixture-drift risk; loading the example file would remove the second copy.
- `lookup_typed_value` on a nominal type without a `definition` (e.g. an `exact-integer` type) raises `KeyError` instead of a structured refusal — inconsistent with `lookup_selector_kind`'s guard; low reachability since vectors are authority-owned, but the refusal boundary should be uniform.
- Extensibility: the design is shape-extensible (a new `value_rule` variant is a well-formed JSON edit; wire schemas are untouched) but a 5th constructor still costs one kernel edit (closed union + the `required_constructors` enumeration — kernel now names five `standard.schema.*` ids, defensible under bADR-0016's closed-for-one-major decision but worth recording as an accepted trade-off), a `standard.schema` version bump, and three sanctioned switch arms in `structured_values` (four today, counting F1) with no mutual-exhaustiveness test. The `static-field` / `local-index` selector mapping exists only at `structured_values.py:496` with no JSON declaration. The envelope `oneOf` fragment is inlined 25× across the wire schemas with no `$defs` — the main driver of the `max_authority_bytes` doubling.

qinhaihong-red:
Implemented at exact head `e84599e2` after evaluating each finding against #636 and the current authority contracts.

| Claim | Decision | Change and evidence | Validation |
| --- | --- | --- | --- |
| F1: static lookup duplicated constructor member knowledge | Fully Adopt | Static admission now projects Record/List lookup through the admitted constructor `value_rule` and selected literal profiles. Mutation tests rename the Record/List definition members and confirm that both conforming consumers still admit the reidentified graph. | Two-consumer authority/composition suite passed: 244 tests. |
| F2: `structured_operations` were unconsumed pseudo-programs with dangling Kernel coordinates | Fully Adopt | Replaced the pseudo-bodies with closed declarative laws for bounded lookup and canonical equality. Static admission, Runtime, RIR projection, and the independent consumer now consume those laws. The nonexistent `kernel/2.0.0/Int64` and `String` references are gone. | Law projection/refusal mutations are rejected symmetrically; permanent two-compiler vector test passed. |
| F3: nominal type-reference shape was host-only | Partially Adopt | Added the nominal reference shape to the existing Kernel typed-envelope admission law and consume it in both implementations. I did not add a second general type-expression DSL: constructor shapes remain owned by their existing closed `value_rule` variants. | Tagged references canonicalize; undeclared members refuse through `StructuredValueFault`. |
| F4: production duplicated the typed-envelope admission law | Fully Adopt | Added one production constant shared by authority admission and structured-value evaluation. The independent consumer remains intentionally separate. | Ruff and Pyright pass; packaged authority admits with zero diagnostics. |
| Low-risk fixture/refusal issues | Fully Adopt where local | The Standard Schema vectors are package-neutral, `test_schema2_model_cli.py` loads the maintained example, fixed nominal lookup returns a structured refusal, and the lookup refusal signal is connected to the declared operation law. | Maintained RPG, periodic, and structured public flows passed: 8 tests. |
| Remaining watch items | Do Not Adopt in this PR | I did not add a new global value-kind abstraction, envelope predicate framework, constructor plugin mechanism, or schema-wide `$defs` rewrite. No current failure justified that scope, and those changes would expand #636 without closing a demonstrated defect. | Existing behavior and current architecture remain unchanged in those areas. |

The authority rebuild also refreshed all affected package/vector/example identities from real compiler output rather than hand-derived values.

Local validation at `e84599e2`:

- authority/composition: `244 passed`
- focused review regressions: `18 passed`
- permanent two-compiler vectors: `1 passed`
- maintained public examples: `8 passed`
- Ruff: passed
- Pyright: `0 errors`

CI is not yet green: `gda-balancing inventory closure` failed and caused the required aggregate plus downstream balancing jobs to skip. I am investigating that job separately and do not claim CI closure yet.

qinhaihong-red:
## Re-review of the claimed fixes at `e84599e2`

I verified each claim directly against the new head (functions read in full, greps re-run). Result: **F1, F2, and F4 are fixed as claimed; the F3 partial adoption is real and its scope argument holds; every "fully adopt where local" low item checks out.** The declined watch items are an acceptable scope call. Three small residuals are recorded below as watch-level; none blocks. The one open DoD item is CI, which the reply itself discloses.

| Finding | Verified evidence at `e84599e2` |
| --- | --- |
| F1 | `_declared_lookup_operation_contract` now delegates to `structured_values.lookup_type_contract`, which projects selector, result type, and refusal signal from the declared `bounded-lookup` law and navigates through `value_rule` member indirection. The `"Int"` literal is gone — List key admission matches key candidates against selected integer literal profiles via `_operation_value_contract_matches`. Load-bearing proof: `test_two_consumers_follow_constructor_member_indirection` renames `fields`→`members` and `element`→`item` in both the `value_rule` and every nominal definition, and both consumers still admit the reidentified graph. |
| F2 | The four `structured_operations` now carry closed declarative laws (`canonical-equal` → `result_contract`; `bounded-lookup` → `selector` / `result_projection` / `refusal_signal`) and are consumed: production reads them through `_structured_operation_law` (static admission via `lookup_type_contract` / `equal_result_contract`, runtime via `lookup_selector_kind`), and the independent consumer reads them in parallel. `kernel/2.0.0/Int64` and `String`: 0 occurrences. Law mutations refuse symmetrically (`test_two_consumers_require_declared_record_lookup_semantics`), and `test_current_slice_refuses_not_yet_delivered_operation_definitions` closes the "declared but undelivered" hole this finding came from. |
| F3 | The kernel typed-envelope admission law now declares `nominal_type_reference` (`coordinate_members`, `optional_kind_member`/`value`), and `nominal_type_key` reads coordinates through it; undeclared members refuse (`test_nominal_reference_refuses_members_outside_the_selected_profile`). Keeping constructor-application shapes owned by their `value_rule` variants — rather than adding a second type-expression DSL — is the right-sized resolution; the member-rename mutation test now guards exactly the drift I was worried about. |
| F4 | One production constant `TYPED_ENVELOPE_ADMISSION_LAW` in `structured_values.py`, imported and used by `admission.py`. The independent consumer stays separate, as the conformance methodology requires. |
| Low items | standard.schema vectors: 0 scenario-package references. `test_schema2_model_cli.py` loads the maintained example file. A nominal type without a `definition` now refuses through `StructuredValueFault` in `_structural_type_contract` (the `KeyError` path is gone). The kernel `lookup` node refusal is renamed `structured-lookup-out-of-range`, matching the declared signal, and the law's `refusal_signal` is consumed at static admission. |

Declining the global value-kind abstraction, envelope predicate, constructor plugin mechanism, and `$defs` rewrite is consistent with how I ranked them (watch, not defects); no objection.

Residual watch notes (non-blocking, no action requested in this PR):

- `_operation_contract_for_structured_type` still classifies inline structural results with a literal `{"list", "ref"}` kind set — no member navigation, but a small kind enumeration outside the `value_rule` vocabulary.
- The runtime bridge from `StructuredValueFault` code to `signal="structured-lookup-out-of-range"` (`execution.py`, the `except StructuredValueFault` arm) remains one hardcoded pair; vectors pin it.
- `equal_typed_values` now fail-closed-checks the law's `result_contract == "kernel-boolean"` but still renders the result type as a literal `kernel/2.0.0/Boolean` coordinate rather than projecting it from the kernel's `fixed_value_contracts`; drift now refuses instead of diverging silently, so this is pinning, not duplication.

Remaining before merge: green CI (the `gda-balancing inventory closure` failure and the skipped aggregate), as already disclosed in the reply. Design-review side: no further findings.

qinhaihong-red:
## Entropy review (complexity-proportionality) — head 2a0e3c2d

Question reviewed: is the entropy this PR introduces (+5615/−424, 57 files) proportionate to #636's goal? This builds on the two earlier review rounds; the delta since `e84599e2` is one line (`tools/ci.py` shard registration) and does not change those verdicts.

**Verdict: Keep.** The bulk of this change is constraint tax, not free-standing mechanism: three hard constraints — authority-owned machine-readable semantics (bADR-0022), production/independent-consumer vector agreement (the two-evaluator clause), and unchanged scalar examples (an explicit acceptance criterion) — directly produce the authority JSON, the vectors, the consumer mirror, and the identity re-pins. Strip that methodology tax and the host code is close to the minimum path #636 asks for. #636's initial scope was itself a human-approved minimum slice (four constructors, three Kernel primitives, one scenario; seven constructor families explicitly deferred), and the PR does not drift from it: no game packages, no #585 semantics, no extra constructors, no new Runtime phase. The fix rounds also consistently picked the lower-entropy direction (declarative laws instead of an execution pipeline for the four operations; a small `nominal_type_reference` declaration instead of a type-expression DSL; declining the predicate framework and `$defs` rewrite). No finding below justifies reopening the implementation.

### Findings (decision-relevant only)

**E-1 (Defer, with a written trigger) — the envelope `oneOf` fragment is inlined 25× across the wire schemas, and drove the `max_authority_bytes` doubling.** The change amplification is real (one envelope-shape change = 25 edits), but a `$defs` consolidation now would be a larger one-shot entropy injection than the problem it removes — declining it in this PR was right. Smaller alternative: record a follow-up trigger — consolidate with the next `standard.schema` version when either a third literal form is needed or a #585 domain package approaches the 512 KB bound.

**E-2 (Simplify, opportunistic) — the three `StructuredValueAuthority` projection constructors repeat ~150 lines of similar indexing.** The three admission stages are real, but they differ only in input shape; a "normalize to a package list → one indexer" shape would remove the representational duplication. No law is duplicated, so this is non-blocking cleanup.

**E-3 (Keep, with an observation line) — `type_reference_closure` declares 7 configurable path members for its single current instance.** Speculative-generality shaped, but defended by the no-host-only-rules constraint: hardcoding the closure would violate machine-readability. Observation line: add no further configurable dimensions before a second instance exists.

**E-4 (Keep, cost noted) — the host-pins-kernel idiom gained ~5 new Python constant mirrors.** Fail-closed pinning is the established, guarded convention (and the fix round already collapsed the typed-envelope law to one shared constant), but its maintenance cost grows linearly with the law count — every kernel evolution is a paired edit. Nothing to do now; worth a conversation if the next slice grows it again.

### Restraint evidence (weighed in favor of Keep)

Metric sampling stayed integer-only rather than growing structured observability; the F3 resolution added one small declared shape instead of a second type DSL; the F2 resolution replaced pseudo-bodies with four small law objects instead of building an operation-execution pipeline; the deferred non-goals stayed deferred.

### Essential complexity (must remain)

Dual-implementation agreement with symmetric mutation refusal; the 17 vectors and the resource budget (both acceptance-listed); the closed discriminated union with fail-closed refusal (the authority boundary itself); the dual lowering chain (zero-change guarantee for the scalar path); the full identity re-pin chain (a consequence of the content-addressed model).

### Assumptions to test (cheapest probes)

- **A1**: admit/lookup/equal compose all structured behavior #585 needs — #636 already defines the trigger (expand only after a concrete #585 path fails); the probe is #585's first vertical slice, no action now.
- **A2**: 512 KB fits #585's domain packages — check the first #585 package's actual size; approaching the bound triggers E-1.
- **A3**: the per-constructor × per-operation law-ownership model does not explode as constructors grow — count new laws when the next constructor lands.

Remaining before merge is unchanged from the previous round: green CI. No entropy finding blocks.

qinhaihong-red:
## Two-axis review — Standards × Spec (#636)

Reviewed range: `df5130b0...84c806b7` (57 files, ~5,600 insertions). The two axes ran as independent reviews; findings are reported separately and not reranked.

**Overall**: the core implementation of #636 is solid — 10 of 13 acceptance criteria have real evidence, no material scope creep, and the new tests exercise the real packaged authority and the real CLI. Both axes independently converge on one family of root-cause issues: **host-private value matching / fallback logic and duplicated host constants**, which breach both the domain rules (hosts are conforming implementations, never authority) and the spec's explicit prohibition. That family (Standards #1/#2 + Spec c1/c2) should be fixed before merge; most other findings are follow-up judgement calls.

---

### Standards axis

#### Documented-standard breaches (hard)

1. **Two divergent copies of one value-contract law** — bADR-0016 Decision: "One closed type interpretation serves every boundary… A host cannot implement a smaller private matcher for one call path"; RULES.md (authoritative-source rule) forbids independently evolving duplicates. Base had two byte-identical matchers; this PR edits both with *different* rules:
   - `domain/authority/runtime_validation.py:16-20` — `if "value_kind" in actual or "value_kind" in formal:` (presence-based)
   - `domain/model/_lowering.py:3438-3444` — `if actual.get("value_kind") == "nominal-structured" or …` (value-based)
2. **Host-invented `value_kind`** — `domain/authority/admission.py:3277` returns `{"type": type_expression, "value_kind": "structured"}`. Bare `"structured"` occurs **zero** times in `kernel.json` or any package JSON (only `nominal-structured`), and `runtime_validation.py:1078-1082` refuses any formal whose `value_kind != "nominal-structured"`, so this contract can never match. Breaches AGENTS.md "no private value matcher… or fallback" and BALANCING-CONTEXT (LDB): "missing content cannot fall back to host behavior".
3. **Naming asserts host authority** — `domain/structured_values.py:32` `class StructuredValueAuthority` + `package_/language_/selected_structured_value_authority()`. This object is a read-only lookup projection of admitted types/operations/profiles — the glossary's **Admitted language index**, whose _Avoid_ list contains "generated authority". AGENTS.md: hosts are "conforming implementations, never authority".

#### Judgement calls (documented standards)

4. `domain/model/_resolution.py:788-812` `_preferred_schema_errors` drops non-preferred `oneOf/anyOf` branch diagnostics by a host heuristic (min pointer count, then `str(schema_path)`), gated on hardcoded `error_path[-2] == "symbols"`. bADR-0015: static stages "report all safely discoverable diagnostics"; sort order is authority-declared.
5. Doc drift: BALANCING-CONTEXT's rewritten "Literal Typing Profile" drops the "close against LDB value inventories" clause that `ARCHITECTURE.md:876` keeps. Also `ARCHITECTURE.md:655` is 136 chars vs the file's ~100 prose wrap, and 879-881 left a ragged re-wrap.
6. `fix(gda-balancing)` used for a test-hash refresh (84c806b7) and a CI shard entry (2a0e3c2d) — AGENTS.md wants truthful types on an independent release train. Scope guard OK (0 files outside the directory). No demo/engine references anywhere; new tests run the real packaged authority + real CLI (no mocks).

#### Baseline smells (judgement calls)

- **Duplicated Code** — `structured_values.py:102/159/216`: three near-identical projections; `execution.py:686-712` repeats the admit→`canonical_bytes` compare→mismatch block twice, and `experiment.py:125-160` a third time; `tests/schema2_bootstrap_conformance_support.py:6900` vs `:9139` reimplements ~250 lines verbatim.
- **Data Clumps** — `structured_authority` + `structured_resource_limit` travel together through ~8 signatures in 5 files; `["resources"]["max_rule_match_steps"]` re-derived in 5 files.
- **Repeated Switches** — `== "nominal-structured"` recurs 24× across 8 domain files.
- **Duplicated Code (inverse map)** — `admission.py:1758-1764` hand-maintains the inverse of `_contract_value_kind` (1398); `"object": "object"` does not round-trip (returns `None`).
- **Speculative Generality** — `execution.py:1250-1252`: `structured_authority=None` defaults that immediately `raise ValueError`; both call sites always pass them.

---

### Spec axis (vs #636)

Verified by reading the diff and running tests at head (`test_structured_values.py` 3 passed; `test_schema2_experiment_cli.py` + `test_schema2_model_lowerer_conformance.py` 166 passed).

**Broadly conformant.** ACs 1–5, 8, 10–13 are met with real evidence: kernel gains `typed-literal`/`lookup`/`equal` + `structured_operations`/`nominal_types`; bADR-0022 and ARCHITECTURE.md amended; the scenario runs `model check/build/inspect` + `experiment check/run`; AC8's nine required negative/boundary vectors exist and are executed by production **and** an independently written consumer (`_consumer_b_evaluate_structured_value_vector`, no `structured_values` import); record member-reorder neutrality is pinned.

#### (a) Missing / partial

1. **AC7 weakly pinned.** Spec: *"changes the structured result or derived Metric"*. In `test_public_structured_selection_...`, the reordered run asserts `reordered_metrics[...] == 9` — identical to the pre-reorder run's `9` and to the authored initial `selection_metric: 9` / `selected_result` (rank 9). The only discriminating assertion is the outcome flip. The README §5 metric change (3→7) needs an extra `rank` edit that **no test covers**.
2. **AC3 "resource bounds" not authority-owned.** `standard.schema@2.3.0` declares `structured_operations[].resource_bounds.max_steps: 1` — never read anywhere. The real budget is host-chosen and inconsistent: production hardcodes `resources.max_rule_match_steps` at 5 sites (`_admission.py:259`, `_lowering.py:2944`, `experiment.py:569`, `evidence.py:684`, `execution.py:1684`) while the vector test uses `max_runtime_projection_steps`.
3. **AC6 partial**: no assertion that the gameplay-alternative run *omits* the terminal-audit set (*"a terminal-audit artifact set only when applicable"*).
4. **AC3 minor**: `standard.schema@2.3.0`'s own new vectors are abstract `"valid"/"invalid"` diagnostic pairs; every executable constructor/lookup/equality vector lives in the scenario package instead of the constructor-owning authority.

#### (b) Scope creep

None material. Constructors are exactly Enum/Record/List/Ref; no Fixed/Decimal/Float/Vector/Set/Map/Distribution; no `game.generation|build|collection`; no reward/rarity/disposition policy. The extra nominal `CandidateRef` is the required `Ref<T>` instantiation — benign.

#### (c) Implemented but weaker than spec

1. **Private value matcher** — spec: *"must not add a private value matcher, serializer, evaluator table, or fallback."* Kernel declares `typed_envelope_profile.type_member`/`value_member`/`admission.envelope_members`, yet the host hardcodes `set(value) == {"type","value"}` (`structured_values.py` `admit_typed_value`, `formula/types.py:187`, `_lowering.py:3248`, `_admission.py:494`) and gates the law by equality against a copied `TYPED_ENVELOPE_ADMISSION_LAW`. `execution.py:_admit_declared_value` additionally shape-sniffs a dict as a *fallback* branch instead of keying on the declaration's `value_kind`.
2. **Duplicate host constant** (AC9 forbids these): `structured_values.py:776` hardcodes `{"id":"Boolean","package":"kernel","version":"2.0.0"}` though `equal_result_contract()` already resolves `kernel-boolean` and kernel.json declares the identity (also duplicated at `runtime_validation.py:506`).
3. A structured **state-write** fault becomes a bare `ValueError` (`execution.py:~2351`), not an authority-owned Diagnostic — a crash path rather than a refusal.

---

**Summary**: Standards axis — 11 findings (3 hard breaches; worst: #1, two now-divergent copies of one value-contract law). Spec axis — 7 findings (worst: c1, hardcoded envelope shape forming the private value matcher/fallback the spec forbids). Both worst findings share one root cause: the host does not consistently key on the admitted authority.

qinhaihong-red:
Addressed the latest two-axis review in `5b4ba48b` after evaluating each finding against #636 and the admitted Kernel/LDB contracts.

### Adopted

- Replaced the divergent Operation value-contract matchers with one shared closed relation.
- Removed the host-only `value_kind: "structured"`; structured contracts now use the admitted `nominal-structured` value kind.
- Renamed `StructuredValueAuthority` and its constructors to `StructuredValueIndex`, matching the glossary's admitted-language-index role.
- Made typed-envelope member names, fixed Boolean result type, value-node contracts, and structured-operation bounds come from the admitted Kernel/LDB projections. The host no longer owns a copied envelope law or Boolean coordinate.
- Removed declaration-bypassing envelope fallback behavior. Scalar result-envelope unwrapping now uses the selected typed-envelope profile and its declared members.
- Aligned the independent consumer with the same authority surfaces while keeping its implementation independent.
- Restored the Literal Typing Profile inventory-closure clause and corrected the cited architecture prose wrapping.
- Strengthened AC6/AC7 evidence: the gameplay-alternative path now asserts that no terminal-audit member is published, and the maintained reorder edit asserts the documented structured result and Metric change to `7`.
- Corrected the structured conformance-vector budget input to `max_rule_match_steps`.

### Partially adopted

- The resource-bound concern was valid, but a second budget mechanism was not warranted. Structured operations now verify their declared `resource_bounds.max_steps` against the selected Kernel node charge; recursive value admission continues to use the existing LDB rule-match budget.
- Fixed-value contracts are no longer accepted by comparing against a host copy. Admission validates their closed shape and requires the declared inventory to equal the contracts referenced by Kernel nodes and vectors.

### Not adopted

- The two `fix(gda-balancing)` commits repaired real CI/test failures, so their types remain truthful.
- Normative scenario vectors remain in `standard.conformance.structured`; this is an explicit human decision in #636, not accidental ownership drift.
- I did not change `_preferred_schema_errors`: the review identified a heuristic but did not demonstrate a safely discoverable independent diagnostic being lost. Expanding schema-diagnostic behavior here would be speculative.
- I did not convert an impossible admitted structured state-write mismatch into a Runtime refusal. The Kernel declares no such refusal for `write-state`; inventing one in the host would create the authority breach the review is trying to prevent.
- The baseline duplication/data-clump observations are not required to close #636 and were not expanded into a broad refactor.

Validation:

- bootstrap/authority/composition suite: `167 passed`
- lowerer/bootstrap high-risk regression suite: `182 passed`
- structured/public and existing scalar/Boolean Experiment slices: passed
- Pyright: `0 errors`
- Ruff lint and format: passed
- `git diff --check`: passed

Remote CI is now rerunning against `5b4ba48b`.

Issue #534, opened 2026-07-21
[gda-balancing] PRD: Standard Schema 2.0 language, runtime, and evidence architecture


qinhaihong-red:
## Refactor supersession — 2026-09-06

The approved refactor in #865 supersedes any assumption that internal package, Operation, Template, or toolkit versions are formally released and permanently immutable. Before the toolkit reaches v1.0, those internal definitions may be changed or withdrawn. v1.0 is the earliest point at which a compatibility commitment may be adopted; it does not automatically establish one. Content integrity, admitted-request/session consistency, exact semantic dependencies, and honest provenance remain required.

The refactor retains the language/compiler/Runtime direction and the full product scope. The old internal version-selection, universal historical-retention, whole-LDB propagation, build-provenance execution binding, and mandatory unused 1.x-converter clauses are superseded by #865. In particular, the old unused-package vector requiring Replay ineligibility is no longer the target contract.

**Mandatory completion gate:** After #874 closes every actual execution dependency, #875 must remove whole-LDB identity, Build receipt/compiler-build provenance, and equivalent unrelated inventory bindings from execution admission, Experiment eligibility, and Replay keys; provenance can remain separately inspectable but cannot cause those eligibility failures. The deletion issue cannot close with only dependency discovery, a new projection, an adapter, a bypass, a feature flag, or a second compatibility path. It must delete the obsolete fields/checks/readers/writers/reason paths and their exclusive tests/documentation, with mutation tests distinguishing harmless provenance changes from real execution changes. Artifact hashes must still truthfully identify their bytes.

The existing `Kernel`/LDB authority ownership and bounded deterministic execution remain; the currently implemented primitive set is a tested starting point, not an immutable completeness claim. The early non-RPG functional witness is owned by #878 before broad genre expansion and is reused by #575 for its complete coverage scope. Functional validation can publish `candidate`/open artifacts; #542–#544 remain deferred under their existing activation conditions. Their authenticated formal-claim requirements are not refactor completion prerequisites.

This dated update changes only the identified contract or scheduling clauses. The original requirements, issue identifiers, historical evidence, and still-valid acceptance scope remain below. It does not claim implementation or conformance completion.

---

## Problem Statement

[...]

PR #639, opened 2026-08-12
test(gda-balancing): prove seeded reward and build tuning

qinhaihong-red:
## Summary

- synchronize the Roguelike dogfood slice with the replacement Kernel/LDB baseline delivered by #640 through PRs #641, #642, #643, and #645
- retain the maintained Model Source, exact Experiment, Formula binding and explanation, seeded reward selection, typed empty-pool refusal, tested declared-fallback behavior, and atomic build replacement through public commands
- prove one complete subprocess designer loop: Model check/build/inspect, Experiment check/run, one authored `rare_weight` edit, recheck/rerun, and an explainable result and Metric change with the same RNG draw
- reconcile the tutorial and macro architecture with the adopted structured Operation primitives and the remaining Record-construction boundary

## Replacement-baseline result

The dogfood path now consumes the generic `operation-execution`, `is-empty`, `require`, and
top-level single-level `guard-block` contracts delivered by #640. Package-owned conformance
vectors prove the `game.generation` and `game.build` behavior. No reward-, build-, package-,
field-, example-, or genre-specific compiler or Runtime dispatch is added by this PR.

With seed `20260812`, stream `reward`, and draw value `3`:

- `rare_weight = 5` selects `volatile_crown` and produces reward/build Metrics `80/90`
- changing only `rare_weight` to `2` preserves the draw, selects `steady_guard`, and produces Metrics `20/30`

The public e2e test verifies the complete configure -> Model check/build/inspect -> Experiment
check/run -> edit -> recheck/rerun loop in separate CLI subprocesses. It also verifies the exact
Formula and Operation explanations, changed Experiment identity, unchanged RNG draw, changed
selection, and changed Metrics.

## Product feedback

- **Adopted refinement:** #640 replaced the insufficient integer-only vector and guard-chain
  baseline with package-owned structured execution evidence plus generic `is-empty`, `require`,
  and `guard-block` behavior. `RewardOption` pairs candidate and selection data.
  `no_reward_on_empty` permits an independent fallback; the checked-in Experiment declares none,
  while focused mutations and Package vectors prove the fallback outcome.
- **Confirmed public path:** Model and Experiment commands expose the Formula binding, Operation
  explanations, seed/draw, ordered options, policy state, selected reward, build decision,
  disposition, trace, and Metrics needed for the tuning loop.
- **Open boundary for HITL:** the Kernel Runtime-node vocabulary does not construct result Records.
  The Experiment authors selection, decision, and next-state Records, and the `game.generation`
  and `game.build` Operations validate their relationships before commit. The human owner must
  judge whether that bounded authoring cost is acceptable; this PR does not introduce a general
  Record-construction primitive.
- **Package-specific boundary:** `game.generation.select-reward-v1` is a binary selector. Its
  current use of indexes `0` and `1` is Package behavior, not a general collection limitation.

## Human review checkpoint

This PR supplies the technical foundation for the player-facing HITL checkpoint now tracked by
#646. It validates the public designer and architecture workflow, but it does not complete #585's
human acceptance. #646 must deliver the exported playable product before the product and
architecture owner records **accept**, **accept with conditions**, or **reopen** on #585.

This slice closes no Tracer, RPG, Roguelike, Variant, Extension, Replay, Evidence,
template-support, Core Extension Invariance, or cross-genre claim.

## Validation at `620fc215`

- `tests/test_e2e_cli.py`: 26 passed, including the complete Roguelike public subprocess loop
- focused seeded-selection and tuning evidence: 2 passed
- Ruff check and format: passed
- `git diff --check`: passed
- required CI: passed for this exact head

Refs #585
Follow-up: #646


qinhaihong-red:
Follow-up review closure for exact head `93455930`:

- Adopted the valid Standards finding. `game.generation` now closes the selected reward and rarity-policy relationships before either state write; `game.build` closes the reward, slot, decision, power, and score relationships before its three writes. Contradictions use the declared rollback outcomes and publish no result fact.
- Added public CLI coverage for 15 independent contradiction mutations (7 generation + 8 build). The cases share each expensive Model build but retain separate specifications, invocation keys, outputs, and labelled assertions.
- Kept the Runtime correction generic: it bridges bare integers and admitted scalar typed envelopes only after the existing `same-value-contract` and `runtime-numeric` admission checks. It adds no Kernel primitive, game/package/field dispatch, or boundary fallback.
- Rebuilt the LDB and four affected Models, then verified all six maintained exact-bound Experiments through public `experiment check`.

Independent exact-head re-review: Standards **no findings**; Spec **no findings**. Focused re-review passed (`7 passed`), `rebuild_schema2_ldb.py --check` passed, and all required CI checks are green.

The remaining #585 human owner accept / condition / reopen decision is intentionally still pending. It is the HITL evidence boundary for this PR, not an automated implementation claim.

qinhaihong-red:
## Review: domain-modular design + entropy

Two reviews were run against this slice: a domain-modular architecture review (completeness / orthogonality / extensibility, judged against BALANCING-CONTEXT.md, ARCHITECTURE.md, the #585 acceptance criteria, and the four existing `game.*` packages) and an entropy review (is the complexity proportionate to the current goal). Method: static review of all 20 changed files, both package semantic closures, the example artifacts, and the test inventory; the CI results claimed in the PR body were not independently re-run.

**Summary.** The design review passes: both packages follow the existing LDB topology exactly, boundaries are clean, dependencies are one-way. The entropy verdict is **Keep**: almost all complexity is driven directly by the #585 acceptance criteria, with no speculative mechanism. One finding spans both reviews and drives the recommendation below: the runtime has no record-construction node, so the slice realizes structured results through an **authored result mirror plus runtime guard chains** — and the dogfood report classifies this as an adopted realization and states "no gap was found", instead of recording the expressiveness boundary as a Gate 5 input as the #585 acceptance criteria require.

**Recommendation: accept with conditions.** All conditions are documentation-level; no code change is requested.

### Design review

Observed structure — all correct:

- `game.build@1.0.0 → game.generation@1.0.0 → {core.quantity, standard.runtime, standard.schema}`; acyclic, one-way. `game.build` reuses `RewardRef`/`RewardSelection`/`RewardDisposition` instead of redefining them.
- Responsibility placement matches the architecture: the Formula and its binding live in the Model Source; seed, candidates, policy, plans, and Metrics live in the Experiment; selection semantics live in `game.generation`; replacement semantics live in `game.build`.
- The only `src` change is a generic scalar-envelope integer projection in the Runtime, with no game- or field-specific branch.
- The bootstrap-authority test change (package index → id lookup) is a correct robustness fix.

Findings:

- **C1 — type surface wider than the operation.** `RewardPool` admits up to 16 candidates, but `select-reward-v1` is structurally bound to indexes 0/1 (binary threshold selection). The versioned operation id gives a clean evolution seam, but the package does not state this bound anywhere a consumer of the types would see it. Suggest one sentence in the package docs or tutorial.
- **C2 — "no gap found" is too strong** (core finding). The runtime operator vocabulary has no record construction, so operations cannot build a `RewardSelection`/`BuildDecision` at runtime; they select among pre-authored results and validate coherence. The path did support the example, but the cost shape of that support (D1, E1 below) is itself an observation that #585 requires to be recorded as an unresolved input to Gate 5, not absorbed into "No unresolved Kernel, compiler, evaluator, or public-command gap was found."
- **O1 — symbol overloading inside operation bodies.** `rare_index`/`common_index` serve three roles (candidate index, boolean marker, arithmetic literal 1/0), and the `equal → if → precondition-greater-than-or-equal` triple repeats 9× / 11× to emulate a boolean assertion. This is a downstream symptom of a missing boolean-precondition node; it does not break module orthogonality, but it reduces the readability of a published authority artifact.
- **O2 — `gameplay-alternative` now carries two meanings.** Existing precedent uses it only for real in-game results (`miss`, `insufficient-resource`). Here `candidate-mismatch`/`plan-mismatch` — contradictions in authored data — share the same kind. Runtime reporting is a defensible stage while no static cross-record check exists, but the shared outcome kind blurs the concept. Tutorial §9 already asks this question; suggest recording it as a future outcome-classification item.
- **D1 — authored derived data (DRY).** `selections[i]` mirrors `candidates[i]` field by field and carries derived values (`policy_after.draw_count + 1`, `selected_index`); `BuildPlan.decision` mirrors the `next_state`/`build_state` delta. The knowledge "what selecting candidate i yields" exists as two hand-edited copies reconciled by runtime guards. Consequence: the one-value tuning claim holds only for values outside the mirror (`rare_weight`). Editing a candidate's `reward_score` is a multi-site edit across `candidates` and `selections`; a missed site surfaces as a runtime `candidate-mismatch`. The tutorial exercises only `rare_weight`.
- **E1 — guard-chain scaling.** The operation bodies are 58/60 nodes versus 2–7 in the existing game packages; guard cost grows with fields × outcomes and counts against `max_steps`. Record as a known cost driver for the Gate 5 Core Extension Invariance decision.
- **E2 — `canonical-equal` weakening (minor).** When either operand carries an admitted integer, equality now compares the projected integers; the type member's value no longer participates. Static typing should keep operands same-typed, so the practical risk is low; worth one recorded sentence.

### Entropy review

**Verdict: Keep.** Every new obligation maps to a #585 acceptance criterion: the six failure/alternative paths are required rows; the 707 test lines map ~1:1 to the criteria; the tutorial matches the established example pattern; `relaxed-pool`, `no-reward`, and `conflict` are all exercised by tests; the runtime projection is ~50 generic lines. No speculative generality found.

- Full-field guard chains: **keep for this slice** — removing them would break the "contradictory authored results" acceptance row. The smaller long-term alternative (guard only the discriminating fields; move the rest to a static cross-record check at `check` time) should be **deferred as a Gate 5 input**, not built now.
- Dual-representation integer projection: **keep** — normalizing at the lookup boundary is not smaller once `canonical-equal`/`write-state` typed-value needs are counted; the mixed-representation error branch is unreachable by construction and cheap.
- The "no gap found" wording: **simplify** — this is the one real entropy source, because it lets the mirror-plus-guards pattern silently become the template for future genre packages. The fix is one documentation line (see conditions).

Assumptions worth testing later: (1) designers can maintain mirror coherence — cheapest probe is a tutorial exercise that edits a candidate score; (2) the guard pattern scales to N candidates — test with a 4+-candidate example during Gate 5 revalidation.

### Conditions for accept

1. Record "no runtime record construction → authored result mirror + guard chains" as an unresolved product/architecture input to Gate 5, and reword the tutorial §8 claim "No unresolved … gap was found" accordingly (C2).
2. Qualify the one-value tuning claim: it holds for values outside the mirror; mirrored values require multi-site edits (D1).
3. Optional: record the `gameplay-alternative` dual meaning (real gameplay outcome vs. authored-data contradiction) as a future outcome-classification item (O2).

No code changes requested; everything else can merge as-is.

qinhaihong-red:
Handled the latest design/entropy review against current head `f4df2ba9` (the comment was written before the documentation-only `0f4d92a0` update). Baseline: current `origin/main` tip `7b075aea`, merge base `b23558c2`; the PR remains mergeable.

| C1: `RewardPool` is wider than `select-reward-v1` | Fully Adopt | Both lists admit up to 16 entries; the Operation can select only indexes 0/1. | The tutorial now states the binary bound, matching-entry requirement, and treatment of later entries. | Checked against the exact Package Release. |
| C2 / D1 / E1: record-construction boundary creates authored mirrors and scaling guard chains | Fully Adopt | The admitted Runtime vocabulary has lookup/equality/select/write nodes but no Record-construction node. Selection/build results are authored and guarded before commit. | The tutorial records this as an `Unresolved product/architecture gap`, qualifies the one-field edit, and assigns Gate 5 re-evaluation without adding a new mechanism now. | `git diff --check`; Package Release bounds and current tutorial path checked. |
| O1: guard-chain repetition suggests a direct boolean-precondition capability | Partially Adopt | The readability and scaling cost is real, but #585 explicitly forbids adding a new core/runtime mechanism to make this bounded slice pass. | Recorded the guard cost as a Gate 5 input. No Kernel node, Runtime phase, compiler dispatch, or evaluator dispatch was added. | Documentation-only diff. |
| O2: `gameplay-alternative` covers both gameplay branches and authored-data contradictions | Fully Adopt for documentation | bADR-0016 defines gameplay outcomes as expected game-semantic branches, while `candidate-mismatch` / `plan-mismatch` report authored-data contradictions. | Added a separate unresolved outcome-classification row and an explicit HITL question. The current typed outcome and rollback remain unchanged. | Checked current package outcomes and bADR-0016. |
| E2: integer projection weakens `canonical-equal` | Do Not Adopt | Admission requires an exact shared `same-value-contract`; numeric execution additionally requires an admitted `runtime-numeric` policy. Runtime never accepts this operation body without that closure. | Retained the generic integer projection and did not couple the tutorial to a host implementation detail. Residual risk exists only if the admitted-RIR boundary is bypassed, which is not a supported public path. | Existing exact-head CI covers admission and Runtime paths. |
| Old custom dogfood classifications | Already closed | Current head already maps observations to the four #585 feedback classes. | No duplicate change. | Verified at `0f4d92a0` and retained in `f4df2ba9`. |

Commit: `f4df2ba9` (`docs(gda-balancing): record roguelike architecture inputs`). All required exact-head CI checks pass; Godot e2e is intentionally skipped for this package-only PR.

qinhaihong-red:
## Re-review: conditions verified at `f4df2ba9`

Rechecked the two documentation-only commits since the reviewed head `93455930` (`0f4d92a0`, `f4df2ba9`). All three accept conditions are satisfied; the review side of "accept with conditions" is resolved.

- **Condition 1 (C2/D1/E1)** — satisfied. Tutorial §2 records the record-construction boundary; §8 replaces "no gap was found" with two `Unresolved product/architecture gap` rows (mirror-and-guard cost, outcome classification) that assign Gate 5 re-evaluation, and the classification column now uses the four #585 feedback classes.
- **Condition 2 (D1)** — satisfied. §5 states why the `rare_weight` edit is one-field and that a mirrored field such as `reward_score` requires the corresponding selection edit, else `candidate-mismatch`.
- **Condition 3 (O2)** — satisfied. Separate unresolved row plus the new authored-data-contradiction entry in the §9 HITL question list.
- **C1** — satisfied. §2 states the binary-selector bound and that later entries do not participate.
- **O1 partial adoption** — appropriate; no new mechanism was requested, and the guard cost is recorded as a Gate 5 input.
- **E2 Do Not Adopt** — accepted. Verified in `domain/authority/admission.py` (~line 3642): admission enforces a shared `same-value-contract` on `canonical-equal` operands, so divergent envelope types cannot reach the projected-integer comparison through a supported public path.
- Terminology spot-check: `Named random stream` and `Discriminated gameplay outcome` are BALANCING-CONTEXT.md glossary terms; the capitalization follows the glossary convention.
- All required CI checks pass on the exact head; the PR is mergeable. Godot e2e skip is expected for a package-only PR.

Non-blocking note: when the Gate 5 tracking issue is next groomed, link the two unresolved §8 rows from it so the inputs are consumed from the durable artifact rather than rediscovered.

The #585 accept / accept-with-conditions / reopen decision remains with the human product/architecture owner.

qinhaihong-red (commented):
## Two-axis review — Standards and Spec (diff `origin/main...f4df2ba9`)

**Summary.** Both axes converge on the new integer-projection helpers in `src/gda_balancing/domain/runtime/execution.py`. Standards axis: 2 hard findings (runtime semantics live only in host code; issue-owned content in a permanent document) plus 4 judgment calls. Spec axis: AC1–AC6 and AC9 verified pass; AC7/AC8/AC10 each have one deviation, and the `canonical-equal` change is the one substantive problem.

### Standards

**S1 (hard) — Runtime semantics exist only in host code.**
`execution.py` adds `_runtime_integer_value` / `_runtime_integer_operand`, so `integer-add/maximum/multiply`, `integer-compare`, `canonical-equal`, and `gameplay-precondition` now accept a typed envelope `{type,value}` as an operand. But `authorities/kernel.json` attaches the typed-envelope law only to `literal_typing` (authoring side); the `add`/`compare` nodes' `operand_constraints` remain `runtime-numeric`, and neither new package's conformance vectors pin the new operand admission. Per `libs/gda-balancing/AGENTS.md` (Development conventions): "the Schema-major Kernel Specification defines bundle interpretation and irreducible semantics … Host implementations are conforming implementations, never authority" (see also bADR-0022). An independent implementation cannot derive this behavior from the Kernel/LDB. Relatedly, the example README §8 claims "The Runtime implementation gap is closed" under "the admitted scalar contract" — a term with zero hits anywhere in the repo.

**S2 (hard) — Issue-owned content written into a permanent document.**
`examples/schema2/roguelike-reward-build/README.md` §8 (observation classification table + Gate 5 action items) and §9 (human review checkpoint). `docs/ARCHITECTURE.md:1404-1406` says: "PRD #534 and the linked issues own detailed observations, acceptance criteria, and live completion status." §9's "Until that review records its result" becomes a stale statement the day this merges. The README also lacks the closing authority back-link its sibling has (`structured-selection/README.md:241-244`) and near-verbatim duplicates the non-claim list with the new ARCHITECTURE entry (single-authority / DRY rule in RULES.md).

**S3–S6 (judgment calls).**
- Naming: `_runtime_integer_value` (returns `None`) vs `_runtime_integer_operand` (raises) — the names do not convey the optional-vs-required difference.
- `tests/test_schema2_experiment_cli.py`: two `if contradiction == "rarity": … elif …` cascades, and a six-times-repeated `next(row for row in specification["scenarios"][0]["assignments"] if …)["value"]["value"]` chain; the suite's existing `parametrize` idiom fits better.
- Guard inconsistency (potential crash): `_runtime_integer_value` checks only `structured_authority is None`, unlike the existing sites (`execution.py:706-711`, `838`) which first check `typed_envelope_profile is not None`; with a `None` profile, `typed_envelope_members` raises `ValueError`.
- Positive: `_package_vector_subject` looks vectors up by package id, removing the magic `packages.3` index from tests.

### Spec (issue #585)

Verified pass: AC1–AC6 and AC9 — public `model`/`experiment` commands end to end; packages under `authorities/packages/` next to the existing `game.*` releases (no private fixture path); static Formula binding plus the mandatory Model explanation; the one-value loop `rare_weight 5→2` with the draw held at 3 and Metrics `80/90 → 20/30`; the trace exposes seed/stream/draw index, ordered candidates, policy state before/after, and the replacement decision; empty pool → typed refusal plus rollback; `no-reward` / `build-conflict` land as `gameplay-alternative`; the README honestly discloses the mirror-and-guard gap.

**P1 (AC7, scope) — `canonical-equal` silently bypasses exact-type checking.**
The fast path in `execution.py` does not just project the integer an admitted envelope carries: when both operands project to an int, it skips the exact-type comparison entirely (previously `structured_values.py:867-872` faulted on a `type` mismatch). Two integers of different nominal type/unit now compare equal silently — a semantic relaxation of admitted nodes, and the very behavior this example depends on. Operand projection has backing (`ARCHITECTURE.md:1159`); the equality relaxation has none. AC7 requires either stopping and recording an unresolved input to Gate 5, or adopting the refinement and updating the authority — neither happened.

**P2 (AC6, correctness) — the new failure path escapes as EXIT_INTERNAL.**
`raise ValueError("admitted equality operands used different representations")` has no matching `except ValueError` in `execution.py` (line 2337 catches only `StructuredValueFault`), so it propagates to `interfaces/cli/dispatch.py:90` → `internal_envelope` / EXIT_INTERNAL. Before this change the same situation raised `StructuredValueFault("language.structured_value_type_mismatch")` — a typed refusal. If this is reachable from authored input it violates AC6 ("build conflict and invalid configuration are not mislabeled as internal failures"). The `ValueError` in `_runtime_integer_operand` has the same problem. Needed: either evidence that static admission blocks mixed representations, or a typed refusal plus a regression test.

**P3 (AC8) — the affected authority was not updated.**
The runtime semantic change is recorded only in the example README §8 observation table (classified "Confirms current design"); `docs/ARCHITECTURE.md:697` still says "exact-type canonical equality", and `standard.runtime@1.1.0` plus its conformance vectors are untouched.

**P4 (AC10) — the non-claim set is incomplete.**
The README lists 6 of the 10 required non-claims (missing Tracer, RPG, Variant, Extension); the new ARCHITECTURE entry misses Tracer and RPG. The PR body's "all required non-claims" overstates.

**P5 — the new refusal paths are untested.**
Zero hits for `different representations` / `did not carry an integer` under `tests/`.

---
Method: two independent review axes (documented standards plus a fixed smell baseline; spec fidelity against #585), run as separate reviewers and reported per axis without cross-ranking.

qinhaihong-red:
Handled the latest review at exact head `2cdf0688`.

| Review item | Decision | Evidence and handling |
| --- | --- | --- |
| P1: projected integer equality skipped exact type | **Fully adopt** | Reproduced the defect at the prior head. Runtime now rechecks the complete envelope type before comparing two projected integers. A different nominal type returns `language.structured_value_type_mismatch`; legal `Quantity` equality remains unchanged. |
| S3 / S6: helper intent and missing envelope-profile guard | **Fully adopt** | Renamed the helpers to `_project_runtime_integer` and `_require_runtime_integer`; projection now returns not-applicable when no typed-envelope profile is selected. |
| S2: issue-owned details and stale HITL text | **Fully adopt** | Removed the detailed classification table, review checklist, and live PR status from the permanent README. The README retains only stable facts and links to the macro architecture. The detailed dogfood record, owner/actions, questions, and live HITL boundary now live in [issue #585](https://github.com/aigengame/godot-agent/issues/585#issuecomment-5264590300). |
| P4: incomplete non-claim set | **Fully adopt** | The maintained README now lists all ten required non-claims once. `ARCHITECTURE.md` keeps only the macro summary and delegates the exact set to the example, avoiding a second drifting list. |
| P2 / P5: possible internal errors and missing refusal evidence | **Partially adopt** | Added a production + independent-admission mutation proving cross-contract equality is rejected before Runtime, plus a public `experiment check` regression proving a non-integer `Quantity` is a static typed refusal. Internal `ValueError` branches remain host-invariant failures because no admitted public RIR can reach them; they were not promoted into a new public refusal contract. |
| S1 / P3: semantics exist only in host code and require an authority change | **Do not adopt as stated; adopt the prose correction** | The behavior is already closed by Kernel `same-value-contract` / `runtime-numeric`, the LDB-owned `core.quantity` `exact-integer` rule, and the documented rule that lookup returns a typed envelope. The host helper bridges admitted runtime representations; it does not add a language rule. Exact-type equality is now preserved, so no Kernel/LDB identity change is warranted. The undefined “admitted scalar contract” wording was removed. |
| S4: parametrize the contradiction loops | **Do not adopt** | The loops intentionally share one expensive Model build while still producing independent specifications, invocation keys, outputs, and labelled assertions. Parametrization would repeat the build for every mutation and reverse the test-time optimization. |
| S5: package-vector lookup improvement | **No action** | Positive observation; retained unchanged. |

Commit: `2cdf0688` (`fix(gda-balancing): preserve typed scalar contracts`).

Validation:

- full `tests/test_schema2_experiment_cli.py`: **160 passed**;
- full `tests/test_schema2_model_cli.py`: **166 passed**;
- pyright: **0 errors**;
- Ruff and `git diff --check`: passed;
- all required CI checks on `2cdf0688`: passed (Godot e2e is the expected package-only skip).

qinhaihong-red (commented):
## Re-review: fixes verified at `2cdf0688`

Every claim in the handling table was checked against the actual code, not the reply. Verdict: the review side is resolved; nothing remains that should block the human accept decision.

- **P1 — fixed.** The `canonical-equal` integer fast path now compares the complete envelope type (`canonical_bytes(left[type_member])` vs right) before comparing values, and a mismatch raises `StructuredValueFault("language.structured_value_type_mismatch", "/type")` — the same typed fault as before the PR. Pinned by `test_runtime_canonical_equality_rechecks_typed_envelope_identity` (Quantity vs RewardRarity, both carrying `1`). The mixed bare-int-vs-envelope case performs no dynamic type compare, which is sound because the Kernel `equal` node's `same-value-contract` operand constraint makes the operand types statically identical, and the new admission mutation test shows a cross-contract rewrite is refused before Runtime.
- **S3/S6 — fixed.** Helpers renamed to `_project_runtime_integer` (returns `None`) / `_require_runtime_integer` (raises); the projection now also returns not-applicable when `typed_envelope_profile is None`, pinned by `test_runtime_integer_projection_without_an_envelope_profile_is_not_applicable`.
- **S2 — fixed.** The observation classification table, review-question list, and the stale "Until that review records its result" text are gone from the README; §8 is now a stable "Product-feedback boundary" that delegates detailed observations and the live HITL decision to issue #585 (the receiving comment exists and carries the moved table), and the closing authority back-link is present.
- **P4 — fixed.** The README states all ten non-claims once; `ARCHITECTURE.md` keeps a macro summary ("closes no coverage, Evidence, template-support, Core Extension Invariance, or cross-genre claim") and delegates the exact set to the example, so no second enumerated list can drift.
- **P2/P5 — partial adoption verified.** `test_package_admission_rejects_canonical_equality_across_value_contracts` proves `admit_authorities` refuses cross-contract equality (with the composition diagnostic subject present), and `test_public_reward_configuration_rejects_a_non_integer_quantity` proves a non-integer `Quantity` in authored input is a static typed refusal with an exact pointer — not EXIT_INTERNAL. Given those, the remaining `ValueError` branches require an already-violated static invariant; classifying them as host-invariant internal failures is defensible.
- **S1/P3 — decline holds.** Verified in the authorities: the Kernel `equal` node carries `{"kind": "same-value-contract"}` (and `add` additionally `runtime-numeric`); `ARCHITECTURE.md` states "Lookup returns a typed envelope …" and "Equality first requires the same canonical type and then compares canonical values", which the fixed code now matches; `admission.py:3642` enforces the constraint. With exact-type equality restored, the host helper is a representation bridge, not a language rule, and the undefined "admitted scalar contract" wording is gone. One non-blocking residual: the envelope-as-numeric-operand behavior is pinned by host tests rather than a conformance vector — worth carrying as a Gate 5 input alongside the two existing unresolved rows.
- **S4 — decline holds.** `_write_built_roguelike_experiment` runs the expensive Model build once; the loop deep-copies the baseline per mutation. Naive parametrization would rebuild per case; the stated rationale is factually accurate.

No new problem introduced by `2cdf0688` was found. All required CI checks pass on the exact head; the PR is MERGEABLE/CLEAN (Godot e2e is the expected package-only skip).

The #585 accept / accept-with-conditions / reopen decision remains with the human product/architecture owner.

qinhaihong-red:
Reviewed exact head `2cdf0688b1958f31a47c05ac233d93d58f7205b1` against PR base `b23558c2c2add7af6b2fdcbb452147f6ed0a09aa`.

## Standards

1. **`game.build` declares an RNG effect that its Operation does not have.** `game.build.replace-reward-v1` contains no `draw` node or RNG-bearing `invoke`, but declares `rng.named-stream`; `build.replace.effects` then fixes that declaration in a vector. bADR-0016 and bADR-0022 require the effect surface to be derived from the program and to agree exactly with the declaration. Remove `rng.named-stream`, update the vector, and refresh affected identities.

2. **The two new packages do not ship the package-owned conformance evidence required for LDB admission.** The `game.generation` and `game.build` vector sets contain dependency, one Enum positive/negative pair, effects, and resource vectors only. bADR-0016 requires positive, negative, boundary, compatibility, deterministic replay, and declared-effect evidence before a package enters the LDB. Add applicable Operation success/gameplay-alternative/rollback cases, list/resource boundaries, compatibility, and replay vectors, then rebind them through the manifests. External pytest coverage is not a manifest-bound normative vector child.

## Spec

1. **[P1] The declared reward fallback is not actually implemented or exercised.** `game.generation.select-reward-v1` never branches on `RarityPolicyKind`; it only verifies that the before/after kinds match and exposes `selected` / `candidate-mismatch`. The fallback test replaces an ordinary candidate at index 1 with `no_reward`, after which normal weighted selection succeeds; only the downstream build Event reports the `no-reward` gameplay alternative. This does not exercise an unsatisfiable-selection fallback or establish `relaxed-pool` eligibility/exhaustion semantics required by #585. Make `game.generation` declare and execute the fallback policy and emit the corresponding gameplay outcome. If the current public contracts cannot express it, follow AC7 and record the Gate 5 gap instead of presenting a sentinel candidate as fallback.

Validation: `git diff --check` passed and all required exact-head CI checks are green. The pending human product/architecture decision is intentionally retained by #585 and is not counted as an implementation defect.

Summary: Standards 2 findings (worst: declared Operation effects disagree with derived closure); Spec 1 finding (declared fallback is not implemented).

qinhaihong-red:
Implemented and independently reviewed at exact head `b1c5d8c6` against
`origin/main@807ffc82`.

The requested `$review` ran separate Standards and #585 Spec axes.

- **Standards:** the first pass found one minor ambiguous reference to “this Package Operation.”
  Commit `b1c5d8c6` now names `game.generation.select-reward-v1` and limits the binary-selector
  statement to package-specific behavior in this release.
- **Spec:** no technical finding. The replacement baseline correctly consumes #640 without
  reimplementing it or widening any Kernel, Runtime, collection, or cross-genre claim.
- **Final re-review:** Standards 0 findings; Spec 0 findings; `git diff --check` passed; the key
  public subprocess e2e passed independently.
- **CI:** all required checks passed at the exact head, including the six gda-balancing shards,
  inventory closure, wheel/subprocess smoke, Pyright, Ruff, build, and co-install checks.

The only remaining gate is #585 AC9: the human product/architecture owner must record **accept**,
**accept with conditions**, or **reopen**. This is a deliberate HITL decision, not an automated
test gap.

libs/gda-balancing/examples/schema2/roguelike-reward-build/README.md
```diff
@@ -367,6 +367,11 @@ This maintained example preserves the stable facts that a reader needs to unders
 - empty-pool and build-conflict paths are declared gameplay outcomes; and
```
qinhaihong-red: **[spec + architecture — substantive]** §8 lists this as a stable fact, but the maintained `experiment.json` ships `"no_reward_on_empty": []` — no declared fallback — so the maintained configuration's empty-pool path is the `game.generation.selection_exhausted` typed refusal, exactly as row 1 of the §7 table above says. The declared-fallback gameplay outcome exists only in the focused-test mutations and the package vectors, not in the shipped artifact. #640 step 10 asked to "remove stale fallback claims and mirrored data".

Two consistent fixes; pick one:
1. Reword here, in the intro (line 22 "an explicit empty-pool fallback"), in §2 (line 56), and in the PR body: the fallback capability is proven by mutation tests and vectors, and the maintained artifact deliberately declares none.
2. Declare a real no-reward fallback in the shipped `experiment.json` so the prose becomes true.

libs/gda-balancing/examples/schema2/roguelike-reward-build/README.md
```diff
@@ -367,6 +367,11 @@ This maintained example preserves the stable facts that a reader needs to unders
...
 - contradictory option and plan data are typed runtime refusals with terminal rollback evidence.
 
+The replacement-baseline rerun confirms the `operation-execution`, `is-empty`, `require`, and
```
qinhaihong-red, line 378: **[standards — DRY, judgement call]** Two mirrors in the added paragraph: (1) the binary-selector sentence restates §2's "considers only indexes `0` and `1`" within the same document — in a step whose brief was to remove mirrored data; (2) "keeps Record construction as an open product and architecture input" is a third home for that fact (with the §8 bullet above and the ARCHITECTURE.md open-boundary bullet). The README's own closing paragraph already assigns the macro boundary to ARCHITECTURE.md — one sentence pointing there would do.

libs/gda-balancing/docs/ARCHITECTURE.md
```diff
@@ -1540,8 +1540,10 @@ issues own detailed observations, acceptance criteria, and live completion statu
     independent consumers agree on the admitted Operation vectors. The maintained Roguelike path
     runs these Operations, and affected authority and example identities are rebuilt against the
     replacement Kernel.
-  - Open boundary: PR #639 must synchronize this replacement evidence before the #585 HITL
-    acceptance decision. Evidence bound to the superseded Kernel identity does not carry forward.
+  - Open boundary: The synchronized designer loop still authors result Records because the Runtime
```
qinhaihong-red: **[standards — terminology, minor]** bADR-0022 writes "Kernel Runtime-node vocabulary" / "closed Runtime-node grammar" (hyphenated); this bullet introduces "Runtime node vocabulary". Also "Package Operations" is not a BALANCING-CONTEXT.md glossary term (the glossary has Operation and Package Release). Worth aligning while the bullet is being touched.

libs/gda-balancing/tests/test_e2e_cli.py
```diff
@@ -343,17 +343,42 @@ def test_formula_model_build_then_inspect_key_path(self, tmp_path):
...
+            str(receipt_path),
+            "--format",
+            "indented",
+        )
+        assert (inspected.returncode, inspected.stderr) == (0, ""), inspected.stdout
+        explanation = json.loads(inspected.stdout)
+        assert {row["id"] for row in explanation["formula_explanations"]} == {
+            "rare-threshold"
+        }
+        assert {row["id"] for row in explanation["operation_explanations"]}.issuperset(
```
qinhaihong-red: **[standards — weak assertion, judgement call]** Asymmetric with the exact-set assert on `formula_explanations` just above: `issuperset` lets an unexpected extra Operation explanation pass, which is weaker than #585 AC2's "inspects the exact Formula/Operation boundary". If the explanation inventory is closed for this model, assert the exact set like the sibling `test_formula_model_build_then_inspect_key_path` does.

libs/gda-balancing/tests/test_e2e_cli.py
```diff
@@ -363,32 +388,88 @@ def test_roguelike_reward_build_key_path(self, tmp_path, monkeypatch):
         _bind_experiment_to_build(experiment, build_receipt)
         experiment_path = tmp_path / "roguelike-experiment.json"
         experiment_path.write_text(json.dumps(experiment), encoding="utf-8")
+        checked_experiment = _run("experiment", "check", str(experiment_path))
```
qinhaihong-red: **[spec — tutorial fidelity, minor]** `_bind_experiment_to_build` (line 388) rewrites the identities before this `check`, so the e2e never validates the checked-in `experiment.json` as shipped, while README §4 tells the reader to check that exact file. Drift is still caught by `_write_built_roguelike_experiment`'s identity-equality assert, so there is no verification hole — but checking the shipped file before binding would match the tutorial it proves.

libs/gda-balancing/tests/test_e2e_cli.py
```diff
@@ -363,32 +388,88 @@ def test_roguelike_reward_build_key_path(self, tmp_path, monkeypatch):
...
-        assert [event["operation"] for event in transitions] == [
+
+        tuned = json.loads(json.dumps(experiment))
+        tuned["id"] = "roguelike.reward-build-feedback.lower-rare-weight"
+        next(
+            row
+            for row in tuned["scenarios"][0]["assignments"]
+            if row["target"]["name"] == "rare_weight"
+        )["value"] = 2
+        tuned_path = tmp_path / "roguelike-tuned.json"
```
qinhaihong-red, line 414: **[standards — duplication, judgement call]** `_run_experiment_variant(name="roguelike-tuned")` writes this same `roguelike-tuned.json` path again with identical bytes; meanwhile the baseline is checked at `roguelike-experiment.json` but run from `roguelike-baseline.json`. Folding the `experiment check` call into `_run_experiment_variant` would remove both the double write and the check-path/run-path split, and shrink the check→build→inspect→check→run spine this test shares with `test_formula_to_experiment_public_key_path`.

libs/gda-balancing/tests/test_e2e_cli.py
```diff
@@ -363,32 +388,88 @@ def test_roguelike_reward_build_key_path(self, tmp_path, monkeypatch):
...
+        assert [event["outcome"] for event in baseline_events] == [
+            {"id": "selected", "kind": "success"},
+            {"id": "replaced", "kind": "success"},
+        ]
+        assert [event["outcome"] for event in tuned_events] == [
             {"id": "selected", "kind": "success"},
             {"id": "replaced", "kind": "success"},
         ]
 
+        def reward_result(events: list[dict]) -> dict:
```
qinhaihong-red, line 450: **[standards — readability, judgement call]** The `["value"]["value"]` double unwrap (fact envelope → structured-value envelope) is the only non-obvious navigation in an otherwise comment-free file; a one-line comment naming the two envelopes, or unpacking in two named steps, would help the next reader.

libs/gda-balancing/tests/test_e2e_cli.py
```diff
@@ -363,32 +388,88 @@ def test_roguelike_reward_build_key_path(self, tmp_path, monkeypatch):
...
+            return next(
+                row for row in events[0]["facts"] if row["name"] == "reward_result"
+            )["value"]["value"]
+
+        assert reward_result(baseline_events)["selected"] == {"key": "volatile_crown"}
+        assert reward_result(tuned_events)["selected"] == {"key": "steady_guard"}
+        assert baseline_events[0]["rng_draws"] == tuned_events[0]["rng_draws"]
```
qinhaihong-red, line 459: **[spec — AC5 evidence, low severity]** The tuned-side comparison stops at outcomes, selected key, draw value, and Metrics: stream name, draw index, ordered candidates, policy state before/after, and disposition are asserted only by the baseline-only focused test (`test_public_seeded_reward_selection_exposes_policy_and_disposition`), and README §5's "the build Event replaces the slot with `steady_guard`" is not directly asserted for the tuned run. The observable-change leg of the loop holds; a disposition/replacement assert on `tuned_events[1]` would close the gap.

qinhaihong-red (commented):
Three-axis review of this sync slice — domain modular architecture, repo standards, and spec fidelity (#585 + #640 step 10).

Verified against the diff: required CI is green at `b1c5d8c6`; the body's numbers (seed `20260812`, stream `reward`, draw `3`, `rare_weight` 5→2 flipping `volatile_crown`→`steady_guard`, Metrics 80/90→20/30) match the artifacts and the new assertions; #641–#645 all sit below the merge-base; the retired "PR #639 must synchronize" boundary is correctly replaced by the real remaining Record-construction boundary; the extended e2e drives only real CLI subprocesses; and the ownership chain (ARCHITECTURE §12.2 entry ← example README ← issue #585, e2e in the `smoke` shard) holds with one-way references.

**One substantive finding, where all three axes converge** (inline on README §8): the maintained `experiment.json` declares no fallback (`"no_reward_on_empty": []`), while the README (intro line 22, §2 line 56, §8 line 367) and this PR body ("retain the … declared no-reward fallback") still claim a declared one. #640 step 10 asked exactly to "remove stale fallback claims".

Not anchorable inline — **[standards]** the squash title: the net diff is two markdown files plus one test file, and `libs/gda-balancing` is its own release-please package, so `feat(gda-balancing)` would cut a minor release with no user-visible feature in it. `test(gda-balancing): …` (96 of the 111 added lines are test code) or `docs(…)` would be truthful; the feature content already released through #641–#645. Precedent #613 titled a dogfood slice `feat`, but there the slice's code landed in the same PR.

The remaining inline comments are low-cost cleanups. None of this gates the #585 HITL decision — that acceptance stays with the human owner.

qinhaihong-red:
Addressed the latest review at `6e11b038`.

- Corrected the fallback claims in the tutorial and PR body. The checked-in Experiment declares no fallback; an empty selection therefore takes the typed Runtime-refusal path. The separate declared-fallback behavior remains proven by focused tests and Package vectors. I did not add a fallback merely to match the prose.
- Removed the repeated binary-selector and Record-construction explanations, and replaced the ambiguous architecture terms with `Kernel Runtime-node vocabulary` plus the exact `game.generation` and `game.build` Operations.
- Replaced the weak Operation-explanation superset assertion with the exact seven-Operation reachable closure. The review's proposed two-entrypoint set omitted the five transitive `quantity.*` Operations exposed by `model inspect`.
- The public e2e path now checks the checked-in `experiment.json` before creating the tuned variant, and the nested reward-value access now uses named steps.
- I did not move `experiment check` into `_run_experiment_variant`: that helper has about 18 call sites, so doing so would add unrelated subprocess checks across the e2e suite. The checked-in and tuned files remain checked explicitly at the relevant call site.
- I did not duplicate the full tuned AC5 assertions in this e2e test. `test_public_seeded_reward_selection_exposes_policy_and_disposition` already proves the stream, index, ordered candidates, policy transition, and disposition; `test_public_reward_tuning_changes_selection_without_changing_the_rng_draw` proves the unchanged draw and exact tuned atomic replacement (`starter_blade` -> `steady_guard`, power `10` -> `30`).
- Updated the PR title to `test(gda-balancing): prove seeded reward and build tuning` and reconciled the PR body with the checked-in artifact and current boundary.

Validation at `6e11b038`:

- `tests/test_e2e_cli.py`: 26 passed
- focused seeded-selection and tuning evidence: 2 passed
- `git diff --check`: passed

Required CI is running for this exact head.

qinhaihong-red:
Follow-up: the first pushed review-fix head exposed a Ruff formatting-only failure. I applied the repository formatter and pushed `620fc215`. Local Ruff check/format and the Roguelike key-path test now pass; the PR body is updated to the new exact head, and required CI has restarted.

qinhaihong-red:
Re-checked the review-fix heads `6e11b038` + `620fc215`. Every finding is resolved or declined with evidence I could verify; nothing from my review blocks the #585 HITL decision.

- **Fallback claims (substantive)** — fixed. README intro/§2/§8 and the PR body now match the shipped `"no_reward_on_empty": []`: empty selection without a fallback is the typed refusal, declared-fallback and build-conflict paths are the gameplay outcomes, and the declared-fallback behavior is attributed to focused tests and Package vectors. Keeping the artifact unchanged instead of adding a fallback to match the prose was the right direction.
- **DRY mirrors** — fixed. The binary-selector fact now has one home (§2, with the exact Operation id) and the third Record-construction restatement is gone.
- **Terminology** — fixed in ARCHITECTURE.md ("Kernel Runtime-node vocabulary", the exact `game.generation`/`game.build` Operations). Residual nit, non-blocking: README line 81 still uses the unhyphenated "Runtime node vocabulary"; fine to leave for a later docs pass.
- **Operation-explanation assertion** — fixed better than suggested: the exact seven-Operation reachable closure. You are right that my proposed two-entrypoint set omitted the five transitive `quantity.*` explanations.
- **Tutorial fidelity** — fixed beyond the suggestion: `experiment check` now runs on the checked-in `experiment.json` as shipped, and the bind step is gone; determinism of the bound identities stays guarded by `_write_built_roguelike_experiment`.
- **Fold `check` into `_run_experiment_variant`** — decline accepted; the 18-call-site count checks out, and the fold would indeed push subprocess checks into unrelated tests.
- **`reward_result` unwrap** — fixed with named envelope steps.
- **Tuned-side AC5 evidence** — decline accepted with verified evidence: `test_public_reward_tuning_changes_selection_without_changing_the_rng_draw` asserts the tuned selection (`steady_guard`, index 0, score 20), the exact tuned replacement (`starter_blade` → `steady_guard`, power 10 → 30), unchanged draws, and tuned Metrics 30/20 — which also directly verifies README §5's replacement claim.
- **Title** — fixed: `test(gda-balancing): …` is truthful for the net diff and no longer cuts a feature release.

Required CI is green at `620fc215` and the merge state is clean. The accept / condition / reopen decision on #585 remains with the human owner.

Issue #640, opened 2026-08-12
[gda-balancing] Admit structured Operation execution and declared selection fallback


qinhaihong-red:
## Parent

- #585
- #534
- #501

## Background

PR #639 dogfoods the public Model, Experiment, Formula, Runtime, and artifact paths with
`game.generation@1.0.0` and `game.build@1.0.0`. It correctly composes the structured-value
capabilities delivered by #636 without Roguelike-specific compiler or Runtime dispatch, but review
found two prerequisites that prevent acceptance.

First, bADR-0016 requires every Package Release to ship manifest-bound executable conformance
evidence before it enters the Language Definition Bundle. The provisional Kernel
`runtime-scenario` vector accepts only signed-int64 Operation inputs and state. It cannot represent
the admitted Enum, Record, List, Ref, or typed Quantity values used by the two Package Releases in
PR #639. Public pytest coverage proves host behavior, but it is not package-owned normative
evidence.

Second, PR #639 claims a selection fallback that it does not implement. Its `relaxed-pool` case
selects an ordinary `no-reward` option from a nonempty pool. It does not detect selection
exhaustion, apply a declared fallback, or distinguish fallback from invalid configuration.

Review of the first #640 design found a third prerequisite. The current Operation body is a linear instruction sequence. `if` selects between two
already produced values; it cannot skip instructions. `require` can continue or refuse, but it
cannot skip an effectful path. Therefore `is-empty` and `require` alone cannot avoid a draw and
out-of-range lookup on the empty path while still executing them on the nonempty path. The
implementation needs one single-level guard block before the fallback design is
executable.

Issue #585 requires implementation to stop when the public contracts cannot support the example.
PR #639 therefore remains paused until this issue closes. Preserve its authority-driven structured
Operations and replace only the incomplete evidence, mirrored pool shape, unsupported fallback,
and incorrect effect/refusal classifications.

Standard Schema 2.0 is still under development. Its Kernel remains provisional until Gate 5 and
Gate 6 complete and a maintainer records `Kernel baseline frozen` in PRD #534. This dogfood result
reopens its architecture gate under bADR-0022. The issue may add the demonstrated generic Kernel
primitives to the provisional 2.0 baseline, must produce a new exact Kernel identity, and must
rebuild affected authority and evidence. After that recorded freeze event, another irreducible
primitive requires the next Schema major.

### Operation execution evidence

1. Replace the provisional package-vector kind `runtime-scenario` with `operation-execution`. It is
[...]

Issue #592, opened 2026-07-27
[gda-balancing] Refactor the LDB into sealed orthogonal package artifacts


qinhaihong-red:
## Parent

- #540

## What to build

Refactor the Standard Schema 2.0 Language Definition Bundle into one sealed, content-addressed artifact graph. The root LDB manifest is the sole authority for the exact closed package inventory. Each root-declared package is represented by one independently bounded directory containing a Package Release manifest and exactly one package-owned conformance-vector set. Admission verifies the complete graph atomically before constructing any non-authoritative in-memory index.

This issue was opened from the human architecture review of #540 and PR #587. The first implementation split the monolithic LDB into root-declared Package Release files and corrected the broad `game.rpg` ownership boundary. Dogfooding then exposed a remaining representation defect: package-owned normative vectors were semantically owned by their Package Release but physically inlined in the same JSON as runtime language semantics. For `game.combat`, vectors occupied about 42% of the file; for `core.quantity`, about 62%. The mixed file makes each package another likely growth monolith and obscures the difference between executable language semantics and executable conformance evidence.

The corrected clean-forward Schema 2.0 contract is:

- a root manifest binds the exact Kernel and a canonical ordered set of Package Release descriptors;
- every package descriptor binds one Package Release manifest by artifact kind, package id/version, canonical content identity, and byte size;
- each Package Release manifest binds exactly one `package-conformance-vector-set` child by artifact kind, content identity, and byte size;
- the vector child binds the owning package id/version and contains the package's complete vector id inventory and vector definitions, including an empty set when a package has no vectors;
- each package uses an independent directory, with the package id's dot-separated namespace mapped to a hyphenated physical directory name, for example `packages/game-combat/game.combat@1.0.0.json` and `packages/game-combat/game.combat@1.0.0.conformance-vectors.json`;
- physical locators and directory names are packaging metadata, not semantic identity or membership authority;
- vector changes propagate through vector-set identity, Package Release content identity, and whole-LDB identity, while Package Release semantic identity remains a projection of runtime semantic closure only;
- loaders read only root- and manifest-declared members and never scan directories for semantic discovery;
- admission rejects incomplete, extra, duplicate, substituted, cyclic, inconsistent, or resource-exceeding graphs before exposing any admitted language;
- flat registries and lookup indexes are derived only after successful admission and are never serialized or edited as semantic authority;
- all shipped manifests and vector children are publicly inspectable and byte-identical between source and installed distributions.

[...]

PR #641, opened 2026-08-12
docs(gda-balancing): specify structured operation execution

qinhaihong-red:
## Summary

- replace the provisional `runtime-scenario` design with the closed `operation-execution` package-vector contract
- specify the generic `is-empty`, `require`, and single-level `guard-block` Kernel primitives required by the #585 product-feedback path
- make authored Operation-body order normative and remove the unused `runtime_program.evaluation_order` phase list from the implementation plan
- require LDB maintenance to execute every manifest-bound vector in one complete candidate graph before it publishes any replacement identity
- define declared `no-reward-on-empty` fallback ownership, refusal, RNG, transaction, and downstream build behavior
- reconcile the decisions across the balancing glossary, macro architecture, genre coverage matrix, and bADR-0014/0016/0017/0022/0023

## Scope

This PR records the completed HITL design for #640. It does not implement the Kernel, LDB packages, evaluators, conformance consumers, or PR #639 synchronization.

Standard Schema 2.0 is still under development. Its Kernel remains provisional until Gate 5 and Gate 6 complete and a maintainer records `Kernel baseline frozen` in PRD #534. This design replaces the exact provisional Kernel identity and requires affected authority and evidence to bind the replacement. It does not complete either gate or record the freeze event.

PR #639 remains paused until #640 is implemented and its authority changes are synchronized. Its authority-driven structured Operations remain valid design evidence; they are not temporary host implementations.

## Resolved boundaries

- `operation-execution` is a Kernel-owned conformance vector form, not a public Runtime primitive or Scenario language.
- `guard-block(node, condition, body, outcome)` is top-level, single-level, and terminal on its true path. False skips the body and continues the outer authored sequence.
- `require` compares an already produced Kernel Boolean with an explicit Boolean `expected` value. A mismatch raises the declared typed refusal and uses the existing Runtime-refusal boundary.
- `is-empty`, `require`, and `guard-block` each cost one step. A skipped guard body consumes no body steps, RNG, writes, or effects. Static closure includes the complete body bound.
- The LDB maintenance workflow executes every vector bound by every Package Release manifest through the production and independent consumers. It publishes the complete replacement graph only after all checks pass; failure leaves the existing admitted graph authoritative.
- Public Runtime, the package resolver, and the identity rebuild tool do not discover or execute conformance vectors.

## Validation

- `git diff --check`
- cross-artifact terminology and reference reconciliation
- issue #640 synchronized with the same design and restored to `ready-for-agent`

Refs #640
Refs #585


qinhaihong-red:
## Consolidated review — /artifact-review · /design-domain-modular-architecture · /entropy-review

Three independent reviews ran against head `789792ea` (base `7b075aea`): a document/artifact review, a domain-modular-architecture review, and an entropy (proportionality) review. Findings below are merged and deduplicated; each item names the lens(es) that produced it. All reviews were read-only.

**Verdict: Changes required.** All three lenses accept the design's direction. The blocking items are internal contradictions inside the record itself — each fixable in a few sentences. One design-level question (the shape of `branch`) is worth taking back to HITL before implementation starts.

### What the reviews confirm as sound (keep)

- `operation-execution` replacing `runtime-scenario` is evidence-driven (the int64-only limit is real: `src/gda_balancing/domain/authority/vector_validation.py:211`), and refusing the old kind with no alias/dual parser is the right call.
- Deriving state inventory from `read-write` ports and deleting `state_names` is a net reduction.
- `no_reward_on_empty` orthogonal to `RarityPolicyKind`, the paired `RewardOption` list (kills the length/index invariant), and the removal of `relaxed-pool` are all entropy reductions driven by #639 review evidence.
- `is-empty` via the Kernel-node + LDB structured-operation-law split mirrors the existing `list-at-v1`/`record-field-v1`/`enum-equal-v1` pattern — the best extensibility decision in the PR.
- `require` resolving its reason against `enclosing-operation.refusals` keeps package vocabulary out of the Kernel.
- The bounded RNG projection, the "no empty vector category for matrix symmetry" rule, and the intent to give the pre-freeze rule a single owner (bADR-0022) are all correct.

### Blocking — internal contradictions in the record

1. **The pre-freeze evolution rule has three copies, and two of them forbid what this PR records.** (artifact R1; architecture H2)
   The designated owner clause (`docs/badr/0022:169-173`) covers only meta-format constructs ("fact kind, term type, premise operator, or judgment construct") — not the Semantic-kernel primitives #640 actually adds. Meanwhile `docs/badr/0022:281-287` ("irreducible kernel operation … requires … a Schema-major review") and the glossary `Semantic kernel` entry (`BALANCING-CONTEXT.md:125-127`, "An irreducible addition is a Schema-major kernel change…") both still state the unconditional rule. The sibling `Schema-major Kernel Specification` entry was updated to defer to bADR-0022; this one was edited by the same PR but kept the old rule.
   *Fix:* one Decision bullet in bADR-0022 whose subject covers Kernel primitives/nodes/judgments/meta-format constructs; the other two sites become one-way references.

2. **"Release freeze" is an undefined, ownerless event that now gates the architecture's strongest constraint.** (architecture H1)
   The 6 references this PR introduces are the only occurrences in the repo. Nothing says who declares freeze or which observable marks it, and the glossary has no entry. Any future Kernel change can self-classify as "pre-freeze" and nobody can refute it.
   *Fix:* bind freeze to an existing observable (e.g. the PRD #534 Schema 2.0 baseline acceptance, or the first published Package Release binding the Kernel identity) in the owner clause, and add a `Release freeze` glossary entry.

3. **The genre-coverage closure paragraph conflicts with the resolved design it was updated to record.** (artifact R2; architecture H3, L5)
   `docs/standard-schema-2.0/genre-coverage.md:142-146`:
   - It still authorizes fallbacks "such as no reward, **currency, or a relaxed pool**" — but this PR removes `relaxed-pool` (`docs/badr/0017:33`) and the declared shape `no_reward_on_empty: List<RewardSelection, max=1>` cannot express a currency fallback.
   - "A policy whose admitted constraints cannot select any result and declares no fallback produces the declared exhaustion refusal" reuses the old *unsatisfiable-policy* subject but assigns it the *exhaustion* refusal, while row `ROGUE-REWARD-01` (line 56) retains the separate `guarantee-policy-unsatisfiable-refused-v1` vector. Issue #640 (item 15) and the new `Selection exhaustion` glossary entry say contradictory configuration is *not* exhaustion. The paragraph's own closing clause forbids implementations to choose between meanings implicitly — the doc now leaves two.
   *Fix:* name the one declarable fallback exactly; keep the three-way split explicit (runtime empty pool without fallback → `selection-exhausted` typed refusal; contradictory/unsatisfiable configuration → typed configuration refusal; declared fallback → `no-reward` outcome).

### Design-level — re-check `branch` before implementation (recommend HITL)

4. **The two-arm recursive `branch` is under-argued, and its blast radius is not recorded.** (entropy 1; architecture M3)
   - Both recorded traces (`docs/badr/0017:39-51`) are one-arm: one arm completes, the other arm's `continue` carries nothing but "the rest of the Operation". No use case exercises `when_false` with a body or both arms continuing.
   - The baseline Operation body is not a free instruction stream: `kernel.json` fixes `evaluation_order = [preconditions, named_draws, bindings, effects, facts]`. A recursive arm body containing draw/lookup/write either rebuilds or abandons that phase model, and the top level degenerates to `[is-empty, is-empty, branch]`. The record does not mention this.
   - `result.kind=refusal` (for `require`) is a **new** result-kind value, and `branch` makes the node grammar recursive for the first time — per the `Runtime program contract`, the role-to-structure mapping identity and every Evaluator Capability Manifest must be re-signed. Not recorded; implementers will underestimate scope.
   - Issue #640's premise "`require` can continue or refuse, but it cannot select an effectful path" reasons about a node that does not exist in the baseline — it is design inference, not a measured result.
   *Alternatives (smaller first):* (a) defer the fallback gameplay-outcome path and `branch` entirely — ship `operation-execution` + `is-empty` + `require`, make empty pool a typed `selection-exhausted` refusal, unpause #639 sooner, and let HITL feedback decide whether the outcome path deserves a control primitive (requires the #585 owner to relax one AC clause); or (b) a single-arm `guard-block` `{node, condition, body, outcome}` — a natural generalization of the existing `precondition-*` node that covers both traces, keeps the body flat, and extends to two arms later purely additively. If the two-arm shape stays, record the phase-model reconciliation and the manifest re-signing consequence explicitly.

### Major

5. **`require`'s terminal consequence is omitted.** (architecture M5) bADR-0014's Decision says a Runtime refusal "terminates the run" and publishes the terminal-audit set; the #640 amendment lists only the rollback inventory. That consequence is the entire difference between refusal and gameplay outcome — and the only reason `no_reward_on_empty` exists. Add one sentence, plus the rationale for classifying no-fallback exhaustion as a configuration refusal rather than an outcome (cf. bADR-0014's "gameplay branches must be modeled as outcomes, not refusals").
6. **bADR-0022:53-90 transcribes unimplemented machine encoding field-by-field.** (entropy 2) Exact members, `semantics.*` bindings, `resource_bounds.max_steps` etc. duplicate what `kernel.json` will own — the same amendment's own Decision says prose explains machine records and cannot be authority. Keep the decision-level constraints (selected-arm-only execution, no charge/RNG/writes on the unselected arm, lexical arm locals, refusal enters the bADR-0014 boundary, `is-empty` is generic over one admitted List); leave exact encodings to `kernel.json` and issue #640.
7. **The development conformance harness: unnamed actor, new concepts over existing capability, three full copies.** (entropy 3; architecture M1, M2; artifact M5)
   - The dual-consumer comparison already exists (`tests/test_schema2_experiment_cli.py:5745` runs all five `game.combat@2.1.0` vectors through the production evaluator and an independent reference evaluator). "Isolated candidate graph" is a new concept with zero prior occurrences; the "only agreement permits rebuild" gate sequences ahead of a tool (`tools/rebuild_schema2_ldb.py --check`) that is not wired into CI at all today.
   - "The maintenance workflow" is never defined; neither is the "every applicable vector" predicate; the independence contract is not referenced.
   - The contract is restated in full in three places (`BALANCING-CONTEXT.md:95-98`, `docs/ARCHITECTURE.md:643-649`, `docs/badr/0016:44-52`), and the ARCHITECTURE copy sits after the §4.2 layer table while declaring itself not a layer — §3.3 (injected candidate contexts) already owns the parent rule.
   *Fix:* keep the negative boundary (resolver / public Runtime / rebuild tool never discover or execute vectors — that part is a genuine decoupling win). Make bADR-0016 the single owner; compress the ARCHITECTURE text to a two-sentence macro consequence under §3.3; name the workflow owner and the applicability predicate. The cheaper, currently-missing gate is wiring `rebuild_schema2_ldb.py --check` into CI; defer the candidate-graph mechanism until a bad rebuild actually costs something.
8. **bADR-0017:39-51 hard-codes a node-by-node Operation body in an accepted bADR.** (entropy 4; architecture M4) The body's owner is the `game.generation` Package Release manifest (bADR-0017:137-139); the #596 amendment is the correct precedent (ownership, not bodies). Keep the observable contract (:25-37); drop the trace or mark it non-normative — it already lives in issue #640.
9. **ARCHITECTURE's evidence-state sections were not synchronized.** (architecture M6) §11's Extensibility row, §12.2's #592 entry, and §13 Gate 2 still present old-Kernel-identity evidence as standing, while §7.1 and two bADRs declare it invalidated. §12 is the designated owner for dogfooding-driven architecture changes but has no #585/#640 entry. Add the §12.2 entry, point §7.1 at it, and qualify §11/§13.
10. **Glossary `Selection exhaustion` is broader than the decidable semantics.** (architecture M7) "contains no selectable option" admits a non-empty pool with no selectable option (e.g. all-zero weights), which the specified `is-empty` control flow cannot detect — it would take the non-empty arm and draw. Change to "is empty" and add the non-empty-but-unselectable case to the entry's exclusion list.
11. **The PR body does not mention `branch` at all.** (entropy 7; architecture L8) The second commit specifies it fully and it is the highest-impact primitive of the three; reviewers scoping from the description will underestimate the change. Update the Summary.

### Minor

- `ROGUE-REWARD-01` lists no vector for the new no-fallback runtime path (`selection-exhausted` refusal); add e.g. `roguelike.reward.selection-exhausted-refused-v1` or state which vector closes it. (artifact M1)
- bADR-0014's #640 amendment is inserted between the two 2026-08-04 amendments, breaking the chronological order used everywhere else. (artifact M2)
- Glossary `Declared fallback` drifts from bADR-0017: omits the `reward_score` commit and narrows "preserves policy state and draw count" to draw count only. (artifact M3; architecture L4)
- One existing node, three names: "conditionals" (ARCHITECTURE:699), "value selection" (bADR-0022:181), "value conditionals" (BALANCING-CONTEXT:122). Pick one — "value selection" matches `semantics.operator=select-value`. (all three lenses)
- `is-empty`'s `result.typing` uses `declared-result`, but every constant-Boolean node in `kernel.json` (`equal`, `less-than`, …) uses `{"kind":"fixed","contract":"kernel-boolean"}`; `declared-result` is the operand-dependent-lookup form, and the same paragraph fixes the result as Kernel Boolean anyway. (architecture L3)
- `branch` reuses `if`'s member names `condition`/`when_true`/`when_false` with opposite semantics (select-a-value vs execute-one-side); distinguish the members or record the contrast. (architecture L2)
- `ROGUE-REWARD-01`'s "consume only a produced reward disposition" collides with the new normative `produced`/`not-produced` result discriminator — on the no-reward path the result is `not-produced` and the disposition is read from a state port. Reword (e.g. "committed"). (architecture L5)
- `is-empty` irreducibility is asserted, not argued, in ARCHITECTURE §7.1; the argument is one sentence (no aggregate/length node exists; out-of-range `lookup` refuses rather than returning a Boolean — verified against `kernel.json`). (architecture L7)
- bADR-0014's "`rollback` is reserved for discarding the currently refusing Event's buffers" now collides with the outcome-side `rollback` state policy this PR uses ("rollback `gameplay-alternative`"); add the half-sentence distinction (outcome rollback keeps RNG draws; refusal rollback rolls back RNG continuation). (architecture L9)
- The two "Semantic kernel" inventories (glossary and bADR-0022 Decision) list the new primitives at different granularity with no stated relationship; state it once, reference elsewhere. (architecture L6)
- bADR-0016 amendment: "…and mismatched expected results or state refuse" garden-paths ("state refuse"); reword to "…are admission refusals". (artifact M6)
- The anti-generalization guard ("no label jump, loop, dynamic lookup, evaluator callback") is repeated in five places; bADR-0022's existing Decision clause is the owner, the rest should reference it. (entropy 7)
- Optional: issue #640's "Kernel Unit is a produced `null` value, not `not-produced`" is recorded nowhere; if doc-worthy, bADR-0022's closed-result sentence is the home. (artifact M7)

### Validation notes

- All 22 resolved-design points of issue #640 were traced to documentation homes; apart from the items above, the placement and single-owner discipline are consistently good (the bADR-0022 ownership of the pre-freeze rule, with ARCHITECTURE §7.1 explicitly declining to define a second policy, is the model pattern).
- `git diff --check` clean; both commits Conventional-Commits-conformant; CI green (docs-only); branch correctly based on `origin/main`.
- Terminology pass ran against the balancing glossary, the touched bADRs, and `kernel.json` (node grammar conventions verified: `family`/`members`/`semantics.operator`/`result.typing` forms all match existing records; `gameplay-alternative` is grounded in the kernel's outcome kinds).
- Limits: the PR is docs-only, so every load-bearing claim stays at evidence state `open`; the `branch`-vs-`evaluation_order` conflict is a specification gap identified from the baseline `kernel.json`, not an implementation-verified failure; suggestion 4(a) requires a decision by the #585 owner and is not a reviewer call.

qinhaihong-red:
Addressed at PR head `d6a1757e` in two commits:

- `d933d9c4` — applies the HITL decisions and reconciles the main design artifacts.
- `d6a1757e` — closes the remaining terminology, evidence-state, and ownership gaps found during the final finding-by-finding check.

### Fully adopted

- Made bADR-0022 the single owner of the provisional-Kernel evolution rule. The rule now covers primitives, nodes, judgments, and meta-format constructs.
- Bound the baseline transition to an observable event: Gate 5 and Gate 6 are complete and a maintainer records `Kernel baseline frozen` in PRD #534.
- Replaced the conflicting fallback text with the exact three-way split: empty pool without fallback -> `selection-exhausted`; declared fallback -> `no-reward`; contradictory or unsatisfiable configuration -> its distinct typed refusal.
- Replaced the recursive two-arm `branch` proposal with the HITL-selected, top-level, single-level `guard-block`. Authored Operation-body order is normative; the unused `runtime_program.evaluation_order` phase list is removed by the implementation plan.
- Closed the `require` terminal behavior, refusal rollback/audit/resource consequences, and the distinction between runtime-refusal rollback and gameplay-outcome rollback.
- Added the no-fallback `selection-exhausted` vector, restored bADR chronology, aligned `reward_score` and policy-state wording, standardized `value selection`, fixed `is-empty` to a fixed Kernel Boolean result, changed the disposition wording to `committed`, and recorded produced Unit `null` versus `not-produced`.
- Narrowed selection exhaustion to an empty eligible list and explicitly excluded a nonempty-but-unselectable list.
- Synchronized ARCHITECTURE evidence status: the #592 witness is bound to the superseded Kernel, #585/#640 now has a maintained-product entry, and Gate 2 states that affected evidence must be rebuilt against the replacement identity.
- Reduced bADR-0017's node-by-node body to a short, explicitly non-normative lowering sketch; the Package Release manifest remains the eventual machine owner.
- Updated the PR description to expose `guard-block`, the candidate-graph boundary, the freeze event, and the effect on #639.

### Partially adopted

- Kept the candidate-graph conformance gate because identity rebuild validation cannot execute Package semantic vectors. bADR-0016 is now the detailed owner; the glossary and ARCHITECTURE copies are compressed to term/macro consequences. The applicability predicate is exact: every vector bound by every Package Release manifest in the complete candidate graph. Public Runtime, resolver, and identity rebuild tooling remain excluded.
- Kept exact provisional machine fields in bADR-0022. This PR is the accepted design target for implementation; `kernel.json` becomes machine authority when #640 implements it. Removing the fields now would reopen the encoding decisions instead of reducing duplicated standing authority.
- Used the existing PRD marker rather than adding a new `Release freeze` glossary term. The record consistently names the observable `Kernel baseline frozen` event, so a second domain term would add vocabulary without a second concept.

### Not adopted

- Did not defer the fallback outcome/control primitive or relax #585. The HITL decision retained the required product behavior and selected the smaller one-arm `guard-block` design.
- Did not replace candidate-graph semantic validation with `rebuild_schema2_ldb.py --check`; the checks cover different failure surfaces.
- Did not apply `branch` member-renaming advice because `branch` no longer exists in the design.

Validation:

- `git diff --check`
- terminology and stale-contract scans across the glossary, ARCHITECTURE, bADRs, and genre matrix
- `is-empty` result encoding checked against the current fixed-Boolean node convention in `kernel.json`
- issue #640 and the PR description reconciled with the same HITL decisions

This is a documentation-only design PR, so no implementation tests are claimed. CI is rerunning for `d6a1757e`.

qinhaihong-red:
## Re-review — `789792ea` → `d6a1757e`

Every claimed fix was verified against the artifacts, not the reply. **Verdict: one required change remains** (a missed copy of the same rule the review blocked on), plus two minor residuals. Everything else checks out, and several fixes exceed the minimum (new falsifiable Validation entries in bADR-0016 and bADR-0022).

### Required

**R1 — a third unconditional Schema-major copy in bADR-0022 was missed.**
`docs/badr/0022:298-304` ("Domain operations are machine-defined compositions…") still says an irreducible kernel operation "requires … and a **Schema-major review**." This is the same finding class the previous review blocked on: the glossary entry and the rule-language Decision bullet were fixed, but this bullet still literally forbids what #640 records (three irreducible kernel operations added pre-freeze without a Schema-major review).
*Fix:* replace the last item with a deferral to the baseline rule, e.g. "…Runtime/Numeric-profile laws, and the frozen-baseline rule above (a Schema-major change after the recorded freeze; a provisional-baseline replacement before it)."

### Minor residuals

- `docs/badr/0014:196`: "`rollback` is reserved for discarding the **currently refusing** Event's uncommitted buffers" is still refusal-only wording, while an outcome's declared `rollback` state policy (existing in `kernel.json`; used by this PR as "rollback `gameplay-alternative`") also discards the current Event's uncommitted writes. The amendment's contrast sentence mitigates this; half a sentence in the Decision bullet would close it.
- Wording drift: "single-level **guarded outcome blocks**" (glossary; bADR-0022 Semantic-kernel bullet) vs "single-level **guard blocks**" (bADR-0022, two later sites). Also, neither Semantic-kernel inventory lists the typed requirement (`require`), though the Runtime-node vocabulary bullet does.

### Verified fixed (spot-check record)

- **Freeze event**: bound to an observable (Gates 5+6 complete, maintainer records `Kernel baseline frozen` in PRD #534); all four copies verbatim-consistent and each defers ownership to bADR-0022. Declining a glossary term is reasonable (marker, not a new concept).
- **Three-way exhaustion split**: the genre-coverage closure paragraph now cleanly separates empty-pool-with-fallback → `no-reward` outcome, empty-pool-without → `selection-exhausted` refusal, nonempty-unsatisfiable-guarantee → `guarantee-policy-unsatisfiable`, and contradictory data → configuration refusal; currency/relaxed-pool explicitly excluded. The row gained `roguelike.reward.selection-exhausted-refused-v1`.
- **`branch` → `guard-block`**: members, single-level/top-level constraints, charging, and static closure are consistent across the glossary, ARCHITECTURE, bADR-0014, and bADR-0022; no stale `branch`/`when_true`/`when_false`/`conditional-block` tokens remain. The empty/nonempty control flow truth-checks out with the new `require(condition, expected, reason)` form.
- **`evaluation_order` removal**: independently verified against the implementation — no code consumes the five-phase values (the key appears only in shape validation), so "unused" is evidence-backed; the new Validation entry ("reordering a body must reorder execution") makes the claim falsifiable.
- **Harness ownership**: bADR-0016 is now the single detailed owner with a named workflow, an exact applicability predicate, and failure semantics (the admitted graph stays authoritative); the ARCHITECTURE copy moved to §3.3 and compressed; the glossary copy reduced to a reference; a mutation-probe Validation entry added.
- All remaining items verified as claimed: bADR-0014 amendment chronology, `reward_score`/policy-state alignment, "value selection" standardization, `is-empty` fixed kernel-boolean typing, "committed" disposition wording, produced Unit `null`, exhaustion narrowed to an empty list with the nonempty-but-unselectable exclusion, ARCHITECTURE §11/§12.2/Gate 2 evidence-state sync plus the new #585/#640 entry, the non-normative bADR-0017 sketch, the `is-empty` irreducibility argument, the garden-path fix, the updated PR body, and issue #640 reconciliation (including `expected` and the `evaluation_order` removal).

### Declined items

All three declines are reasoned and acceptable. One accepted risk to carry into implementation: the exact machine fields kept in bADR-0022 will need a follow-up ADR edit if any member changes during implementation — that rationale lives only in the PR reply, not in the record.

### Validation

`git diff --check` clean; CI green for `d6a1757e` (Godot e2e skipping is the standing PR-CI configuration); conventional commit messages; terminology pass rerun on all changed passages including the three new glossary entries (consistent with bADR-0017/0022).

qinhaihong-red:
Addressed at exact PR head `0d86025a` (base tip and merge base `7b075aea`).

| bADR-0022 still required a Schema-major review for every irreducible Kernel operation | Fully Adopt | The unconditional sentence contradicted the accepted pre-freeze replacement rule and the three primitives specified by #640 | The Domain-operation bullet now defers to the frozen-baseline rule: replace and reidentify the provisional baseline before the recorded freeze; use the next Schema major after it | The unconditional `Schema-major review` wording is absent; the remaining rule points to the single baseline policy |
| bADR-0014 described `rollback` as refusal-only | Fully Adopt | The current Kernel outcome contract admits both `commit` and `rollback` state policies for completed outcomes | The Decision now distinguishes Event-refusal rollback from a completed outcome's rollback state policy, including the different RNG behavior | Checked against `kernel.json` and the existing #640 amendment |
| Semantic-kernel inventories used two guard terms and omitted `require` | Fully Adopt | `Guard block` / `guard-block` is the canonical term, and `require` is part of the accepted Runtime-node vocabulary | Both inventories now say `typed requirements` and `single-level guard blocks` | Residual scans find no `guarded outcome block(s)` in the affected artifacts |

The reconciliation pass also found one tracker drift that the re-review had marked as fixed: issue #640 still used the old operand-dependent `declared-result` typing for `is-empty`. The issue now uses `{"kind": "fixed", "contract": "kernel-boolean"}` and the canonical `single-level guard block` wording.

Validation:

- `git diff --check`
- stale-rule and terminology scans across the glossary, architecture records, and genre documents
- live verification of issue #640 after the update
- clean worktree after commit `0d86025a`

This remains a documentation-only design change; no implementation test result is claimed. CI is rerunning for `0d86025a`.

PR #649, opened 2026-08-14
refactor(gda): adopt per-command-group modules over the shared descriptor core (ADR-0040)

qinhaihong-red:
## What

Restructures `src/gda` into the modular architecture recorded in **ADR-0040** (new in this PR), then applies a deep-module pass on the result. `libs/gda-balancing` and `examples/` are untouched.

- **ADR-0040** adopts the per-command-group split that ADR-0023 left as an open follow-on ("revisit in its own ADR"); ADR-0023 gets the dated Outcome note. Both of ADR-0023's attached conditions hold: registration stays consolidated on the descriptor + live Typer tree, and the split consolidates the import surface.
- `gda/commands/` — one module per Command group (16 modules: scene, node, script, resource, shader, theme, project, export, game, diag, logger, perf, input, screen, daemon, meta). Each owns its whole vertical slice: params/result models, renderers, group-only classifiers, descriptors, recipe implementations, and Typer bodies, exposing one `register(root)`. The five recipe modules (`export_run`, `script_run`, `screen_ops`, `daemon_ops`, `skill_ops`) are absorbed by their groups.
- `gda/dispatch.py` — the CLI dispatch tails and runner seams, extracted from `cli.py`; `cli.py` shrinks 3624 → 100 lines (composition root only). `models.py` 4272 → ~490, `render.py` 822 → 75, `errors.py` 701 → 430 — all reduced to the shared single-authority core (error/envelope + schema models, value projection, shared classifier machinery, registries untouched).
- The `cli.py` whole-file suppressions (`F811` ruff per-file-ignore, `reportRedeclaration` pyright comment) are dropped — nothing was masked.
- Deep-module pass (3 commits): the 12 byte-identical live `classify_*` pass-through wrappers collapse into a kind-based default in `HeadlessCommand.execute`; the four daemon lifecycle preconditions are named once (three identical copies plus one already-drifted wording, now explicit); symbols crossing module seams get public names (dispatch tails, ADR-0035 projection constants, the sanctioned `logger→diag` / `shader→script` edge symbols, `make_failure`); the argv model-construction→`BadParameter` rule is stated once in `dispatch.params_or_bad_parameter`.

## Dependency direction

[...]

PR #650, opened 2026-08-15
test(gda): guard the cross-language live contracts at unit level

qinhaihong-red:
## What

Tier 1 of the daemon-module assessment (adversarial review of `gda/daemon/*` via design-domain-modular-architecture + improve-codebase-architecture, adjudicated in-session): a set of **pure unit-level contract guards** plus one dead-code removal. No behavior change beyond deleting an unreachable function.

**Why these guards matter here specifically:** PR CI runs no Godot e2e (schedule/dispatch only), so a one-sided rename of any cross-language contract passed every PR gate and could land on main, surfacing only in the nightly e2e tier. The project already guards this class three times (`LOG_BEGIN` mirror, error-code registry drift, `MAX_WINDOW_FRAMES` mirror); this PR extends the same idiom to the remaining unguarded contracts.

- **Op-name drift** (`tests/test_live_contract_guards.py`): the 13 relayed LIVE operations read off the live Typer tree must equal the harness's `const OP_*` table in `gda_harness.gd`, both directions; `LOG_OPS` (`diag-errors`, `logger-tail`) must be part of the LIVE surface. Extraction has a vacuity floor so an empty regex match fails loudly. Negative-probed three ways (value perturbation, extra harness op, broken extraction) — all fail as intended.
- **`LAUNCH_MARKER` mirror**: Python `session.LAUNCH_MARKER` vs the harness const — a mismatch makes every Engine session silently inert.
- **Wire-shape binding**: the raw dicts `daemon/diag.py` builds (`parse_log_records` / `parse_errors`) are validated through the `logger tail` / `diag errors` descriptors' own output models, round-tripping callstack frames, `gda_log` `fields`, and the closed level/origin enums — drift that previously surfaced only as e2e `contract_violation` now fails a unit test.
- **Timeout ordering invariant**: `OP_TIMEOUT < LIVE_REQUEST_TIMEOUT` and `CONNECT_TIMEOUT + OP_TIMEOUT < LIVE_REQUEST_TIMEOUT` — a stuck harness must surface as the daemon's typed `live_timeout`, not a client socket timeout (worst case today has 5 s of margin, previously unasserted).
- **Dead code**: `diag.parse_log` deleted — the raw `diag log` op it served was superseded by `gda logger tail --raw` (ADR-0022 Outcome, ADR-0026); its tests were its only consumer.

## Layout independence

The new tests import no result model or descriptor by module path — descriptors come from the live Typer tree (`gda_command`, ADR-0023) and models from `descriptor.output_model`. They are therefore correct on both the current layout and the pending #649 restructure, in either merge order.

## Verification

- `ruff check` / `ruff format --check` / `pyright` (0 errors) clean on both commits.
- Fast suite: 1191 baseline → **1197** with the guards → **1194** after deleting `parse_log`'s 3 tests.

Assessment context: the Tier 2 items (Session-log ownership in `DaemonPaths`, the server↔session seam) are deliberately deferred until the next live-stack change wave; Tier 3 items (daemon observability, corrupt-handshake error code, the cross-language op-contract ADR) need their own decisions.

### DIFF
--- a/src/gda/daemon/diag.py
+++ b/src/gda/daemon/diag.py
@@ -21,7 +21,8 @@ error's ordered ``callstack`` (frame ``[0]`` equals the ``at:`` location).
 
 The parsing is **best-effort**: a line that is neither a recognized ``<TYPE>:``
 header nor the ``   at:`` follow-on (a backtrace, an interleaved print line) is
-skipped for ``errors`` and never raises. ``log`` keeps every line verbatim.
+skipped for ``errors`` and never raises. The verbatim whole-log view is served by
+:func:`parse_log_records` with ``raw`` set (ADR-0026).
 """
 
 import json
@@ -135,18 +136,6 @@ def parse_errors(data: str | bytes, limit: int | None = None) -> list[dict]:
     if limit is not None and limit >= 0:
         return errors[-limit:] if limit else []
     return errors
-
-
-def parse_log(data: str | bytes, limit: int | None = None) -> list[str]:
-    """The raw captured output lines from a Session log — minimal parsing (#224).
-
-    The full captured stream (print output AND error lines) verbatim, one entry
-    per line. ``limit`` tails the most recent ``N`` lines. Empty input -> ``[]``.
-    """
-    lines = _lines(data)
-    if limit is not None and limit >= 0:
-        return lines[-limit:] if limit else []
-    return lines
 
 
 # --- The structured `LogRecord` channel (#281, ADR-0026) -----------------------
--- a/tests/test_diag_log_parser.py
+++ b/tests/test_diag_log_parser.py
@@ -5,11 +5,11 @@ the diag parse logic is fast-unit-testable. Godot writes errors as a two-line
 pair (``<TYPE>: <message>`` then ``   at: <function> (<file>:<line>)``) and
 print output as plain lines, both into the one ``--log-file`` the daemon owns
 (verified against the engine's ``core/io/logger.cpp``). The parser turns errors
-into ``{level, message, function?, file?, line?}`` and ``log`` into raw lines,
-best-effort: it never fails on a malformed/continuation line.
+into ``{level, message, function?, file?, line?}``, best-effort: it never fails
+on a malformed/continuation line.
 """
 
-from gda.daemon.diag import parse_errors, parse_log
+from gda.daemon.diag import parse_errors
 
 # A realistic Godot --log-file capture: print output interleaved with the engine's
 # two-line error pairs across all four ErrorType strings, a multi-line backtrace,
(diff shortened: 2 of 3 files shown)


qinhaihong-red:
## Review — Standards / Spec

Reviewed exact head `afde56ed83a0380cc4265ddc9f0ee06788fc9385` against base `807ffc82490ae52378850ce27443df6aad4a8c49`.

### Standards

No findings.

### Spec

**[P2] Exercise every valid `level` / `origin` enum member in the descriptor-model guard**

The new [`SESSION_LOG` fixture and wire-record guard](https://github.com/aigengame/godot-agent/blob/afde56ed83a0380cc4265ddc9f0ee06788fc9385/tests/test_live_contract_guards.py#L152-L207) only produce levels `info`, `warning`, and `error`, and origins `null`, `engine`, `script`, and `gda_log`. They never send the valid `debug` level or `shader` origin through the descriptor's `output_model`, even though [ADR-0026 closes those enums over `debug | info | warning | error` and `engine | script | shader | gda_log`](https://github.com/aigengame/godot-agent/blob/afde56ed83a0380cc4265ddc9f0ee06788fc9385/docs/adr/0026-structured-runtime-log-protocol.md#L39-L45).

I reproduced the gap by substituting a consumer model that rejects `debug` and `shader`; the new guard still passes. That leaves producer/consumer drift for those two valid wire values outside the PR gate. Please add a `debug` `gda_log` record and a `SHADER ERROR` record to the representative sample, then assert that both survive the descriptor-model round trip.

No other standards or spec issues found.

### Verification

- `git diff --check`
- Ruff check and format check
- Pyright: 0 errors
- Focused tests: 22 passed
- Fast suite: 1194 passed, 390 deselected
- Real Godot e2e was not rerun; the corresponding CI job is skipped on this PR

qinhaihong-red:
Review handled — the one finding is **fully adopted**, fixed in `b516ce46`.

|---|---|---|---|---|
| [P2] The wire-shape guard never sends the valid `debug` level or `shader` origin through the descriptor's output model, so producer/consumer drift on those two ADR-0026 enum members stays outside the PR gate | Fully adopt | Reviewer's repro confirmed (a consumer model rejecting `debug`/`shader` passed the old guard); the fixture indeed produced only info/warning/error and engine/script/gda_log/None | `SESSION_LOG` gains a `debug` `gda_log` record and a `SHADER ERROR` block (with `at:` line); both round trips are asserted (`debug` record message, `shader` origin at `error` level, `shader_error` diag entry with file/line); the enum assertions tighten from subset (`<=`) to **set equality** on both `origin` and `level`, so a fixture regression that drops a member also fails | Re-ran the reviewer's probe inverted: a narrow model without `debug`/`shader` now fails the guard's validate step rejecting exactly `['debug', 'shader']`. Gates at `b516ce46`: ruff / format / pyright clean, focused file 6 passed, fast suite 1194 passed |

Head is `b516ce46`; e2e tier remains skipped on PR CI by design (unchanged).

PR #676, opened 2026-08-15
docs(gda): document the Control-root zero-size layout in scene create help and the skill

qinhaihong-red:
## Summary

Dogfooding source GDA-DF-006: `gda scene create ui/main.tscn --root-type Control` creates a valid root with zero anchors/offsets and no viewport-filling layout. Container minimum sizes can make descendants visible and mask the problem until `gda game rect` shows the root, and its child layers, at `[0, 0]` size.

This is a documentation-only change: no input model, schema, or CLI-flag change. It states the behavior, the fix, and the check in two surfaces:

- **`gda scene create --help`** (`src/gda/commands/scene.py`, `create()` docstring): a paragraph after the existing one-line summary, stating the zero-anchor/zero-offset behavior, the fix (`anchor_right`/`anchor_bottom` to `1` via `gda node set`), and the `gda game rect` check.
- **The bundled skill** (`src/gda/skill/SKILL.md`, "Scene authoring" section): a paragraph placed right after the existing Control-layout/`offset_*` paragraph (the natural neighbor), stating the same behavior, fix, and check in the skill's prose voice.

README.md is untouched — this is a per-command layout pitfall, not onboarding material (repo convention: no per-command pitfalls in README).

**Review round 2** (external adversarial review, see the PR comments below) found the round-1 wording overstated the claim: it said any Control root renders as a literal zero-size rect, but that is only true for a root class with no intrinsic minimum size. A class with an intrinsic minimum (`Button`, `Label`, …) renders at that minimum instead — still not the viewport, but not `[0, 0]` either. Both surfaces are now reworded to state this precisely, and a new test (`tests/test_skill_surface_sync.py::test_scene_create_control_root_note_agrees_with_skill`) guards the two hand-maintained prose surfaces against drifting apart again — they had already diverged once mid-review (a scope-only fix landed in SKILL.md, `ca00f092`, without the matching `scene.py` edit).

**Review round 3** (recheck of round 2's fix) found the new test itself was too weak: it searched its tokens across the whole bundled skill, so `game rect` / `node set` could be satisfied by an unrelated occurrence, and it did not guard the semantic facts round 2 actually corrected (zero anchors/offsets, the intrinsic-minimum conditional). Scoped the assertion to just the Scene-authoring paragraph (anchor-extracted, failing loudly if the anchor disappears) and widened the checked tokens to cover the corrected facts. No prose or `--help`/`gda schema` change — test-only.

[...]

Issue #670, opened 2026-08-15
CLI surface consistency: aliases, did-you-mean, and unified script path forms


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-024** (2026-08-10) — `daemon install` does not exist; `daemon start` implicitly installs the harness, and the unknown command gives no structured hint.
- **GDA-DF-025** (2026-08-10) — `scene inspect` is unknown; the operation is `scene get`; unknown subcommands render human help instead of the structured convention.
- **GDA-DF-033 "Script command naming is not self-consistent"** (2026-08-11; numbering collides in the source document — identified by title and date) — `script check` is absent; the operation is `script validate`.
- **GDA-DF-032 "Related CLI discovery forms are inconsistent"** (2026-08-12, same collision note) — `gda help X`, `gda --schema`, `gda version`, `gda game get-property`, `script run --script <path>` are all rejected without pointing at their working siblings; `gda info` rejects `--project` that most engine-spawning commands accept.
- **GDA-DF-041** (2026-08-13, no-redirect half) — `gda analyze` and `gda doctor` are unknown with no closest-command hint. (The `analyze` capability half is #663's scope; the hint mechanism built here serves its spelling.)

The validate/run **path-form unification** (GDA-DF-019) was split out to **#675** during triage, with its ADR-0031 amendment direction decided there.

## What to build

Decisions recorded at triage (2026-08-15):

- **Documented-surface drift → implement.** ADR-0005 and CONTEXT.md name `gda version`, `gda help`, and `daemon install` (ADR-0017 extension) as part of the surface; the dogfooding record shows agents actually type them. Implement all three: `version` and `help` as the ADR-0005 `Meta command`s, `daemon install` as an explicit idempotent install/sync operation (resolving GDA-DF-024's discoverability gap at the source).
- **`gda info` accepts `--project`** — validated (structured error on an invalid path), for uniform orchestration argv.
[...]

Issue #669, opened 2026-08-15
schema: argv-binding metadata and a discriminated input-event union


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-003** (2026-08-08, reconfirmed 2026-08-09) — emitted schemas state required fields but not their CLI spelling: `screen capture` needs `--output` while `input mouse-click` needs positional `x y`; `input action` requires positional `ACTION` and rejects `--action`. The JSON input contracts are accurate yet insufficient to construct argv without `--help` round-trips or guessing.
- **GDA-DF-037** (2026-08-12, reconfirmed) and **GDA-DF-032 "Live input-sequence action schema is harder to infer than single-action input"** (2026-08-13; the document's GDA-DF numbering collides across sessions — identified by title and date) — the shared sequence event shape encodes no per-kind validity: the per-kind fields appear only in the schema's description prose, and help shows only a mouse example. `pressed` is valid only on `mouse_button` events (not on `mouse_click` or any other kind); action events need separate press and `release` forms, so schema-driven automation adds a failed invocation to each new live-control workflow.

## What to build

- **Derive** argv-binding metadata (positional vs option, option name) for every field **from the live Typer/Click parameters** at the two schema sites — the per-command `--schema` (which already iterates the command's params) and the whole-surface manifest (which already walks the live Click tree) — per ADR-0012 and ADR-0023 §2. Do **not** add an argv field to the `HeadlessCommand` descriptor: a hand-maintained spelling table would be exactly the parallel registry ADR-0023 rejects; the Typer signature is already the single source of the binding.
- Publish the `input sequence` event type as a discriminated union with per-kind required and forbidden fields; include a copyable action example (press event, process frames, release event) beside the mouse example in help.
- On rejection, name the accepted field for the event kind alongside the rejected one (e.g. `release` for an action event). Note: the `pressed`-on-action mistake is already rejected loudly today; the gap is that the message does not name the accepted alternative and the union is not machine-checkable.

Boundary: #670 owns rejection-message hints for near-miss commands/options; this issue owns the emitted contract. Check whether the manifest addition changes the gda-mcp wire schema and record the answer.

## Acceptance criteria

- [ ] An agent can construct a valid argv for any command from the emitted schema alone (spelling and position included).
[...]

Issue #659, opened 2026-08-15
Structured version and installation-provenance self-description


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-018** (2026-08-09) — `gda --version --json` is rejected by the root parser; evidence tooling must parse human text or reuse a recorded version, weakening exact-environment provenance. The root-help half of this entry is tracked in #671; **#671 owns the shared root `--json` option in the CLI entry point — this issue owns the version payload.**
- **GDA-DF-043** (2026-08-14) — the PATH-resolved gda was an editable install; mid-run, its source checkout changed revision and 27 strict-suite invocations failed with a Python `NameError` before Godot. `gda --version` reported only `gda 0.8.0`, with no editable/source-root/revision disclosure, so the runner could not preflight or pin the implementation.

## What to build

- A structured version/provenance surface — a `Meta command` alongside `gda info` (its naming interacts with #670's `doctor` discussion) or `gda --version --json` — returning:
  - gda version;
  - executable path and interpreter path;
  - install kind (wheel or editable); for editable installs: source root, Git revision, and dirty state when resolvable (e.g. from `direct_url.json`); wheels report that no source checkout backs them;
  - the detected Godot version **only when it can be obtained without spawning the engine**, otherwise omitted with a stated reason;
  - a schema/protocol version only if one is introduced — gda has none today; do not invent one for this issue.
- The surface must not perform a `Headless launch`: the motivating environment is exactly the restricted profile where engine spawns crash (GDA-DF-030/042 record `gda info --json` among the crashing commands).
- This enables production runners to pin/verify the implementation before long strict runs.

## Acceptance criteria

- [ ] The structured invocation returns, at minimum: gda version, executable path, interpreter path, install kind; plus source root/revision/dirty state for editable installs.
- [ ] The result is produced without launching Godot; the engine version appears only when obtainable without a spawn.
- [ ] Wheel and editable installs are covered by tests.
- [ ] Documented in the skill/README as the provenance preflight for long runs.

## Blocked by

None - can start immediately. Related: #671 (root `--json` option owner), #670 (Meta-command naming).

qinhaihong-red:
> *This was generated by AI during triage.*

[...]

PR #677, opened 2026-08-15
feat(gda): accept --json on the root and schema discovery surfaces

### LINKED ISSUE #671: Root discovery surfaces accept --json (help, schema)
## Dogfooding source

- **GDA-DF-022** (2026-08-10) — `gda schema --json` is rejected with `No such option: --json`, even though the skill teaches "always pass `--json`" and describes `schema` as the whole-surface JSON manifest.
- **GDA-DF-018** (2026-08-09, root-help part) — `gda --json --help` is rejected, so a machine client must special-case exactly the discovery surface before it can discover subcommands. (The `--version --json` half of this entry is tracked in #659.)

**Two distinct parser sites** (important — a root-only fix leaves `schema` broken): `gda --json --help` is rejected by the **root** parser (the root callback declares only `--version`); `gda schema --json` is rejected by the **`schema` subcommand's** parser (it declares only `--schema`). Both must be fixed.

## What to build

- Accept `--json` on `schema` and at the root, at minimum as an idempotent no-op where JSON is already the output. Making JSON the default is an optional addition, **not** an alternative — a client that follows the documented "always pass `--json`" rule must stop getting exit 2.
- Root help under `--json` is either structured or explicitly documented as an exception in the skill.
- Ownership: **this issue owns the root-level `--json` option**; #659 consumes it for the version payload. Boundary with siblings: manifest *content* is #669; near-miss *spellings* are #670; this issue is flag acceptance only.

## Acceptance criteria

- [ ] `gda schema --json` exits 0 and returns the manifest under the same structured-output convention as operational commands.
- [ ] `gda --json --help` either returns structured output or the skill lists root help as an explicit exception.
- [ ] Every surface that does not accept `--json` is listed as an explicit exception in the skill (a rule that stays true whichever default is chosen).
- [ ] A contract test pins the accepted invocations at both parser sites.

## Blocked by

None - can start immediately. Related: #659 (consumes the root option), #669, #670 (boundaries above).

qinhaihong-red:
Closes #671

## Problem

The bundled Skill teaches ONE rule — "always pass `--json`" — but discovery rejected it at
**two different parser sites**, so a machine client had to special-case exactly the surface it
uses to discover the others (a root-only fix would have left `schema` broken):

- **root parser** (GDA-DF-018, root-help part): the root callback declared only `--version`, so
  `gda --json --help` — and `gda --json <group> <command>` — exited `2` with `No such option: --json`;
- **`schema` subcommand parser** (GDA-DF-022): it declared only `--schema`, so `gda schema --json`
  exited `2` — although `schema` *is* the whole-surface JSON manifest.

## Change

- `src/gda/cli.py` — the root callback declares `--json`.
- `src/gda/headless.py` — **the root flag is inherited, not inert.** This layer owns both halves of
  the contract: `ROOT_JSON_META_KEY` (click's context-meta dict, shared by every context in the
  tree) plus a `set_root_json(ctx, value)` setter the root callback CALLS, and a `json_option()`
  callback that reads it. Knowledge runs downward (cli → headless), with no private parameter name
  crossing the seam. A command's own `--json` wins when given; otherwise the root's applies. Click finishes the root group's params before it
  parses a subcommand, so the value is already there — no argv re-parsing. Because it lives on the
  one shared option, every call site inherits it with no per-command wiring, including the
  `--params-json` dispatch path (it reads `ctx.params`, after the callback has run).
  Accepting a flag that silently returned human text would have been worse than the loud usage
  error it replaces, so `gda --json skill` and `gda skill --json` now produce identical bytes.
- `src/gda/commands/meta.py` — `gda schema` declares `--json` and keeps its original one-line
  `typer.echo(...)` body. The option is declared but not read — the same idiom as `schema` /
  `params_json` on every other command — because this command's only output *is* the JSON
  manifest, so the flag cannot change it. `gda schema` stays byte-identical by construction.
- `src/gda/skill/SKILL.md` — the `--json placement` paragraph in `## Discovery` (below). This is
  the shipped agent guidance, so the contract lands HERE, in this head — not deferred to a rebase.
- `README.md` + `docs/README.zh-CN.md` / `README.ja.md` / `README.es.md` — the READMEs called
  `gda schema` the `--json` exception, which this branch makes false. Corrected in all four
  languages, plus the root spelling on the global-flags `--json` row; translations re-stamped with
  `uv run python scripts/update_readme_i18n.py` (gated by `tests/test_readme_i18n_sync.py`).
- `tests/test_json_flag_acceptance.py` (12 tests) — pins both parser sites, the
  root-vs-command equivalence (meta command, domain command via the fake runner, `--params-json`
  path), and the honest order-dependence the root flag has toward an eager root callback (below).
- `tests/support.py`, `tests/test_cli_version.py`, `tests/test_export_run_commands.py` — Typer
  renders help through Rich, which colorizes under GitHub Actions but not under a local
  `uv run pytest`, so a help-text assertion can pass locally and fail in CI on escape sequences
  alone (it did, on the first push). Two modules already carried a private copy of the same ANSI
  stripper; it is promoted to `tests/support.plain_text` (three consumers) and the help-text
  assertions in those three modules go through it. Other test modules still assert on help text
  without stripping — they match short substrings that Rich does not break up, so they are left
  alone. Reproduce the CI rendering locally with `GITHUB_ACTIONS=true uv run pytest`.

Scope kept to flag **acceptance/placement**: manifest *content* is #669, near-miss *spellings* are
#670, the `--version` JSON payload is #659.

### What #659 inherits (order-dependence, stated honestly)

The root `--json` is eager, so it lands in the root context before the *other* eager root options
run — **but only for the `--json`-first spelling**. Click processes eager params in argv order, so
`gda --version --json` still runs the version callback first, with the flag unbound. A root JSON
payload that must serve both orders therefore cannot be rendered from inside the eager `--version`
callback; that is #659's design call. `test_root_version_stays_text_in_both_json_orders` pins today's
text output for both orders, flagged in-comment as the expectation #659 flips, and the constraint
itself is stated in `cli.py`'s comment on the option. (An earlier probe test asserting the
processing order was removed in review: it introspected Typer internals to protect a choice #659
has not made yet.) A SUBCOMMAND is unaffected either way.

## Public-surface diff (every intended change, enumerated)

Baselines were captured before the first edit; the base tree is re-run from `git archive d0de6bd2`
for the comparisons below.

1. `gda schema` (stdout manifest) — **byte-identical**, `sha256 c0710da1…8befd7` on base, on this
   branch, and with `--json`.
2. `gda --help` — one added option row (3 wrapped lines):
   ```
   │ --json             Emit the invoked command's result as JSON — the same as   │
   │                    passing --json after the command; the root itself prints  │
   │                    no JSON payload (`--help` stays text).                    │
   ```
3. **bare `gda`** (the no-args help path) — the same 3 added lines, since it prints the root help.
   Its exit code is unchanged (`2` on base and on this branch — Click's `no_args_is_help`).
4. `gda schema --help` — one added option row plus two docstring lines:
   ```
    `--json` is accepted and idempotent: the manifest is already the JSON
    result, so the flag changes nothing and an agent can pass it uniformly.

   │ --json            Emit the result as a single JSON object.                   │
   ```
5. **Every other `--help` surface is unchanged**: an 86-surface sweep (root, every group, every
   command in the manifest, `info` / `skill` / `schema`) diffed base-vs-branch reports exactly two
   changed surfaces — `<root>` and `schema`. The per-command `--json` help text is untouched.
6. `gda skill` (the emitted `SKILL.md`) — diff vs base is **exactly** the new Discovery paragraph:
   one blank line plus the 7-line `**--json placement.**` block quoted below, nothing else.
7. `README.md` and the three translations — the `gda schema` "exception" sentence corrected and the
   global-flags `--json` row extended; translation markers re-stamped.
8. `gda schema --schema` and the aggregate manifest — unchanged: contracts are model-derived, and
   `schema` gained no `--params-json`, so it stays non-dispatchable (ADR-0012 Plan A).

### Behavior diff

| Invocation | Base | This branch |
| ---------- | ---- | ----------- |
| `gda schema --json` | exit `2` (`No such option`) | exit `0`, manifest (byte-identical to `gda schema`) |
| `gda --json --help` | exit `2` | exit `0`, byte-identical to `gda --help` |
| `gda --json <command>` | exit `2` | exit `0`, same output as `gda <command> --json` |
| `gda <command> --json` | unchanged | unchanged |

## Validation (all run in the worktree, on the final tree)

| Gate | Result |
| ---- | ------ |
| `uv run pytest -m "not e2e"` | `1207 passed, 390 deselected` |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` (CI colorization) | `1207 passed, 390 deselected` |
| `uv run ruff check` | `All checks passed!` |
| `uv run ruff format --check` | `427 files already formatted` |
| `uv run pyright` | `0 errors, 0 warnings, 0 informations` |
| 86-surface `--help` byte-diff, base vs branch | only `<root>` and `schema` differ |
| `gda schema` sha256, base vs branch vs `--json` | identical (`c0710da1…8befd7`) |

Real invocations (no Godot needed):

| Invocation | Exit | First line |
| ---------- | ---- | ---------- |
| `gda schema --json` | `0` | `{"commands":[{"name":"daemon start",…` |
| `gda schema` | `0` | same bytes as above |
| `gda --json schema` | `0` | same bytes as above |
| `gda --json --help` | `0` | ` Usage: gda [OPTIONS] COMMAND [ARGS]...` (identical to `gda --help`) |
| `gda --json skill` | `0` | `{"name":"gda","version":"0.9.0","content":"---` |
| `gda skill --json` | `0` | identical bytes to `gda --json skill` |
| `gda skill` | `0` | `---` (text default unchanged) |
| `gda --json skill --json` | `0` | identical bytes (both levels is not a conflict) |
| `gda --json skill --params-json {}` | `0` | `{"name":"gda",…` (the ADR-0015 path honors it) |
| `gda schema --json --schema` | `0` | `{"input":{…` (`--schema` still wins) |
| `gda --json --version` / `gda --version --json` | `0` | `gda 0.9.0` (text; the JSON payload is #659) |
| `gda --json` (no command) | `2` | `Missing command.` — unchanged usage error |
| `gda scene --json` | `2` | group parsers take only `--help` (third parser site, out of scope) |

Godot-spawning rows (`gda --json info` / `gda info --json`) were **not** run: the shared e2e lock
`scratchpad/gda-e2e.lock` was held by another worktree, and this slice needs no engine. The same
sentinel dispatch path is covered engine-free by `test_root_json_reaches_a_domain_command`
(`scene get` against the fake runner, both spellings byte-equal), and the real-process path by
`gda --json skill` above.

## Acceptance criteria

- [x] `gda schema --json` exits `0` and returns the manifest under the same structured-output
      convention as operational commands.
- [x] `gda --json --help` returns exit `0` (text help; documented as an exception in the Skill
      lines below).
- [x] "Every surface that does not accept `--json` is listed as an explicit exception in the Skill"
      — **landed in this head** (`src/gda/skill/SKILL.md`, `## Discovery`), no longer deferred to a
      rebase: PR-body text is not shipped agent guidance. Guarded by token-level assertions in
      `tests/test_json_flag_acceptance.py`.
- [x] A contract test pins the accepted invocations at both parser sites
      (`tests/test_json_flag_acceptance.py`).

## The Skill block that landed

In `src/gda/skill/SKILL.md`, `## Discovery`, directly after the existing bullet list:

```markdown

**`--json` placement.** `gda --json <group> <command>` and `gda <group> <command> --json` mean the
same thing — a root `--json` applies to the command it invokes — so either spelling works, as does
both at once. `gda schema --json` is accepted too, and idempotent: the manifest is already JSON.
Two limits: help output is always TEXT (`gda --json --help` returns the same help, never JSON), and
two spellings are still usage errors (exit `2`) — `gda <group> --json` (a group's parser takes only
`--help`; pass the flag to the command) and a bare `gda --json` with no command
(`Missing command.`).
```

Every claim in that block was verified against this branch: `gda --json skill` == `gda skill --json`
== `gda --json skill --json` (byte-equal JSON), `gda skill` still text, `gda --json --help` == `gda
--help` (byte-equal), `gda schema --json` == `gda schema` (sha256-equal), `gda scene --json` → `2`,
`gda --json` → `2`. The `<group> <command>` equivalence is additionally pinned engine-free by the
`scene get` fake-runner test.

## Review rounds

**Round 2 (internal):** root `--json` made genuinely inherited rather than inert; the
`emit_result` / `render_surface_manifest` / registry-exemption mechanism reverted so `gda schema`
keeps its plain `typer.echo` body under a declared-but-unread option; the eagerness claim in
`cli.py` corrected to the real argv-ordered guarantee; the skill block rewritten and re-verified;
bare `gda` added to the surface enumeration.

**Round 3 (external adversarial review, head `4e179fd5`):** all five claims adopted —

| Claim | Decision | Change |
| ----- | -------- | ------ |
| 1 + 5 — the placement contract must ship in this head, not in PR text | fully adopted | `SKILL.md` `## Discovery` paragraph landed + token-level guard test; AC re-marked satisfied |
| 2 — `_inherit_root_json` hard-coded `cli.py`'s private param name | fully adopted | `ROOT_JSON_META_KEY` + `set_root_json()` owned by `gda.headless`; the root callback calls it |
| 3 — README + translations still call `gda schema` the `--json` exception | fully adopted | corrected in all four languages, restamped |
| 4 — eagerness probe couples to Typer internals for a future #659 choice | fully adopted | test deleted; the constraint stays in the `cli.py` comment and the both-orders behavioral test |

**Coordination note:** sibling PR #680 also edits `README.md` + the three translations (a different
row) and restamps. It merges first, so this branch needs a rebase and a re-run of
`uv run python scripts/update_readme_i18n.py` afterwards; the orchestrator sequences that. Sibling
#676 edits `SKILL.md` in a different region (~line 172); this change touches only `## Discovery`.
The `gda <group> --json` third parser site is tracked separately as #683.

No sibling-owned file was touched beyond the coordinated README/SKILL regions above
(`error_codes.py`, ADR-0002 untouched).


qinhaihong-red:
Adversarial review pinned to `4e179fd5d94b0ff9a0b66ea0c124cd95bde0d17f` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

**[P1] Land the public `--json` placement contract in this exact head — `src/gda/skill/SKILL.md:3,33,57-63`.**

The implementation adds equivalent root/leaf placement and deliberately leaves text-help, group-parser, and bare-root exceptions, but the bundled Skill still says only “Always pass `--json`” and documents none of those exceptions. Issue #671 explicitly requires every non-accepting surface to be listed. PR-body text deferred to a future rebase is not shipped agent guidance. Rebase the sibling Skill change into this PR, add the stated placement/exception block, and guard it with a focused contract test. Claim evidence state: `open`; the exact head does not contain the required contract.

**[P2] Do not make lower `headless` code discover a root-parser implementation detail — `src/gda/headless.py:81-100`.**

`headless._inherit_root_json()` walks upward with `ctx.find_root()` and hard-codes the root callback's private Python parameter name, `"json_output"`, owned by `cli.py`. This reverses the documented `cli -> commands -> dispatch -> headless` knowledge direction and creates two editable authorities for the same option. Renaming that Python parameter while retaining public `--json` makes root placement silently return human text. Propagate the preference through an explicit lower-owned context contract (or a dedicated CLI-options boundary), and test public root/leaf equivalence without depending on the private parameter name.

## `review`

### Standards

**[P2] Synchronize public CLI documentation — `README.md:360-361` and the three translated mirrors.**

`src/gda/commands/meta.py:353` now accepts `gda schema --json`, but the English, Chinese, Japanese, and Spanish command-reference passages still call `gda schema` the exception to `--json`. Update all four synchronized surfaces to say that the flag is accepted and idempotent because the normal output is already JSON.

**[P3] Possible Speculative Generality — `tests/test_json_flag_acceptance.py:178-207`.**

The callback-introspection/monkeypatch test protects a future #659 design choice, not current observable behavior, which is already covered by the black-box two-order test. It couples the suite to Typer internals and can fail on harmless refactors. Remove this test and `_UNBOUND`, retaining the public behavior assertion.

### Spec

**[P1] Acceptance criteria 2 and 3 are still unmet — `src/gda/skill/SKILL.md:57-63`.**

Root help remains textual and direct probing confirms group-local placement such as `gda scene --json --help` is rejected, yet the Skill lists neither as an exception. Update the bundled Skill in this PR to document text-only root help and enumerate every rejecting placement/surface, including group help; align discovery examples such as `gda schema --json`.

Validation: exact base/head and merge-base pinned; `git diff --check`, Ruff, format, Pyright, 36 focused tests, and all 1,207 non-e2e tests passed. The 390 Godot-dependent/e2e tests were not run.

Summary: DDMA — 2 findings (worst P1); Standards — 2 (worst P2); Spec — 1 (worst P1).

qinhaihong-red:
## Review reply — all five claims adopted (head `48b97ff5`)

Thanks — the review was reviewed against head `4e179fd5` and every claim was adopted. Summary
first, evidence per row.

| Claim | Decision | Evidence it was real | Change | Validation |
| ----- | -------- | -------------------- | ------ | ---------- |
| **1 + 5** — the `--json` placement contract lived only in the PR body, deferred to a post-merge rebase; ACs 2/3 unmet | **Fully adopted** | PR text is not shipped guidance: an agent reads the bundled `SKILL.md`, which said nothing about placement | `src/gda/skill/SKILL.md` `## Discovery` gained the `**--json placement.**` paragraph (quoted below), landed in THIS head. Region-disjoint from #676 (~line 172) | `gda skill` output diff vs base is exactly that paragraph (1 blank + 7 lines, nothing else); every claim in it re-run on this head (table below) |
| **2** — `_inherit_root_json` reached up into `cli.py` via `ctx.find_root().params.get("json_output")`, hard-coding a private parameter name | **Fully adopted**, via the lower-owned contract | Renaming the root parameter would have broken inheritance silently; the option layer knew a CLI-module name | `gda.headless` now owns both halves: `ROOT_JSON_META_KEY` (click's context-meta dict, nested from the parent context) + `set_root_json(ctx, value)`. The root callback CALLS the setter; the option callback reads the key. Direction is cli → headless | Verified click's vendored `Context.meta` is inherited from the parent (`self._meta = getattr(parent, "meta", {})`). Behavior unchanged: the black-box equivalence tests (`--json skill` == `skill --json`, domain command, `--params-json`) still pass and remain the loud guard |
| **3** — `README.md` ~line 360 and the three translations still called `gda schema` the `--json` exception | **Fully adopted** | The sentence became false on this branch | Corrected in `README.md`, `docs/README.zh-CN.md`, `README.ja.md`, `README.es.md` in-language ("accepted, and changes nothing because the manifest is already JSON"), plus the root spelling on the global-flags `--json` row; re-stamped with `uv run python scripts/update_readme_i18n.py` | `tests/test_readme_i18n_sync.py` passes in the fast suite. **Coordination:** sibling #680 also edits README + translations + restamp (different row); it merges first and this branch then needs a rebase + re-stamp — noted in the PR body |
| **4** — the callback-introspection/monkeypatch eagerness test coupled to Typer internals and protected a future #659 choice | **Fully adopted** | It asserted parameter-processing order, not shipped behavior, and reached into `app.registered_callback` + the declared option object | `test_root_json_visibility_to_an_eager_root_option_is_argv_ordered` and the `_UNBOUND` sentinel deleted | `test_root_version_stays_text_in_both_json_orders` still pins current behavior for both argv orders; the argv-order constraint #659 inherits stays as a comment on the option in `cli.py` |

### The paragraph that landed (`SKILL.md`, `## Discovery`)

> **`--json` placement.** `gda --json <group> <command>` and `gda <group> <command> --json` mean the
> same thing — a root `--json` applies to the command it invokes — so either spelling works, as does
> both at once. `gda schema --json` is accepted too, and idempotent: the manifest is already JSON.
> Two limits: help output is always TEXT (`gda --json --help` returns the same help, never JSON), and
> two spellings are still usage errors (exit `2`) — `gda <group> --json` (a group's parser takes only
> `--help`; pass the flag to the command) and a bare `gda --json` with no command
> (`Missing command.`).

Guarded by token-level assertions in `tests/test_json_flag_acceptance.py` (both equivalent
spellings, both exit-2 spellings, the exit code) — deliberately not in `test_skill_surface_sync.py`,
which a sibling PR edits.

### Re-verification on head `48b97ff5`

Every claim in the paragraph, run:

| Invocation | Exit | Result |
| ---------- | ---- | ------ |
| `gda --json skill` / `gda skill --json` / `gda --json skill --json` | `0` | identical JSON bytes |
| `gda skill` | `0` | text (default unchanged) |
| `gda --json skill --params-json {}` | `0` | JSON (ADR-0015 path honors the root flag) |
| `gda schema` / `gda schema --json` / `gda --json schema` | `0` | identical bytes |
| `gda --json --help` | `0` | byte-identical to `gda --help` |
| `gda <group> --json` (`scene`), `gda <group> --json --help` | `2` | group parser takes only `--help` (third parser site, tracked as #683) |
| `gda --json` (no command) | `2` | `Missing command.` |
| `gda --json --version` / `gda --version --json` | `0` | text `gda 0.9.0` (the JSON payload is #659) |

Gates, all actually run on this head:

| Gate | Result |
| ---- | ------ |
| `uv run pytest -m "not e2e"` | 1207 passed, 390 deselected |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1207 passed, 390 deselected |
| `uv run ruff format --check` | 427 files already formatted |
| `gda schema` sha256 vs base | identical (`c0710da1…8befd7`) |
| 86-surface `--help` byte-diff vs base | only `<root>` and `schema` differ |
| `gda skill` diff vs base | exactly the new Discovery paragraph |

Issue #672, opened 2026-08-15
scene create: document the Control-root zero-size layout consequence (or offer a preset)


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-006** (undated entry in the 2026-08-08 block). `gda scene create ui/main.tscn --root-type Control` creates a valid root with zero offsets and no viewport-filling layout. Container minimum sizes can make descendants visible, which hid the problem until `gda game rect` showed the root and its child layers at `[0, 0]` size.

## What to build

Documentation only. State the Control-root consequence and its fix in `scene create` help and the skill guidance:

- The behavior: a `Control` root is created with default zero anchors/offsets and does not fill the viewport.
- The fix: after `scene create`, set the root's anchors/offsets to fill the viewport (via `gda node set`; the dogfooded project set the composition root to the viewport rectangle during startup).
- The check: point at `gda game rect` as the command that exposes a zero-size root at runtime.

## Out of scope / follow-up

A layout preset flag on `scene create` (e.g. emit a full-rect root) is enhancement-shaped — it changes the input model and the emitted `--schema` — and is not part of this documentation issue. File it separately if wanted. (Feasibility note: whether a full-rect preset persists as plain properties or needs `set_anchors_preset()` at runtime was not verified.)

## Acceptance criteria

- [ ] `scene create` help or the skill states the zero-size Control-root behavior, the documented fix above, and the `game rect` check.

## Blocked by

None - can start immediately.

qinhaihong-red:
> *This was generated by AI during triage.*

Triage: moved to `ready-for-agent`. The issue body (revised 2026-08-15 after a full artifact review against the dogfooding source, the ADRs, and the current code) is the authoritative agent brief — current behavior, deliverables, acceptance criteria, and scope boundaries are specified there.

Issue #664, opened 2026-08-15
scene: static validation and bounded startup preflight


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-040** (2026-08-12) — `gda scene` has no `validate` operation; `scene get` parses the tree but yields no validity verdict on whether external resources resolve and script bindings load, so admission flows infer a weaker static result.
- **GDA-DF-030 "Static validation cannot preflight authored encounter invariants"** (2026-08-11; the document's GDA-DF numbering collides across sessions — identified by title and date) — script and scene validation (the available `scene get` read) passed while the first live launch rejected every encounter assembly due to a deterministic application invariant; compilation success can be misread as scene readiness. The notes ask for a bounded scene-start preflight that waits for `_ready`, captures assertion/error diagnostics, and returns a structured startup verdict distinct from static validation.

## What to build

Two additions; **slice 1 is `scene validate`**, the startup preflight is slice 2.

1. `scene validate <path>` (slice 1): parse the scene, resolve external resources and attached scripts, and return a stable validity verdict with diagnostics — distinct from `scene get`.
2. Bounded startup preflight (slice 2): boot the target scene through a gda-owned runner as a one-shot `Headless launch` — **not** a daemon `Engine session` — wait for `_ready` up to a configured bound (frames or seconds), capture script errors, and return a structured startup verdict. Script-error stderr parsing reuses the parser owned by #651.

## Acceptance criteria

- [ ] `scene validate` reports broken external resources and unloadable script bindings without starting an `Engine session`.
- [ ] The startup preflight distinguishes "compiles and loads" from "reaches `_ready` without script errors".
- [ ] A scene that never reaches `_ready` returns a structured timeout verdict within the configured bound, with captured diagnostics.
- [ ] Both results are structured and covered by e2e tests on a known-bad and known-good scene.

## Blocked by

None - can start immediately. Related: #651 (owns the script-error parser).

qinhaihong-red:
> *This was generated by AI during triage.*

Triage: moved to `ready-for-agent`. The issue body (revised 2026-08-15 after a full artifact review against the dogfooding source, the ADRs, and the current code) is the authoritative agent brief — current behavior, deliverables, acceptance criteria, and scope boundaries are specified there.

Issue #655, opened 2026-08-15
script run: configurable timeout that preserves captured output and fails fast on aborted scripts


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-012** (2026-08-08, reconfirmed 2026-08-09) — a script error aborted the test before `quit()`; the engine stayed alive, and after 120 s gda returned `launch_timeout` whose diagnostics contained only the timeout message. The script error Godot had already printed was discarded; rerunning the engine directly exposed it immediately.
- **GDA-DF-032 "Fixed one-shot timeout hides progress in production-sized suites"** (2026-08-11; the document's GDA-DF numbering collides across sessions — identified by title and date) — a healthy suite that grew past ~125 s was indistinguishable from a hang; no partial stdout, no elapsed data, and no way to raise the fixed 120 s ceiling.

## What to build

Scope: `gda script run`. The discard lives in the shared `Headless launch` primitive, so the sentinel channel (60 s default) and the export channel (600 s default) are affected too — they are named follow-up work, not covered by this issue's ACs.

- Expose a per-invocation `--timeout` for `script run`.
- On timeout, return the captured output inside the existing failure contract: the last N KiB of stdout/stderr go into the `Error envelope` `diagnostics` (the cap is fixed and stated in the message), together with elapsed seconds and a termination phase from a small enumerated set (e.g. `launched` / `script_started` / `killed_at_timeout`). Note: `diagnostics` is a string today; promoting these to structured envelope fields changes the uniform failure ABI (ADR-0004) and needs a recorded decision — string diagnostics are the smallest first step.
- Surface known Godot script-error lines captured before the timeout as diagnostics, reusing the parser owned by #651.
- Terminate early when the entry script aborts: support an **opt-in, user-declared completion marker** — gda kills the run and returns the captured error when the script's stderr shows a fatal script error and the declared marker has not appeared. ADR-0031 rejected imposing a gda-owned sentinel wrapper on user scripts; this marker is caller-declared, not imposed, and is not the ADR-0002 op-dispatch sentinel.

## Acceptance criteria

- [ ] A script whose runtime error prevents `quit()` returns the captured script error within a stated bound (seconds, not the full timeout) when a completion marker is declared.
[...]

PR #678, opened 2026-08-15
fix(gda): make gda the verdict authority for script run (ADR-0031 amendment)

### LINKED ISSUE #651: script run: script errors, missing scripts, and failing tests can all report success
## Dogfooding source

- **GDA-DF-007** (2026-08-08, ~6 reproductions) — dependency/entry parse errors and runtime script errors leave nested `exit_status: 0`; a suite-owned `completed:true` sentinel can even survive dependency compilation failure.
- **GDA-DF-017** (2026-08-09, ~5 reproductions) — a test that calls `SceneTree.quit(1)` yields nested `exit_status: 1` but the `gda` process exits `0`, so shell `&&` chains and conventional CI accept the failure.
- **GDA-DF-032 "Missing script can report wrapper success"** (2026-08-12, marked critical in the notes) — a nonexistent `res://` script returns `{"exit_status":0}` with the load errors only in `stderr`.

(The document's GDA-DF numbering collides across sessions; entries are identified here by title and date.)

Scope note: the missing-script and load/compile-failure shapes are defects — no reading of the contract calls them a successful run. The `quit(1)` passthrough is a **contract-change request**: ADR-0031 deliberately promotes the Raw run as the public `script run` result, where a non-zero `exit_status` can be a success.

## What to build

Make gda the verdict authority for `gda script run` instead of trusting the raw engine process exit code:

- Parse known Godot script-error stderr lines (Parse Error, SCRIPT ERROR, resource load failures) into structured diagnostics on the result. This issue owns that parser; #655 (timeout path) and #664 (scene startup preflight) reuse it.
- Treat a missing `res://` entry script and an entry/dependency load or compile failure as failed operations with new registered `Gda error code`s (e.g. `script_not_found`, `script_compile_failed`) added to the error-code registry — never as success.
- Under a documented strict flag, make an explicitly non-zero nested exit status exit the CLI with a **registered** exit code from the exit-code registry (the public CLI ABI — e.g. the operation-failure code, or a new registered code). Do not propagate the child status verbatim: a script's `quit(3)` would alias unrelated registry codes.
- Preserve the raw process status and stderr as secondary evidence in the structured result.

Decision prerequisite: the strictness contract needs an explicit decision against ADR-0031 (opt-in `--strict` vs a default change), recorded as an ADR-0031 amendment or superseding record. The two defect shapes above are covered by the same record.

## Acceptance criteria

- [ ] By default: a missing `res://` entry script returns a structured failure with a registered `Gda error code`, not `exit_status: 0`.
- [ ] By default: an entry script (or dependency) that fails to parse/compile produces a non-success verdict even when the engine process exits zero.
- [ ] Under the documented strict mode: an explicit `quit(1)` makes the `gda` process exit with a registered non-zero exit code, observable without parsing the JSON result. Default behavior is unchanged pending the ADR decision.
- [ ] Known script-error stderr lines are surfaced as structured diagnostics.
- [ ] The ADR-0031 amendment (or superseding record) lands with the change.
- [ ] Regression tests cover: missing script, parse-error dependency, runtime script error, explicit failing test.

## Blocked by

The ADR-0031 strictness decision above. No code dependency.

qinhaihong-red:
`gda script run` passed the engine's exit status through unconditionally, so several failures reported success. Godot exits **0** for a `--script` entry point it never ran, reporting the reason on stderr only.

Closes #651

## Behavior matrix

| shape | before | after |
| --- | --- | --- |
| missing `res://` entry script (GDA-DF-032) | `{"exit_status": 0}`, exit 0 | `script_not_found`, exit 4 |
| entry or preloaded dependency fails to parse/compile (GDA-DF-007) | `{"exit_status": 0}`, exit 0 | `script_compile_failed`, exit 4 |
| entry compiles but is not a `SceneTree`/`MainLoop` | `{"exit_status": 0}`, exit 0 | `incompatible_script_type`, exit 4 — but see the caveat below: this shape more often leaves the engine idling, surfacing as `launch_timeout` |
| runtime GDScript error the script survived (GDA-DF-007) | only in `stderr` prose | still a **success**, plus classified `diagnostics` |
| deliberate `quit(1)` (GDA-DF-017) | success, gda exits 0 | unchanged by default; `--strict` → `script_failed`, exit 4, carrying the script's own stdout+stderr |
| any of the above via a non-canonical spelling (`res://dir/../x.gd`) | verdict bypassed → phantom success, or misrouted code | same verdict as the canonical spelling |
| a running script's `load("res://missing.tres")` fails | success, `diagnostics: []` | still a **success**, now carrying `resource_load_failed` diagnostics |

The verdict is read from the engine's error stream, never from the exit code, and is keyed on the **entry script's own `res://` path — in canonical form**. Godot resolves an address before naming it in an error, so one canonical lexical identity (`.`/`..` segments and duplicate slashes collapsed) is fixed at the operation's input boundary and used for the spawn argv, every comparison and every message; the parser canonicalizes each address it publishes, and `entry_load_failure` canonicalizes both sides again. Keying on that path — a running script that itself fails to load some *other* resource stays a success. When a run emits several causes the earliest-stage one wins (missing > compile > parse > load > not-a-main-loop); a test pins that precedence against the code map.

**`--strict` is opt-in.** Per the triage decision the Raw-run promotion is why a user script keeps free use of its exit codes, so the default is untouched. The child status is **not** propagated verbatim — a `quit(3)` would alias `EXIT_VERSION` — so strict exits the registered `EXIT_OPERATION` (4).

## Two new modules

**`src/gda/engine_log.py`** — the engine's error-line format (`Logger::log_error`: the two-line `<TYPE>: <message>` / `at: <fn> (<file>:<line>)` pair plus the GDScript backtrace), **extracted from `gda.daemon.diag`**. It was a Phase-2 module, so a Phase-1 channel importing it reversed ADR-0000's component order and ADR-0040's module direction. Now a leaf importing nothing from `daemon`/`commands`, with both consumers importing downward. `gda.daemon.diag` re-exports `parse_errors`, so every existing consumer and its pinned tests are source-compatible; the move is mechanical and proved by the untouched suites (89 daemon unit tests, plus the diag/logger e2e).

**`src/gda/script_errors.py`** — what those lines *mean for a script run*. This issue owns it; #655 (timeout path) and #664 (scene startup preflight) build on it.

```
ScriptErrorKind        # closed public enum: parse_error | runtime_error | script_missing
                       # | load_failed | compile_failed | not_a_main_loop | resource_load_failed
                       # all but runtime_error mean the named resource could not be loaded or run
ScriptError            # pydantic model: {kind, message, path, line}; path always canonical
canonical_res_path(path: str) -> str
parse_script_errors(stderr: str) -> list[ScriptError]
entry_load_failure(errors: Sequence[ScriptError], script: str) -> ScriptError | None
```

Recognition is a **closed set**, so `diagnostics` stays curated rather than a second encoding of the whole error stream. A `kind` alone never decides the verdict: whether a load failure ended the *run* depends on whether it names the entry, which is `entry_load_failure`'s job.

## Public-surface diff (complete)

`gda schema` diff vs. the base commit: **`script run` only**, no other command's entry changed, no top-level change.

- **New CLI option** `gda script run --strict`.
- **Input schema** (`ScriptRunParams`): `+ strict: boolean, default false`. A params field, not argv-only, so `--params-json` / gda-mcp opt in identically (ADR-0015).
- **Output schema** (`ScriptRunResult`): `+ diagnostics: ScriptError[]` (present-but-empty on a clean run), plus two new `$defs` (`ScriptError`, `ScriptErrorKind`). `exit_status` / `stdout` / `stderr` unchanged; `kind`, `constraints` and the `error` envelope unchanged.
- **`ScriptErrorKind` enum values** (public ABI): `parse_error`, `runtime_error`, `script_missing`, `load_failed`, `compile_failed`, `not_a_main_loop`, **`resource_load_failed`** (added for the resource-load sentences).
- **`ScriptError.path` semantics**: always the **canonical** form of the address, on every diagnostic.
- **`gda skill` output**: the `script` row's `run` clause now documents `--strict` and the never-ran failures. SKILL.md is otherwise untouched (sibling-owned regions avoided).
- **Parser-module extraction: no surface change at all** — verified by the schema diff below.
- **Command description / `--help`**: rewritten to document the three never-ran verdicts and `--strict`, including exactly what the `script_failed` envelope's `diagnostics` carries, and the caveat that the not-a-main-loop shape may instead surface as `launch_timeout`.
- **`ScriptErrorKind` / `ScriptError.kind` description strings** (public ABI via `gda schema`): corrected to say five never-ran kinds, not three.
- **Error-code registry** (Python + ADR-0002 table, kept in lockstep by `tests/test_error_registry.py`):
  - `+ script_not_found` — `operation` / `classifier` / exit `4`
  - `+ script_failed` — `operation` / `classifier` / exit `4`
  - `~ script_compile_failed` — description broadened; category/source/exit **unchanged**
  - `~ incompatible_script_type` — description broadened to cover `script run`'s entry-point requirement; category/source/exit **unchanged**
  - Both broadenings are description-only and so are **not** a `gda schema` diff (the schema's `error` is the envelope shape, which does not enumerate codes) — but they are a public-doc diff, enumerated here.
- **`GdaError.diagnostics` content for `script_failed`**: both script streams under fixed labels. `diagnostics` is a free-form `str` (ADR-0004), so this is **not** a shape change; ADR-0002's prose is synced to say so.
- **No new exit codes** — `exit_codes.py` untouched.
- **No ADR-0004 envelope shape change.**

## ADR amendments

- **ADR-0031** — dated amendment (repo convention: the decision text is preserved, not rewritten). It draws the line the original left implicit: **gda is the authority on whether the engine RAN the script; the script stays the authority on what its own exit status means.** "Synthesize `script_failed`" stays **rejected as the default**; `--strict` is the opt-in expression of it. It also records that the two diagnostics channels are deliberately asymmetric — structured `ScriptError[]` on success results, prose on the ADR-0004 failure envelope (including the child's numeric status, which survives only as message prose). Giving the failure channel structured diagnostics is an ADR-0004 decision, deferred to its dedicated owner **#687** (#655's timeout envelope names the same constraint and adopts #687's outcome).
- **ADR-0002** — one dated scope note, plus the two living registry table rows. The note records that **both** rules it qualifies presuppose the sentinel channel, which ADR-0031's passthrough shape does not have: "stderr is never parsed for the outcome" (there, the error stream is the only evidence of the outcome) and "copies stderr into diagnostics" (there, `--strict` carries the script's own stdout *and* stderr, `diagnostics` being a free-form string). Neither rule is relaxed. Per repo convention the decision text is **untouched** — the ADR-0002 prose diff is pure insertion, exactly like the ADR-0031 amendment.

## Acceptance criteria

| #651 criterion | Status | Where |
| --- | --- | --- |
| By default, a missing `res://` entry script returns a structured failure with a registered code, not `exit_status: 0` | **Satisfied** | `script_not_found`, exit 4; unit + real-engine e2e, incl. non-canonical spellings |
| By default, an entry script (or dependency) that fails to parse/compile is a non-success verdict even though the engine exits 0 | **Satisfied** | `script_compile_failed`, exit 4; unit + real-engine e2e, incl. non-canonical spellings |
| Under `--strict`, an explicit `quit(1)` exits with a registered non-zero code, observable without parsing the JSON; default unchanged | **Satisfied** | `script_failed`, exit 4 (`EXIT_OPERATION`, never the child's status); default arm pinned by its own test |
| Known script-error stderr lines surfaced as structured diagnostics | **Satisfied** | 7-kind `ScriptError[]` incl. resource-load failures; canonical `path`, `line` where the engine reports one |
| The ADR-0031 amendment lands with the change | **Satisfied** | Dated amendment; ADR-0002 gets dated scope notes only, accepted prose untouched |
| Regression tests: missing script, parse-error dependency, runtime script error, explicit failing test | **Satisfied** | All four at unit tier and, except the engine-nondeterministic not-a-main-loop shape, at real-engine e2e tier |
| Preserve the raw process status and stderr as secondary evidence in the structured result | **Satisfied on the success path; PARTIAL on failure paths** | See below |

**The one partial, stated plainly.** On the **success** path the criterion is fully met: `exit_status`, `stdout` and `stderr` are carried verbatim, alongside the typed `ScriptError[]` diagnostics. On **failure** paths the evidence is preserved but **not typed**:

- the parsed `ScriptError[]` is *not* carried — `GdaError.diagnostics` is a free-form `str` (ADR-0004), so a verdict envelope carries the engine's stderr as prose;
- under `--strict` the script's own stdout **and** stderr are carried, but as labelled prose sections (`--- script stdout ---` / `--- script stderr ---`);
- the child's numeric exit status survives only inside the message text, not as a field.

Typed failure-path evidence would change the ADR-0004 `GdaError` envelope — a cross-cutting ABI decision that **#687 owns** (a dedicated issue created at review reconciliation; #655's authoritative scope covers only the string-diagnostics timeout form), and that this PR's own ADR-0031 amendment explicitly defers rather than pre-empting. The gap is recorded there, not left implicit, and is also recorded on issue #651 so the eventual closure is honest. **Residual risk, accepted:** a `--strict` consumer reads the status from message text and splits the labelled sections to separate the streams.

## Validation

Run in the worktree; every gate below was executed, and re-executed after the review round.

| gate | result |
| --- | --- |
| `uv run pytest -m "not e2e"` | **1246 passed**, 396 deselected |
| `uv run ruff format --check` | 428 files formatted |
| `uv run pytest -m e2e tests/test_e2e_script_run.py` (real Godot 4.6.3) | **18 passed** in 7.4s |
| `uv run pytest -m e2e` — diag / logger / script (guards the parser-module move on the real daemon path) | **58 passed** |
| `uv run pytest -m e2e` — script / scene-security / params-json / mcp-stdio / mcp-tracer | **67 passed** |
| `gda schema` + `gda script run --help` before/after diff | only the intended diffs above |

Unit fixtures are stderr **recorded verbatim from real Godot 4.6.3 runs**, not invented lines; the e2e arms then prove the engine really does exit 0 for a script that never ran. The `--strict` evidence row was re-verified by a real run: default → exit 0 with `exit_status: 1`; `--strict` → exit 4 with the suite's printed `FAIL test_damage…` lines inside `diagnostics` (empty before this round, since the suite writes nothing to stderr).

## Decisions worth a reviewer's attention

- **`script_compile_failed` and `incompatible_script_type` are reused, not minted.** Both already name these exact conditions for `script attach`. Reuse-vs-mint is decided by **semantic match**, not by a code's `source` field — `source` records canonical origin, and reusing across sites is established practice in `errors.py`. `script_not_found` is minted because `path_not_found` *means* something else: "a file the operation was asked to act on is absent" vs. "the engine could not load the entry script, so the run never happened", which agents branch on differently.
- **No e2e arm for the not-a-main-loop shape.** The engine has **two** behaviours for it and only one is a phantom success: it either errors and exits 0 (the shape this fix closes — pinned at unit level against real captured stderr), or it errors and **keeps running**, falling through to the project's main loop and idling forever. A clean fixture project does the latter, which is already a failure via `launch_timeout` (the #655 path), so gda never reports success either way. The exit-0 variant is not reproducible from project contents — the import-cache and broken-sibling-script hypotheses were both tested and refuted — so forcing an e2e arm would mean a 120s timeout per run for no added signal. Both behaviours are recorded in the e2e module docstring.

## Shipped agent guidance

The `--strict` guidance lands in `src/gda/skill/SKILL.md` itself — PR-body text is not shipped agent guidance. One clause on the `script` row (line 71), clear of the regions sibling PRs edit:

> `run` executes a `res://` script one-shot and passes its `exit_status`/`stdout`/`stderr` through — a non-zero `quit()` is still success, so read `exit_status`, or pass `--strict` to get a `script_failed` failure (exit 4) whose `diagnostics` carries the script's own stdout and stderr; a script that never ran — missing, or a failed parse/compile — always fails

`README.md` is owned by #672 this wave and was **not** touched.

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -130,7 +130,9 @@ operation, and parse codes the CLI assigns).
 | `unsupported_property_type` | `operation` | `operation` | `4` | An Object-typed property expects a type `node set` / `resource set` cannot yet assign a `res://` resource to: a script `class_name`-typed property (deferred to the ADR-0032 resolver) or an Object property with no declared engine class (ADR-0033, #363). |
 | `no_search_match` | `operation` | `operation` | `4` | A search-replace script edit found no occurrence of the search string. |
 | `invalid_line_range` | `operation` | `operation` | `4` | A line-range script edit specified lines outside the script's bounds, or end before start. |
-| `script_compile_failed` | `operation` | `operation` | `4` | A script could not be attached to a node because it does not compile. |
+| `script_compile_failed` | `operation` | `operation` | `4` | A script does not compile, so the requested work could not proceed: `script attach` refuses to bind it to a node, and `script run` reports that the entry script (or a dependency it preloads) never ran (#651). |
+| `script_not_found` | `operation` | `classifier` | `4` | A `script run` entry script does not exist in the project, so the engine never ran it — read from stderr, since the engine still exits 0 (#651). |
+| `script_failed` | `operation` | `classifier` | `4` | A `script run --strict` script ran to completion and chose a non-zero exit status; strict mode maps that opted-in failure onto the uniform error envelope. Never reported without `--strict` (ADR-0031 amendment, #651). |
 | `incompatible_script_type` | `operation` | `operation` | `4` | A script compiles but its native base type is incompatible with the target node's type. |
 | `signal_not_found` | `operation` | `operation` | `4` | A requested signal does not exist on the source node. |
 | `already_connected` | `operation` | `operation` | `4` | A signal is already connected to the target node's method. |
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -63,19 +63,34 @@ An operation failure payload has this wire shape:
 The GDScript payload owns only `code` and `message`. `gda` owns the public
 `GdaError` wrapper: it validates that the code is registered, assigns the
 `operation` category, preserves the message, and copies stderr into diagnostics.
+(`diagnostics` is a free-form string, so a channel with better evidence to offer
+may fill it differently: `script run --strict` carries the user script's own
+stdout *and* stderr there under fixed labels, because for that failure the
+script's output is the diagnostic — see the ADR-0031 amendment.)
 
 ### stderr as advisory diagnostics
 
-stderr is still **never** parsed for the success/failure *outcome* or for stable
-error codes — those come only from the exit code and the stdout sentinel, as
-above. A command **may**, however, surface engine error text from stderr as
-**advisory, best-effort diagnostics** on its *success* result, when a useful
-detail is available nowhere else. `script validate` (#118) is the established
-case: when it reports `valid=false`, the per-error `line` and `message` exist
-only in the engine's stderr (no bound API exposes them), so `gda` parses them
-into the result's `diagnostics`. This stays within the contract: the diagnostics
-are advisory (they may hold only the first error, and `column` is unavailable on
-the standard build), and they never determine the outcome or a stable code.
+On **this sentinel channel**, stderr is still **never** parsed for the
+success/failure *outcome* or for stable error codes — those come only from the
+exit code and the stdout sentinel, as above. A command **may**, however, surface
+engine error text from stderr as **advisory, best-effort diagnostics** on its
+*success* result, when a useful detail is available nowhere else. `script
+validate` (#118) is the established case: when it reports `valid=false`, the
+per-error `line` and `message` exist only in the engine's stderr (no bound API
+exposes them), so `gda` parses them into the result's `diagnostics`. This stays
+within the contract: the diagnostics are advisory (they may hold only the first
+error, and `column` is unavailable on the standard build), and they never
+determine the outcome or a stable code.
+
+> **Scope note (2026-08-15, #651).** The rule above governs the sentinel channel,
+> where an outcome is always available from the exit code plus the sentinel.
+> ADR-0031's `script run` is a **different execution shape**: the entry script is
+> the user's own, so it emits no sentinel, and the engine exits `0` even when it
+> never ran the script. There, the engine's error stream is the *only* evidence of
+> the outcome, so that channel **does** derive its verdict from parsed stderr —
+> keyed on recognized engine sentences, never on free-text matching. This does not
+> relax the sentinel rule; it records that the sentinel rule presupposes a sentinel.
+> See the ADR-0031 amendment.
 
 ## `GdaError.code` registry
 
@@ -133,7 +148,7 @@ operation, and parse codes the CLI assigns).
 | `script_compile_failed` | `operation` | `operation` | `4` | A script does not compile, so the requested work could not proceed: `script attach` refuses to bind it to a node, and `script run` reports that the entry script (or a dependency it preloads) never ran (#651). |
 | `script_not_found` | `operation` | `classifier` | `4` | A `script run` entry script does not exist in the project, so the engine never ran it — read from stderr, since the engine still exits 0 (#651). |
 | `script_failed` | `operation` | `classifier` | `4` | A `script run --strict` script ran to completion and chose a non-zero exit status; strict mode maps that opted-in failure onto the uniform error envelope. Never reported without `--strict` (ADR-0031 amendment, #651). |
-| `incompatible_script_type` | `operation` | `operation` | `4` | A script compiles but its native base type is incompatible with the target node's type. |
+| `incompatible_script_type` | `operation` | `operation` | `4` | A script compiles but its base type is incompatible with the requested use: `script attach`'s target node type, or `script run`'s requirement that a one-shot entry script extend `SceneTree`/`MainLoop` (#651). |
 | `signal_not_found` | `operation` | `operation` | `4` | A requested signal does not exist on the source node. |
 | `already_connected` | `operation` | `operation` | `4` | A signal is already connected to the target node's method. |
 | `connection_not_found` | `operation` | `operation` | `4` | A requested signal-to-method connection does not exist on the source node. |
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -105,6 +105,27 @@ the standard build), and they never determine the outcome or a stable code.
 > `invalid_path` already does for a CLI-side path rejection. The GDScript mirror is
 > derived from operation-source *membership*, which such reuse does not change.
 
+> **Scope note, extended 2026-08-16 (#651 review recheck) — naming the sentences
+> this qualifies.** Two accepted sentences read as an exclusive rule and are the
+> ones the paragraph above narrows: the registry section's "GDScript mirrors only
+> the rows whose source is `operation`, because only those codes can be reported by
+> headless operations", and the `Source` column's implied "the source that may
+> report it".
+>
+> Both stay true of what they actually govern — **mirror membership and the
+> sentinel channel**: `operations.gd` declares exactly the `operation`-source rows,
+> and only those may come back through the ADR-0002 sentinel as an operation's own
+> report. Neither constrains the **Python classifier**, which may assign any
+> registered code whose semantics match the failure it recognized, including an
+> `operation`-source one — a practice `invalid_path` established long before #651.
+>
+> So: `source` is an **origin-and-membership** field, never an emitter whitelist.
+> Nothing derives behavior from the exclusive reading — `classify_run` keys on
+> operation-source *membership* to validate a sentinel-reported code, and the
+> mirror test keys on the same membership — so this is a correction to the recorded
+> meaning, not a change to the contract. `src/gda/error_codes.py`'s module
+> docstring is the single home of this definition.
+
 ## `GdaError.code` registry
 
 `GdaError.code` values are a public ABI for agents. Their authoritative source is
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -97,6 +97,13 @@ the standard build), and they never determine the outcome or a stable code.
 >
 > Neither relaxes the sentinel rules; both record that those rules presuppose a
 > sentinel. See the ADR-0031 amendment.
+>
+> One consequence for the registry below: a row's `source` names the code's
+> **authoritative origin channel**, not an exclusive list of what may emit it. When
+> the classifier recognizes the same semantic failure from the engine's output
+> rather than from a sentinel, it may assign an `operation`-source code — as
+> `invalid_path` already does for a CLI-side path rejection. The GDScript mirror is
+> derived from operation-source *membership*, which such reuse does not change.
 
 ## `GdaError.code` registry
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -63,34 +63,40 @@ An operation failure payload has this wire shape:
 The GDScript payload owns only `code` and `message`. `gda` owns the public
 `GdaError` wrapper: it validates that the code is registered, assigns the
 `operation` category, preserves the message, and copies stderr into diagnostics.
-(`diagnostics` is a free-form string, so a channel with better evidence to offer
-may fill it differently: `script run --strict` carries the user script's own
-stdout *and* stderr there under fixed labels, because for that failure the
-script's output is the diagnostic — see the ADR-0031 amendment.)
 
 ### stderr as advisory diagnostics
 
-On **this sentinel channel**, stderr is still **never** parsed for the
-success/failure *outcome* or for stable error codes — those come only from the
-exit code and the stdout sentinel, as above. A command **may**, however, surface
-engine error text from stderr as **advisory, best-effort diagnostics** on its
-*success* result, when a useful detail is available nowhere else. `script
-validate` (#118) is the established case: when it reports `valid=false`, the
-per-error `line` and `message` exist only in the engine's stderr (no bound API
-exposes them), so `gda` parses them into the result's `diagnostics`. This stays
-within the contract: the diagnostics are advisory (they may hold only the first
-error, and `column` is unavailable on the standard build), and they never
-determine the outcome or a stable code.
+stderr is still **never** parsed for the success/failure *outcome* or for stable
+error codes — those come only from the exit code and the stdout sentinel, as
+above. A command **may**, however, surface engine error text from stderr as
+**advisory, best-effort diagnostics** on its *success* result, when a useful
+detail is available nowhere else. `script validate` (#118) is the established
+case: when it reports `valid=false`, the per-error `line` and `message` exist
+only in the engine's stderr (no bound API exposes them), so `gda` parses them
+into the result's `diagnostics`. This stays within the contract: the diagnostics
+are advisory (they may hold only the first error, and `column` is unavailable on
+the standard build), and they never determine the outcome or a stable code.
 
-> **Scope note (2026-08-15, #651).** The rule above governs the sentinel channel,
-> where an outcome is always available from the exit code plus the sentinel.
-> ADR-0031's `script run` is a **different execution shape**: the entry script is
-> the user's own, so it emits no sentinel, and the engine exits `0` even when it
-> never ran the script. There, the engine's error stream is the *only* evidence of
-> the outcome, so that channel **does** derive its verdict from parsed stderr —
-> keyed on recognized engine sentences, never on free-text matching. This does not
-> relax the sentinel rule; it records that the sentinel rule presupposes a sentinel.
-> See the ADR-0031 amendment.
+> **Scope note (2026-08-15, #651) — the two rules above presuppose the sentinel
+> channel, which ADR-0031's `script run` does not have.** Both the "stderr is never
+> parsed for the outcome" rule and the "copies stderr into diagnostics" wrapper
+> behaviour are stated for the sentinel pipeline, where an outcome is always
+> available from the exit code plus the stdout sentinel, and where the engine's
+> stderr is the only diagnostic on offer. `script run` (ADR-0031) is a **different
+> execution shape**: the entry script is the user's own, so it emits no sentinel,
+> and the engine exits `0` even when it never ran the script. Two consequences,
+> both scoped to that channel:
+>
+> - **Outcome.** The engine's error stream is the *only* evidence that the script
+>   never ran, so that channel **does** derive its verdict from parsed stderr —
+>   keyed on recognized engine sentences, never on free-text matching.
+> - **Diagnostics.** `GdaError.diagnostics` is a free-form string, so a channel
+>   with better evidence to offer may fill it differently: `script run --strict`
+>   carries the user script's own stdout *and* stderr there under fixed labels,
+>   because for that failure the script's output *is* the diagnostic.
+>
+> Neither relaxes the sentinel rules; both record that those rules presuppose a
+> sentinel. See the ADR-0031 amendment.
 
 ## `GdaError.code` registry
 
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -11,9 +11,15 @@ status: accepted
 > left implicit, and adds one opt-in inversion.
 >
 > **1. A script that never ran is a failure, by default.** Godot reports **three** such shapes
-> **only on stderr — and still exits `0`** (verified against Godot 4.6.3): a missing `--script` entry
-> point; an entry script (or a dependency it preloads) that fails to parse or compile; and one that
-> compiles but does not extend `SceneTree`/`MainLoop`, so it can never be a one-shot entry point.
+> **only on stderr — and still exits `0`**: a missing `--script` entry point; an entry script (or a
+> dependency it preloads) that fails to parse or compile; and one that compiles but does not extend
+> `SceneTree`/`MainLoop`, so it can never be a one-shot entry point. The first two were verified
+> against Godot 4.6.3 as reliably reproducible. The **third is not**: its exit-0 form is real and
+> captured, but the engine more often prints nothing and simply keeps idling — falling through to
+> the project's main loop — which surfaces as `launch_timeout` (#655). That is a failure by another
+> route, so gda never reports success for it either way; only the exit-0 form is what this decision
+> has to correct.
+>
 > Passing that status through made `gda script run res://does-not-exist.gd` return
 > `{"exit_status": 0}`, which no reading of the contract calls a successful run. These are
 > **defects, not the contract**: the passthrough exists because gda does not know a user script's
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -10,16 +10,29 @@ status: accepted
 > decision below is unchanged for a **completed** run; this note draws the line the original text
 > left implicit, and adds one opt-in inversion.
 >
-> **1. A script that never ran is a failure, by default.** Godot reports a missing `--script` entry
-> point, and an entry script (or a dependency it preloads) that fails to parse or compile, **only on
-> stderr — and still exits `0`** (verified against Godot 4.6.3). Passing that status through made
-> `gda script run res://does-not-exist.gd` return `{"exit_status": 0}`, which no reading of the
-> contract calls a successful run. These are **defects, not the contract**: the passthrough exists
-> because gda does not know a user script's *semantics*, and a script that never started has none.
-> They now classify to the registered `script_not_found` / `script_compile_failed` **Error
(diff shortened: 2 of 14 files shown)


qinhaihong-red:
Adversarial review pinned to `e3579edd8af2d11b3b128cc5136abe50273d9ec1` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

**[P1] Canonicalize `res://` identity before deciding the entry verdict — `src/gda/script_errors.py:217`.**

`entry_load_failure()` compares a diagnostic path with the caller's raw spelling. Godot canonicalizes dot segments: a real Godot 4.6.3 run of broken `res://bad.gd` invoked as `res://dir/../bad.gd` exited 0 and reported `res://bad.gd`; this head therefore returned `ScriptRunResult(exit_status=0)` instead of `script_compile_failed`. A similarly spelled missing entry can also receive the wrong code. Establish one canonical lexical resource identity at the input boundary (or compare canonical equivalents), and cover alternate-spelling missing and compile-failed entries with real-engine regressions.

**[P2] Put the shared engine-log parser below both CLI and daemon consumers — `src/gda/script_errors.py:48`.**

The new script verdict path imports `gda.daemon.diag.parse_errors`, making bottom-layer command/core behavior depend on the higher daemon implementation. This contradicts ADR-0000's component order and ADR-0040's directed module graph. Extract the dependency-free engine error-line parser to a neutral lower module; let `daemon.diag` and `script_errors` import it downward, while script-specific verdict classification remains owned by `script_errors`.

Claim evidence state for both findings is `open`: live canonicalization disconfirms raw-string identity, and the exact import graph disconfirms the claimed downward dependency.

## `review`

### Standards

**[P1] ADR-0031's “missing/non-compiling entries fail” contract is bypassed by alternate resource spellings — `src/gda/script_errors.py:217`.**

The live reproduction above is a phantom success on the exact head. Normalize or reject noncanonical input before verdict matching; add `.`, `..`, duplicate-slash, missing, and compile-failed real-engine cases.

**[P2] The error-code registry's `source` authority is no longer truthful — `src/gda/error_codes.py:358-365`.**

`ErrorCodeSpec` defines `source` as what may report a code, and ADR-0002 derives GDScript mirroring from that field. The new comment redefines it as “canonical origin,” while `commands/script.py:779-783` classifier-emits operation-source codes. Preserve semantic reuse, but make reporting sources multi-valued or add a separate reportability axis, then reconcile the registry, ADR table, and mirror derivation.

**[P2] The new parser dependency reverses ADR-0040's direction — `src/gda/script_errors.py:48`.**

This is the same architectural edge identified above. Move the neutral log parser below both consumers; do not make Phase 1 depend on daemon ownership or on speculative future consumers.

### Spec

**[P1] Compare canonical resource identities — `src/gda/script_errors.py:217`.**

Issue #651 requires an entry or dependency parse/compile failure to produce a non-success verdict even when Godot exits zero. Alternate valid `res://` spellings still violate that acceptance criterion and can also misclassify `script_not_found`.

**[P2] Parse the requested resource-load failures — `src/gda/script_errors.py:297`.**

An exact-head run of a script calling `load("res://missing_resource.tres")` returned success with `diagnostics: []`, while stderr contained both `Cannot open file` and `Failed loading resource`. Issue #651 explicitly includes resource-load lines among the known structured diagnostics. Add those patterns and a runtime fixture; only turn them into an entry verdict when canonical entry identity matches.

**[P2] Preserve typed raw evidence on failure results — `src/gda/commands/script.py:779`, `src/gda/errors.py:436,493`.**

The issue requires raw process status and stderr as secondary evidence in the structured result. Never-ran and strict branches return a `GdaError`: parsed `ScriptError[]` is discarded, and the raw status is absent as data (strict mode keeps it only in message prose). Add an optional typed script-run evidence object carrying raw status, stderr, and parsed diagnostics.

Validation: exact base/head and merge-base pinned; `git diff --check` passed; review subpasses passed 166 focused tests at the broadest boundary. I independently reproduced both the canonical-path phantom success and the empty-diagnostics resource-load result with Godot 4.6.3. Full suite and MCP stdio end-to-end were not run.

Summary: DDMA — 2 findings (worst P1); Standards — 3 (worst P1); Spec — 3 (worst P1).

qinhaihong-red:
## Review response — 6 adopted, 1 partially adopted, 1 declined

Thanks for the review; the two P1-class reproductions were correct and both were real defects. Everything is on **`bfd7751a`**, pinned against the head you reviewed (`e3579edd`).

| # | Finding | Decision | Evidence | Change / retained behavior | Validation |
| 1–3 | **Non-canonical `res://` spellings bypass the entry verdict** (P1 ×3). `res://dir/../bad.gd` → engine reports `res://bad.gd` → no match → phantom success; also misroutes `script_not_found` | **Fully adopted** | Reproduced on a clean project: a broken entry via `res://sub/../bad.gd` returned `exit_status: 0`. The missing-entry capture shows *both* spellings in one stderr — the resource layer reports canonical `res://gone.gd`, while `main.cpp` echoes the raw argv verbatim, which is exactly why the raw comparison matched only the generic "Can't load script:" line and downgraded the code | New `canonical_res_path()` establishes **one canonical lexical identity** (collapse `.`/`..` and duplicate slashes; purely lexical, no filesystem access — the missing-file case has nothing to stat). Fixed at the operation's input boundary and used for the **spawn argv**, every comparison, and every message; the parser canonicalizes every address it publishes; `entry_load_failure` canonicalizes **both sides** again so a caller passing hand-built errors cannot defeat it | 11 unit cases on the canonicalizer (incl. `res://`, `res://../outside.gd`, non-`res://` passthrough, idempotence); parametrized unit + **real-engine e2e** arms for missing and compile-failed entries via `./`, `../` and duplicate-slash spellings — 6 new e2e cases, all green |
| 4–5 | **`script_errors` imports `gda.daemon.diag`**, reversing ADR-0000 component order / ADR-0040 module direction (P2 ×2) | **Fully adopted** | Correct as stated: a Phase-1 channel reached up into a Phase-2 package | Extracted the dependency-free error-line parser to a new leaf module **`src/gda/engine_log.py`** (imports nothing from `daemon`/`commands`); `gda.daemon.diag` and `gda.script_errors` both import downward. `daemon.diag` **re-exports `parse_errors`**, so the daemon server and every pinned test stay source-compatible. Script-specific verdict classification stays in `script_errors` | Pure move, proved by **untouched** suites: 89 daemon/diag/logger unit tests and 58 diag/logger/script **e2e** tests pass with no edits; the `gda schema` diff is unchanged by the move |
| 6 | **Registry `source` meaning drifted** — comment redefined it as "canonical origin" while the classifier emits operation-source codes (P2 Standards) | **Partially adopted** | The inconsistency in the recorded prose was real; the underlying design was not | Recorded truth made exact instead: `source` names the code's **authoritative origin channel**, not an exclusive emitter list; the classifier may additionally assign an operation-source code when it recognizes the same semantic failure from the engine stream (**`invalid_path` precedent, which predates this PR**); mirror derivation keys on operation-source **membership**, which classifier reuse neither adds to nor removes from. Matching sentence added to the ADR-0002 dated note | `tests/test_error_registry.py` (ADR-0002 table ↔ Python registry ↔ `operations.gd` mirror) green, unchanged |
| 7 | **Resource-load failure lines are not parsed** — `load("res://missing.tres")` → success with `diagnostics: []` (P2 Spec) | **Fully adopted** | Correct; #651 names resource load failures among the known stderr lines | Both engine sentences parsed — `Cannot open file '<path>'` (format loader) and `Failed loading resource: <path>.` (ResourceLoader; the trailing period is sentence punctuation, not part of the address). New `resource_load_failed` kind carrying the resource path. They flip the entry verdict **only** when the canonical entry identity matches, so a runtime load of another resource stays a success-with-diagnostics | Real-engine e2e for the runtime case (success, `exit_status 0`, diagnostics naming `res://missing.tres`); unit case for the entry-preload case (engine blames the **entry** with a parse error + load failure, resource named only in the message → `script_compile_failed`); the missing-entry fixture updated — it emits this sentence too, and `script_missing` still outranks it |
| 8 | **Add an optional typed script-run evidence object on failure results** (P2 Spec) | **Not adopted** | — | This is precisely the ADR-0004 failure-envelope shape change that this PR's own ADR-0031 amendment **explicitly defers to #655**, and the amendment already records the resulting limitation in writing: failure-path diagnostics are prose, the structured `ScriptError[]` is success-only, and the child's numeric status survives only as message prose. Adopting it here would change a cross-cutting ABI (`GdaError`) inside a behavior-fix PR, ahead of the decision that owns it. **Residual risk, accepted and recorded:** a `--strict` consumer must read the exit status from the message text rather than a field, and must split the labelled `--- script stdout ---` / `--- script stderr ---` sections to separate the streams. The labelled prose is the evidence the current shape allows, and it is stable and documented | Covered by the existing strict-evidence unit tests (exact layout, including the empty-stream case) and a real-engine e2e asserting a suite's `print()` output survives into the envelope |

### Also declined, for the record

- **Escaping the `--- script stdout/stderr ---` labels against a script that prints them.** Accepted under ADR-0009's Trusted-project assumption — the same assumption that lets `script run` execute arbitrary project code in the first place. A script that forges the label can already do anything the label could mislead about.
- **Human-mode rendering of a strict failure as single-line JSON.** Pre-existing behavior of the shared failure channel across the whole surface, not introduced here; noted as a follow-up candidate rather than changed inside this PR.

### Validation on `bfd7751a`

`uv run pytest -m "not e2e"` → **1246 passed**; `ruff check`, `ruff format --check`, `pyright` → clean; `pytest -m e2e tests/test_e2e_script_run.py` (real Godot 4.6.3) → **18 passed**; diag/logger/script e2e → **58 passed** (the parser-move guard). The `gda schema` diff versus the base commit is still confined to `script run` — the module extraction changes no public surface.

qinhaihong-red:
Follow-up recheck pinned to exact head `bfd7751aee78a68bd662c23749cc6513db2b3d05` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

The canonical `res://` identity defect, parser dependency inversion, and resource-load diagnostic gap are closed. Focused validation passed (`91` unit tests and `18` real-Godot script-run e2e tests). Two P2 residues remain.

**[P2] Reconcile the authoritative meaning of `source` — `src/gda/error_codes.py:3-6,37-43,358-369`; `docs/adr/0002-headless-structured-output-contract.md:101-113`.**

The new note defines `source` as an authoritative origin and explicitly non-exclusive, while the module/dataclass still define it as the source that may report the code, and both the module and ADR still say only `operation` rows can be reported by headless operations. Those claims cannot all be true once the classifier emits operation-source codes. The fix added an exception note but did not reconcile the registry/ADR/mirror authority requested by the finding.

Smallest closure: choose one stable meaning and update every authoritative definition consistently. If the field owns origin/mirror membership rather than reportability, name and document that meaning directly (or model reportability separately) instead of retaining contradictory “may report” / “only those codes can be reported” prose.

**[P2] The failure-path evidence contract remains unresolved — `src/gda/errors.py:418-440,473-497`.**

Declining the proposed typed evidence object is a defensible ADR-0004 scope decision, but it does not close issue #651's stated requirement to preserve raw process status/stderr as secondary evidence and surface known script-error lines as structured diagnostics. On the current head, a never-ran failure still returns only raw stderr in the generic string `error.diagnostics`, with the engine status absent; strict mode keeps the child status only in message prose and combines streams under string headers. The parsed `ScriptError[]` available on the success channel is discarded on both failure paths.

Smallest closure: either add an ADR-0004-compatible optional typed evidence extension, or explicitly reconcile issue #651/ADR-0031 so failure-path typed diagnostics and raw status are deferred to #655 without claiming this PR satisfies that part of #651. The implementation proposal was declined; the underlying contract mismatch was not disproved.

qinhaihong-red:
## Recheck response — both P2 residues closed on `2d8f0b0d`

Thanks for the recheck. Both residues were fair; one was a real inconsistency in the recorded meaning, the other a scope-honesty problem in how this PR described itself.

| Residue | Decision | Change / retained | Validation |
| --- | --- | --- | --- |
| **1 — `source` still contradictory across authorities.** The dated ADR-0002 note called it a non-exclusive authoritative origin, while `error_codes.py`'s module docstring and the `ErrorCodeSpec` field doc still said "the source that may report it", and both the module and ADR-0002 still stated only `operation` rows are reported by headless operations | **Fully adopted**, one stable meaning | **`source` names a code's authoritative origin channel and governs GDScript-mirror membership; it is not an exclusive emitter list — the Python classifier may additionally assign any code whose semantics match (the `invalid_path` precedent).** The module docstring is now the single home of that definition; `ErrorCodeSource` and `ErrorCodeSpec.source` point at it instead of restating it, and the #651 per-code comment dropped its duplicated copy. ADR-0002's accepted prose is **untouched** — the dated scope note is extended instead, and now names the two accepted sentences it qualifies ("GDScript mirrors only the rows whose source is `operation`, because only those codes can be reported by headless operations" and the `Source` column's implied "the source that may report it"), so a reader hits the qualification from either direction | Audited every consumer: **nothing derives behavior from the exclusive reading.** `classify_run` keys on operation-source *membership* to validate a sentinel-reported code, and the mirror test keys on the same membership — classifier reuse adds and removes no member. `tests/test_error_registry.py` green unchanged; full fast suite **1246 passed** |
| **2 — the PR claimed #651's raw-evidence AC was satisfied, while failure paths discard the parsed `ScriptError[]`, keep the status in message prose, and use a generic string** | **Adopted via reconciliation** — no code change | The PR body now carries an explicit **acceptance-criteria table**. The raw-evidence criterion is marked **satisfied on the success path, PARTIAL on failure paths**, listing exactly what is and is not typed there (no `ScriptError[]`; labelled prose streams under `--strict`; numeric status in message text only), with typed failure-path evidence deferred to **#655**, which owns the ADR-0004 envelope decision. No claim of full satisfaction remains. The ADR-0031 amendment gained one clause bounding the same thing, so the ADR and the issue record say the same thing; the deferral rationale and residual risk are unchanged from the previous round | Recorded on issue #651 as well, so the eventual closure at promotion time is honest rather than inferred from a PR body |

**Residue 2 was closed by reconciliation, not by code** — deliberately. Implementing the typed evidence object here would change `GdaError` for every command in the surface, inside a behavior-fix PR, ahead of the decision that owns it. The declined-with-rationale position from the previous round stands; what changed is that this PR no longer overstates what it delivers.

### Validation on `2d8f0b0d`

`uv run pytest -m "not e2e"` → **1246 passed**; `ruff check`, `ruff format --check`, `pyright` → clean. `gda schema` is **byte-identical** to `bfd7751a` — verified by regenerating it from a stashed tree and comparing, since docstring edits must not leak into the published surface. This round touched documentation and comments only; no e2e rerun, as no behavior changed.

qinhaihong-red:
Second follow-up recheck pinned to exact head `2d8f0b0d982be74a74955d9474b48434611e6fcc` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

The `source` definition is now coherent across `error_codes.py` and ADR-0002; that residue is closed. Two repository-artifact residues remain.

**[P2] The deferred failure-evidence work still has no authoritative owner.**

PR #678, ADR-0031, and the new #651 comment now honestly mark typed failure-path evidence as partial and say that #655 owns the remainder. However, #655's authoritative body does not own that delivery: it says structured envelope fields require a separate ADR-0004 decision, calls string diagnostics the smallest first step, and has no acceptance criterion for a typed child status or failure-path `ScriptError[]`. Because this PR says `Closes #651`, merging it will close the source requirement while the claimed destination does not track the deferred work.

Smallest closure: amend #655's authoritative scope and acceptance criteria to explicitly own the ADR-0004 decision plus typed failure-path status/diagnostics, or create a dedicated linked issue that does. Then point #651 and ADR-0031 to that actual owner. The reconciliation prose is now honest about the gap, but the handoff is not yet executable.

**[P3] Remove the stale parser-layering statement from the PR body.**

The PR body correctly describes `gda.engine_log` as the extracted neutral leaf, but its final “Layering note” still says `gda.script_errors` imports `gda.daemon.diag.parse_errors` and defers extraction until a third consumer appears. Current source imports `gda.engine_log` directly. These two statements contradict each other in the same PR artifact; remove or update the stale note.

Validation at this head: `tests/test_error_registry.py` passes (`13 passed`); current imports confirm both consumers point downward to `gda.engine_log`; required CI is green. No behavior change was needed for the closed `source` residue.

qinhaihong-red:
Second-recheck response — both residues closed on `b50a69c2`.

| [P2] The deferred failure-evidence work has no authoritative owner: #655's body scopes only the string-diagnostics timeout form, so `Closes #651` would close the source requirement with the destination not tracking the remainder. | Fully adopt (dedicated-issue option) | Confirmed against #655's authoritative body — its ACs cover timeout behavior; the ADR-0004 decision is named but unowned. | Created **#687** ("Typed failure-path evidence on the Error envelope (ADR-0004 decision)", milestone gda dogfooding): owns the ADR-0004 decision, the typed status/`ScriptError[]` delivery for `script run` failure envelopes, and the #655 timeout-envelope alignment, with an explicit decline path. Repointed the ADR-0031 amendment (`#655 owns it` → `#687 owns it`, commit `b50a69c2`), the PR body's AC table and deferral paragraph, and posted a follow-up on issue #651 naming the actual owner. | ADR edit is one clause in this PR's own amendment text (accepted prose untouched); fast suite unaffected (docs-only commit); CI green on `b50a69c2`. |
| [P3] Stale "Layering note" in the PR body still describes `gda.script_errors` importing `gda.daemon.diag.parse_errors` and defers extraction. | Fully adopt | The note contradicted the same document's `gda.engine_log` description; source imports `gda.engine_log` directly since `bfd7751a`. | Removed the stale note from the PR body (the extraction is already documented in the "engine_log" section). | PR body only; no code change. |

PR #680, opened 2026-08-15
fix(gda): restore pre-install project state on daemon uninstall and report the full mutation set

### LINKED ISSUE #654: daemon uninstall: restore pre-install project state and report the full mutation set
## Dogfooding source

- **GDA-DF-009** (2026-08-08) — `daemon uninstall` returns `{"removed":true}` but leaves `addons/gda_harness/gda_harness.gd.uid` and the harness directory.
- **GDA-DF-020** (2026-08-09, 10 recorded recurrences across packages) — uninstall leaves a generated empty `[autoload]` section in `project.godot`, keeping the tracked file modified after every live-QA session.
- **GDA-DF-039** (2026-08-12) — framing: the harness install writes to `project.godot` and `addons/`; unattended automation can accidentally commit or ship the harness, so install/uninstall must be a reviewed, fully reversible mutation with a receipt.

## What to build

- On uninstall: remove the harness script's `.uid` sidecar — it is what currently keeps the addon directory non-empty (empty-directory removal already exists).
- On uninstall: when removing the harness autoload entry leaves the `[autoload]` section empty, remove the section header too. This is decidable at uninstall time; do not record pre-install state anywhere, and never write marker files into the project.
- Report every removed path and section in the structured uninstall result; extend the `daemon start` result (which already reports `installed_harness`/`harness_synced`) to name the exact created paths and sections.

## Out of scope

- A read-only/no-injection mode: ADR-0018 records that the harness must be an installed autoload (Godot exposes no `--autoload` flag), and install is already idempotent (`changed=false` when the harness is present). If that idempotency guarantee needs documentation or a regression test, file it separately.

## Acceptance criteria

- [ ] After start → stop → uninstall on a project with no pre-existing autoload section: no `gda_harness.gd`, no `.uid` sidecar, no `addons/gda_harness/` directory, and `project.godot` is byte-identical to its pre-install content.
- [ ] A pre-existing user autoload section survives install/uninstall unchanged.
- [ ] Install/uninstall preserves the project file's existing line endings, or the byte-identical guarantee is explicitly scoped to LF files.
- [ ] Structured results enumerate the full mutation/removal set.
- [ ] A daemon regression test covers the full round-trip on a tracked project.

## Blocked by

None - can start immediately.

qinhaihong-red:
Closes #654

## Problem

`gda daemon uninstall` returned `{"removed": true}` while leaving residue that kept a tracked project dirty after every live-QA session:

- **GDA-DF-009** — `addons/gda_harness/gda_harness.gd.uid` survived. Because that file kept `addons/gda_harness/` non-empty, the existing empty-directory removal never fired either, so the addon directory survived too.
- **GDA-DF-020** (10 recurrences) — the generated `[autoload]` section survived as an empty header, leaving `project.godot` modified in git after every session.
- **GDA-DF-039** — install writes to `project.godot` and `addons/`, and the results said only *whether* something happened, never *what*, so unattended automation could commit or ship the harness without a reviewable record.

## Change

**Uninstall restores the pre-install state.** It now also removes the engine-generated `.uid` sidecar, and drops the `[autoload]` header (plus the one blank separator line the install appended) when stripping the harness entry leaves the section with no keys. Both decisions are read off the file as it stands **at uninstall time** — no pre-install state is recorded and no marker file is written into the project, per the issue.

**Line endings are preserved.** `project.godot` is read and written with newline translation off and rejoined with the terminator its first line uses. Python's default text mode was silently rewriting a CRLF project file to LF on any autoload edit — a whole-file diff from a one-line change. Three malformed-input shapes still come back changed and are enumerated as exceptions in the `install.py` module docstring, where the guarantee lives: mixed terminators normalize to the first one, a file with no final terminator gains one, and a CR-only file comes back CRLF. All three were verified by probe; none is reachable for a `project.godot` Godot itself wrote, since its `ConfigFile` writer emits uniformly LF-terminated lines and always terminates the last one.

**Both halves issue a mutation receipt.** `install_harness` / `uninstall_harness` return the exact set they touched, surfaced on the CLI results.

**`gda export run` stays byte-identical.** ADR-0028's transactional strip calls `uninstall_harness`, so widening the removal would have made it delete the `.uid` sidecar and never restore it. `_HarnessSnapshot` now reads its file list from the new `harness_artifacts()` in the installer rather than restating the paths, so the strip and the restore cannot drift again.

### Scoped out by design (documented in code)

Two states are not restored, for two different reasons:

- the `[autoload]` section the harness entry sat in is dropped even if it was **already empty** before the install — restoring that one genuinely would need the recorded pre-install state the issue forbids, and Godot's own `ConfigFile` writer never emits an empty section. (An empty `[autoload]` section the harness never joined is left alone: removal is scoped to the section the entry came out of.)
- an `addons/` directory gda created is left in place — but for a different reason: an empty directory is invisible to git, so it causes none of the tracked-file churn this change is about, and `addons/` is the shared Godot-convention directory another addon may be about to populate. Install reports it in `created_paths`, so the one path not reversed is still on the receipt.

## Public-surface diff (intended, complete)

Verified by byte-diffing `gda schema` and `gda daemon {start,stop,status,uninstall} --help` plus `gda --help` before the first edit and after the last.

**`gda schema`** — 4 new optional array properties and 2 description changes; **no** `required` list changed, so every existing consumer stays valid. The new properties are optional by absence from `required`; the emitted schema carries **no `default` key** on them (pydantic `default_factory`), matching the repo's other optional arrays such as `export run`'s `warnings`:

| Command | Change |
| --- | --- |
| `daemon start` | new output property `created_paths: string[]` |
| `daemon start` | new output property `created_sections: string[]` |
| `daemon uninstall` | new output property `removed_paths: string[]` |
| `daemon uninstall` | new output property `removed_sections: string[]` |
| `daemon uninstall` | `output.description` extended (the result model's docstring) |
| `daemon uninstall` | command `description` extended (the Typer docstring, same text as the help diff below) |

No other command, no top-level schema key, and no input model changed.

**`--help`** — `gda daemon uninstall` only. `start`, `stop`, `status` and the root help are byte-identical.

The new prose describes the wider removal and the enumerated result. It also repairs a pre-existing rendering defect in the same paragraph: the old text read *"removal is paired and crash-safe — the  entry is stripped first"* because Typer renders the docstring through Rich, which reads `[autoload]` as a markup tag and drops it. Section names are now spelled without brackets in that docstring (with a code comment recording why); the bracketed spelling survives in the result-model field descriptions, which reach agents as JSON and are never Rich-rendered.

**Human render** — `render_daemon_uninstall` now names the removal set (`harness uninstalled: res://…/gda_harness.gd, res://…/gda_harness.gd.uid, res://addons/gda_harness, [autoload] in project.godot`) instead of a bare `harness uninstalled`. Naming what left the project is the point of this command. `render_daemon_start` is deliberately unchanged: its headline is the daemon coming up, and its `(installed harness)` clause already carries the signal — the paths are in the JSON.

## Validation

All run in the worktree at `e3e77f10`.

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | **1211 passed**, 391 deselected |
| `uv run ruff format --check` | 426 files already formatted |
| `uv run pyright` | 0 errors, 0 warnings, 0 informations |
| `uv run pytest -m e2e -rs tests/test_e2e_daemon.py tests/test_e2e_harness_install.py tests/test_e2e_export_run.py` | **35 passed, 1 skipped** (real Godot 4.6) |
| `gda schema` + all `daemon --help` byte-diff | only the rows enumerated above |

The one e2e skip is pre-existing and unrelated: `test_e2e_export_run.py:355` forces missing templates via `XDG_DATA_HOME`, which only the Linux engine honors as its data dir.

### Making the e2e non-vacuous

The round-trip test first asserted the `.uid` sidecar was gone without proving one had ever existed. Probing a real Godot 4.6 showed a plain game run (`--headless --path <p> --quit`) does **not** write the sidecar — it appears once the project is *scanned* (`--import`, or a human opening the editor on it, which is exactly ADR-0018's concurrent external editor and how the dogfooding project grew one). The test now performs that scan, fails loudly if no sidecar appears, and checks the scan left `project.godot` alone so the closing byte comparison measures the uninstall rather than the engine.

### Tests added

- `tests/test_harness_install.py` — `.uid` removal; empty-section removal; byte-identical round trip with no pre-existing section; a pre-existing user section unchanged; CRLF preservation; a mid-file empty section that must not merge its neighbours; the created/removed receipts; `res://addons` not claimed when the project already had it; the dangling-entry case; `harness_artifacts`.
- `tests/test_daemon_lifecycle_payload.py` — the receipt on the `daemon start` and `daemon uninstall` results, including the empty receipt on an idempotent repeat start.
- `tests/test_export_run_operation.py` — the export strip removes the `.uid` sidecar during the native export and restores it byte-exactly.
- `tests/test_e2e_daemon.py` — the full round trip on a real project (above).
- `tests/test_render.py` — updated for the enumerated human render.

## Acceptance criteria

| Criterion | Status |
| --- | --- |
| start → stop → uninstall leaves no `gda_harness.gd`, no `.uid`, no `addons/gda_harness/`, and `project.godot` byte-identical | Met — e2e `test_daemon_round_trip_restores_the_project_it_started_from`, fast `test_install_then_uninstall_leaves_project_godot_byte_identical` |
| a pre-existing user autoload section survives install/uninstall unchanged | Met — `test_round_trip_leaves_a_pre_existing_user_autoload_section_unchanged` (byte comparison), `test_uninstall_keeps_an_autoload_section_that_still_has_keys` |
| line endings preserved, or the guarantee explicitly scoped | Met by preserving — `test_round_trip_preserves_crlf_line_endings`; the three malformed-input exceptions (mixed terminators, no final terminator, CR-only) are enumerated in the `install.py` module docstring and on `uninstall_harness`, where the guarantee lives |
| structured results enumerate the full mutation/removal set | Met — the 4 new fields above |
| a daemon regression test covers the full round trip on a tracked project | Met — the e2e above, made non-vacuous against a real engine scan |

## Artifacts synced

`docs/adr/0018` gains a dated `> Outcome (2026-08-15, #654)` note: point 3's removal is widened to a full reversal, point 1's "reports the effect" grows from a boolean to an enumeration, and the two deliberately-unrestored states are recorded with their rationale. The decision text is preserved as the point-in-time record.

## Docs landed with the behaviour

- `README.md:482` — the `daemon uninstall` row now names the full teardown (autoload entry, harness files, `.uid` sidecar, an `[autoload]` section left with no keys, `project.godot` back to its pre-install bytes, and the enumerated result). Ownership was checked: slice #672 finished without touching README, so there is no conflict.
- `docs/README.zh-CN.md:479`, `docs/README.es.md:494`, `docs/README.ja.md:498` — the same row in each language, followed by `uv run python scripts/update_readme_i18n.py` to re-stamp the sha256 markers. `tests/test_readme_i18n_sync.py` fails on any `README.md` byte change without the re-stamp; it was confirmed red before the re-stamp and green after.
- `docs/adr/0018` — dated `> Outcome (2026-08-15, #654)` note (see above).
- `docs/adr/0028` — dated `> Outcome (2026-08-15, #654)` note: decision 1's snapshot file list now comes from `harness_artifacts()` (script + `.uid`) rather than the two files its text names. The decision text is preserved as the point-in-time record.

`src/gda/skill/SKILL.md` and `src/gda/error_codes.py` needed no change — no new error code was introduced.

## Review round (`f68959a4`)

Doc-accuracy fixes from the independent review. No behaviour change; one unit test added for a case that was described but never exercised.

- **Byte-identity was under-scoped.** It claimed one exception (mixed terminators). Probing found two more, both now enumerated: a file with no final terminator gains one, and a CR-only file comes back CRLF (`_line_ending` only tells CRLF from LF, while `splitlines` also splits a bare CR). Left documented rather than coded around, since Godot's own writer cannot produce any of the three.
- **The `addons/` rationale was wrong.** It cited the absence of recorded pre-install state, which cannot be the operative reason — the same absence applies to the `[autoload]` section that *is* removed. Corrected in `install.py`, `commands/daemon.py` (two field descriptions) and the ADR-0018 note.
- **"`removed` true with both receipt lists empty" was attributed to the wrong case.** It needs the harness files already gone *and* a sibling autoload holding the section up; the dangling-entry case does report a removed section. Corrected in the dataclass docstring, the renderer comment and the test comment, and `test_uninstall_reports_an_empty_receipt_when_only_the_entry_remained` now covers the genuine case.
- Test-module docstring: "no `addons/` residue" → "no `addons/gda_harness/` residue".

Two review suggestions were declined by the reviewer and are not implemented: softening the `daemon uninstall --help` byte-identity phrasing (the help stays concise; the exceptions live at the guarantee's home), and extending `render_daemon_start` with the created paths (the deliberate asymmetry stands).

**Round-2 gates** — `pytest -m "not e2e"` **1212 passed** (includes the i18n sync test green after the re-stamp), `ruff check` pass, `ruff format --check` 426 formatted, `pyright` 0 errors. No Godot e2e rerun: this round changes no behaviour and the added test is unit-tier.

**Round-2 surface check** — all five `--help` outputs byte-identical to round 1. `gda schema` differs in exactly two `description` strings (`daemon start.created_paths`, `daemon uninstall.removed_paths`), which is the intended effect of the `addons/` rationale fix landing in those field descriptions. Command set, property set, property keys, types and both `required` lists are unchanged.

## Review round 2 — external adversarial review (`8bb56ab5`)

An external review of `f68959a4` raised 12 findings across DDMA / Standards / Spec. All were adopted; two were P1 correctness holes I had not found. The full finding-by-finding reply is posted as a PR comment.

**P1 — `gda export run` could leave the dev project stripped.** The snapshot was captured and the harness stripped *outside* the `try/finally`. The strip is itself a multi-step mutation (entry, script, sidecar, directory), so a failure part way through skipped the restore entirely. Reproduced with an injected sidecar-unlink `OSError`: `project.godot` left stripped and the harness deleted. The strip now runs inside the guarded region. Regression `test_export_restores_the_project_when_the_STRIP_itself_fails` was confirmed red against the old ordering and green against the fix.

**P1 — a failed `daemon start` left the project mutated with no receipt.** The install precedes the daemon, so a start that never became ready returned `daemon_not_running` over a project it had silently written to. Such a start now undoes its install and reports the outcome in the failure's `diagnostics`. Taken as a rollback rather than moving the install to lazy session launch, which is a larger redesign than the defect warrants. *(The rollback shipped here was receipt-driven; the recheck below showed that was insufficient and replaced it with snapshot-exact restoration.)*

**P2/P3 — the artifact list had two owners.** `harness_artifacts()` claimed to be the single enumeration of what the install owns, but `_remove_files()` carried its own script/UID tuple, so the claim only held for the export snapshot. Removal now iterates the authority and derives its receipt from the same entries; a test extends the authority and asserts deletion *and* the receipt follow.

**P2 Spec — empty-section removal was over-broad.** It dropped *every* key-less `[autoload]` section, so a project carrying its own unrelated empty one had it silently deleted — contradicting this change's own "a pre-existing empty section is not gda's to remove" scoping. Reproduced, then scoped to the section the harness entry was actually removed from.

**P2 ×2 — the byte-identity scope was documented three different ways.** `install.py` said three exceptions, the ADR-0018 note said one, the README row said none. `install.py` is now named as the authority; the ADR note carries a one-line summary plus a pointer (so the two cannot drift), and the README row plus all three translations carry a short "(for files Godot's own writer produces)" clause — not the full list, which is not README material. Markers re-stamped via `scripts/update_readme_i18n.py`.

**P2 Spec — LF file missing its final newline: documentation only.** The round-2 adjudication stands — Godot's `ConfigFile` writer always terminates its last line, so the input is unreachable for an engine-written file, and coding around it would mean a per-line terminator model for inputs Godot cannot produce. The claim-4 doc sync *is* the fix: the scope now appears consistently in all three places. **Residual risk:** a hand-edited or tool-mangled `project.godot` lacking a final newline gains one across install→uninstall. Cosmetic, one byte, and visible in review.

**P2 Spec — the e2e never proved cleanliness on a *tracked* project.** The AC says "on a tracked project" and the byte assertions could only see files the test named. The round trip now runs in a real git working tree (`git init`, baseline commit, only `.godot/` ignored) and asserts `git status --porcelain` is empty. Verified live, not vacuous: un-ignoring `.godot/` makes it fail with `?? .godot/`, confirming `.godot/` is the *only* thing needing an ignore — so any harness residue would surface. This also demonstrates the `addons/` rationale directly: the now-empty directory survives on disk and is correctly invisible to git.

**Round-2 gates** — `pytest -m "not e2e"` **1224 passed** (incl. the i18n sync test green after re-stamp), `ruff check` pass, `ruff format --check` clean, `pyright` 0 errors, and real-engine e2e `tests/test_e2e_daemon.py tests/test_e2e_harness_install.py tests/test_e2e_export_run.py` → **35 passed, 1 skipped** (the same pre-existing Linux-only template skip). Most new tests are unit-tier; only the tracked-project round trip needs an engine.

**Round-2 surface check** — `gda schema` and all five `--help` outputs are **byte-identical** to the previous head. The rollback's report is runtime `diagnostics` prose in an existing error code, so ADR-0004's envelope shape and the schema are untouched.

**Coordination note** — this round touches `README.md` and the three translations (plus the sha256 re-stamp). Sibling PR #677 edits a different row in the same files; whichever merges second will need a re-stamp on rebase.

## Recheck round — snapshot-exact restoration (`97ae7ec7`)

The recheck closed six of the seven groups and demonstrated a remaining **P1**, which I reproduced on the exact head before changing anything.

**The residue.** The round-2 rollback was driven by the install *receipt*, and a receipt cannot describe how to undo an install that created nothing. Two paths were affected:

- **Resync.** A project with a correct pre-existing autoload entry and a **stale harness body**: install re-materializes the body, so `changed=True, synced=True, created_paths=()`. The rollback had no prior bytes to put back and did nothing — the failed start silently replaced tracked content and reported an **empty** footprint.
- **Re-point.** A pre-existing entry pointing elsewhere was **removed** by the rollback rather than restored to its original target.

Both are the exact defect class #654 exists to kill (GDA-DF-020/039), so "rolls back exactly" was false as written.

**The fix.** `daemon start` now captures a `HarnessSnapshot` *before* installing and restores it on any post-install failure, composing the two mechanisms this code already owned:

- the **snapshot** returns every captured file to its pre-install bytes — this is what a receipt cannot do, and it covers rewritten bodies and re-pointed entries alike;
- the **receipt** contributed the directories the install created. *(Superseded by the round below: the snapshot now records absent directories itself, and the restore takes nothing from the receipt.)*

Absent artifacts are captured faithfully as `None` and restored by deletion, so the fresh-install case is reversed by the same mechanism as the resync case. The snapshot is held in memory for one operation and never written to disk — #654's "no recorded pre-install state" rule is about persisted markers left in the user's project, and `export run`'s in-process snapshot is the existing precedent.

**Where the snapshot type now lives.** Promoted out of `commands/export.py` into **`src/gda/harness/install.py` as the public `HarnessSnapshot`**, next to the `harness_artifacts()` that supplies its file list; `commands/export.py` imports it from there. That is the direction `commands/` already imports (`commands/export.py` → `harness/install.py`), so no cycle and no new dependency edge. One transactional mechanism now serves both callers instead of a second implementation — the same single-authority point the earlier round raised about `harness_artifacts()`.

`HarnessInstall.wrote_autoload_entry` — added the previous round purely so the receipt could describe its own undo — is **dropped**. Restoring `project.godot` byte-exactly subsumes it.

**Red-then-green on the reviewer's repro (a).** Against the pre-fix head `8bb56ab5`, the stale body is not restored (`harness body byte-identical to pre-start: False`), verified by re-running the scenario with the working tree stashed back to that commit. With the fix, `test_restore_puts_a_stale_harness_body_back_unchanged` and `test_failed_start_restores_a_stale_harness_body_it_resynced` both pass, asserting the stale body comes back **verbatim, still stale**.

**Regressions added** (all unit-tier — none needs an engine):

| Case | Test |
| --- | --- |
| (a) resync + failure → every file byte-identical, stale stays stale | `test_restore_puts_a_stale_harness_body_back_unchanged`, `test_failed_start_restores_a_stale_harness_body_it_resynced` |
| (a) same, seen as a git diff on a tracked fixture | `test_failed_start_leaves_a_tracked_project_porcelain_clean` |
| re-pointed entry restored to its original target | `test_restore_puts_a_re_pointed_autoload_entry_back` |
| (b) fresh install + failure → no residue | `test_restore_undoes_a_fresh_install_completely` |
| (c) `changed=false` → no actions | `test_restore_of_an_unchanged_install_does_nothing` |
| (d) restore failure → residual delta enumerated | `test_failed_start_reports_the_residual_delta_when_restore_also_fails`, `test_snapshot_pending_names_what_a_failed_restore_leaves_behind` |
| created dir that meanwhile gained content is kept | `test_restore_keeps_a_created_dir_that_meanwhile_gained_content` |

**Recheck gates** — `pytest -m "not e2e"` **1229 passed**; `ruff check` / `ruff format --check` / `pyright` clean; real-engine e2e (`test_e2e_daemon.py`, `test_e2e_harness_install.py`, `test_e2e_export_run.py`) **35 passed, 1 skipped** — rerun in full because the snapshot move touches `export run`'s path. `gda schema` and all five `daemon --help` outputs remain **byte-identical**.

**Prose synced** to the now-true statement: ADR-0018's outcome note (restoration is snapshot-driven, and why a receipt is insufficient by construction), ADR-0028's note (the strip moved inside the guarded region; the snapshot type's new home), and this PR body's round-2 section.

## Recheck round 2 — the transaction now opens before the install (`50fad779`)

The second recheck confirmed both prior reproductions were closed and demonstrated one remaining same-class **P1**, which I reproduced on the exact head before changing anything.

**The residue.** `daemon start` captured the snapshot and then called `install_harness()` **outside** the guarded region. The install is itself multi-step — it materializes the harness file first and writes the autoload config **second** — so a config write that fails escapes with the harness already on disk, before any restore can run and with no `HarnessInstall` receipt ever produced.

Reproduced through the real CLI with a real filesystem fault (`chmod 0444 project.godot` on an otherwise writable project), exactly as reported:

```
PermissionError: [Errno 13] Permission denied: '/tmp/reviewrepro/project.godot'
--- residue ---
addons/gda_harness/gda_harness.gd    <-- left behind
addons/                              <-- created by gda, left behind
```

**Transaction shape chosen: bring `install_harness()` inside the guarded region** (option 1 of the two offered). The snapshot is taken first, the install runs inside the `try`, and the failure path restores the already-captured snapshot. I did not make `install_harness()` self-restoring (option 2): it would have to own a snapshot of its own while `daemon start` already holds one, which is precisely the second mechanism the brief said not to add — and it would change `install_harness`'s contract for every other caller, including the already-running branch and the tests.

**Restoration is now receipt-independent**, which is what makes that shape work — a half-finished install has no receipt to consult. `HarnessSnapshot` records absent **directories** alongside absent files, so the restore removes an `addons/gda_harness/` (or `addons/`) that gda created while keeping one the project already had. That subsumes `restore_install()`, which is **dropped** along with its receipt argument, so there is one transaction mechanism rather than two. `HarnessInstall`'s receipt goes back to being purely what the caller *reports*, never an input to an undo.

**The exception still surfaces unchanged** — same `PermissionError`, same traceback, same exit — so the error classification is untouched. Only the residue is gone. Verified against the same real-CLI repro:

```
PermissionError: [Errno 13] Permission denied: '/tmp/reviewrepro2/project.godot'
--- residue check ---
main.tscn  project.godot        <-- nothing else; addons/ never survived
```

**Red-then-green on repro (a).** Against `97ae7ec7`: `no harness residue: False`, `no addons dir: False`. Against `50fad779`: both clean, and the three new regressions pass.

**Regressions added** (unit-tier, driven by a real read-only file rather than a stub):

| Case | Test |
| --- | --- |
| (a) read-only `project.godot` → PermissionError surfaced, no harness/`.uid`/directory residue, project byte-identical | `test_failed_install_restores_a_project_whose_config_cannot_be_written` |
| (a) the same, as a git diff on a tracked fixture | `test_failed_install_leaves_a_tracked_project_porcelain_clean` |
| (b) same fault with a PRE-EXISTING stale harness → stale bytes and sidecar restored verbatim, pre-existing directories kept | `test_failed_install_keeps_a_pre_existing_harness_untouched` |
| the snapshot records which directories the install would create | `test_snapshot_records_the_directories_the_install_would_create` |
| (c) all prior snapshot/rollback tests | unchanged and green |

A note on (b): with a *correct* pre-existing autoload entry the install writes no config at all, so the read-only fault never fires. The test therefore uses a half-installed project (stale harness files, entry absent), which is what makes the config write — and so the fault — actually happen. Worth knowing in its own right: this fault only bites when `project.godot` genuinely has to change.

**Test hygiene:** the read-only fixture restores the original mode in teardown, so pytest's `tmp_path` cleanup cannot fail and mask a real assertion failure.

**Gates** — `pytest -m "not e2e"` **1233 passed**; `ruff check` / `ruff format --check` / `pyright` clean; real-engine e2e (`test_e2e_daemon.py`, `test_e2e_harness_install.py`, `test_e2e_export_run.py`) **35 passed, 1 skipped**. `gda schema` and all five `daemon --help` outputs remain **byte-identical**.

**Prose synced** in ADR-0018's outcome note: the transaction opens *before* the install, and the restore is driven solely by the snapshot (a receipt is insufficient twice over — nothing created on a resync/re-point, and nothing at all on a half-finished install).


qinhaihong-red:
Adversarial review pinned to `f68959a4db4840779b157a0131faecb1dcd88544` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

**[P1] Restore the snapshot when harness stripping itself fails — `src/gda/commands/export.py:608-619`.**

The snapshot is captured, but `uninstall_harness(project)` runs before the `try/finally`. An exact-head injected `.uid` unlink `OSError` escaped with both `project.godot` and the harness script unrestored. Enter the `try` before uninstall, restore in `finally`, and add a partial-strip-failure regression.

**[P1] Do not leave an unreceipted install after daemon readiness failure — `src/gda/commands/daemon.py:478-486`.**

Harness installation precedes spawn/readiness, but mutation fields exist only on `DaemonStartResult`. An exact-head forced readiness failure returned `daemon_not_running` while leaving `project.godot` changed and the harness installed, with no receipt. Install only after readiness (for example at lazy session launch), or perform exact rollback on every post-install failure; if a later install fails, stop the daemon.

**[P2] Make `harness_artifacts()` the actual artifact authority — `src/gda/harness/install.py:214-223,463-466`.**

The helper claims to be the single list, but `_remove_files()` repeats the script/UID tuple. Make removal iterate the authority and derive its receipt from the same entries.

**[P2] Reconcile ADR-0018's byte-identity scope — `docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md:97-99`.**

The ADR calls mixed terminators the one exception, while the implementation's own contract documents three: mixed terminators, no final terminator, and CR-only input. Record the accepted outcome consistently.

## `review`

### Standards

**[P2] ADR-0028 restoration excludes uninstall failures — `src/gda/commands/export.py:608-619`.**

This is the same independently reproduced restore hole described above. Put uninstall inside the guarded region and test partial stripping.

**[P2] Failed start silently leaves the project mutated — `src/gda/commands/daemon.py:478-486`.**

ADR-0018 makes receipts part of both lifecycle halves, but readiness failure returns a plain error after installation. Defer installation or restore an exact pre-start snapshot.

**[P2] Public byte-identity claims omit two known exceptions — ADR-0018:97-99 and `README.md:482`.**

`src/gda/harness/install.py:50-57` documents three changing inputs, the ADR names only mixed terminators, and the README promises unconditional restoration. Synchronize the ADR, README, and translated mirrors.

**[P3] Possible Duplicated Code — the declared artifact authority is not consumed — `src/gda/harness/install.py:463-466`.**

Extending `harness_artifacts()` does not extend uninstall because removal owns a second tuple. Iterate the helper and derive resource-path receipts from it.

### Spec

**[P1] A failed start loses the required mutation receipt — `src/gda/commands/daemon.py:478`.**

Issue #654 requires install/uninstall to be fully reversible and receipted, and structured results to enumerate the mutation set. Roll back on post-install failure or include typed mutation evidence in that failure.

**[P2] Uninstall removes unrelated empty `[autoload]` sections — `src/gda/harness/install.py:388`.**

The issue permits removing the section made empty by removing the harness entry. `_drop_empty_autoload_sections()` instead repeatedly removes every empty `[autoload]` section. An exact-head probe containing one unrelated empty section plus the harness section returned two removals and deleted both. Restrict deletion to the section span from which the harness entry was removed.

**[P2] LF byte identity still fails without a final newline — `src/gda/harness/install.py:303`, `README.md:482`.**

An LF project missing only its final newline gains one after install/uninstall. Preserve the original terminator state, or explicitly scope the public guarantee and issue-compliant tests.

**[P2] The regression does not prove cleanliness on a tracked project — `tests/test_e2e_daemon.py:796`.**

The acceptance criterion asks for the round trip on a tracked project. The test compares selected bytes and paths in a temp directory but never initializes/tracks a repository or asserts clean porcelain. Commit a baseline with generated Godot state ignored, then assert the worktree is clean after uninstall.

Validation: exact base/head and merge-base pinned; `git diff --check` passed. The broadest subpass ran 113 focused tests plus the real-engine round-trip e2e; existing tests passed. I independently reproduced both P1 mutation leaks and the unrelated-section deletion. Full suite, full e2e tier, lint, and type checking were not run.

Summary: DDMA — 4 findings (worst P1); Standards — 4 (worst P2); Spec — 4 (worst P1).

qinhaihong-red:
Thanks — this was a genuinely useful review. Two of the findings were correctness holes I had not found, and both reproduced exactly as described. **All 12 findings adopted**, two of them partially (with the rationale below). Fixed in `8bb56ab5`.

Findings are grouped where they share substance, as they were raised across DDMA / Standards / Spec.

| # | Finding (group) | Decision | Evidence | Resolution | Validation |
| 1 | **P1 — export strip runs outside `try/finally`**, so a mid-strip failure skips the restore (×2: DDMA + Spec) | **Fully adopted** | Reproduced with your injected `.uid` unlink `OSError`: `project.godot` left stripped, harness `.gd` deleted, no restore | Strip moved **inside** the guarded region (`8bb56ab5`, `src/gda/commands/export.py`), so the `finally` covers a partial strip, not just a failed export | `test_export_restores_the_project_when_the_STRIP_itself_fails` — confirmed **red** against the old ordering, green against the fix |
| 2 | **P1 — `daemon start` installs before spawn/readiness**; a readiness failure returns `daemon_not_running` over a mutated project with no receipt (×2: DDMA + Spec) | **Fully adopted**, via receipt-driven rollback | Reproduced by forcing `_await_ready` → `None` | Any post-install start failure now rolls back exactly what the install receipt reports and states the outcome in the failure's `diagnostics`; a rollback failure reports the footprint to remove by hand. `changed=false` → no rollback actions. Exception arm re-raises unchanged | 4 tests in `tests/test_daemon_lifecycle_payload.py`: readiness failure → pre-start bytes; pre-existing install untouched; spawn raises → no residue; rollback fails → footprint in `diagnostics` |
| 3 | **P2/P3 — `harness_artifacts()` is not actually the single authority**; `_remove_files()` owns a second script/UID tuple (×2) | **Fully adopted** | Confirmed by reading: the "extend it once" property held only for the export snapshot | Removal now iterates `harness_artifacts()` and derives its receipt from the same entries (`_res_path`) | `test_uninstall_removal_is_driven_by_the_harness_artifacts_authority` — extends the authority, asserts deletion **and** the receipt follow |
| 4 | **P2 ×2 — byte-identity scope documented three ways**: `install.py` says three exceptions, ADR-0018 says one, README says unconditional | **Fully adopted** | Confirmed by reading all three | `install.py` module docstring named as the **authority**; ADR-0018 note carries a one-line summary + pointer (so they cannot drift); README row + all three translations carry a short `(for files Godot's own writer produces)` clause. Markers re-stamped via `scripts/update_readme_i18n.py` | `tests/test_readme_i18n_sync.py` green after re-stamp |
| 5 | **P2 Spec — `_drop_empty_autoload_sections()` removes EVERY empty `[autoload]` section** | **Fully adopted** | Reproduced with your two-section shape: both sections removed, `removed_sections` reported `('[autoload]', '[autoload]')` | Removal scoped to the section the harness entry was actually removed from (`_remove_autoload` now tracks the touched header indices; `_drop_emptied_autoload_sections`) | `test_uninstall_leaves_an_unrelated_pre_existing_empty_autoload_section` — asserts the exact surviving text and `removed_sections == ("[autoload]",)` |
| 6 | **P2 Spec — LF file missing only its final newline still gains one** | **Partially adopted — documentation only, no code** | Confirmed by probe (along with a third case you did not raise: a CR-only file comes back CRLF) | Both now enumerated in `install.py`'s exception list, and the scope is synced to the ADR + README per #4. **No code**: Godot's `ConfigFile` writer always terminates its last line, so the input is unreachable for an engine-written file; coding around it means a per-line terminator model for inputs Godot cannot produce | Covered by the doc sync in #4. **Residual risk stated plainly:** a hand-edited or tool-mangled `project.godot` without a final newline gains one across install→uninstall — one byte, cosmetic, visible in review |
| 7 | **P2 Spec — the e2e never proves cleanliness on a TRACKED project** | **Fully adopted** | Agreed: byte assertions can only see files the test names | Round trip now runs in a real git working tree — `git init`, baseline commit, only `.godot/` ignored — and asserts `git status --porcelain` is empty | `test_daemon_round_trip_restores_the_project_it_started_from`. **Verified non-vacuous:** un-ignoring `.godot/` makes it fail with `?? .godot/`, proving `.godot/` is the only thing needing an ignore, so any harness residue would surface |

### Two notes on the partial adopts

**#6 (no code).** I want to be explicit that this is a judgement call, not an oversight: the case is real and reproducible, and I am choosing documentation over code because the only producer of `project.godot` that matters cannot generate the input. If you think a hand-edited project file is a realistic enough source to warrant the terminator model, say so and I will reopen it.

**#4 (README).** The README row gets a short parenthetical rather than the exception list — the row is a one-line command summary and the rationale belongs at the guarantee's home. The `(for files Godot's own writer produces)` clause is there to stop the row reading as an unconditional promise.

### On #2, why rollback rather than moving the install

Installing at lazy session launch would remove the window entirely, and it is the more principled shape. I did not take it because it relocates an ADR-0018 decision (the install is a deliberate, reported `daemon start` effect) and touches the daemon's session lifecycle — a redesign disproportionate to this defect. The rollback closes the observable hole and, on failure, tells the user precisely what is left. If the window should be closed structurally, that deserves its own issue and ADR amendment.

### Surface impact

None. `gda schema` and all five `daemon --help` outputs are **byte-identical** to the previous head — the rollback reports through `diagnostics` prose on an existing error code, so ADR-0004's envelope shape is untouched.

### Gates

`pytest -m "not e2e"` **1224 passed**; `ruff check` / `ruff format --check` / `pyright` clean; real-engine e2e (`test_e2e_daemon.py`, `test_e2e_harness_install.py`, `test_e2e_export_run.py`) **35 passed, 1 skipped** — the skip is pre-existing and Linux-only (`XDG_DATA_HOME` template forcing).

qinhaihong-red:
Follow-up recheck pinned to exact head `8bb56ab5d1f93d5decb440d82c24bd4e72e47eb2` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

The export restore hole, clean-project failed-start residue, artifact authority, unrelated-autoload deletion, documentation scope, and tracked-project e2e findings are closed. Focused validation passed (`101` tests plus the real tracked-project daemon round trip). One failed-start integrity residue remains.

**[P1] Failed-start rollback still cannot restore pre-existing harness state — `src/gda/harness/install.py:347-374,571-617`; `src/gda/commands/daemon.py:524-538`.**

The receipt records only paths/sections created by this install. If the harness already exists but is stale, `install_harness()` rewrites it (`changed=True`, `synced=True`) while `created_paths=()`. `rollback_install()` therefore has no prior bytes to restore. It also documents that a pre-existing autoload entry re-pointed by the install is removed rather than restored.

An exact-head reproduction with a pre-existing correct autoload entry plus a stale harness body, followed by forced readiness failure, produced:

```
Failure
project_restored True
harness_restored False
diagnostics ''
```

So “any post-install start failure now rolls back exactly” is still false for a normal version-resync path: failed start silently leaves tracked harness content changed and reports no remaining footprint.

Smallest closure: capture and restore the exact pre-start bytes/config values for every existing artifact or entry that install may rewrite/re-point, or defer those mutations until start succeeds. If safe-direction changes are intentionally retained, enumerate them in failure diagnostics and narrow the rollback contract instead of claiming exact restoration.

qinhaihong-red:
You were right, and the repro was exact. **Fully adopted** in `97ae7ec7` via snapshot-exact restoration.

## The residue, confirmed

Reproduced on `8bb56ab5` before changing anything, both halves of it:

```
install: changed=True synced=True created_paths=() wrote_entry=False
rollback undone: ()
project_restored: True
harness_restored: False        <-- the stale body was silently replaced

re-point restored: False       <-- pre-existing entry removed, not restored
```

The diagnosis holds exactly as you put it: a receipt describes what an install **created**, and an install that re-materializes a stale body or re-points an existing entry creates *nothing*. So the receipt had no prior bytes to put back, the rollback was a no-op, and the failure envelope reported an empty footprint — a failed start changing tracked content with nothing to show for it. "Rolls back exactly" was false as written, on precisely the defect class #654 exists to kill.

## Fix: snapshot + receipt, composed

`daemon start` now captures a `HarnessSnapshot` **before** `install_harness()` and restores it on any post-install failure:

- the **snapshot** returns every captured file to its pre-install bytes — covering the two cases a receipt cannot express;
- the **receipt** contributes the one thing a file snapshot does not model: the directories the install created, removed only when now empty (a directory that meanwhile gained other content is kept).

Absent artifacts are captured faithfully as `None` and restored by deletion, so the fresh-install case is reversed by the same mechanism as the resync case — one path, not two. In-memory for one operation, never persisted; #654's "no recorded pre-install state" rule is about markers left in the user's project, and `export run`'s in-process snapshot is the precedent.

`HarnessInstall.wrote_autoload_entry` — which existed only so the receipt could describe its own undo — is **dropped**, since restoring `project.godot` byte-exactly subsumes it.

## Where the snapshot type lives now

Promoted out of `commands/export.py` into **`src/gda/harness/install.py` as the public `HarnessSnapshot`**, directly beside the `harness_artifacts()` that supplies its file list. `commands/export.py` imports it from there — the direction that module already imports, so no cycle and no new edge. Both transactional callers (the export strip and the failed-start restore) now share one mechanism, which is the same single-authority point you raised earlier about `harness_artifacts()`.

## Red-then-green on repro (a)

| | Result |
| --- | --- |
| Repro (a) against `8bb56ab5` (receipt-only rollback) | **RED** — `harness body byte-identical to pre-start: False` (re-run with the working tree stashed back to that commit) |
| Repro (a) against `97ae7ec7` | **GREEN** — body restored verbatim, **still stale** |

## Regressions

All unit-tier — none needs an engine.

| Case | Test |
| --- | --- |
| (a) resync + forced readiness failure → every file byte-identical, stale stays stale | `test_restore_puts_a_stale_harness_body_back_unchanged`, `test_failed_start_restores_a_stale_harness_body_it_resynced` |
| (a) the same, as a git diff on a **tracked** fixture → porcelain-clean | `test_failed_start_leaves_a_tracked_project_porcelain_clean` |
| re-pointed entry restored to its original target | `test_restore_puts_a_re_pointed_autoload_entry_back` |
| (b) fresh install + failure → no residue | `test_restore_undoes_a_fresh_install_completely` |
| (c) `changed=false` → no actions | `test_restore_of_an_unchanged_install_does_nothing` |
| (d) restore failure → residual delta enumerated | `test_failed_start_reports_the_residual_delta_when_restore_also_fails`, `test_snapshot_pending_names_what_a_failed_restore_leaves_behind` |
| created dir that meanwhile gained content is kept | `test_restore_keeps_a_created_dir_that_meanwhile_gained_content` |

One refinement worth flagging: on a restore failure the diagnostics now report the **measured** residual delta (`HarnessSnapshot.pending()` — which captured files still differ from their prior bytes) rather than a footprint predicted from the receipt. Same reason as the main fix: the receipt cannot see a rewrite.

## Validation

`pytest -m "not e2e"` **1229 passed**; `ruff check` / `ruff format --check` / `pyright` clean; real-engine e2e (`test_e2e_daemon.py`, `test_e2e_harness_install.py`, `test_e2e_export_run.py`) **35 passed, 1 skipped** — rerun in full because the snapshot move touches `export run`'s path; the skip is the pre-existing Linux-only template one. `gda schema` and all five `daemon --help` outputs remain **byte-identical**.

Prose synced to the now-true statement in ADR-0018's outcome note (restoration is snapshot-driven, with why a receipt is insufficient by construction), ADR-0028's note (the strip's move inside the guarded region, and the snapshot type's new home), and the PR body.

qinhaihong-red:
Second follow-up recheck pinned to exact head `97ae7ec73ff20cb1955446db31dc841e99bafebd` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

The stale-harness resync and re-point reproductions are closed: the previous exact repro now reports both `project_restored True` and `harness_restored True`. One same-class P1 transaction hole remains.

**[P1] A partial `install_harness()` failure still escapes before snapshot restoration — `src/gda/commands/daemon.py:538-545`; `src/gda/harness/install.py:441-447`.**

`daemon start` captures the snapshot, then calls `install_harness()` *before* entering the `try`. The install materializes the harness before it writes the autoload edit. If that config write fails, no `HarnessInstall` receipt is returned and the restore path never runs.

I reproduced this with a real filesystem failure, not a mocked rollback: make `project.godot` read-only (`0444`) and start against an otherwise writable project. `_materialize()` creates the harness, `_write_config()` raises `PermissionError`, and the operation leaves:

```
PermissionError [Errno 13] Permission denied: '.../project.godot'
project_restored True
harness_residue True
addons_residue True
```

So the normal readiness-failure path is snapshot-exact now, but the install transaction itself can still fail after mutation and leave residue without restoration or a structured mutation footprint—the same project-integrity class as the prior P1.

Smallest closure: make `install_harness()` exception-safe, or include the install inside a transaction that can restore both captured file bytes and pre-install directory existence even when no `HarnessInstall` result was produced. Add a regression that fails the config write after harness materialization and asserts no harness file or created directory remains.

Validation at this head: the two focused files pass (`78 passed`); the old stale-body repro passes; the read-only-file probe above independently reproduces the remaining path. Required CI is green.

qinhaihong-red:
Confirmed and **fully adopted** in `50fad779`. Reproduced through the real CLI with your exact fault before changing anything.

## The residue

`daemon start` captured the snapshot and then called `install_harness()` **outside** the guarded region. Your diagnosis is exactly right: the install is itself multi-step — it materializes the harness file first and writes the autoload config **second** — so a config write that fails escapes with the harness already on disk, before any restore can run and with no `HarnessInstall` receipt ever produced.

`chmod 0444 project.godot`, then `gda daemon start`, on `97ae7ec7`:

```
PermissionError: [Errno 13] Permission denied: '/tmp/reviewrepro/project.godot'
--- residue ---
addons/gda_harness/gda_harness.gd    <-- left behind
addons/                              <-- created by gda, left behind
```

## Transaction shape chosen

**Option 1 — `install_harness()` moves inside the guarded region.** The snapshot is taken first, the install runs inside the `try`, and the failure path restores the already-captured snapshot.

I did not take option 2 (making `install_harness()` self-restoring). It would need a snapshot of its own while `daemon start` already holds one — the second mechanism the brief said not to add — and it would change `install_harness`'s contract for every other caller, including the already-running self-sync branch.

**Restoration is now receipt-independent**, which is what makes option 1 work: a half-finished install has no receipt to consult. `HarnessSnapshot` now records absent **directories** alongside absent files, so the restore removes an `addons/gda_harness/` (or `addons/`) that gda created while keeping one the project already had — verified in both directions. That subsumes `restore_install()`, which is **dropped** along with its receipt argument, so there is one transaction mechanism instead of two. `HarnessInstall`'s receipt goes back to being purely what the caller *reports*, never an input to an undo.

**The exception surfaces unchanged** — same `PermissionError`, same traceback, same exit — so the classification is untouched. On `50fad779`, same repro:

```
PermissionError: [Errno 13] Permission denied: '/tmp/reviewrepro2/project.godot'
--- residue check ---
main.tscn  project.godot        <-- nothing else; addons/ never survived
```

## Red / green on repro (a)

| | Result |
| --- | --- |
| Repro (a) against `97ae7ec7` | **RED** — `no harness residue: False`, `no addons dir: False` |
| Repro (a) against `50fad779` | **GREEN** — no harness, no `.uid`, no created directory, `project.godot` byte-identical, porcelain-clean |

## Regressions

Driven by a real read-only file, not a stub.

| Case | Test |
| --- | --- |
| (a) read-only `project.godot` → PermissionError surfaced, no residue, project byte-identical | `test_failed_install_restores_a_project_whose_config_cannot_be_written` |
| (a) the same, as a git diff on a tracked fixture | `test_failed_install_leaves_a_tracked_project_porcelain_clean` |
| (b) same fault with a PRE-EXISTING stale harness → stale bytes and sidecar restored verbatim, pre-existing directories kept, nothing deleted | `test_failed_install_keeps_a_pre_existing_harness_untouched` |
| the snapshot records which directories the install would create | `test_snapshot_records_the_directories_the_install_would_create` |
| (c) prior snapshot/rollback tests (readiness-failure path) | unchanged and green |

One finding while writing (b), worth recording: with a *correct* pre-existing autoload entry the install writes **no config at all**, so the read-only fault never fires and the start proceeds to the readiness path. The test therefore uses a half-installed project (stale harness files, entry absent) to make the config write — and so the fault — actually happen. The fault only bites when `project.godot` genuinely has to change.

Test hygiene as you asked: the read-only fixture restores the original mode in teardown, so `tmp_path` cleanup cannot fail and mask a real assertion failure.

## Validation

`pytest -m "not e2e"` **1233 passed**; `ruff check` / `ruff format --check` / `pyright` clean; real-engine e2e (`test_e2e_daemon.py`, `test_e2e_harness_install.py`, `test_e2e_export_run.py`) **35 passed, 1 skipped** — the skip is the pre-existing Linux-only template one. `gda schema` and all five `daemon --help` outputs remain **byte-identical**.

ADR-0018's outcome note is synced: the transaction opens *before* the install, and the restore is driven solely by the snapshot — a receipt is insufficient twice over, creating nothing on a resync/re-point and not existing at all on a half-finished install.

qinhaihong-red:
Third follow-up recheck pinned to exact head `50fad779c6450b1d0bd90a2a6e7e828414735998` against merge-base `d0de6bd22d912697eb8b01798ade90dd235adf32`.

The previous P1 is closed: both the real read-only `project.godot` repro and the stale-harness repro restore cleanly, and the focused files pass (`82 passed`). One P2 recovery-reporting residue remains.

**[P2] `HarnessSnapshot.pending()` omits directory residue — `src/gda/harness/install.py:346-357`; `src/gda/commands/daemon.py:411-418`.**

The snapshot now owns absent directories as part of exact restoration, but `pending()` still compares only `self.files`. I forced the innermost directory removal to raise after the captured files had already been restored. The exact head returned:

```
Failure
harness_dir_residue True
diagnostics 'the harness install could NOT be rolled back (injected directory removal failure); these files still differ from their pre-start state: '
```

A real snapshot-owned residual remains, but the measured residual list is empty. This breaks the failure path’s promise to tell the user what still needs repair.

Smallest closure: include every still-present `absent_directories` entry in `pending()`, change the diagnostic wording from “files” to “paths” or “artifacts,” and add a regression that injects `rmdir` failure after file restoration and asserts the remaining harness directory is named.

Validation: the old read-only-install and stale-resync repros pass; 82 focused tests pass; this directory-only restore-failure probe independently reproduces the residue; required CI is green.

qinhaihong-red:
Third-recheck response — the recovery-reporting residue is closed on `1125bbce`.

| [P2] `HarnessSnapshot.pending()` compares only `self.files`, so a directory-only restore failure (rmdir fails after the files are already back) reports an empty residual list while `addons/gda_harness/` remains. | Fully adopt | Reproduced via the injected-rmdir path: files restored, directory removal raises, old `pending()` returned `()`. | `pending()` now appends every captured-absent directory that still exists, after the byte-differing files; the failure prose says "these paths" instead of "these files" (`src/gda/harness/install.py`, `src/gda/commands/daemon.py`). New regression `test_failed_start_names_directory_residue_when_only_the_rmdir_fails` injects the rmdir failure after file restoration and asserts the leftover `res://addons/gda_harness` is named while the already-restored file is not. The pre-existing `test_snapshot_pending_names_what_a_failed_restore_leaves_behind` expectation widened to the new files-then-directories shape. | `1125bbce`. Red/green: the new regression FAILS against `50fad779`'s `pending()` (stash probe) and passes with the fix. Full fast suite 1234 passed; ruff check/format clean; pyright 0 errors. `gda schema` byte-identical before/after (stash-compare, sha `1a88eae0…`) — runtime prose only, no surface change. |

qinhaihong-red:
Recheck response — the ADR-0018 wording residue is closed on `3d680b5f`: the outcome note's restore-failure clause now reads "which paths (files and created directories) still differ", matching `pending()`'s behavior since `1125bbce`. One-line doc sync; no code or surface change.

Issue #679, opened 2026-08-15
[gda-balancing] Add a local HTTP execution service


qinhaihong-red:
## Background

Local interactive clients need to admit and execute new Experiment Specifications without spawning
a new CLI process, publishing a temporary filesystem artifact set, or inventing a client-specific
case format for every interaction. The existing CLI remains the human and automation interface, but
it is not a long-lived client coordination protocol.

The identified consumers all run locally. They can use different local UI technologies, including
a browser-based UI. This issue does not design a remotely published or multi-tenant service.

This issue owns only the engine- and application-agnostic gda-balancing service. Consumer UI,
content projection, executable discovery, source-file location, and product distribution belong to
their consumer issues. In particular, #646 is a downstream consumer; this service does not depend
on Godot or Reward Run.

## Goal

Add a loopback-only HTTP execution service to the existing gda-balancing wheel. The service must
reuse the current Application and Domain behavior and preserve existing Kernel, LDB, Model,
Experiment, Runtime, artifact, identity, outcome, and refusal ownership.

Keep the application-agnostic Execution HTTP API separate from local process control. The API owns
execution requests and outcomes. The local companion host owns the in-process server lifecycle,
loopback binding, its process capability, readiness state, and shutdown. The owning client remains
responsible for executable discovery, child-process launch, and forced termination after its own
timeout.

The first delivery supports complete Experiment runs. That boundary is deliberately changeable:
the first concrete client need for Runtime steps, declared external input, active-run observation,
or another interaction must reopen the protocol immediately and add the smallest required
authority-preserving capability. Do not wait for a second consumer, but do not prebuild unused
interaction machinery.

## Aligned technical design

- Add the foreground operational command
  `gda-balancing serve --host 127.0.0.1 --port 0`. It emits one JSON readiness record to stdout;
  logs use stderr. It is not a new Runtime or Experiment command group.
- Keep `serve` as one Command descriptor in the existing registry. Add only a
  `foreground-service` execution marking and a typed readiness result to select an Interface-owned
  foreground runner. The descriptor remains the source for input, help, `--schema`, Surface
  manifest, readiness, usage/internal errors, and conformance fixtures; do not add a second command
  list, schema path, or dispatch framework.
- Flush exactly one readiness document after the local host accepts requests, then write no more
  stdout content. Normal authenticated shutdown exits with code 0 and no terminal result. Before
  readiness, argument validation retains the existing usage contract; a socket-binding or other
[...]

Issue #646, opened 2026-08-14
[gda-balancing] Deliver an extensible playable HITL reward/build MVP


qinhaihong-red:
## Parent

- #585

## Dependency

- Blocked by #679, which owns the generic gda-balancing HTTP execution service.

## Background

Reward Run already proves a player-facing Godot shell, blockout gameplay, Tween feedback,
mouse/keyboard input, localization, display preferences, feedback capture, and the Content
composition seam. Its current reward outcomes are generated before play through
'playtest/tools/generate_reward_cases.py'.

That generated path fixes every trial in advance. A running playtest cannot admit a changed
Experiment, so a player cannot change reward tuning and immediately perceive the next result. The
generator also creates a playtest-specific case schema and duplicate data pipeline that would have
to be repeated for Combat, Periodic Effect, and later playable products.

#679 now owns the generic technical service. This issue owns only the local Reward Run product and
its reusable Godot client boundary. It must not redefine the HTTP protocol or gda-balancing
semantics.

## Goal

Turn Reward Run into a local, live HITL product. A player launches the playable through one local
entry point, changes a player-facing reward control, completes a short trial, perceives the selected
reward and build replacement, and records feedback. The playable hides Model, Experiment, Formula,
Runtime, trace, Metric, typed-value, artifact, identity, and HTTP details.

The playtest is a repository-local verification product, not a separately distributed game. This
issue does not package Python, freeze gda-balancing, publish a companion executable, or require a
standalone exported product.

## Godot architecture

Keep one-way dependencies:

    UI -> Reward Content -> Systems -> Add-ons -> Godot

- Add 'addons/gda_balancing_client/' as a reusable technical library. A thin bootstrap creates one
  'GdaExecutionClient' and injects it into Reward Content. It is not an Autoload, service locator,
  event bus, plugin registry, 'EditorPlugin', or required 'plugin.cfg'.
- The Add-on owns executable discovery, child-process lifetime, readiness, the process credential,
  HTTP, protocol compatibility, session/revision handles, and transport/service errors.
- Its public operations are limited to start, session creation, revision creation, exact-revision
  run, session deletion, and shutdown. It contains no Reward, Combat, or Effect concepts.
- Reward Content owns maintained source locations, Experiment editing, artifact-to-feature
  projection, application flow, and player-facing failure/retry behavior.
- Systems receive only reward/build gameplay values and own reusable state transitions. UI calls
  Content and observes player-facing signals; it does not parse service or gda-balancing data.
[...]

PR #682, opened 2026-08-15
feat(gda-balancing): add local HTTP execution service

qinhaihong-red:
## Summary

- add `gda-balancing serve` as a descriptor-owned foreground command in the existing CLI surface;
- add a loopback-only, capability-protected `/v1` HTTP service to the existing wheel;
- admit complete Model Source and Experiment values into process-local Execution sessions;
- support immutable Experiment revisions, exact revision selection, synchronous runs, deletion, and
  shutdown without adding a second Domain or artifact authority;
- reuse one publication-independent Experiment execution use case from both CLI and HTTP adapters;
- return the existing success, verdict, and Runtime-refusal artifact sets inline; and
- accept and reconcile bADR-0026 and the affected architecture, glossary, and CLI decisions.

## Scope

The service is engine- and application-agnostic. It contains no Godot, game, playtest, browser,
remote-hosting, persistence, worker, task, or partial-update behavior. PR #646 is a downstream local
consumer and does not define this protocol.

Complete Experiment runs and fixed-Model Execution sessions are initial delivery boundaries, not
permanent restrictions. A concrete client need can extend the protocol with the smallest
authority-preserving capability without waiting for a second consumer.

The Execution HTTP API contains exactly five routes:

- `GET /v1/status`;
- `POST /v1/execution-sessions`;
- `POST /v1/execution-sessions/{session_id}/experiment-revisions`;
- `POST /v1/execution-sessions/{session_id}/runs`; and
- `DELETE /v1/execution-sessions/{session_id}`.

The local companion host separately owns authenticated `POST /v1/shutdown`, loopback binding,
readiness, the process capability, graceful shutdown, and post-readiness fault termination.

## Validation

- HTTP service suite: `33 passed`;
- complete Experiment shard: `184 passed`;
- complete Composition shard: `427 passed, 29 skipped`;
- layer and isolation checks: `38 passed`;
- inventory closure: 1,582 tests, no missing, overlapping, uncovered, or unexpected-shard tests;
- Ruff and Pyright on the final changed implementation and test files;
- source CLI and HTTP semantic-artifact parity;
- built-wheel service startup, packaged-authority run, and shutdown;
- temporary uncommitted Godot 4.6 `HTTPRequest` consumer: session creation, complete Experiment run,
  seven inline artifacts, and shutdown passed; and
- full required CI matrix passed at the previous implementation head; the review-only regression
  commit has started a fresh matrix.

Tracks #679.


qinhaihong-red:
## Review summary (base 6a9375a -> head 8019ed3)

Reviewed with the design-domain-modular-architecture method and a two-axis (Standards/Spec)
review. Overall: the new inbound Interface fits the bADR-0025 layer model and the bADR-0026
ownership boundaries; CLI and HTTP share the same Application/Domain execution path; and no new
authority, identity family, Repository, or speculative mechanism is introduced. One
architecture-level behavior needs a focused runtime probe before merge; the remaining items are
minor DRY observations.

### Findings

1. **Post-readiness route-level faults likely do not terminate the process (open evidence, needs
   a runtime probe).** bADR-0026 requires that after readiness, a later internal fault writes one
   sanitized `internal_error` envelope as its final log record and exits with code 4 while the
   readiness document remains on stdout. In the production composition, both `create_api_v1` and
   `_local_companion_app` register an `Exception -> unexpected_internal_error` handler. Starlette's
   exception middleware catches route-handler exceptions and returns the sanitized 500 response, so
   the exception does not propagate to `FatalApplicationFaultMiddleware`; the process keeps serving
   instead of exiting 4, and no final stderr record is written. `test_post_readiness_application_fault_stops_the_local_host`
   injects a bare ASGI app without those handlers, so it proves the middleware path but not the
   production composition.
   Validation needed: inject a fault inside a real route after readiness and assert exit 4,
   readiness retained on stdout, and the final `internal_error` envelope on stderr. Depending on the
   result, either let route faults reach the fault middleware (or report the fault from the handler)
   or narrow the bADR-0026 wording to "sanitized 500 and continue".

2. **Duplicated 64 KiB envelope constant (minor, DRY).** `SMALL_HTTP_REQUEST_BYTES`
   (service_errors.py) and `_PROTOCOL_ENVELOPE_BYTES` (api_v1.py) are two independently editable
   copies of 65_536 with different meanings; share one constant or document why they are independent.

3. **Duplicated exception-handler registration (minor).** `api_v1` and `_local_companion_app`
   register the same handler set; the envelope authority already lives in service_errors, so this is
   a judgement call.

4. **`run_serve` ignores the dispatch-provided stderr channel (minor).** The descriptor contract
   says the foreground runner receives stderr for operational logs, but uvicorn writes to process
   stderr directly. This is the same stream in production, but the declared channel contract is not
   exercised.

5. **Repeated three-way outcome dispatch (judgement).** Success/verdict/refusal `isinstance`
   cascades appear in experiment_execution, api_v1, and experiment_run; acceptable for the small
   closed set and consistent with the existing typed-union style.

### Verified confirmations

- One descriptor for `serve`; `--schema`, readiness, help, manifest projection, and conformance
  fixtures derive from it; no parallel command or schema path.
- Session/revision semantics match bADR-0026: fixed Resolved Model, immutable revisions keyed by
  content identity, idempotent admission, explicit revision selection, no implicit active revision.
- CLI/HTTP parity: identical artifacts and content identities; no Artifact-set receipt synthesized;
  the Build receipt stays publication-independent.
- Dependency gates: `api_v1` does not import `local_host`; Application/Domain do not import
  Starlette or Uvicorn; no upward imports; the service imports no game or engine module.
- Closed protocol: Pydantic `extra="forbid"`, bounded reads before parsing, complete service-error
  mapping, per-session ordering with process-wide serial execution.
- No scope creep: no persistence, Repository, task system, WebSocket, quota framework, or
  standalone packaging.

Standards axis: 0 hard violations, 3 judgement calls. Spec axis: 1 material finding (item 1) and 1
minor observation (item 4). CI was green at the pinned head; item 1 is not covered by the current
suite.

qinhaihong-red:
Handled at exact head `b3493f2e`.

| Claim | Decision | Evidence and handling |
| --- | --- | --- |
| A production route fault is converted to HTTP 500 and cannot stop the process. | Partially adopt | The behavior claim does not reproduce. Starlette 1.6.0 places the general `Exception` handler in `ServerErrorMiddleware`; it sends the sanitized 500 response and then re-raises the exception. A real `create_api_v1` route fault reaches `FatalApplicationFaultMiddleware` and stops the host. The original test did not cover that production composition, so `b3493f2e` adds the missing regression. Together with the descriptor conformance test, this proves HTTP 500, host termination, exit 4, retained readiness stdout, and the final sanitized stderr envelope. No production or bADR change was needed. |
| The two 64 KiB constants should be shared. | Do not adopt | They own different policies: `SMALL_HTTP_REQUEST_BYTES` is the complete bound for small request bodies; `_PROTOCOL_ENVELOPE_BYTES` is additional framing allowance for large source-bearing requests. bADR-0026 already distinguishes them. Sharing one constant would couple independent limits. |
| The two exception-handler maps should be deduplicated. | Do not adopt | They belong to separate Starlette boundaries and are not identical: the Execution API owns media-type mapping, while the companion host owns shutdown and mount behavior. A shared registration helper would hide this boundary without removing meaningful complexity. |
| `run_serve` must route Uvicorn logging through the injected `stderr`. | Do not adopt | The runner currently emits no application-owned operational logs. In the public CLI, Uvicorn already writes to process stderr, while post-readiness fatal output is serialized by dispatch through the injected stream. Custom Uvicorn handler plumbing would serve only an internal alternate-stream scenario with no current consumer or failure evidence. |
| The three outcome dispatches should be unified. | Do not adopt | They are separate boundary projections: Application constructs the execution outcome, CLI selects publication behavior, and HTTP selects the response model. Keeping the closed three-way branches explicit preserves those ownership boundaries. |

Validation for the review change:

- `tests/test_http_service.py`: 33 passed;
- foreground CLI fault contract: 1 passed;
- Pyright: 0 errors;
- Ruff check and format: passed;
- inventory closure: 1,582 tests, with no missing, overlapping, uncovered, or unexpected-shard tests;
- `git diff --check`: passed.

The required CI matrix has restarted for `b3493f2e`.

PR #688, opened 2026-08-16
Promote the gda dogfooding wave 1 slices to main

qinhaihong-red:
## Wave 1 promotion — gda dogfooding milestone

Promotes the wave-1 slices from `feat/gda-dogfooding-dev` to `main` (branch model: slice PRs
merge into the dev branch; `main` receives the work through this promotion PR). Tracks the
review-and-merge process for the whole wave.

**Merge method note:** merge commit recommended (NOT squash) — the dev branch carries one clean
squashed commit per issue, and squashing the promotion would collapse them into one commit,
breaking release-please's changelog/version derivation on `main`.

### Merge order and status

| # | Slice PR | Issue | Status |
| --- | --- | --- | --- |
| 1 | #676 — docs(gda): document the Control-root zero-size layout in scene create help and the skill | #672 | ✅ merged `f78fb310` |
| 2 | #680 — fix(gda): restore pre-install project state on daemon uninstall and report the full mutation set | #654 | ✅ merged `1b98fcdc` |
| 3 | #681 — fix(gda): keep the harness serving live operations while the SceneTree is paused | #656 | ✅ merged `910cf474` |
| 4 | #677 — feat(gda): accept --json on the root and schema discovery surfaces | #671 | ✅ merged `ddd423c8` (rebased onto the dev branch + i18n re-stamped after #680, gates re-run green on the rebased head) |
| 5 | #678 — fix(gda): make gda the verdict authority for script run (ADR-0031 amendment) | #651 | ✅ merged `37dfa709` |

### Review trail

Every slice went through: implementation → independent adversarial review → adjudicated fix
rounds → external (maintainer-channel) adversarial review with per-claim reply tables on each
PR → recheck rounds until closed. See each PR's comment thread for the full claim → decision →
evidence → commit mapping.

### Wave validation

- Per-PR CI green on every final head (unit/lint/type/build/co-install/scope guards).
- Targeted real-engine e2e run locally per slice, serialized across worktrees.
- Full-e2e CI: dispatched on the dev-branch ref after the last merge — run
  [31926275507](https://github.com/aigengame/godot-agent/actions/runs/31926275507) (`run-e2e: true`) — **completed: success** (full suite incl. the real-Godot e2e tier, on the integrated dev branch at `37dfa709`).

### Follow-ups created during the wave (tracked, not blocking)

#683 (group-level `--json` parser site), #684 (`SceneTree.suspended` limit), #685 (human-mode
[...]

PR #692, opened 2026-08-16
feat(gda): report structured install provenance from --version --json

### LINKED ISSUE #659: Structured version and installation-provenance self-description

qinhaihong-red:
Closes #659

> **Base note (resolved).** GitHub auto-deleted `feat/gda-dogfooding-dev` when promotion
> PR #688 merged; this PR briefly targeted `main` as the content-equivalent fallback. The
> orchestrator recreated the integration branch at `main`'s tip (`c11f07db`) and this PR
> targets `feat/gda-dogfooding-dev` again, per the wave branch model.

Dogfooding GDA-DF-018 and GDA-DF-043: `gda --version --json` was rejected, so an evidence
collector had to parse one human line; and the `PATH`-resolved `gda` turned out to be an
editable install whose source checkout changed revision mid-run, with `gda --version`
disclosing none of it.

## What landed

`gda --version --json` now emits the install-provenance payload. Both argv orders work;
bare `gda --version` is unchanged human text.

```json
{
  "gda_version": "0.9.0",
  "executable": "/…/w2-659/.venv/bin/gda",
  "interpreter": "/…/w2-659/.venv/bin/python",
  "package_path": "/…/w2-659/src/gda",
  "install_kind": "editable",
  "source": { "root": "/…/w2-659", "revision": "5a61a813…", "dirty": false },
  "godot": {
    "binary": "/Users/…/Godot.app/Contents/MacOS/Godot",
    "version_unavailable_reason": "the engine reports its version only from a running process, and this surface never launches Godot; run `gda info --json` for it"
  }
}
```

### Design decisions (and their rationale)

**`--version --json` only — no new Meta command.** The ACs ask for a *structured
invocation*, and `--version --json` is the exact spelling the report used; a second named
command would be a second authority for one payload, would enter `gda schema` as a tool,
and would have to negotiate a name with #670's `doctor` discussion. Instead the payload
lives in a CLI-free `gda.provenance` module with one entry point,
`build_version_provenance()`, so a later `gda doctor` (#670) can embed it verbatim. **This
slice deliberately squats no name #670 may want.** `gda schema` is byte-identical, so
gda-mcp's generated tool surface is untouched.

**`--version` is no longer eager.** Click orders parameter processing by
`(not is_eager, argv index)`, so an eager `--version` runs before a later `--json` and can
never see it — the constraint recorded in #671's comment and its
`test_root_version_stays_text_in_both_json_orders`. Non-eager, `--version` sorts after
*every* eager parameter, so the eager root `--json` is bound first in both orders. The flag
travels through #671's existing `ctx.meta` contract; this PR adds only the read half
(`gda.headless.root_json`) beside `set_root_json`, and moves the write into the `--json`
option's own callback so it lands before any other root option is processed.

*Behavior trade, stated up front:* click's own eager `--help` now wins over `--version`
when both are given (`gda --version --help` prints help, not the version). That is the
conventional precedence; it is pinned by `test_help_wins_over_version_when_both_are_given`.

**Install kind from PEP 610 — three values, because two would lie.** `dir_info.editable == true`
→ `editable`; `archive_info` or a non-editable `dir_info` → `wheel` (a built copy, so no
checkout to report). The version string cannot distinguish them and `pip show` is unavailable
in a uv environment.

The load-bearing part is the READ seam, which is three-state (`RecordState`), because
collapsing any two arms is how a damaged *editable* install gets reported as an immutable
wheel — the exact false provenance this surface exists to prevent:

| Read arm | Observation | Verdict |
| --- | --- | --- |
| `ABSENT` | the reader says "no such file" | **`wheel`** — the ONLY observation that earns it, and the shape of an ordinary index install |
| `PRESENT` | any content, whitespace included, handed over verbatim | `editable` / `wheel` / `unknown`, per `classify_install` |
| `UNREADABLE` | the reader raised, or the distribution is not installed | **`unknown`** — gda failed to look; it did not learn that nothing is there |

On the `PRESENT` arm, `unknown` covers unparseable JSON (a whitespace-only document included),
a non-object document, a non-object `dir_info`, and an `editable` that is not the boolean PEP
610 types it as — so `"false"` and `1` are off-spec, not truthy. `unknown` never claims a
checkout even when the damaged record names a good one, and no arm fails the preflight: the
payload is still emitted with everything it does know. `classify_install` stays pure, so every
branch is testable without an installer.

*Formerly a known hole, closed on the second recheck (`ffa105be`):* `Distribution.read_text()`
suppresses `PermissionError`, `IsADirectoryError` and friends into the same `None` it returns
for a missing record. The seam now reads the record from the distribution's own metadata
directory, where `FileNotFoundError` alone is positive missing-file evidence (`absent` →
`wheel`); every other read failure — a mode-000 record, a directory sitting in the record's
place — is `unreadable` → `unknown`. An exotic `Distribution` without a metadata path degrades
its ambiguous `None` to `unreadable`, never to the confident-wheel arm. The seam fixture now
builds REAL dist-info directories, so all arms are produced by the real reader.

**`package_path` makes `install_kind` falsifiable.** Every other field is read from
distribution metadata or `sys.argv[0]` — i.e. from what an *installer recorded*, not from
what Python *loaded*. Under a `sys.path` shadow a wheel install truthfully reports `wheel`
while a mutable source tree is what actually runs, which would defeat the stated purpose
("ties your results to the code that produced them"). `package_path` reports the directory
the `gda` package was imported from, read from the loaded module, so the divergence is
visible instead of silent.

**Git is read from `source_root` and nowhere else.** `$GIT_DIR` **wins over `git -C`**, and
gda is invoked from exactly the places that set it — a git hook, `git rebase --exec`,
`git bisect run`, a CI wrapper. Unscrubbed, the payload paired this checkout's `root` with
another repository's `revision`. The provenance git calls drop `GIT_DIR`, `GIT_WORK_TREE`,
`GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY` and `GIT_COMMON_DIR`, and set
`GIT_OPTIONAL_LOCKS=0` so a read-only preflight never contends for the agent's own
`index.lock`.

**Untracked source counts as dirty, whatever the git config says.** `git status` is called
with `--untracked-files=all`, not left to the repository's `status.showUntrackedFiles`:
with that set to `no`, an untracked module sitting in an editable checkout — importable, so
code that can run — reported `dirty: false`, contradicting this module's stated contract.
The verdict is a property of the tree, not of someone's config.

**The engine version key is omitted, never null.** #659's contract is "only when obtainable,
otherwise omitted with a stated reason". A `null` would claim gda looked and found nothing,
when it declined to look; so `GodotProvenance` drops whichever of
`version`/`version_unavailable_reason` does not apply and exactly one is ever present —
gda's omitted-never-null convention. `SourceCheckout` deliberately keeps ITS nulls: there
the null is the finding (gda looked at a named checkout and could not resolve its state).

**No engine launch.** The motivating environment is the restricted profile where an engine
spawn crashes, so the Godot *version* — obtainable only from a running engine — stays
`null` with a stated reason, while the *binary* gda would use is reported spawn-free.
`binary` resolves through `$GDA_GODOT` → the built-in default; that is the whole precedence
reachable here, because `--godot` is a per-**command** option and this surface is the root
(a later command's explicit `--godot` still overrides it). Reading the Git revision does
run `git`; that is a different process from the one under suspicion, and it degrades to
`null` when unavailable.

**No schema/protocol version.** gda has none; inventing one would create a contract nothing
honors. Pinned by `test_payload_declares_no_schema_or_protocol_version`.

**Scope call:** an unresolvable Git revision/dirty state is reported as `null` with no
per-field reason string. The issue asks for a stated reason only for the engine version.

## Real invocations (this worktree's editable install)

```
$ gda --version
gda 0.9.0
exit=0

$ gda --version --json
{"gda_version":"0.9.0","executable":"/Users/lelej/projects/godot-agent-worktree/w2-659/.venv/bin/gda","interpreter":"/Users/lelej/projects/godot-agent-worktree/w2-659/.venv/bin/python","package_path":"/Users/lelej/projects/godot-agent-worktree/w2-659/src/gda","install_kind":"editable","source":{"root":"/Users/lelej/projects/godot-agent-worktree/w2-659","revision":"5a61a8130b727884961866f654f3907b65d22e98","dirty":false},"godot":{"binary":"/Users/lelej/Applications/Godot.app/Contents/MacOS/Godot","version_unavailable_reason":"the engine reports its version only from a running process, and this surface never launches Godot; run `gda info --json` for it"}}
exit=0

$ gda --json --version
{"gda_version":"0.9.0",…,"package_path":"/Users/lelej/projects/godot-agent-worktree/w2-659/src/gda",…}   # byte-identical to the line above
exit=0
```

(The payload above is this branch describing itself at the merge candidate `5a61a813`: an
editable install, its source checkout named, its revision pinned — the exact disclosure
GDA-DF-043 lacked. `dirty` reads `false` because the worktree is committed and clean at
this head; the `true` path is covered by
`test_editable_install_reports_a_dirty_working_tree`.)

**The `$GIT_DIR` redirect, reproduced against the fixed CLI:**

```
other repo HEAD    = f674459a5b5bafc3890ba2a674023fd24cddf83c
this worktree HEAD = 5a61a8130b727884961866f654f3907b65d22e98

# the mechanism — plain git, -C pointed at THIS worktree:
$ GIT_DIR=<other>/.git GIT_WORK_TREE=<other> git -C <this worktree> rev-parse HEAD
f674459a5b5bafc3890ba2a674023fd24cddf83c          # the OTHER repo's HEAD

# gda after the fix:
$ GIT_DIR=<other>/.git GIT_WORK_TREE=<other> gda --version --json
root    : /Users/lelej/projects/godot-agent-worktree/w2-659
revision: 5a61a8130b727884961866f654f3907b65d22e98   # this checkout's own
```

## Surface diff — every change enumerated

Baseline captured before the first edit, re-dumped after both rounds.

| Surface | Diff |
| --- | --- |
| `gda schema` | **byte-identical to the pre-work baseline** (no command added; the payload model is not in the manifest, so `package_path` does not leak in) |
| `gda --help` | 2 option help strings only: `--version` now says "…; with --json, emit structured install provenance instead (no Godot is spawned)."; `--json`'s trailing "the root itself prints no JSON payload" replaced with "with --version it emits structured install provenance (`--help` stays text)" — it was about to become false |
| per-command `--help` | unchanged |
| `gda skill` | +8 lines: one **Provenance** bullet in `## Setup` |
| README | +1 row in the **Global flags** table (`--version` was not listed there at all) |
| `docs/README.{zh-CN,es,ja}.md` | the same row, translated in-language; markers re-stamped with `scripts/update_readme_i18n.py` |

## Gates (all run, all green, after the review fixes)

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1342 passed, 406 deselected |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1342 passed, 406 deselected |
| `uv run ruff format --check` | 432 files already formatted |

No Godot e2e run: the surface must not spawn the engine, and
`test_the_surface_never_launches_godot` proves it by tripping `gda.runner.launch` (the one
`Headless launch` primitive) and by asserting the invocation's recorded `subprocess.run`
calls are non-empty and all `git`.

## Tests

`tests/test_version_provenance.py` (39 tests) covers: both argv orders producing the same
payload; the bare human line; `--help` precedence; the required field set; the absence of a
schema/protocol key; **editable** installs (real temp git repo — revision, clean and dirty,
non-repo → `null`, non-`file:` URL → no checkout, **an inherited `GIT_DIR`/`GIT_WORK_TREE`
pointing at a second real repo asserting this checkout's revision is what comes back**, and
**a fixture repo configured `status.showUntrackedFiles=no` that still reports `dirty: true`**
— that test also asserts plain `git status --porcelain` sees the tree as clean, so it is its
own proof the config was in force); **wheel** installs (no PEP 610 record, local-wheel
`archive_info`, absent and explicit non-editable `dir_info`) via faked distribution metadata
— a real wheel install is not producible from inside this editable checkout, so that half is
unit-tier by construction, not by convenience; **`unknown`** (unparseable JSON, non-object
document, non-object `dir_info`, seven off-spec `editable` values, and a damaged record that
names a real checkout it must not vouch for) both through `build_version_provenance` and
through the CLI; `package_path` naming the loaded module, plus a **real subprocess
`PYTHONPATH` shadow** that copies the package elsewhere and asserts the field exposes the
divergence; **all three read arms pinned at the reader** — missing → `ABSENT`, malformed and
whitespace-only → `PRESENT` (verbatim), a raising reader and an uninstalled distribution →
`UNREADABLE` — plus the two once-collapsed arms end to end through the CLI with a `wheel`
control, so the tests prove a distinction rather than a constant (mutation-checked: restoring
either collapse fails five tests); the engine version key's **absence** plus its reason, and the mirror case where a
resolved version omits the reason; the no-launch guard; and a real out-of-process
`python -m gda --version --json`.

`tests/test_json_flag_acceptance.py`: #671's hand-off test
`test_root_version_stays_text_in_both_json_orders` is **flipped** to
`test_root_version_is_structured_in_both_json_orders`, plus a companion pinning that the
flag-less spelling stays text.


qinhaihong-red:
**Changes requested.** The provenance boundary is cohesive, but three exact-head failures can still make the preflight assert stronger provenance than it has.

Reviewed refs:

- Head: `114d0fccaf68b3877400489e38407900ce3c4ca8`

## Domain architecture

### [P1] Force untracked-file enumeration before reporting an editable checkout as clean

`src/gda/provenance.py:258`

`_source_checkout()` runs `git status --porcelain` without an explicit untracked-file policy. Git therefore honors repository or user configuration such as `status.showUntrackedFiles=no` and can hide untracked Python files. The implementation then reports `dirty: false` even though the module contract at `src/gda/provenance.py:94-97` explicitly counts untracked importable source as dirty.

This breaks the central ownership invariant of `gda.provenance`: it is the single authority tying a long run to the editable source state that produced it.

Minimum fix: invoke Git with a configuration-independent policy such as `status --porcelain --untracked-files=all`, and add a regression that sets `status.showUntrackedFiles=no`, creates an untracked importable file, and still expects `dirty is True`.

Exact-head reproduction: with that Git configuration and an untracked `runtime_shadow.py`, `_source_checkout()` returned `dirty=False`; adding `--untracked-files=all` reported the file.

The remaining structure should be retained: `gda.cli` is the inbound root adapter; `gda.provenance` owns payload models and assembly; `gda.binary` owns Godot-binary resolution; and the runtime flow remains directed from the Click callback through metadata/Git/path adapters to the Pydantic result. No cycle or speculative command was introduced.

Load-bearing claims remain `open` because the repository has no claim-evidence-state vocabulary:

- `DA-692-1`: `gda.provenance` is the single payload owner. Evidence: models and assembly are colocated; disconfirmed if a future `doctor` re-derives the fields.
- `DA-692-2`: this surface never launches Godot. Evidence: the exact call graph and no-launch tests; disconfirmed by a future dependency on an engine-spawn path.
- `DA-692-3`: `dirty` detects importable untracked source. The reproduced Git-configuration counterexample currently disproves this at the advertised scope.

## Standards

No Standards finding survived exact-head verification. The diff follows the applicable ownership, CLI taxonomy, structured-output, lint/format, and type-check rules; no baseline smell was substantiated.

## Spec

### [P2] Invalid provenance metadata is reported as a definite wheel install

`src/gda/provenance.py:167,295-309`

Issue #659 requires an install kind so production runners can pin or verify the implementation. `read_direct_url()` collapses an absent record and malformed or unreadable metadata into the same `None`; `build_version_provenance()` then converts all of those cases into `install_kind="wheel"` with `source=null`.

Impact: a damaged editable installation can be confidently misreported as an immutable wheel—the exact false provenance this preflight exists to prevent.

Minimum fix: distinguish “record absent” from “record invalid/unreadable”; fail closed or report an explicitly unresolved kind rather than asserting `wheel`. Validate that `dir_info.editable` is the boolean `true`, and add a CLI-level malformed-metadata regression.

### [P2] The unavailable Godot version is present despite the omission contract

`src/gda/provenance.py:118-120,310-314`; `src/gda/cli.py:104-105`

Issue #659 says the detected Godot version appears only when it can be obtained without spawning the engine, otherwise it is omitted with a stated reason. The model always serializes `"version": null` beside that reason.

Impact: presence-sensitive consumers do not receive the promised distinction between an observed version and an unavailable field.

Minimum fix: omit only `godot.version` when it is unavailable, retain `version_unavailable_reason`, and assert key absence in the contract test.

## Validation

- The three counterexamples above were independently reproduced on the exact head.
- Focused suites reported `58 passed` on the architecture pass and `40 passed` on the Standards/Spec pass.
- `git diff --check` passed; focused Ruff checks passed.
- Required GitHub CI is green. The Godot e2e job is skipped.
- A real wheel installation was not independently exercised; wheel behavior is covered through distribution-metadata seams.

Summary: Domain architecture — 1 finding, worst P1. Standards — 0 findings. Spec — 2 findings, worst P2.

qinhaihong-red:
Thanks — all three verified on `114d0fcc` and **all three fully adopted** in `d05a286d`.

| # | Finding | Decision | Fix | Regression |
| 1 | P1 — `status.showUntrackedFiles=no` hides untracked source, so `dirty` contradicted the stated contract | **Fully adopted** | `git status --porcelain --untracked-files=all` — the flag overrides the config, so the verdict is a property of the tree, not of someone's git config | `test_dirty_ignores_the_repositorys_untracked_files_config`: fixture repo with `status.showUntrackedFiles=no` + an untracked `runtime_shadow.py`, expects `dirty is True`. It also asserts plain `git status --porcelain` reports the tree **clean**, so the test carries its own proof the config was in force — it fails on the old code path by construction. |
| 2 | P2 — "record absent" and "record malformed" both became a confident `wheel`, misreporting a damaged editable install as immutable | **Fully adopted** | Seam split: `read_direct_url_text()` returns the RAW text (blank → absent) and a pure `classify_install()` decides. New `InstallKind.UNKNOWN` for a record that exists but is off-spec: unparseable JSON, non-object document, non-object `dir_info`, or an `editable` that is not the boolean PEP 610 types it as — **truthiness is no longer the test**. `unknown` never claims a checkout, even when the damaged record names a good one. | 8 tests: `test_malformed_metadata_is_unknown_not_wheel`, `test_a_non_object_record_is_unknown`, `test_a_non_object_dir_info_is_unknown`, `test_a_non_boolean_editable_is_unknown_not_truthiness` (7 off-spec values incl. `"false"`, `"true"`, `1`, `0`), `test_unknown_never_claims_a_checkout`, `test_malformed_metadata_reaches_the_cli_as_unknown` (**CLI level**, as requested), `test_classify_install_is_pure_and_total`, plus `test_an_explicit_non_editable_dir_info_is_a_wheel` for the `false` side. |
| 3 | P2 — `"version": null` beside the reason, but the contract is "omitted with a stated reason" | **Fully adopted** | `GodotProvenance` gained a wrap `model_serializer` that drops whichever of `version` / `version_unavailable_reason` does not apply, so **exactly one is ever present** — the omitted-never-null convention, noted in the field docstring. | `test_godot_version_key_is_omitted_with_a_stated_reason` asserts **key absence** (`"version" not in payload["godot"]`) and pins the exact key set; `test_a_resolved_godot_version_would_omit_the_reason_instead` pins the mirror case so a future spawn-free version source cannot leave a stale null reason. |

**Deliberate non-change, stated for the record:** `SourceCheckout.revision`/`dirty` keep their explicit nulls rather than following finding 3's omission rule. The two nulls mean different things — for `godot.version` a null would claim gda looked and found nothing when it declined to look, while for a named checkout the null **is** the finding ("looked, could not resolve"). Both docstrings now say so, so the asymmetry reads as a decision rather than an oversight.

**Doc duty:** the third install kind is now in the Skill bullet (`wheel`, `editable`, or `unknown` when the install metadata cannot be read — gda will not guess) and in the README `--version` row, with the three translations re-translated in-language and re-stamped via `scripts/update_readme_i18n.py`.

**Payload shape now** (this branch describing itself at the merge candidate — note the absent `godot.version` key):

```json
{
  "gda_version": "0.9.0",
  "executable": "/…/w2-659/.venv/bin/gda",
  "interpreter": "/…/w2-659/.venv/bin/python",
  "package_path": "/…/w2-659/src/gda",
  "install_kind": "editable",
  "source": { "root": "/…/w2-659", "revision": "d05a286d…", "dirty": false },
  "godot": {
    "binary": "/Users/…/Godot.app/Contents/MacOS/Godot",
    "version_unavailable_reason": "the engine reports its version only from a running process, and this surface never launches Godot; run `gda info --json` for it"
  }
}
```

**Validation at `d05a286d`** — `uv run pytest -m "not e2e"` 1337 passed / 406 deselected; the `GITHUB_ACTIONS=true` variant likewise 1337 passed; `ruff check` clean; `ruff format --check` 432 files already formatted; `pyright` 0 errors, 0 warnings. `gda schema` re-diffed **byte-identical to the pre-work baseline** (the payload model is not in the manifest, so neither the new kind nor the key omission leaks into it); `gda --help` unchanged; the only `gda skill` diff is the Provenance bullet. No Godot e2e — the surface still spawns no engine, guarded by `test_the_surface_never_launches_godot`. PR body refreshed with the new claims, payload example and re-run evidence.

Not adopted, noted only: ADR-0005's `gda version` naming ambiguity — no action here, carried to the #670 `doctor` discussion.

qinhaihong-red:
**Recheck: one P2 remains open.** The untracked-file and Godot-version findings are fully fixed at `d05a286d`, but the install-metadata finding is only partially fixed: non-empty malformed JSON is now `unknown`, while a blank record or a read failure still becomes a confident `wheel`.

### [P2] Keep blank and unreadable `direct_url.json` distinct from an absent record

`src/gda/provenance.py:222-227,241-242`

`Distribution.read_text()` uses `None` for an absent record. The new seam additionally returns `None` when the record is blank and when reading it raises `OSError`; `classify_install(None)` then reports `InstallKind.WHEEL`.

That leaves two arms of the original counterexample open:

```text
non-empty malformed JSON -> unknown  (fixed)
off-spec editable value  -> unknown  (fixed)
blank existing record    -> wheel    (still open)
read_text raises OSError -> wheel    (still open)
```

I reproduced both remaining arms independently on the exact head. The `OSError` probe produced `raw=None, kind=wheel`; a whitespace-only record likewise produced `raw=None, kind=wheel`. The current test named `test_read_direct_url_text_treats_a_blank_record_as_absent` pins the latter collapse rather than distinguishing absence from invalid content. There is no regression covering the `OSError` arm.

**Impact:** a damaged editable install with an empty or unreadable PEP 610 record can still be asserted to be an immutable wheel, which is the false provenance the original finding asked this boundary to prevent. It also contradicts the new function contract at lines 211-216 and the PR reply's claim that metadata which cannot be read is never guessed as `wheel`.

**Minimum fix:** preserve an explicit `absent / present / unreadable` distinction at the read seam. Only a truly absent record should classify as `wheel`; whitespace-only content and read failures should classify as `unknown` without failing the preflight. Add seam plus CLI/build-level regressions for a whitespace-only record and a distribution reader that raises `OSError`.

The other two findings are closed:

- `git status --porcelain --untracked-files=all` and its configured-repository regression correctly make `dirty` independent of `status.showUntrackedFiles`.
- `godot.version` is omitted when unavailable, with the reason retained; the mirror case omits the reason when a version is present.

Recheck validation at immutable head `d05a286da5c472eae71347d80185d6e3f6a3a7af`: `34` focused provenance tests passed, focused Ruff passed, and `git diff --check` passed. Current required CI is green; Godot e2e is skipped.

qinhaihong-red:
Recheck acknowledged — you are right, and the finding-2 row in my previous reply overclaimed. **Fully adopted in `5a61a813`.**

## What round 1 of finding 2 actually missed

I split "malformed" out of "absent" but left the read seam returning `Optional[str]`, and **two arms folded straight back into `None` → confident `wheel`**:

| Arm | Round-1 behaviour | Why it was wrong |
| --- | --- | --- |
| (a) whitespace-only `direct_url.json` | `raw.strip()` falsy → `None` → `wheel` — and `test_read_direct_url_text_treats_a_blank_record_as_absent` **pinned that collapse as correct** | a record that exists but says nothing is off-spec, not absent |
| (b) `read_text()` raises `OSError` | swallowed by the `except` → `None` → `wheel`, with no regression at all | gda failed to look; it did not learn that nothing is there |

Both contradicted the function's own contract and my "never guessed as wheel" claim. A two-state seam cannot carry a three-state fact — that was the real defect, not the two symptoms.

## The fix — a three-state read seam

Seam shape chosen: **an explicit `RecordState` tag on a frozen `DirectUrlRecord`** (`read_direct_url_record() -> DirectUrlRecord`), not a sentinel and not exception-to-verdict. A sentinel fights the type checker at every call site, and exception-to-verdict would have put the read policy back in the caller; a tagged record makes the three arms exhaustively matchable and keeps `classify_install` pure with a total, side-effect-free signature.

| Read arm | Observation | Verdict |
| --- | --- | --- |
| `ABSENT` | reader returns `None` ("no such file") | **`wheel`** — the ONLY observation that earns it (PEP 610 default, unchanged) |
| `PRESENT` | any content, whitespace included, handed over **verbatim** | `editable` / `wheel` / `unknown`, per the existing off-spec handling |
| `UNREADABLE` | reader raised, or the distribution is not installed | **`unknown`**, and the preflight still emits its payload |

No blank-collapse at the seam: `"   \n"` now travels through as `PRESENT`, and `classify_install`'s existing `JSONDecodeError` branch yields `unknown` — no new classifier logic. The enum docstring's "refusal to guess" claim is now true for all three arms: only `ABSENT` reaches `wheel`.

## Regressions

**At the reader** (all four arms, driven through the real `Distribution.read_text` via a fake metadata reader, not a faked verdict): `test_a_missing_record_is_absent`, `test_a_malformed_record_is_present_and_handed_over_verbatim`, `test_a_whitespace_only_record_is_present_not_absent` (the rewrite of the test that pinned the wrong behaviour), `test_an_unreadable_record_is_unreadable_not_absent`, `test_an_uninstalled_distribution_is_unreadable_not_absent`.

**End to end** — both once-collapsed arms at CLI level: `test_a_whitespace_only_record_reaches_the_cli_as_unknown` and `test_an_unreadable_record_reaches_the_cli_as_unknown`, each asserting `install_kind == "unknown"`, `source is None`, **exit 0, and a complete payload** (degrade, never die). Plus `test_a_missing_record_reaches_the_cli_as_wheel` as the control, so the pair proves a distinction rather than a constant.

`test_classify_install_is_pure_and_total` now walks every arm directly; the eight `unknown`/`wheel`/`editable` classification tests are unchanged and green.

**Mutation-checked, not asserted:** restoring either collapse (`except → ABSENT`, `not raw.strip() → ABSENT`) fails exactly 5 tests — `test_a_whitespace_only_record_is_present_not_absent`, `test_an_unreadable_record_is_unreadable_not_absent`, `test_an_uninstalled_distribution_is_unreadable_not_absent`, and both CLI arms. Mutation reverted; suite green.

## One hole now documented rather than claimed away

`importlib.metadata.read_text` **suppresses some of its own failures** — notably `PermissionError` — and returns `None`, which is indistinguishable from a missing file at this seam. That case still reports `wheel`. It is stated in the seam docstring and in the PR body; closing it would mean reading the metadata path directly instead of through `importlib.metadata`, which I did not judge in scope here. Say the word if you want it filed.

## Docs

The skill and README wording — "`unknown` when the install metadata cannot be read — gda will not guess" — is literally true for all three arms, so it is unchanged and **no re-translation was needed**. Verified: the `gda skill` output diffs clean against the previous head.

## Validation at `5a61a813`

`uv run pytest -m "not e2e"` **1342 passed** / 406 deselected; the `GITHUB_ACTIONS=true` variant likewise **1342 passed**; `ruff check` clean; `ruff format --check` 432 files already formatted; `pyright` 0 errors, 0 warnings. `gda schema` re-diffed **byte-identical to the pre-work baseline**; `gda --help` unchanged; `gda skill` unchanged. No Godot e2e — the surface still spawns no engine. PR body updated: the install-kind section now documents the three read arms as a table (replacing the claim this recheck corrected), with the known hole stated and the evidence re-run at this head.

qinhaihong-red:
**Second recheck: the two reported seam-level arms are fixed, but the original P2 remains open on the real `PathDistribution` read path.** Whitespace now reaches the classifier as `PRESENT → unknown`, and a reader that visibly raises becomes `UNREADABLE → unknown`. However, `importlib.metadata.PathDistribution.read_text()` suppresses several read failures itself and returns the same `None` it returns for absence; the current seam then records `ABSENT → wheel`.

Reviewed refs:

- Base tip: `c11f07db202d6bffad3de37a45159fa20bfe8165`
- Head: `5a61a8130b727884961866f654f3907b65d22e98`

### [P2] Do not let `PathDistribution.read_text()` turn unreadable/corrupt metadata into an absent record

`src/gda/provenance.py:243-269,284-287`

The stdlib implementation suppresses `PermissionError`, `IsADirectoryError`, `NotADirectoryError`, `FileNotFoundError`, and `KeyError`, returning `None` for all of them. `read_direct_url_record()` cannot distinguish that ambiguous `None` and currently marks it `ABSENT`; `classify_install()` then returns a confident `WHEEL`.

I independently reproduced two real-reader cases on the exact head:

```text
mode-000 direct_url.json:
  path exists                 = true
  direct Path.read_text       = PermissionError
  PathDistribution.read_text  = None
  record state                = absent
  CLI install_kind            = wheel

direct_url.json is a directory:
  entry is directory          = true
  PathDistribution.read_text  = None
  record state                = absent
  install_kind                = wheel
```

These are not a new feature request: they are the remaining real implementation path for the original requirement that existing but invalid or unreadable PEP 610 metadata must not be asserted to mean an immutable wheel. Documenting the hole is honest, but does not close that correctness claim.

It also leaves the public wording false at this head:

- `README.md:549`: `unknown` “when the install metadata cannot be read”
- `src/gda/skill/SKILL.md:33`: “gda will not guess”

### Minimum fix direction

Keep the new `DirectUrlRecord` / `RecordState` model and the pure classifier; the three-state architecture is appropriate. Change only the reader adapter so `ABSENT` is produced from positive missing-file evidence, not from `PathDistribution.read_text() is None` alone:

1. Resolve the distribution metadata entry/path and read it through a path-backed adapter whose outcomes stay distinct: missing entry → `ABSENT`; bytes/text, including empty or whitespace → `PRESENT`; any other read/type failure → `UNREADABLE`.
2. If a platform/reader can supply only an ambiguous `None`, do not promote that ambiguity to `ABSENT` without independent evidence that the entry is missing; degrade it to `UNREADABLE` instead.
3. Add real-reader regressions, not another faked verdict:
   - a `PathDistribution` whose underlying `Path.read_text` raises `PermissionError` → reader `UNREADABLE`, CLI `unknown`, exit 0;
   - a metadata entry named `direct_url.json` that is a directory → `UNREADABLE` / `unknown`;
   - an actually missing entry as the control → `ABSENT` / `wheel`.

This keeps the fix local to the read boundary, preserves the useful tagged-record design, and makes the README/Skill promise true without changing the classifier or public payload shape.

Validation at immutable head `5a61a8130b727884961866f654f3907b65d22e98`: `39` focused provenance tests passed, focused Ruff passed, and `git diff --check` passed. The focused suite currently covers a fake reader that raises, but has no `PathDistribution` regression for a failure the stdlib suppresses. Required CI is green; Godot e2e is skipped and is not relevant to this no-engine surface.

qinhaihong-red:
Second-recheck response — the remaining P2 is closed on `ffa105be` (handled directly by the orchestrator).

| [P2] `PathDistribution.read_text()` suppresses `PermissionError`/`IsADirectoryError`/`NotADirectoryError`/`FileNotFoundError`/`KeyError` into the same `None` as absence; the seam promoted that ambiguity to `ABSENT` → confident `wheel` (mode-000 and directory-shaped records reproduced). | Fully adopt (the suggested fix direction, with one correction) | Reproduced both arms; also found `locate_file` — the reviewer's suggested entry point — resolves installed files against **site-packages**, not the dist-info directory, so it locates the wrong path for a metadata file (first attempt's tests caught this immediately). | The reader adapter now reads `direct_url.json` from the distribution's own metadata directory (`PathDistribution._path`, covering directory- and archive-backed installs): `FileNotFoundError` alone → `ABSENT` (positive missing evidence); any other read failure (`OSError` incl. permission/directory shapes, `KeyError` on zip backends) → `UNREADABLE` → `unknown`; success — any content — → `PRESENT`. An exotic `Distribution` without `_path` can only offer the ambiguous `read_text` `None`, which degrades to `UNREADABLE`, never to `ABSENT` (the recheck's rule 2). `RecordState`/`classify_install` untouched. The test fixture was rebuilt to construct REAL dist-info directories on disk and point `Distribution.from_name` at a real `PathDistribution`, so the mode-000, directory-shaped, missing (control) and malformed arms all exercise the true reader — no faked verdicts. CLI-level: the mode-000 record through the real CLI → `install_kind: "unknown"`, `source: null`, complete payload, exit 0. | Red-proof: with the previous seam restored (stash probe), exactly the three new real-failure tests fail; green with the fix. `tests/test_version_provenance.py` 40 passed; full fast suite 1343 ×2 variants; ruff/format/pyright clean; `gda schema` byte-identical (sha `06f2669e…`). README:549 / SKILL.md:33 wording ("cannot be read — gda will not guess") is now literally true with no text change. |

qinhaihong-red:
**Third recheck: one P2 remains.** The real `PathDistribution` ambiguity reported in the second recheck is fixed at `ffa105be`: a genuinely missing record is `ABSENT`, while mode-000 and directory-shaped records are `UNREADABLE`, and the archive-backed missing-record control also remains `ABSENT`. However, the replacement adapter regresses the previously fixed raised-reader arm for `Distribution` implementations without `_path`.

Reviewed refs:

- Previous head: `5a61a8130b727884961866f654f3907b65d22e98`
- Current head: `ffa105bea9e0a2abce1537f4b26f3586fea937dd`
- Base tip: `c11f07db202d6bffad3de37a45159fa20bfe8165`

### [P2] Keep failures from the no-`_path` reader fallback inside the `UNREADABLE` boundary

`src/gda/provenance.py:277-282`

When `metadata_dir is None`, the code calls `dist.read_text("direct_url.json")` outside the guarded read block. A custom `importlib.metadata.Distribution` is not required to expose the private `PathDistribution._path`, and its `read_text()` can still fail while accessing metadata. That failure currently escapes instead of becoming `RecordState.UNREADABLE`.

I independently reproduced this with a concrete `Distribution` subclass that has no `_path` and whose `read_text()` raises `PermissionError`:

```text
5a61a813: gda --version --json -> exit 0, install_kind = unknown, complete JSON
ffa105be: gda --version --json -> exit 1, no JSON, PermissionError escapes
```

This contradicts the function contract that “a reader that raises is `UNREADABLE`” and the preflight's degrade-with-evidence behavior. It also makes the author's claim that the no-`_path` fallback can only return an ambiguous `None` too narrow: the abstract reader can raise before producing either text or `None`.

**Minimum fix:** keep the current path-backed branch and classifier unchanged. Wrap only the fallback `dist.read_text()` call in the same failure boundary (at minimum `OSError`, plus the other reader/backend failures already recognized by this adapter) and return `UNREADABLE`; retain `None -> UNREADABLE` and text -> `PRESENT`. Add a seam regression and a CLI-level regression for a no-`_path` `Distribution` whose reader raises, asserting `unknown`, exit 0, and a complete payload.

Validation on the immutable current-head snapshot: `40` focused provenance tests passed; focused Ruff, Pyright, and `git diff --check` passed. Required GitHub CI is green; Godot e2e is skipped and is not relevant to this no-engine surface.

qinhaihong-red:
Third-recheck response — the fallback regression is closed on `b0bba630` (handled directly by the orchestrator).

| [P2] The no-`_path` fallback called the abstract `Distribution.read_text()` OUTSIDE the guarded block, so a custom distribution whose reader raises escaped as a traceback (exit 1, no JSON) — regressing the raised-reader arm for that shape, and disproving the docstring's "can only offer the ambiguous None" claim. | Fully adopt (the minimum fix as suggested) | Reproduced by the recheck; confirmed the branch was the one read outside the failure boundary. | The fallback now shares the same boundary as the path-backed branch: `except (OSError, KeyError, ValueError)` → `UNREADABLE`; the ambiguous `None` stays `UNREADABLE`; only text is `PRESENT`. The docstring's claim widened to name the raise. Two regressions: a seam test with a pathless `Distribution` whose `read_text` raises `PermissionError` → `UNREADABLE`, and the same shape through the real CLI → `install_kind: "unknown"`, `source: null`, complete payload, exit 0. | Red-proof: with the previous seam restored (stash probe) exactly the two new tests fail (the CLI one with the escaping traceback); green with the fix. `tests/test_version_provenance.py` 42 passed; full fast suite 1345 ×2 variants; ruff/format/pyright clean; `gda schema` byte-identical (sha `06f2669e…`). |

Issue #658, opened 2026-08-15
script validate: resolve the owning project or refuse when the script is outside the selected root


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-035** (2026-08-11). Running `gda script validate examples/card-games/kung-fu/<script>.gd` from the repository root opened the script but resolved its `res://` preloads against the repository root, producing many false missing-file and derived type-inference errors on a valid project script. The single project-root mistake was buried under downstream noise.

## What to build

ADR-0006 fixes project resolution at `--project` > `$GDA_PROJECT` > cwd for the whole Phase-1 surface, and it explicitly **rejected** deriving the project from the target path (undefined outside any project; a call spanning multiple paths could imply different roots). This issue therefore requires the ADR-conformant behavior:

- When the script path lies outside the resolved project, **refuse before parsing** with a structured diagnostic (fits the existing `project_not_found` / a sibling code) that names both the script's location and the resolved project — never emit false dependency errors.
- Add a `project_root` field to the `script validate` success result (`ScriptValidateResult`; this is a declared schema addition), reporting the resolved project.
- Nearest-`project.godot` discovery is **out of scope** pending an ADR-0006 amendment; do not implement it here.

Related: #663's batch mode must resolve every path against the one resolved project — ADR-0006's rejection rationale names exactly that hazard. #670 owns the validate/run path-form unification.

## Acceptance criteria

- [ ] Validating a project script from outside the resolved project refuses with a structured diagnostic naming the mismatch; it never reports false dependency errors.
- [ ] The `script validate` success result reports the resolved project root in a `project_root` field.
- [ ] A regression test covers a script under a nested project validated from an ancestor directory.

## Blocked by

None - can start immediately (the refusal path conforms to ADR-0006 as decided).

qinhaihong-red:
> *This was generated by AI during triage.*

Triage: moved to `ready-for-agent`. The issue body (revised 2026-08-15 after a full artifact review against the dogfooding source, the ADRs, and the current code) is the authoritative agent brief — current behavior, deliverables, acceptance criteria, and scope boundaries are specified there.

Issue #653, opened 2026-08-15
Preflight and redirect the Godot user:// log/user-data root for the Headless launch


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`). This is the most frequently reconfirmed defect family in the document (about 30 distinct reconfirmation notes — a conservative count):
- **GDA-DF-001** (2026-08-08) and **GDA-DF-014** (2026-08-08) — in a filesystem-restricted process where the project is writable but `~/Library/Application Support/Godot/` is not, every command that performs a `Headless launch` (`script validate/run`, `info`, `scene`, …) fails before project code: Godot cannot create `user://logs`, then crashes with signal 11 in `RotatedFileLogger::rotate_file()`. gda reports `engine_crashed` with a noisy backtrace instead of a typed environment refusal.
- **GDA-DF-042** (2026-08-13/14) and **GDA-DF-030 "Engine commands crash under the restricted execution profile"** (2026-08-12, with 2026-08-13 reconfirmations; the document's GDA-DF numbering collides across sessions — identified by title and date) — same failure across `script run`, `info`, `project info`; sequential and concurrent alike; `XDG_DATA_HOME`/`XDG_CONFIG_HOME` overrides do not redirect the resolved `user://logs` location.
- **GDA-DF-028** (2026-08-10) — beyond logs: twelve persistence/settings suites failed or timed out under the restricted profile because `user://` itself cannot be redirected; the byte-identical suites passed with the standard path writable. An unwritable `user://` looks like a broad product regression.

Key evidence: passing a wrapper executable through `--godot` that prepends `--log-file /tmp/....log` removed the crash entirely, so log-path redirection is proven sufficient for the crash half. The daemon's `Engine session` is already immune because it passes `--log-file` (ADR-0022); the gap is the one-shot `Headless launch` path. Restricted agent sandboxes are a recurring gda environment in the dogfooding record.

## What to build

- Preflight the resolved user-data/log root before spawning Godot; on failure return a new `Classifier error code` (ENVIRONMENT category, e.g. `user_data_unwritable`) instead of entering the engine crash path.
- Expose a first-class per-invocation `--log-file` and/or `--user-data-root` (or default one-shot headless operations to an isolated writable temporary root), instead of requiring callers to manufacture a wrapper executable.
[...]

Issue #293, opened 2026-06-25
docs(skill): sync SKILL.md live-ops tables to the surface + add a drift gate


qinhaihong-red:
## What to build

The bundled `src/gda/skill/SKILL.md` live-ops command tables have drifted from the actual `gda`
command surface. Bring them current **and** add a test gate so they cannot silently drift again.
The headless tables are already in sync; the drift is isolated to the live section.

Two parts, shipping in one PR:

**(A) Content sync** — fix the live-ops tables to match the surface:
- Drop the phantom `diag log` — it was removed and is superseded by `logger tail --raw` (ADR-0026).
- Add a `logger` group row with `tail` (live; `--raw` returns the verbatim lines that replaced
  `diag log`, `--level <min>` filters by minimum severity, `--limit N`).
- Surface `daemon start --scene` (boot a chosen scene; #285) alongside the existing `--windowed`.
- Optional polish: note that `diag errors` now carries a `callstack`, and mention runtime
  logs/diagnostics in the discovery blurb.

**(B) Drift gate** — a unit test that parses SKILL.md's `group → commands` tables and asserts they
exactly match the live `build_surface_manifest(app)` (the same authoritative source `gda schema`
projects from). A future surface change without a matching SKILL.md edit then fails CI. This applies
the ADR-0023 "projection, not a parallel registry to keep in sync" principle to the Skill's tables —
closing the gap that ADR-0024's version-lock leaves open (it guarantees the file ships at the
package version, not that its prose matches the surface).

## Acceptance criteria

- [ ] SKILL.md lists `logger | tail`; the phantom `diag log` is gone; `daemon start --scene` is mentioned
- [ ] A unit test asserts SKILL.md's group→command tables match `build_surface_manifest(app)`, and fails on an intentional drift
- [ ] The gate passes against the corrected SKILL.md; `pytest -m "not e2e"` green

## Blocked by

None - can start immediately.

PR #693, opened 2026-08-16
feat(gda): accept both script-path forms for script run (ADR-0031 amendment)

### LINKED ISSUE #675: script run: accept both script path forms (amend ADR-0031's res://-only scope)
> *This was generated by AI during triage.*

## Parent

Split from #670 during triage (2026-08-15): the ADR-bearing path-unification bullet, separated so the hint/alias work can proceed independently.

## What to build

`gda script run` rejects the project-relative script path form that every other `script` command accepts through the shared path normalization; callers must rewrite to `res://…`.

**Decision (recorded at triage):** amend ADR-0031's `res://`-only scope. Its stated rationale — "consistent with the rest of the `script` command group" — no longer holds: the rest of the group accepts both forms. `script run` accepts both the project-relative and `res://` forms through the same normalization and reports the canonical `res://` path in the result. The ADR-0031 amendment (a dated note preserving the original decision text) lands with the change.

Dogfooding source: **GDA-DF-019** (2026-08-09), `goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`.

## Acceptance criteria

- [ ] One script-path representation (project-relative or `res://`) works for both `script validate` and `script run`.
- [ ] The result reports the canonical `res://` path.
- [ ] The ADR-0031 scope amendment lands with the change.
- [ ] Regression covers both path forms on both commands.

## Blocked by

None - can start immediately (the amendment direction is decided above).

### LINKED ISSUE #699: normalize_path: a ~nosuchuser path crashes with a raw RuntimeError surface-wide
## Review-round source

Found during PR #693's review (wave 2); pre-existing and surface-wide, not introduced there. The shared path normalization calls `Path.expanduser()`, which raises `RuntimeError` (not `ValueError`) for `~nosuchuser/...` — pydantic does not convert it, so every command using the shared `NormalizedPath` (`script get`/`validate`/`run`, and siblings) crashes with an uncaught `RuntimeError: Could not determine home directory.` traceback instead of a structured Error envelope. This contradicts the module's own promise that path problems come back as a structured `GdaError`, never a crash.

## What to build

Catch the expansion failure at the shared normalizer and surface it as the existing `invalid_path` refusal (message naming the unexpandable `~user` prefix). One fix at the single home; every consumer inherits it.

## Acceptance criteria

- [ ] `gda script validate '~nosuchuser/x.gd' --json` (and any NormalizedPath consumer) returns a structured `invalid_path` envelope, exit 4 — no traceback.
- [ ] A unit regression pins the RuntimeError conversion at the normalizer; one command-level test proves the envelope end-to-end.
- [ ] Valid `~/...` expansion behavior is unchanged.

## Blocked by

None - can start immediately.

qinhaihong-red:
Closes #675
Closes #699

`gda script run` refused the project-relative script path that every other `script` command accepts, so a caller had to rewrite a path between `script validate` and `script run` (dogfooding GDA-DF-019). It now accepts both forms through the same normalization chain and reports the canonical `res://` path.

## Base branch

Targets `feat/gda-dogfooding-dev` per the wave branch model. GitHub auto-deleted that branch when promotion PR #688 merged, so this PR was briefly opened against `main`; the orchestrator has since recreated the branch at main's tip (`c11f07db`) and retargeted the PR back to it. The branch is rebased onto `c11f07db`.

## Decision (ADR-0031, second dated amendment)

ADR-0031 scoped `script run` to `res://` only, on the rationale that this is "consistent with the rest of the `script` command group (which all act on `res://`)". That rationale does not hold — the rest of the group takes a project-relative path too. The amendment is appended after #651's, with **zero prior lines modified**.

## Design — one address, no second normalizer

A project-relative path is lifted onto the `res://` root it is already relative to, then put through the **same** `canonical_res_path` #651 introduced. The lift adds no normalization rule of its own; the shared helper still does all `./`, `..` and duplicate-slash collapsing, for either spelling. Both forms therefore converge on ONE address before any launch, so the argv, the entry-load verdict matching and every message keep the single canonical identity #651 established.

`ScriptRunParams.path` now carries the shared `NormalizedPath` like every other path field, so the argv and `--params-json` paths normalize identically (ADR-0015) and a `~` prefix is refused as the absolute path it means, instead of becoming a nonsense `res://~/…`.

## Intended surface diffs (enumerated)

Baseline captured before the first edit; `gda schema` diff is confined to `script run` — the other **66 commands are byte-identical**.

`gda schema`, `script run` only:
1. Command description — the new path-form paragraph; "A non-res:// path" → "An absolute path".
2. `input` model description — both accepted forms, the `NormalizedPath` rationale.
3. `input.properties.path.description` — names both forms. (Property *shape* unchanged: `NormalizedPath` adds no schema keyword.)
4. `output` model description — what `path` is and why.
5. `output.properties.path` — **added** (declared schema addition, per the amendment).
6. `output.required` — gains `path`.

`gda script run --help`: the three text changes above (command docstring, argument help), no option/flag changes.

Behavior:
7. A project-relative path is accepted instead of `invalid_path`.
8. `invalid_path` message: `script run requires a res:// script path, got: X` → `script run requires a project-relative or res:// script path, got: X`. Same code, same category, same exit 4. Verified **not** projected into `gda schema` or `--help` — error message text is not part of the self-description surface.
10. **Refusal set** named in the help/schema prose. Refused before any launch: an absolute path; another engine scheme (`user://`, `uid://`); a path naming the project root (`""`, `"."`, `"sub/.."`, `res://`, `res://.`); and a path escaping above the root (`".."`, `"../outside.gd"`, and their `res://` spellings).
11. **Agent-visible code delta** (improvement, previously unenumerated): bare `res://` and `res://.` moved from `script_compile_failed` (base — decided post-launch, after spawning the engine) to `invalid_path` (head — decided pre-launch). The `..` escapes moved from a phantom `exit_status: 0` **success** to `invalid_path`. Category and exit 4 unchanged for the failure cases.

`gda skill` (the shipped SKILL.md, ADR-0024) — **exactly one line** changes, the `script` row's parenthetical note; the rest of the 277-line emission is byte-identical:
9. `run` executes a \`res://\` script one-shot → `run` executes a project script one-shot (address it project-relative or as \`res://\`, the same as \`script validate\`). The row's command list before the first `(` is untouched, so #293's surface-sync gate reads the same tokens.

Not changed: `render_script_run` (human output stays the script's passthrough; the canonical path is on the JSON result and already in every failure message).

## Validation

- `uv run pytest -m "not e2e"` — **1350 passed** (incl. the skill surface-sync gate, #293); also green with `GITHUB_ACTIONS=true` (help text touched).
- `uv run ruff check` / `ruff format --check` / `uv run pyright` — clean (0 errors).
- Godot **4.6.3 e2e**, `tests/test_e2e_script_run.py` — **36 passed**, including **wave-1's canonical-spelling arms unchanged and green**, the F1/F2/R2 refusal arms (`""`, `.`, `sub/..`, `user://`, `uid://`, `..`, `sub/../..`, `../outside.gd`, `res://..`, `res://../outside.gd`), and project-relative aliases (`no-such-script.gd`, `sub/../no-such-script.gd`, `./no-such-script.gd` and the `suite.gd` equivalents).
- Godot e2e `tests/test_e2e_script.py` — **the full suite, 56 passed** (not just the validate subset), including the new both-forms `script validate` arm.
- All gates re-run after the rebase onto current `main`.

Real-engine dogfood of the reported defect:

```
$ gda script run tests/logic.gd --project P --json      # was: invalid_path
{"path":"res://tests/logic.gd","exit_status":0,"stdout":"…ran ok\n",…}
$ gda script run res://tests/logic.gd --project P --json
{"path":"res://tests/logic.gd","exit_status":0,…}
$ gda script run tests/broken.gd --project P --json      # verdict still fires
{"error":{"code":"script_compile_failed","message":"script run: res://tests/broken.gd did not run — …"}}
$ gda script run /abs/tests/logic.gd --project P --json  # still refused
{"error":{"code":"invalid_path","message":"script run requires a project-relative or res:// script path, got the absolute path: '/abs/…'"}}
```

## Review round (findings F1, F2, F4, F5 adopted — `2d82e8c5`)

**F1 (MEDIUM-LOW, verified regression) — fixed.** A degenerate relative path lost its pre-launch guard. `normalize_path("")` → `.` → lifted to `res://.` → **engine launched**; the engine's `Can't load script: res://.` parses with the sentence period stripped to `res://`, which never matches `res://.`, so `entry_load_failure` returned `None` and the run reported **exit 0 SUCCESS** for a run that never happened. Reproduced on the real engine before fixing (`{"path":"res://.","exit_status":0,...}`). Reachable as `gda script run "$SCRIPT"` with an unset variable. Same for `.`, `./`, `sub/..`. The base refused all of these pre-launch.

**F2 (LOW-MEDIUM, verified regression) — fixed.** Non-`res://` engine schemes were silently mangled: `user://x.gd` → `res://user:/x.gd`, `uid://cabc…` → `res://uid:/cabc…`, spawning the engine and quoting an address the caller never typed. Reproduced before fixing. The refusal message is generalized accordingly — it hardcoded `got the absolute path:`, which would have misdescribed these.

**Guard's final shape.** One acceptance gate, `_project_scoped_res_path(script) -> str | None`, replaces the inline absolute-check-then-lift. It *accepts* the two portable forms and returns the single canonical address (so the canonical identity is still fixed at exactly one place), and returns `None` — refusal — for absolute, another engine scheme, and a collapse to the project root (`res://` or `res://.`). The raw input is kept for the message, which still quotes what the user typed.

The underlying `res://.` parser hole pre-exists on base and is **not** touched here; the guard closes the reachable path. Recorded by the orchestrator as a follow-up candidate, along with the pre-existing `~nosuchuser` `RuntimeError` in the shared `normalize_path`.

**F4 (record accuracy) — fixed.** The amendment's stated rationale for keeping absolute refused was falsifiable (the engine *does* run `--script <abs path>`, and `script validate` accepts absolute today). Replaced with the two verified rationales: (i) the engine reports a failed run under the **`res://` spelling even for an absolute in-project path**, so accepting absolute without a res:// mapping would break #651's canonical-identity match and reopen the phantom success; (ii) `--script <abs path OUTSIDE the project>` **does execute**, so accepting absolute would widen the Project-code execution surface past ADR-0009's Trusted project — a trust decision needing its own ADR.

**F5 (wording) — fixed in four places** (SKILL.md:79, `errors.py` message/docstring, the amendment, plus the `run` docstring/params for consistency). "the same as `script validate`" overstated parity — validate also accepts and echoes a third form, absolute. Now scoped to "the two **portable** forms", with the asymmetry stated explicitly rather than implied.

## Sibling-PR merge caveat

PR #695 (#658) also edits `src/gda/commands/script.py` — `ScriptValidateResult` and shared context lines two blank lines from my hunk. A clean merge is expected, but it is a shared **file**, not just a shared test file as this body previously implied. #695 also touches shared `normalize_path` (`"://" in path` → `is_engine_virtual_path(path)`), behaviorally equivalent for my chain. **Serialized merge expected**; my position is before #692's and this PR is currently MERGEABLE.

## Review round 2 (R2-F1, R2-F2 + doc sync — `fd2e6b69`)

**R2-F1 / R2-F2 (both verified, both fixed).** The gate refused a path that *collapses to* the project root but not one that *escapes above* it. Reproduced on the real engine before fixing:

- `gda script run ..` → `res://..` → launched → engine's `Can't load script: res://..` parses back as `res://.` (sentence period stripped) → no match → **`{"path":"res://..","exit_status":0}`**, the same phantom-success class as F1, one level up. Same for `sub/../..`.
- `gda script run ../outside.gd` → **executed a script outside the project**, printing `OUTSIDE THE PROJECT RAN`. This one contradicted my own amendment: rationale (ii) for refusing absolute paths is "it would widen the Project-code execution surface past ADR-0009's Trusted project" — which the relative spelling then did anyway. The fix makes that sentence true rather than aspirational.

**Fix.** After canonicalization, refuse any address whose remainder still leads with a `..` segment. Implemented as a **first-segment** test, not a string prefix: `..foo.gd` is a legal filename and stays accepted — verified end-to-end (it runs), with unit arms so a naive `startswith("..")` cannot regress it. `../../etc/passwd` needs nothing extra but is covered by the same rule.

**Honest scoping.** Both shapes are reachable on the `res://` spelling on the **base** branch too (base `res://..` also phantom-succeeds; base `res://../outside.gd` also runs). This PR **widens reachability** — via the ordinary project-relative form — rather than creating the holes. The residual parser weakness (the trailing-period strip that mis-reads `res://..`) is **not** touched here; it belongs to the error parser and is tracked separately.

**Doc sync.** The amendment's "what stays refused" list now names the root/escape spellings and records the `res://` / `res://.` code move (surface item 11). The `ScriptRunParams` docstring — which ships into `gda schema` as the input description — was reflowed to remove an orphan short line left by the previous round; verified in the emitted schema.

**Refusals quote the raw *operation* input**, post-`NormalizedPath`, not the raw argv: `""` and `./` both report `got: '.'`. Inherent to ADR-0015 (the model is the single normalization home) and correcting an over-broad claim in an earlier revision of this body.

## External review (qinhaihong-red / codex, pinned to `fd2e6b69`) — adopted (`890254bc`)

Three axes, **one unique finding**, verified and adopted at the shared normalizer's altitude.

**The finding.** Moving `ScriptRunParams.path` to the shared `NormalizedPath` sent `~nosuchuser/script.gd` into `Path.expanduser()`, which raises `RuntimeError` → **bare traceback, exit 1**, on *both* argv and `--params-json` (what gda-mcp forwards). On the exact base, `script run` returned `invalid_path` exit 4 — the old res://-only gate refused before expansion. So for `script run` this is a **changed-path regression this PR introduced**, against ADR-0031's pre-launch structured refusal and ADR-0015's no-bare-traceback rule. Reproduced on base and head before fixing.

**Fixed at the normalizer, not with a script-run band-aid.** `normalize_path` is now **total**: an unresolvable `~user` prefix passes through unchanged. Every `NormalizedPath` consumer inherits it.

**Why pass through rather than raise** — the brief suggested a `ValueError` for pydantic to translate; I prototyped it and it is **insufficient**, which is worth recording. The argv body constructs its params model *directly*, so a `ValueError` becomes a `ValidationError` that **still escapes as a bare traceback**; only `--params-json` caught it (as `invalid_params`). Only totality repairs both channels. Normalization is also a convenience, not a validity check — validity stays with the consumer that opens the path.

**`script run` regains its exact base envelope** via its own gate, which now refuses a leading `~`. A home reference is a filesystem address form, not a project-relative one, so lifting it to `res://~user/x.gd` would repeat the `res://user:/x.gd` splice the gate already refuses. Both tilde outcomes now converge: resolvable `~/x.gd` → expands to absolute → refused as absolute; unresolvable → refused as tilde. Same code, same message.

**Envelope shapes, measured (not assumed):**

| command | code | category | exit | traceback |
| `script run` (argv **and** `--params-json`) | `invalid_path` | operation | 4 | none — **base parity** |
| `script validate` / `script get` / `script delete` / `scene get` | `path_not_found` | operation | 4 | none |
| `resource get` | `invalid_path` | operation | 4 | none |

Non-`script run` consumers differ in **code** — they reach the engine, which reports the path it cannot find, whereas `script run` refuses pre-launch at its own gate. Structured and traceback-free in every case, which is the contract; the code difference is inherent to where each command decides, and is documented rather than papered over.

**#699 subsumed.** The same crash pre-existed on base for the sibling script commands and was filed as #699. The single-home fix closes it — hence `Closes #699` above.

**Surface:** `gda schema` and `gda script run --help` are **byte-identical** to the previous head (verified); this round is runtime behavior only.

**Coverage:** a unit regression pins the `RuntimeError` conversion at the normalizer itself plus a guard that resolvable `~/…` expansion still works; a command-level test on `script validate` (both channels) proves surface-wide inheritance; `script run` gets both-channel base-parity tests; the operation-level gate table gains `~nosuchuser/x.gd` and `~/x.gd`.

## Acceptance criteria

- [x] One script-path representation works for both `script validate` and `script run` — e2e-pinned on both commands.
- [x] The result reports the canonical `res://` path — new `ScriptRunResult.path`.
- [x] The ADR-0031 scope amendment lands with the change.
- [x] Regression covers both path forms on both commands — unit (operation + CLI + `--params-json`) and e2e.

**Wave-1 verdict tests stayed green**, and the interaction is now guarded directly: a broken entry addressed project-relatively must still classify `script_compile_failed`. If the lift bypassed the canonical identity, that would silently regress to the phantom success #651 removed.

## Sibling-owned changes flagged (NOT made here)

- **SKILL.md — RESOLVED, now landed here.** Originally flagged for #659 (the wave's skill owner). #659's PR #692 has since finished and edits a disjoint region (Setup), so per the ownership-lease rule the clause lands in this PR instead — PR-body text is not shipped agent guidance, and the AC's doc surface belongs in the artifact. Landed in `43283471`; only the punctuation was adjusted from the proposal (a parenthetical rather than a third em-dash pair, which would have collided with the clause that follows). The rest of the row (non-zero `quit()` is still success, `--strict`, never-ran always fails) is unaffected. This PR's merge position is before #692's.
- **`src/gda/errors.py` (adjacent to #653's `error_codes.py`).** Only the `script_path_invalid_failure` message and docstring changed. **No new or changed error code, category or exit code** — `invalid_path` / `operation` / exit 4 throughout, so ADR-0002 and the registry are untouched. Flagging for awareness only; I do not disagree with #653's ownership.
- **`validate` (owner: #658).** Untouched. `script validate` still echoes the path spelling it was given, as every sentinel operation does — canonicalizing one operation's echo would trade this inconsistency for a different one, so it is left alone deliberately (recorded in the amendment). My only edit under #658's area is an appended e2e test in `tests/test_e2e_script.py`; expect a trivial rebase if #658 lands first.

## Out of scope

An **absolute** path stays refused. ADR-0031's ABI edge names "a non-`res://` **or absolute**" path; only the first half is lifted, matching the issue's decision text ("the project-relative and `res://` forms") and the dogfooding evidence. Recorded in the amendment rather than left implicit.


qinhaihong-red:
**Changes requested.** All three independent review axes reproduced the same public-input regression: an unresolvable named-user path escapes the structured failure boundary.

Reviewed refs:

- Base / three-dot merge-base: `c11f07db202d6bffad3de37a45159fa20bfe8165`
- Head: `fd2e6b69311fb49de49c39565f4fd3079c59dfb2`

## Domain architecture

### [P2] Keep an unresolvable `~user` path inside the structured path boundary

`src/gda/commands/script.py:548` (throw site: `src/gda/models.py:262`)

Moving `ScriptRunParams.path` to `NormalizedPath` invokes `Path.expanduser()` without handling `RuntimeError`. At the exact head, both argv and equivalent `--params-json` invocations with a nonexistent `~user/script.gd` exit `1`, emit no JSON, and expose `RuntimeError: Could not determine home directory.` The exact base returns the structured `invalid_path` envelope with exit `4` on both routes.

This violates ADR-0031's pre-launch structured-refusal contract and ADR-0015's shared argv/JSON boundary; the JSON route is also what gda-mcp forwards.

Minimum fix: make shared normalization total for unresolvable tildes, translate the failure consistently through both structured adapters, and add paired argv and `--params-json` regressions.

The remaining topology should be retained: argv and `--params-json` are inbound adapters; the `script` slice owns params/results, admission, recipe, descriptor, and canonical result; `NormalizedPath` owns generic normalization; `canonical_res_path` owns lexical `res://` identity; and `errors.py`, `dispatch_recipe`, and `runner.launch` keep distinct failure, composition, and outbound responsibilities. Dependencies remain directed and acyclic.

- `DA-693-1`: both portable spellings converge before launch. Unit/CLI evidence supports it; Godot e2e passed in the review, but the repository CI job is skipped.
- `DA-693-2`: every refused path is a structured pre-launch failure. The `~user` counterexample currently disproves the advertised scope.
- `DA-693-3`: accepted paths stay within the Trusted-project surface. Lexical root/escape tests support it; Windows-specific spellings and symlink resolution remain unverified boundaries.

## Standards

### [P1] Unexpandable tilde paths bypass the Error envelope

`src/gda/commands/script.py:548`

The traceback described above is a hard breach of ADR-0031 (“all decided before any launch as `invalid_path`”) and ADR-0015's rule that public input failures are never bare tracebacks. It can also tear down MCP-backed callers rather than returning a branchable `GdaError`.

Minimum fix: make tilde normalization total, route the unexpandable form to the existing structured failure, and cover both input channels. No additional baseline smell was retained.

## Spec

### [P2] Shared normalization makes one project-relative input crash

`src/gda/commands/script.py:548`

Issue #675 requires `script run` to accept project-relative and `res://` forms through the same normalization. The new normalization route instead raises for an unknown named user. This is a changed-path regression: the fixed base returned `invalid_path`.

Minimum fix: make `normalize_path` total for unresolved named-user expansion and assert structured behavior through argv and `--params-json`.

The remaining acceptance criteria appear complete: the ordinary forms work for validate/run, successful run results report canonical `res://`, the dated ADR amendment preserves the original text, and both commands have two-form regression coverage.

## Validation

- I independently reproduced head exit `1` with a Rich/Python traceback and base exit `4` with the structured `invalid_path` envelope.
- Full non-e2e: `1341 passed, 426 deselected`.
- Focused real-Godot 4.6.3 e2e: `14 passed`.
- Ruff check/format, Pyright, and `git diff --check` passed.
- Required CI is green; its Godot e2e job is skipped. The complete e2e suite was not rerun.

Summary: Domain architecture — 1 finding, worst P2. Standards — 1 finding, worst P1. Spec — 1 finding, worst P2.

qinhaihong-red:
Thanks — the finding is real, reproduced on both base and head, and **fully adopted** in `890254bc`. Grouping it as one issue across the three axes it was reported on.

## The finding

Moving `ScriptRunParams.path` to the shared `NormalizedPath` sent `~nosuchuser/script.gd` into `Path.expanduser()`, which raises `RuntimeError` — **bare traceback, exit 1**, on both argv and `--params-json` (the channel gda-mcp forwards). On the exact base, `script run` returned a structured `invalid_path` with exit 4, because the old `res://`-only gate refused the path before expansion.

So for `script run` this is a **changed-path regression this PR introduced**, not a pre-existing wart: it breaks ADR-0031's "pre-launch structured refusal, never a crash" edge and ADR-0015's no-bare-traceback rule. Evidence:

```
# base (c11f07db)
$ gda script run '~nosuchuser/x.gd' --project P --json
{"error":{"category":"operation","code":"invalid_path",...}}   # exit 4

# head before the fix (fd2e6b69)
RuntimeError: Could not determine home directory.              # exit 1, raw traceback
```

## Adopted at the shared normalizer, not as a script-run band-aid

The same crash already reached **every** `NormalizedPath` consumer on base — `script get`, `script validate`, `scene get` and friends all died identically (tracked as #699). Fixing it in `script run` alone would have left that untouched, so the fix went into `normalize_path` itself, which is now **total**: an unresolvable `~user` prefix passes through unchanged. This PR therefore **subsumes and closes #699**.

One correction worth recording, since the obvious fix does not work. Raising a `ValueError` for pydantic to translate is **insufficient** — I prototyped it. The argv command bodies construct their params model *directly*, so the `ValueError` becomes a `ValidationError` that **still escapes as a bare traceback**; only `--params-json` caught it (as `invalid_params`). Only making the normalizer total repairs both channels. It is also the semantically honest choice: normalization is a convenience that saves the caller a shell, not a validity check — validity belongs to whoever opens the path.

`script run` regains its **exact base envelope** through its own gate, which now refuses a leading `~`. A home reference is a filesystem address form, not a project-relative one, so lifting it into `res://~user/x.gd` would repeat the `res://user:/x.gd` splice that gate already refuses. Both tilde outcomes now converge on one refusal: a resolvable `~/x.gd` expands to an absolute path and is refused as absolute; an unresolvable one is refused as a tilde.

## Resulting envelopes (measured)

| command | code | category | exit | traceback |
| `script run` (argv **and** `--params-json`) | `invalid_path` | operation | 4 | none — base parity |
| `script validate` / `script get` / `script delete` / `scene get` | `path_not_found` | operation | 4 | none |
| `resource get` | `invalid_path` | operation | 4 | none |

Non-`script run` consumers differ in **code**: they reach the engine, which reports the path it cannot find, while `script run` refuses pre-launch at its own gate. Structured and traceback-free in every case — the code difference is inherent to where each command decides, and is documented in the PR body rather than smoothed over.

## Validation

- Unit regression pinning the `RuntimeError` conversion at the normalizer, plus a guard that resolvable `~/…` expansion still works.
- A command-level test on `script validate` (**both** channels) proving surface-wide inheritance — the argv half is the one a raising normalizer would have failed.
- `script run` both-channel base-parity tests; the operation-level gate table gains `~nosuchuser/x.gd` and `~/x.gd`.
- `pytest -m "not e2e"` 1350 passed (same under `GITHUB_ACTIONS=true`); ruff check, ruff format --check, pyright clean.
- Godot 4.6.3 e2e: `test_e2e_script_run.py` 36 passed, `test_e2e_script.py` 56 passed.
- `gda schema` and `gda script run --help` **byte-identical** to the previous head — this round is runtime behavior only.

The two remaining items from earlier rounds stay as recorded follow-ups and are deliberately not fixed here: the error parser's trailing-period strip that mis-reads `res://..`, which belongs to the parser and not to a path-form amendment.

Issue #687, opened 2026-08-16
Typed failure-path evidence on the Error envelope (ADR-0004 decision)


qinhaihong-red:
## Review-round source

Created during PR #678's review reconciliation to give the deferred failure-evidence work an authoritative owner. #651 delivered structured `ScriptError[]` diagnostics and raw status/stderr evidence on the SUCCESS path; on failure paths the evidence is prose only — `error.diagnostics` carries raw or labeled stderr/stdout text, the child's numeric exit status survives only in the message, and the parsed `ScriptError[]` is discarded (recorded in the ADR-0031 amendment and in #651's scope-reconciliation comment). #655 names the same constraint for its timeout envelope ("promoting these to structured envelope fields changes the uniform failure ABI (ADR-0004) and needs a recorded decision") but its acceptance criteria deliberately cover only the string-diagnostics form. Until this issue, no artifact owned the decision or the typed delivery.

## What to build

1. **The ADR-0004 decision**: a recorded decision (ADR-0004 amendment or a new ADR) on whether and how the uniform failure ABI carries optional TYPED evidence — e.g. an optional, operation-scoped evidence object on the `Error envelope` — without breaking the envelope's stable `category`/`code`/`message` contract or every existing consumer.
2. **If adopted, the `script run` delivery**: failure envelopes (`script_not_found`, `script_compile_failed`, `incompatible_script_type`, `script_failed`) carry the child's numeric exit status as data and the parsed `ScriptError[]` instead of discarding it; `--strict`'s labeled prose streams remain or are superseded per the decision.
3. **The #655 alignment**: the timeout envelope's elapsed seconds, termination phase, and capped streams adopt the same typed shape (#655's string-diagnostics form ships first and is not blocked by this issue).
4. If the decision DECLINES typed evidence, record why and close this issue with the prose contract affirmed; #651/#655's disclosures then stand as the permanent contract.

## Acceptance criteria

- [ ] The ADR-0004 decision is recorded (adopt or decline, with the shape and its consumers analyzed).
- [ ] If adopted: `gda script run` failure envelopes carry typed status + structured diagnostics **(delivered: `script_errors` on all four verdicts and both gda-ended runs; `exit_status` on `script_failed` only — see the scope-reconciliation note)**; schema, help, and the bundled skill are synced; regression tests pin the new shape and JSON-mode invariance for untouched codes.
- [ ] If adopted: #655's timeout envelope uses the same shape (or a follow-up on #655 is filed with the mapping).
[...]

PR #694, opened 2026-08-16
fix(gda): tell a sandbox-denied window server from an absent display (#667)

### LINKED ISSUE #667: daemon start --windowed: distinguish sandbox permission denial from an unavailable display
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-029** (2026-08-10, reconfirmed on a second package). From a filesystem-restricted macOS process, `daemon start --windowed` returned `live_windowed_unavailable`; the byte-identical command started successfully in the same desktop session once allowed outside the sandbox. Automation can misclassify a permission boundary as a machine-capability failure and silently skip required rendered QA.

**Current behavior, stated precisely:** the probe's error message already names the failing API and lists sandbox as a candidate cause ("no usable macOS window-server session (CGSessionCopyCurrentDictionary returned NULL — e.g. SSH / CI / sandbox)…") — that shipped before the dogfooded 0.8.0. What is missing is (a) any **machine-readable** probe/platform field, and (b) any classification: the CG API returns NULL with no error code, so today gda cannot tell sandbox denial from a genuinely absent window-server session.

## What to build

**Step 1 — spike:** determine whether macOS exposes any signal that separates sandbox-denied window-server access from an absent session (the current probe has only two outcomes: NULL and usable).

**Step 2 — depending on the spike:**
- If a signal exists: classify permission denial as a new `Classifier error code` (ENVIRONMENT category, exit bucket mirroring `live_windowed_unavailable`), registered in the error-code registry. It is distinct from `live_display_unavailable`, the capture-time harness code — do not overload that.
- If no signal exists: close by recording that finding — the error message and docs state that the two causes are indistinguishable on macOS and how to tell them apart operationally (retry outside the sandbox).

Either way: add a machine-readable field to the `Error envelope` naming the probe (e.g. `CGSessionCopyCurrentDictionary`) and the platform, instead of prose only; document when windowed daemon startup requires desktop-session access.

## Acceptance criteria

- [ ] The spike's finding (signal exists / does not exist) is recorded on this issue.
- [ ] The failure result carries a machine-readable probe/platform field.
- [ ] If classification is possible: a sandbox-denied probe returns the new registered code, not `live_windowed_unavailable`; otherwise the indistinguishability is documented in the error and docs.
- [ ] Docs/skill state the desktop-session requirement.

## Blocked by

None - can start immediately (the spike is the first step).

qinhaihong-red:
## Summary

`gda daemon start --windowed` reported `live_windowed_unavailable` for two different
facts about the world: no window server is reachable, and the window-server lookup was
refused outright. They need opposite reactions, so automation read a sandbox boundary
as a machine-capability gap and silently skipped rendered QA (dogfooding GDA-DF-029).

**Spike result: a usable signal exists**, so this takes the classification branch of
the issue. Its strength was **overstated in the first round and corrected on review** —
see below and the [correction comment](https://github.com/aigengame/godot-agent/issues/667#issuecomment-5310776173).

## The spike, as corrected

`CGSessionCopyCurrentDictionary` returns NULL with no error code, so it cannot say WHY.
`bootstrap_look_up` of the window-server mach service does add information — but less
than the first round claimed.

| profile (desktop host) | `com.apple.windowserver.active` | unregistered control name |
| --- | --- | --- |
| unsandboxed | `KERN_SUCCESS` | `BOOTSTRAP_UNKNOWN_SERVICE` (1102) |
| `(allow default)` + deny only that name | **`BOOTSTRAP_NOT_PRIVILEGED` (1100)** | `1102` |
| `(deny default)`, no mach-lookup | **`1100`** | **`1100`** |
| `(allow default)` + deny `file-write*` only | `KERN_SUCCESS` | — |

**What 1100 proves: "the policy refused THIS lookup" — nothing more.** macOS evaluates
a seatbelt `mach-lookup` deny per name and *before* resolving it, so a blanket-deny
profile refuses an unregistered name too (row 3). The first round inferred "the host
HAS a window server" from 1100; that is false, and would have told a display-less CI
Mac under a broad-deny profile that its host has a display, with a remediation that
cannot succeed — the same inverted-failure direction used to reject
`sessionHasGraphicAccess`.

**A control lookup characterises the confinement** without licensing an existence
claim: control `1102` → the denial is **name-specific** (sandbox signature); control
`1100` → **blanket** confinement; any other control answer → **unknown breadth**, which
claims only the target refusal (round-3: folding unreadable statuses into "blanket"
would invent a fact). All three report the same code and all three say gda cannot tell
whether a window server exists.

A **filesystem-only** restriction does not trip the probe at all (row 4), so this never
fires on unrelated confinement.

Also measured and rejected: `launchctl managername` (reports "Aqua" in every row) and
`SessionGetInfo`'s `sessionHasGraphicAccess` (stays true under a denying sandbox, but
can also be true on an auto-logged-in CI box with no reachable window server).

## What changed

- **New registered code `live_windowed_permission_denied`** — ENVIRONMENT, exit 127
  (the same bucket as its sibling), classifier-source, not GDScript-mirrored. It means
  *the window-server lookup was refused*, which holds in both control branches.
  Deliberately not overloading `live_display_unavailable`, the harness's capture-time
  code.
- **`gda.display` returns a verdict, not a reason string** — `WindowedUnavailable
  {code, reason, probe}`. Both refusal sites (the `daemon start` fail-fast and the
  daemon's authoritative launch-boundary guard) relay the code the probe decided. The
  denial probe runs only on the failure path; the control lookup runs only after a
  refusal, so a healthy desktop pays nothing.
- **ADR-0004 amendment** — one optional `probe` `{name, platform}` key on the Error
  envelope. Added as a dated note in the ADR's existing `> **Outcome/Amendment**`
  block at the top of the file (the repo's convention for that ADR); no accepted-prose
  line was modified.
- Docs: ADR-0002 registry row, `docs/command-catalog.md` lifecycle entry, and the
  desktop-session requirement + code-branch guidance in the bundled skill's live
  section.

### Commit `0b6be192`

`examples/platformer/panda-adventure/tests` is a **separate pytest root that PR CI
enforces as its own tier**, and my first head failed it: three e2e gates there imported
the renamed helper, which surfaced as a collection `ImportError`. That commit updates
those three gates plus the gADR-0007 seam description. On review, the same three files'
`_NO_DISPLAY_CODES` skip sets also gained `live_windowed_permission_denied`, so a
display-state race between the pre-check and the live call skips rather than fails.

## The envelope change is additive — proof

`emit_failure` serializes with `exclude_none`, so `probe` is **omitted, never `null`**,
for every failure that does not set one. Guarded through the **real CLI** by an ungated
test (a probe-less `binary_not_found`), not only by hand-rolled model dumps.

Schema diff (`gda schema`, all 67 commands), enumerated exhaustively:

1. `$defs.EnvironmentProbe` added;
2. `GdaError.properties.probe` added — `anyOf[EnvironmentProbe, null]`, `default: null`,
   **not** in `required`;
3. `GdaError.description` extended by one clause.

Unchanged: every existing `GdaError` property byte-identical, `required` unchanged, and
`input` / `output` / `kind` / `constraints` unchanged for all 67 commands. Exactly
**one** distinct error schema across the whole surface, so ADR-0004's "byte-for-byte the
same for every command" invariant holds. `gda daemon start --help` is byte-identical.

**Size: +15.3%** (361781 → 417257 bytes), unchanged by the relay work — `LiveError`
is an internal daemon↔CLI wire model and does not appear in the published `--schema`
surface (verified). The error schema is repeated per command, so
description text is paid ~67 times: the field descriptions added on review (F4) are
+5.4pp of that, after tightening them from a first pass that cost +9.8pp. Model
rationale is kept in comments for the same reason.

`gda-mcp` needs no change: it maps only `input`/`output` and passes the envelope through
to its `is_error` channel.

## Scope boundary with #687

#687 owns whether the failure ABI carries operation-scoped **typed evidence** (a
script's exit status, parsed `ScriptError[]`, #655's timeout fields). This amendment
does not decide that and does not pre-empt its shape: `probe` is a fixed two-string
answer to "which host call decided this environment verdict". The amendment states
explicitly that the two axes compose and that #687 stays free to adopt, reshape, or
decline typed evidence.

**The relay carries `probe` too** (round-3 review). The refusal that actually gates
every live op is the daemon's lazy-launch guard, not the CLI's optional fail-fast, so
leaving the field off the relay made the AUTHORITATIVE path the poorer one while #667's
AC is unqualified. The live envelope therefore takes the same optional key —
`LiveError` is `{code, message, probe?}` — and it is validated end to end (daemon reply
→ `classify_live` → public JSON) on the real builders.

The **cross-language contract is untouched**: the GDScript-emitted headless sentinel
keeps its own strict, probe-less model (`OperationError`, `extra="forbid"`), because a
GDScript operation has no host probe to report. Two models, one per channel. On the
wire the key is omitted when absent, so every other live reply and every headless
envelope is byte-identical.

## Test tiering

**Ungated** (every CI tier runs these; the verdict logic is a pure function of injected
host answers, so no display and no sandbox are needed):

- `tests/test_display_probe.py` — the verdict logic (capability vs both denial
  branches), the status→breadth mapping via a fake lookup, that the control lookup is
  skipped unless the target was refused, the degradation guard (probe raising falls back
  to the old verdict), and a **regression guard that no branch reasserts host-has-a-
  window-server**.
- `tests/test_models.py` — the `exclude_none` ABI guard through the real CLI, plus
  envelope round-trips pinning that the only key the convention drops is `probe`.
- `tests/test_daemon_lifecycle_payload.py` — the denied start returns the new code at
  ENVIRONMENT/127, still pre-launch and non-mutating, with the probe field.
- `tests/test_daemon_server.py` — the daemon relay reports the probe's code, wire
  envelope shape unchanged.
- `tests/test_error_registry.py` — registry + `classify_live` flow.

**Capability-gated** (skips honestly off a capable host): the real-seatbelt leg in
`test_display_probe.py`, gated on macOS + a working `sandbox-exec` + an actual desktop
session to deny access TO. It exercises **both** profiles (name-specific and blanket)
and **ran and passed** on this host.

**The gate policy itself is the round-3 P1 fix.** The e2e gates used to skip on every
no-display code, which meant a sandboxed run greened the suite with the rendered
acceptance unexecuted — the GDA-DF-029 behaviour, reintroduced on the test side. There
is now ONE owner for the reaction per context (`tests/support.py`; the game gets its own
`tests/display_gate.py`, being a separate pytest root that cannot import the toolkit's
test package):

- capability (`live_windowed_unavailable`, `live_display_unavailable`) → **skip**;
- `live_windowed_permission_denied` → **fail**, carrying the re-run remediation.

`_needs_display` became a fixture because a `skipif` can only skip. Coverage, stated
precisely (the recheck caught an earlier overclaim here):

- **pre-flight** — the repo's windowed e2e (via the `windowed_host` fixture) and all
  three game gates, through `require_windowed_host()`.
- **post-start race path** — the game's three gates via `handle_no_display_code()`, and
  the repo's windowed e2e via `assert_windowed_ok()` on exactly the branches that can
  see a display refusal: `daemon start --windowed`, the first live op, and
  `screen capture` / `screen frames`. Other assertions are deliberately not wrapped.

Both reactions are table-tested ungated (including the fall-through cases, so a real
regression still fails with the command's own output), verified against a real seatbelt
profile (confined → `Failed`, unconfined → proceeds), and demonstrated on a real root
e2e test with an injected refusal: capability → `SKIPPED`, permission → `Failed` with
the remediation.

## Gates

- `uv run pytest -m "not e2e"` — 1327 passed (also green with `GITHUB_ACTIONS=true`)
- panda-adventure tier — 480 passed
- `uv run ruff check` / `ruff format --check` / `uv run pyright` — clean
- Godot e2e on a desktop host, under the shared serial lock: `test_e2e_daemon.py` +
  `test_e2e_screen.py` — 32 passed, windowed legs actually executed (not skipped)
- End-to-end reproduction: the real CLI under a real seatbelt sandbox emits
  `live_windowed_permission_denied` with `probe:
  {"name":"bootstrap_look_up(com.apple.windowserver.active)","platform":"darwin"}`

## Merge notes

- **#653 owns `error_codes.py` / ADR-0002 this wave and merges first.** This branch adds
  a registry entry and an ADR row there. Reviewer checked the sibling hunks and they sit
  elsewhere in both files, so a clean rebase is the expected case rather than a conflict.
- **Shared files with other in-flight slices:** `src/gda/errors.py` and
  `src/gda/headless.py` are also touched by #692/#696, in disjoint hunks.
- **`SKILL.md`** is owned by #659 and also edited by #696. This edit is in the live
  section's prerequisites prose, region-disjoint from the command tables.
- README deliberately untouched — flagged rather than assumed: its `--windowed` mention
  is a one-line aside, and the requirement detail belongs with the command docs and the
  skill. Say the word if you want a short clause there.

Closes #667

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -64,6 +64,17 @@ The GDScript payload owns only `code` and `message`. `gda` owns the public
 `GdaError` wrapper: it validates that the code is registered, assigns the
 `operation` category, preserves the message, and copies stderr into diagnostics.
 
+> **Scope note (2026-08-17, #667) — the shape above is the GDScript-emitted
+> *headless* payload, and stays exactly that.** The Phase-2 **live** channel reuses
+> this envelope but is emitted by Python (the daemon and the daemon IPC client), and
+> it carries one OPTIONAL extra key: `probe`, the host-probe context behind a windowed
+> refusal (ADR-0004 amendment). The two channels validate against **different models**
+> — `OperationErrorEnvelope` (headless) stays `extra="forbid"` and probe-less, because
+> a GDScript operation has no host probe to report; `LiveErrorEnvelope` accepts the
+> optional key. The key is omitted when absent, so every payload either language
+> emitted before is byte-identical. This widens no cross-language contract: GDScript
+> neither emits nor reads `probe`.
+
 ### stderr as advisory diagnostics
 
 stderr is still **never** parsed for the success/failure *outcome* or for stable
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -222,6 +222,7 @@ operation, and parse codes the CLI assigns).
 | `live_display_unavailable` | `live` | `classifier` | `6` | A live `screen` capture ran on a headless engine session (the dummy DisplayServer cannot read pixels); start the daemon windowed with `gda daemon start --windowed` (Phase 2, #222). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 | `live_windowed_unavailable` | `environment` | `classifier` | `127` | A windowed live session (`gda daemon start --windowed`) was requested but the host has no usable DisplayServer (no on-console GUI session / no `$DISPLAY`), so the session cannot come up; refused before spawning Godot (Phase 2, #345). |
+| `live_windowed_permission_denied` | `environment` | `classifier` | `127` | A windowed live session (`gda daemon start --windowed`) was requested and the host HAS a window server, but this process is denied access to it (e.g. a sandbox); retry outside the restriction rather than treating the host as display-less (Phase 2, #667). |
 
 ## Considered options
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -222,7 +222,7 @@ operation, and parse codes the CLI assigns).
 | `live_display_unavailable` | `live` | `classifier` | `6` | A live `screen` capture ran on a headless engine session (the dummy DisplayServer cannot read pixels); start the daemon windowed with `gda daemon start --windowed` (Phase 2, #222). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 | `live_windowed_unavailable` | `environment` | `classifier` | `127` | A windowed live session (`gda daemon start --windowed`) was requested but the host has no usable DisplayServer (no on-console GUI session / no `$DISPLAY`), so the session cannot come up; refused before spawning Godot (Phase 2, #345). |
-| `live_windowed_permission_denied` | `environment` | `classifier` | `127` | A windowed live session (`gda daemon start --windowed`) was requested and the host HAS a window server, but this process is denied access to it (e.g. a sandbox); retry outside the restriction rather than treating the host as display-less (Phase 2, #667). |
+| `live_windowed_permission_denied` | `environment` | `classifier` | `127` | A windowed live session (`gda daemon start --windowed`) was requested but this process is denied the window-server lookup (e.g. a sandbox), so gda cannot tell whether the host has one; re-run outside the restriction to find out rather than recording the host as display-less (Phase 2, #667). |
 
 ## Considered options
 
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -24,6 +24,45 @@ status: accepted
 > the surrounding prose cannot drift. Like `kind`, gda-mcp ignores it (ADR-0012),
 > so it stays a strict superset — backward compatible.
 
+> **Amendment (2026-08-16, #667):** the uniform failure envelope (`GdaError`, the
+> `error` half of every command's `--schema`) gains ONE optional key — `probe`, an
+> `EnvironmentProbe` `{name, platform}` naming the host call that decided an
+> ENVIRONMENT failure `gda` resolved by probing the machine rather than by running
+> the engine (`CGSessionCopyCurrentDictionary`,
+> `bootstrap_look_up(com.apple.windowserver.active)`, `$DISPLAY / $WAYLAND_DISPLAY`).
+> Motivation: `--windowed` refusals had two causes — no window server on the host,
+> versus this process denied access to one — distinguishable only by reading English
+> prose, so automation recorded a sandbox boundary as a machine-capability gap and
+> silently skipped rendered QA (dogfooding GDA-DF-029).
+>
+> Three properties keep it additive:
+>
+> - **Optional and OMITTED, never `null`.** `emit_failure` serializes with
+>   `exclude_none`, so every failure that sets no probe emits byte-identically to
+>   before. Only the two windowed-refusal codes set one today.
+> - **The stable trio is untouched.** `category` / `code` / `message` (and
+>   `diagnostics`) keep their contract; `probe` is context ABOUT the classification,
+>   never a substitute for branching on `code`. `gda-mcp` passes the envelope through
+>   to its `is_error` channel unchanged and needs no adapter work (ADR-0012).
+> - **Schema-derived, zero per-command cost.** `error` is still the one shared
+>   `GdaErrorEnvelope` schema, identical for every command — it changed once, for all.
+>
+> **Scope boundary with #687.** #687 owns the separate, larger decision on whether the
+> failure ABI carries operation-scoped typed EVIDENCE — a script's numeric exit status,
+> parsed `ScriptError[]`, a timeout's elapsed seconds (#651, #655). This amendment
+> deliberately does NOT decide that and does not pre-empt its shape: `probe` answers
+> only "which host call decided this environment verdict", a fixed two-string context
+> with no per-operation variation. The two axes compose — an evidence key adopted by
+> #687 would sit alongside `probe` under the same "optional keys are omitted when
+> absent" rule established here — and #687 stays free to adopt, reshape, or decline
+> typed evidence on its own merits.
+>
+> **Not carried on the daemon relay.** The daemon→CLI wire envelope is ADR-0002's
+> `{code, message}` (extra keys rejected), so the launch-boundary guard inside
+> `gda-daemon` relays the code and prose but no `probe`; the field is populated where
+> the probe RAN in-process. Widening the cross-language wire contract is out of scope
+> here.
+
 ADR-0000 lists `--schema` as a core capability without defining it. We fix its
 semantics here, and deliberately scope out an overloaded interpretation.
 
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -30,16 +30,19 @@ status: accepted
 > ENVIRONMENT failure `gda` resolved by probing the machine rather than by running
 > the engine (`CGSessionCopyCurrentDictionary`,
 > `bootstrap_look_up(com.apple.windowserver.active)`, `$DISPLAY / $WAYLAND_DISPLAY`).
-> Motivation: `--windowed` refusals had two causes — no window server on the host,
-> versus this process denied access to one — distinguishable only by reading English
-> prose, so automation recorded a sandbox boundary as a machine-capability gap and
-> silently skipped rendered QA (dogfooding GDA-DF-029).
+> Motivation: `--windowed` refusals had two causes — no window server was detected
+> (`live_windowed_unavailable`), versus the window-server lookup was REFUSED
+> (`live_windowed_permission_denied`, which proves confinement, not that one exists) —
+> distinguishable only by reading English prose, so automation recorded a sandbox
+> boundary as a machine-capability gap and silently skipped rendered QA (dogfooding
+> GDA-DF-029).
 >
 > Three properties keep it additive:
 >
 > - **Optional and OMITTED, never `null`.** `emit_failure` serializes with
 >   `exclude_none`, so every failure that sets no probe emits byte-identically to
->   before. Only the two windowed-refusal codes set one today.
+>   before. Only the two windowed-refusal codes — `live_windowed_unavailable` and
+>   `live_windowed_permission_denied` — set one today.
 > - **The stable trio is untouched.** `category` / `code` / `message` (and
 >   `diagnostics`) keep their contract; `probe` is context ABOUT the classification,
 >   never a substitute for branching on `code`. `gda-mcp` passes the envelope through
@@ -57,11 +60,19 @@ status: accepted
 > absent" rule established here — and #687 stays free to adopt, reshape, or decline
 > typed evidence on its own merits.
 >
-> **Not carried on the daemon relay.** The daemon→CLI wire envelope is ADR-0002's
-> `{code, message}` (extra keys rejected), so the launch-boundary guard inside
-> `gda-daemon` relays the code and prose but no `probe`; the field is populated where
-> the probe RAN in-process. Widening the cross-language wire contract is out of scope
-> here.
+> **Carried on the daemon relay too** (revised on review, 2026-08-17). The refusal
+> that actually gates every live op is the daemon's lazy-launch guard, not the CLI's
+> optional fail-fast, so leaving `probe` off the relay made the AUTHORITATIVE path the
+> poorer one — and #667's acceptance criterion is unqualified. The live channel's
+> envelope therefore takes the same optional key: `LiveError` is `{code, message,
+> probe?}` and `classify_live` carries it into the public `GdaError`.
+>
+> The two channels keep **separate models**, which is what keeps this narrow: the
+> headless sentinel (`OperationError`, GDScript-emitted) stays strict, `extra="forbid"`
+> and probe-less, because a GDScript operation has no host probe to report and widening
+> it would invite a key the other language can never fill. On the wire the key is
+> omitted when absent, so every other live reply and every headless envelope is
+> byte-identical to before.
 
 ADR-0000 lists `--schema` as a core capability without defining it. We fix its
 semantics here, and deliberately scope out an overloaded interpretation.
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -708,12 +708,17 @@ headless is unaffected (4.4+, cross-platform).
   requires the host's desktop session — an on-console GUI login on macOS, `$DISPLAY` /
   `$WAYLAND_DISPLAY` on Linux — because a windowed Godot aborts during `DisplayServer`
   registration without one; it is checked pre-launch (#345) and refused with one of two
-  ENVIRONMENT codes (#667): `live_windowed_unavailable` when the host has no window server
-  (skip rendered QA here) and `live_windowed_permission_denied` when it has one this process
-  is denied access to, e.g. a sandbox (re-run outside the restriction). Both carry the
-  deciding host call as the envelope's `probe` `{name, platform}` (ADR-0004 amendment). On
-  macOS a sandbox that hides the window server rather than refusing the lookup is
-  indistinguishable from an absent session — re-running outside the sandbox is how to tell.
+  ENVIRONMENT codes (#667): `live_windowed_unavailable` when nothing refused the probe and no
+  session is reachable (skip rendered QA here) and `live_windowed_permission_denied` when the
+  window-server lookup itself was refused, e.g. a sandbox (re-run outside the restriction).
+  The second code does NOT mean the host has a window server — macOS refuses the lookup
+  before resolving the name, so a broadly-confined process is refused whether or not one
+  exists; it means only that gda was not allowed to ask, and re-running outside the
+  restriction is what settles it. Conversely a sandbox that hides the window server rather
+  than refusing the lookup is indistinguishable from an absent session. A refusal that
+  originates in `daemon start --windowed` carries the deciding host call as the envelope's
+  `probe` `{name, platform}` (ADR-0004 amendment); the daemon-relayed form carries the code
+  and prose only, since the daemon→CLI wire envelope has no `probe` key.
 
 Out of scope (editor context, ADR-0017): UndoRedo-aware mutation, the editor's
 open-scene tree, saving the open scene, editor errors/screenshots, run-editor-script,
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -704,7 +704,16 @@ headless is unaffected (4.4+, cross-platform).
   and crash-survivable like `diag` (ADR-0022). The opt-in rich `gda_log()` protocol layers on in a
   follow-up slice (#282).
 - **lifecycle (the `daemon` command group):** `gda daemon start` / `stop` / `status`, and `gda daemon
-  install` / `uninstall` for the `gda harness` (ADR-0018).
+  install` / `uninstall` for the `gda harness` (ADR-0018). `daemon start --windowed` additionally
+  requires the host's desktop session — an on-console GUI login on macOS, `$DISPLAY` /
+  `$WAYLAND_DISPLAY` on Linux — because a windowed Godot aborts during `DisplayServer`
+  registration without one; it is checked pre-launch (#345) and refused with one of two
+  ENVIRONMENT codes (#667): `live_windowed_unavailable` when the host has no window server
+  (skip rendered QA here) and `live_windowed_permission_denied` when it has one this process
+  is denied access to, e.g. a sandbox (re-run outside the restriction). Both carry the
+  deciding host call as the envelope's `probe` `{name, platform}` (ADR-0004 amendment). On
+  macOS a sandbox that hides the window server rather than refusing the lookup is
+  indistinguishable from an absent session — re-running outside the sandbox is how to tell.
 
 Out of scope (editor context, ADR-0017): UndoRedo-aware mutation, the editor's
 open-scene tree, saving the open scene, editor errors/screenshots, run-editor-script,
--- a/examples/platformer/panda-adventure/docs/gadr/0007-visual-smoke-seam.md
+++ b/examples/platformer/panda-adventure/docs/gadr/0007-visual-smoke-seam.md
@@ -33,12 +33,17 @@ The model:
   when a future slice adds a player-visible feature, it adds a checkpoint (a
   capture and/or a check) to this scenario — never a new windowed session.
 - **Display-gated: a desktop tier, not a CI gate.** Gated exactly like
-  `test_e2e_screenshot.py`: `gda.display.windowed_unavailable()`
-  pre-checks the window server, and the daemon's typed no-display refusals
-  (`live_windowed_unavailable` / `live_windowed_permission_denied` /
-  `live_display_unavailable`, #345, #667) are
-  honored as skips — visible under `-rs`, never silent. CI keeps the headless
-  seams; the visual-smoke seam runs pre-merge on a real desktop.
+  `test_e2e_screenshot.py`, through the one shared policy in `tests/display_gate.py`
+  (pre-check and post-start race path alike, #345, #667). The reaction depends on
+  WHICH refusal, because they are not the same fact:
+  - **capability** (`live_windowed_unavailable` / `live_display_unavailable`) — the
+    host cannot show a window, so the tier **skips**, visibly under `-rs`, never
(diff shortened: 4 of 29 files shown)


qinhaihong-red:
**Changes requested.** The new verdict and probe model are useful, but downstream policy still collapses permission denial into a capability skip, and the authoritative lazy-launch path drops the required probe.

Reviewed refs:

- Head: `fd297119816d967288768d31f5e41138bcd920b8`

## Domain architecture

### [P1] Windowed gates collapse the permission verdict back into a capability skip

`tests/test_e2e_screen.py:64-69`; `tests/test_e2e_daemon.py:973-975`; the three Panda Adventure windowed e2e modules and `examples/platformer/panda-adventure/docs/gadr/0007-visual-smoke-seam.md:35-41`

`src/gda/display.py` and the shipped Skill define opposite reactions: `live_windowed_unavailable` may skip rendered QA, while `live_windowed_permission_denied` requires retry outside the restriction. The downstream callers erase that distinction: every precheck verdict is skipped, and the Panda gates put the permission code in `_NO_DISPLAY_CODES`.

Impact: a sandboxed run still greens the suite with rendered acceptance unexecuted—the exact GDA-DF-029 behavior #667 exists to remove. Showing the skip under `-rs` does not restore the required reaction.

Minimum fix: centralize test-side reaction policy per context; skip only the capability verdict and fail with the retry remediation for `live_windowed_permission_denied`. Apply it to both the initial probe and post-start race paths, reconcile gADR-0007, and table-test both reactions.

### [P2] The authoritative lazy-launch boundary drops the machine-readable probe

`src/gda/commands/daemon.py:535-542`; `src/gda/daemon/server.py:181-199`

The CLI fail-fast preserves the complete `WindowedUnavailable`, but the authoritative daemon boundary deliberately strips `probe` and relays only code/prose. That makes the same verdict produce different public evidence depending on which pre-launch boundary observes it.

This conflicts with #667's acceptance criterion and ADR-0004's claim that the host call which decided an environment failure is carried in `probe`.

Minimum fix: add a typed live-relay error payload that can carry `EnvironmentProbe`, preserve it through `classify_live` into `make_failure`, and keep the headless operation-owned `{code,message}` envelope strict. Validate the complete daemon-reply → live-classifier → public JSON path.

### [P3] Changed executable-spec comments retain the falsified host-existence model

`tests/test_display_probe.py:3-6`; `src/gda/daemon/session.py:58`; `src/gda/daemon/server.py:65-69,267-274`

These comments still describe a permission denial as inability to reach a DisplayServer “that exists,” or describe only `live_windowed_unavailable`. The corrected issue authority says the refusal proves confinement, not host existence.

Minimum fix: state only that the lookup was refused and host existence remains unknown; name both relay codes.

The intended architecture should otherwise be retained: `gda.display` owns probing and verdict classification; CLI start is an optional fail-fast; gda-daemon owns the authoritative session-launch guard; `WindowedDisplayUnavailable` transports the verdict without re-deciding it; and Panda Adventure remains a downstream consumer in its separate context.

- `DA-694-1`: denial proves confinement, not host existence. Pure target/control mapping supports it; real unrestricted seatbelt comparison was not available in this review.
- `DA-694-2`: optional `probe` is additive. Shared-model and `exclude_none` tests support untouched-envelope compatibility.
- `DA-694-3`: one verdict has one downstream meaning. Findings 1 and 2 currently disprove this across the complete runtime path.

## Standards

### [P1] Rendered-QA callers collapse the new verdict distinction

The caller locations above unconditionally skip every `WindowedUnavailable`, conflicting with `docs/command-catalog.md:711-717` and `src/gda/skill/SKILL.md:96-105`, which assign opposite reactions. Minimum fix: skip only capability/capture-unavailable codes and make permission denial a rerun-required failure.

### [P3] Relay documentation still describes only the old capability verdict

`src/gda/daemon/session.py:58`; `src/gda/daemon/server.py:65-69,267-274`

The exception, seam, and launch-boundary comments still describe only “no usable DisplayServer” / `live_windowed_unavailable`. This violates the cross-artifact synchronization rule in `RULES.md`.

### [P3] Possible Duplicated Code

The same `_NO_DISPLAY_CODES` set and unconditional skip behavior recur across the three Panda gates, with the same shape repeated in root e2e tests. This duplication helped the incorrect semantic migration spread. A test-only gate helper with table-tested reactions would give the rule one owner.

## Spec

### [P1] Permission denial is still treated as a no-display skip

Issue #667 says automation misclassified a permission boundary as a capability failure and silently skipped rendered QA. These exact callers still skip both cases. Fail with the rerun-outside-restriction instruction on `live_windowed_permission_denied`; skip only actual capability absence.

### [P2] The authoritative daemon result drops required probe data

`src/gda/daemon/server.py:190-199`; pinned by `tests/test_daemon_server.py:254-282` and `tests/test_error_registry.py:178-205`

The acceptance criterion requires a machine-readable probe/platform field. A refusal detected at the authoritative lazy-launch guard ends with `error.probe == None`. Carry optional probe data through the daemon error payload and `classify_live`, then assert it on the real public envelope.

### [P3] Unexpected control statuses are reported as blanket confinement

`src/gda/display.py:255-259`

The corrected spike licenses control `1102` as name-specific denial and control `1100` as blanket confinement. The current `else` maps every other status—including success or an unexpected bootstrap error—to blanket confinement. A focused probe with target `1100` and control `5` reproduced `blanket`.

Minimum fix: match `1100` explicitly and represent all other control outcomes as unknown breadth; add cases for control `0` and an arbitrary error.

### [P3] The falsified existence claim remains in changed prose

`docs/adr/0004-schema-flag-self-description.md:33-35`; `tests/test_display_probe.py:3-6`

Phrases such as “denied access to one” and “the one that exists” still assert that a WindowServer exists. Replace them with the narrower, supported fact: this lookup was refused.

## Validation

- Exact-head focused tests: `160 passed, 2 skipped`; both skips were real seatbelt-profile legs.
- On the confined macOS review process, the probe returned `live_windowed_permission_denied`; all three Panda windowed tests reproduced as skips.
- `git diff --check`, targeted Ruff check, and Ruff format check passed.
- Required CI is green, but its Godot e2e job is skipped.
- The full suite and an unrestricted real-desktop seatbelt run were not independently rerun.

Summary: Domain architecture — 3 findings, worst P1. Standards — 3 findings, worst P1. Spec — 4 findings, worst P1.

qinhaihong-red:
## Round-3 review: all ten findings adopted — `66ac9d97`

Thanks — the P1 is the one that matters most, and it is a genuine self-inflicted
regression: the fix for GDA-DF-029 reintroduced the same defect one layer down, on the
test side. Mapping below by group.

> **Orchestrator note:** Group A **reverses an earlier orchestrator instruction** to me.
> After round 2 I was told to add `live_windowed_permission_denied` to the game's
> `_NO_DISPLAY_CODES` skip sets, on a display-state-race argument. That argument only
> holds for a capability *disappearing* between the pre-check and the live call; a
> permission verdict says the whole run is confined, which cannot be raced into
> passing. The instruction is withdrawn and this round implements the opposite.

### Group A (P1) — the gates collapsed the distinction back into a skip

**Adopted in full.** Confirmed: `tests/test_e2e_screen.py`, `tests/test_e2e_daemon.py`
and the three game gates all skipped on every no-display code, so a sandboxed run
reported a green suite with the rendered acceptance never executed. The docs promised
opposite reactions and the tests delivered one.

Fixed with a centralized policy — one owner per context, because the duplication across
five call sites is exactly how the wrong reaction spread:

- `tests/support.py`: `WINDOWED_CAPABILITY_CODES`, `WINDOWED_PERMISSION_DENIED_CODE`,
  `handle_no_display_code(code, detail)` (post-start race path) and
  `require_windowed_host()` (pre-flight), plus a `windowed_host` fixture in
  `tests/conftest.py`.
- `examples/platformer/panda-adventure/tests/display_gate.py`: the game's own copy. It
  is a separate pytest root and cannot import the toolkit's test-support package, so
  the policy is restated rather than shared, with a note in both that they must stay in
  step. (Put in its own module rather than `conftest.py`, so tests are not importing a
  conftest.)

Reactions: capability (`live_windowed_unavailable`, `live_display_unavailable`) → skip;
`live_windowed_permission_denied` → **fail**, carrying "Rendered acceptance did NOT
execute … Re-run outside the sandbox/restriction". Applied to **both** the initial probe
and the post-start race paths, everywhere. `_needs_display` became
`pytest.mark.usefixtures("windowed_host")` because a `skipif` can only skip.
gADR-0007's "honored as skips" bullet is reconciled to the split.

**Validation:** both reactions table-tested ungated (`test_the_no_display_reaction_policy`,
`test_the_preflight_gate_applies_the_same_policy`), plus a real-seatbelt check —
confined → `Failed` with the remediation, unconfined → proceeds.

### Group B (P2) — the authoritative boundary dropped `probe`

**Adopted; disclosure was indeed not delivery.** The lazy-launch guard is the refusal
that gates every live op, so reporting less there than at the optional CLI fail-fast had
it backwards.

The live wire envelope now carries the optional key: `error_envelope(code, message,
probe=None)` omits it when absent, `error_reply(..., probe=...)` passes it, and a new
`LiveError` / `LiveErrorEnvelope` pair carries it through `classify_live` into
`make_failure`. The **headless GDScript sentinel is untouched** — `OperationError` stays
`extra="forbid"` and probe-less, because a GDScript operation has no host probe to
report, so this widens no cross-language contract.

Guards extended deliberately, not fought:
`test_windowed_denied_relays_the_permission_code_not_the_capability_one` now asserts
`{code, message, probe}` with the value; a new sibling pins that a probe-less relay
keeps the narrow `{code, message}`; `tests/test_sentinel_result.py` gains an
omitted-when-absent guard and a two-model guard (headless rejects the key, live accepts
it). `test_a_relayed_windowed_refusal_carries_probe_to_the_public_json` walks the whole
daemon-reply → classifier → emitted-JSON path on the real builders.

Docs: the ADR-0004 amendment's wire-limit paragraph is **rewritten** (it was stale the
moment this landed), and ADR-0002 gains a dated scope note recording that its
`{code, message}` shape is the GDScript headless payload and stays exactly that, with
the live channel validating against a different model.

### Group C (P3) — the falsified model survived the site-list sweep

**Adopted.** Swept by CLAIM this time (grep across `src/`, `tests/`, `docs/`,
`examples/`), which is the right unit — my round-2 sweep was scoped to an enumerated
site list, which is why these survived. Remaining hits fixed:

- `tests/test_display_probe.py` module docstring ("may not reach the one that exists");
- `docs/adr/0004` amendment motivation ("denied access to one") — now names **both**
  codes explicitly, as does the "two windowed-refusal codes" bullet;
- three fabricated test messages that implied existence
  (`test_error_registry.py`, `test_models.py`, `test_daemon_lifecycle_payload.py`).

`src/gda/daemon/session.py` and `src/gda/daemon/server.py` were already fixed in
`fd297119`. Every surviving grep hit is now either a negation ("cannot tell", "proves
nothing about") or the wording test's forbidden-phrase list. That ungated wording guard
already covers the message text; extending it to sweep doc files was considered and left
out as a grep-over-prose test with a poor cost/benefit ratio — say the word if you want it.

### Group D (P3) — unexpected control statuses were read as blanket

**Adopted.** `_classify_denial` now matches `1100` explicitly for blanket and `1102` for
name-specific; everything else (control succeeded, arbitrary error) returns a third
`unknown-breadth` verdict whose message claims only the target refusal. Table cases
added for control `0` and control `5`, and the no-existence-claim guard is parametrized
over all three breadths.

### Gates (`66ac9d97`)

`pytest -m "not e2e"` 1343 passed (and with `GITHUB_ACTIONS=true`); panda tier 480
passed; ruff check + `format --check` clean; pyright 0 errors; `test_display_probe.py`
31 passed including the Group-D cases and both real-seatbelt legs; `test_e2e_daemon.py`
+ `test_e2e_screen.py` 32 passed foreground under the shared serial lock, on this
unconfined desktop host — where Group A requires them to pass rather than skip.

Schema delta re-derived: unchanged at **+15.3%**, three structural items
(`$defs.EnvironmentProbe`, the optional `probe` property not in `required`, one
description clause), all existing properties byte-identical, one distinct error schema
surface-wide. `LiveError` is an internal daemon↔CLI wire model and does **not** appear
in the published `--schema` surface (verified). `gda daemon start --help` byte-identical.

qinhaihong-red:
**Recheck: the false-green P1 is fixed, but one P2 and a small prose-sync residue remain.** Permission denial now fails instead of silently skipping rendered acceptance, and the probe relay / denial-breadth fixes are sound. The root e2e post-start path, however, does not use the new reaction policy that the reply says is applied “everywhere.”

Reviewed refs:

- Base tip: `c11f07db202d6bffad3de37a45159fa20bfe8165`
- Head: `a2d1fb054410d6dae2a06bd8b49fe9c2a2d32279`

### [P2] Wire the root e2e post-start race paths into the shared reaction policy

`tests/support.py:741-754`; `tests/test_e2e_screen.py:101-106,131-155`; `tests/test_e2e_daemon.py:1283-1303`

`handle_no_display_code()` correctly implements the required split: capability codes skip, while `live_windowed_permission_denied` fails with the re-run remediation. Panda Adventure calls it at its post-start failure sites. In the root test context, though, the helper is called only by its unit test; no root e2e calls it. The root windowed tests still assert `returncode == 0` directly after `daemon start`, the first lazy live op, and capture/frames.

Therefore, if display capability disappears between the precheck and lazy launch, the same capability verdict skips in Panda Adventure but fails in the root e2e tier. Permission denial does fail now, so the original false-green P1 is genuinely closed; the remaining behavior is a false-red and still disproves the “one verdict has one downstream meaning” claim. It also contradicts the reply's statement that the policy is applied to both pre-flight and post-start race paths everywhere.

**Minimum fix:** on the root windowed e2e start / first-live-op / capture failure branches, parse the structured error code and call the existing `handle_no_display_code()` before the ordinary assertion. Keep unknown failures as assertions and permission denial as a loud failure.

### [P3] Reconcile comments with the split reaction and third denial breadth

- `examples/platformer/panda-adventure/tests/test_e2e_screenshot.py:95-107` still says a slipped display refusal is “a skip … not a failure.”
- `examples/platformer/panda-adventure/tests/test_visual_smoke_e2e.py:482-484` still says “skip, not fail.”
- `src/gda/display.py:246-255` documents two denial branches and says “Neither branch,” although `_classify_denial()` now has the third `unknown-breadth` branch.

The code has the intended three-way behavior; these changed executable-spec comments should name it accurately.

The other groups are closed:

- **Probe relay:** the authoritative daemon refusal carries `EnvironmentProbe` through `LiveErrorEnvelope` → `classify_live` → public JSON, while the headless sentinel remains strict and probe-less.
- **Existence wording:** the falsified “denial proves a WindowServer exists” claim is gone; remaining hits are correct negations/explanations.
- **Unexpected control status:** independently reproduced as `1102 → name-specific`, `1100 → blanket`, and `0/5 → unknown-breadth`.

Validation at the immutable head: `185 passed, 2 skipped` in the focused root suites; Panda Adventure non-e2e `534 passed, 1 skipped, 19 deselected`; focused Ruff and `git diff --check` passed. I also independently reproduced probe preservation through the live classifier. Required CI is green; GitHub's Godot e2e job is skipped, and I did not independently rerun the real-Godot/seatbelt legs.

qinhaihong-red:
## Recheck: both residues adopted — `88bf5b42`

### P2 — the root e2e post-start branches never called the policy

**Adopted, and the overclaim was mine.** My round-3 reply said the policy was "applied
to both paths everywhere"; it was not. Only the game's post-start sites called
`handle_no_display_code()`. The repo's own windowed e2e still asserted
`returncode == 0` straight after the windowed start and the capture, so a capability
refusal arriving mid-run **skipped in the game tier and turned into a false RED in the
root tier** — one verdict, two meanings, which is the same class of defect as the
false-green, just inverted.

Fix, kept to the reviewer's minimum. `tests/support.py` gains `assert_windowed_ok(result)`
next to the rest of the policy:

```python
if result.returncode != 0:
    handle_no_display_code(gda_error_code(result.stdout))
assert result.returncode == 0, result.stdout + result.stderr
```

So a display refusal is classified first (capability → skip, permission → loud fail with
the remediation) and everything else falls through to the ordinary assertion, carrying
the command's own output. **Three branch groups wired, nothing else:**

| file | branches |
| --- | --- |
| `tests/test_e2e_screen.py` | `daemon start --windowed` + `screen capture` (single-frame test); start + `capture --inline`; start + `screen frames` |
| `tests/test_e2e_daemon.py` | `daemon start --windowed` + `screen capture` in the paused-capture test |

**Demonstrated, not just reasoned** — a throwaway `-p` plugin makes
`daemon start --windowed` return an injected refusal while the pre-flight probe still
runs for real on this desktop, so it exercises precisely the new branch, on a real root
e2e test (`test_windowed_daemon_captures_a_single_viewport_frame`):

- injected `live_display_unavailable` →
  `SKIPPED [1] tests/support.py:754: windowed session unavailable in this environment (live_display_unavailable)`
  — this is the case that previously produced a false RED.
- injected `live_windowed_permission_denied` →
  `Failed: the windowed session was refused because this RUN is confined … Rendered
  acceptance did NOT execute … Re-run outside the sandbox/restriction.`

Also tabled ungated (`test_the_post_start_assertion_applies_the_policy_before_asserting`)
over six cases including `operation_failed` and unparseable output, so the fall-through
is pinned and the helper cannot start swallowing real regressions.

### P3 — three comments lagging the behaviour

**Adopted.** All three described semantics that no longer hold:

- `examples/.../tests/test_e2e_screenshot.py` — both race-path comments (the start
  branch and the capture branch) said a slipped refusal is honored "as a skip … not a
  failure"; they now state the split, naming `live_windowed_permission_denied` as the
  loud-fail case.
- `examples/.../tests/test_visual_smoke_e2e.py` — same correction on the first-live-op
  comment.
- `src/gda/display.py` — `_macos_window_server_denial`'s docstring still said "Neither
  branch licenses a claim about the host's display" after `_classify_denial` grew a
  third; it now lists the unknown-breadth branch and says "None of the three".

### Gates (`88bf5b42`)

`pytest -m "not e2e"` 1349 passed (and with `GITHUB_ACTIONS=true`); panda tier 480
passed; ruff check + `format --check` clean; pyright 0 errors; `test_display_probe.py`
37 passed including the new post-start table and both real-seatbelt legs;
`test_e2e_daemon.py` + `test_e2e_screen.py` **32 passed, exit 0** foreground under the
shared serial lock on this unconfined desktop — the new branches only reroute failure
paths, so the green path is unchanged.

qinhaihong-red:
**Second recheck: the helper and almost every named branch are fixed, but one original root call site still bypasses the policy.** The screen start/capture/frames paths and the paused-capture start/capture path now have one verdict → one reaction; the three stale comments are also corrected. `test_daemon_status_surfaces_the_windowed_display_mode`, however, still directly asserts the windowed start result after its precheck.

Reviewed refs:

- Base tip: `c11f07db202d6bffad3de37a45159fa20bfe8165`
- Head: `88bf5b42a37663737449c5bc122e07a0c76e1a92`

### [P2] Route the daemon-status windowed start through `assert_windowed_ok()` too

`tests/test_e2e_daemon.py:967-975`

The test currently remains:

```python
require_windowed_host()
assert run("daemon", "start", "--windowed").returncode == 0
```

The test-side precheck and the CLI's own `daemon start --windowed` check are two separate observations. If display capability disappears between them, the command returns the structured capability refusal, but this direct assertion still turns it into a false red. The same verdict now skips in the repaired screen / paused-capture / Panda branches.

This is the same reaction-policy gap, not new scope: `tests/test_e2e_daemon.py:973-975` was one of the original finding's explicit locations. The previous recheck fix covered the later paused-capture test in this module but missed the daemon-status test.

**Minimum fix:** import `assert_windowed_ok` beside `require_windowed_host` and replace the direct assertion with:

```python
assert_windowed_ok(run("daemon", "start", "--windowed"))
```

No new helper or test mechanism is needed; the existing six-case table already pins capability skip, permission failure, and ordinary-error fall-through.

Everything else from the recheck is closed:

- `assert_windowed_ok()` applies the correct capability / permission / ordinary-failure split.
- All three `test_e2e_screen.py` windowed start + capture/frames groups use it.
- The paused-capture test uses it for windowed start and capture.
- Both Panda race-path comments now describe the split reaction.
- `_macos_window_server_denial()` now documents all three breadths and says “None of the three.”

Validation at immutable head `88bf5b42a37663737449c5bc122e07a0c76e1a92`: `35 passed, 2 skipped` in `test_display_probe.py`; both root e2e modules collect all `32` tests; focused Ruff and `git diff --check` passed. Required CI is green; GitHub's Godot e2e job is skipped, and I did not independently run the real-Godot e2e tier.

qinhaihong-red:
Second-recheck response — the last bypass is closed on `e5cbcc37` (handled directly by the orchestrator).

| [P2] `test_daemon_status_surfaces_the_windowed_display_mode` still asserted the windowed start's bare returncode after its precheck — a between-observations capability slip became a false red at one of the original finding's named sites (the prior round repaired only the later paused-capture test in this module). | Fully adopt (the minimum fix as suggested) | Confirmed the site was the one location left outside the policy: `grep` shows every other root windowed start/capture/frames branch routes through `assert_windowed_ok()`. | The direct assertion is replaced with `assert_windowed_ok(run("daemon", "start", "--windowed"))`, with a comment naming the two-observations race and the split reaction. No new mechanism — the existing six-case table already pins capability skip, permission failure, and ordinary-error fall-through. | `tests/test_e2e_daemon.py -m e2e` → 27 passed under the compliant shared lock on this unconfined desktop; full fast suite 1349 passed; ruff check/format clean. |

Issue #663, opened 2026-08-15
script validate: batch/project-wide mode with one aggregate verdict


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-008** (2026-08-08, reconfirmed twice — four scripts/four launches, then six scripts/six launches) — one invocation accepts exactly one `.gd` path; validating six related scripts costs six serialized engine launches. Omitting PATH is a text-only usage error; `res://` is `invalid_path`.
- **GDA-DF-041** (2026-08-13, capability half) — there is no project-wide static-analysis command (`gda analyze` is unknown). The no-redirect half of that entry (the unknown command offers no hint) is owned by **#670**, whose structured-hint mechanism also serves the `analyze` spelling.

## What to build

Two modes, named consistently:

- **Batch mode (required):** accept repeated PATH arguments on `script validate`; a single engine launch validates the whole batch and returns per-file diagnostics plus one aggregate verdict in one structured result. This is the form the evidence demands (production packages change 4–6 related scripts together).
- **Project mode (optional extension, own AC):** an `--all` (or directory) form that validates every project script the same way.
- ADR-0006 constraint: every path resolves against the one resolved project; a batch whose paths span projects is refused (this is exactly the hazard ADR-0006's rejection rationale names — cross-reference #658).

## Acceptance criteria

- [ ] Batch mode: one invocation validates N scripts; a single engine launch performs the batch; the result carries per-file diagnostics and one aggregate verdict.
- [ ] The aggregate `valid` is false when any file is invalid; the command still exits 0, per the existing validate contract (an invalid script is a successful validation operation).
- [ ] A batch whose paths resolve outside the one resolved project is refused with a structured diagnostic.
- [ ] Project mode, if delivered: `--all` validates every project script with the same result shape.
- [ ] Schema/help documents batch mode (and project mode if delivered).

## Blocked by

None - can start immediately. Related: #670 (owns near-miss hints, including `analyze`), #658 (project resolution).

qinhaihong-red:
> *This was generated by AI during triage.*

Triage: moved to `ready-for-agent`. The issue body (revised 2026-08-15 after a full artifact review against the dogfooding source, the ADRs, and the current code) is the authoritative agent brief — current behavior, deliverables, acceptance criteria, and scope boundaries are specified there.

PR #695, opened 2026-08-16
fix(gda): refuse a script validate outside the resolved project and report the root

### LINKED ISSUE #658: script validate: resolve the owning project or refuse when the script is outside the selected root

qinhaihong-red:
Closes #658

## What changed

`gda script validate` compiled a script against the resolved project even when
that project does not own it, so every `res://` dependency resolved against the
wrong root. The engine answered with a cascade of missing-file errors plus the
type errors *derived* from them, on a script that is perfectly valid in its own
project (GDA-DF-035) — the one project-context mistake buried under the noise it
caused.

Two changes, both decided at the CLI because that is where ADR-0006's resolved
project lives (the engine is *told* the project through `--path`, never asked):

1. **Refusal before parsing.** A script outside the resolved project is refused
   with the existing `project_not_found` code, naming both the file's location
   and the resolved project. No engine is spawned, so the false cascade is never
   produced.
2. **`project_root` on the success result.** Every successful validate reports
   the project it compiled against (`null` when projectless), so a reader can
   attribute a `valid=false` verdict to its root instead of inferring it.

## What this does NOT fix — stated plainly

**The exact GDA-DF-035 invocation still produces the false cascade on this head.
This PR makes it attributable (`project_root`), not absent.** Running
`gda script validate examples/<game>/<script>.gd` from a repository root that is
not itself a Godot project resolves no project, so there is nothing for the
target to be "outside" of; the engine still resolves `res://` against its cwd and
still reports the missing dependencies. What changed is that the result now says
`"project_root": null`, which is the reading the dogfooding session lacked.

The same holds for a script in a project **nested under** the resolved one: it is
*contained*, so it is not refused. Closing either case requires deriving the
project from the target path — exactly what ADR-0006 rejected — so it needs an
amendment, not an implementation choice here.

Follow-up: **#697** (ADR-0006 amendment: derive-vs-refuse for a target under a
different `project.godot`) carries both reproduced boundaries, the
no-escape-hatch consequence, the `target_outside_project` registry question, and
the README doc note. Both non-refusal boundaries are pinned by tests here
(`test_a_project_nested_inside_the_resolved_one_is_not_refused`,
`test_script_validate_verdict_is_root_dependent_and_names_the_root`) so they stay
deliberate rather than accidental.

## Acceptance criteria → deliverable

| AC (issue #658) | Delivered by | Verified by |
|---|---|---|
| Refuses with a structured diagnostic naming the mismatch; never false dependency errors | `script_outside_project_failure` + the containment check in `_script_validate_recipe`, run **before** any engine spawn | `test_script_validate_refuses_a_script_outside_the_resolved_project` (unit: asserts `fake.calls == []`; e2e: asserts no diagnostic and empty `diagnostics`) |
| Success result reports the resolved project root in `project_root` | `ScriptValidateResult.project_root`, stamped by the recipe | `test_script_validate_reports_the_resolved_project_root`, `…_null_project_root_when_projectless`, e2e `…_verdict_is_root_dependent_and_names_the_root` |
| Regression covers a script under a nested project validated from an ancestor directory | e2e `test_script_validate_verdict_is_root_dependent_and_names_the_root` — same script, two roots, opposite verdicts, each labelled by `project_root` | real engine |

## Error code: reused, not minted

`project_not_found` (existing, `operation` category, exit 4). Semantic match: the
failure is that **no project usable for this target was resolved**, and the
remedy is the project context (`--project`) — the same class of mistake, and the
same agent branch, as `script run`'s projectless ABI edge, which already uses
this code. Sibling #663's batch-mode AC uses the same "outside the one resolved
project" semantics.

The registry description and the ADR-0002 row enumerated three producers, and
this refusal is a contradicting fourth (a project *was* resolved and *is* valid),
so both texts now name it — no new code, no new row, no exit-code change. The
structural pin in `tests/test_error_registry.py`
(`test_adr_registry_matches_python_authoritative_registry`) is green; note it
compares only `(category, source, exit_code)`, so the description column is
**unpinned** — the two description texts were verified to agree by inspection
(word-for-word, differing only in the ADR table's markdown backticks).

## Containment: how a target is judged

Three revisions came out of review; the rule now is:

1. An **engine-virtual** path (`res://`, `user://`, `uid://`, matched by PREFIX) is
   inside by construction — gda makes no filesystem claim about one.
2. A filesystem path is **anchored where the engine anchors it**
   (`project_anchored`): relative at the *resolved project*, absolute untouched.
   Launched with `--path game`, the engine opens `game/deck.gd` for the input
   `deck.gd` no matter where gda was invoked, and the README promises the same.
3. The anchored candidate is then read up to two ways, and belongs to the project
   when **either** says so:
   - **Resolved** (`Path.resolve()`, symlinks followed) — "are these two spellings
     the same place?" (`/tmp` ↔ `/private/tmp`). Always consulted.
   - **Lexical** (`os.path.abspath`, symlinks preserved) — "did the caller address
     this through the project's own tree?" This is what makes the monorepo
     shared-addon layout (`game/addons/lib -> ../../libs/lib`) work: Godot follows
     the same link, so the file really is in the project's `res://` namespace.
     Consulted **only** when neither the candidate nor the project spelling carries
     a `..` component — see below.

**Why the `..` guard.** A `..` that follows a symlink collapses textually to a
place the filesystem never visits: with `game/pivot -> ../outside/deep`, the input
`game/pivot/../deck.gd` reads as `game/deck.gd` lexically while really naming
`outside/deck.gd`. Trusting the lexical reading there accepted a genuinely outside
target and the engine compiled it. Without a `..`, a symlink can only redirect
*downward* from a path that does start inside the project — the legitimate case
above — so the guard costs nothing there. When it withholds the lexical reading,
containment falls back to the resolved reading alone: a symlinked-in file named
*with* a `..` is refused, and the caller can name it without the `..`.

## Public-surface diff (complete, byte-diffed before/after)

`gda schema`: **one** command entry changes, `script validate` — everything else
is byte-identical (`kind` stays `headless`; `input`, `error` and `constraints`
unchanged). The registry/ADR prose above is **not** in the manifest; re-verified
byte-identical after that edit.

1. `script validate` manifest `description` — gains the two project-context
   paragraphs (the command docstring is the manifest description).
2. `ScriptValidateResult` schema `description` — gains one paragraph about
   `project_root`.
3. `ScriptValidateResult.properties.project_root` — **new** property,
   `{"anyOf": [{"type":"string"},{"type":"null"}]}`, appended after `diagnostics`
   so the existing property order is untouched.
4. `ScriptValidateResult.required` — gains `project_root`:
   `["path", "valid"]` → `["path", "valid", "project_root"]`. Required **and**
   nullable, because every emitted verdict carries the key (`null` = projectless),
   so an agent reads it unconditionally instead of probing for it. It is also
   reported as an **absolute** path now; `--project game` used to emit the bare
   relative `"game"`.

`gda script validate --help`: gains exactly the same two paragraphs (no option,
argument, or first-line change).

`gda skill`: **one** hunk, in the existing `script validate` verdict paragraph —
tells the agent to read `project_root` before acting on a `valid=false`, since a
verdict full of missing-`res://` errors usually means the wrong project, not a
broken script (`null` = projectless).

Human-readable output: an **invalid** verdict gains one leading line,
`  project: <root>` (or `  project: (none resolved: projectless)`), placed
*before* the diagnostics — when the root is wrong, every line under it is an
artefact of that one mistake. A `valid` verdict is unchanged.

`README.md` is untouched; its matching gap is recorded in #697.

## Review round: what was wrong, and whether I caused it

An external adversarial review found five issues; all adopted, each reproduced
before fixing and now covered by a regression.

**The relative-path false positive was mine.** `gda script validate deck.gd
--project game`, run from the workspace above the project, was refused as outside.
I verified base behaviour directly against the engine — the pre-refusal chain
passes the path through and the engine, launched with `--path game`, opens
`game/deck.gd` and reports `valid: true`. So this was **a regression my refusal
introduced**, not a pre-existing mismatch that the fix now corrects. It is also
what the README documents ("point `gda` at a project once; relative paths then
resolve *inside* it"), which makes it a documented-behaviour break. Fixed by
anchoring containment where the engine anchors.

**Two containment bypasses**, both reproduced with the engine actually compiling
the outside script: the symlink-then-`..` pivot, and a filesystem path merely
*containing* `://` being classified as engine-virtual (a colon is legal in a POSIX
filename). Both refused now, with the engine never launched.

**Two contract/doc gaps**: `project_root` was absent from the published `required`
list, and neither the README nor `docs/command-catalog.md` mentioned the field or
the refusal.

## Docs

`README.md` (+ the `zh-CN` / `es` / `ja` translations, updated in-language and
re-stamped via `scripts/update_readme_i18n.py`) and `docs/command-catalog.md` now
document `project_root` and the refusal.

> **Coordination:** #692 also has README edits in flight. This PR merges first;
> #692 then rebases and re-stamps the i18n markers.

## Design note

The refusal and the stamping must apply to **both** input paths, and
`--params-json` bypasses the argv body. The descriptor `recipe` (ADR-0023) is the
one hook both paths share, so `script validate` now carries one — while still
running the same sentinel op through `cmd.execute` (as the export recipe does for
its preflight) and the same `classify_script_validate`, so stderr diagnostics are
parsed in exactly as before. `kind` stays `HEADLESS`.

The engine-virtual path rule (`"://"` means the engine resolves it against the
project) now has **one owner**, `gda.project.is_engine_virtual_path`, read by
both `models.normalize_path` and the containment check.

## Validation

Real-engine evidence, reproduced before and after:

```
# from the ancestor (projectless) — the GDA-DF-035 cascade, now attributable
{"path":"inner/deck.gd","valid":false,"diagnostics":[
  {"line":3,"message":"Parse Error: Preload file \"res://scripts/card.gd\" does not exist."},
  {"line":6,"message":"Parse Error: Cannot infer the type of \"c\" variable ..."}],
 "project_root":null}

# pointed at a project that does not own it — refused, no engine run
{"error":{"category":"operation","code":"project_not_found",
 "message":"/…/workspace/inner/deck.gd is outside the resolved Godot project /…/outer: its res:// dependencies would resolve against the wrong root, so nothing was parsed. …"}}

# through an in-project symlink (game/addons/cardlib -> ../../libs/cardlib):
# valid, and its res://addons/cardlib/card.gd preload resolved THROUGH the link
{"path":"/…/game/addons/cardlib/deck.gd","valid":true,"diagnostics":[],
 "project_root":"/…/game"}
```

Gates (all run on the final tree, all green):

- `uv run pytest -m "not e2e"` — 1326 passed, 412 deselected.
- `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` — 1326 passed.
- `uv run ruff check` — clean; `uv run ruff format --check` — 430 files formatted.
- `uv run pyright` — 0 errors, 0 warnings.
- Godot e2e, foreground under the shared lock: `-m e2e -k script_validate` — **12
  passed**, including the three new review regressions (ancestor-cwd relative
  target on both channels, the symlink-`..` pivot, the colon-bearing outside path).
  An earlier round ran the full `tests/test_e2e_script.py` +
  `tests/test_e2e_params_json.py` — 58 passed.
- `uv run pytest tests/test_readme_i18n_sync.py` — 5 passed (translations re-stamped).

## Tests added

- `tests/test_project.py` — 11 containment cases: inside, sibling tree, `res://`/
  `user://`/`uid://`, symlinked *project* spelling (both directions), **directory
  symlinked into the project**, **file symlinked into the project**, **`..`
  escape still outside**, relative path, non-existent path, nested-project scope
  line.
- `tests/test_script_commands.py` — refusal (asserting the engine was **never
  called**), the identical refusal via `--params-json`, `project_root` reported,
  `project_root` null when projectless, `res://` never refused, and the two human
  output shapes.
- `tests/test_e2e_script.py` — the AC3 regression, the real-tier refusal, and the
  symlinked-addon pair (accepted through the link / refused from outside).
- `tests/test_command_descriptor_registry.py` — `script-validate` added to the
  pinned recipe set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,6 +421,14 @@ does not compile is still a successful operation: exit `0`, no top-level `error`
 and `valid: false` with `error_string` / `diagnostics`. Operation problems such as
 a missing file still use the normal Error envelope.
 
+The result also reports `project_root`: the project the script was compiled against,
+which is the root its `res://` dependencies resolved to (`null` when no project was
+resolved). Read it before acting on a `valid: false` — a verdict full of missing
+`res://` dependencies, plus the type errors derived from them, usually means the wrong
+project rather than a broken script. A script *outside* the resolved project is refused
+up front with `project_not_found`, naming both the file and the project, instead of
+reporting those false errors; pass `--project` for the project that owns the file.
+
 **`project`** — the project as a whole (settings, autoloads, static analysis)
 
 | Command | What it does |
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=7e7ef59df7e407b072ba03aa8dd6fb0325c4043a1e064f1ffb98cafa3230dc6a -->
+<!-- gda-readme-i18n: source=README.md sha256=bb3fb3e81210910899b8cfd05144fa9f969d9d214a0436f5e002a1f496c37f0e -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -432,6 +432,15 @@ script que no compila sigue siendo una operación exitosa: salida `0`, sin `erro
 de nivel superior, y `valid: false` con `error_string` / `diagnostics`. Los
 problemas de operación, como un archivo inexistente, siguen usando el Error envelope
 normal.
+
+El resultado también informa `project_root`: el proyecto contra el que se compiló el
+script, es decir, la raíz a la que se resolvieron sus dependencias `res://` (`null`
+cuando no se resolvió ningún proyecto). Léelo antes de actuar sobre un `valid: false`:
+un veredicto lleno de dependencias `res://` inexistentes, más los errores de tipo
+derivados de ellas, suele significar el proyecto equivocado y no un script roto. Un
+script *fuera* del proyecto resuelto se rechaza de entrada con `project_not_found`,
+nombrando tanto el archivo como el proyecto, en lugar de informar esos errores falsos;
+pasa `--project` con el proyecto al que pertenece el archivo.
 
 **`project`** — el proyecto en su conjunto (ajustes, autoloads, análisis estático)
 
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=7e7ef59df7e407b072ba03aa8dd6fb0325c4043a1e064f1ffb98cafa3230dc6a -->
+<!-- gda-readme-i18n: source=README.md sha256=bb3fb3e81210910899b8cfd05144fa9f969d9d214a0436f5e002a1f496c37f0e -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -436,6 +436,14 @@ offset プロパティを明示的に設定してください。
 コンパイルできないスクリプトも成功した操作として扱われます。終了コードは `0`、トップレベルの
 `error` はなく、`valid: false` と `error_string` / `diagnostics` が返ります。存在しない
 ファイルなどの操作上の問題は、通常どおり Error envelope を使います。
+
+結果には `project_root` も含まれます。これはスクリプトのコンパイル対象となったプロジェクト、
+つまり `res://` 依存の解決先となったルートです(プロジェクトが解決されなかった場合は `null`)。
+`valid: false` に対処する前にこれを確認してください。存在しない `res://` 依存とそこから派生した
+型エラーばかりの結果は、通常スクリプトの不具合ではなくプロジェクトの選択ミスを意味します。
+解決されたプロジェクトの**外**にあるスクリプトは、その偽のエラー群を報告する代わりに、
+ファイルとプロジェクトの両方を示した `project_not_found` で事前に拒否されます。その場合は
+そのファイルを所有するプロジェクトを `--project` で指定してください。
 
 **`project`** — プロジェクト全体(設定、オートロード、静的解析)
 
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=7e7ef59df7e407b072ba03aa8dd6fb0325c4043a1e064f1ffb98cafa3230dc6a -->
+<!-- gda-readme-i18n: source=README.md sha256=bb3fb3e81210910899b8cfd05144fa9f969d9d214a0436f5e002a1f496c37f0e -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -417,6 +417,12 @@ Cursor 没有 `mcp add` 命令——请通过上面的 JSON 或 Settings → MCP
 对 `script validate --json`，要读取结果对象里的 `valid` 字段。脚本编译失败仍然算一次成功操作：
 退出码为 `0`，没有顶层 `error`，并以 `valid: false` 携带 `error_string` / `diagnostics`。
 缺失文件等操作层问题仍然采用标准的 Error envelope。
+
+结果还会报告 `project_root`：脚本编译时所针对的项目，也就是其 `res://` 依赖解析到的根目录
+（未解析到项目时为 `null`）。在处理 `valid: false` 之前请先读它——若诊断里满是缺失的 `res://`
+依赖以及由此派生的类型错误，通常说明项目选错了，而不是脚本本身有问题。位于所解析项目**之外**
+的脚本会被提前拒绝并返回 `project_not_found`，同时指明文件与项目，而不是报出那一连串假错误；
+此时请用 `--project` 指定真正拥有该文件的项目。
 
 **`project`** — 作为整体的项目（设置、autoload、静态分析）
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -156,7 +156,7 @@ operation, and parse codes the CLI assigns).
 | `save_failed` | `operation` | `operation` | `4` | A scene could not be packed or saved. |
 | `delete_failed` | `operation` | `operation` | `4` | A file could not be removed from disk. |
 | `file_changed_externally` | `operation` | `operation` | `4` | A read-modify-write operation's target file changed on disk between the read and the write, so the write was refused to avoid clobbering the external edit. |
-| `project_not_found` | `operation` | `operation` | `4` | gda ran without a usable resolved Godot project: an operation needed one and none was resolved, or an explicit `--project` was empty, or a `--project`/`$GDA_PROJECT` does not name a Godot project (no `project.godot`). |
+| `project_not_found` | `operation` | `operation` | `4` | gda has no resolved Godot project usable for the requested target: an operation needed one and none was resolved, or an explicit `--project` was empty, or a `--project`/`$GDA_PROJECT` does not name a Godot project (no `project.godot`), or the target lies outside the resolved project (whose `res://` dependencies would then resolve against the wrong root). |
 | `path_not_found` | `operation` | `operation` | `4` | A requested file does not exist. |
 | `not_a_scene` | `operation` | `operation` | `4` | A requested file cannot be loaded as a `PackedScene`. |
 | `parent_not_found` | `operation` | `operation` | `4` | A requested parent node path does not resolve to a node in the scene. |
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -459,6 +459,21 @@ parsed Python-side and may carry only the **first** error. **`column` is always 
 standard Godot build (the engine exposes no column for a parse error). `validate` reuses existing
 codes only (no new ones).
 
+**Project context** (#658): the result carries `project_root` — the project the script was
+compiled against, i.e. the root its `res://` dependencies resolved to, absolute, and `null` when
+gda ran projectless. It is **required and nullable**, so every verdict carries the key. It exists
+because a script compiled against the wrong project reports every `res://` dependency as missing
+plus the type errors derived from them, which reads as a broken script; `project_root` is what
+tells the two apart. A target **outside** the resolved project is **refused before parsing** with
+`project_not_found` naming both the file and the project, rather than emitting that false cascade.
+Containment follows the engine's own addressing: a relative path is anchored at the resolved
+project (not gda's cwd), an engine-virtual path (`res://`, `user://`, `uid://`) is inside by
+construction, and a file reached through a symlink into the project counts as inside — except when
+a `..` traversal could cross that symlink, where only the fully resolved location decides. gda
+never derives the project from the target path (ADR-0006), so a script under a project **nested
+inside** the resolved one is contained and not refused; `project_root` is what surfaces that
+mismatch, pending the ADR-0006 amendment tracked in #697.
+
 | Command | Description |
 | --- | --- |
 | `gda script create` | Create a `.gd` script (template or content) |
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -24,6 +24,7 @@ from typing import Optional, Protocol, runtime_checkable
 import typer
 from pydantic import BaseModel, Field, model_validator
 
+from gda import dispatch
 from gda.binary import resolve_godot_binary
 from gda.dispatch import dispatch_domain, dispatch_recipe
 from gda.errors import (
@@ -32,6 +33,7 @@ from gda.errors import (
     classify_run,
     script_did_not_run_failure,
     script_exit_status_failure,
+    script_outside_project_failure,
     script_path_invalid_failure,
     script_run_project_not_found_failure,
     unresolvable_binary_failure,
@@ -45,6 +47,7 @@ from gda.headless import (
     project_option,
 )
 from gda.models import CREATED_DIRS_DESC, NormalizedPath
+from gda.project import path_outside_project
 from gda.runner import RunResult, launch
 from gda.script_errors import (
     ScriptError,
@@ -504,6 +507,10 @@ class ScriptValidateResult(BaseModel):
     ``diagnostics`` is a best-effort list parsed from the engine's stderr (the
     only place line/message are available); it may hold only the first error, and
     is empty when the script is valid or nothing could be parsed.
+
+    ``project_root`` names the project the script was compiled against, so a
+    reader can tell a real compile error from one caused by the wrong project
+    context without re-deriving gda's resolution (#658).
     """
 
     path: str
@@ -522,6 +529,18 @@ class ScriptValidateResult(BaseModel):
         description=(
             "Best-effort advisory diagnostics parsed from the engine's stderr "
             "(line + message). May hold only the first error; empty when valid."
+        ),
+    )
+    project_root: str | None = Field(
+        default=None,
+        description=(
+            "The Godot project this script was compiled against — the root its "
+            "res:// dependencies resolved to (ADR-0006: --project, then "
+            "$GDA_PROJECT, then the current directory). Null when gda ran "
+            "projectless (no project resolved), where only filesystem paths "
+            "resolve; a script whose res:// dependencies were reported missing "
+            "with a null or unexpected root here has a project-context problem, "
+            "not a source problem."
         ),
     )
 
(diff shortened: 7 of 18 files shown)


qinhaihong-red:
**Changes requested.** The new containment authority has both false-negative and false-positive cases: some outside paths bypass refusal, while an ordinary project-relative path is rejected.

Reviewed refs:

- Head: `f5d2e6ac362a92865c254bdd28083099f0321ef9`

## Domain architecture

### [P1] A symlink followed by `..` bypasses the outside-project refusal

`src/gda/project.py:97-98`

The lexical branch returns `None` before consulting the resolved location. That is incorrect when `..` is interpreted after crossing a symlink. Exact-head reproduction:

```text
game/pivot -> ../outside/deep
input:     game/pivot/../deck.gd
abspath:   game/deck.gd
resolve:   outside/deck.gd
result:    path_outside_project(...) == None
```

A real Godot 4.6.3 run accepted the path, compiled the physical outside script against `game`, and returned a false dependency diagnostic. This is not #697's derive-vs-refuse decision: the target is already outside under real filesystem resolution, and the helper claims to reject it.

Minimum fix: do not accept lexical containment before accounting for symlink-sensitive parent traversal. Add a unit counterexample and a real-engine regression proving no launch or false diagnostic occurs.

### [P2] `"://"` is not a valid project-virtual-path discriminator

`src/gda/project.py:31-42,93-94`

Any raw string containing `"://"` is classified as engine-virtual and skips containment. A valid POSIX outside path can be spelled `/workspace/outside://deck.gd`; it resolves to the filesystem file `/workspace/outside:/deck.gd`. On the exact head, `path_outside_project()` returned `None`, and the engine opened the outside file.

This merges two different rules: “a documented Godot virtual prefix” and “an arbitrary string contains a scheme separator.” Minimum fix: recognize the supported prefixes exactly (`res://`, `user://`, `uid://`) and add the outside-path regression.

The retained topology is appropriate: `gda.commands.script` owns the validate workflow/result; `gda.project` owns resolution and containment; `gda.dispatch` supplies the shared argv/JSON path; and `HeadlessCommand.execute` plus the runner remain the outbound adapter. No new context or DDD building block is needed, but the reusable containment owner is not safe until these cases are closed.

- `PR695-CONTAINMENT`: simple inside/sibling/symlink/relative tests pass, but the symlink-then-`..` counterexample disproves complete containment.
- `PR695-VIRTUAL-PATH`: documented Godot prefixes pass through, but the valid POSIX `"://"` spelling disproves the predicate.
- `PR695-RECIPE-PARITY`: focused argv/`--params-json` tests pass; revalidate both input channels after repairing containment.

## Standards

### [P1] Project-relative paths are rejected before launch

`src/gda/project.py:95-100`

The new check anchors a relative target at the invoker cwd. This conflicts with `README.md:136-138`, which says that after selecting a project, relative paths resolve inside it. From an ancestor workspace, both argv and `--params-json` reject `deck.gd --project game`; direct Godot with `--path game` validates the script successfully.

Minimum fix: anchor relative filesystem targets at the resolved project and cover both input channels from an ancestor cwd.

### [P2] The schema does not guarantee `project_root`

`src/gda/commands/script.py:534-545`

The nullable field has `default=None`, so it is absent from `output.required`, although every public success result is stamped with it and the shipped Skill tells agents to read it. This weakens the descriptor-derived contract.

Minimum fix: make the public result field required-but-nullable, use a separate internal sentinel shape or stamp before public-model validation, and assert `project_root` appears in `required`.

### [P2] Public documentation remains inconsistent with the changed interface

`README.md:419-422`; `docs/command-catalog.md:446-460`

These surfaces still omit `project_root`; the catalog also omits the new `project_not_found` refusal. This conflicts with the repository's cross-artifact synchronization rule. #697 does not repair the current exact-head artifacts.

No additional baseline smell was substantiated.

## Spec

### [P1] Containment uses the wrong root for relative targets

`src/gda/project.py:95-100`

Issue #658 requires refusal only when the script lies outside the resolved project. With cwd `workspace`, `--project game`, and `deck.gd`, the script is inside the selected project under the actual Godot path semantics, but both public input channels refuse `/workspace/deck.gd` as outside.

Minimum fix: anchor the relative path at the selected project and add an ancestor-cwd regression using relative project and script arguments.

### [P2] Any raw path containing `://` bypasses containment

`src/gda/project.py:31-42`

Issue #658 requires outside scripts to be refused and never emit false dependency errors. The valid POSIX spelling above is treated as virtual, launches Godot, and returns the false missing-`res://` cascade. Recognize only the documented virtual schemes.

Nearest-project discovery, nested-under-resolved ownership, case folding, and ordinary cross-project symlinks were not retained as findings because #658 explicitly defers those decisions.

## Validation

- Focused suite: `115 passed`; `git diff --check` passed.
- The architecture review ran two targeted real-Godot 4.6.3 counterexamples.
- I independently reproduced `symlink/../deck.gd → outside/deck.gd` returning `None` and raw `/outside://deck.gd` bypassing containment on the exact head.
- Both public relative-path failures were reproduced; direct Godot served as the control.
- Required CI is green; the repository Godot e2e job is skipped.

Summary: Domain architecture — 2 findings, worst P1. Standards — 3 findings, worst P1. Spec — 2 findings, worst P1.

qinhaihong-red:
Thanks — this was a strong review. **All five groups adopted.** Every finding was
reproduced against the real engine *before* fixing, and each is now a regression
test. Fixed in `7447e6ce` (head was `f5d2e6ac` when you reviewed).

## A — relative targets anchored at the invoker cwd (P1, false-positive refusal)

**Adopted, and the honest verdict is that I caused it.** You asked whether this was
a regression or a pre-existing mismatch: I ran the pre-refusal chain directly —
`godot --headless --path game --script operations.gd -- script-validate
'{"path":"deck.gd"}'` from the workspace above the project — and it returns
`{"path":"deck.gd","valid":true}`. There is no `workspace/deck.gd`; the engine
anchored the relative path at the project. So base validated this invocation fine
and **my containment check introduced the false positive**, on behaviour the README
documents ("point `gda` at a project once; relative paths then resolve *inside*
it"). Reproduced on `f5d2e6ac`:

```
$ cd workspace && gda script validate deck.gd --project game --json
{"error":{"category":"operation","code":"project_not_found",
 "message":"/…/workspace/deck.gd is outside the resolved Godot project game: …"}}
```

Note the message also leaked the bare relative `game`.

**Fix:** a single anchoring rule, `gda.project.project_anchored` — relative at the
*resolved project*, absolute untouched — used by the containment check, so it and
the engine cannot disagree about which file a relative argument names. **Chain
verified:** the op still receives the path unchanged (`fake.calls == [("script-validate",
{"path": "deck.gd"})]`) because the engine does the anchoring; the e2e proves the
whole chain against a real engine. `project_root` is now reported **absolute**, so
`--project game` no longer emits `"project_root": "game"`.

**Tests:** unit `test_a_relative_path_is_anchored_at_the_project_not_the_cwd` (incl.
a relative *project* spelling) and `test_project_anchored_matches_how_the_engine_addresses_a_path`;
CLI on **both channels** — `…accepts_a_relative_target_from_an_ancestor_cwd` and
`…relative_target_parity_on_the_params_json_path`, both with relative project **and**
script arguments; plus `…still_refuses_a_relative_target_that_climbs_out`; e2e
`…relative_target_from_an_ancestor_cwd_agrees_with_the_engine` (argv + `--params-json`,
asserting the two results are identical).

## B — symlink then `..` bypassed containment (P1)

**Adopted**, reproduced exactly as you described on `f5d2e6ac` — accepted with
`valid: true`, and the engine compiled `outside/deck.gd`:

```
lexical abspath: /…/reproB/game/deck.gd      ← inside  (accepted)
real resolve():  /…/reproB/outside/deck.gd   ← outside (what was compiled)
```

**Fix:** your contained shape. The lexical reading is consulted **only** when
neither the anchored candidate nor the project spelling carries a `..` component;
otherwise the resolved reading decides alone. I extended the guard to the project
spelling as well as the candidate — a `..` in the root spelling can desynchronise
the lexical root the same way, and being stricter here only ever refuses more.
Round-2's symlinked-into-project acceptance is unaffected (those inputs carry no
`..`), which the retained tests confirm.

**Tests:** unit `test_a_symlink_followed_by_dot_dot_does_not_pass_as_inside`; e2e
`…refuses_a_symlink_dot_dot_pivot_out_of_the_project`, asserting the refusal names
the *real* location and that **no verdict was produced** (`'"valid"' not in stdout`,
empty `diagnostics`) — i.e. no launch, no false diagnostic.

## C — `"://" in path` is not a scheme test (P2)

**Adopted**, reproduced: `/…/outside://deck.gd` was classified virtual, skipped
containment, and the engine opened the outside file with `valid: true`.

**Fix:** `is_engine_virtual_path` now matches the documented prefixes via a new
`ENGINE_VIRTUAL_PREFIXES = ("res://", "user://", "uid://")` tuple.

**Consumer check, as asked.** The predicate has exactly two consumers:
`gda.models.normalize_path` and `path_outside_project` — both now agree, and
tightening it strictly *narrows* what `normalize_path` passes through unexpanded
(a colon-bearing filesystem path now gets `~`-expansion like any other, which is
the correct treatment). On **#693's `script run` gate**: `script run` is res://-only
and refuses `user://`/`uid://`/absolute outright, which is stricter than `validate`
— deliberately, since `validate` has always accepted absolute filesystem paths
(that is the projectless standalone case, pinned by existing e2e). `validate` is not
*looser* on the axis this finding is about: a disguised filesystem path is no longer
waved through. It does still skip containment for a genuine `user://`/`uid://`
target, which is correct — gda has no filesystem claim to make about an
engine-resolved address, and such a target's own `res://` deps still resolve
against the resolved project. Unifying the accepted path *forms* across
`validate`/`run` is #670/#675's scope, not this PR's.

**Tests:** unit `test_a_colon_bearing_filesystem_path_is_not_treated_as_virtual`
(and the same sequence *inside* the project still contained); CLI
`…does_not_refuse_a_colon_bearing_path_as_virtual`; e2e
`…refuses_an_outside_path_that_merely_contains_a_scheme`.

## D — `project_root` absent from `required` (P2)

**Adopted.** It is now **required and nullable**: a `before` validator supplies the
key for the internal sentinel parse — the engine never reports it, since ADR-0006
keeps the project CLI-side and the engine is *told* it via `--path` — and the recipe
stamps the resolved project immediately after. The leniency is inward-facing only;
the published contract lists the field. This is a schema diff, enumerated in the PR
body: `required` goes `["path","valid"]` → `["path","valid","project_root"]`.

**Tests:** `test_script_validate_schema_emits_model_derived_contract_without_other_args`
now asserts `"project_root" in doc["output"]["required"]` and pins the nullable
`anyOf`; the sample-payload gate carries `project_root` in both a rooted and a
projectless instance. That gate is what caught the change — a sample public result
without the key no longer validates, which is exactly the guard you wanted.

## E — README + command-catalog omissions (P2)

**Adopted.** `README.md` and `docs/command-catalog.md` now document `project_root`
and the refusal (the catalog also records the containment semantics and the
nested-project scope line, cross-referencing #697). The three translations are
updated **in-language** and re-stamped with `scripts/update_readme_i18n.py`;
`tests/test_readme_i18n_sync.py` passes. **Known consequence, as flagged:** #692
also has README edits in flight — this PR merges first, then #692 rebases and
re-stamps. Noted in the PR body too.

## Gates on `7447e6ce`

- `pytest -m "not e2e"` — 1326 passed; `GITHUB_ACTIONS=true` variant — 1326 passed.
- `ruff check` clean; `ruff format --check` 430 files; `pyright` 0 errors.
- Godot e2e (foreground, serialized under the shared lock): `-m e2e -k script_validate`
  — **12 passed**, including the three new A/B/C regressions.
- Surface re-derived: `gda schema` still changes **exactly one** command entry
  (`script validate`: `description`, `output`), `kind` stays `headless`, `input` /
  `error` / `constraints` byte-identical; `--help` and `gda skill` unchanged from the
  previous round.

One thing I did **not** change, for the record: a script under a project *nested
inside* the resolved one is still contained and not refused. That is the ADR-0006
boundary this PR deliberately leaves open, tracked in #697.

Issue #668, opened 2026-08-15
Scoped import surface for clean-worktree resource loading


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-010** (2026-08-08) — in a clean worktree with no `.godot/imported` cache, `preload()`/`ResourceLoader.load()` of PNGs fails under `script run` ("no recognized resource loader") although the same project loads them in an imported session; one-shot tests can fail for an import-cache reason absent in the real game.
- **GDA-DF-038** (2026-08-12) — the blunt alternative, a direct editor reimport, touched 2,148 auto-imported entries and left source-adjacent sidecars/cache residue outside the intended diff; unattended asset admission can stage tool-owned noise or misread stale importer state as source drift.

## What to build

A scoped import operation (a `--import`/`--ensure-imports` option on `script run`, or a dedicated command) that:

- imports only the requested dependencies into the project cache;
- reports, per asset, whether the cache was hit or missed;
- classifies every file it creates as cache-owned or source-adjacent, under an explicit cache root;
- supports a dry run that reports the full mutation inventory without mutating;
- returns a machine-readable completion summary (imported / skipped / failed counts and paths).

Plain `script run` must not trigger an import pass. (`script run` itself is not side-effect-free — it executes a project script, per the Project-code execution surface — the guarantee here is only that gda adds no import pass.)

**Decisions for triage:**
- Command-group name: `resource import` under the existing `resource` group, or a new `asset` group (CONTEXT.md names groups after the domain object acted on).
- Trust framing: the import pass runs engine importer code against project content — it stays within the `Trusted project` assumption (ADR-0009); record whether it widens the Project-code execution surface, as ADR-0031/0033 did for their additions.

Related: #653 adds a different per-invocation root (user-data/log, not the project import cache) — choose the option names together.

## Acceptance criteria

- [ ] In a project with no `.godot/imported`, a script that `preload()`s a `.png` succeeds after the new import step and fails without it.
- [ ] The result reports per-asset cache hit/miss and the machine-readable completion summary.
- [ ] Dry run reports the full mutation inventory and writes nothing.
- [ ] Every created file is classified (cache-owned vs source-adjacent) in the result.
- [ ] Plain `script run` behavior is unchanged.

## Blocked by

None - can start immediately (the two triage decisions above are recorded on this issue before implementation).

[...]

PR #696, opened 2026-08-16
fix(gda): preflight and redirect the Godot user-data root for the headless launch

### LINKED ISSUE #653: Preflight and redirect the Godot user:// log/user-data root for the Headless launch

qinhaihong-red:
## Summary

In a filesystem-restricted process where the project is writable but Godot's
application-data directory is not, **every** command performing a `Headless launch`
died with signal 11 before any project code ran, and `gda` reported it as
`engine_crashed` plus a C++ backtrace instead of a typed environment refusal
(dogfooding GDA-DF-001/014/042/030/028, ~30 reconfirmations).

`gda` now **owns the engine log target of every headless launch**, in the launch
primitive (`gda.runner.launch`) once, so the sentinel, export and `script run`
channels all inherit it.

Closes #653

## Root cause, verified against the engine source

The crash is **not** the user-data directory failing to be created — that is a
non-fatal `ERR_FAIL_COND_MSG` (`core/os/os.cpp`, `OS::ensure_user_data_dir`). The
crash is the **file logger**: `Main::setup` attaches a `RotatedFileLogger`, which
dereferences the `FileAccess` it failed to open, so an unopenable log target
segfaults in `rotate_file()` before project code runs.

Reproduced locally against Godot 4.6.3 with an unwritable
`$HOME/Library/Application Support`:

```
ERROR: Could not create directory: 'user://logs'.
handle_crash: Program crashed with signal 11
[2] RotatedFileLogger::rotate_file() ...
[5] Main::setup(char const*, int, char**, bool) ...
```

Confirmed by a second, independent trigger of the same code path: `--log-file
/dev/null` also segfaults in `rotate_file()` — the open fails, not the directory
creation. That is why the preflight below **creates and opens** the target rather
than merely stat-ing the directory.

## Engine-verified redirect mechanism

| Want | Mechanism | Evidence |
| --- | --- | --- |
| Redirect the log | `--log-file <path>` | `main/main.cpp:1512` parses it; `main/main.cpp:2320-2336` uses it as `base_path` **and forces `max_files = 1`** ("Ensure log file name respects the specified override by disabling log rotation"), and opens it with `ACCESS_FILESYSTEM`, not `ACCESS_USERDATA` — so it is never resolved through `user://` |
| Redirect `user://` | the platform data variable | there is **no `--user-data-dir` flag**: the spelling appears nowhere in `main/` or `core/`, and passing it changes nothing (verified empirically — the crash was unchanged) |

`user://` resolution, mirrored in `gda.runner.engine_data_path`:

- macOS — `$HOME/Library/Application Support` (`platform/macos/os_macos.mm:433-441`:
  `get_data_path()` returns `get_config_path()`, which reads `HOME`)
- Linux/BSD — `$XDG_DATA_HOME` when absolute, else `$HOME/.local/share`
  (`platform/linuxbsd/os_linuxbsd.cpp:911-916`)
- Windows — `%APPDATA%` (`platform/windows/os_windows.cpp:2418-2424`)

`XDG_DATA_HOME` does **not** redirect `user://` on macOS — confirmed empirically,
and independently corroborated by this repo's own pre-existing skip reason in
`tests/test_e2e_export_run.py:355`.

Default log path is `user://logs/godot.log` with `max_log_files = 5` and
`enable_file_logging.pc = true` (`main/main.cpp:2312-2314`), i.e. on desktop every
default headless run creates and **rotates** one per-project log — which is the
concurrency contention this issue also asks about.

## What changed

1. **Every launch carries `--log-file` at a target `gda` has already created.**
   The creation *is* the preflight: it proves the engine's own
   `FileAccess::open(..., WRITE)` will succeed, and it mirrors the truncate-before-spawn
   the daemon already does for a `Session log` (ADR-0022).
2. **Default target = a private temporary file**, removed after the run. A read-only
   application-data directory is therefore no longer fatal at all, and — since
   `--log-file` also disables rotation — concurrent invocations never share a
   rotation-sensitive file.
3. **A placement `gda` cannot make usable is refused before the spawn** as the new
   `user_data_unwritable` code, with prose diagnostics naming the resolved binary,
   the platform-resolved user-data path and the attempted log target. The ADR-0004
   envelope *shape* is unchanged.
   With an explicit root, the preflight covers **both** halves: the log target *and*
   the platform-derived data path the engine appends to the root
   (`<root>/Library/Application Support` on macOS). Creating the root alone is not
   enough — the derived path can be unusable while the root is perfectly writable,
   and gda would then have promised a `user://` redirect it did not deliver.
4. **`--user-data-root <dir>`** (root option, `$GDA_USER_DATA_ROOT` as the env half)
   places both the log and `user://` under a caller-chosen directory, by overriding
   the platform variable above for the child process.

Placement lives entirely in the launch primitive, so no channel plumbs it and the
`RunnerFactory` seam is untouched.

## Option naming: the pair-form for #668

**Form: `--<resource>-root <dir>`** — a per-invocation directory that `gda`
redirects one engine-owned write target into, with a `GDA_<RESOURCE>_ROOT` env
twin at the same precedence (flag > env > engine default), matching
`--godot`/`$GDA_GODOT` and `--project`/`$GDA_PROJECT`.

So #668's project-import-cache root should be **`--import-cache-root`** /
`$GDA_IMPORT_CACHE_ROOT`. Rationale for the shape:

- it names the *resource* being relocated, not the mechanism (`--log-file` would
  have been mechanism-specific and could not have grown to cover `user://`);
- `root` — not `dir` or `path` — because the value is a directory the engine's own
  layout is created *under*, not the final path. This matters: the resolved
  location is platform-specific (`<root>/Library/Application Support/...` on macOS
  vs `<root>/godot/...` on Linux), so the contract is "lands under this directory"
  and the exact path is reported, never assumed;
- root-level rather than per-command, because it is process-wide *environment
  placement* that applies to whichever command runs, on every channel that spawns
  an engine. Declaring it once also keeps it off all ~66 per-command signatures and
  out of `gda schema`.

## Surface diff (enumerated)

Baseline captured before the first edit; diffed after.

- `gda schema` — **byte-identical**. The option is CLI-level, not an input-model
  field, so no per-command contract changed.
- `gda script run --help` — **byte-identical**.
- `gda --help` — one intended addition, `--user-data-root TEXT`, plus the column
  re-wrap Typer applies to the pre-existing `--version` / `--json` / `--help` rows
  because the option column widened to fit the new name. No wording changed.
- `gda skill` (i.e. `src/gda/skill/SKILL.md`, emitted verbatim) — two intended
  additions, enumerated below.
- **Authoritative contract reconciled** (no new ADR): CONTEXT.md's `Headless launch`
  and `Raw run` entries described an argv of `[binary, --headless, *args]` and a
  `launch_failure` covering only missing-binary/timeout, both of which this change
  outgrew; a new `User-data placement` term now carries the shared-language
  definition. ADR-0031 gets a dated, **append-only** amendment on the launch
  primitive it reuses — zero accepted-prose lines modified.
- New error code `user_data_unwritable` in the ADR-0002 table and
  `gda/error_codes.py`. **No new exit bucket** — it reuses the existing ENVIRONMENT
  exit `127`, alongside `binary_not_found`, `live_unsupported_platform` and
  `live_windowed_unavailable`.

### Registry row added (both places)

| Code | Category | Source | Exit Code | Meaning |
| `user_data_unwritable` | `environment` | `runner` | `127` | The log or user-data placement for the launch could not be made usable, so the launch was refused. |

The description is deliberately about the **placement**, not about Godot's own
directory: the same code covers a private temporary log target that could not be
created — where Godot's location may be perfectly writable — and a redirected root
whose derived data path is unusable. Naming only the latter sent readers to fix the
wrong directory; the diagnostics say which one it was, and a paired test pins that
each refusal blames only its own.

`tests/test_error_registry.py` (the ADR↔Python drift guard) stays green.

### `gda skill` output diff

The Skill is a shipped doc surface, so the guidance lives in the artifact, not in
this PR body. Two additions to `## Setup` and the exit-code table:

```diff
@@ ## Setup @@
+- **User data (headless runs)** — each headless run's engine log goes to a private
+  temporary file, so a read-only Godot application-data directory is not fatal and
+  concurrent runs never contend. If a script needs a writable `user://`, pass
+  `gda --user-data-root DIR <group> <command>` (or set `$GDA_USER_DATA_ROOT`) to
+  place the log and `user://` under `DIR`; a target `gda` cannot create is refused
+  as `user_data_unwritable` before the engine starts. Two limits: Godot reads the
+  **export templates** from that same directory, so do not combine it with
+  `export run` (the export reports no installed templates); and live sessions are
+  unaffected either way — the daemon owns their log.

@@ ## Structured output & errors @@
-| `127` | Godot binary not found |
+| `127` | environment unusable: `binary_not_found`, `user_data_unwritable`, `live_unsupported_platform`, `live_windowed_unavailable` |
```

The exit-code row is a correctness fix this change forces: `user_data_unwritable`
reuses exit `127`, so the old row would have told an agent that `127` means only a
missing binary. It now enumerates all four codes on that exit (review F6).

Ownership note: the wave's Skill owner is #659 (PR #692), which merges first, so
this branch takes the serialized rebase. **Three collision points, not one:**

1. `src/gda/skill/SKILL.md` — the `## Setup` region (#692 adds a Provenance bullet
   there; this adds a User-data bullet).
2. `src/gda/cli.py` — the `from gda.headless import set_root_json` import line, which
   #692's rebase deletes and this PR extends with a second import.
3. `src/gda/cli.py` — the root-callback body: this PR's `set_user_data_root(...)` is
   appended directly after the `set_root_json(ctx, json_output)` line that #692
   deletes.

Point 3 is a **silent** sever — the flag would parse and do nothing. It is now
covered by `test_the_root_option_reaches_the_launch`, verified red-proof by
no-op-ing the hand-over (review F2).

## Validation

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1333 passed, 415 deselected |
| same, `GITHUB_ACTIONS=true` | 1333 passed |
| `uv run ruff format --check` | passed |
| e2e `tests/test_e2e_user_data.py` | 9 passed |
| e2e regression `info`, `script` | 56 passed |
| e2e regression `script_run`, `op_robustness` | 22 passed |
| e2e regression `export`, `export_run` | 15 passed, 1 skipped (pre-existing Linux-only skip) |
| e2e regression `daemon`, `scene` | 50 passed |

All e2e runs were foreground under the shared lock, with a bounded retry that fails
the gate on exhaustion. Every permission change is restored in fixture teardown;
verified afterwards that no test directory was left unwritable and that the real
`~/Library/Application Support/Godot` was untouched.

### The e2e regression is guarded against passing for the wrong reason

`tests/test_e2e_user_data.py` builds a real unwritable application-data directory
under a fake `HOME` (tree pre-created, only the final component locked — not a bare
empty `HOME`, which perturbs Godot's first-run behaviour). It includes a **control
arm** that spawns the raw engine *without* `gda`'s `--log-file` and asserts it does
**not** complete; if the sandbox ever stopped reproducing the restriction, the
control fails rather than the protected arms silently passing.

These fixtures deliberately leave file logging at the **engine default** rather than
using `conftest.project_godot`, which disables it — disabling it is exactly what
would hide the bug, since with no `RotatedFileLogger` there is no crash to survive.

Worth noting: `tests/conftest.py` documents that the e2e tier had to disable file
logging project-side (#180) precisely because concurrent launches raced in
`rotate_file()`. This change fixes that contention at the source, for all users
rather than for `gda`'s own fixtures. The conftest workaround is left in place
(out of scope here).

## Acceptance criteria

| AC | Status |
| --- | --- |
| Unwritable app-data dir → typed structured environment error before project code, no signal-11 backtrace | **Met, in the stronger form.** The engine crash path is not entered at all: with the default isolated target the command *completes* (`test_commands_complete_under_a_read_only_app_data_dir`, `test_script_run_completes_...`). The typed `user_data_unwritable` refusal covers the residual case where `gda`'s own target is unwritable too (`test_unwritable_log_target_is_a_typed_environment_error`). Deviation is deliberate — see below. |
| New code registered in the ENVIRONMENT category with its exit bucket | Met — table above; reuses exit `127`, no new bucket |
| A documented per-invocation option (or default isolated root) lets `script validate/run` and `info` complete under a workspace-restricted profile | Met by the **default** isolated root, proven at the real-engine tier for `script validate` and `script run`. **`info` is corrected, not proven:** see the non-repro note below |
| Failure diagnostics name resolved binary, `user://` root, log path | Met — asserted at unit and e2e tier |
| Concurrent default invocations don't share a crashing log target | Met — per-invocation targets (unit) plus 6 concurrent real-engine invocations (e2e) |
| Persistence case covered or explicitly split | Met via the **strong** branch: `--user-data-root` makes `user://` redirectable for `script run` (`test_user_data_root_makes_user_writable_again` asserts the write lands *and* resolves under the requested root). No script-side "declares writable user data" mechanism was built. |

### Correction: `gda info` never crashed

The issue lists `info` among the commands that died. That does **not** reproduce on
Godot 4.6.3, and this PR no longer claims it does. `info` is projectless, and
`Main::setup` builds the file logger only when `!project_manager && !editor && …`;
a projectless run takes the project-manager branch, so no `RotatedFileLogger` is
ever constructed and there is nothing to crash. Verified directly — the projectless
sentinel op under the restriction emits its result with zero `handle_crash` lines,
with *and* without this fix.

An earlier revision of this PR asserted an `info` arm that therefore passed with the
fix removed. It has been dropped rather than kept as a green smoke, and the finding
is recorded in the e2e module docstring (review F4a). `info` is still covered for the
*refusal* path, which does apply to it: gda creates a log target for every launch,
projectless included.

### Deliberate deviation on AC 1 — formally amended on the issue

Recorded as an accepted amendment to the acceptance criteria: https://github.com/aigengame/godot-agent/issues/653#issuecomment-5313363800

The issue offered two shapes — refuse with a typed error, *or* default to an
isolated writable root. I implemented both, with the isolation as the default,
because refusing is wrong for the common case: `gda info`, `scene create` and most
commands never touch `user://`, so failing them because the application-data
directory is read-only would preserve the very pain this issue is about. `gda`
refuses only what it cannot make work — its own log target.

## Flagged for the orchestrator

- **SKILL.md now lands in this PR** (adjudicated: the Skill owner #659 / PR #692
  merges first, so this hunk takes the serialized rebase). Diff enumerated above.
- **Known bound.** A `gda` process killed by a signal cannot remove its temporary
  log directory; the leftover is a tiny file in `$TMPDIR`, which the OS reaps.
  Normal exits (including every failure path) clean up — asserted by
  `test_default_log_target_is_removed_after_the_run`.
- `--user-data-root` also moves the export templates and editor settings Godot
  reads from that directory — so an `export run` under it finds none (measured:
  `templates_installed` flips true→false). Now stated in both the option help and
  the Skill; it is opt-in and per-invocation, and the default never touches the
  child environment.
- **Explicit non-goal: the live/daemon channel.** `--user-data-root` redirects
  HEADLESS launches only. A live `Engine session` is spawned by the daemon, which
  already owns its `--log-file` (ADR-0022) and so is immune to this failure by
  construction; it does not read this option. The option help and the Skill are
  scoped to say so, rather than promising "this invocation" generally (review F5).

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -78,18 +78,37 @@ _Avoid_: consistency, coherence, sync
 The one-shot `godot --headless` spawn primitive that the Phase-1 channels share —
 the sentinel op-dispatch runner, the native-export runner, and the `gda script run`
 user-script runner (ADR-0031). Given the binary, an argv tail, an optional working
-directory, and a timeout, it builds `[binary, --headless, *args]`, captures bytes
-with the timeout, and normalizes the outcome into a `Raw run` (the single home of
-the spawn / timeout / launch-failure / UTF-8-decode handling). Each channel
-contributes only its argv tail and the export-only cwd.
+directory, and a timeout, it builds `[binary, --headless, --log-file <gda-owned
+path>, *args]`, captures bytes with the timeout, and normalizes the outcome into a
+`Raw run` (the single home of the spawn / timeout / launch-failure / UTF-8-decode
+handling). Each channel contributes only its argv tail and the export-only cwd. It
+also owns the launch's `User-data placement` — resolved and preflighted here, once,
+so no channel plumbs it (#653).
 _Avoid_: spawn helper, subprocess wrapper
+
+**User-data placement**:
+Where one `Headless launch` puts the engine's log and, when redirected, `user://`.
+gda always owns the **log** target and passes it as `--log-file`, because Godot
+builds its file logger before any project code runs and dies with signal 11 if it
+cannot open the log — and because the engine default is one per-project rotated
+file that concurrent invocations contend over. By default the target is a private
+temporary file, so a read-only application-data directory is not fatal; the
+per-invocation `--user-data-root` (env `GDA_USER_DATA_ROOT`) instead places the log
+*and* `user://` under a caller-chosen directory, since Godot has no
+`--user-data-dir` flag and the platform data variable is the only lever. The
+placement is **created, not inspected**, before the spawn — that creation IS the
+preflight — and a placement gda cannot make usable is a typed refusal
+(`user_data_unwritable`) rather than an engine crash. Headless only: a live
+`Engine session`'s log is daemon-owned (ADR-0022).
+_Avoid_: log redirect, user dir, sandbox
 
 **Raw run**:
 The normalized outcome a `Headless launch` returns — `{stdout, stderr, exit_code,
 launch_failure}`, unparsed — before any classification. `launch_failure` is set
-only when the primitive synthesized the result (binary missing, timed out) rather
-than the engine returning one, so the classifier keys environment failures on
-that typed reason, not on the overloaded exit code. Those launch-backed channels
+only when the primitive synthesized the result (binary missing, timed out, or the
+`User-data placement` was refused) rather than the engine returning one, so the
+classifier keys environment failures on that typed reason, not on the overloaded
+exit code. Those launch-backed channels
 all return the one `RunResult` shape. Normally internal, it is **promoted to a
 public result by `gda script run`** — the one operation whose success result *is*
 a Raw run (minus `launch_failure`, which is lifted out into an `Error envelope`),
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -143,7 +143,7 @@ operation, and parse codes the CLI assigns).
 | --- | --- | --- | --- | --- |
 | `binary_not_found` | `environment` | `runner` | `127` | The Godot binary could not be launched. |
 | `launch_timeout` | `environment` | `runner` | `124` | Godot launched but did not return before the runner timeout. |
-| `user_data_unwritable` | `environment` | `runner` | `127` | Godot's user-data location is not writable, so the launch was refused. |
+| `user_data_unwritable` | `environment` | `runner` | `127` | The log or user-data placement for the launch could not be made usable, so the launch was refused. |
 | `unsupported_version` | `version` | `version_gate` | `3` | The detected Godot version is below the supported minimum. |
 | `engine_crashed` | `operation` | `classifier` | `4` | Godot terminated abnormally, such as by signal death. |
 | `operation_failed` | `operation` | `classifier` | `4` | The engine or operation failed without a valid registered operation error envelope. |
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -143,6 +143,7 @@ operation, and parse codes the CLI assigns).
 | --- | --- | --- | --- | --- |
 | `binary_not_found` | `environment` | `runner` | `127` | The Godot binary could not be launched. |
 | `launch_timeout` | `environment` | `runner` | `124` | Godot launched but did not return before the runner timeout. |
+| `user_data_unwritable` | `environment` | `runner` | `127` | Godot's user-data location is not writable, so the launch was refused. |
 | `unsupported_version` | `version` | `version_gate` | `3` | The detected Godot version is below the supported minimum. |
 | `engine_crashed` | `operation` | `classifier` | `4` | Godot terminated abnormally, such as by signal death. |
 | `operation_failed` | `operation` | `classifier` | `4` | The engine or operation failed without a valid registered operation error envelope. |
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -253,3 +253,17 @@ added incrementally under ADR-0025 if a concrete need appears.
   dispatch channels.
 - **Large output** streams through stdout like any headless result; if a future need surfaces, the
   ADR-0002 result-file escape hatch applies, but it is not committed here.
+
+> **Amendment 2026-08-17 (#653).** The shared `launch()` primitive this decision reuses has since
+> grown a third responsibility, so what `script run` inherits from it is wider than described
+> above. `launch()` now also owns each launch's **`User-data placement`**: it always passes a
+> gda-owned `--log-file` (Godot builds its file logger before any project code and dies with
+> signal 11 if it cannot open the log; the engine default is additionally one per-project rotated
+> file that concurrent invocations contend over), and it preflights that placement by creating it.
+> Two consequences for this ADR's own contract, neither of which changes the passthrough decision:
+> the argv is `[binary, --headless, --log-file <path>, *args]` rather than `[binary, --headless,
+> *args]`; and `Raw run.launch_failure` has a third value, `USER_DATA_UNWRITABLE`, which — like the
+> existing two — is lifted out of the promoted `script run` success result into an `Error envelope`
+> (`user_data_unwritable`). The per-invocation `--user-data-root` also makes `user://` writable for
+> a `script run` under a restricted profile. See CONTEXT.md's `User-data placement` entry for the
+> shared-language definition.
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -148,8 +148,9 @@ def main(
         "so a read-only application-data directory is not fatal and concurrent "
         "runs do not share one log; pass this when `user://` itself must be "
         "writable. Godot reads the export templates and editor settings from that "
-        "same directory, so an export run under it finds no installed templates. "
-        "A live session is unaffected: the daemon owns its log (ADR-0022).",
+        "same directory, so a release/debug 'export run' under it finds no "
+        "installed templates unless you place them there ('--mode pack' needs "
+        "none). A live session is unaffected: the daemon owns its log (ADR-0022).",
     ),
 ) -> None:
     """An agent-facing Godot CLI with structured output."""
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -38,6 +38,7 @@ from gda.commands import (
 )
 from gda.headless import root_json, set_root_json
 from gda.provenance import build_version_provenance, render_version_line
+from gda.runner import USER_DATA_ROOT_ENV, set_user_data_root
 
 app = typer.Typer(
     name="gda",
@@ -138,6 +139,17 @@ def main(
         "--json after the command; with --version it emits structured install "
         "provenance (`--help` stays text).",
     ),
+    user_data_root: Optional[str] = typer.Option(
+        None,
+        "--user-data-root",
+        help="Directory to place Godot's user data under for this invocation: the "
+        f"engine log and `user://` (overrides ${USER_DATA_ROOT_ENV}). By default "
+        "gda redirects only the engine log, to a private temporary file, so a "
+        "read-only application-data directory is not fatal and concurrent runs do "
+        "not share one log; pass this when `user://` itself must be writable. It "
+        "also moves the export templates and editor settings Godot reads from that "
+        "directory.",
+    ),
 ) -> None:
     """An agent-facing Godot CLI with structured output."""
     # This callback keeps gda a command *group* so meta commands like `gda info`
@@ -150,8 +162,13 @@ def main(
     # inert: handing it to the shared option layer makes the invoked command's own
     # `--json` inherit it, so the root and post-command spellings mean the same
     # thing — accepting a flag that silently returned human text would be worse than
-    # the loud usage error it replaced. The hand-over happens in the option's own
-    # callback (`_record_root_json`), so this body has nothing left to do.
+    # the loud usage error it replaced. The --json hand-over happens in the
+    # option's own callback (`_record_root_json`), so it needs no line here.
+    # `--user-data-root` is a root option because it is process-wide ENVIRONMENT
+    # placement, not an operation parameter: it applies to whichever command runs,
+    # on every channel that spawns an engine, and the runner reads it there (#653).
+    # Handing it over here keeps the knowledge running downward.
+    set_user_data_root(user_data_root)
 
 
 meta_commands.register(app)
--- a/src/gda/cli.py
+++ b/src/gda/cli.py
@@ -142,13 +142,14 @@ def main(
     user_data_root: Optional[str] = typer.Option(
         None,
         "--user-data-root",
-        help="Directory to place Godot's user data under for this invocation: the "
-        f"engine log and `user://` (overrides ${USER_DATA_ROOT_ENV}). By default "
-        "gda redirects only the engine log, to a private temporary file, so a "
-        "read-only application-data directory is not fatal and concurrent runs do "
-        "not share one log; pass this when `user://` itself must be writable. It "
-        "also moves the export templates and editor settings Godot reads from that "
-        "directory.",
+        help="Directory to place Godot's user data under for a HEADLESS launch: "
+        f"the engine log and `user://` (overrides ${USER_DATA_ROOT_ENV}). By "
+        "default gda redirects only the engine log, to a private temporary file, "
+        "so a read-only application-data directory is not fatal and concurrent "
+        "runs do not share one log; pass this when `user://` itself must be "
+        "writable. Godot reads the export templates and editor settings from that "
+        "same directory, so an export run under it finds no installed templates. "
+        "A live session is unaffected: the daemon owns its log (ADR-0022).",
     ),
 ) -> None:
     """An agent-facing Godot CLI with structured output."""
--- a/src/gda/error_codes.py
+++ b/src/gda/error_codes.py
@@ -87,6 +87,13 @@ ERROR_CODES: tuple[ErrorCodeSpec, ...] = (
         "Godot launched but did not return before the runner timeout.",
     ),
     ErrorCodeSpec(
+        "user_data_unwritable",
+        ErrorCategory.ENVIRONMENT,
+        EXIT_NOT_FOUND,
+        ErrorCodeSource.RUNNER,
+        "Godot's user-data location is not writable, so the launch was refused.",
+    ),
+    ErrorCodeSpec(
         "unsupported_version",
         ErrorCategory.VERSION,
         EXIT_VERSION,
--- a/src/gda/error_codes.py
+++ b/src/gda/error_codes.py
@@ -91,7 +91,13 @@ ERROR_CODES: tuple[ErrorCodeSpec, ...] = (
         ErrorCategory.ENVIRONMENT,
         EXIT_NOT_FOUND,
         ErrorCodeSource.RUNNER,
-        "Godot's user-data location is not writable, so the launch was refused.",
+        # Deliberately about the PLACEMENT, not about Godot's own directory: the
+        # same code covers a private temporary log target that could not be created
+        # (where Godot's location may be perfectly writable) and a redirected root
+        # whose derived data path is unusable. Naming only the latter sent readers
+        # to fix the wrong directory; the diagnostics name which one it was.
+        "The log or user-data placement for the launch could not be made usable, "
+        "so the launch was refused.",
     ),
     ErrorCodeSpec(
         "unsupported_version",
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -208,7 +208,8 @@ def classify_launch_or_crash(raw: RunResult, binary: Path | None) -> Failure | N
         # user-data directory, and the log path.
         return make_failure(
             "user_data_unwritable",
-            "Godot's user data location is not writable; the launch was refused",
+            "the log or user data placement for this launch is not usable; "
+            "the launch was refused",
             raw.stderr,
         )
     if raw.exit_code < 0:
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -20,6 +20,8 @@ engine. The decision tree, top to bottom (``code`` in parentheses; the four
 - launch NOT_FOUND → environment / binary_not_found  (runner could not launch it)
 - launch TIMEOUT   → environment / launch_timeout     (runner launched it but it
   hung past the timeout)
+- launch USER_DATA_UNWRITABLE → environment / user_data_unwritable (the engine log
+  target gda owns could not be created, so the launch was refused, #653)
 - exit < 0  → operation   / engine_crashed         (engine killed by a signal)
 - exit ≠ 0  → operation   / <operation code>        (operation reported a structured
   failure via the ADR-0002 error envelope — e.g. path_not_found)
@@ -196,6 +198,17 @@ def classify_launch_or_crash(raw: RunResult, binary: Path | None) -> Failure | N
         return make_failure(
             "launch_timeout",
             "Godot launched but did not return before the timeout",
+            raw.stderr,
+        )
+    if raw.launch_failure is LaunchFailure.USER_DATA_UNWRITABLE:
+        # Refused before the spawn (issue #653): the engine builds its file logger
+        # ahead of any project code and dies with signal 11 when it cannot open the
+        # log, so this environment problem would otherwise arrive as an
+        # `engine_crashed` backtrace. The runner's diagnostics name the binary, the
+        # user-data directory, and the log path.
+        return make_failure(
+            "user_data_unwritable",
+            "Godot's user data location is not writable; the launch was refused",
             raw.stderr,
         )
     if raw.exit_code < 0:
--- a/src/gda/runner.py
+++ b/src/gda/runner.py
@@ -212,12 +212,14 @@ class UserDataUnwritable(OSError):
         cause: str,
         *,
         data_path: Optional[Path] = None,
+        data_location: Optional[str] = None,
         log_file: Optional[Path] = None,
         log_location: Optional[str] = None,
     ) -> None:
         super().__init__(cause)
         self.cause = cause
         self.data_path = data_path
+        self.data_location = data_location
         self.log_file = log_file
         self.log_location = log_location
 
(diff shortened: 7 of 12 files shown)


qinhaihong-red:
When launching three Godot instances in parallel, the three processes competed for the same `user://logs` path, triggering a log rotation crash in Godot 4.6.3 itself.

Can this PR resolve the issue?

qinhaihong-red:
Yes — for Godot instances launched through gda, this PR resolves exactly that failure, by two independent mechanisms:

1. **No shared log target by default.** Every one-shot Headless launch now injects a gda-owned `--log-file`, and under the default configuration each invocation gets its own private directory (`$TMPDIR/gda-log-<random>/godot.log`). Three parallel gda commands therefore write three distinct files — the contended `user://logs/godot.log` path is never used. This is pinned by a red-proven regression: the concurrency test asserts that no `user://logs` directory appears under the (fake) HOME data path at all; with the `--log-file` injection removed, it fails naming the shared logs directory the three processes would have fought over.

2. **Rotation is disabled even when a target IS shared.** `--log-file` makes the engine substitute its logger with `max_files = 1` — rotation off entirely (`main/main.cpp`, the `--log-file` branch). The 4.6.3 crash you hit lives in `RotatedFileLogger::rotate_file()`; with rotation disabled that code path is unreachable. So even if you pass the SAME `--user-data-root DIR` to three parallel invocations (they then share `DIR/logs/godot.log`), the rotation race cannot trigger — the worst case is interleaved log content in one file. For parallel runs, prefer the default (distinct temp targets) or distinct roots per invocation if you want clean per-run logs.

Boundaries, for precision:

- Covers every gda-launched one-shot headless process — the injection lives once in the Headless-launch primitive, so the sentinel ops, `script run`, and `export run` channels all inherit it.
- The daemon's Engine sessions were already immune (ADR-0022 passes a per-session `--log-file`).
- Godot processes started OUTSIDE gda keep the stock `user://logs` rotating logger and can still crash under parallel contention on 4.6.3 — the same `--log-file` remedy applies, but that is beyond this PR's reach.

This is acceptance criterion 5 of #653 ("concurrent invocations with the default configuration do not share a crashing log target"), verified in `tests/test_e2e_user_data.py` against the real engine.

qinhaihong-red:
**Changes requested.** Centralizing the one-shot log target in `Headless launch` is the right boundary and closes the shared-rotation crash, but the explicit root is not fully preflighted and several public contracts disagree with the exact behavior.

Reviewed refs:

- Head: `ddc76a16c217b86bcfd4bb1f7a31a55553fc6774`

## Domain architecture

### [P1] The preflight proves the log path, but not the redirected `user://` data path

`src/gda/runner.py:236`

With an explicit root, `user_data_placement()` creates `root` and `root/logs/godot.log`, but it does not create or probe the platform-derived data path. On macOS that path is `<root>/Library/Application Support`.

Exact-head reproduction used a root whose log directory was writable but whose derived data path was blocked by an existing file. `user_data_placement()` succeeded; real `script validate` then launched project code, emitted “Could not create directory …/Library/Application Support,” and exited `0` with a successful JSON result.

Impact: `--user-data-root` can report success while `user://` is unavailable. A persistence script can fail or spend its timeout even though the option exists to make this failure attributable before execution.

Minimum fix: when an explicit root is active, create and probe the platform-derived `data_path` before spawning, map failure through `UserDataUnwritable`, and add a real-engine negative arm asserting the typed refusal and no project execution.

### [P2] An explicit empty flag loses flag-over-environment precedence

`src/gda/runner.py:79,117`

`set_user_data_root("")` stores `None`, after which `$GDA_USER_DATA_ROOT` wins. This contradicts the documented `flag > env > default` contract and the established `--godot` / `--project` behavior, where an explicit empty value is an error rather than silent fallback.

Minimum fix: preserve “option absent” separately from “present but empty,” reject the empty flag, and add a flag-plus-environment precedence test.

### [P2] The authoritative Headless-launch definitions still describe the old contract

`CONTEXT.md:77`; `docs/adr/0031-headless-script-run-passthrough-execution.md:109`

The glossary still says the primitive builds exactly `[binary, --headless, *args]` and that `launch_failure` covers only missing binary and timeout. The exact head now injects an owned `--log-file` and adds `USER_DATA_UNWRITABLE`.

Impact: the architecture authority omits the new responsibility and failure branch, so a later channel or refactor can follow the documented contract and bypass the log owner.

Minimum fix: reconcile the `Headless launch` / `Raw run` glossary entries and add a short amendment or outcome to the relevant launch ADR. No new topology or standalone ADR is necessary.

The main boundary should be retained: sentinel, native export, and `script run` all delegate to one `gda.runner.launch`; the root option hands configuration downward without channel-specific policy; default launches use distinct temporary log targets; and gda-daemon retains separate ownership of its persistent Session log.

- `DDMA-696-C1`: one owner covers all three one-shot channels. Static call trace and focused real-engine coverage support it.
- `DDMA-696-C2`: default concurrent launches do not share the rotated log. Unique temporary placement and the six-process e2e support it.
- `DDMA-696-C3`: an explicit root makes both log and `user://` usable or refuses. The blocked derived-path counterexample currently disproves this.
- `DDMA-696-C4`: the daemon boundary remains independent. The live Session continues to own its own absolute `--log-file`.

## Standards

### [P3] Export guidance rejects a supported `pack` workflow

`src/gda/skill/SKILL.md:35-38`; `src/gda/cli.py:120-121`

The guidance says not to combine `--user-data-root` with `export run` because templates move. However, `src/gda/commands/export.py:519-539` explicitly exempts `--mode pack`, and `tests/test_e2e_user_data.py:406-447` proves that rooted pack export succeeds.

Minimum fix: state that release/debug exports need templates under the redirected root while `--mode pack` remains supported.

### [P3] The public error meaning misclassifies private-temp-log allocation failures

`src/gda/error_codes.py:90-94`; `docs/adr/0002-headless-structured-output-contract.md:146`; `src/gda/runner.py:228-249`

`user_data_unwritable` is defined as “Godot's user-data location is not writable,” but the default branch also emits it when `tempfile.mkdtemp()` cannot allocate the private log directory, even if Godot's user-data location is writable.

Impact: the stable code and diagnostics can direct remediation toward the wrong path. Broaden the registered meaning to an unusable log/user-data placement or give temp-log allocation a distinct verdict; test default-temp and explicit-root refusals separately.

## Spec

### [P1] AC1 and the exact default behavior disagree

`src/gda/runner.py:228-235`; `tests/test_e2e_user_data.py:197-246`

Issue #653 requires an unwritable application-data directory to produce a typed refusal before project code. The exact-head default instead redirects only the log, leaves `child_env=None`, and explicitly tests that `script validate` succeeds and `script run` executes project code under that restriction.

The alternative behavior avoids the crash, but the implementation and authoritative acceptance criterion still disagree. Either implement the pre-spawn refusal as written or amend #653 to accept default log-only redirection and successful command execution.

### [P2] Failure diagnostics do not name the resolved paths required by AC4

`src/gda/runner.py:256-284`

On macOS, the redirected data path resolves to `<root>/Library/Application Support`, but explicit-root diagnostics report only `user data: under <root>`. If private temporary-directory allocation fails, diagnostics say `log file: a private temporary directory`, not an attempted path.

Minimum fix: carry the computed data path and attempted log target into the refusal formatter and assert the platform-resolved values.

## Validation

- Unit/focused placement tests: `38 passed` on the architecture pass and `51 passed` on the Standards/Spec pass.
- Exact-head real-Godot user-data e2e: `8 passed`, including the six-process concurrency arm.
- The derived macOS data-path failure and explicit-empty flag precedence were reproduced.
- Static and focused runtime checks confirmed that sentinel, `script run`, and native export all reach the shared `launch()` primitive.
- `git diff --check` passed. Required CI is green; its Godot e2e job is skipped.
- Full e2e and unrestricted alternate-platform runs were not performed.

Summary: Domain architecture — 3 findings, worst P1. Standards — 2 findings, worst P3. Spec — 2 findings, worst P1.

qinhaihong-red:
Thanks — this was a genuinely valuable pass. **All seven findings adopted**, six in
code/docs at `6f225990` and one (F) as a formal amendment on the issue. Every
finding was reproduced before fixing, and every new guard was verified red-proof by
breaking the fix and re-running.

| # | Sev | Decision | Evidence | Where |
| A | P1 | Adopted | reproduced verbatim | `6f225990` |
| B | P2 | Adopted | reproduced | `6f225990` |
| C | P2 | Adopted | confirmed drift | `6f225990` |
| D | P3 | Adopted | your reading is right | `6f225990` |
| E | P2 | Adopted | confirmed | `6f225990` |
| F | P1 spec | Adopted via issue amendment | — | [#653 comment](https://github.com/aigengame/godot-agent/issues/653#issuecomment-5313363800) |
| G | P2 | Adopted | confirmed | `6f225990` |

### A (P1) — derived data path never preflighted

Reproduced exactly as described: a root with `<root>/Library` as a regular file
preflighted clean, the engine printed `Could not create directory …`, exited 0, and
`script run` reported **`exit_status: 0` SUCCESS** with `OPEN_OK=false`. The worst
possible outcome — a false success on the very promise the option makes.

Fixed by preflighting the derived path in the engine's own shape. The probe
**creates** rather than inspects, deliberately: `OS::ensure_user_data_dir` calls
`make_dir_recursive` on `<data_path>/app_userdata/<name>`, so the probe creates and
removes a throwaway directory there. A permission-bit check would have missed a
read-only mount, a full filesystem or an immutable flag. It runs only when an
explicit root is active — the default branch must not preflight the engine's own
location, since most commands never touch `user://`.

Same input now:

```
code = user_data_unwritable   category = environment   exit 127
gda:   binary:    …/Godot
gda:   user data: …/root/Library/Application Support (--user-data-root)
gda:   log file:  …/root/logs/godot.log
gda:   cause:     [Errno 20] Not a directory: '…/root/Library/Application Support'
```

Real-engine negative arm added (`test_a_root_whose_derived_data_path_is_blocked_is_refused`)
asserting the typed refusal **and** that the project never executed. Red-proof:
removing the probe fails it plus both unit arms.

### B (P2) — empty flag demoted to absent

Reproduced. `set_user_data_root` now stores the value verbatim, and the resolver
distinguishes option-absent from present-but-empty: an empty **flag** raises (your
`--godot` precedent — "an explicit value is a deliberate choice; an empty one is a
mistake we surface"), while an empty **environment** value still falls through to the
default, since unset and blank are the same intent there. Three tests, including the
flag-plus-environment precedence case; red-proof against restoring `value or None`.

### G (P2) — diagnostics named paraphrases, not paths

Right, and it defeated AC 4's purpose. `UserDataUnwritable` now carries the paths it
actually resolved and attempted, and the formatter reports those: the
platform-derived data path (not "under `<root>`") and the attempted log target, or
`tempfile.gettempdir()` when the failure preceded having a log path at all. Asserted
against `engine_data_path(data_path_env(root))` rather than a literal.

### E (P2) — registry meaning too narrow

Agreed, the shared code was misdirecting remediation. Broadened in both authorities
(`error_codes.py` + the ADR-0002 row) to the *placement* being unusable. Added a
**paired** test that the two refusal shapes blame different directories — the
default-temp refusal must not name the root, and the explicit-root refusal must not
claim gda is "not redirecting user://".

### C (P2) — authoritative contract lagged

Confirmed. CONTEXT.md's `Headless launch` and `Raw run` entries reconciled, plus a new
`User-data placement` term to own the concept, and a dated **append-only** amendment
on ADR-0031's launch-primitive reuse (zero accepted-prose lines touched). No new ADR.

### D (P3) — pack-mode wording

You're right, and my own e2e proves rooted `--mode pack` works. Both the Skill and the
option help now scope the caveat to release/debug ("unless you place them there") and
state that `--mode pack` needs none.

### F (P1 spec) — AC 1 deviation

Taken via your second branch, as an issue-side amendment rather than a code change:
[#653 comment](https://github.com/aigengame/godot-agent/issues/653#issuecomment-5313363800).
It records the stronger form (default isolated target → no crash path, commands
succeed; the typed refusal covers gda's own placement being unusable), the reasoning
that the crash is the *logger* rather than the data directory, and two further
corrections to the issue text — that `gda info` never crashed (projectless, so
`Main::setup` builds no `RotatedFileLogger`), and that the persistence half landed via
the strong branch. Cross-linked from the PR body's deviation note.

### Validation

Fast suite **1333 passed** (and again under `GITHUB_ACTIONS=true`); `ruff check`,
`ruff format --check`, `pyright` 0 errors; user-data e2e **9 passed** foreground under
the shared lock, permissions restored in teardown. Surface re-derived: `gda schema`
and `gda script run --help` still **byte-identical** to the pre-change baseline; the
only surface deltas are the root `--help` option text and the two `gda skill` hunks,
both enumerated in the PR body.

qinhaihong-red:
## Follow-up recheck

**Changes requested — the substantive launch/placement fixes are closed, with one P3 diagnostics-contract residue remaining.** I rechecked the review response and issue amendment against exact current head `ad941a55a9cd04967b4b5354bdbfd93774786a0e`.

The following review items are now closed:

- the platform-derived `user://` path is preflighted for an explicit root;
- an explicit empty `--user-data-root` is no longer collapsed into absence or replaced by `GDA_USER_DATA_ROOT`;
- `CONTEXT.md` and ADR-0031 describe the current user-data placement contract;
- the export guidance correctly distinguishes template-dependent release/debug runs from supported pack runs;
- the registry/ADR wording now covers both log and user-data placement failures;
- the original AC1 conflict is resolved by the formal issue amendment; and
- real placement failures report the resolved Godot binary, platform data path, and attempted log target.

### Remaining finding

- **[P3] Align the empty-root refusal with the amended diagnostics claim — `src/gda/runner.py:432-447`, `src/gda/errors.py:203-208`.** `GDA_USER_DATA_ROOT=/from/env gda --user-data-root '' info --json` now correctly refuses before launch with `user_data_unwritable`, does not fall back to the environment value, and reports `cause: explicit --user-data-root is empty`. However, this `ValueError` branch bypasses `_user_data_unwritable_stderr`, so it reports neither the resolved Godot binary nor user-data/log fields. That conflicts with the issue amendment's statement that these refusals name the binary, platform-resolved user-data path, and attempted log target, and with the inline registry comment saying every `USER_DATA_UNWRITABLE` diagnostic contains all three. This does not reopen the precedence or refusal behavior; those are correct. **Minimal fix direction:** either route this branch through the shared formatter and explicitly render unavailable fields as `unresolved` / `not attempted`, or narrow the amendment, PR description, and inline comment so the three-path guarantee applies only after placement targets exist.

Validation on the immutable current-head snapshot:

- `106` focused tests passed;
- `9` real-Godot user-data e2e tests passed;
- focused Ruff, Ruff format, Pyright, and `git diff --check` passed;
- required current CI checks are green; the CI Godot e2e job is skipped.

I did not rerun the full test suite in this follow-up.

qinhaihong-red:
Follow-up-recheck response — the P3 diagnostics residue is closed on `33e23898` (handled directly by the orchestrator).

| [P3] The empty-root `ValueError` branch bypassed `_user_data_unwritable_stderr`, so its refusal named neither the resolved binary nor the user-data/log fields — conflicting with the amended three-path guarantee and the registry's inline comment. | Fully adopt (the route-through-the-shared-formatter direction) | Reproduced with your exact invocation (`GDA_USER_DATA_ROOT=/from/env gda --user-data-root '' info --json`): the old branch emitted only the cause line. | The branch now constructs a `UserDataUnwritable` and routes through the shared formatter; `UserDataUnwritable` gains a `data_location` string fallback mirroring the existing `log_location`, and the formatter renders unavailable fields explicitly — `user data: unresolved (--user-data-root is empty)`, `log file: not attempted (no placement was prepared)` — with a case-specific remedy ("pass a non-empty --user-data-root directory…"). The three-path guarantee now holds unconditionally, so no wording narrowing was needed. | Real CLI at `33e23898`: the same invocation returns `user_data_unwritable` / exit 127 with all three fields named. The empty-flag unit test strengthened to assert binary + both explicit markers + the remedy; `tests/test_user_data_placement.py` 31 passed; full fast suite 1333; ruff/format/pyright clean; `gda schema` byte-identical (`06f2669e…`). |

PR #702, opened 2026-08-17
Promote the gda dogfooding wave 2 slices to main

qinhaihong-red:
## Wave 2 promotion — gda dogfooding milestone

Promotes the wave-2 slices from `feat/gda-dogfooding-dev` to `main`. Tracks the
review-and-merge process for the whole wave.

**Merge method note:** merge commit (NOT squash) — the dev branch carries one clean squashed
commit per issue; squashing the promotion would break release-please's derivation on `main`.
The squash-only gate lives in the branch ruleset (`godot-agent-branch-rules`), temporarily
widened and restored at promotion time.

### Merge order and status

| # | Slice PR | Issue | Status |
| --- | --- | --- | --- |
| 1 | #695 — fix(gda): refuse a script validate outside the resolved project and report the root | #658 | ✅ merged `ca8d5ddc` |
| 2 | #693 — feat(gda): accept both script-path forms for script run (ADR-0031 amendment) | #675 (+ subsumes #699) | ✅ merged `b2910854` (integration fix: containment-layer tilde totality, found by the serial-merge gate) |
| 3 | #692 — feat(gda): report structured install provenance from --version --json | #659 | ✅ merged `1c15c257` (rebased + i18n re-stamped after #695; translations rebuilt three-way) |
| 4 | #696 — fix(gda): preflight and redirect the user:// root for the Headless launch | #653 | ✅ merged `9fb05e9e` (three-point rebase after #692; flag-sentinel test green; duplicate skill bullet from keep-both audited out) |
| 5 | #694 — fix(gda): classify sandbox-denied window-server access (ADR-0004/0002 amendments) | #667 | ✅ merged `1f3b96e0` (registry hunks auto-merged clean) |

### Review trail

Every slice: implementation → internal adversarial review + adjudicated fix rounds → external
adversarial reviews (codex channel) with per-claim reply tables → rechecks until closed. See
each PR's thread.

### Wave validation

- Per-PR CI green on every final head; targeted real-engine e2e per slice under the serialized lock.
- Full-e2e CI: dispatched on the dev-branch ref after the last merge — run
  [32011771629](https://github.com/aigengame/godot-agent/actions/runs/32011771629) (`run-e2e: true`) — **completed: success** (full suite incl. the real-Godot e2e tier, on the integrated dev branch at `1f3b96e0`).

### Follow-ups created during the wave

#697 (ADR-0006 derive-vs-refuse, needs-triage), #698/#700/#701 (ready-for-agent); #699 subsumed
[...]

PR #705, opened 2026-08-17
feat(gda): configurable script-run timeout that preserves captured output and fails fast on aborted scripts (#655)

### LINKED ISSUE #655: script run: configurable timeout that preserves captured output and fails fast on aborted scripts

qinhaihong-red:
Closes #655

> **Base-history note.** This slice was branched from `feat/gda-dogfooding-dev` at `08cf5738`. That
> branch was auto-deleted when wave 2 was promoted (PR #702), so the PR was opened against `main` as
> the only resolvable equivalent; the orchestrator has since recreated the integration branch at
> main's tip and retargeted this PR. **Base is now `feat/gda-dogfooding-dev` @ `190367d7`** and the
> diff is unchanged throughout.

## What changed

**The primitive gains a second capture strategy** (`src/gda/runner.py`), selected by a caller passing a `LaunchWatch`. Both pipes are read on their own threads (portable — a pipe is not selectable on Windows, and the threads are also what stops a child that outgrows the OS pipe buffer from deadlocking) through one incremental UTF-8 decoder per stream, so a chunk boundary inside a multi-byte sequence does not corrupt a character. A timeout then **preserves** the capture, the wall clock is measured, and the watch can end the run early.

**`script run` gains** (`src/gda/commands/script.py`):
- `--timeout <s>` — per-invocation ceiling, default the previous 120s, honored and reflected in the failure.
- a richer `launch_timeout` envelope — the captured partial output (tail-capped, cap stated), elapsed seconds, one enumerated termination phase, and the recognized script errors.
- `--completion-marker <line>` — an opt-in **liveness contract** (final shape after the recheck round below): declaring it asserts the script keeps producing output until the marker line says it finished, so when a recognized error **attributable to the entry script** appears on stderr, the declared marker has **not**, and the run then goes silent for a fixed 3s, gda ends the run and reports the new `script_aborted` with the captured error — deterministically, on every platform.

## Decisions, with rationale

**Timeout code: REUSED `launch_timeout`, not minted.** The condition is exactly what that code names — Godot launched and did not return before the timeout — and ADR-0031 already recorded this path under it. Per ADR-0002 the code is reused and the message discriminates. Exit `124` unchanged, so no shell consumer's branch moves.

**Abort code: MINTED `script_aborted`** (operation / classifier / exit 4). No registered code names this condition:
- `launch_timeout` would be untrue — gda did not wait for the timeout, it decided not to. Reporting it would erase the very distinction the abort exists to make.
- `script_failed` means "ran to completion and chose a non-zero status", is documented as *never reported without `--strict`*, and sends an agent to read an exit status that does not exist here. Reusing it would have falsified that recorded guarantee.
- `script_compile_failed` / `script_not_found` say the entry never **loaded**; here it loaded and ran, then died partway.

I own `error_codes.py` + the ADR-0002 table row this wave; the registry equality test (`test_adr_registry_matches_python_authoritative_registry`) is green, and as a CLASSIFIER-source code it is correctly **not** GDScript-mirrored (`test_gdscript_operation_error_codes_mirror_python_operation_subset` green).

**No envelope fields — deliberate.** Elapsed, phase and the error lines are prose in `message` / `diagnostics`. Promoting them changes ADR-0004's uniform failure ABI, which **#687 owns**; ADR-0031's #651 amendment already records that this issue adopts #687's outcome. Stated in the code comments (`_ended_run_diagnostics`, `_classify_ended_run`, `TerminationPhase`) and in the ADR note.

**The abort enforces a declared contract; it does not detect death.** A GDScript runtime error aborts only the function that raised it, so a script can survive one and still reach its `quit()` — this repo's e2e suite pins exactly such a run as a **success** — and whether a silent run can still finish is not observable from outside the process (the recheck round below falsified both silence-only and CPU-idleness rules on real paired runs). Declaring the marker is what settles it: the caller asserts the script keeps printing until the marker line, so error + marker absent + a full silence window *is* the run being dead, by declaration. Any output line resets the window; the escape hatches (print progress, or omit the marker) are stated in the help, the ADR, and the failure message itself.

**Parsers reused, never re-implemented.** The captured errors go through the existing `gda.engine_log` → `gda.script_errors.parse_script_errors` stack, and are rendered in the same `<kind>: <path>:<line>: <message>` layout the human renderer already uses. A capture whose errors would satisfy the #651 never-ran verdict is deliberately **not** re-verdicted: ADR-0031 records that shape as failing "by another route", and narrowing it is a separate decision (noted, not done).

**Marker is caller-declared and is not the ADR-0002 sentinel.** ADR-0031 rejected imposing a gda-owned sentinel wrapper on a user-authored entry script. Nothing is required of or injected into the script; with no marker declared the watch never aborts. The distinction is made explicit in the CLI help, the params description, the class docstring, the ADR note, and a new `CONTEXT.md` term whose `_Avoid_` list leads with "sentinel".

**Termination phases deviate from the issue's illustrative set, with a measured reason.** The issue suggested `launched` / `script_started` / `killed_at_timeout`. Measuring Godot 4.6.3 showed it prints its version banner to stdout within ~0.1s of a normal spawn, so "output arrived" does **not** prove the *script* started — `script_started` would have been a false name. The set is `launched` / `output_seen` / `aborted_on_error`: both timeout phases *are* "killed at timeout", so that member carries nothing the code does not, while what a reader cannot infer is whether the run got anywhere.

## Sentinel and export channels: unchanged, proven

They pass no watch, so they keep the **buffered** strategy — output discarded on timeout, the `gda: <label> timed out after <n>s` diagnostic (published prose in their envelopes) byte-identical, and no `elapsed_seconds`. Moving them across is the follow-up the issue names; it is a one-line switch at the call site, not a second mechanism, and it is out of scope here because their timeout results are part of their own error envelopes. What is **not** duplicated is the failure mapping: both strategies return through one shared `_timeout_result` / `_not_found_result`.

Proof:
- **`gda schema` diff**: every differing leaf key is under `commands[59]` (`script run`) — **2 added input properties** (`timeout`, `completion_marker`), which is 12 added leaf keys, plus its own `description`. Zero removals, no other command touched.
- **targeted unit guard**: `test_the_buffered_path_still_discards_output_and_keeps_its_diagnostic` pins the discard, the exact diagnostic string and the absent clock; `test_streaming_maps_a_missing_binary_to_the_same_not_found_result` pins that both strategies share one mapping.
- **real-engine e2e**: `test_e2e_export_run.py`, `test_e2e_info.py`, `test_e2e_script.py`, `test_e2e_user_data.py` — 80 passed, 1 pre-existing documented skip (Linux-only templates arm).

## Every intended surface diff

| Surface | Diff |
| --- | --- |
| `gda script run --help` | `+ --timeout FLOAT` (default 120.0, help names the 3 termination phases); `+ --completion-marker TEXT` (help names the 3.0s silence window and the two conditional-guarantee caveats); one added docstring paragraph. Option column reflows — no other text changes. |
| `gda script run --schema` / `gda schema` | `input.properties` gains **2** properties — `timeout` (number, `exclusiveMinimum: 0`, default 120.0) and `completion_marker` (string\|null, `minLength: 1`, default null) — which is 12 added leaf keys; both descriptions publish the phase set and the 3.0s window; `description` updated. Output + error schemas unchanged. 0 removed, 1 changed, no other command touched. |
| Error registry | `+ script_aborted` (operation / classifier / exit 4) in `error_codes.py` and the ADR-0002 table. |
| `RunResult` (internal) | `+ elapsed_seconds`; `LaunchFailure` `+ ABORTED`. Not public — `ScriptRunResult` is built field-by-field and is unchanged. |
| `CONTEXT.md` | `Headless launch` + `Raw run` updated; new `Completion marker` term. |
| `docs/adr/0031` | Dated append-only amendment. Zero edits to accepted prose. |
| `SKILL.md` | `script` row gains both options. |
| `README.md` | **Untouched** — it does not document `script run` at all today, so no i18n restamp. That gap is pre-existing and reported below. |

## Validation

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1528 passed, 444 deselected |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1528 passed, 444 deselected |
| `uv run ruff format --check` | 436 files already formatted |
| e2e `tests/test_e2e_script_run.py` (real Godot, serial under the shared lock) | **39 passed** in 25s (36 baseline + 3 new arms) |
| e2e sentinel/export regression (4 suites, same lock) | 80 passed, 1 pre-existing skip |

**AC status** — all four met:

1. *Abort-before-quit returns the captured script error within a stated bound when a marker is declared* — `test_script_run_returns_the_captured_error_of_an_aborted_run` (real engine): `script_aborted`, under 30s against a 60s ceiling, carrying `Nonexistent function 'missing_method'` and `runtime_error: res://aborts.gd:6`. The bound is stated in the message (the marker, the 3s silence window, and the ceiling that was not reached).
2. *A run exceeding `--timeout` returns captured partial output (cap stated) + elapsed + one enumerated phase* — `test_script_run_timeout_returns_partial_output_elapsed_and_a_phase` (real engine): `SUITE START` present and `SUITE DONE` absent, `elapsed 4.…s`, `termination phase 'output_seen'`, `16384 characters`.
3. *`--timeout` honored and reflected* — `test_the_caller_timeout_is_what_reaches_the_launch`, `test_timeout_reaches_the_launch_from_argv`, `test_timeout_and_marker_reach_the_launch_through_params_json`, plus the reflected value asserted in both envelopes.
4. *Regressions cover the aborting-entry-with-marker and slow-but-healthy-small-timeout cases* — both, at unit level (`test_script_run_operation.py`) and against the real engine, plus a third arm proving the no-marker path still waits the ceiling out.

## Reported for follow-up (not done here)

1. **Sentinel + export channels still discard output on timeout** — the issue names this as follow-up. Needs an issue.
2. **README does not document `gda script run`** at all (the `script` table has no `run` row), a pre-existing gap from #343. Filling it would drag the three translations through the i18n restamp, so it is out of scope here. Needs an issue.
3. **Re-verdicting a timeout capture** — a captured `script_missing` / `not_a_main_loop` error under the timeout envelope could arguably become the #651 entry-load verdict instead. ADR-0031 records that shape as failing "by another route", so narrowing it is its own decision, not a silent change here.

---

## Review round (independent review of the first push)

Head is now `4f4f9754`. One HIGH defect and the small fixes, all adopted.

**F1 — HIGH, orphaned engine (fixed, red-proofed).** The streaming capture left the child running on any exception out of its poll loop: `_end_process` sat on the happy path inside the `try`, and the `finally` only closed pipes. The buffered strategy gets this free (`subprocess.run` kills its child when an exception leaves its `with` block), so this was a regression against it — and not cosmetic, since the runs this strategy exists for are exactly the ones that do not stop on their own, so an orphan idles forever and repeated interruptions accumulate engines contending over `user://`.

The whole teardown moved into the `finally`, preserving the order that was already load-bearing: reap (so the pipes reach EOF and the readers end) → join the readers → close the pipes. The reap is conditional, so the normal path pays nothing.

*Red proof*: `test_streaming_reaps_the_child_when_the_poll_loop_is_interrupted` was written first and failed on both arms against the pre-fix code — `AssertionError: the engine outlived the interrupted launch` for `[RuntimeError]` and `[KeyboardInterrupt]` — then passed after the fix. It wraps the real `subprocess.Popen` to hold the child and assert its fate, and kills any survivor so the red run does not leak the orphan it forbids. `KeyboardInterrupt` is covered explicitly: it is a `BaseException` that would slip past `except Exception`, and it is the case that bites when gda sits in its own process group and the signal never reaches the engine.

**F2 (fixed)** — `_elapsed` returned exactly `0.0` while its docstring promised a ceiling fallback. The fallback is wired through as `at_least`, and each call site passes the truthful lower bound its own rule guarantees: the ceiling for a timeout, the silence window for an abort. Pinned by `test_an_unmeasured_run_falls_back_to_its_rule_s_lower_bound`.

**F3 (fixed)** — the capture strategies are now `buffered` / `streaming`. "One-shot" already means "one process per call" in CONTEXT.md (`Headless operation`, `Live operation`, `Headless launch`), so overloading it inside the shared-language authority was a real collision. Renamed in CONTEXT.md, the `launch` / `_spawn` / `_spawn_streamed` docs, the ADR-0031 amendment, the test comments and the test name; every surviving "one-shot" is the pre-existing process sense, checked by grep.

**F4 / F5 (fixed)** — base-history note rewritten as resolved history above; the "11 added input properties" over-count corrected to 2 properties = 11 leaf keys, in both places.

**F6 / F7 (fixed)** — the phase set (`launched` / `output_seen` / `aborted_on_error`) is now named in the `--timeout` help and the `timeout` params description, so the members are discoverable while prose stays the carrier. The `completion_marker` params description now carries the 3.0s figure (JSON/MCP callers read only the schema) plus both caveats: a run that keeps printing after a fatal error never goes silent and so waits out `--timeout` **by design** — that is what keeps a script which survives an error and still reaches its own `quit()` a success — and a `--timeout` at or below the window leaves the marker inert.

**Nits, all adopted** — F9 `_complete_lines` now states the real reason (only `split("\n")` distinguishes a complete final line from a partial one, which is what `parts.pop()` holds back); F10 the aborted result's `exit_code` is `0`, since `EXIT_OPERATION` implied a mapping onto `script_aborted`'s exit 4 that does not exist and the field is never read (the unused import is gone); F11 the boundary `assert completion_marker is not None` is replaced by a structured degradation — the builder names the condition without the string, so an unreachable state yields vaguer prose rather than a traceback that `-O` would strip — pinned by `test_the_abort_envelope_names_the_condition_without_a_marker_string`; F12 CONTEXT.md's wrap orphans are reflowed and "`elapsed_seconds` is absent" is now "is `None`".

The ADR-0031 amendment also gains the reap guarantee, since it is a contract of the streaming strategy and not merely an implementation detail.

### Re-run gates

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1532 passed, 444 deselected |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1532 passed, 444 deselected |
| `ruff check` / `ruff format --check` / `pyright` | passed / 436 formatted / 0 errors |
| e2e `test_e2e_script_run.py` (foreground, shared lock) | 39 passed in 25s |
| `test_launch.py` incl. the F1 regression (real children, under the lock) | 21 passed |
| e2e buffered-channel regression (`export_run`, `info`) | 9 passed, 1 pre-existing skip |

### Re-derived surface enumeration (vs the pre-change baseline)

Unchanged in shape after F6/F7 added description text: **only `commands[59]` (`script run`) differs** — 12 added leaf keys under its **2** new input properties, `description` changed, **0 removed**, no other command touched. `--help` adds the two options (their help now naming the phase set, the 3.0s window and the caveats) plus one docstring paragraph.

---

## External adversarial review round

Head is now `245cc129`. Five findings (seven axes), all adopted — full mapping in [the reply comment](#issuecomment-5316627690). Headlines:

- **The early abort could kill a run that would have completed.** It armed on any parsed diagnostic plus silence, and neither is evidence a run cannot finish: a GDScript runtime error aborts only the raising function, so a script can survive one — *inside the entry script itself* — and then compute for a long time without printing. Now four conditions must hold: the error must name the **entry script**, the marker must be absent, the run must be silent for a window, and the process must consume **no measurable CPU** across a further window. Measured against Godot 4.6.3: an aborted entry idles at ~0.01s CPU/wall-second, a quiet worker burns ~1.0s. Where CPU cannot be measured, gda does not abort. Note attribution alone would *not* have saved the reviewed run — the CPU check is the decisive gate.
- **Capture setup sat outside the reaping boundary**, orphaning the child on a setup failure — the same defect `4f4f9754` fixed for the loop, reached through setup. Constructors moved inside; teardown tolerates a partially initialized pair.
- **A non-finite `--timeout` was accepted on both routes**, making the ceiling unreachable and the run unbounded. Refused by the shared model and an argv guard; since JSON Schema cannot express finiteness, the description says so.
- **The output cap counted characters, not bytes** (~3x over for multibyte). Now a 16 KiB UTF-8 byte tail trimmed to a character boundary.
- **The marker matched as a substring**, so `NOT DONE YET` disarmed the abort. Now whole-line equality; blank markers refused on both routes.

### Re-derived enumeration

Only `commands[59]` (`script run`) differs: **2 input properties** = **12 added leaf keys** (+1 vs the previous round for `completion_marker.pattern`), `description` changed, **0 removed**, no other command touched.

### Gates

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` (+ `GITHUB_ACTIONS=true`) | 1561 passed each |
| `ruff check` / `ruff format --check` / `pyright` | passed / 436 formatted / 0 errors |
| e2e `test_e2e_script_run.py` (foreground, shared lock) | 41 passed in 45s |
| `test_launch.py` — reap, setup-failure ×2, CPU probe | 25 passed |
| e2e buffered-channel regression | 9 passed, 1 pre-existing skip |

---

## Recheck round: the marker becomes a declared liveness contract

Head is now `b36a32ff`. The recheck (pinned to `245cc129`) proved the CPU-idleness gate wrong in **both directions**: a run blocked in a legitimate wait (`OS.execute`, an `await`) consumes no CPU while alive, so idleness cannot distinguish it from a corpse; and a host where CPU time cannot be read at all (restricted macOS, Windows without `ps`) silently lost the early abort entirely — a genuinely dead entry ran to the full ceiling as `launch_timeout`, forfeiting the issue's seconds-bound AC on platforms the repository promises. The root fact: **whether a run that printed an error can still finish is not decidable by observation from outside the process**, by gda or by an engine signal (during an `await`, the main loop iterates just as an abandoned one does), so every observational rule inherits one of the two failure directions.

The resolution is a deterministic caller contract, now **explicitly defined by issue #655's dated amendment** (2026-08-18 — the original "fatal script error" wording is not decidable by observation, so the requirement itself was amended rather than reinterpreted): declaring `--completion-marker` asserts the script keeps producing output until the marker line says it finished, so an **entry-attributable error + marker absent + one full 3s silence window** ends the run — a pure function of the observed text and the clock, identical on every platform. The contract's price is stated where it is declared, not hidden: a script that survives an entry error and then works past the bound in total silence is ended even though it would have finished; its escape hatches (print progress — any line resets the window — or omit the marker) are named in the help, the ADR amendment, and the failure message. `cpu_seconds` and its `/proc`-and-`ps` plumbing are **deleted**, `LaunchWatch.observe` returns to a pure `(text, elapsed)` signature, the abort lands after ~one window (~3s) instead of two, and the ADR amendment records the falsified alternatives so the probe cannot come back as an "improvement".

The e2e suite now pins **both directions of the contract** against the real engine: the silent survivor is ended by contract with the marker declared (red-proofed: fails against the previous implementation) and completes untouched without it; the progress-printing survivor completes **with** the marker declared; the dead-entry abort still lands in seconds (bound tightened to <10s against a 120s ceiling).

### Re-derived enumeration

Only `commands[59]` (`script run`) differs from the pre-#655 baseline, and this round changes **description text only**: `commands[59].description` and `input.properties.completion_marker.description` (contract wording). **0 keys added, 0 removed** vs the previous head; no other command touched. The bundled `SKILL.md` `script` row gains the contract duty ("keeps printing until that line").

### Gates

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` (+ `GITHUB_ACTIONS=true`) | 1559 passed each (−2: deleted CPU-probe tests) |
| `ruff check` / `ruff format --check` / `pyright` | passed / 436 formatted / 0 errors |
| e2e `test_e2e_script_run.py` (foreground, shared lock) | 43 passed in 45s |
| Red-proof of the contract arm against `245cc129` | failed as expected (survivor spared by the CPU gate) |

**Specification authority (final recheck):** head `d3fa8397`. Issue #655 now carries an explicit dated amendment defining the marker contract, its stated price, both escape hatches, and the amended acceptance criteria; the ADR amendment and the watch docstring cite that amendment instead of reading the contract into the issue's original "fatal error" sentence.


qinhaihong-red:
## Adversarial review — changes requested

Reviewed the exact three-dot range `190367d7ba98816dd0246bd4724f3d115dfaec59...33cda1afb7f5c33e883944d894a902886ca00b9b` (merge-base `08cf573817e48087b36417c247943a2641a89985`) against issue #655, `CONTEXT.md`, the accepted ADRs, and repository rules. I found five distinct open problems. The requested axes are kept separate below; repeated findings are independent corroboration.

## Domain architecture

- **[P2] Put capture initialization inside the child-reaping boundary — `src/gda/runner.py:800-807`.** After `Popen` succeeds, both `_StreamCapture` constructors, including `Thread.start()`, run before the teardown `try/finally`. If either raises, the child remains alive. An exact-head injected setup failure reproduced `child_alive_after_setup_exception=True`. This violates ADR-0031's unconditional no-outliving guarantee. **Minimal direction:** establish the lifecycle `try/finally` immediately after `Popen`, support partially initialized captures, and add first-/second-capture setup-failure tests.

- **[P2] Reject non-finite timeout values — `src/gda/commands/script.py:646-648,2022-2023`; `src/gda/runner.py:820`.** `Field(gt=0)` and the argv guard accept positive infinity. Both `--timeout inf` and params JSON containing `Infinity` reached launch dispatch on the exact head; `elapsed >= timeout` can then never become true. **Minimal direction:** require a finite positive value in the shared params model and keep an explicit finite-value argv guard, with argv and params-JSON regression tests.

- **[P3] Reconcile the output-cap unit with the authoritative brief — `src/gda/errors.py:588-600,624-629`; ADR-0031:315-320.** Issue #655 requires a fixed last-N-**KiB** cap, but the implementation and new ADR amendment redefine it as 16,384 **characters**. Exact-head reproduction retained 16,384 `你` characters, or 49,152 UTF-8 bytes. **Minimal direction:** enforce a UTF-8 byte/KiB cap while retaining valid text, or explicitly approve and reconcile a specification change across the issue and ADR.

The modular direction should otherwise be retained: `runner` owns process/pipe/deadline/termination mechanics; the `script` vertical slice owns marker policy, termination phases, and script-specific classification; `LaunchWatch` preserves the downward dependency; `script_errors` and `error_codes` remain the single parser and failure-code authorities. No new context, service, repository, event bus, or broader abstraction is justified.

Load-bearing claims `DA-705-1` (complete streaming-process lifecycle ownership) and `DA-705-2` (every configured timeout is a finite lifecycle bound) are currently disproved by the first two findings. Marker-policy ownership and single-authority error classification remain open, with the current focused evidence supporting them.

## Standards

1. **[P2] Match the declared completion-marker line, not a substring — `src/gda/commands/script.py:992`.** `any(self._marker in ln for ln in lines)` treats marker `DONE` as present after `NOT DONE YET`. A later recognized error plus silence then waits for the full timeout instead of returning `script_aborted`. This conflicts with the Completion marker definition in `CONTEXT.md:128-139` and ADR-0031:334-343. Compare complete lines and add a containing-line negative test.

2. **[P2] The documented bounded-launch invariant admits infinity — `src/gda/commands/script.py:646-648,2022-2023`.** This independently corroborates the DDMA timeout finding. Reject non-finite values on both public input routes.

No additional meaningful Fowler-baseline smell was found. The large changes otherwise follow the repository's accepted vertical-slice and deep-module ownership.

## Spec

- **[P2] The early-abort policy can kill a run that would complete — `src/gda/commands/script.py:983-998`.** Issue #655 asks to terminate when the *entry script aborts* after a fatal script error. The watch instead arms on any `parse_script_errors()` result and treats three seconds of output silence as proof of abortion, although the shared parser includes recoverable runtime/resource-load errors. A real Godot 4.6.3 paired reproduction used the same script in both arms: with a declared marker, gda killed it after 3.29s as `script_aborted`; without the marker, it completed its quiet work, printed `SUITE DONE`, and exited 0 after about 5s. **Minimal direction:** require evidence that the entry script cannot continue rather than any parsed diagnostic plus silence, and add this silent-survivor case as a regression test.

- **[P3] The tail cap is characters rather than the required KiB — `src/gda/errors.py:588-600,624-629`.** This independently corroborates the DDMA authority conflict; multibyte output exceeds the promised byte bound.

Validation on exact head: `git diff --check` passed; Ruff passed; Pyright reported zero errors; the full non-e2e suite reported `1530 passed, 2 skipped, 444 deselected`; and three targeted real-Godot e2e acceptance arms passed. GitHub CI is green except that the Godot e2e job is skipped, so it does not cover the real-engine counterexample above.

**Review summary:** Domain architecture: 3 findings, worst P2. Standards: 2 findings, worst P2. Spec: 2 findings, worst P2.

qinhaihong-red:
Thanks — all five findings (seven axes) adopted, head `245cc129`. Two of them were reproduced against the real engine before I touched anything, and one of your conclusions turned out to be stronger than stated. Mapping below.

---

### 1 — P2, capture init outside the reaping boundary → **fixed**

Confirmed: both `_StreamCapture` constructors (each starting a reader thread) ran after `Popen` but before the lifecycle `try`, so a setup failure orphaned the child — the same defect `4f4f9754` fixed for the loop, reached through setup instead. It contradicted the guarantee the ADR amendment had just recorded, which makes it worse than a leak.

The constructors moved inside the boundary, and the teardown now tolerates a partially initialized pair: reap → join whatever captures exist → close the pipes (always closable, since `Popen` gave us both).

*Evidence*: `test_streaming_reaps_the_child_when_capture_setup_fails[1|2]` — parametrized over both positions because they fail differently: failing on the **first** leaves nothing to join, failing on the **second** leaves a live stdout reader the teardown must still drain before closing. Both were **red before the fix** (`the engine outlived a capture-setup failure on #N`). `245cc129`.

### 2 — P2 ×2, non-finite `--timeout` → **fixed on both routes**

Confirmed on both axes: `Field(gt=0)` and the argv guard each pass `inf`, after which `elapsed >= timeout` is unsatisfiable and the run gda advertises as *bounded* is unbounded — the precise inversion of the option's purpose.

Enforced in the shared params model (`allow_inf_nan=False`, so argv and `--params-json` cannot diverge, ADR-0015) plus an explicit `math.isfinite` argv guard. `nan` is covered by the same check (it compares false against everything and fails identically).

One honest limitation worth recording: **JSON Schema cannot express finiteness**, so a schema-driven client sees only `exclusiveMinimum: 0`. Rather than fake it with an arbitrary `maximum`, the `timeout` description now states the finite requirement, so the constraint is at least discoverable in the published schema.

*Evidence*: `test_a_non_finite_timeout_is_refused_on_argv[inf|Infinity|-inf|nan|1e400]` (exit 2, launch never reached) and `test_a_non_finite_timeout_is_refused_through_params_json[Infinity|-Infinity|NaN|1e400]` (structured `invalid_params`). `1e400` is included because it reaches `inf` without any JSON extension. `245cc129`.

### 3 — P3 ×2, character cap vs last-N-KiB → **fixed, byte cap adopted**

Confirmed: 16,384 CJK characters encode to 49,152 bytes, so a bound meant to keep an inline JSON payload small silently did not for non-ASCII output.

Adopted your direction: `SCRIPT_OUTPUT_TAIL_CAP_BYTES = 16 KiB`, encode → keep the byte tail → decode. The boundary trim uses `errors="ignore"` rather than `"replace"`, deliberately: the truncation is *gda's own doing*, so emitting `U+FFFD` would misreport the engine's output as malformed. Anything genuinely malformed was already replaced when the capture was decoded and survives unchanged. Message wording is now `16384 UTF-8 bytes (16 KiB)`; the ADR amendment text is synced.

*Evidence*: `test_the_output_cap_bounds_utf8_BYTES_not_characters` (asserts the exact surviving character count, the absence of `U+FFFD`, and the byte bound) and `test_output_within_the_cap_is_never_re_encoded`. `245cc129`.

### 4 — P2, substring marker match → **fixed**

Confirmed: `any(marker in line ...)` made `NOT DONE YET` satisfy the marker `DONE`, disarming the abort for a run that had in fact died. Since CONTEXT.md and the ADR both define the marker as a **line** the script prints, the line is the right unit.

Now whole-line equality with both sides stripped. That change creates a hazard substring matching did not have in the same way — a blank marker equals every blank line — so blank is refused on **both** input routes (`pattern: \S` in the model, an argv guard) and treated as undeclared in the watch itself, so no third call site can reintroduce it.

*Evidence*: `test_a_line_merely_containing_the_marker_does_not_disarm_the_abort` (the containing-line negative you asked for: the run is still correctly ended), `test_the_marker_matches_its_line_ignoring_surrounding_whitespace`, `test_a_blank_marker_is_treated_as_undeclared`, plus argv/JSON refusal arms. `245cc129`.

### 5 — P2, the abort could kill a run that would complete → **fixed; and your finding is stronger than stated**

Reproduced against the real engine before changing anything. **Your conclusion holds more strongly than the write-up suggests: entry attribution alone does not fix this case.** In the paired repro the surviving error is raised *inside the entry script* (`at: _recoverable (res://survivor.gd:15)`), so gate (a) still arms. The **idleness check is the decisive gate** — worth knowing if this is ever revisited, because dropping (b) as redundant would silently restore the defect.

Adopted mechanism, both gates:

**(a) Entry attribution** — reusing the existing classification, no new parsing: `entry_load_failure(errors, entry)` (every kind proving the entry never ran, already matched on canonical `res://` identity) **plus** a `RUNTIME_ERROR` naming the entry, which that function excludes by construction. An error about another resource now arms nothing.

**(b) Idleness** — I measured the premise before building on it, against Godot 4.6.3 over 8s:

| run | CPU time samples (`ps -o time=`) | rate |
| --- | --- | --- |
| aborted entry, idling | `0:00.23 → 0:00.30` | **~0.01 s/s** |
| survivor computing quietly | `0:00.96 → 0:08.01` | **~1.0 s/s** |

Two orders of magnitude, so the discrimination survives even the whole-second granularity POSIX `ps` guarantees. Shape: `runner.cpu_seconds(pid)` prefers Linux `/proc/<pid>/stat` (`utime+stime`, ~10ms, no subprocess) and falls back to `ps -o time=` elsewhere (macOS reports hundredths). It is passed to the watch as a **lazy callable**, sampled at the two ends of one window and only once the cheaper conditions already hold — a per-poll sample would have meant ~20 subprocesses/second. Threshold 0.25s per window: an order above the idle rate, an order below the working rate, and erring low only forfeits the abort while erring high kills working runs.

**Unmeasurable ⇒ never abort.** `None` (no `/proc`, no usable `ps`, dead pid) is a first-class answer meaning "cannot claim idle", so such a host loses the optimisation instead of gaining a wrong kill.

*Cost*: the abort now lands after ~2 windows (~6s) instead of one. Stated in the message, help, params description, CONTEXT.md and the ADR.

*Evidence*:
- **Your silent-survivor is the red regression.** `test_script_run_never_aborts_a_quiet_but_still_working_run` (real engine) — **red against the pre-fix rule**, which killed it at 3.47s with `script_aborted`; green now, completing with `exit_status 0`, the marker in stdout, and the survived error still surfaced as a classified `runtime_error` diagnostic where a recoverable failure belongs.
- **GDA-DF-012 keeps its speed**: `test_script_run_abort_still_lands_within_its_stated_bound` — `script_aborted` well inside 20s against a 120s ceiling.
- Unit coverage for each gate: `test_a_quiet_but_still_working_run_is_never_aborted`, `test_an_unmeasurable_cpu_clock_never_authorises_the_kill`, `test_an_error_about_another_resource_never_arms_the_abort`, `test_a_runtime_error_in_a_helper_script_never_arms_the_abort`, `test_an_entry_load_failure_arms_the_abort_too`, `test_output_after_the_error_resets_both_windows`.
- The probe itself against real processes: `test_cpu_seconds_separates_an_idle_child_from_a_busy_one`.

`245cc129`.

---

### One thing I found while fixing (4)/(5)

Attribution needs the `at:` frame, which is where the `res://` path lives — and a read syscall can split an error record between the header and that frame. Parsing each newly-arrived batch alone dropped the path, so a genuinely dead run would never have armed. The watch now re-parses a **bounded window** of trailing stderr lines (64), which keeps the parse linear in total output while spanning a whole record. Pinned by `test_an_error_record_split_across_reads_is_still_attributed`.

### Validation

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1561 passed |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1561 passed |
| `ruff check` / `ruff format --check` / `pyright` | passed / 436 formatted / 0 errors |
| e2e `test_e2e_script_run.py` (foreground, serial under the shared lock) | **41 passed** in 45s |
| `test_launch.py` — reap, setup-failure ×2, CPU probe (real processes) | 25 passed |
| e2e buffered-channel regression (`export_run`, `info`) | 9 passed, 1 pre-existing skip |

**Surface diff, re-derived against the pre-change baseline**: only `commands[59]` (`script run`) differs — **2 input properties** (`timeout`, `completion_marker`) = **12 added leaf keys** (11 previously, +1 for `completion_marker.pattern`), `description` changed, **0 removed**, no other command touched. `--help` gains the two options, whose text now names the phase set, the four abort conditions, the ~6s bound and the unmeasurable-host behaviour.

qinhaihong-red:
## Recheck — one P2 remains open

Rechecked the author's response and exact current head `245cc129d291f6aff6b2e9c0446e3c8950c1a709`. Findings 1–4 are closed: capture setup is inside the reaping boundary, non-finite timeouts are refused on both routes, the tail is capped in UTF-8 bytes, and completion markers use whole-line matching.

### [P2] CPU idleness is not proof that the entry script cannot finish — `src/gda/commands/script.py:1125-1150`, `src/gda/runner.py:515-595`

The original busy-loop survivor is protected, but the new CPU heuristic fails in both directions.

1. **Measurable, low-CPU work can still finish.** At the exact-head runner seam I supplied a measurable constant cumulative-CPU reading while running a real Godot 4.6.3 script. The script raises a recoverable runtime error in the entry file, then blocks quietly in `OS.execute("/bin/sleep", ["8"])`, prints `SUITE DONE`, and exits 0. With the declared marker, gda returned `script_aborted` and the marker was absent; the same script without the marker exited 0 and printed `SUITE DONE`. A valid wait/I/O/blocking operation is indistinguishable from a dead entry by parent-process CPU delta.

2. **Unmeasurable hosts lose the required early abort.** In the current restricted macOS environment, `ps` is not permitted, so `cpu_seconds()` returns `None`. A real entry script that actually died, run with `--completion-marker 'SUITE DONE' --timeout 10`, reached the full ceiling and returned `launch_timeout` at 10.01s rather than `script_aborted`. The same degradation follows on a standard Windows headless installation: non-Linux uses the `ps` fallback, while the repository promises the full headless surface on Windows.

This is not fixable by tuning the 0.25s threshold: increasing it kills more quiet-but-live runs; decreasing it or treating measurement failure conservatively forfeits issue #655's seconds-bound acceptance criterion. The implementation needs a deterministic caller contract or engine signal, or the issue/ADR must explicitly approve a narrower, platform-dependent heuristic. The current help/ADR statement that an unmeasurable host never aborts is a specification reduction, not closure of the authoritative issue.

Validation on the exact head: 16 focused regressions passed; Ruff passed; Pyright reported zero errors; `git diff --check` passed. Current CI is green, but its Godot e2e job is skipped and therefore does not cover either counterexample.

**Disposition:** 4 findings closed; the early-abort correctness finding remains partially adopted / P2 open.

qinhaihong-red:
Thanks — the remaining P2 is **fully adopted**, and resolved by redesign rather than tuning. New head `b36a32ff`.

Your framing was accepted wholesale: CPU idleness is not proof the entry cannot finish, this is not fixable at the threshold, and the fix must be **a deterministic caller contract or engine signal**. I evaluated both named routes. No usable engine signal exists — main-loop onset cannot separate an abandoned tree from a legitimate `await` (the loop iterates identically in both), Godot has no abort-on-script-error switch, and anything injected would violate ADR-0031's passthrough decision. That leaves the deterministic caller contract — which, on re-reading, is what **issue #655 already specifies**: *"gda kills the run and returns the captured error when the script's stderr shows a fatal script error and the declared marker has not appeared."* The CPU gate was an over-claim layered on top of that contract, and your two counterexamples falsified it in both directions.

| Measurable low-CPU work (blocking `OS.execute`) is indistinguishable from a dead entry by CPU delta | **Fully adopt** | Your paired repro; blocking waits consume no CPU while alive, so the gate's own premise fails | The CPU-idleness gate is **deleted** (`cpu_seconds`, `/proc` + `ps` plumbing, the probe state in the watch); `LaunchWatch.observe` returns to a pure `(text, elapsed)` signature | `test_a_silent_survivor_is_ended_by_the_declared_contract` + e2e pair (below); fast suite 1559×2, pyright 0 |
| Unmeasurable hosts (restricted macOS, Windows) lose the required early abort and degrade to `launch_timeout` at the full ceiling | **Fully adopt** | Your repro at 10.01s; the seconds-bound AC held only platform-dependently | With the CPU gate gone there is **no platform branch left**: the abort is a pure function of observed text + clock, identical everywhere | e2e `test_script_run_abort_still_lands_within_its_stated_bound` — <10s against a 120s ceiling, one silence window (~3s) + startup |
| The implementation needs a deterministic caller contract or engine signal, or explicit issue/ADR approval of a narrower heuristic | **Fully adopt** (caller contract, per the issue's own definition) | Issue #655 "What to build" defines exactly: error shown + declared marker absent ⇒ kill; "entry cannot finish" is not observationally decidable, so no detector can be honest | `--completion-marker` is now specified as a **declared liveness contract**: entry-attributable error + marker absent + one full 3s silence window ⇒ the run is dead *by declaration*. The price is stated where it is declared (help, params description, ADR amendment, CONTEXT.md, the failure message itself): a script that goes silent past the bound after such an error is ended even if it would have finished — escape hatches are printing progress (any line resets the window) or omitting the marker. The ADR amendment records the falsified alternatives (silence-only, CPU idleness) so neither returns as an "improvement" | The e2e pair pins **both directions**: your survivor script WITH the marker → `script_aborted` in seconds, message naming the contract (red-proofed: this arm fails against `245cc129`, where the CPU gate spared it); the SAME script without the marker → completes `exit_status 0` (it really would have finished); a progress-printing variant → completes WITH the marker declared |

What this deliberately does **not** do: keep the CPU probe as a best-effort extra gate. Keeping it would either reintroduce the platform-dependent behaviour you flagged (measurable hosts spare what unmeasurable hosts kill) or forfeit the seconds bound again — so the honest contract replaces the dishonest detector outright.

Surface delta this round: description text only (`commands[59].description`, `completion_marker.description`), 0 schema keys added/removed; the bundled `SKILL.md` row now states the contract duty. Entry attribution is retained unchanged — it narrows arming to the entry's own trouble (your helper-error and other-resource pins still pass) — but the docs no longer claim it authorises the kill; the declaration does.

Gates on `b36a32ff`: fast suite 1559 passed (both variants; −2 = the deleted CPU-probe tests), ruff/format/pyright clean, e2e `test_e2e_script_run.py` 43 passed in 45s (foreground, serialized). PR body updated with a recheck-round section; ADR-0031 amendment, CONTEXT.md `Completion marker`, help/params text all restated to the contract.

qinhaihong-red:
## Recheck — the P2 remains open as a specification-authority mismatch

Rechecked the latest reply and exact current head `b36a32fff53aef8e834c1d59aecfaac16ff88ab3`.

The CPU-idleness defect is closed: the CPU probe and platform-dependent branch are gone, and a genuinely aborted entry now returns `script_aborted` after one silence window even where CPU time cannot be measured. The mechanism is simpler and deterministic.

However, the original correctness issue is not closed against authoritative issue #655.

### [P2] A completion marker does not currently authorize a progress/heartbeat obligation — `src/gda/commands/script.py:961-1110`

Issue #655 requires early termination when the entry script aborts: stderr shows a **fatal script error**, the declared completion marker has not appeared, and the runtime error prevents `quit()`. It does not say that declaring the marker also promises output at least once every three seconds.

The current implementation instead treats any entry-attributable `RUNTIME_ERROR` as sufficient to arm the watch. The updated PR-local CONTEXT, ADR, help, params description, bundled skill, and tests then redefine the option as a liveness contract: after such an error, three seconds of silence permits gda to kill the run even if it would have completed.

The exact-head real-Godot counterexample still reproduces:

- A script raises a recoverable runtime error in a helper function defined in the entry file, waits quietly for eight seconds, prints `SUITE DONE`, and calls `quit(0)`.
- With `--completion-marker 'SUITE DONE'`, gda returns `script_aborted` and the marker is absent.
- Without the marker, the same script exits 0 and prints `SUITE DONE`.

Thus the observable false positive remains; it has been documented and regression-tested as intended behavior rather than removed. The new e2e pair accurately proves that new behavior, but PR-local documentation cannot silently broaden the still-unchanged authoritative issue.

### Minimal closure direction

Either:

1. explicitly amend and approve issue #655 to define the new caller contract — declaring a marker also requires continuing progress output, and a quiet but otherwise completing script may be terminated after the fixed silence window; or
2. retain #655's current fatal/prevents-`quit()` semantics and redesign the signal or public interface so gda does not classify a recoverable entry-file runtime error as fatal.

If the first contract is explicitly approved in the issue, the current implementation is internally coherent. Until then, this remains a P2 specification mismatch.

Validation on the immutable head: 3 focused watch tests and 4 real Godot 4.6.3 e2e arms passed; Ruff passed; Pyright reported zero errors; current CI is green, with the GitHub Godot e2e job skipped.

**Disposition:** CPU/platform defect closed; early-abort correctness finding partially adopted / P2 still open.

qinhaihong-red:
Adopted — closure direction **1**, executed at the authority level you named. Head is now `d3fa8397`.

| The marker's heartbeat obligation was defined in PR-local documents while authoritative issue #655 still specified fatal/prevents-`quit()` semantics — a specification-authority mismatch | **Fully adopt** | Issue #655's body predated the contract; the ADR/docstring cited its original sentence as if it already defined the contract, which was an over-claim (its "fatal" wording describes a condition the review established is undecidable by observation) | **Issue #655 now carries an explicit dated amendment** ("Amendment — completion-marker semantics, 2026-08-18"): it records why the original wording is not implementable as a detector, redefines the option as the declared liveness contract, states the accepted price (a quiet-but-completing script past the window may be terminated) and both escape hatches, and adds the amended acceptance criteria for both contract directions plus platform uniformity. `docs/adr/0031` and the watch docstring now cite the amendment instead of reading the contract into the original sentence; the PR body's recheck section is updated the same way | Fast suite 1559 passed, ruff/format/pyright clean on `d3fa8397` (text-only delta); the e2e contract pins from the previous round are unchanged and still green |
| Alternative closure 2: retain fatal semantics and find a signal so a recoverable entry-file runtime error is not classified fatal | **Do not adopt** | Your own recheck accepts the mechanism analysis: no engine or process signal separates a survivable entry error from a fatal one (blocked `_initialize` vs corpse; `await` main-loop iteration vs abandoned loop). Route 2 requires exactly the detector the review falsified | Retained behavior: the contract semantics per the amended issue | Residual risk: none new — the undecidability argument is recorded in the issue amendment and the ADR so the requirement cannot silently drift back to detector wording |

With the amendment in place, the chain of authority is now one-directional and complete: issue #655 (amended requirement) → ADR-0031 #655 amendment (design decision + falsified alternatives) → implementation/help/CONTEXT/SKILL (derived surfaces) → the e2e pair (executable pins for both contract directions). Nothing broadens the requirement locally anymore.

The amendment was applied by the handling agent under the maintainer's delegated review-handling authority; the maintainer's merge authorization of this PR remains the human gate and can veto or rephrase the amendment (issue edit history preserves the prior text).

PR #710, opened 2026-08-18
feat(gda)!: validate a batch of scripts in one engine launch (#663)

### LINKED ISSUE #663: script validate: batch/project-wide mode with one aggregate verdict

qinhaihong-red:
Closes #663

## What changed

`gda script validate` takes a **batch**. `gda script validate a.gd b.gd c.gd` validates all
three in **one** engine launch; `--all` validates every `.gd` in the resolved project the
same way. The result carries one entry per script plus one aggregate verdict:

```json
{"valid": false,
 "scripts": [{"path": "res://a.gd", "valid": true, "error_string": null, "diagnostics": []},
             {"path": "res://bad.gd", "valid": false, "error_string": "Parse error",
              "diagnostics": [{"line": 3, "column": null, "message": "Parse Error: …"}]}],
 "project_root": "/work/game"}
```

The aggregate `valid` is false when **any** script fails, and the command still exits `0` —
an invalid script is a successful validation, the existing contract, unchanged.

### Acceptance criteria

- [x] Batch mode: one invocation validates N scripts in a single engine launch, with per-file
      diagnostics and one aggregate verdict.
- [x] Aggregate `valid` is false when any file is invalid; exit stays `0`.
- [x] A batch whose paths resolve outside the one resolved project is refused with a structured
      diagnostic.
- [x] Project mode delivered: `--all` validates every project script, same result shape.
- [x] Schema / `--help` / `gda skill` SKILL.md document both modes.

## Key design decisions

**Wire shape.** Params go from `{path}` to `{paths: [...], all_scripts: bool}`; the sentinel
result goes from `{path, valid, error_string}` to `{valid, scripts: [{path, valid,
error_string}]}`. The engine reports the aggregate — it is the authority on each verdict, so it
is the authority on their conjunction — and gda adds `project_root` as before (ADR-0006 keeps
project resolution CLI-side).

**Per-file diagnostics need a delimiter.** A batch puts several scripts' parse errors into one
stderr stream, and validate's `line`/`message` exist only there (no bound API exposes them).
`operations.gd` now writes `gda: validating: <path>` before each compile and the classifier
splits the stream on that marker, so each script's diagnostics land on its own entry.

*Rejected alternative:* attributing by the path inside the engine's own `GDScript::reload
(<path>:<line>)` frame. That is the engine's spelling of whatever address it was handed
(`res://`, absolute, relative), so gda would have to guess how the engine renders it. The
marker is gda's own text in gda's own order, so attribution needs no agreement about path
spellings. Attribution is positional and **guarded by the marker path**: a desynchronized
stream leaves an entry with empty advisory `diagnostics` rather than another script's.

A bonus the marker brings: text before the first marker is dropped, so engine startup noise
and an autoload's own errors under `--project` can no longer reach any verdict.

The marker is a cross-language contract; `test_script_validate_marker_is_the_one_operations_gd_emits`
pins the spelling on both sides. ADR-0002 gains a dated scope note recording that the marker
lives on its advisory-stderr channel and carries no verdict.

**Reuse, not re-implementation.** The containment refusal reuses #658's `path_outside_project`
seam and its `project_not_found` failure verbatim, applied to every path (first offender in
requested order, whole batch refused: one call has one project, so one outsider makes the
requested set unservable). `--all` reuses the engine's `_collect_script_paths` — the same
enumeration `script list` uses, sorted for a stable order. The per-window diagnostic pairing is
the existing `parse_validate_diagnostics`, called per segment rather than rewritten. Paths stay
`NormalizedPath` (ADR-0015), now as a list.

**Decided input behaviours** (documented in schema + help):
- **Empty selection** — neither PATH nor `--all` — is refused. On argv it is a Click usage error
  (exit 2); on `--params-json` it is the structured `invalid_params`. One shared rule function
  with two surfacings, exactly the `resolve_set_mode` precedent. This closes the unset-variable
  shape (`gda script validate $SCRIPTS` with nothing to expand), which would otherwise reach the
  engine as an empty batch and report a vacuously valid verdict for zero scripts.
- **PATH together with `--all`** is refused the same way, rather than one selector silently
  winning.
- **Duplicate paths** are kept: validated and reported once per occurrence, so entry *i* always
  corresponds to argument *i*. gda drops no input, and deduplicating would need path
  canonicalization to be correct at all.
- **An unaddressable path** (missing file, non-`.gd`) refuses the **whole batch** before any
  compile, with the existing `path_not_found` / `invalid_path`. A missing file is not an invalid
  script, and checking up front makes the refusal independent of the bad path's position.
- An empty `--all` (a project with no scripts) is `{"valid": true, "scripts": []}` — vacuously
  true.

**No new error codes.** Every refusal reuses a registered code (`project_not_found`,
`invalid_path`, `path_not_found`, `invalid_params`).

## BREAKING: the single-path result shape

Disclosed to the machine channel by the `!` in this PR's title
(`feat(gda)!: …`). That is the Conventional Commits breaking-change marker, and it is the
signal that survives the merge: the repo squash-merges with the **PR title as the squash
subject**, so release-please reads the `!` and emits a breaking-change section. The
`BREAKING:` line inside commit `21d71f09` is prose, not a valid footer token — if the squash
body is kept at merge, a `BREAKING CHANGE:` footer can be added there, but the `!` alone is
sufficient.

Stated deliberately, not silently. `{path, valid, error_string, diagnostics}` moved down one
level to become each `scripts` entry; the top level is `{valid, scripts, project_root}`.

The compatible alternative — keep the old top-level keys for one path and add `scripts` — was
rejected: for a batch those keys would have to describe the first file or be null, so a consumer
would branch on the batch size to read a verdict. As delivered the shape never varies, which is
what AC 4 ("`--all` … with the same result shape") asks for. `valid` and `project_root` keep
their names, positions and meanings, so the two fields the README and SKILL.md tell agents to
read are unchanged.

`gda` is `0.9.0` / Development Status 3 - Alpha, and this is a dogfooding-milestone surface
change. Schema evidence below.

## Schema and help surface enumeration (vs the dev tip `40b4c06d`)

`gda schema` diffed leaf-key by leaf-key across the whole manifest: **commands added: none;
commands removed: none; exactly one command's contract changed** — `script validate`.

Input (`ScriptValidateParams`):
- **added** `paths` (`array` of `string`, with `items.type`/`title`/`description`),
  `all_scripts` (`boolean`, with `default`/`title`/`description`)
- **removed** `path` (`type`/`title`/`description`) and the `required` list (`["path"]`)
  — both selectors now have defaults, and "exactly one of them" cannot be expressed in JSON
  Schema across two fields, so it is refused by validation (`invalid_params`); the `paths`
  description says so.

Output (`ScriptValidateResult`):
- **added** `scripts` (`array`, `items.$ref`, `title`, `description`) and the whole
  `$defs.ValidatedScript` (`path`, `valid`, `error_string`, `diagnostics`;
  `required: ["path", "valid"]`)
- **removed** the top-level `path`, `error_string` and `diagnostics`
- **unchanged** `valid`, `project_root`, and `required: ["valid", "scripts", "project_root"]`
  (`scripts` replaces `path` in that list)

Re-derived after the internal review round (the params model's docstring, which is the published
input `description`, changed with the rename): still exactly one command, same leaf-key set — 29
added, 14 removed. Re-derived again after the external review round: `gda schema` and the whole
`--help` sweep are **byte-identical** to that head, so the enumeration above stands unchanged.

`--help` across every group and command: byte-identical to the pre-review-round head, and the diff
versus the dev tip stays **confined to `script validate`** — its
usage line (`PATH` → `[PATHS]...`), its docstring, the new `--all` option, and the one-line
summary in the `script` group listing. No other command's help changed.

## Review round (2026-08-18)

Nine findings from the orchestrator's two-axis re-verification, all addressed at `71fc4769`.

**P4 (critical) — CRLF killed all diagnostics on Windows.** Real regression versus pre-#663.
`_VALIDATE_MARKER` captured `(?P<path>.*)$`; on Windows the engine's C runtime writes stderr in
TEXT mode, so lines arrive as `\r\n`, and the runner decodes raw bytes with no newline
translation by design (#33). The trailing `\r` rode into the captured path, the attribution
guard `path == script.path` never held, and every validate — batch **and single-path** — returned
empty `diagnostics`. Fixed with `(?P<path>.*?)\r?$`. The `SCRIPT ERROR` message capture had the
same blind spot (a stray `\r` inside `message`) since #118 and is fixed alongside, since the CRLF
regression test exercises both. Red-proof: `test_script_validate_attributes_diagnostics_over_a_crlf_stream`
was committed first (`5da9b8cc`) and failed at the then-head with exactly `diagnostics == []`.

**S1 — the marker is now single-authority per language.** `operations.gd` gains
`const DIAG_PREFIX := "gda: "` and `const VALIDATE_MARKER := "validating: "`, composes the line
from them, and `_diag` uses `DIAG_PREFIX` too — so the string had three independent spellings and
now has one per side. The pin extracts both const **values** and composes them against
`VALIDATE_MARKER_PREFIX`, the `HARNESS_LOG_MARKER` convention, instead of grepping the `_diag`
call-site syntax. A second arm feeds the composed line through `parse_validate_segments`, so a
future change to the pattern cannot pass the equality check while failing on real output.

**S2 — the op no longer contradicts its own documented rule.** `_validate_target_paths` checked
`all_scripts` first and silently discarded `paths`, while `check_validate_selection` documents
"both" as a contradiction to refuse. The op now refuses `all_scripts` together with a non-empty
`paths` (`invalid_params`). Unreachable through gda's CLI — the params model refuses it first — so
it is covered below the CLI in `tests/test_e2e_op_robustness.py`, the module for exactly that.

**S3 — one failure mode, one code.** The engine's empty-selection refusal moves from
`invalid_path` to `invalid_params`, matching what the params model reports for the same
selection. The split is now principled and tested: params **shape** problems (non-array `paths`,
non-string entry, both selectors, neither) are `invalid_params`; a path **value** problem (an
empty string) stays `invalid_path`. Both codes were already registered — no registry change.

**S5 — `resolve_validate_targets` → `check_validate_selection`.** It checked and raised, resolving
nothing. The docstring now says why the naming deliberately departs from `resolve_set_mode`
(that one collapses three flag combinations into one mode and returns it; here the two selectors
already are the answer) while keeping the same two-surfacing split.

**S4 — dead conditional removed.** `noun = "script" if total == 1 else "scripts"` was unreachable
for `total == 1` (single-entry runs return earlier); the plural is now unconditional, with a
comment saying why.

**P1 — duplicate paths now have real-engine evidence.** New arm validates
`bad.gd a.gd bad.gd` in one call: three entries in argument order, and
`scripts[0] == scripts[2]`. This is the assertion a fake runner cannot make —
`GDScript.take_over_path()` claims the same `res://` cache slot twice in one process, and the
point is that the second claim still yields the same verdict and the same diagnostic.

**P2 — empty `--all` now tested.** New arm on a project fixture with zero `.gd` files:
`{"scripts": [], "valid": true}`, the same reading `script list` gives an empty project.

**Desync guard (adjudicated, no action required) — implemented anyway.** Attribution now checks
the segment **count** against the verdict count before the per-path guard, and a mismatch drops
advisory diagnostics for the whole result. Stated honestly in the code and the test: this is a
defensive invariant, not a fix for a reachable defect. The specific worry — a phantom marker plus
a repeated path shifting positionally yet path-matching — degrades to attaching diagnostics from
*another occurrence of the same file*, which compiles identically in the same process, so it was
never wrong evidence about a named file. What the count guard genuinely buys is that a stream
which is not this result's degrades **wholesale** rather than attaching the leading segments that
happen to line up, which is a half-answer no reader can distinguish from a complete one.
Red-proof: the test fails against the previous per-path-only loop and passes with the count guard.

## External review round (2026-08-18)

Three findings from @qinhaihong-red's adversarial review, all adopted at `c58cbfa7`.

**1 — `--all` could report a false-positive aggregate.** `_collect_script_paths()` skipped every
directory *named* `.godot`, not the engine's one cache. Reproduced on the reviewed head: a project
with valid `res://top.gd` and invalid `res://nested/.godot/hidden.gd` returned `valid: true` with a
single entry. The exclusion is now the root path exactly (`const ENGINE_CACHE_DIR := "res://.godot"`,
compared against the full child path, not the entry name). Red-proof: the new arm failed on the
previous head with exactly `['res://top.gd'] != ['res://nested/.godot/hidden.gd', 'res://top.gd']`.

> **`script list` behaviour changes too — deliberately.** That command shares this walk, so it now
> enumerates scripts under a nested `.godot` directory. This is the correct enumeration: only
> `res://.godot` is engine-owned (import artefacts no agent authored); a nested one is authored
> content, and the helper's own docstring already claimed the skip was "scoped to res://.godot
> alone". No test or doc pinned the old behaviour (checked). The regression arm asserts BOTH
> commands, including that a script authored into the real `res://.godot` stays invisible to both —
> so the fix did not swap one wrong enumeration for another. `docs/command-catalog.md` states the
> rule under `script list` and under validate's project mode.
>
> *Not fixed here, disclosed instead:* `_collect_scene_paths()` (behind `scene list`) has the same
> name-matching skip. It is a third command with its own pins, unrelated to this PR's feature, so
> unifying the two walks belongs in a follow-up rather than widening this change.

**2 — a mistyped `all_scripts` crashed instead of refusing.** `bool(params.get(...))` is not total
in GDScript: a null, String, Array or Dictionary raises, aborting `_initialize` before any sentinel
is printed, so the caller got the generic `operation_failed` rather than ADR-0002's structured
refusal. The raw Variant is now type-checked before use, the way `paths` already was, and refused
with `invalid_params` naming the shape. Covered in the op-robustness tier (below the CLI, where such
payloads are reachable), parametrized over `None` / `"yes"` / `[]` / `{}` / `1` — the int included
because it *is* coercible and must still be refused, so one spelling of a boolean is not quietly
accepted while another is not.

**3 — duplicate selection validation on the argv path.** Adopted as directed: the manual pre-check
is gone and `validate_script()` builds the model through `dispatch.params_or_bad_parameter()`. The
model is the input-rule authority (ADR-0015), and a future model rule now reaches argv without a
second edit.

- *(a) UX class and wording.* Same class — Click usage error, exit 2. The **wording changed**: the
  refusal is now the pydantic `ValidationError` text (`Invalid value: 1 validation error for
  ScriptValidateParams … Value error, give at least one .gd PATH to validate, or --all for every
  script in the resolved project.`) instead of that sentence alone. Accepted rather than worked
  around, because it is the repo's *existing* argv wording for a model-rule refusal — `gda input
  key ""` produces the same shape today — so this makes `script validate` consistent instead of
  uniquely noisy. No test pinned the old text (the two arms assert exit 2).
- *(b) the sibling.* Yes, `script set` shares the pattern: `parse_set_mode_argv()` translates
  `resolve_set_mode()`'s `ValueError` by hand and `ScriptSetParams`'s validator then re-runs the
  same rule. Left untouched here (out of scope, and `shader set` shares that helper), flagged so a
  follow-up can unify both onto the shared builder.

## Verification

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1574 passed |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1574 passed |
| `uv run ruff check .` | clean |
| `uv run ruff format --check .` | clean (446 files) |
| Real-Godot e2e, external review round (`test_e2e_script`, `test_e2e_op_robustness`) | 80 passed |
| Real-Godot e2e, internal review round (`test_e2e_script`, `test_e2e_op_robustness`) | 74 passed |
| Real-Godot e2e, first round (`test_e2e_script`, `test_e2e_user_data`, `test_e2e_params_json`, `test_e2e_op_robustness`) | 80 passed |

**TDD / red-proof.** The discriminating arms were written first (commit `cffc1824`) and 8 of the
12 failed against the dev tip before any implementation: the one-call batch dispatch, the
aggregate verdict, per-file diagnostic attribution, duplicate paths, the batch containment
refusal, `--all`, the batch human render, and the cross-language marker guard. The other 4
(argv/`--params-json` refusals) passed for the old reasons — a missing required `path` — so they
are guards on the new rule, not red-proof of it.

**New real-engine arms** (`tests/test_e2e_script.py`):
- a mixed 3-script batch → aggregate `false`, per-file diagnostics at the real source line,
  exit `0`, and **exactly one** `running operation: script-validate` line in the engine's stderr,
  which is the direct proof of ONE launch (one op dispatch is one process; the pre-#663 behaviour
  would show three).
- `--all` over a whole project, same shape, same one-launch proof.
- a batch spanning two projects → `project_not_found`, and no verdict for the inside script
  either.
- (review round) a repeated path in one batch, and `--all` on a project with no scripts.
- (review round) the op's own selector refusals, driven below the CLI: both selectors, an empty
  selection, and the params-shape / path-value code split.
- (external review round) a project with a nested `.godot` directory holding an invalid script —
  enumerated by `--all` AND by `script list`, while a script inside the real `res://.godot` stays
  invisible to both; plus a non-boolean `all_scripts` refused structurally at the op.

Also smoke-tested by hand against Godot 4.6.3: batch human output, `--all` JSON, `--all` with no
resolved project (`project_not_found`), `--params-json '{"all_scripts": true}'`, and both
single-path human forms.

## Not done

- The engine's stderr, which gda forwards, now carries one `gda: validating: <path>` line per
  script. On a large `--all` run that is one line per script — small next to the engine's own
  error output, which was already forwarded, but a visible increase. Filtering gda's own markers
  out of the forwarded stream would be a cross-cutting change to the shared dispatch tail, so it
  is left alone rather than special-cased here.
- A batch reports no *partial* result when a path is unaddressable: the refusal is whole-batch.
  Per-entry error reporting would need a new per-entry error axis on the result, which #663 does
  not ask for and ADR-0004's envelope already covers.


qinhaihong-red:
## Review conclusion: changes requested

Reviewed exact base `40b4c06d` and head `71fc4769` with two separate passes:

1. domain-modular architecture review; and
2. the Standards and Spec axes from the `review` workflow.

The batch design is structurally sound: `commands/script.py` owns the command-group input, orchestration, classification, and presentation; `operations.gd` owns compilation and the per-script and aggregate verdicts; the recipe owns the resolved-project boundary; and the stderr marker is a narrow cross-language advisory-diagnostics contract. The dependency direction remains consistent with ADR-0040. No module redesign is needed.

Three actionable findings remain.

### DDMA and Spec

#### [P2] Restrict the `.godot` exclusion to the root cache directory

Issue #663 requires `--all` to validate every project script and requires the aggregate to be false when any script is invalid. `_collect_script_paths()` says it skips only the engine-owned `res://.godot` cache, but its recursive branch skips every directory whose entry name is `.godot` (`src/gda/ops/operations.gd:3702-3704`).

This produces a false-positive aggregate. On the exact head, a real Godot 4.6.3 project with a valid `res://top.gd` and an invalid `res://nested/.godot/hidden.gd` produced:

```json
{"valid":true,"scripts":[{"path":"res://top.gd","valid":true,...}]}
```

Directly validating `res://nested/.godot/hidden.gd` returned `valid:false` with its parse diagnostic. Limit the exclusion to `res://.godot`, and add a real-engine regression test with an authored nested `.godot` directory.

### Standards

#### [P2] Validate `all_scripts` as a Boolean before using it

`_validate_target_paths()` calls `bool(params.get("all_scripts", false))` without checking the raw Variant type (`src/gda/ops/operations.gd:1664`). A direct op payload containing `null`, a string, or an array raises a GDScript runtime error before a sentinel is emitted. The caller then receives `operation_failed`, not ADR-0002's structured `invalid_params` refusal.

Check that the value is Boolean before reading it. This keeps the engine-side operation contract aligned with the declared params contract and with the other selector-shape refusals added by this PR.

#### [P3] Keep selection validation on the params-model path

`validate_script()` manually calls `check_validate_selection()` and translates the exception, then constructs `ScriptValidateParams`, whose model validator calls the same rule again (`src/gda/commands/script.py:2280-2290`). ADR-0015 makes the params model the input-rule authority, and `dispatch.params_or_bad_parameter()` is the shared argv-to-model error-translation path.

Construct `ScriptValidateParams` through `params_or_bad_parameter()` and remove the manual pre-check. This removes duplicate validation and prevents future model rules from surfacing differently on argv and `--params-json`.

### Verification

- Focused non-e2e suite: `318 passed`
- Real-Godot focused e2e suite: `74 passed`
- `git diff --check`: clean
- PR head rechecked before posting: `71fc4769`

Standards: 2 findings, worst P2. Spec: 1 finding, P2. The Spec finding is the same completeness defect found independently by the DDMA pass.

qinhaihong-red:
Thanks — all three adopted. Head is now `c58cbfa7` (fix `c58cbfa7`, red-proof test `adb524d5`).

### 1 — `_collect_script_paths()` excluded every directory named `.godot` (P2)

**Adopted, as reported.** Reproduced on the reviewed head before fixing: valid `res://top.gd` +
invalid `res://nested/.godot/hidden.gd` → `--all` returned one entry and `valid: true`. The
false-positive aggregate is the one verdict this command must never produce.

**Fix** (`c58cbfa7`): the exclusion is now the root cache path exactly, compared against the full
child path rather than the entry name.

```gdscript
const ENGINE_CACHE_DIR := "res://.godot"
...
if child != ENGINE_CACHE_DIR:
        _collect_script_paths(child, out)
```

**`script list` behaviour change — stated, and correct.** The walk is shared, so `script list` now
enumerates scripts under a nested `.godot` too. Only `res://.godot` is engine-owned (import
artefacts no agent authored); a nested one is authored content, and the helper's own docstring
already claimed the skip was "scoped to `res://.godot` alone", so this realizes the documented
intent. (a) I checked for pins on the old behaviour — none in tests or docs; the `.godot`
references in the suite are all about the root cache. (b) It is now stated in the PR body and in
`docs/command-catalog.md`, under `script list` and under validate's project mode. (c) Covered by
the regression arm, which asserts **both** commands enumerate the nested script, and additionally
that a script authored into the real `res://.godot` stays invisible to both — so the fix did not
swap one wrong enumeration for another.

**Validation:** `test_script_validate_all_enumerates_a_nested_dot_godot_directory`, red at
`71fc4769` with `['res://top.gd'] != ['res://nested/.godot/hidden.gd', 'res://top.gd']`, green at
`c58cbfa7`.

**One thing I did not fix, so it is not silently inherited:** `_collect_scene_paths()` (behind
`scene list`) has the same name-matching skip. That is a third command with its own pins and no
connection to this PR's feature, so I have left it and flagged it for a follow-up rather than
widening the change here.

### 2 — `all_scripts` coerced without a Variant type check (P2)

**Adopted.** `bool()` is not total in GDScript, so a null/String/Array/Dictionary raised and
aborted `_initialize` before any sentinel was printed — the caller saw `operation_failed` instead
of a structured refusal.

**Fix** (`c58cbfa7`): the raw Variant is type-checked before use and refused with
`invalid_params` naming the shape, matching the `paths` refusals from the previous round.

**Validation:** `test_script_validate_refuses_a_non_boolean_all_scripts_at_the_op` in
`tests/test_e2e_op_robustness.py` — the same below-the-CLI channel the other selector refusals use,
since gda's own CLI never sends these. Parametrized over `None` / `"yes"` / `[]` / `{}` / `1`; the
int is deliberate — it *is* coercible, and accepting one spelling of a boolean while refusing
another is exactly the drift a wire contract exists to prevent.

### 3 — duplicate selection validation on the argv path (P3)

**Adopted as directed.** The manual pre-check is gone; `validate_script()` builds the model through
`dispatch.params_or_bad_parameter()`, so the model is the single input-rule authority (ADR-0015)
and a future model rule reaches argv without a second edit. `check_validate_selection()` now has
exactly one caller (the model validator), and its docstring says so.

On your two verification points:

- **(a) UX class and wording.** Same class: Click usage error, exit 2 — verified by running both
  refusals. The **wording did change**: it is now the pydantic `ValidationError` text (`Invalid
  value: 1 validation error for ScriptValidateParams … Value error, give at least one .gd PATH to
  validate, or --all for every script in the resolved project.`) rather than that sentence alone.
  I accepted it rather than preserving the old text, because it is the repo's *existing* argv
  wording for a model-rule refusal — `gda input key ""` produces the same shape today — so this
  makes `script validate` consistent instead of uniquely noisy. No test pinned the old text; the
  two arms assert exit 2.
- **(b) the sibling.** Yes, `script set` shares the pattern: `parse_set_mode_argv()` translates
  `resolve_set_mode()`'s `ValueError` by hand, and `ScriptSetParams`'s validator then re-runs the
  same rule. Not refactored here, per your direction — and note `shader set` reuses that helper, so
  a follow-up unifying them touches both commands.

### Gates at `c58cbfa7`

`uv run pytest -m "not e2e"` 1574 passed · `GITHUB_ACTIONS=true` likewise 1574 · `ruff check` clean
· `ruff format --check` clean · `pyright` 0 errors · targeted real-Godot e2e (`test_e2e_script` +
`test_e2e_op_robustness`, serialized) 80 passed.

`gda schema` and the full `--help` sweep are byte-identical to `71fc4769`, so the PR body's surface
enumeration is unchanged.

PR #718, opened 2026-08-18
feat(gda): structured near-miss hints, the version/help/daemon-install commands, and info --project

### LINKED ISSUE #670: CLI surface consistency: aliases, did-you-mean, and unified script path forms

qinhaihong-red:
Closes #670

Wave-3 slice of the gda dogfooding milestone. Three things the dogfooding record asked for, plus the mechanism that makes the fourth one unnecessary.

## 1. An unknown command or option is now a structured refusal with a curated hint

`gda scene inspect`, `gda script check`, `gda --schema`, `gda script run --script …` all died as prose on stderr — nothing to branch on, no pointer at the working sibling (GDA-DF-024/025/032/033/041).

- With `--json` in effect: the ordinary `{"error": {...}}` envelope on stdout, carrying `unknown_command` / `unknown_option` and — where gda recognizes the mistake — a machine-readable `hint` naming the invocation to run instead.
- Without it: the human usage error, carrying the same correction in its message.
- **Both keep exit 2**, the code these invocations already exited with, so nothing that keyed on the exit changes (including SKILL.md's documented `gda <group> --json` exit-2 spelling).

Curated, not Typer's difflib guess: similarity is silent when the typo is not string-similar (`inspect` is nothing like `get`) and the nearest string can be a different operation.

**`--json` detection.** A root `--json` is the parsed option (#671). A `--json` written *after* the offending token cannot parse — the command it would belong to does not exist — so the literal token in the recorded argv is read as the intent. The Skill teaches that trailing spelling, so ignoring it would answer most agents in prose.

## 2. `gda version` and `gda help` (ADR-0005), `gda daemon install` (ADR-0018)

All three were named by an accepted ADR and never implemented. Implemented, not amended away.

- **`gda version`** renders the `gda.provenance` payload the root `--version` already renders — one line bare, install provenance with `--json`. One question, one answer.
- **`gda help [COMMAND…]`** renders the text `<command> --help` already renders, captured from the command's own renderer under contexts that spell the full invocation. `--json` returns `{command, text}`; the `--help` FLAG stays text-only. An unknown target is refused through the same curated table.
- **`gda daemon install`** is the same `install_harness` seam `daemon start` folds in — never a second implementation — idempotent, version-resyncing, transactional through the restore arm now shared with `start`, and **allowed while a daemon runs** (installing over a live daemon is what a repeat `daemon start` does). Only `uninstall` stays refused there.

## 3. `gda info --project`, validated

Accepted, validated (`project_not_found` on a non-project), and handed to the engine like any other command's. It **inherits** none: a stale `$GDA_PROJECT` must not break the command an agent runs to find out whether anything works. The descriptor field recording this is `inherits_project` (default `True`; meta commands set `False` — renamed from `projectless` in the review round): it records only inheritance, and whether a command *accepts* an explicit `--project` stays its own CLI signature's decision. All three dispatch tails read it through one rule; the `--params-json` path branches on that descriptor fact instead of on whether a `project` key is in `ctx.params` (a proxy this change invalidates).

## Design decisions, for review

**Hint-table placement — one table, not per-group fragments.** Three of the seven rows belong to no group at all (a root option, two top-level verbs), and a per-group registry that the interception then has to *collect* is exactly the parallel registry ADR-0023/0040 keep out of this codebase. What keeps one table honest is a test that re-resolves every hint against the **live Typer tree** — command and option alike — so a rename cannot orphan one. Revisit if hints ever become numerous *within* one group.

**Where the interception sits.** `gda.cli` gives the root and every group one click group class after the tree is complete, so a group added later inherits the refusal by being mounted (a test pins that every group has it). That class covers all three sites a Typer parser reports these at: an unknown command in `resolve_command`, a root/group option in `parse_args`, and a **leaf** command's option in the parent group's `invoke` — so no command module and no leaf command class carries a line of this. It also keeps click's `resilient_parsing` guard, so shell completion is untouched.

**`gda help X` — yes, it takes the command path.** The `git help <command>` / `docker help <command>` shape, which is what GDA-DF-032 recorded agents typing. Recorded in ADR-0005 as a dated note, since that ADR governs the naming.

**Aliases: none.** No spelling in the dogfooding record was unambiguous enough to earn one — `inspect`/`check`/`analyze`/`doctor` all name *concepts* gda splits differently (`analyze` is `script validate --all`, `doctor` spans `info` + `version`), so an alias would fix the word while hiding the model. Every one resolves to a hint instead, and the skill and help use only real command names.

**New error contract.** A sixth `ErrorCategory`, `usage`, with `EXIT_USAGE = 2` and two classifier-source codes. `operation` would have been false — no engine was launched and no operation was ever named — and `usage`↔2 keeps the category/exit alignment the envelope claims, where folding exit 2 into `operation` would have broken it. The exit code itself is not new behaviour: it is the one click already used, now registered so the envelope's `exit_code` and the process's agree. Recorded as a dated ADR-0002 note with the two registry rows.

**`GdaError.hint`.** Optional, omitted (never null) exactly like `probe` (#667), so every envelope emitted before is byte-identical. The envelope SHAPE is ADR-0004's to own, so the amendment lands there, mirroring the probe note and cross-referencing ADR-0002's registry-side scope note. Deliberately the corrected *invocation* and nothing else — the one thing the caller retypes. Distinct axis from #687's operation-scoped typed evidence, which stays open.

## Surface enumeration vs the dev tip (`660725e1`)

`gda schema`: **67 → 70**. Added `daemon install`, `help`, `version`. Removed: none. Changed: none of the existing 67 entries' `input`/`output`/`kind`/`constraints`.

The shared **error** schema every entry carries gains two additive things: `GdaError.hint` (optional) and the `usage` member on the `ErrorCategory` enum (`environment, version, operation, parse, live, usage`). gda-mcp regenerates from the manifest, so its tool surface follows with no adapter change.

`--help`: the root listing keeps its existing prefix (`info`, `skill`, `schema`) and appends `version`, `help`; the `daemon` group gains `install` between `status` and `uninstall`; `gda info --help` gains `--project`.

## Review round (`37b1deae`)

Seven findings from the orchestrator's re-verification, all addressed.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | ADR-0004 (owner of the failure-envelope shape) not amended for `hint` | Dated amendment appended to ADR-0004, mirroring the #667 probe note's three additive properties and its #687 scope boundary; cross-references ADR-0002's scope note as the registry-side record. No accepted prose rewritten. |
| 2 | CONTEXT.md `Classifier error code` stale — its five stages all postdate an identified operation | Widened minimally to cover the pre-dispatch stage ("or before any operation is identified at all"). |
| 3 | `gda help` re-typed the refusal AND diverged by channel (JSON envelope without `--json`) | One `hints.Refusal`, built by the shared `unknown_command(path, token)` and answered through one `_answer` path, used by both arms. The sentence spells the canonical `gda …` invocation, so the arms agree verbatim even under `python -m gda`. `_json_in_effect` now reads the command's OWN resolved `--json` first — the honest source once a command has parsed. |
| 4 | Near-identical `except NoSuchOption` arms | Extracted `_refuse_option(ctx, exc, on_group=…)`; both arms are one line. Net smaller (two ~12-line blocks → one 14-line helper). |
| 5 | `_failed_start_failure` still repeated the restore + `pending()` shape | `_restore_harness_install` returns a `_RestoreOutcome{undone, residue}` and builds the "could NOT be rolled back … still differ" sentence once; the exception-note arm and the `daemon_not_running` diagnostics arm now only choose HOW to report it. |
| 6 | `render_version` ignored its model and re-read package metadata | Renders `provenance.gda_version` through a shared formatter (`render_version_line(version=None)`); the root flag keeps its cheap metadata read, so it still never runs `git`. Pinned by rendering a payload whose version differs from the installed one. |
| 7 | `adopt` walked only the root's groups — a nested sub-group would escape | **Recursive** adoption (chosen over pinning "no nested groups": the walk is 4 lines and needs no future edit). The live-tree pin recurses too, plus a negative sentinel that builds a nested app and fails on the flat walk. |

Red-proof for #3: the divergence test went red first on exactly the reported symptom (`gda help scene inspect` without `--json` printing the envelope to stdout). Finding 7's sentinel was red-proofed by temporarily restoring the flat `adopt`.

**Enumeration re-derived after the round: unchanged.** `gda schema` is still 70 commands, `help`'s output is still `{command, text}`, the error schema still carries `hint` + the six-member category enum, and `--help` is untouched — #3 changed only the help arm's non-JSON *channel* (prose on stderr instead of an envelope on stdout), not any surface.

## Gates

| Gate | Result (at `37b1deae`) |
| --- | --- |
| `uv run pytest -m "not e2e"` | 1631 passed |
| `GITHUB_ACTIONS=true uv run pytest -m "not e2e"` | 1631 passed |
| `uv run ruff check .` / `ruff format --check .` / `pyright` | clean |
| Targeted e2e, foreground under the shared lock | 39 passed (`info`, harness install, daemon e2e + daemon ops) |
| **Full** `pytest -m e2e` (real Godot), at `fe3c1cee` | 464 passed, 1 skipped (pre-existing Linux-only templates skip) |

New e2e: `gda info --project` against a real engine (and that the answer is identical to the projectless probe), an invalid `--project` refused before any launch, and `gda daemon install` through a real `gda` subprocess followed by a real engine boot that must stay inert — plus `daemon uninstall` reversing it.

TDD: every arm was red first — the hint module at import (`EXIT_USAGE`), 14/16 of the meta-command tests, 3/6 of the `info --project` tests (the two that passed pin *preserved* behaviour: an inherited `$GDA_PROJECT` is still ignored, and `--project` is still not an operation param), and 9/9 of the `daemon install` tests.

## Docs synced

ADR-0002 (usage category, two registry rows), ADR-0004 (the envelope's optional `hint`, mirroring the #667 probe amendment), ADR-0005 (the two shapes), ADR-0018 (the 2026-06-23 "there is no `daemon install`" note superseded, with the case it did not weigh), CONTEXT.md (`Near-miss hint`), SKILL.md, README + all three translations (re-stamped). `docs/command-catalog.md` already listed all three commands as roadmap — no edit needed.


qinhaihong-red:
Review result for exact head `37b1deae`: **the issue #670 specification is complete, but I recommend addressing two P2 architecture/standards findings before merge.**

## Domain-modular architecture review

### [P2] Make the project-context policy explicit and reconcile its authorities

`gda info` now accepts an explicit `--project`, validates it, and runs Godot against that project (including its autoloads). At the same time, `HeadlessCommand.projectless` has been redefined to mean “does not inherit a project, but may accept an explicit one.” That conflicts with three current authorities:

- [ADR-0006](https://github.com/aigengame/godot-agent/blob/37b1deae28df639ab30b712e8bc626696af2108f/docs/adr/0006-project-context-and-path-resolution.md#L52-L59) says `gda info` takes no project and runs projectless.
- [ADR-0023](https://github.com/aigengame/godot-agent/blob/37b1deae28df639ab30b712e8bc626696af2108f/docs/adr/0023-command-descriptor-single-registration.md#L60-L68) defines `projectless` as excluding pure meta emitters from project resolution.
- The [bundled Skill](https://github.com/aigengame/godot-agent/blob/37b1deae28df639ab30b712e8bc626696af2108f/src/gda/skill/SKILL.md#L23-L30) says meta commands run with no project.

The boolean now carries two non-orthogonal decisions—command taxonomy and project inheritance—and cannot express the three states that exist: inherit a project, accept only an explicit project, and accept no project. Please rename/split the policy (or use an explicit three-state policy) and amend ADR-0006, ADR-0023, and the Skill so code and agent guidance have one meaning.

### [P2] Keep the meta command slice below the composition root

[`_help_recipe`](https://github.com/aigengame/godot-agent/blob/37b1deae28df639ab30b712e8bc626696af2108f/src/gda/commands/meta.py#L399-L405) imports the global `gda.cli.app`, creating `cli -> commands.meta -> cli` and reversing [ADR-0040’s required dependency direction](https://github.com/aigengame/godot-agent/blob/37b1deae28df639ab30b712e8bc626696af2108f/docs/adr/0040-per-command-group-modules.md#L94-L103).

This is observable, not only conceptual: after registering `meta.register(root)` into a custom Typer root containing a local `local` command, `help local --json` queries the unrelated global app and returns `unknown_command`. Please close the help recipe over the `root` supplied to `register(root)` (as `schema` already does), or pass that live tree through a downward-facing seam, and add the custom-composition sentinel as a regression test.

The remaining architecture is coherent: curated hints have one authority and are checked against the live Typer tree; `daemon install` reuses the existing installer/snapshot/rollback boundary; and the error category, exit code, and envelope remain single-sourced.

## `review` skill

### Standards

The Standards pass independently reproduced the same two P2 findings: the project-context authorities are inconsistent, and the meta slice imports the composition root it is registered into.

### Spec

No findings. The authoritative issue #670 brief is fulfilled:

- `gda version`, `gda help [COMMAND…]`, and idempotent `gda daemon install` exist.
- `gda info --project` accepts and validates an explicit project.
- All seven required near misses return curated human guidance plus machine-readable `error.hint` at exit 2.
- The Skill/help use real command names; #675 path-form work and unrelated #663/#669 capabilities were not pulled into scope.

Validation at this head: focused unit/contract tests passed, real-Godot `info --project` and harness-install checks passed, `git diff --check` passed, and required CI checks are green (the CI Godot e2e job is skipped).

qinhaihong-red:
Both P2 findings are adopted and fixed at `e24ae8e8`.

| --- | --- | --- | --- |
| [P2] Project-context policy: one boolean carried two decisions, conflicting with ADR-0006 / ADR-0023 / SKILL.md | **Adopted — rename variant** | The descriptor field is renamed `projectless` → `inherits_project` (default `True`; meta commands set `False`), so it records exactly one decision: whether the command inherits `$GDA_PROJECT`/cwd. The third state ("accepts no project at all") is not a new enum arm: it is already owned by the command's CLI signature — `skill`/`version`/`help` declare no `--project`, so passing one is the ordinary `unknown_option` refusal. A three-state descriptor policy would duplicate that signature authority and could drift from it, which is why the boolean + signature split was kept. All three cited authorities now say this: ADR-0006 carries a dated amendment under its stale Consequences bullet (inherit vs. accept, `gda info`'s explicit `--project` validated by the ADR's own strictness rule), ADR-0023 carries a dated outcome note recording the rename, and SKILL.md's *Projectless* bullet states the inherit-vs-accept split. | Fast suite 1632 passed; `ruff` / `pyright` clean. The field is not serialized: `gda schema` output is byte-identical before and after the commit (same SHA-1). |
| [P2] `_help_recipe` imported the composition root (`gda.cli.app`), reversing ADR-0040's direction | **Fully adopted** | The `help` descriptor is now built inside `register(root)` and its recipe closes over that `root` — the same closure rule `gda schema` follows. `gda.commands.meta` no longer imports `gda.cli` anywhere. Your custom-composition repro is the new regression sentinel: a custom root carrying a `local-only` command the global app lacks; `help local-only --json` must answer `gda local-only`. | Red-proofed: the sentinel failed against `37b1deae` with exactly the reported symptom (exit 2, `unknown_command` from the global tree) and passes at `e24ae8e8`. Real-engine spot check: `test_e2e_info.py` + `test_e2e_scene.py` (27 passed) — covering both an inheriting and a non-inheriting dispatch through the renamed field. |

No behavior or surface change for any shipped invocation: the schema manifest hash is unchanged, and the divergence was reachable only through custom composition of the meta slice.

qinhaihong-red:
Recheck result for exact head `e24ae8e8`: **the implementation fixes are correct, but one P2 authority/documentation residue remains before both findings can be considered fully closed.**

## Closed: help uses its registered composition root

This finding is fully resolved. The help descriptor is now built inside `register(root)` and its recipe closes over that root; `gda.commands.meta` no longer imports `gda.cli`. The new custom-composition sentinel reproduces the old failure against `37b1deae` and passes here: `help local-only --json` returns `gda local-only` from the custom tree.

## Partially closed: project inheritance versus explicit project acceptance

The implementation choice is sound and smaller than adding a three-state enum. `inherits_project` now records exactly one decision—whether `$GDA_PROJECT`/cwd is inherited—while each command signature remains the authority for whether explicit `--project` is accepted. ADR-0006, ADR-0023, and the bundled Skill now describe that split consistently.

However, the same old contract remains in two additional tracked locations:

- [ADR-0014](https://github.com/aigengame/godot-agent/blob/e24ae8e83e51a4c2a0a3013ade8e02c82ac3cb73/docs/adr/0014-gda-mcp-project-context-resolution.md#L21-L25) still says meta commands such as `info` “take no project” and reject `--project` outright. `info` now accepts an explicit flag; what it ignores is the MCP-provided/inherited `GDA_PROJECT` environment context.
- [`gda.mcp.project_context`'s module contract](https://github.com/aigengame/godot-agent/blob/e24ae8e83e51a4c2a0a3013ade8e02c82ac3cb73/src/gda/mcp/project_context.py#L16-L20) repeats that `info` “takes no project.” Its later inline comment already uses the correct wording (“never inherit”), so the module-level explanation should match it.

The PR body’s **“3. `gda info --project`, validated”** section also still says the `projectless` field “grows” to combine inheritance and explicit validation. That describes the pre-fix design; the current field is `inherits_project`, and explicit acceptance belongs to the CLI signature.

The minimum remaining change is documentation-only: add a dated ADR-0014 amendment, update the module docstring, and refresh that PR-body paragraph. No further implementation change or enum is needed.

Validation at `e24ae8e8`: 207 focused unit/contract tests passed; 27 real-Godot `info`/`scene` e2e tests passed; the custom-root sentinel passed; `skill`/`version`/`help --project` still return `unknown_option`; the pre/post-fix `gda schema` SHA-256 is identical; and required CI checks are green (the CI Godot e2e job is skipped).

qinhaihong-red:
All three documentation residues are fixed at `b7b04d91`; the PR body is refreshed.

| Residue | Resolution |
| --- | --- |
| ADR-0014 still said meta commands "take no project" and reject `--project` outright | Dated amendment added directly under the decision paragraph: `gda info` now accepts a validated explicit `--project` (#670, ADR-0006 amendment); "take no project" / "reject outright" still describe `skill`/`version`/`help`. What the ADR relies on is called out as unchanged — meta commands never **inherit** an injected `GDA_PROJECT` (the `inherits_project` field, ADR-0023), so gda-mcp still needs no per-command knowledge and never forges a `--project` flag. The original decision prose is preserved point-in-time. |
| `gda.mcp.project_context` module docstring repeated "take no project" | Now reads "meta commands (`info`), which never inherit a project (#670), ignore it" — matching its own inline comment and the field's contract. |
| PR-body section 3 described the pre-fix design (`projectless` "grows") | Rewritten: the field is `inherits_project` (default `True`; meta commands `False`), recording only inheritance; accepting an explicit `--project` stays each command signature's decision. |

I also swept the repo for the same contract elsewhere (`take(s) no project`, `no --project`, meta-command project wordings): every remaining hit is either the amendment-covered ADR-0014 decision prose, an accurate statement about the option-less meta commands, or unrelated (`Godot takes no project lock`, `gda project info`).

Documentation-only round: no code path changed. Gates at `b7b04d91`: fast suite 1632 passed, `ruff`/format clean.

Issue #36, opened 2026-06-10
--schema eager-callback: fix precedence, usage validation, and value binding


qinhaihong-red:
## Parent

#17 — PRD: gda domain command surface. Found during the high-effort code review of #18's implementation (merged via #41).

## What to build

The shared \`_schema_option\` eager callback introduced three regressions (all verified via \`CliRunner\`):

- **Precedence vs \`--help\`.** Both flags are now eager, so the winner depends on argv order: \`gda info --schema --help\` emits schema JSON instead of help. Previously \`--help\` always won.
- **Swallows usage errors.** The callback exits 0 before Click validates positional/required args, so \`gda scene create --schema a b c\` exits 0 with schema output instead of a usage error (exit 2). A malformed argv that happens to include \`--schema\` is indistinguishable from a valid \`--schema\` call.
- **Wrong value binding.** The callback returns \`None\` on the non-emitting path, and Click binds the callback's return as the parameter value, so every command body receives \`schema=None\` instead of the declared \`bool\` default \`False\` — a latent type-contract violation (harmless only because no body currently reads it).

Fix the schema-flag mechanism so it emits only when intended, does not suppress usage validation, and binds a correct boolean.

## Acceptance criteria

- [ ] \`--help\` takes precedence over \`--schema\` regardless of argv order.
- [ ] An invocation with extra/unknown positional args still fails with a usage error even when \`--schema\` is present.
- [ ] Command bodies receive \`schema=False\` (not \`None\`) when the flag is absent.
- [ ] Existing \`--schema\` emission behavior (model-derived \`{input, output}\`, no Godot spawned) is preserved, with tests for each of the three fixes.

## Blocked by

None - can start immediately.

PR #719, opened 2026-08-18
feat(gda)!: argv-binding metadata in the emitted schemas and a discriminated input-event union

### LINKED ISSUE #669: schema: argv-binding metadata and a discriminated input-event union

qinhaihong-red:
Closes #669

## What this fixes

Two dogfooding defects on the emitted contract:

- **GDA-DF-003** — the schema stated a command's required fields but not their CLI
  spelling, so argv could not be built from it: `screen capture` takes its required
  output as `--output`, `input mouse-click` takes `x y` positionally, `input action`
  takes a positional ACTION it rejects as `--action`.
- **GDA-DF-037 / GDA-DF-032** — one flat `input sequence` event shape carried every
  kind's fields, so the per-kind rules lived only in prose, a foreign field was
  accepted and silently ignored (a `release` on a key event PRESSED the key), and the
  refusal of `pressed` on an action event never said to use `release`.

## Design

**1. `argv`, a projection of the live Typer/Click parameters.** Both schema sites —
the per-command `--schema` and the `gda schema` manifest — now carry an `argv` list
built by one shared derivation (`command_argv_bindings`), the same shape
`command_constraints` already has. Each binding says how to write one parameter:

```json
{"name": "output", "input_property": "output", "kind": "option",
 "option": "--output", "position": null, "required": true,
 "flag": false, "multiple": false, "json_value": false}
```

`position` orders the positionals, `option` gives the spelling, `flag` marks an
option that takes no value, `multiple` one that is repeated per value,
`json_value` one whose single token carries the value's JSON (`input sequence
--events` is the only one on the surface — an agent reading `input` sees an array
and would otherwise write one token per element), and `required` is the DECLARED
requirement, not the relaxed one the `--schema` probe parses with (#36).

The Typer signature stays the single source: no `argv` field was added to the
`HeadlessCommand` descriptor, because a hand-maintained spelling table is the
parallel registry ADR-0023 §2 rejects and would rot the moment a signature changed.
Reading the registered parameters also makes the projection correct on every
dispatch channel at once — the sentinel path, the EXPORT / LIVE kinds, and the
recipe commands (`export run`, the `daemon` lifecycle, `screen`), which is asserted
by a test rather than assumed.

`argv` is the argv form of the OPERATION parameters — the set `--params-json` is
exclusive with (ADR-0015) — so the cross-cutting flags every command shares stay
out. It is not a one-to-one image of `input`: **every required property has a
binding** (pinned by a test), while an optional property the CLI computes from flags
rather than takes directly (`script set`'s `mode`, from `--replace` / `--search`)
has none. `input_property` links a binding back to the property it fills, derived
from the parameter name or the long option; it is `null`, never guessed, for the two
flags on the surface that rename their property (`project list --all` →
`include_defaults`, `skill --dir` → `install_dir`). 130 of 132 bindings link.

**2. `input sequence`'s event type is a discriminated union on `type`.** One variant
per kind, each declaring exactly the fields it accepts, so the emitted schema is a
`oneOf` with a discriminator mapping and every kind's required fields and
`additionalProperties: false` boundary are machine-checkable. `mouse_button`'s
"exactly one of `pressed: true` / `release: true`" rule is published as schema too —
that kind is the one an agent reaches for to build a drag, which is the case
GDA-DF-037 named — and a corpus test requires the published rule and the enforcing
validator to agree on every combination of the two fields.

**3. A foreign field is refused by name AND answered.** The message says which
spelling this kind uses for its press/release phase (`release` for an action,
`released` for a key, `pressed`/`release` for a mouse button), or that the kind has
none and which kinds do. Every part of it is derived from the variant models — the
union's own membership included — so it cannot drift from what they accept.

**4. Help.** The `--events` help gains the action press/release pair beside the mouse
drag example. Both are spelled with a space after each `:` and `,` so the help
renderer WRAPS them instead of ellipsizing an unbreakable word — the old one-line
mouse example was cut off mid-JSON at every terminal width, so it was never copyable.

## gda-mcp wire-schema impact (the recorded answer)

**The `argv` addition changes no registered MCP tool's wire schema.** gda-mcp maps
`input` → `input_schema` and `output` → `output_schema` and ignores every other entry
key (ADR-0012); `argv` is a SIBLING of those halves. A test emits each command's
contract with and without the bindings and requires both halves to be identical, so
this is asserted, not argued.

**One tool's `input_schema` does change, for an unrelated reason in the same PR:**
`gda_input_sequence`, because that command's own params model became the per-kind
union. Measured by diffing `gda schema` before and after, across all 70 commands:

| entry key | commands whose value changed |
| --- | --- |
| `input` | `input sequence` only (its own model's union) |
| `output` / `error` / `description` / `kind` / `constraints` | none |
| new keys | `argv` |

Both schema lookup paths were checked, and they stay pinned separately:

1. **The aggregate manifest → gda-mcp tool registration** (`gda.mcp.server.build_server`,
   pinned by `tests/test_mcp_registration.py` and `tests/test_schema_aggregate.py`).
2. **The per-command `--schema` document** (ADR-0004, pinned by
   `tests/test_schema_command.py`) — gains the additive `argv` key.

The two are asserted to agree binding-for-binding
(`test_entry_argv_matches_the_commands_own_schema_argv`) without being merged.

`libs/gda-balancing`'s own two `wire_schema` lookup paths
(`artifact_wire_schema_identity` / `wire_schema_identity_for_kind`) are a different
subsystem with a separate release train; nothing there derives from gda's command
surface and no file under `libs/` is touched.

## BREAKING CHANGE

`input sequence` no longer accepts an event field belonging to another event kind.
Five shapes that validated before are now rejected — for example
`{"type": "key", "key": "A", "release": true}` and
`{"type": "mouse_move", "x": 1, "y": 2, "button": "right"}`. None of them ever did
anything: the harness reads only the kind's own fields, so those values were
discarded, and the `release`-on-a-key case silently PRESSED the key. Turning them
loud is the point of the change. The published `input_schema` for the
`gda_input_sequence` MCP tool changes shape with it (a flat object → a `oneOf` with a
discriminator), so a client that generated code from it should regenerate.

`button: null` stays accepted (it means left, as before) — the old flat shape
defaulted it to null, so a producer replaying a dumped event would otherwise have
broken.

## Validation

**Gates** — `uv run pytest -q -m "not e2e"` 1661 passed; `ruff check`, `ruff format
--check`, `pyright` all clean. PR CI green.

One CI-only failure was found and fixed on this branch: the help renderer emits SGR
colour sequences when it believes it is writing to a terminal, which it does on
GitHub Actions and not in a local run, so the new copyable-example assertion passed
locally and failed only in CI. The mechanism was reproduced locally with
`GITHUB_ACTIONS=true uv run pytest` before fixing it; the fast suite now passes in
both modes and is run in both.

**Real-engine repro of GDA-DF-003** (Godot 4.x, macOS). A script builds argv for
seven commands from the emitted schema ALONE — no `--help`, no guessing — and runs
them against a real project. Before this change the script cannot run at all: there
is no `argv` key. After, every command succeeds, including the spellings that used to
require a round trip:

```
$ gda scene create res://Main.tscn --root-type Node2D          exit 0
$ gda node add res://Main.tscn --type Sprite2D --name Hero     exit 0   (param node_type → --type)
$ gda script create res://hero.gd --extends Sprite2D           exit 0   (param extends_type → --extends)
$ gda script validate res://hero.gd                            exit 0   (variadic positional)
$ gda resource create res://box.tres --type RectangleShape2D   exit 0
$ gda node set res://Main.tscn --node Hero --property z_index --value 5 exit 0
$ gda node get res://Main.tscn --node Hero                     exit 0
```

**Real-engine verification of the union (GDA-DF-032/037)** — the live e2e drives a
real `gda daemon start`, a real engine session and the real harness with the new
per-kind payload: `tests/test_e2e_input.py` 10 passed, including
`test_input_sequence_physics_frame_hold_matches_predicted_displacement` (an action
press at `physics_frame: 0` and its `release: true` at `physics_frame: N`, verified
against the predicted displacement) and `test_input_sequence_drags_mouse_with_held_button_mask`
(the `mouse_button` phase events). The engine-facing payload is unchanged in kind:
the harness reads every sequence field through `event.get(key, default)`, so an event
that now omits the other kinds' keys behaves exactly as before.

**Rejection message, before → after** (`--params-json`, engine-free):

| event | before | after |
| --- | --- | --- |
| `{"type":"action","action":"jump","pressed":true}` | `'pressed' is only valid on a 'mouse_button' sequence event.` | `'pressed' is not valid on an 'action' sequence event: an 'action' event spells its press/release phase 'release'; 'pressed' is accepted on: mouse_button.` |
| `{"type":"key","key":"A","release":true}` | **accepted** — the key was PRESSED | `'release' is not valid on a 'key' sequence event: a 'key' event spells its press/release phase 'released'; 'release' is accepted on: mouse_button, action.` |
| `{"type":"mouse_move","x":1,"y":2,"pressed":true}` | `'pressed' is only valid on a 'mouse_button' sequence event.` | `'pressed' is not valid on a 'mouse_move' sequence event: a 'mouse_move' event has no press/release phase; 'pressed' is accepted on: mouse_button.` |
| `{"type":"nope",…}` | `Input should be 'key', 'mouse_click', …` | same list of wire values (the enum reports its value, so the tag list never names a Python symbol) |

**Other targeted e2e** (serialized under the shared lock): `test_e2e_input`,
`test_e2e_mcp_stdio`, `test_e2e_params_json` re-run after the review fixes — 16
passed; earlier rounds also ran `test_e2e_mcp_tracer`, `test_e2e_info`,
`test_e2e_mcp_live`, `test_e2e_daemon`, `test_e2e_game` — 41 passed.

## Review round

The branch was put through two independent adversarial reviews (spec-vs-issue and
code-standards) before this was opened for human review. Everything they found is
fixed in `66323231`, each with a guard so the class cannot return silently: the
union's membership is now read off the union instead of re-listed beside it; the
message builder no longer raises a bare `KeyError` from inside a refusal path; the
mouse-button phase rule is published; `button: null` is accepted again;
`json_value` was added; ADR-0012's "no wire schema changes" sentence was corrected;
and three docs (including the published field descriptions) stopped describing a
null-`input_property` rule the code does not implement.

## Residual risks / known limits

- `gda schema` grows ~6% (507 KB → 537 KB) for the 132 bindings. Field descriptions
  are deliberately terse for that reason.
- A foreign field masks a co-occurring missing required one: the pre-validation
  refusal fires first, so `{"type":"action","pressed":true}` reports `pressed` and
  only then the missing `action`. Two round trips where one would do.
- Three per-event rules stay enforced model-side only, as before this change: the
  one-clock rule (a relation between two nullable fields — expressing it in schema
  would wrongly reject `{"frame": 1, "physics_frame": null}`), the modifier
  vocabulary, and the window ceiling. Candidates for a follow-up, out of scope here.
- **Considered and declined:** renaming the two Python parameters behind `--all` /
  `--dir` to match their properties would make all 132 bindings link. It changes two
  unrelated command groups to make this metadata tidier, both properties are
  optional, and the null is honest rather than wrong — revisit if a REQUIRED property
  ever ends up unlinked (the new test fails loudly if one does).
- **Not changed, deliberately:** the README's one-line `--schema` flag description —
  it is still true, and every README edit forces a re-translation of the three
  localized READMEs (`test_readme_i18n_sync`) for a non-load-bearing clause. The
  agent-facing surfaces that carry the contract are updated: ADR-0004, ADR-0012, the
  command catalog, the bundled Skill, and the emitted schema itself.
- Scope boundary with #670 (near-miss command/option hints) is untouched.

### DIFF
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -156,7 +156,8 @@ status: accepted
 > authoritative. The projection is the argv form of the operation parameters, not a
 > one-to-one image of `input`: every REQUIRED property has a binding, held by a
 > test, while an optional property the CLI computes from flags rather than takes
-> directly (`script set`'s `mode`, from `--replace` / `--search`) has none. On the
+> directly (`script set`'s and `shader set`'s `mode`, from `--replace` /
+> `--search`) has none. On the
 > aggregate entry the `argv` key is required, its list possibly empty — the same
 > "key always present" guarantee `constraints` has.
 
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -112,8 +112,8 @@ status: accepted
 > **Amendment (2026-08-18, #669):** the per-command `--schema` object gains a SIXTH,
 > additive key — `argv`, the list of `ArgvBinding`s naming how each parameter `input`
 > describes is written on a command line (`kind` positional/option, `position` or
-> `option` spelling, `required`, `flag`, `multiple`, and the `input_property` it
-> fills). Motivation: the contract stated a command's required fields but not their
+> `option` spelling, `required`, `flag`, `multiple`, `json_value`, and the
+> `input_property` it fills). Motivation: the contract stated a command's required fields but not their
 > CLI spelling, so an agent could not build argv from it — `screen capture` takes its
 > required output as `--output`, `input mouse-click` takes `x y` positionally, and
 > `input action` takes a positional ACTION it rejects as `--action` (dogfooding
@@ -121,12 +121,15 @@ status: accepted
 >
 > It is additive on the same properties the notes above establish:
 >
-> - **A sibling of the schema halves, never a key inside them.** `input` / `output`
->   stay byte-identical, so gda-mcp's `input_schema` / `output_schema` are unchanged
->   and it needs no adapter work — it maps only `input` / `output` / `description`
->   (ADR-0012). Measured across the whole surface at the time of the change: one
->   `input` differs (`input sequence`, for its own model's per-kind union), and no
->   `output` / `error` / `description` / `kind` / `constraints` does.
+> - **A sibling of the schema halves, never a key inside them.** Adding the key
+>   leaves `input` / `output` byte-identical — a test emits each command's contract
+>   with and without the bindings and requires both halves to match — so gda-mcp's
+>   `input_schema` / `output_schema` are unchanged by it and it needs no adapter
+>   work; it maps only `input` / `output` / `description` (ADR-0012). Measured
+>   across the whole surface at the time of the change, exactly one `input` differs
+>   for an unrelated reason (`input sequence`, whose own params model became a
+>   per-kind union in the same PR), and no `output` / `error` / `description` /
+>   `kind` / `constraints` differs at all.
 > - **Derived, never declared.** The bindings are read off the LIVE Typer/Click
 >   parameters at emission time, through one projection shared by the per-command
 >   `--schema` and the aggregate manifest (ADR-0012's live-tree walk, ADR-0023 §2).
@@ -136,16 +139,26 @@ status: accepted
 >   Reading the registered parameters also makes it correct on every dispatch channel
 >   at once — the sentinel path, the EXPORT / LIVE kinds, and the recipe commands.
 > - **Zero per-command cost**, like the halves it joins: a new command's spelling
->   appears because it registered parameters, not because anyone wrote it down.
+>   appears because it registered parameters, not because anyone wrote it down. The
+>   binding expresses a positional, an option, a valueless flag, a repeated value
+>   and a JSON-encoded one; a registration test fails if the surface ever grows a
+>   Click shape it cannot write (a `--x/--no-x` pair, an n-ary option, a counting
+>   option), so the contract is extended deliberately rather than emitting a
+>   binding no caller can follow.
 >
 > Two boundaries. `argv` covers the OPERATION parameters — the same set
 > `--params-json` treats as exclusive of the individual arguments (ADR-0015) — so the
 > cross-cutting flags every command shares (`--json` / `--schema` / `--params-json` /
 > `--godot` / `--project`) stay out; they are not per-command information. And
-> `input_property` is `null`, not guessed, where a CLI form has no 1:1 property (two
-> flags selecting one field, or a differently-named option): a wrong link would read
-> as authoritative. On the aggregate entry the `argv` key is required, its list
-> possibly empty — the same "key always present" guarantee `constraints` has.
+> `input_property` is `null`, not guessed, where the flag's spelling renames the
+> property it fills (`project list --all` fills `include_defaults`, `skill --dir`
+> fills `install_dir` — the only two on the surface): a wrong link would read as
+> authoritative. The projection is the argv form of the operation parameters, not a
+> one-to-one image of `input`: every REQUIRED property has a binding, held by a
+> test, while an optional property the CLI computes from flags rather than takes
+> directly (`script set`'s `mode`, from `--replace` / `--search`) has none. On the
+> aggregate entry the `argv` key is required, its list possibly empty — the same
+> "key always present" guarantee `constraints` has.
 
 ADR-0000 lists `--schema` as a core capability without defining it. We fix its
 semantics here, and deliberately scope out an overloaded interpretation.
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -150,14 +150,16 @@ status: accepted
 > `--params-json` treats as exclusive of the individual arguments (ADR-0015) — so the
 > cross-cutting flags every command shares (`--json` / `--schema` / `--params-json` /
 > `--godot` / `--project`) stay out; they are not per-command information. And
-> `input_property` is `null`, not guessed, where the flag's spelling renames the
-> property it fills (`project list --all` fills `include_defaults`, `skill --dir`
-> fills `install_dir` — the only two on the surface): a wrong link would read as
-> authoritative. The projection is the argv form of the operation parameters, not a
-> one-to-one image of `input`: every REQUIRED property has a binding, held by a
-> test, while an optional property the CLI computes from flags rather than takes
-> directly (`script set`'s and `shader set`'s `mode`, from `--replace` /
-> `--search`) has none. On the
+> `input_property` is `null`, not guessed, where a parameter's property is revealed
+> by neither its name nor its long option: a wrong link would read as authoritative.
+> The surface has **no such parameter** — where an option renames the property it
+> fills (`project list --all` → `include_defaults`, `skill --dir` → `install_dir`)
+> the Python parameter carries the property's name, so the link resolves — and a
+> test holds that every **directly supplyable** property has a binding. The nullable
+> value stays part of the contract for a future parameter that cannot be resolved,
+> rather than being forced into a guess. The one exemption from that guard is a
+> property the CLI COMPUTES rather than takes (`script set`'s and `shader set`'s
+> `mode`, from `--replace` / `--search`), which its own description declares. On the
 > aggregate entry the `argv` key is required, its list possibly empty — the same
 > "key always present" guarantee `constraints` has.
 
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -109,6 +109,44 @@ status: accepted
 > an operation reported. The two compose under the same omitted-when-absent rule, and
 > #687 stays free to decide its axis on its own merits.
 
+> **Amendment (2026-08-18, #669):** the per-command `--schema` object gains a SIXTH,
+> additive key — `argv`, the list of `ArgvBinding`s naming how each parameter `input`
+> describes is written on a command line (`kind` positional/option, `position` or
+> `option` spelling, `required`, `flag`, `multiple`, and the `input_property` it
+> fills). Motivation: the contract stated a command's required fields but not their
+> CLI spelling, so an agent could not build argv from it — `screen capture` takes its
+> required output as `--output`, `input mouse-click` takes `x y` positionally, and
+> `input action` takes a positional ACTION it rejects as `--action` (dogfooding
+> GDA-DF-003). The self-description was accurate and still insufficient.
+>
+> It is additive on the same properties the notes above establish:
+>
+> - **A sibling of the schema halves, never a key inside them.** `input` / `output`
+>   stay byte-identical, so gda-mcp's `input_schema` / `output_schema` are unchanged
+>   and it needs no adapter work — it maps only `input` / `output` / `description`
+>   (ADR-0012). Measured across the whole surface at the time of the change: one
+>   `input` differs (`input sequence`, for its own model's per-kind union), and no
+>   `output` / `error` / `description` / `kind` / `constraints` does.
+> - **Derived, never declared.** The bindings are read off the LIVE Typer/Click
+>   parameters at emission time, through one projection shared by the per-command
+>   `--schema` and the aggregate manifest (ADR-0012's live-tree walk, ADR-0023 §2).
+>   An `argv` field on the `HeadlessCommand` descriptor was rejected: a
+>   hand-maintained spelling table is a second authority for a fact the Typer
+>   signature already owns, and it would silently rot the moment a signature changed.
+>   Reading the registered parameters also makes it correct on every dispatch channel
+>   at once — the sentinel path, the EXPORT / LIVE kinds, and the recipe commands.
+> - **Zero per-command cost**, like the halves it joins: a new command's spelling
+>   appears because it registered parameters, not because anyone wrote it down.
+>
+> Two boundaries. `argv` covers the OPERATION parameters — the same set
+> `--params-json` treats as exclusive of the individual arguments (ADR-0015) — so the
+> cross-cutting flags every command shares (`--json` / `--schema` / `--params-json` /
+> `--godot` / `--project`) stay out; they are not per-command information. And
+> `input_property` is `null`, not guessed, where a CLI form has no 1:1 property (two
+> flags selecting one field, or a differently-named option): a wrong link would read
+> as authoritative. On the aggregate entry the `argv` key is required, its list
+> possibly empty — the same "key always present" guarantee `constraints` has.
+
 ADR-0000 lists `--schema` as a core capability without defining it. We fix its
 semantics here, and deliberately scope out an overloaded interpretation.
 
--- a/docs/adr/0012-gda-mcp-tool-surface-generation.md
+++ b/docs/adr/0012-gda-mcp-tool-surface-generation.md
@@ -21,6 +21,17 @@ status: accepted
 > `constraints` — so the addition is backward compatible. On the aggregate entry
 > the `constraints` key is always present, its value nullable.
 
+> **Outcome (2026-08-18, #669):** each manifest entry gained an additive `argv`
+> field (the command's `ArgvBinding` list — how each parameter is written on a
+> command line, ADR-0004's amendment of the same date), so the entry above is now
+> `{name, description, input, output, error, kind, constraints, argv}`. gda-mcp's
+> mapping is unchanged — it ignores `kind` / `constraints` / `argv` — and the two
+> halves it DOES map are untouched, so no registered tool's wire schema changes.
+> The bindings come from the same live-tree walk this ADR established: the walker is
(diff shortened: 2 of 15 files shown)


qinhaihong-red:
## Review conclusion

**Request changes.** I reviewed exact head `9ab1711ccae1862367892a52582d928e1cdb8a10` against base `8cec419600d415e10b0d9882ca236a201c25682a`, keeping the domain/modular-architecture, Standards, and Spec axes separate. The event-union path is coherent and passed focused plus real-engine validation, but the schema-to-argv contract does not yet satisfy #669 for every field and every command.

### Domain and modular architecture

1. **[P1] The input-property-to-argv relationship has no complete authority.** `command_argv_bindings()` can associate a Click parameter with a Pydantic input property only when the parameter name or option spelling happens to match. At this head:

   - `project list` publishes `input.include_defaults`, but its `--all` binding has `input_property: null`.
   - `skill` publishes `input.install_dir`, but its `--dir` binding has `input_property: null`.

   A schema-only client therefore cannot deterministically translate concrete input objects containing these fields without guessing or reading help/source. `tests/test_schema_aggregate.py::test_every_required_input_property_has_an_argv_binding` misses both because it narrows the invariant to required fields. Prefer aligning the live Python parameter names with the model properties while retaining the external `--all` / `--dir` spellings, then guard every directly supplyable input property. This preserves the no-parallel-registry direction.

2. **[P2] The accepted ADR set is internally inconsistent.** ADR-0015 still says that the aggregate schema deliberately does not encode CLI binding, rejects enriching it with binding metadata, and states that `--schema` is unchanged. This PR adopts that metadata in ADR-0004 and ADR-0012. Amend or explicitly partially supersede ADR-0015 while preserving the unchanged boundary that gda-mcp dispatches through `--params-json`.

Apart from these two gaps, the responsibility placement and source dependencies are directed: the live Click tree owns CLI spelling, the input models own field semantics, `headless.py` owns the shared projection, and `surface.py` owns aggregate traversal. No new dependency cycle was found.

### Standards

1. **[P2] `ArgvBinding` accepts states its contract declares impossible.** The model and its emitted JSON Schema do not tie `kind` to `option` and `position`. Both of these currently validate:

   - `kind="argument"`, `option="--x"`, `position=null`, `flag=true`
   - `kind="option"`, `option=null`, `position=0`

   Actual emitted bindings are protected by a separate inventory test, but `gda schema --schema` does not describe the promised mutually exclusive shapes. Use a `kind`-discriminated union or equivalent model and JSON-Schema constraints.

2. **[P3] The new sequence variants duplicate public validation facts.** `KeySequenceEvent` and `ActionSequenceEvent` independently redeclare constraints already owned by `InputKeyParams` and `InputActionParams`: required/min-length fields, defaults, phase fields, and action-strength bounds. There is no derivation or parity guard, so the single-event and sequence contracts can drift. A shared field definition or a complete model/schema parity test would satisfy the repository's single-authority rule without requiring a broad redesign.

### Spec

1. **[P1] #669's core acceptance criterion remains partial.** The issue asks for argv-binding metadata for **every field** and for an agent to construct argv for **any command from the emitted schema alone**. The two null associations above require guessing, while the new aggregate guard silently weakens that requirement to every *required* property. The computed `script set` / `shader set` `mode` properties are different because their descriptions explicitly say supplied values are ignored; the two real optional flags already establish the acceptance gap.

The other #669 requirements passed review: all five sequence-event kinds publish discriminated required/forbidden field sets; the mouse-button phase rule is schema-checkable; wrong-kind errors name the accepted alternative; and the gda-mcp wire-schema impact is recorded.

### Validation

- `git diff --check 8cec4196...9ab1711c` — passed
- Focused schema/aggregate/input tests — `173 passed`
- Real Godot input e2e, serial — `10 passed`
- Required PR checks — green at the reviewed head; the GitHub Godot e2e job is skipped, with the relevant input e2e covered locally

qinhaihong-red:
All five findings are addressed at `8a33adbf`.

| --- | --- | --- | --- |
| [P1] `include_defaults` / `install_dir` had no reachable spelling (`input_property: null`); the guard covered required properties only — same root as the Spec P1 | **Adopted, your preferred variant** | The Python parameter names now match the model properties (`all_settings` → `include_defaults`, `dir` → `install_dir`) while the external `--all` / `--dir` spellings are unchanged, so `_bound_property` links both by name — the surface now has **zero** null links, with no descriptor field and no parallel registry. The guard is extended to every **directly-supplyable** input property; the exclusion is mechanical (a property whose description says "a value passed in is ignored") and pinned to exactly `{script set: mode, shader set: mode}`, so the phrase cannot become a silent escape hatch. ADR-0004's amendment, the published `input_property` description, and SKILL.md no longer cite the null cases; `input_property` stays nullable by contract for a future unresolvable parameter, pinned by a unit test on `_bound_property` itself. | Red-proofed twice, independently: at `9ab1711c` the extended guard fails naming exactly `project list: include_defaults` and `skill: install_dir`. Renames touch parameter names, not models: `input`/`output` halves stay byte-identical (schema-halves test), and against the pre-#669 baseline only `input sequence`'s `input` differs. `project list --all` exercised in real-engine e2e (`test_e2e_project.py` + `test_e2e_project_analysis.py`, 40 passed under the serial lock). |
| [P2] ADR-0015 inconsistent with the adopted binding metadata | **Adopted** | Dated 2026-08-19 amendment, historical prose untouched: it names the two superseded sentences, and states what remains in force — gda-mcp still forwards `--params-json` verbatim, reconstructs no argv, ignores `argv`, so the rejected option's cost (typed-argv encoding inside gda-mcp) never materializes. Two audiences, two channels, one params model. | Doc-only; amendment discipline (dated note, no rewrite). |
| [P2] `ArgvBinding` accepts impossible states | **Adopted via your "equivalent" branch** | `_check_spelling` model validator rejects the five impossible combinations; `_ARGV_BINDING_SPELLING_SCHEMA` publishes the same pairing as `json_schema_extra` `oneOf` — the `_MOUSE_BUTTON_PHASE_SCHEMA` pattern. A kind-discriminated union was deliberately not taken: it changes the field set and adds a shared base for the same checkability. | 9-case corpus through both `jsonschema` and pydantic requiring identical verdicts, plus non-vacuity (both verdicts occur). Red-proofed by unpublishing the rule. |
| [P3] Sequence variants redeclare single-op constraints with no drift guard | **Adopted as a parity test** | `test_a_sequence_variant_keeps_its_single_frame_ops_constraints` compares the mirrored pairs' constraint metadata and defaults (not descriptions, which legitimately differ), asserting ≥10 fields actually compared. No extraction — the duplication is now constrained by the guard, per the repo's consistency-test channel. | Red-proofed by dropping `le=1.0` from `ActionSequenceEvent.strength`. |
| [P1 Spec] AC1 partial for optional properties | **Adopted** | Same fix as the first row: every directly-supplyable property now carries a binding, so an agent can construct argv for any command from the emitted schema alone; the computed `mode` properties are excluded on the stated ground that their descriptions declare supplied values ignored. | The extended guard + the zero-null probe over all 70 commands / 132+ bindings. |

Gates at `8a33adbf`: fast suite 1665 passed (also under `GITHUB_ACTIONS=true`), ruff / format / pyright clean, PR CI green (7 pass, Godot e2e job dispatch-only as usual — input & project e2e run locally under the serial lock).

qinhaihong-red:
## Recheck conclusion

Rechecked exact head `8a33adbffcf8ec82d02b1e071695ab6220685adc` against the five findings in the prior review and the author's reply. Four findings are fully closed: both schema-to-argv/Spec findings, ADR-0015 consistency, and the original `ArgvBinding` `kind`/`option`/`position` invalid states. One P3 residue remains.

### [P3] The parity guard still misses removed or retyped mirrored fields

`tests/test_input_commands.py::test_a_sequence_variant_keeps_its_single_frame_ops_constraints` compares only:

```python
shared = set(variant.model_fields) & set(params.model_fields)
```

It then checks `metadata` and `default`, with only a global `compared >= 10` floor. A mirrored field that disappears from one model is removed from `shared`; annotation and `default_factory` changes are not compared either.

Reproduction against this head: remove `KeySequenceEvent.released` and `ActionSequenceEvent.strength` from the comparison view. The guard still reaches `drift == []`, `compared == 10`, and passes. Thus the original drift class is only partially guarded even though the `le=1.0` red proof works.

Minimal fix: pin the expected mirrored fields for each model pair, assert each field exists on both sides, then compare `annotation`, `metadata`, `default`, and `default_factory` where parity is required. This preserves the chosen consistency-test approach and needs no model extraction.

### Validation

- Focused schema/input/project/skill tests: `255 passed`
- Real Godot `project list --all` e2e: `2 passed`
- Required PR checks: green at `8a33adbf`

No other prior finding remains open.

qinhaihong-red:
The P3 residue is fixed at `a504cc19`.

| Residue | Resolution | Validation |
| --- | --- | --- |
| The parity guard's dynamic intersection shrinks on removal/rename, and annotation / `default_factory` were not compared | Your minimal fix, as suggested: the pairing table now **pins each pair's mirrored fields** (12 across the four pairs); a pinned field missing on either side is itself reported as drift, and the comparison covers `annotation`, `metadata`, `default`, and `default_factory`. The one deliberate divergence — `MouseClickSequenceEvent.button` accepting an explicit null for old flat-shape dumps — is a pinned exemption that must be exactly `Optional[<the op's annotation>]` and must name a pinned field, so it can neither widen nor go stale silently. No extraction; the consistency-test approach is preserved. | Red-proofed on all three drift classes at this head: (1) a single-side removal of `KeySequenceEvent.released` fails with "pinned mirrored field missing"; (2) retyping `ActionSequenceEvent.strength` to `int` fails on the annotation comparison; (3) the original `le=1.0` drop still fails. Fast suite 1665 passed, ruff / format / pyright clean, PR CI green. |

qinhaihong-red:
The stale-exemption residue is fixed at `ea6b4e16`.

| Residue | Resolution | Validation |
| --- | --- | --- |
| A resolved nullable divergence kept passing: `Optional[Optional[T]]` collapses to `Optional[T]`, so the exemption never went stale loudly — the pinned-membership assertion only caught a removed field | As suggested: before applying the exemption the guard now requires the op-side annotation to be non-nullable (checked structurally: `repr(there.annotation | None) == repr(there.annotation)` detects the collapse); a resolved divergence is reported as drift naming the exemption entry to delete. The variant side forgetting its nullability is already caught by the expected-annotation comparison, so both stale forms now fail. The comment that previously over-claimed "fails loudly" for both forms is corrected to attribute each form to its mechanism. | Red-proofed: making `InputMouseClickParams.button` nullable (your reproduction) fails with `stale exemption ('MouseClickSequenceEvent', 'button'): … delete the exemption`; restored, the suite is green. Fast suite 1665 passed, ruff / format / pyright clean. |

Issue #31, opened 2026-06-10
Headless op dispatch must guarantee exactly one clean process exit


qinhaihong-red:
## Parent

#17 — PRD: gda domain command surface. Found during the high-effort code review of #18's implementation (branch \`feat/18-scene-create-get-tracer\`, not yet merged).

## What to build

The \`operations.gd\` dispatcher uses an \`ok\` bool plus \`if ok: quit()\`. If an operation hits an uncaught GDScript runtime error (e.g. a future op, or any internal failure), it aborts **before** \`quit()\` is reached and the headless \`SceneTree\` main loop runs forever. Through \`gda\` this burns the full 60s runner timeout and is then **misclassified** as \`environment / launch_timeout\` (exit 124) rather than an operation failure.

Restructure the dispatch so the process **always** exits, exactly once, with the right code: an operation returns a result-or-error value and \`_init\` alone owns emitting the result and calling \`quit()\` (one call site). A pre-dispatch/usage error and an unknown operation must also exit non-zero with a stable, structured error code rather than the generic \`operation_failed\`. This also removes the \`_fail\`/\`_fail_op\` near-duplication (one failure helper).

Note: the related structured-failure *channel* decision (how an op reports its error code — stderr marker vs the ADR-0002 sentinel) is tracked separately; this issue is about the control flow guaranteeing a single exit and giving dispatcher-level failures a stable code via whatever channel exists.

## Acceptance criteria

- [ ] An operation that raises a runtime error exits promptly and non-zero (no 60s hang), and \`gda\` classifies it as an operation failure, not \`launch_timeout\`.
- [ ] \`quit()\` is reached on every path; no path can emit a result and then have the exit code reset to 0.
- [ ] Dispatcher-level failures (unknown operation, missing/invalid params payload) carry a stable error code, not the generic \`operation_failed\`.
- [ ] \`_fail\`/\`_fail_op\` are unified into a single failure helper.

## Blocked by

None - can start immediately.

Issue #30, opened 2026-06-10
scene get: read scene state without executing scene code


qinhaihong-red:
## Parent

#17 — PRD: gda domain command surface — Phase-1 headless increment. Found during the high-effort code review of #18's implementation (branch \`feat/18-scene-create-get-tracer\`, not yet merged).

## What to build

\`gda scene get\` currently loads the \`.tscn\` and **instantiates** it to walk the tree. Instantiating runs the \`_init\` of any script attached to a node, so *reading* a scene executes arbitrary project code, and any \`print()\` that code emits lands on stdout ahead of the real result — where a sentinel-shaped line is taken as the (forged) result because the parser takes the first sentinel match.

Read the structured tree **without instantiating**: derive node name/type/children from the packed scene's state introspection (e.g. \`PackedScene.get_state()\` → \`SceneState\`), not from a live node tree. A read must never execute scene code.

## Acceptance criteria

- [ ] \`gda scene get\` on a scene whose root has an attached script with a side-effecting \`_init\` (e.g. writes a file or prints) does **not** execute that script — no side effect, no script output on stdout.
- [ ] The reported tree (root name/type, nested children) is unchanged for script-free scenes (round-trip with \`scene create\` still matches).
- [ ] An e2e regression test (\`@pytest.mark.e2e\`) builds a scene with a script that would print a forged \`<<<GDA:RESULT>>>…\` block and asserts the real tree is reported and the side effect did not happen.

## Blocked by

None - can start immediately.

qinhaihong-red:
Post-merge finding (2026-06-11): the operation-level guarantee here holds (`scene get` / `node list` never instantiate), but it does not extend to the process level — a project's **autoloads** are constructed by the engine on every `--project` run, before the operation script executes. Verified on Godot 4.6. Tracked as #61; this issue's claim should be read as scoped to *scene scripts*, not all project code.

PR #720, opened 2026-08-19
feat(gda): static scene validation and a bounded startup preflight (#664)

### LINKED ISSUE #664: scene: static validation and bounded startup preflight

qinhaihong-red:
Closes #664

Two commands, delivered as two slices, closing the two dogfooding defects the issue names.

## What was broken

Reproduced against Godot 4.6.3 before writing anything, with the real engine:

```
$ gda scene get res://missingres.tscn --json      # script AND texture both missing
{"path":"res://missingres.tscn","root":{"name":"MissingRes","type":"Node2D",
 "children":[{"name":"Spr","type":"Sprite2D","children":[]}]}}   # exit 0, looks healthy
```

Loading a broken scene **succeeds**: Godot substitutes null for an `[ext_resource]` it cannot
resolve, prints an error to stderr, and still hands back a usable `PackedScene`. So `scene get`
reports a healthy tree for a scene whose script and texture are both gone (**GDA-DF-040**), and
nothing in the surface said whether a scene would actually come up (**GDA-DF-030** — static
validation passed while the first live launch rejected every encounter assembly).

## What was built

**`gda scene validate PATH` — static.** Loads the scene without INSTANTIATING it, so none of the
scene's own node scripts run (issue #30; the project's autoloads still start, as for every
`--project` op). It resolves every dependency the `.tscn` declares and compiles every `.gd` it
attaches. One problem per file that did not resolve: `missing_resource`, `unloadable_resource`
(present on disk but no loader — typically an asset that was never imported, which a non-editor
engine cannot load at all) or `script_compile_failed`. Each carries the declared `type` and the
node paths referencing it, read from the file's own text because the engine drops an unresolvable
reference from what it loads. Invalid is a successful operation (exit 0, `valid: false`).

**`gda scene preflight PATH` — dynamic.** Instantiates the scene, adds it under a one-shot
engine's tree root (which runs its `_ready` and the project's autoloads), keeps it alive for
`--frames` idle frames so startup work landing after `_ready` still prints, and reports `status`
(`ready` / `not_ready` / `timeout`) plus the script errors read off stderr with #651's parser.
`started` requires both halves — reaching `_ready` is not reaching it cleanly, which is exactly
AC2's distinction.

## Design decisions

**A separate `scene preflight` command, not `scene validate --startup`.** The issue asks for "a
structured startup verdict **distinct from static validation**", and a flag switching the result
shape would break ADR-0023's one-descriptor/one-input/one-output registration and its `--schema`
projection. Two commands, two models, two verdicts.

**Under `scene`, HEADLESS kind, no daemon.** The preflight is a one-shot `godot --headless` launch
of gda's own payload — nothing drives or observes the scene from outside, so it is not a
`Live operation` and needs no `Engine session` (ADR-0017/0019). CONTEXT.md gains the
`Startup preflight` term precisely because that boundary is easy to get wrong.

**The preflight calls the launch primitive directly, but dispatches an ordinary sentinel op.** Its
entry script is gda's own `operations.gd`, so the engine can and does report a structured verdict
— but the runner seam's BUFFERED capture discards the child's output at a timeout, and for a
scene whose `_ready` never returns that discarded text is the entire evidence. So the recipe calls
`launch()` with a capture-only watch (#655's streaming path). The argv is the shared
`gda.runner.sentinel_args`, extracted for its second caller so the two channels cannot drift on
how an op is spelled.

**A run gda ends at its bound is a `timeout` VERDICT, not `launch_timeout`.** Elsewhere a timeout
means gda could not get you an answer; here the question was whether this scene comes up within
the bound, so the bound being reached IS the answer (AC3 asks for exactly this: "a structured
timeout verdict within the configured bound, with captured diagnostics"). Recorded as a dated
scope note on ADR-0002, together with the other place this command sits close to its rules: a
success result's own verdict field (`started`) derived from the recognized sentences — the
success/failure OUTCOME still comes only from the exit code plus the sentinel.

**The #651 parser is reused, not forked or widened.** `gda.script_errors.parse_script_errors` is
called as-is (its module docstring already named this consumer). Recognition stays the closed set
it owns.

**Two bounds, each honest about what it measures.** Readiness is settled before the first observed
frame, so `--frames` is NOT a bound on coming up: it is the observation window that lets startup
work landing after `_ready` run and print (measured at ~6.5ms/frame, so the default 10 costs
~65ms — and an e2e test pins that a frame-five failure is caught at the default and missed at
`--frames 1`). `--timeout` is the wall-clock ceiling. A `_ready` that blocks takes the whole engine
with it, so no in-engine budget can ever fire there — only the CLI's ceiling can, which is why AC3
is served by `--timeout` and not by a fake in-engine timer.

**`project_root` on both results**, applying #658's rule where it matters just as much: every
problem either command reports is a `res://` resolution outcome, and the wrong root reports
everything as broken.

**No new error codes, no new envelope keys.** Refusals reuse the scene group's existing ladder
(`path_not_found` / `not_a_scene`, plus `missing_dependency` for a scene the engine cannot
instantiate at all — the same code and wording `scene get-exports` already uses for that exact
condition). ADR-0002's registry is untouched.

**`operations.gd` gained a bounded pending tail.** The preflight needs frames, which the "quit on
the first idle frame" control flow (issue #31) does not give. `_process` now counts and CAPS the
frame BEFORE calling the tick, so a tick aborted by an uncaught runtime error — which makes
`_process` return its default `false` and be called again — still ends the run at the cap instead
of erroring forever. Every other op is untouched (no pending tick → the original path).

**Two shared pieces gained a second consumer and moved down**: `LaunchFn` (the injected launch
seam) to `gda.runner`, and the `<path>:<line>: <message>` script-error line to `gda.render`.

## Review round

Two independent adversarial reviews (spec/convention and correctness) ran against the branch
before this PR opened. Every finding they backed with evidence is fixed, each with a regression
test — the four that mattered:

| Finding | Fix |
| --- | --- |
| A binary `.scn` reported `valid: true` whatever was broken in it (the dependency set comes from the scene's TEXT) | `scene validate` takes a `.tscn`; a `.scn` is refused with `invalid_path`, naming why |
| A dependency broken from a `[sub_resource]` (an `AtlasTexture`'s atlas, a script-backed `Resource`) hard-fails the whole load, so gating on it answered `not_a_scene` about the very dependency the command exists to report | the dependency scan runs first and its answer outranks the load |
| Readiness is settled *before* the first observed frame (the op runs inside `MainLoop::initialize`, before the tree has its root), so sampling it on a frame reported a splash scene that hands off in `_ready` as `not_ready` | the verdict is LATCHED from the instance's `ready` signal; the three doc strings stating the wrong mechanism (one published in `--schema`) are corrected |
| A project calling `get_tree().quit()` during startup left a clean exit with no sentinel → `contract_violation`, blaming gda for what the scene did | reported as `operation_failed` naming the cause; the payload's own exit defaults to failure, so a clean exit with no result can only be someone else's `quit(0)` |

Also fixed: "runs no project code" was false in four places (autoloads start for every `--project`
op, and `GDScript.reload()` executes static initializers — which validate was doing a second time
for every healthy script; the compile check now asks the already-loaded script first);
CONTEXT.md's `Project-code execution surface`, the single authority, now names the preflight as
its widest point; the frames/timeout relationship is documented on both options; the
cross-language enum constants are pinned on both sides; an unused Protocol is gone.

## Validation

- Gates: `pytest -m "not e2e"` 1695 passed, `ruff check`, `ruff format --check`, `pyright` strict
  — all clean.
- e2e (real engine, serialized under the shared lock): 17 new tests in
  `tests/test_e2e_scene_validate.py` + `tests/test_e2e_scene_preflight.py` + the op-robustness
  arms, run together with `tests/test_e2e_scene.py` — 57 passed. The runner refactor was also
  covered end to end by `test_e2e_script_run.py` + `test_e2e_node.py` + `test_e2e_info.py`
  (149 passed).
- Real-engine before/after captured by hand on a scratch project (good scene, missing script +
  missing texture, non-compiling attached script, never-imported `.png`, error-in-`_ready`,
  `assert(false)` in `_ready`, `_ready` that never returns, corrupt `.tscn`):
  - `scene validate` — `valid res://good.tscn`; two `missing_resource` problems with node
    attribution; `script_compile_failed` pointing at `gda script validate`; `unloadable_resource`
    for the unimported asset; `not_a_scene` (exit 4) for the corrupt file.
  - `scene preflight` — `ready`; `ready with errors` naming `res://err.gd:5` and the failed
    assertion; and the hang case returning `{"status":"timeout","started":false}` in **3.2s** for
    `--timeout 3`.
- TDD: the verdict shapes and the timeout verdict were written as failing tests first.

## Residual risks / deliberately out of scope

- **`push_error()` is not a recognized diagnostic.** A project that reports an application
  invariant with `push_error` (rather than an `assert` or a real runtime error) produces an
  engine `ERROR:` record whose only script attribution is its GDScript backtrace, which #651's
  closed recognition set does not cover — so such a scene reports `started: true`. Assertions and
  genuine runtime errors ARE caught (verified). Widening the parser changes what `script run`
  reports too, so it belongs to its own issue rather than this one. Draft follow-up: *"Recognize a
  project-raised `push_error` as a structured script diagnostic (#651 parser widening)"* — worth
  filing, since it is the most common way a Godot project reports exactly the invariant violation
  GDA-DF-030 describes. **Filed as #722** (needs-triage: the widening is a shared-parser
  contract decision) — reviewers, do not re-report.
- **`not_ready` has no demonstrated trigger.** With the readiness signal latched, every scene
  measured reports `ready`; the value is the honest else-branch of the latch (the propagation
  never reached the scene) rather than a state anyone has produced. Its *projection* is unit-tested
  from a canned payload, so the contract is pinned even though the engine path is not.
- **`--frames` and `--timeout` are not cross-checked.** A window large enough to outrun the
  ceiling reports the `timeout` verdict for a healthy scene. The relationship is documented on
  both options; a cross-check would need a per-frame cost estimate that depends on the machine and
  on what the scene does, and that false precision would be worse than the stated rule.
- **A `uid://`-spelled `[ext_resource]` path is untested.** Godot 4 writes both `uid=` and `path=`
  and gda reads `path=`, so this is not reachable through any scene gda or the editor writes
  (ADR-0036); a hand-written one would go through `ResourceLoader.exists`, which resolves uids.
- **The engine's stdout is not surfaced** by the preflight (gda's stdout carries only the result
  object, ADR-0002), so a scene's `print()` output is not visible; its stderr IS forwarded, like
  every other headless command's engine diagnostics. Pinned by an e2e test.
- **Node attribution is best-effort**, and one gap is named in the code: a multi-line property
  value whose continuation line starts with `[` closes the node block early, so later references
  in that node lose their `nodes` entry. The dependency itself is still found and reported.
- **Not built** (not asked for, and each is its own decision): a batch form for either command
  (`script validate`'s `--all` shape — worth a follow-up, since a gate typically checks every
  scene a change touched and each call is one engine launch), an outside-the-project refusal for
  `scene validate` (#658 built that for scripts as its own issue), recursive validation into
  instanced sub-scenes (each sub-scene validates on its own; **filed as #721** — reviewers,
  do not re-report), and any `--fail-on-invalid` /
  strict exit-code mode.


qinhaihong-red:
Orchestrator re-verification round complete at `ec3f3a8d` (DDMA + parallel Standards/Spec reviews + independent gates + a live hang-scene probe). All five #664 acceptance criteria verified, four empirically against real Godot. Eight findings were adopted and fixed in this round — recorded here so later reviewers do not re-derive them:

1. Sentinel-presence checking moved into `gda.parser.result_sentinel_start` — one home, used by `parse_result` itself; a BEGIN-without-END payload stays a parse failure (regression-tested against the tempting re-implementation).
2. The `frames >= 1` e2e comment now states the real reason (the cap is checked before the per-frame tick that records the verdict).
3. The engine side requires `frames` (`invalid_params`); the CLI/model owns the only default. Red-proofed.
4. `timeout` is excluded from the op's wire params (the bound is enforced by `launch()`).
5. The absent-project-root validator is extracted once as `gda.models.ProjectRootedResult`; `script validate` / `scene validate` inherit it — `--schema` output byte-identical before/after.
6. `ResourceLoader.load` runs only when no problems were found, so a known-broken scene pays neither the load nor its stderr cascade.
7. The unreachable `return` in `_ext_resource_ids_in_line` is gone (the loop exit is the return).
8. ADR-0002's `launch_timeout` row points at the 2026-08-19 scope note recording the preflight's timeout-as-verdict exception.

Deliberately NOT changed (assessed and declined): the `operations.gd` module-var grouping (GDScript-side restructuring costs more than the smell) and renaming `--frames` (both options document the real semantics). Known residual gaps are filed and linked in the PR body: composed-scene recursion → #721, `push_error` visibility → #722 — please do not re-report those two.

Gates at `ec3f3a8d`: fast suite 1696 passed, ruff/format/pyright clean, e2e under the serial lock 58 + 103 passed, PR CI green.

qinhaihong-red:
## Review conclusion: changes required

Reviewed at exact head `ec3f3a8d` against #664, the accepted gda ADRs, and the repository's Standards/Spec axes. The command-group/descriptor/runner structure is sound, but two independently reproduced P1 false-positive verdicts mean the new gates are not yet safe to trust.

### P1 — `scene validate` does not validate script bindings

`_scene_dependency_problem()` (`src/gda/ops/operations.gd:693-729`) checks external `.gd` resources for loadability/compilation, but it does not check whether a script's native base can bind the node that references it, and it does not cover embedded `GDScript` sub-resources.

Real Godot 4.6.3 probes on this head produced both false positives:

- a `Node2D` whose attached script `extends Resource` emitted `Script inherits from native type 'Resource', so it can't be assigned to an object of type 'Node2D'`, while `scene validate` returned `{"valid":true,"problems":[]}`;
- a scene with a syntax error in an embedded `GDScript` emitted a parse error, while `scene validate` again returned `valid: true`.

This misses #664's explicit “unloadable script bindings” requirement. The repository already owns the compatibility rule in `_op_script_attach()` (`operations.gd:1919-1939`); scene validation needs a static equivalent/shared authority and a result kind that can represent these failures. Please add real-engine regressions for incompatible external bindings and broken embedded scripts.

### P1 — `scene preflight` can report a failed binding as a clean start

`run_scene_preflight_operation()` derives `started` only from the ready status and `parse_script_errors()` (`src/gda/commands/scene.py:875,895`). The incompatible binding above is a deterministic engine `ERROR`, but it is outside #651's current recognized set. Preflight therefore returned:

```json
{"started":true,"status":"ready","diagnostics":[]}
```

The intended script never bound and its `_ready` could not run. This is distinct from the already-filed `push_error()` widening in #722; it is an undisclosed binding failure in the core #664 path. Keep one parser authority, but make this high-signal engine failure affect the structured verdict and pin it end to end.

### P2 — raw operation params silently truncate fractional `frames`

`_preflight_frames()` accepts any `float` and converts it with `int(raw)` (`operations.gd:876-888`). Direct sentinel input with `frames: 1.5` proceeded to the scene-path check instead of returning `invalid_params`; on a real scene it would silently observe one frame. The operation-side contract should accept only mathematically integral numbers.

### P2 — a dependency finding bypasses scene admission

When `_scene_dependency_problems()` finds anything, `_op_scene_validate()` skips the `PackedScene` load entirely (`operations.gd:618-627`). A file ending in `.tscn` that contained a missing `[ext_resource]` plus non-scene text was accepted as a scene verdict (`valid: false`) rather than refused as `not_a_scene`. Dependency diagnosis may outrank a load failure for a real scene, but the target still needs an independent scene-identity admission check.

### Contract and authority drift

- **P2:** `SceneProblem.path` and `project_root` documentation promise `res://` outcomes, but supported projectless filesystem scenes return absolute filesystem dependency paths. Either describe that contract honestly or canonicalize/restrict it.
- **P2:** `CONTEXT.md` says static validation runs “without running anything”, contradicting ADR-0009: a `--project` launch runs autoloads, and script compilation can run static initializers. Qualify this as not instantiating/running the target scene.
- **P3:** all four READMEs and `docs/command-catalog.md` say scene reading “always succeeds”, while this PR itself proves a broken dependency reached through a sub-resource can make the whole scene load fail. Narrow the claim to the measured cases.
- **P3:** `CONTEXT.md` still describes all sentinel channels as buffered and names only `script run` as streaming, although `scene preflight` now uses streaming capture through `_CaptureOnlyWatch`.

### Validation

- focused unit/command tests: `62 passed`;
- serialized real-engine scene/preflight/op e2e: `35 passed`;
- `git diff --check`: clean;
- manual exact-head probes reproduced every behavioral finding above;
- PR CI is green for Python/ruff/pyright, while the PR's Godot e2e job is skipped.

I intentionally did not re-report composed-scene recursion (#721) or `push_error()` visibility (#722).

qinhaihong-red:
All seven findings are addressed at `27c0dc30`. Every behavioral claim was first re-reproduced against a real Godot 4.6.3 before fixing; every fix is pinned by an e2e regression on the same fixture shape you probed.

| --- | --- | --- | --- |
| [P1] `scene validate` misses incompatible bindings and embedded scripts | **Adopted** | A binding scan over the loaded scene's `SceneState` (runs only when the dependency walk found nothing, so the load exists): an embedded `[sub_resource GDScript]` that does not compile is `script_compile_failed` named by its `::id` sub-resource path; a compiled script whose native base cannot bind its node is the new `incompatible_script` kind — the same rule `node script attach` enforces (`OP_ERROR_INCOMPATIBLE_SCRIPT_TYPE`'s vocabulary), asked statically via `get_instance_base_type()` × `ClassDB.is_parent_class`. A node without its own type (instanced/inherited child) is skipped honestly. | Reproduced first (`valid:true` on both probes at `ec3f3a8d`); after: `incompatible_script` with node attribution / `script_compile_failed` at `res://….tscn::GDScript_1`. e2e ×2 added. |
| [P1] preflight reports a failed binding as a clean start | **Adopted** | One parser authority kept: `gda.script_errors` now recognizes the engine's deterministic refusal sentence (`Script inherits from native type '…', so it can't be assigned …`) as `incompatible_script` — path-less by construction, so it can never name an entry script and never decides a `script run` verdict (guard-tested). `started = ready AND no diagnostics` then flips on its own. Distinct from #722: this is the engine's own fixed-shape refusal, not a project-authored `push_error`. | Reproduced first (`started:true, diagnostics:[]`); after: `started:false` + the `incompatible_script` diagnostic. Verbatim-stderr unit tests + an e2e; the full `script run` e2e suite passes unchanged. |
| [P2] fractional `frames` silently truncates | **Adopted** | The op refuses a non-integral number (`invalid_params: "frames must be a whole number: 1.5"`) before `int()` is asked. | Reproduced first (proceeded to the path check); direct-sentinel e2e added. |
| [P2] a dependency finding bypasses scene admission | **Adopted** | Scene-identity admission from the file's own `[gd_scene]` header, decided BEFORE any diagnosis — garbage text carrying an `[ext_resource]` line is `not_a_scene`, while a real scene with broken deps keeps its verdict (the sub-resource load-failure rationale is untouched). | Reproduced first (a scene verdict for garbage); after: `not_a_scene`. e2e added. |
| [P2] `path`/`project_root` promise res:// outcomes | **Adopted** | Both descriptions now state the contract honestly: projectless dependency paths are plain filesystem paths, and an embedded script's identity is the `::id` form. | Schema description text only; fast suite green. |
| [P2] CONTEXT.md "without running anything" vs ADR-0009 | **Adopted** | Qualified: static validation never instantiates or runs the TARGET scene; autoloads and script static initializers still run, as on every `--project` op (ADR-0009). | Doc-only. |
| [P3] "always succeeds" ×5 / stale capture-strategy prose | **Adopted** | All four READMEs + the catalog narrowed to the measured cases (a sub-resource-broken dependency can fail the whole load); CONTEXT.md's `Headless launch` term now names `scene preflight` as the second streaming-capture channel. i18n hashes re-stamped. | Doc-only; i18n sync test green. |

Gates at `27c0dc30`: fast suite 1698 passed, ruff / format / pyright clean; serialized real-engine e2e 106 passed (scene validate / preflight / op robustness / scene / script run). The `incompatible_script` additions to `ScriptErrorKind` and `SceneProblemKind` are additive enum values on already-published models; no envelope key changed.

qinhaihong-red:
## Re-review conclusion: two residuals remain

Rechecked the latest reply and exact head `27c0dc30`. The two former P1 false positives, fractional `frames`, projectless-path description, and the named README/CONTEXT/catalog drift are fixed and independently reproduced as closed. I am not repeating them below.

### P2 — scene-identity admission is still bypassable

`_op_scene_validate()` (`src/gda/ops/operations.gd:619-621`) treats any text beginning with the prefix `[gd_scene` as having a scene header:

```gdscript
if not text.lstrip(...).begins_with("[gd_scene"):
```

That is not an actual header check. On Godot 4.6.3, this non-scene `.tscn`:

```text
[gd_scenery]

[ext_resource type="Texture2D" path="res://missing.png" id="1_missing"]

this is not a Godot scene
```

still returned exit `0` with `{"valid":false,"problems":[{"kind":"missing_resource",...}]}` instead of refusing it as `not_a_scene`. The original no-header fixture is fixed, but the underlying admission property is only partially closed. Please recognize a complete valid `[gd_scene ...]` header rather than a string prefix, and pin an adjacent malformed-header regression.

### P3 — the published help/schema contract still describes the old behavior

- `gda scene validate --help` is generated from `src/gda/commands/scene.py:1111-1143` and still says scene loading succeeds “whatever is broken,” says only attached `.gd` files are compiled, and lists only the original three problem kinds. This contradicts both the measured sub-resource failure and the new embedded/incompatible-binding behavior.
- `gda scene validate --schema` still describes `SceneProblem` and `problems` as unresolved dependencies only (`scene.py:342-347, 415-419`), although an embedded or incompatible script is now a problem too.
- The published `ScriptError.kind` description (`src/gda/script_errors.py:238-245`) says every kind except `runtime_error` means a named resource could not load or run. The new path-less `incompatible_script` kind means the script compiled but could not bind, so that description is no longer true.

### Re-validation at `27c0dc30`

- focused unit/command tests: `64 passed`;
- serialized real-engine scene/preflight/op e2e: `40 passed`;
- `git diff --check`: clean;
- PR CI is green for Python/ruff/pyright; the PR's Godot e2e job remains skipped.

The behavioral gate should not be approved as fully closed until the scene-admission predicate is tightened. The contract-text residue is smaller but should be reconciled in the same fix because `--help` and `--schema` are authoritative agent-facing surfaces.

qinhaihong-red:
Both residuals are fixed at `9b096fcc`.

| Residual | Resolution | Validation |
| --- | --- | --- |
| [P2] admission was a string prefix — `[gd_scenery]` passed | `_has_scene_header` now requires the section NAME to be exactly `gd_scene`: after the tag comes the section's closing bracket or the whitespace before its attributes; a bare `[gd_scene` at EOF is not a header. The load keeps the final word on a header that lies. | Real-engine three-way probe: your `[gd_scenery]` fixture → `not_a_scene`; a bare `[gd_scene]` header still admits; the incompatible-binding scene still reports its verdict. Pinned by an adjacent malformed-header e2e regression on your exact fixture. |
| [P3] help/`--schema` prose described the old behavior | `scene validate --help` now states the current contract: loading survives *most* breakage, every **bound** script (referenced or embedded) is compiled, the base-compatibility check runs, all four problem kinds are listed (with the `::id` naming), and `not_a_scene`'s two triggers are named. `SceneProblem`'s docstring and the `problems` description cover binding problems and state the two problem classes' mutual exclusion (dependency problems suppress the binding scan, which needs the load). `ScriptErrorKind`'s class doc and the published `kind` description now except `incompatible_script` from "the named resource could not be loaded or run" — it loaded and compiled but could not be bound, and names no resource. | Doc/description-only; the `--schema` projections carry the new text via the models. |
Gates at `9b096fcc`: fast suite 1698 passed, ruff / format / pyright clean; serialized real-engine e2e 41 passed (scene validate incl. the new malformed-header regression, preflight, op robustness).

qinhaihong-red:
## Second re-review conclusion: both residuals remain partially open

Rechecked the latest reply and exact head `9b096fcc`. The `[gd_scenery]` fixture is now refused, and most of the source-level help/schema prose is corrected. Two directly related edges still prevent the residuals from being fully closed.

### P2 — an unclosed `[gd_scene ...` header still bypasses admission

`_has_scene_header()` (`src/gda/ops/operations.gd:664-672`) checks only that the character after `[gd_scene` is `]`, a space, or a tab. It does not require an attribute-bearing header to close with `]`.

On Godot 4.6.3, this non-scene file:

```text
[gd_scene load_steps=2 format=3

[ext_resource type="Texture2D" path="res://missing.png" id="1_missing"]

this is not a Godot scene
```

still returned exit `0` with `{"valid":false,"problems":[{"kind":"missing_resource",...}]}` rather than `not_a_scene`. The reply's “load keeps the final word on a header that lies” does not hold on this path: once dependency problems exist, `_op_scene_validate()` skips the load (`operations.gd:633-647`). Please require a complete, closed `[gd_scene ...]` header and add this adjacent malformed-header regression.

### P3 — the actual published help/schema output still has contract residue

- The `gda scene validate --help` source text is updated, but its unescaped `[gd_scene]`, `[ext_resource]`, and `[sub_resource]` tokens are consumed as Rich/Typer markup. The actual rendered help says `no complete  header` and similarly drops the dependency/sub-resource names. Please escape the markup and pin the rendered help, not only the docstring source.
- `SceneProblemKind` (`src/gda/commands/scene.py:313-321`) still projects into `--schema` as “what ... found wrong with one of a scene's dependencies”, although the enum now includes the non-dependency `incompatible_script` binding problem. `SceneProblem`, `problems`, and `ScriptError.kind` themselves are corrected; this is the remaining schema description.

### Re-validation at `9b096fcc`

- focused unit/command tests: `44 passed`;
- serialized real-engine scene/preflight/op e2e: `41 passed`;
- `git diff --check`: clean;
- PR CI is green for Python/ruff/pyright; the PR's Godot e2e job remains skipped.

The two fixes are moving in the right direction, but the current behavior and rendered contract do not yet support the reply's “Both residuals are fixed” conclusion.

qinhaihong-red:
Both edges are closed at `77c37f63`.

| Edge | Resolution | Validation |
| --- | --- | --- |
| [P2] an unclosed `[gd_scene …` header still bypassed admission | `_has_scene_header` now also requires the header LINE to close with `]` (the section-name check stays). You were right that "the load has the final word" does not hold on this path — with dependency problems present the load is skipped by design — so the function's comment now states that admission must be decided here, completely; the load's final word applies only to an admitted, closed header over a lying body. | Your exact fixture reproduced first (a `missing_resource` verdict at `9b096fcc`), then refused as `not_a_scene`; probed adjacent cases: a real attribute-bearing closed header still admits (and reports its dependency verdict), the bare `[gd_scene]` form still admits. The fixture is a pinned e2e regression. |
| [P3] rendered help dropped the bracketed tokens; `SceneProblemKind`'s schema doc still said "dependencies" | The three tokens are Rich style-tag shapes, so they are now markup-escaped (`\[gd_scene]` …) and the RENDERED help reads "no complete [gd_scene] header" / "an unresolvable [ext_resource] referenced from a [sub_resource]" — pinned by a test on the rendered output, not the docstring source (red-proofed: it failed against the unescaped text). `SceneProblemKind`'s class doc now names both problem classes the verdict reports. | Rendered-help pin test + live `--help` output check; fast suite 1699 passed, ruff / format / pyright clean; serialized real-engine e2e 42 passed (incl. the new unclosed-header regression). |

Issue #657, opened 2026-08-15
daemon: explicit engine-session readiness contract for the first live read


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-031 "Daemon startup can report success before live diagnostics are ready"** (2026-08-11, reconfirmed on a second package; the document's GDA-DF numbering collides across sessions — identified by title and date). After a successful `daemon start --windowed`, the first `diag errors` returned `engine_session_not_running`; an intervening live game read established the session and the unchanged diagnostic then succeeded.

Correction against the current implementation (this reframes the dogfooding entry): `daemon start` returns the **daemon** PID and socket and launches no `Engine session` — sessions launch lazily on the first live op, a recorded decision (ADR-0017 outcome). `diag errors` deliberately does not launch a session (a read-only diagnostic must not run project code, and a relaunch would truncate the Session log — ADR-0022), and its failure already carries a typed code with remediation text. There is no race and no defect; the gap is **start-time ergonomics**: an unattended workflow has no documented way to reach "the session serves" other than issuing an unrelated warm-up live read.

## What to build

Give agents an explicit, bounded way to establish and await session readiness, leaving the lazy-launch default intact:

- Preferred: a bounded wait command (e.g. `daemon wait-ready --timeout`) or an explicit session-launch operation whose success means subsequent live reads serve.
- Alternative (bigger decision): launch the session eagerly at `daemon start`. This amends ADR-0017's lazy-launch outcome and runs project code at start time (interacts with ADR-0022 and the ADR-0009 reasoning above); it must not be chosen implicitly.
- Document the lazy-launch lifecycle in help and the skill so `engine_session_not_running` on a first `diag errors` is expected, not alarming.

## Acceptance criteria

- [ ] An agent can deterministically reach a state where `diag errors` serves, using a documented bounded wait command — no unrelated warm-up read, no sleep loops.
- [ ] The skill and command help state the lazy-launch lifecycle and the readiness contract.
- [ ] A regression test starts a daemon and reaches a served `diag errors` through the documented path as the first live call.

## Blocked by

The lazy-vs-eager decision above if the eager option is considered (ADR-0017 amendment). The bounded wait command needs no ADR change.

qinhaihong-red:
> *This was generated by AI during triage.*

[...]

PR #723, opened 2026-08-20
refactor(gda): daemon Tier 2 — DaemonPaths session log, session seam, socket-lifecycle tests (#674)

### LINKED ISSUE #674: daemon Tier 2: Session-log path into DaemonPaths; server↔session seam with socket-lifecycle tests
## Provenance

Tier 2 of the daemon-module assessment (adversarial review of the daemon modules, adjudicated in-session on 2026-08-15; Tier 1 — the cross-language contract guards — merged as PR #650). Deliberately parked "behind the next live-stack change wave"; that wave is now tracked in this milestone as #656 (pause-proof harness) and #657 (engine-session readiness contract), so this issue rides with them. Not a dogfooding finding itself — it is in this milestone because the milestone owns the wave that unparks it.

## What to build

Two behavior-preserving deliverables on the server↔session boundary:

1. **Session-log path becomes part of `DaemonPaths`.** Today the daemon server re-derives the Session-log path by string-splitting the CLI socket's filename to recover the project slug — a parallel derivation outside `DaemonPaths`, the declared single home of the per-project on-disk identity (ADR-0021). Add the Session-log path as a `DaemonPaths` field derived in the same place as the sockets and pidfile; the server consumes it instead of re-deriving.
2. **Server↔session seam + socket-lifecycle tests.** The daemon server constructs and manages the engine session directly, so its accept/dispatch/cleanup logic is only exercised end-to-end. Introduce a seam that lets unit tests drive the server against a fake session (launch success, launch failure, session death mid-request), and add socket-lifecycle tests: bind, `sun_path`-limit refusal, stale-socket cleanup, double-start, and cleanup on stop/signal. Tests binding a real UDS must use the short-runtime-dir fixture (`sun_path` overflows under pytest's long tmp paths).

Both are refactor + test work: no behavior change, no schema change.

## Acceptance criteria

- [ ] `DaemonPaths` carries the Session-log path; no caller re-derives it from socket filenames; existing e2e behavior is unchanged.
- [ ] Unit tests cover the server's session lifecycle against a fake session: lazy launch on first live op, launch failure surfaced as the typed error, session death mid-request.
- [ ] Socket-lifecycle unit tests cover bind, UDS length refusal, stale-socket cleanup, double-start, and stop/signal cleanup.
- [ ] The fast suite runs these without a Godot binary or a real engine session.

## Blocked by

None technically — scheduled to land in the same wave as #656/#657 per the daemon assessment, so the restructuring and those behavior changes are reviewed and e2e-verified together.

qinhaihong-red:
## What

W4 slot 1, first half (the #674 preparatory refactor; #657 wait-ready stacks on the final server shape after this merges). Two behavior-preserving deliverables on the daemon's server↔session boundary, per the issue brief:

1. **Session-log path into `DaemonPaths`.** The path is now a field of the per-project on-disk identity (ADR-0021), derived in `daemon_paths()` beside the sockets and pidfile. `DaemonServer` consumes `paths.session_log`; the string-splitting re-derivation (`_session_log_path`) is gone. No caller derives it from a socket filename anymore.
2. **Server↔session seam + socket-lifecycle tests.** `DaemonServer` takes a `SessionLaunch` (a Protocol matching `launch_session`, which stays the default), so unit tests drive the **real serve loop** — real UDS bind, accept, reply, cleanup — against a fake session. New `tests/test_daemon_socket_lifecycle.py` covers, in the fast suite with no Godot binary:
   - bind + status over the socket; stale-slot reclaim (abandoned socket files + unlocked pidfile); double-start; stop-op cleanup; termination-signal cleanup; the UDS `sun_path` length refusal at `daemon start` (typed `daemon_not_running`, previously untested);
   - the session lifecycle through the seam: lazy launch on the first live op + reuse; launch failure as typed `engine_session_not_running`; **death mid-request** (a real `EngineSession` over a broken socketpair → `engine_disconnected`) followed by relaunch on the next op; and every launch receiving the one `DaemonPaths`-derived log path.

## One found defect, fixed (disclosed deviation from "no behavior change")

The mandated double-start test proved a real destructive race at head: `serve()` bound its sockets **before** taking the pidfile lock, so a LOSING second start unlinked the winner's live socket files during `_bind`, then failed at the lock, and its cleanup unlinked the slot again — turning one live daemon into zero reachable ones (new client connects hit an unlinked path and read "not running"). `serve()` now acquires the pidfile lock — the mutual exclusion AND liveness primitive — **first**, so a loser fails before touching the winner's slot. Red-proofed: the test fails on the pre-fix ordering with exactly that destruction, and passes on the fix. Liveness as the CLI reads it (lock held AND socket bound AND project matches) is unchanged; `daemon start`'s readiness poll (`daemon_pid`) is unaffected.

## Validation

- Fast suite: **1713 passed**; pyright 0 errors; ruff check/format clean.
- Behavior-preserving claim: `gda schema` + all-`--help` dump is **byte-identical** to the base (dev tip `16cc7bb8`).
- Real-engine daemon e2e (`tests/test_e2e_daemon.py`, Godot 4.6.3, serial under the shared lock): **27 passed**.
- The new lifecycle suite runs real UDS binds via the short-runtime-dir fixture (`sun_path` limit) and needs no engine.

Closes #674

### DIFF
--- a/src/gda/daemon/discovery.py
+++ b/src/gda/daemon/discovery.py
@@ -46,6 +46,14 @@ class DaemonPaths:
     cli_socket: Path  # CLI <-> daemon UDS
     harness_socket: Path  # daemon <-> harness UDS (path injected into the session)
     pidfile: Path  # liveness + the recorded canonical project path
+    # The daemon-owned Session log (#224): the engine session is launched with
+    # `--log-file` on this path so the daemon can read the running game's
+    # errors/output back to serve `gda diag` / `gda logger`. Under the private
+    # runtime dir (NOT `user://logs` — that shared path caused #180), keyed by the
+    # same slug as the sockets/pidfile; `RotatedFileLogger` truncates it each
+    # launch, making it session-bound (ADR-0020). Derived here with the rest of
+    # the identity (#674) — no consumer re-derives it from a socket filename.
+    session_log: Path
 
 
 def _runtime_dir(env: Mapping[str, str]) -> Path:
@@ -85,6 +93,7 @@ def daemon_paths(project: Path, env: Mapping[str, str] | None = None) -> DaemonP
         cli_socket=runtime / f"{slug}.cli.sock",
         harness_socket=runtime / f"{slug}.harness.sock",
         pidfile=runtime / f"{slug}.pid",
+        session_log=runtime / f"{slug}.session.log",
     )
 
 
--- a/src/gda/daemon/discovery.py
+++ b/src/gda/daemon/discovery.py
@@ -117,16 +117,25 @@ def acquire_pidfile(paths: DaemonPaths, pid: int):
     ``os.kill`` PID-liveness (which a reused PID could spoof). The OS releases the
     lock when the daemon exits or crashes, so liveness self-heals with no cleanup.
     Raises ``OSError`` if another live daemon already holds it (a start race).
+
+    Opened WITHOUT truncation, and truncated only AFTER the lock is won (#723
+    review): ``open("w")`` zeroed the file before ``flock`` could refuse, so a
+    LOSING start erased the live winner's recorded identity — ``daemon_pid``
+    then read the running daemon as not running.
     """
     import fcntl
 
     ensure_runtime_dir(paths)
-    handle = open(paths.pidfile, "w", encoding="utf-8")
+    handle = os.fdopen(
+        os.open(paths.pidfile, os.O_RDWR | os.O_CREAT, 0o644), "r+", encoding="utf-8"
+    )
     try:
         fcntl.flock(handle.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
     except OSError:
         handle.close()
         raise
+    handle.seek(0)
+    handle.truncate()
     handle.write(f"{pid}\n{paths.project}\n")
     handle.flush()
     return handle
--- a/src/gda/daemon/server.py
+++ b/src/gda/daemon/server.py
@@ -20,6 +20,7 @@ from gda.daemon.protocol import error_reply, read_message, result_reply, write_m
 from gda.daemon.session import (
     EngineSession,
     SceneMismatch,
+    SessionLaunch,
     WindowedDisplayUnavailable,
     launch_session,
 )
@@ -49,6 +50,7 @@ class DaemonServer:
         windowed: bool = False,
         scene: str | None = None,
         display_check: Optional[Callable[[], Optional[WindowedUnavailable]]] = None,
+        launch: Optional[SessionLaunch] = None,
     ) -> None:
         self.paths = paths
         self.godot = godot
@@ -67,6 +69,11 @@ class DaemonServer:
         # so tests drive the guard without a real display; defaults to the shared
         # gda.display probe. Consulted only for a windowed session.
         self._display_check = display_check or windowed_unavailable
+        # The server↔session seam (#674): how an engine session is launched.
+        # Injectable so unit tests drive the whole serve loop against a fake
+        # session — no Godot binary, no real engine. Resolved at construction
+        # against the module global, so the default stays the real launch.
+        self._launch: SessionLaunch = launch or launch_session
         self._token = secrets.token_hex(16)
         self._stopping = False
         self._listener: socket.socket | None = None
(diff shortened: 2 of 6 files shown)


qinhaihong-red:
Adversarial review pinned to `base=16cc7bb8`, `head=21cad634`.

Conclusion: this is not ready to merge. I kept the DDMA, Standards, and Spec axes separate below; the repeated double-start finding is intentional because it fails all three axes for different reasons.

## Design Domain Modular Architecture

### [P1] The pidfile lock does not own the daemon slot's complete critical section

`src/gda/daemon/server.py:93` now calls `acquire_pidfile()` before binding, but `src/gda/daemon/discovery.py:124-126` opens the pidfile with `"w"` before attempting `flock`. A losing second start therefore truncates the winner's liveness record before it fails to acquire the lock.

The reverse lifecycle has the same ownership gap: `src/gda/daemon/server.py:343-360` releases the pidfile lock before closing and unlinking the sockets/pidfile. A successor can acquire the lock and bind in that window, after which the predecessor unlinks the successor's complete slot.

Deterministic replays on this head:

- Double-start: `daemon_pid=111` -> loser raises `BlockingIOError` -> pidfile becomes `''` -> `daemon_pid=None`.
- Stop/start handoff: successor reaches `daemon_pid=222` with all three paths present -> predecessor cleanup completes -> all three paths are absent and `daemon_pid=None`.

The runtime impact is that a daemon can still hold live listener fds while status, live attach, and stop report it as not running; the running-daemon uninstall guard can also be bypassed.

Minimum correction: open the pidfile without truncating, acquire the lock, then truncate/write; during cleanup, remove the owned slot and release the lock last. Extend the tests to assert the winner's pidfile contents and `daemon_pid(paths)`, plus a stop/start handoff regression.

### [P2] The server-session seam is still typed as the concrete adapter

`SessionLaunch` is defined in `src/gda/daemon/session.py:143-165` and returns concrete `EngineSession`; `DaemonServer` also stores and returns that concrete type. The new fake sessions consequently require `cast(EngineSession, ...)` at `tests/test_daemon_socket_lifecycle.py:262` and `:313`, bypassing the type checker at the exact seam this refactor introduces.

The server side should own a structural `SessionHandle` contract for the capabilities it consumes (`log_file`, `alive`, `request`, `close`), and `SessionLaunch` should return that contract. `EngineSession` can remain the concrete implementation in `session.py`.

The `DaemonPaths.session_log` move itself is sound: `daemon_paths()` is now the single path-derivation authority, and I found no remaining consumer that reconstructs the log path from a socket filename.

## Standards

1. **[P1, hard breach]** The double-start truncation violates ADR-0021's socket + pidfile + held-lock liveness contract. The new test at `tests/test_daemon_socket_lifecycle.py:154-174` only connects directly to the winner's socket and never re-checks its pidfile or `daemon_pid(paths)`.
2. **[P2, hard breach]** Releasing the mutual-exclusion lock before deleting the slot permits the predecessor/successor cleanup race above, also violating ADR-0021.

Together these are a judgement-call **Shotgun Surgery** smell: one slot-atomicity rule is distributed across acquire, bind, and cleanup, and the ordering change updated only part of that rule.

## Spec

### [P1] Issue #674's double-start acceptance criterion is not met

The issue requires behavior-preserving work and explicit double-start coverage. The patch preserves the winner's socket files but destroys its discovery identity, so the acceptance criterion is only partially implemented and the test gives a false positive.

No other open Spec finding remains. The latest real-SIGTERM child-process test closes the earlier Linux test-model failure and now exercises the production signal path correctly.

Validation on exact head `21cad634`:

- focused daemon tests: `31 passed`;
- real Godot daemon e2e outside the sandbox: `27 passed`;
- current fast CI: green;
- `git diff --check`: clean.

qinhaihong-red:
## Review response — all findings fixed at head `c3a9f181`

| --- | --- | --- |
| [P1] The pidfile lock does not own the slot's complete critical section (pre-lock truncation; lock released before the unlinks) | **Fully adopt** | Both halves verified by writing the tests you asked for first — both red on `21cad634` — then fixed. (1) `acquire_pidfile` opens `O_RDWR\|O_CREAT`, takes the `flock`, and only then truncates and rewrites: a losing start can no longer erase the winner's recorded identity. (2) `_cleanup` releases the lock **last**, after the socket/pidfile unlinks: while it is held a successor's acquire fails, so no successor slot can be bound inside the window and then destroyed. |
| [P1 Standards/Spec] The double-start test asserted too little (sockets only) | **Fully adopt** | The test now also asserts the winner's `read_pidfile` contents `(pid, project)` and `daemon_pid(paths) == winner` after the loser fails — the full ADR-0021 socket + pidfile + held-lock contract. Red on the old head via exactly your replay (pidfile becomes `''`, `daemon_pid=None`). |
| [P1] Stop/start handoff race | **Fully adopt** | New deterministic test: `os.unlink` is instrumented to record, for each of the three slot paths, whether the pidfile lock was still held at that moment — all three must be `True` (red on the old head: all `False`, since the lock released first). The same test then starts a successor after the stop and asserts `daemon_pid` reports it with a served `__status__`. |
| [P2] The seam is typed as the concrete adapter | **Fully adopt** | The server now owns a structural `SessionHandle` Protocol (`log_file`, `alive`, `request`, `close`); `SessionLaunch` moved to `server.py` (the contract's owner) and returns `Optional[SessionHandle]`. `EngineSession` remains the concrete implementation in `session.py`; both test fakes conform by shape and the two `cast(EngineSession, …)` at the seam are gone (the remaining casts are the orthogonal fake-`Popen` ones). |

Validation at `c3a9f181`: fast suite **1714 passed**; pyright **0 errors** including the lifecycle test file (no seam casts); ruff check/format clean; `gda schema` + all-`--help` still **byte-identical** to base `16cc7bb8`; real-Godot daemon e2e **27 passed** (serial under the shared lock). PR CI is running on the new head.

qinhaihong-red:
Closed as a side effect of renaming the head branch to the conventional form: GitHub treated the rename as `head_ref_deleted` and dropped the PR (reopen is refused once the recorded head name is gone). Superseded by **#724** from `refactor/674-daemon-session-seam` with the **identical head `c3a9f181`** — the adversarial review and the point-by-point response in this thread apply to it verbatim.

PR #725, opened 2026-08-20
feat(gda): daemon wait-ready — a bounded, explicit engine-session readiness wait (#657)

### LINKED ISSUE #657: daemon: explicit engine-session readiness contract for the first live read

qinhaihong-red:
## What

W4 slot 1, second half (stacked on #674's merged server shape, per the wave plan). `gda daemon wait-ready`: the documented, bounded way to establish the lazily-launched engine session, per issue #657 (GDA-DF-031 reframed — no race, no defect; a start-time ergonomics gap).

- **Contract**: a session launches lazily on the **first operation that REQUIRES one** — the precise form of ADR-0017's lazy launch, which this slice's dated amendment restates, since `diag errors` / `logger tail` are live operations the daemon serves from the Session log and they deliberately launch nothing (ADR-0022). `wait-ready` is the explicit way to *be* that operation: success (`{pid, launched}`) means the in-game harness has connected **and** presented its token, so subsequent live reads serve — including a first `diag errors`. Idempotent while a session is alive (`launched: false`, nothing relaunched), and `launched` is reported by the launch owner rather than sampled around it, so a call that launched can never report `false`.
- **Shape**: a `kind = LIVE` command in the `daemon` group (`diag errors` precedent: daemon-served, harness-never-relayed; routed and classified through the live channel — `DaemonRunner`, `classify_live`, the schema/constraints self-description all come free). The daemon answers a new `daemon-wait-ready` op through the **same launch boundary every session-needing op uses** — `_session_or_refusal`, extracted from the live-op branch and returning a typed `_Established(session, launched)` — so a failed launch, scene mismatch, or windowed refusal is byte-identically the same typed error on both paths.
- **Bound**: `--timeout` ((0, 50], model-owned per ADR-0015, argv via `params_or_bad_parameter`, **re-enforced at the daemon IPC boundary** as a typed `invalid_params`) rides to the daemon, and the launch boundary turns it into **one absolute instant** covering everything it does on the caller's clock: retiring the session being replaced, the spawn, the harness connect, each handshake frame, and the teardown of a failed launch. It is a budget for **waiting and for committing to new work**, not a hard wall clock — nothing here can interrupt a synchronous call already in flight, and gda is not adding threads and cancellation to pretend otherwise. Three rules instead: no phase gets a fresh grace; every timed wait uses what remains, computed where the wait begins; and once the instant is spent the boundary commits to nothing further. Three edges it has to survive, because the daemon serves one request at a time: a socket timeout bounds INACTIVITY, so a frame read in chunks recomputes its budget per read; an engine that ignores `SIGTERM` is escalated with what the deadline has left; and collecting a killed child is a background duty (best-effort), never a wait. What teardown retires is the process GROUP the session owns — captured at spawn, since a reaped leader can no longer name it — and it waits for that group to empty rather than for the leader to exit. The live channel's own 60s ceiling now covers the whole round trip too, on the same absolute-instant rule.
- **Serving state**: a session serves only while its **process and its harness channel** can answer. The channel is latched stale on a dropped or closed connection **and on a relay that timed out** — the one-op-at-a-time RPC carries no request id and a timed-out frame is not drained, so a late reply would be read as the *next* operation's reply — and the next session-needing op rebuilds through the same launch boundary. `live_timeout` therefore costs a relaunch, and runtime state does not survive it (ADR-0020). The relay read is bounded the same way as the handshake frames, so `did not return within 30s` is true of a trickled reply too.
- **Guard**: `DAEMON_SERVED_OPS` (= LOG_OPS + wait-ready) is the single **runtime** authority for daemon-answered ops: `_handle` routes on it, the PR #650 cross-language op-table guard reads it, and a declared member with no handler refuses loudly instead of silently relaying. `WAIT_READY_TIMEOUT_MAX` is the one bound constant, consumed by both the params model (`le=`) and the wire check.
- **Docs** (skill owner this wave): ADR-0017 gains the dated amendment above (launch trigger, serving state, and the single deadline the launch boundary owns); ADR-0040 records the `daemon` group's one `kind = LIVE`, recipe-less command. Command help, SKILL.md (prereq paragraph + daemon row), README daemon table + all three translations, and the command catalog all use the precise launch rule and state that a first `diag errors` right after `start` reporting `engine_session_not_running` is expected.

## Acceptance criteria → evidence

1. *Deterministic served state, no warm-up read, no sleep loops* — e2e `test_wait_ready_makes_a_first_diag_errors_serve`: real `daemon start` → `daemon wait-ready` (`launched: true`) → `diag errors` **as the first read** serves → repeat wait-ready (`launched: false`) → stop. Real Godot 4.6.3.
2. *Skill and help state the lifecycle and readiness contract* — see the docs list above; the SKILL surface-sync guard pins the daemon row.
3. *Regression test through the documented path as the first live call* — the e2e above, plus socket-level fast tests driving the REAL serve loop over real UDS: launch-once + idempotent repeat + caller bound passed through; failed launch relays the typed refusal.

Plus the review-driven regressions, each red-proofed against the head it was found on:

- a silent handshake peer cannot hold the launch past the deadline (at the launcher, and end to end with the next control request still served);
- a session whose harness channel broke, and a session whose relay timed out, are both rebuilt by the next `wait-ready` instead of certified;
- `launched == (a launch actually happened on this call)` under a liveness that flips between samples;
- every declared `DAEMON_SERVED_OPS` member is intercepted, never relayed;
- the wire boundary refuses negative / zero / over-limit / `inf` / `nan` / boolean / string timeouts;
- a SIGTERM-ignoring engine child cannot hold the round trip, or the request enqueued *while it runs*, past the caller's bound — whether it is the child of the launch being bounded or of the stale session being retired;
- a spawn that outruns the whole budget yields a refusal, not a session, even with both handshake frames already queued;
- a peer that trickles either handshake frame one byte at a time cannot outlast the bound, and neither can a trickled op reply;
- teardown does not wait for the kernel to collect the child, and on the ordinary path the child is collected;
- the termination wait is measured when it begins, not when it was decided;
- what teardown retires is the process GROUP the engine owns — proved for a leader that ignores `SIGTERM` and for one that obeys it while a descendant does not;
- launch preparation that crosses the deadline spawns nothing, and a deadline spent by the spawn itself is reported as that, never as a missing auth token;
- the owner still knows the group after a liveness check has reaped the leader, and a descendant may finish its own `SIGTERM` cleanup inside the remaining deadline;
- a trickled reply cannot hold the CLI past `LIVE_REQUEST_TIMEOUT`.

## Validation

- Fast suite: **1748 passed**; pyright 0 errors; ruff clean (`format` included); `git diff --check` clean; i18n stamps unchanged.
- Red-proofs, each against the head its finding was raised on — `a35b49a4` (5, one of which *hangs* there rather than failing), `59671962` (3), `b8ac5706` (5), `cdb3c900` (3), `29b22d34` (4), `b196c829` (1), and `dc04d62b` (3: a descendant orphaned once the leader was reaped; a descendant's cleanup cut short; a trickled reply holding the CLI past its ceiling).
- Two assertions are deliberately not red on the head that introduced their subject, and say so rather than hiding in a count: "a killed child is still collected" (the reap already existed on `cdb3c900`) and the ordinary-path collection assertion — the daemon-thread reaper is unjoined and swallows failures, so no test can prove collection in general.
- Real-engine e2e, serial under the shared lock: **59 passed** across the daemon and live suites (`daemon`, `game`, `diag`, `logger`, `perf`, `input`, `harness install`, `mcp live`, `screen`).
- Surface: one intended `gda schema` change across the whole slice beyond the new command itself — `daemon wait-ready`'s timeout description now states what the budget covers instead of implying a hard wall clock. Every other command's schema entry is byte-identical.

Closes #657

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,8 +534,8 @@ reported as `script_aborted` with the captured error.
 
 | Command | What it does |
 | ------- | ------------ |
-| `daemon start` | Start the per-project daemon and install the in-game harness; the engine session launches on the first live op (`--windowed` for `screen` capture). |
-| `daemon wait-ready` | Establish the lazily-launched engine session and wait, bounded, until live reads serve — the documented way to be the first live op, so a first `diag errors` right after `daemon start` does not report `engine_session_not_running`. |
+| `daemon start` | Start the per-project daemon and install the in-game harness; the engine session launches on the first operation that requires one (`--windowed` for `screen` capture). |
+| `daemon wait-ready` | Establish the lazily-launched engine session and wait, bounded, until live reads serve — the documented way to trigger the session launch, since the read-only `diag`/`logger` reads never launch one and a first `diag errors` right after `daemon start` reports `engine_session_not_running` by design. |
 | `daemon stop` | Stop the project's daemon and any running engine session. |
 | `daemon status` | Report the daemon's state (running, windowed mode, session). |
 | `daemon install` | Install the in-game `gda` harness into the project without starting a daemon, reporting every path and section it created. Idempotent, and re-syncs a harness from an older `gda`. `daemon start` does this itself, so run it only to make that `project.godot` change deliberately — to review or commit it. |
--- a/README.md
+++ b/README.md
@@ -535,7 +535,7 @@ reported as `script_aborted` with the captured error.
 | Command | What it does |
 | ------- | ------------ |
 | `daemon start` | Start the per-project daemon and install the in-game harness; the engine session launches on the first operation that requires one (`--windowed` for `screen` capture). |
-| `daemon wait-ready` | Establish the lazily-launched engine session and wait, bounded, until live reads serve — the documented way to trigger the session launch, since the read-only `diag`/`logger` reads never launch one and a first `diag errors` right after `daemon start` reports `engine_session_not_running` by design. |
+| `daemon wait-ready` | Establish the lazily-launched engine session so live reads serve. Its `--timeout` is the daemon-side budget shared by launch waits and new-work decisions; a synchronous launch call can delay when expiry is observed. This is the documented way to trigger the session launch, since the read-only `diag`/`logger` reads never launch one and a first `diag errors` right after `daemon start` reports `engine_session_not_running` by design. |
 | `daemon stop` | Stop the project's daemon and any running engine session. |
 | `daemon status` | Report the daemon's state (running, windowed mode, session). |
 | `daemon install` | Install the in-game `gda` harness into the project without starting a daemon, reporting every path and section it created. Idempotent, and re-syncs a harness from an older `gda`. `daemon start` does this itself, so run it only to make that `project.godot` change deliberately — to review or commit it. |
--- a/README.md
+++ b/README.md
@@ -535,6 +535,7 @@ reported as `script_aborted` with the captured error.
 | Command | What it does |
 | ------- | ------------ |
 | `daemon start` | Start the per-project daemon and install the in-game harness; the engine session launches on the first live op (`--windowed` for `screen` capture). |
+| `daemon wait-ready` | Establish the lazily-launched engine session and wait, bounded, until live reads serve — the documented way to be the first live op, so a first `diag errors` right after `daemon start` does not report `engine_session_not_running`. |
 | `daemon stop` | Stop the project's daemon and any running engine session. |
 | `daemon status` | Report the daemon's state (running, windowed mode, session). |
 | `daemon install` | Install the in-game `gda` harness into the project without starting a daemon, reporting every path and section it created. Idempotent, and re-syncs a harness from an older `gda`. `daemon start` does this itself, so run it only to make that `project.godot` change deliberately — to review or commit it. |
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=ce484830762f3beaff4c979423c5ae400f9e443d077393808ea1b566fefce7ff -->
+<!-- gda-readme-i18n: source=README.md sha256=8cb5ae07a2d875db3554b7840b6fc90efdca5fa1bd84da220a2de393f6871949 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -549,7 +549,7 @@ silencio en ambos flujos se termina en segundos en lugar de esperar al límite, 
 | Comando | Qué hace |
 | ------- | ------------ |
 | `daemon start` | Arranca el daemon por proyecto e instala el harness dentro del juego; la sesión del motor se lanza en la primera operación que la requiere (`--windowed` para la captura de `screen`). |
-| `daemon wait-ready` | Establece la sesión del motor de lanzamiento perezoso y espera, con límite, hasta que las lecturas live sirvan — la forma documentada de disparar ese lanzamiento explícitamente: las lecturas de solo lectura `diag`/`logger` nunca lanzan una sesión, así que un primer `diag errors` justo después de `daemon start` informa `engine_session_not_running` por diseño. |
+| `daemon wait-ready` | Establece la sesión del motor de lanzamiento perezoso para que las lecturas live respondan. `--timeout` es el presupuesto del daemon compartido por las esperas de lanzamiento y las decisiones de iniciar trabajo nuevo; una llamada de lanzamiento síncrona en curso puede retrasar la detección del vencimiento. Es la forma documentada de disparar ese lanzamiento explícitamente: las lecturas de solo lectura `diag`/`logger` nunca lanzan una sesión, así que un primer `diag errors` justo después de `daemon start` informa `engine_session_not_running` por diseño. |
 | `daemon stop` | Detiene el daemon del proyecto y cualquier sesión del motor en ejecución. |
 | `daemon status` | Informa el estado del daemon (en ejecución, modo con ventana, sesión). |
 | `daemon install` | Instala el harness `gda` dentro del proyecto sin iniciar un daemon, informando cada ruta y sección que creó. Idempotente, y resincroniza un harness de un `gda` anterior. `daemon start` ya lo hace por su cuenta, así que ejecútalo solo para realizar ese cambio en `project.godot` de forma deliberada — para revisarlo o confirmarlo. |
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=6c49dbe1f8202e9cfa46bc8fd79e1c78060238936b0d52e56ec01fd4f2f53e70 -->
+<!-- gda-readme-i18n: source=README.md sha256=3c57da598896485491b50b52cd19a411aecd42efff28a1242fe06c190f8e4462 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -549,6 +549,7 @@ silencio en ambos flujos se termina en segundos en lugar de esperar al límite, 
 | Comando | Qué hace |
 | ------- | ------------ |
 | `daemon start` | Arranca el daemon por proyecto e instala el harness dentro del juego; la sesión del motor se lanza en la primera operación live (`--windowed` para la captura de `screen`). |
+| `daemon wait-ready` | Establece la sesión del motor de lanzamiento perezoso y espera, con límite, hasta que las lecturas live sirvan — la forma documentada de ser la primera operación live, para que un primer `diag errors` justo después de `daemon start` no informe `engine_session_not_running`. |
 | `daemon stop` | Detiene el daemon del proyecto y cualquier sesión del motor en ejecución. |
 | `daemon status` | Informa el estado del daemon (en ejecución, modo con ventana, sesión). |
 | `daemon install` | Instala el harness `gda` dentro del proyecto sin iniciar un daemon, informando cada ruta y sección que creó. Idempotente, y resincroniza un harness de un `gda` anterior. `daemon start` ya lo hace por su cuenta, así que ejecútalo solo para realizar ese cambio en `project.godot` de forma deliberada — para revisarlo o confirmarlo. |
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=3c57da598896485491b50b52cd19a411aecd42efff28a1242fe06c190f8e4462 -->
+<!-- gda-readme-i18n: source=README.md sha256=ce484830762f3beaff4c979423c5ae400f9e443d077393808ea1b566fefce7ff -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -548,8 +548,8 @@ silencio en ambos flujos se termina en segundos en lugar de esperar al límite, 
 
 | Comando | Qué hace |
 | ------- | ------------ |
-| `daemon start` | Arranca el daemon por proyecto e instala el harness dentro del juego; la sesión del motor se lanza en la primera operación live (`--windowed` para la captura de `screen`). |
-| `daemon wait-ready` | Establece la sesión del motor de lanzamiento perezoso y espera, con límite, hasta que las lecturas live sirvan — la forma documentada de ser la primera operación live, para que un primer `diag errors` justo después de `daemon start` no informe `engine_session_not_running`. |
+| `daemon start` | Arranca el daemon por proyecto e instala el harness dentro del juego; la sesión del motor se lanza en la primera operación que la requiere (`--windowed` para la captura de `screen`). |
+| `daemon wait-ready` | Establece la sesión del motor de lanzamiento perezoso y espera, con límite, hasta que las lecturas live sirvan — la forma documentada de disparar ese lanzamiento explícitamente: las lecturas de solo lectura `diag`/`logger` nunca lanzan una sesión, así que un primer `diag errors` justo después de `daemon start` informa `engine_session_not_running` por diseño. |
 | `daemon stop` | Detiene el daemon del proyecto y cualquier sesión del motor en ejecución. |
 | `daemon status` | Informa el estado del daemon (en ejecución, modo con ventana, sesión). |
 | `daemon install` | Instala el harness `gda` dentro del proyecto sin iniciar un daemon, informando cada ruta y sección que creó. Idempotente, y resincroniza un harness de un `gda` anterior. `daemon start` ya lo hace por su cuenta, así que ejecútalo solo para realizar ese cambio en `project.godot` de forma deliberada — para revisarlo o confirmarlo. |
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=6c49dbe1f8202e9cfa46bc8fd79e1c78060238936b0d52e56ec01fd4f2f53e70 -->
+<!-- gda-readme-i18n: source=README.md sha256=3c57da598896485491b50b52cd19a411aecd42efff28a1242fe06c190f8e4462 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -545,6 +545,7 @@ offset プロパティを明示的に設定してください。
 | コマンド | 機能 |
 | ------- | ------------ |
 | `daemon start` | プロジェクトごとのデーモンを起動し、ゲーム内ハーネスをインストールします。エンジンセッションは最初の Live 操作で起動します(`screen` キャプチャには `--windowed`)。 |
+| `daemon wait-ready` | 遅延起動されるエンジンセッションを確立し、Live 読み取りが応答するまで上限つきで待ちます — 「最初の Live 操作」になるための正規の方法で、`daemon start` 直後の最初の `diag errors` が `engine_session_not_running` を報告しなくなります。 |
 | `daemon stop` | プロジェクトのデーモンと、実行中のエンジンセッションを停止します。 |
 | `daemon status` | デーモンの状態(実行中か、ウィンドウモードか、セッション)を報告します。 |
 | `daemon install` | デーモンを起動せずにゲーム内 `gda` harness をプロジェクトへインストールし、作成したパスとセクションをすべて報告します。冪等で、古い `gda` の harness は現行版に再同期されます。`daemon start` 自身がこれを行うため、その `project.godot` の変更を意図的に(レビューやコミットのために)行いたいときにだけ実行します。 |
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=3c57da598896485491b50b52cd19a411aecd42efff28a1242fe06c190f8e4462 -->
+<!-- gda-readme-i18n: source=README.md sha256=ce484830762f3beaff4c979423c5ae400f9e443d077393808ea1b566fefce7ff -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -544,8 +544,8 @@ offset プロパティを明示的に設定してください。
 
 | コマンド | 機能 |
 | ------- | ------------ |
-| `daemon start` | プロジェクトごとのデーモンを起動し、ゲーム内ハーネスをインストールします。エンジンセッションは最初の Live 操作で起動します(`screen` キャプチャには `--windowed`)。 |
-| `daemon wait-ready` | 遅延起動されるエンジンセッションを確立し、Live 読み取りが応答するまで上限つきで待ちます — 「最初の Live 操作」になるための正規の方法で、`daemon start` 直後の最初の `diag errors` が `engine_session_not_running` を報告しなくなります。 |
+| `daemon start` | プロジェクトごとのデーモンを起動し、ゲーム内ハーネスをインストールします。エンジンセッションは、それを必要とする最初の操作で起動します(`screen` キャプチャには `--windowed`)。 |
+| `daemon wait-ready` | 遅延起動されるエンジンセッションを確立し、Live 読み取りが応答するまで上限つきで待ちます — セッション起動を明示的にトリガーする正規の方法です。読み取り専用の `diag`/`logger` はセッションを起動しないため、`daemon start` 直後の最初の `diag errors` は設計上 `engine_session_not_running` を報告します。 |
 | `daemon stop` | プロジェクトのデーモンと、実行中のエンジンセッションを停止します。 |
 | `daemon status` | デーモンの状態(実行中か、ウィンドウモードか、セッション)を報告します。 |
 | `daemon install` | デーモンを起動せずにゲーム内 `gda` harness をプロジェクトへインストールし、作成したパスとセクションをすべて報告します。冪等で、古い `gda` の harness は現行版に再同期されます。`daemon start` 自身がこれを行うため、その `project.godot` の変更を意図的に(レビューやコミットのために)行いたいときにだけ実行します。 |
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=ce484830762f3beaff4c979423c5ae400f9e443d077393808ea1b566fefce7ff -->
+<!-- gda-readme-i18n: source=README.md sha256=8cb5ae07a2d875db3554b7840b6fc90efdca5fa1bd84da220a2de393f6871949 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -545,7 +545,7 @@ offset プロパティを明示的に設定してください。
 | コマンド | 機能 |
 | ------- | ------------ |
 | `daemon start` | プロジェクトごとのデーモンを起動し、ゲーム内ハーネスをインストールします。エンジンセッションは、それを必要とする最初の操作で起動します(`screen` キャプチャには `--windowed`)。 |
-| `daemon wait-ready` | 遅延起動されるエンジンセッションを確立し、Live 読み取りが応答するまで上限つきで待ちます — セッション起動を明示的にトリガーする正規の方法です。読み取り専用の `diag`/`logger` はセッションを起動しないため、`daemon start` 直後の最初の `diag errors` は設計上 `engine_session_not_running` を報告します。 |
+| `daemon wait-ready` | 遅延起動されるエンジンセッションを確立し、Live 読み取りを応答可能にします。`--timeout` は、デーモン側の起動待機と新しい処理の開始判断で共有する予算です。実行中の同期起動呼び出しにより、期限切れの検出が遅れることがあります。これはセッション起動を明示的にトリガーする正規の方法です。読み取り専用の `diag`/`logger` はセッションを起動しないため、`daemon start` 直後の最初の `diag errors` は設計上 `engine_session_not_running` を報告します。 |
 | `daemon stop` | プロジェクトのデーモンと、実行中のエンジンセッションを停止します。 |
 | `daemon status` | デーモンの状態(実行中か、ウィンドウモードか、セッション)を報告します。 |
 | `daemon install` | デーモンを起動せずにゲーム内 `gda` harness をプロジェクトへインストールし、作成したパスとセクションをすべて報告します。冪等で、古い `gda` の harness は現行版に再同期されます。`daemon start` 自身がこれを行うため、その `project.godot` の変更を意図的に(レビューやコミットのために)行いたいときにだけ実行します。 |
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=ce484830762f3beaff4c979423c5ae400f9e443d077393808ea1b566fefce7ff -->
+<!-- gda-readme-i18n: source=README.md sha256=8cb5ae07a2d875db3554b7840b6fc90efdca5fa1bd84da220a2de393f6871949 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -517,7 +517,7 @@ problem 列表只对它到达的那个阶段完整，并非一次覆盖两个阶
 | 命令 | 作用 |
 | ------- | ------------ |
 | `daemon start` | 启动按项目运行的 daemon 并安装游戏内 harness；引擎会话会在第一个需要它的操作时启动（`screen` 截图需加 `--windowed`）。 |
-| `daemon wait-ready` | 建立惰性启动的引擎会话并在有界等待内等到 Live 读取可服务——这是显式触发会话启动的规范做法：只读的 `diag`/`logger` 读取从不启动会话，所以 `daemon start` 之后的第一次 `diag errors` 按设计会报 `engine_session_not_running`。 |
+| `daemon wait-ready` | 建立惰性启动的引擎会话，使 Live 读取可服务。`--timeout` 是 daemon 端由启动等待和新工作决策共享的预算；正在执行的同步启动调用可能延后过期状态的观察。这是显式触发会话启动的规范做法：只读的 `diag`/`logger` 读取从不启动会话，所以 `daemon start` 之后的第一次 `diag errors` 按设计会报 `engine_session_not_running`。 |
 | `daemon stop` | 停止项目的 daemon 以及任何正在运行的引擎会话。 |
 | `daemon status` | 报告 daemon 的状态（是否运行、窗口模式、会话）。 |
 | `daemon install` | 在不启动 daemon 的情况下把游戏内 `gda` harness 安装进项目，并报告它创建的每个路径与配置段。幂等，并会把旧版 `gda` 装下的 harness 同步为当前版本。`daemon start` 自己就会执行这一步，因此只有在你想显式地做出这次 `project.godot` 改动（例如便于审阅或提交）时才需要它。 |
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=6c49dbe1f8202e9cfa46bc8fd79e1c78060238936b0d52e56ec01fd4f2f53e70 -->
+<!-- gda-readme-i18n: source=README.md sha256=3c57da598896485491b50b52cd19a411aecd42efff28a1242fe06c190f8e4462 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -517,6 +517,7 @@ problem 列表只对它到达的那个阶段完整，并非一次覆盖两个阶
 | 命令 | 作用 |
 | ------- | ------------ |
 | `daemon start` | 启动按项目运行的 daemon 并安装游戏内 harness；引擎会话会在第一个 Live 操作时启动（`screen` 截图需加 `--windowed`）。 |
+| `daemon wait-ready` | 建立惰性启动的引擎会话并在有界等待内等到 Live 读取可服务——这是“第一个 Live 操作”的规范做法，如此 `daemon start` 之后的第一次 `diag errors` 就不会报 `engine_session_not_running`。 |
 | `daemon stop` | 停止项目的 daemon 以及任何正在运行的引擎会话。 |
 | `daemon status` | 报告 daemon 的状态（是否运行、窗口模式、会话）。 |
 | `daemon install` | 在不启动 daemon 的情况下把游戏内 `gda` harness 安装进项目，并报告它创建的每个路径与配置段。幂等，并会把旧版 `gda` 装下的 harness 同步为当前版本。`daemon start` 自己就会执行这一步，因此只有在你想显式地做出这次 `project.godot` 改动（例如便于审阅或提交）时才需要它。 |
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=3c57da598896485491b50b52cd19a411aecd42efff28a1242fe06c190f8e4462 -->
+<!-- gda-readme-i18n: source=README.md sha256=ce484830762f3beaff4c979423c5ae400f9e443d077393808ea1b566fefce7ff -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -516,8 +516,8 @@ problem 列表只对它到达的那个阶段完整，并非一次覆盖两个阶
 
 | 命令 | 作用 |
 | ------- | ------------ |
-| `daemon start` | 启动按项目运行的 daemon 并安装游戏内 harness；引擎会话会在第一个 Live 操作时启动（`screen` 截图需加 `--windowed`）。 |
-| `daemon wait-ready` | 建立惰性启动的引擎会话并在有界等待内等到 Live 读取可服务——这是“第一个 Live 操作”的规范做法，如此 `daemon start` 之后的第一次 `diag errors` 就不会报 `engine_session_not_running`。 |
+| `daemon start` | 启动按项目运行的 daemon 并安装游戏内 harness；引擎会话会在第一个需要它的操作时启动（`screen` 截图需加 `--windowed`）。 |
+| `daemon wait-ready` | 建立惰性启动的引擎会话并在有界等待内等到 Live 读取可服务——这是显式触发会话启动的规范做法：只读的 `diag`/`logger` 读取从不启动会话，所以 `daemon start` 之后的第一次 `diag errors` 按设计会报 `engine_session_not_running`。 |
 | `daemon stop` | 停止项目的 daemon 以及任何正在运行的引擎会话。 |
 | `daemon status` | 报告 daemon 的状态（是否运行、窗口模式、会话）。 |
 | `daemon install` | 在不启动 daemon 的情况下把游戏内 `gda` harness 安装进项目，并报告它创建的每个路径与配置段。幂等，并会把旧版 `gda` 装下的 harness 同步为当前版本。`daemon start` 自己就会执行这一步，因此只有在你想显式地做出这次 `project.godot` 改动（例如便于审阅或提交）时才需要它。 |
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -138,9 +138,8 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > which is how an exhausted budget came back whole for a replacement launch.
 >
 > What the instant is, precisely: a **budget for waiting and for committing to new work**, not a
-> hard wall clock. It cannot be the latter — nothing here can interrupt a synchronous system
-> call already in flight, and gda is not going to introduce threads and cancellation to pretend
-> otherwise. So the guarantee is stated as three rules. No phase gets a fresh grace. Every timed
+> hard wall clock. A hard wall-clock limit would require cancellation around synchronous calls;
+> this design does not add that mechanism, so the guarantee is stated as three rules instead. No phase gets a fresh grace. Every timed
 > wait uses what remains, computed where the wait begins rather than where it was decided. And
 > once the instant has passed, the boundary **commits to nothing further** — it does not launch,
 > and a launch already committed is torn down. A slow liveness check, a slow filesystem write, a
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -116,7 +116,8 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > and `gda logger tail` are live operations the daemon serves from the Session log it owns, and they
 > deliberately launch nothing (ADR-0022). The precise rule, and the wording every public surface
 > uses, is that a session launches on **the first operation that REQUIRES an Engine session**.
-> `gda daemon wait-ready` is the explicit, bounded way to BE that operation — the documented
+> `gda daemon wait-ready` is the explicit way to BE that operation, with one daemon-side
+> wait/commit budget — the documented
 > alternative to firing a throwaway read — so an agent can establish the session up front and have
 > its first real read serve, including a first `diag errors`. It is one command in the `daemon`
 > group carried on the live channel (`kind = LIVE`, the `diag errors` precedent); the group-module
@@ -139,7 +140,8 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 >
 > What the instant is, precisely: a **budget for waiting and for committing to new work**, not a
 > hard wall clock. A hard wall-clock limit would require cancellation around synchronous calls;
-> this design does not add that mechanism, so the guarantee is stated as three rules instead. No phase gets a fresh grace. Every timed
+> this design does not add that mechanism, so the guarantee is stated as three rules instead.
+> No phase gets a fresh grace. Every timed
 > wait uses what remains, computed where the wait begins rather than where it was decided. And
 > once the instant has passed, the boundary **commits to nothing further** — it does not launch,
 > and a launch already committed is torn down. A slow liveness check, a slow filesystem write, a
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -130,6 +130,15 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > re-review). A stale session is rebuilt through the same lazy-launch boundary, so `live_timeout`
 > costs a relaunch and, per ADR-0020, the runtime state does not survive it. Correlating replies
 > instead would change the cross-language harness protocol and is left to its own decision.
+>
+> **One deadline, owned by the launch boundary.** Everything that boundary does on a caller's
+> clock draws from a single deadline: retiring the session being replaced, the connect, each
+> handshake frame, and the teardown of a failed launch. Two consequences are not obvious. A
+> socket timeout bounds *inactivity*, so a frame read in chunks must recompute the remaining
+> budget per read or a trickling peer holds the reader indefinitely — the daemon serves one
+> request at a time, so that is every later request too. And retirement is not free: an engine
+> that ignores `SIGTERM` is escalated with what the deadline has left, not with a fresh grace
+> of its own.
 
 ## Decision
 
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -143,7 +143,8 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > indefinitely — the daemon serves one request at a time, so that is every later request too.
 > Retirement is not free: an engine that ignores `SIGTERM` is escalated with what the deadline
 > has left, not with a fresh grace. Collecting a killed child is a duty but not the caller's
-> time, so it is neither charged nor dropped — it happens in the background. And the spawn is
+> time, so it happens in the background — best-effort, since SIGKILL cannot be caught and the
+> cost of a rare miss is one process-table entry the daemon's own exit clears. And the spawn is
 > the one step that cannot be interrupted once begun: the deadline is checked before it and
 > governs everything after, so a spawn that outruns the budget ends in a refusal rather than in
 > a session that came up late.
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -132,13 +132,27 @@ tracked by the Phase-2 PRD (#6) and the gda-daemon feature (#7).
 > instead would change the cross-language harness protocol and is left to its own decision.
 >
 > **One deadline, owned by the launch boundary.** Everything that boundary does on a caller's
-> clock draws from a single deadline: retiring the session being replaced, the connect, each
-> handshake frame, and the teardown of a failed launch. Two consequences are not obvious. A
-> socket timeout bounds *inactivity*, so a frame read in chunks must recompute the remaining
-> budget per read or a trickling peer holds the reader indefinitely — the daemon serves one
-> request at a time, so that is every later request too. And retirement is not free: an engine
-> that ignores `SIGTERM` is escalated with what the deadline has left, not with a fresh grace
-> of its own.
+> clock draws from a single deadline: retiring the session being replaced, the spawn, the
+> connect, each handshake frame, and the teardown of a failed launch. It travels as an
+> **absolute instant**, never as a duration — a duration restarts whatever clock receives it,
+> which is how an exhausted budget came back whole for a replacement launch. When the instant
+> has passed, the boundary REFUSES rather than launching past it.
+>
+> Four consequences are not obvious. A socket timeout bounds *inactivity*, so a frame read in
+> chunks must recompute the remaining budget per read or a trickling peer holds the reader
+> indefinitely — the daemon serves one request at a time, so that is every later request too.
+> Retirement is not free: an engine that ignores `SIGTERM` is escalated with what the deadline
+> has left, not with a fresh grace. Collecting a killed child is a duty but not the caller's
+> time, so it is neither charged nor dropped — it happens in the background. And the spawn is
+> the one step that cannot be interrupted once begun: the deadline is checked before it and
+> governs everything after, so a spawn that outruns the budget ends in a refusal rather than in
+> a session that came up late.
+>
+> The same rule applies to the op relay's own `live_timeout` ceiling, which had the same
+> inactivity shape. Accepted deliberately, and beyond what the `wait-ready` slice needed: a
+> trickled reply now reaches that ceiling where before it completed, and since a timed-out relay
+> marks the channel stale, the consequence is a relaunch and the loss of runtime state. The
+> alternative was to keep publishing a bound that was not one.
 
 ## Decision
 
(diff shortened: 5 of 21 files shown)


qinhaihong-red:
## Review — changes required (`a35b49a4`)

I reviewed the exact diff from `3bfc9ae4` to `a35b49a4` along four independent axes: Markdown artifact quality, domain-modular architecture, repository standards, and issue #657's specification. The happy path works, but the readiness contract is still false on several failure and lifecycle edges.

1. **[P1] `--timeout` does not bound the complete readiness handshake.**

   `src/gda/daemon/session.py:251-295` applies the caller's timeout only to `harness_listener.accept()`. After a peer connects, the token and scene-verification `read_frame(conn)` calls run on a blocking accepted socket with no remaining deadline. A focused probe connected within a `0.05s` timeout but sent no token; the launch was still blocked after `0.20s` and returned only after the peer closed. The CLI's 60-second timeout does not cancel that daemon-side read, so the single-threaded daemon can remain stuck and stop serving later live/control requests.

   Use one monotonic launch deadline across `accept`, token, scene verification, and teardown. Add regressions for a peer that sends neither token nor verification, and prove that the daemon responds to a later control request after the timeout.

2. **[P1] `wait-ready` can report success after the daemon already knows the harness channel is disconnected.**

   `EngineSession.alive()` checks only `proc.poll()`. An `EngineSession.request()` can therefore return `engine_disconnected` for a closed harness peer without invalidating the cached session; while the Godot process remains alive, the next `wait-ready` returns success with `launched: false`, and the next relayed read disconnects again. This contradicts issue #657's contract that success means subsequent live reads serve.

   Mark the session channel stale on EOF/OSError and make readiness include channel availability (or perform a bounded daemon-to-harness readiness probe). `wait-ready` should close and rebuild a stale handle through the shared launch boundary.

3. **[P2] `launched` is inferred outside the launch owner and has a real TOCTOU error.**

   `DaemonServer._handle_wait_ready()` samples `alive()` before `_ensure_session()` samples it again. If the process exits between those checks, `_ensure_session()` launches a replacement but the result still says `launched: false`. A deterministic fake reproduced `actual_launches == 1` with `launched == false`.

   Have `_ensure_session()` / `_session_or_refusal()` atomically return a typed `{session, launched}` outcome instead of predicting the fact in the caller.

4. **[P2] The new ownership guards are not yet single authorities at runtime.**

   - `DAEMON_SERVED_OPS` is consumed only by the cross-language test; production routing still independently decides daemon-served membership with `if op in LOG_OPS` / `if op == WAIT_READY_OP`. Adding a nonexistent member to the tuple still lets the current guard pass. Derive routing and the guard from one handler/category table, or add a mutation-sensitive parity test proving every declared member is intercepted rather than relayed.
   - The `(0, 50]`, finite timeout rule is enforced by `DaemonWaitReadyParams` but not at the daemon IPC boundary in `server.py:248-255`; numeric negative, over-limit, NaN, and infinity values continue to the launcher. Reuse the same validation at wire decode and return a typed refusal.

5. **[P2] Reconcile the public terminology and accepted architecture records.**

   The new implementation deliberately makes one command in the `daemon` group recipe-less and `kind=LIVE`, while `src/gda/commands/daemon.py:1-25` still says every daemon operation runs a recipe and ADR-0040 still describes the group as “recipe-backed, not LIVE.” The transport choice can be valid, but the accepted record and module contract need to state the exception.

   The Markdown surfaces also say that a session launches on “the first live op,” although `diag errors` and `logger tail` are daemon-served Live operations that deliberately do not launch one. Use the precise rule: the session launches on the first operation that requires an Engine session; `wait-ready` explicitly triggers that same launch boundary. Apply the wording consistently to the Skill, command catalog, README, and translations. The documented “bounded” claim also depends on finding 1 being fixed.

### Validation

- `42 passed` across the focused descriptor, Skill/README sync, daemon socket, wait-ready, and live-contract suites
- `1 passed` for the new real-Godot readiness e2e
- `git diff --check` passed; rendered command help and `--schema` were inspected
- Current required CI checks are green; the GitHub `Godot e2e tests` job is skipped
- The full 1,724-test suite, pyright, and ruff were not rerun locally

The happy-path slice and shared refusal extraction are sound, but findings 1–3 keep the command from establishing the promised bounded, serving state reliably. Requesting changes.

qinhaihong-red:
## Review response — all five findings fixed at head `59671962`

| --- | --- | --- |
| [P1] `--timeout` bounds only the accept, not the whole handshake | **Fully adopt** | `launch_session` now derives ONE monotonic deadline and applies the remaining budget to the accept, the token frame, and the scene-verification frame (`conn.settimeout(_remaining())` before each read). Two regressions: a launcher-level test (a peer connects and stays silent → `None` within the bound, with a "sent no auth token within the launch deadline" diagnostic), and an end-to-end one — the REAL `launch_session` inside the serve loop with a silent peer occupying the harness socket → `wait-ready` returns the typed refusal within its bound **and the next `__status__` still serves**. Red-proof: on the reviewed head the launcher test does not fail — it **hangs the suite outright** (the unbounded read), which is the finding demonstrated at its worst. |
| [P1] Success over a dead harness channel | **Fully adopt** | `EngineSession` latches `_channel_stale` on every observed EOF/OSError (and the no-connection branch), and `alive()` now includes the channel. A broken-channel session with a live engine process is dead to the daemon: the next `wait-ready` closes it and rebuilds through the shared launch boundary. Regression (red on the reviewed head): relay breaks → `engine_disconnected`; next `wait-ready` → `launched: true` (second launch); next read serves. Disclosed scope: a `live_timeout` deliberately does NOT latch stale — a slow op is not a broken channel, and the possibly-dirty-stream question after a late reply predates this slice (noted in the code). |
| [P2] `launched` inferred outside the launch owner (TOCTOU) | **Fully adopt** | `_ensure_session` / `_session_or_refusal` now return a typed `_Established(session, launched)` — the fact travels with the decision; the caller's pre-sample is gone. Regression with a flip-flopping liveness fake pins the invariant `launched == (a launch actually happened on this call)`; red on the reviewed head (launch happened, reported `false`). |
| [P2] Ownership guards not runtime authorities | **Fully adopt** | (a) `_handle` now routes on `DAEMON_SERVED_OPS` itself (`_handle_daemon_served` dispatches members; an undeclared member refuses loudly instead of relaying), and a mutation-sensitive parity test drives every declared member through the real loop with a session cached, asserting none reaches the relay. (b) The finite `(0, 50]` rule is re-enforced at the daemon IPC boundary — negative, zero, over-limit, `inf`, `nan`, booleans and strings are refused with a typed `invalid_params` (red on the reviewed head: silently forwarded). The cap is one constant, `WAIT_READY_TIMEOUT_MAX` in `server.py`, consumed by both the params model (`le=`) and the wire check. |
| [P2] Terminology and records | **Fully adopt** | ADR-0040 carries a dated amendment for the group's one `kind = LIVE` command; the `gda.commands.daemon` module docstring states the exception explicitly. Every surface now uses the precise rule — a session launches on the first operation that **requires** one, and the read-only `diag`/`logger` reads never do — in the Skill (prereq paragraph, daemon row, example comment), README + all three translations, the command catalog, the command help, and the result-field description. The "bounded" claim is now true per finding 1. |

Validation at `59671962`: fast suite **1730 passed** (5 new regressions, all red-proofed against `a35b49a4` — the F1 launcher test *hangs* there rather than failing); pyright 0 errors; ruff clean; i18n re-stamped; real-Godot daemon e2e **32 passed** (serial under the shared lock). PR CI running on the new head.

qinhaihong-red:
## Re-review — changes still required (`59671962`)

I rechecked the previous findings against the exact current head `5967196295b88b109a3a0713335de09386cd07db` and the latest response. The original unbounded handshake reads, broken-channel EOF/OSError handling, `launched` TOCTOU, runtime routing authority, wire-timeout validation, and most repository documentation drift are fixed. Two runtime residues and one documentation residue remain.

1. **[P1] `live_timeout` leaves a response-unsafe session eligible for `wait-ready` success.**

   `EngineSession.request()` deliberately does not latch `_channel_stale` on `TimeoutError` (`src/gda/daemon/session.py:124-132`). The protocol has no request ID and the timed-out frame is not drained, so a late response is indistinguishable from the next operation's response. On this head I reproduced:

   ```text
   first request -> live_timeout
   session.alive() -> true
   wait-ready -> {launched: false}
   second request -> receives the first request's late response
   ```

   This violates #657's success postcondition that subsequent live reads serve, and can return a validly-framed but semantically wrong result. Mark the channel stale after a live timeout and rebuild through the shared launch boundary, or introduce response correlation/reliable draining; add the sequence above as a regression.

2. **[P2] The caller's readiness deadline still does not cover teardown.**

   `launch_session()` exhausts the monotonic deadline in the handshake and then `_terminate()` starts a fresh `proc.wait(timeout=5)` (`src/gda/daemon/session.py:272-320,366-381`). With a child that ignores `SIGTERM`, `timeout=0.3` returned after **5.308s**, blocking the daemon's single serve loop for the same interval. The `(0, 50]` cap preserves the separate 60s transport envelope, but it does not satisfy the documented per-call timeout.

   Pass the same absolute deadline into teardown; when no budget remains, escalate immediately instead of starting another five-second grace period. The regression should measure the whole wait-ready round trip and prove a following control request is served promptly.

3. **[P2] Reconcile the PR's Markdown with the fixed head.**

   The PR body still says "first live op" / "be that first op", describes only a harness-connect wait, calls the params model the sole timeout owner, and reports the old 1,724-test / two-socket-test validation. Update Contract, Shape, Bound, Guard, and Validation to the current implementation and evidence. As a smaller accepted-record residue, ADR-0017 lines 27 and 89 still say "first live op" while current docs use the precise "first operation that requires an Engine session" rule.

### Recheck validation

- Focused daemon suites: **124 passed**
- Fast non-e2e suite: **1,728 passed, 2 skipped**
- Real-Godot readiness acceptance e2e: **1 passed**
- Ruff, pyright, and `git diff --check`: passed
- Required GitHub checks are green; the GitHub Godot e2e job is skipped

The fixes substantially improve the slice, but finding 1 still permits `wait-ready` to certify a channel that can return the wrong operation's response, so this is not ready to merge yet.

qinhaihong-red:
## Re-review response — all three residues fixed at head `b8ac5706`

Baseline: base tip `3bfc9ae4`, merge base `3bfc9ae4`, reviewed head `59671962`, new head `b8ac5706`. Both runtime findings were reproduced on `59671962` before anything was changed.

| [P1] `live_timeout` leaves a response-unsafe session eligible for `wait-ready` success | **Fully adopt** | Reproduced verbatim on `59671962`: op A times out → `alive()` true → `wait-ready {launched: false}` → op B reads A's late payload and returns it as B's result | `request()` latches `_channel_stale` on `TimeoutError` too | 2 regressions, both red on `59671962` |
| [P2] The readiness deadline does not cover teardown | **Fully adopt** | Reproduced: a SIGTERM-ignoring child made `launch_session(timeout=0.3)` return after **5.306s** | `_terminate(proc, grace)` takes the grace as a BUDGET; the launch paths pass what remains of the one monotonic deadline | 1 regression measuring the round trip **and** the request behind it, red on `59671962` (5.3s) |
| [P2] PR Markdown and ADR-0017 still describe the pre-fix state | **Fully adopt** | PR body's Contract/Shape/Bound/Guard/Validation, ADR-0017 lines 27 and 89 | PR body rewritten to the current implementation and evidence; ADR-0017 gains a dated amendment, with both historical lines pointing at it | — |

### 1 — the timed-out relay

The earlier reading ("a slow op is not a broken channel") priced only the churn of relaunching. Your repro shows the real price, and it is worse: this protocol carries no request id and the timed-out frame is never drained, so a late reply is indistinguishable from the next operation's reply. A channel that can answer with **another operation's** result is not serving, whatever the engine process is doing — so the timeout now latches the channel stale exactly like a dropped connection, and the next session-needing op rebuilds through the shared launch boundary.

I did not add reply correlation or a drain: correlation changes the cross-language harness protocol and deserves its own decision, and a drain cannot know whether or when the late frame arrives. The accepted cost — a relaunch after an op that outran `OP_TIMEOUT`, with runtime state gone (ADR-0020) — is now stated in the ADR-0017 amendment, the command catalog, the Skill, and the `live_timeout` registry entry, rather than left as an open comment in the code.

Regressions (both red on `59671962`):

- `test_a_timed_out_relay_leaves_the_session_dead_to_the_daemon` — your exact sequence: op times out, the harness answers late, the engine process is still alive, `alive()` must be false (was `true`).
- `test_wait_ready_rebuilds_a_session_whose_relay_timed_out` — through the real serve loop: `wait-ready` → relay times out → `wait-ready` must relaunch (`launched: true`, was `false`) → the next read serves from the fresh session.

### 2 — teardown inside the deadline

`_terminate` now takes `grace` as a budget rather than a fixed pause, and `launch_session`'s failure paths pass `deadline - now`; with nothing left the escalation to SIGKILL is immediate. `EngineSession.close()` keeps the 5s default — no caller deadline exists there.

One consequence I checked rather than assumed: killing sooner does not cost the failed-launch diagnostics. Godot's file logger flushes on every error and, in a debug build, on every print (`core/io/logger.cpp`, `application/run/flush_stdout_on_print.debug` defaults true in `main/main.cpp`), so the Session log the daemon tails is already on disk when the child is killed.

Regression `test_a_teardown_cannot_outlast_the_readiness_deadline` (red on `59671962` at 5.3s): a real `launch_session` inside the serve loop spawns a child that ignores SIGTERM; the `wait-ready --timeout 0.3` round trip must finish under 2s **and** the `__status__` behind it must be served promptly.

### 3 — records

ADR-0017 carries a dated amendment (2026-08-21, #657) for the two facts that were drifting: the launch trigger stated precisely — the first operation that **requires** a session, which `wait-ready` explicitly is — and what keeps a session serving (process **and** channel, with the no-request-id reason for the timeout case). Lines 27 and 89 keep their point-in-time prose and now point at it, rather than being rewritten to today's state. The PR body's Contract, Shape, Bound, Guard and Validation now match the implementation, and gained a Serving-state bullet.

### Validation at `b8ac5706`

- Fast suite **1733 passed**; pyright 0 errors; ruff clean (`format` included); `git diff --check` clean.
- Red-proof of the 3 new regressions against `59671962`: all 3 fail there (`assert True is False`; `launched: false`; `teardown held the round trip for 5.3s`).
- Real-engine e2e, serial under the shared lock: daemon suite **28 passed**; live suites (`game`, `diag`, `logger`, `perf`, `input`, `harness install`, `mcp live`) **26 passed**.
- Surface: `gda schema` and the rendered help are **byte-identical** to `59671962` — the `live_timeout` registry description is maintainer-facing and reaches no published surface. The only shipped-surface change is one added SKILL.md sentence about the discarded session.

Not verified / disclosed:

- One unrelated flake seen once in a full-suite run under load: `tests/test_launch.py::test_streaming_timeout_preserves_the_output_the_child_already_wrote` (a #655 streaming test) asserted `'' == 'SUITE START\n'`. It passed on 4 subsequent runs including a full re-run, and nothing in this diff touches the headless launch path. I have not filed it — happy to, if you want it tracked rather than watched.

qinhaihong-red:
The one flake disclosed in the validation above is now tracked as #728 (test-infra only — the streaming-capture guarantee itself is intact). It belongs to #655's suite, not to this slice, so this PR's diff is unchanged.

qinhaihong-red:
## Second re-review — changes still required (`b8ac5706`)

I rechecked the latest response and the exact current head `b8ac570653761af776cb7ecb4de6b938952c40f1` along the same four independent axes: Markdown artifacts, domain-modular architecture, repository standards, and issue #657's specification.

The previous `live_timeout` residue is fixed: a timed-out relay now makes the channel stale, `wait-ready` rebuilds it, and a late reply on the old channel cannot be consumed by the next operation. The PR-body and ADR amendment drift from the previous re-review is also fixed. Two bounded-readiness gaps and one teardown residue remain.

1. **[P1] A partial-frame trickle can still outlive the absolute readiness deadline.**

   `launch_session()` sets a relative socket timeout once before reading the token frame and once before reading the verification frame (`src/gda/daemon/session.py:304-333`). `_recv_exactly()` then performs repeated `recv()` calls (`src/gda/daemon/protocol.py:54-64`), and each successful partial read restarts that relative inactivity timeout. A peer can therefore keep the single-threaded daemon occupied indefinitely by sending each next byte just before the timeout.

   On this head, with `timeout=0.05`, focused trickle probes completed successfully after **0.722s** for the token frame and **1.093s** for the verification frame. Recompute the remaining absolute-deadline budget before every `recv`, and add token- and verification-partial-frame regressions.

2. **[P1] Retiring a stale cached session is outside the caller's `wait-ready` deadline.**

   `_ensure_session()` calls `self._session.close()` before starting the new launch (`src/gda/daemon/server.py:416-433`). `EngineSession.close()` retains an independent five-second termination grace, so this work is not charged to the caller's readiness deadline. With a live-timed-out cached session backed by a real child that ignores SIGTERM, `wait-ready --timeout 0.3` took **5.002s**; a following control request was blocked for the same interval.

   The caller/use-case layer needs one deadline covering both stale-session retirement and replacement launch, passing only the remaining budget to each phase. The current regression mocks `_terminate` to a no-op, so it does not exercise this composition.

3. **[P2] The SIGKILL path does not explicitly reap the child.**

   `_terminate()` calls `proc.kill()` after the grace expires but does not call `wait()` afterward (`src/gda/daemon/session.py:389-415`). A focused stubborn-child probe returned with `Popen.returncode is None`; `waitid(..., WNOWAIT)` showed an exited-but-unreaped SIGKILL child. Explicitly reap after the kill so the lifecycle owner completes teardown deterministically.

### Four-axis disposition

- **Artifact Review — changes required:** the updated PR body, command catalog, and Skill still overstate the whole readiness attempt as caller-bounded because findings 1 and 2 are counterexamples. The ADR-0017 historical wording/amendment is otherwise reconciled.
- **Domain-modular architecture — changes required:** the launch boundary does not yet own one deadline across frame consumption, stale-handle retirement, replacement launch, and teardown; the kill path also leaves ownership incomplete.
- **Standards — blocking:** the stale-close path violates the documented public timeout contract. Non-blocking maintainability note: `_terminate` is monkeypatched repeatedly across the two affected test files; a shared fixture plus at least one real-teardown serve-loop test would reduce the chance of hiding this interaction again.
- **Spec — changes required:** findings 1 and 2 directly violate issue #657's bounded-readiness requirement and can block later control traffic despite the caller timeout.

### Validation

- The three new author regressions: **3 passed**
- Focused daemon suites: **127 passed**
- Fast non-e2e suite: **1,731 passed, 2 skipped, 499 deselected**
- Real-Godot readiness acceptance e2e: **1 passed**
- Ruff, pyright, and `git diff --check`: passed
- Required GitHub checks are green; the GitHub `Godot e2e tests` job is skipped

The response-unsafe `live_timeout` problem is closed, but the two P1 paths still let `wait-ready` exceed its caller-provided bound and monopolize the daemon serve loop, so this is not ready to merge yet.

qinhaihong-red:
## Second re-review response — all three fixed at head `cdb3c900`

Baseline: base tip `3bfc9ae4`, merge base `3bfc9ae4`, reviewed head `b8ac5706`, new head `cdb3c900`. Each finding was reproduced on `b8ac5706` before anything changed.

| Claim | Decision | Reproduced on `b8ac5706` | Change | Validation |
| [P1] A partial-frame trickle outlives the absolute deadline | **Fully adopt** | 0.04s/byte on a 0.05s bound: token frame held **0.735s**; verification frame held **2.085s** and the launch then **succeeded** | The framing layer takes an absolute deadline and recomputes the budget before every `recv` | 2 parametrized regressions (token / verification), both red |
| [P1] Retiring a stale cached session is outside the deadline | **Fully adopt** | `wait-ready --timeout 0.3` against a stale session whose engine ignores SIGTERM: **5.002s**, with the request behind it blocked | The launch boundary derives ONE deadline and hands retirement, then the launch, only what the previous phase left | 1 regression with a REAL teardown — no `_terminate` mock — measuring the round trip, the request behind it, and the budget the launch received; red at 5.0s |
| [P2] The SIGKILL path does not reap | **Fully adopt** | `returncode` stayed `None` after teardown | Reaped after the kill, within a grace that is a formality (0.5s) rather than a wait | 1 regression, red with "the killed child was never reaped" |

### 1 — the trickle

You are right about the shape, and the fix belongs one layer down: `read_frame` now takes an absolute `time.monotonic()` instant and `_recv_exactly` recomputes `deadline - now` before **every** `recv`, raising `TimeoutError` rather than issuing a read that could still block. Setting the socket timeout once per frame could only ever bound inactivity.

**One extension beyond the finding, deliberate and disclosed.** `EngineSession.request()` had the identical shape — `OP_TIMEOUT` set once, the reply read in as many chunks as the harness sends — so `the engine session did not return within 30s` was not true of a trickled reply. It now uses the same absolute deadline. The harness is trusted (ADR-0009), so this is honesty about a published bound rather than a defence, and it is one argument at a call site I was already changing; leaving a bound I could see to be false seemed worse than the small scope increase. Guarded by `test_a_trickled_reply_cannot_outlast_the_relays_own_bound` (red on `b8ac5706`: the trickled reply was waited out and returned as a **success**).

### 2 — retirement inside the deadline

`_ensure_session` now derives `deadline = now + (caller timeout or CONNECT_TIMEOUT)` and spends it in order: `close(remaining)` on the session being replaced, then `launch(remaining)`. `SessionHandle.close` grew a `grace` parameter with a default, so the callers that have no clock — daemon shutdown, the tests — are unchanged; the structural Protocol made the seam's fakes update with it.

A consequence worth stating: with no `--timeout`, retirement now eats into `CONNECT_TIMEOUT` instead of leaving it whole. That is the honest reading of a bound — the caller asked for a total, not a total per phase.

Your note that the existing regression mocked `_terminate` was exactly right, and it cost me a second round to see: the new test takes a real child and does **not** request the mock. Two related repairs came out of writing it —

- The mock is now one named helper (`tests.support.no_engine_teardown`) whose docstring says what it hides and that a teardown test must not call it. That is your non-blocking maintainability point, adopted at the size it deserves.
- The stubborn child now **announces itself** before the test proceeds. The SIGTERM handler is installed by the child, so a teardown fired ~1ms after spawn kills it during interpreter startup by the default disposition: my first reap regression passed on `b8ac5706` for that reason alone (`returncode` was `-15`, not `-9`) and was no regression at all. The readiness line is the synchronisation.

### 3 — the reap

Adopted. `_REAP_GRACE = 0.5` is the only interval teardown spends outside the caller's budget, and the code says so: a SIGKILL'd process is reapable in microseconds, and a child that somehow outlasts even that is left unreaped rather than allowed to block the serve loop.

### Records

ADR-0017's amendment gains a third paragraph — the one deadline the launch boundary owns, and the two non-obvious consequences (inactivity-vs-read, retirement is not free). The command catalog's `--timeout` sentence now names what the bound actually covers, so the "bounded" claim in the PR body, catalog and Skill is true rather than aspirational.

### Validation at `cdb3c900`

- Fast suite **1738 passed**; pyright 0 errors; ruff clean (`format` included); `git diff --check` clean.
- Red-proof of the 5 new regressions against `b8ac5706`: all 5 fail there, with the numbers in the table above.
- Real-engine e2e, serial under the shared lock: **54 passed** across the daemon and live suites (`daemon`, `game`, `diag`, `logger`, `perf`, `input`, `harness install`, `mcp live`) — the framing change touches every daemon read, so the whole live tier was re-run rather than the readiness test alone.
- `gda schema` byte-identical to `b8ac5706`.

Residual risk I have not closed: the CLI↔daemon leg still reads its frames with no deadline, because no caller of it promises one — a slow client there costs the serve loop, not a violated bound. If you want that bounded too it is the same one argument, but it wants its own decision about what the bound should be.

qinhaihong-red:
## Third re-review — changes still required (`cdb3c900`)

I rechecked the latest response against the exact current head `cdb3c900175429282a5f2217458f64de7f181d20` along the same four independent axes: Markdown artifacts, domain-modular architecture, repository standards, and issue #657's specification.

The previous partial-frame findings are fixed: token, verification, and relay-reply reads now recompute the remaining absolute deadline before every `recv`. Stale-session retirement now receives the caller's remaining budget, the duplicated teardown mock is consolidated, and the normal real SIGKILL path is reaped by its owner. One blocking deadline-owner residue remains.

1. **[P1] The readiness attempt still does not use one caller-owned absolute deadline.**

   The server creates a deadline, but passes the launcher a relative duration with `max(_left(), 0.001)` (`src/gda/daemon/server.py:429-475`). An exhausted budget is therefore revived for a replacement launch. `launch_session()` then truncates the log and calls `Popen` before creating a new deadline from that duration (`src/gda/daemon/session.py:258-305`). This is a second clock, not the absolute instant owned by `_ensure_session`.

   A controlled production-launch-seam probe used a `0.05s` caller timeout, spent `0.04s` retiring the stale session, then delayed `Popen` by `0.04s` with both handshake frames already queued. The command still succeeded with `{launched: true}` after **0.087s**. A simpler `0.05s` launch probe with a delayed `Popen` likewise succeeded after **0.125s**.

   The reap path creates another independent allowance: after the caller's grace expires, `_terminate()` waits a fixed `_REAP_GRACE = 0.5` outside the caller budget and suppresses a second `TimeoutExpired` (`src/gda/daemon/session.py:418-445`). A controlled slow-reap probe produced:

   ```text
   wait-ready timeout: 0.01s
   actual result:       success, launched=true, after 0.501s
   replacement budget: 0.001s
   child reaped:        false
   ```

   In an actual serve-loop variant, the following status request remained queued for about **0.462s**. This contradicts ADR-0017 and the catalog's claim that `--timeout` bounds the whole daemon-side readiness attempt on ONE deadline; it also means the PR-body statement that the teardown always reaps the child is not true on the path the code explicitly permits.

   Have `_ensure_session` pass the same absolute deadline through `close` and `launch_session`; do not launch when it is exhausted, and do not recreate it after `Popen`. Reap within the remaining budget or asynchronously. If the intended product contract is instead `timeout + 0.5s`, state that explicitly and remove the whole/one-deadline claims.

   The current regressions do not pin the public bound: their `0.3s` requests accept anything below `2.0s`, the launcher assertion allows any value `<= 0.3`, and the status request is sent only after wait-ready returns. Tighten them to the caller timeout plus a small scheduling tolerance, make a restored full-duration mutation fail, and enqueue the control request while wait-ready is still active.

### Four-axis disposition

- **Artifact Review — changes required:** ADR-0017, the command catalog, Skill, and PR body still overstate the timeout as covering the whole attempt. Terminology and ASD-STE100 prose otherwise pass.
- **Domain-modular architecture — P1 blocking:** the server is the deadline owner, but deadline authority is duplicated by the launcher and bypassed by the reap allowance. Frame ownership, stale-session ownership, and normal reap ownership are otherwise sound.
- **Standards — two hard violations:** the post-budget reap and the post-`Popen` clock reset violate the documented one-deadline contract. The previous Duplicated Code smell is closed by `tests.support.no_engine_teardown`; no new baseline smell was found.
- **Spec — one P1 plus one P2 scope note:** the P1 above leaves issue #657's documented bounded wait only partially implemented. Separately, changing every live relay reply from an inactivity timeout to an absolute `OP_TIMEOUT` is outside #657's `start -> wait-ready -> first diag errors` path. The DDMA axis finds that small DRY extension coherent, but the Spec axis records it as behavior that should be explicitly accepted or tracked separately because timeout now makes the session stale and a later relaunch loses runtime state.

### Validation

- Focused daemon suites: **133 passed**
- Fast non-e2e suite: **1,736 passed, 2 skipped, 499 deselected**
- Real-Godot readiness acceptance e2e: **1 passed**
- Ruff, pyright, and `git diff --check`: passed
- Current required GitHub checks are green; the GitHub `Godot e2e tests` job is skipped
- I did not independently rerun the author's full 54-test real-engine matrix

The three reported mechanics are substantially improved, but the remaining split-clock path still allows `wait-ready` to succeed after its caller-provided timeout and to block later control traffic, so this is not ready to merge yet.

qinhaihong-red:
## Third re-review response — the split clock is closed at head `29b22d34`

Baseline: base tip `3bfc9ae4`, merge base `3bfc9ae4`, reviewed head `cdb3c900`, new head `29b22d34`. Both mechanics were reproduced on `cdb3c900` before anything changed.

| Claim | Decision | Reproduced on `cdb3c900` | Change |
| --- | --- | --- | --- |
| [P1] The launcher receives a duration and restarts the clock after `Popen` | **Fully adopt** | `launch_session` with a 0.05s budget and a 0.12s spawn **succeeded** after 0.123s | The seam carries the caller's absolute instant; an expired one is a refusal at both ends |
| [P1] The reap allowance is spent outside the budget | **Fully adopt** | `_terminate(grace=0.0)` returned after **0.502s** | Collection moved to a background reaper; teardown returns as soon as it has killed |
| [P2] The regressions do not pin the public bound | **Fully adopt** | — | Bounds tightened to the caller's timeout plus a named slack; the control request is enqueued *while* wait-ready runs |
| [P2 scope] The relay's absolute `OP_TIMEOUT` is outside #657's path | **Explicitly accepted, recorded** | — | Stated as a deliberate out-of-slice acceptance in ADR-0017, with its consequence |

### 1 — one instant, not two clocks

`SessionLaunch` now takes `deadline` instead of `timeout`, and `launch_session` takes it **before** it truncates the log or spawns anything. Every phase after — the accept, both frames, the teardown — draws from that instant. `_ensure_session` passes the same instant it is honouring, and refuses to launch at all when retirement consumed the budget (with the reason in the launch diagnostics, since an empty refusal reads as a mystery).

One limit I will not claim away: the spawn is the single step that cannot be interrupted once begun. The deadline is checked immediately before it and governs everything after it, so a spawn that outruns the budget ends in a **refusal** rather than in a session that came up late — which is what the probe now shows (`None` after 0.123s, where `cdb3c900` returned a live session). ADR-0017 and the catalog say this explicitly rather than implying the spawn is interruptible.

### 2 — collection is a duty, not the caller's time

You are right that a fixed allowance is a second budget however small it is. It is gone. `_terminate` kills and returns; one short-lived daemon thread owns `wait()`. Neither charged nor dropped.

Writing the regression for this corrected something in my own last round: a real SIGKILL is collectable in microseconds, so a test with a real child cannot see the allowance at all — my previous timing assertion would have passed on `cdb3c900`. The property only shows against a child the kernel has not made reapable yet (uninterruptible I/O, a loaded host), which is exactly the worst case the bound has to hold in, so the test now uses a stand-in whose collection blocks until released. It is red on `cdb3c900` at 0.56s.

### 3 — the regressions now pin the bound

- The `0.3s` requests assert against `0.3 + _SCHEDULING_SLACK` (0.35s, named and justified in one place) instead of `2.0s`, so a restored full-duration wait fails them.
- The control request behind wait-ready is **enqueued while it is still being served**, on its own thread, and its own latency is asserted — the daemon serves one at a time, so that is what a queued caller actually waits.
- The launcher assertion is now about the instant it received: no later than the request plus the caller's timeout (plus that request's own transit, since the daemon starts the clock when it *receives* the call — a bound of 25s, the default, would fail it).
- One new regression pins the spawn case directly: both handshake frames pre-queued, a 0.12s spawn, a 0.05s deadline, and the launch must still refuse.

### 4 — the relay ceiling, accepted rather than assumed

Recorded in the ADR-0017 amendment as what it is: an acceptance beyond what this slice needed. A trickled reply now reaches `live_timeout` where before it completed, and because a timed-out relay marks the channel stale, the consequence is a relaunch and the loss of runtime state. I kept it rather than tracking it separately because the alternative is to go on publishing a bound that is not one — but the record now names the trade instead of leaving it to be rediscovered. Happy to split it out if you would rather it landed on its own.

### Validation at `29b22d34`

- Fast suite **1740 passed**; pyright 0 errors; ruff clean; `git diff --check` clean.
- Red-proof against `cdb3c900`: 3 of the 4 new/tightened regressions fail there (launch succeeded past its bound; a replacement launched after the budget was exhausted; teardown spent 0.56s). The fourth — "a killed child is still collected" — is *not* red there, because the reap exists on that head; it pins the property so moving it to the background could not silently drop it.
- Real-engine e2e, serial under the shared lock: **54 passed** across the daemon and live suites.
- `gda schema` byte-identical to `cdb3c900`.

Still open, unchanged from my last reply: the CLI↔daemon leg reads its frames with no deadline, because no caller of it promises one. It costs the serve loop, not a violated bound, and picking a value for it wants its own decision.

qinhaihong-red:
## Fourth re-review — changes still required (`29b22d34`)

I rechecked the latest response against the exact current head `29b22d34236882136d6aefee75c56b6954fea526`, replaying only the previous residue and its adjacent deadline/lifecycle edges.

The previous late-success path is fixed: a `Popen` call that itself outlives the deadline now ends in a refusal rather than returning a live session. The fixed post-budget reap allowance is also gone from the caller path, and the relay `OP_TIMEOUT` scope extension is now explicitly accepted in ADR-0017. Two blocking deadline-owner gaps and two adjacent residues remain.

1. **[P1] Stale-session retirement still restarts the caller's deadline as a duration.**

   `_ensure_session()` creates its deadline only after the initial `self._session.alive()` call, then converts the instant once into `close(max(deadline - monotonic(), 0))` (`src/gda/daemon/server.py:420-434`). `EngineSession.close()` / `_terminate()` receive that duration; after connection close, `proc.poll()`, and signal work have consumed time, `_terminate()` still gives `proc.wait()` the original full duration (`src/gda/daemon/session.py:171-178, 444-454`).

   A deterministic production-seam probe used `timeout=0.05s`; `poll()` consumed `0.08s`, after which `wait(timeout≈0.05)` received the full original grace. `wait-ready` returned after **0.222s** (`launches=0`, `wait_timeouts=[≈0.05, None]`). Refusing the replacement is correct, but it does not make the whole caller-owned bound true.

   Create the absolute deadline at entry to the launch boundary, pass that instant through `alive` / `close` / `_terminate`, and recompute `max(deadline - monotonic(), 0)` immediately before every potentially blocking wait. An uninterruptible step may itself overrun, but it must not be followed by a revived grace period.

2. **[P1] Blocking pre-spawn preparation can cross the deadline and still launch the engine.**

   `launch_session()` checks `_left()` before truncating the Session log (`src/gda/daemon/session.py:272`), but performs the potentially blocking `Path.write_bytes()` and then calls `Popen` without another deadline gate (`src/gda/daemon/session.py:292-303`). With a `0.05s` deadline and a `0.12s` log write, `Popen` was still called once, **0.076s after the deadline had passed**.

   Recheck the same instant immediately before `Popen`; if it is spent, record a spawn-preparation deadline reason and do not launch. Also check immediately after `Popen` returns: the already-fixed slow-spawn case refuses, but currently reports that the harness sent no auth token even when token and verification frames were pre-queued. That is a false phase attribution; it should report that the deadline expired during spawn and tear down with zero remaining budget.

3. **[P2, adjacent lifecycle] SIGKILL escalation does not cover the owned process group.**

   Teardown sends `SIGTERM` to the process group but escalates with `proc.kill()`, which kills only the leader (`src/gda/daemon/session.py:440-447`). In a real leader-plus-descendant probe where both ignored `SIGTERM`, teardown returned in `0.053s` and reaped the leader, but the descendant remained alive. If the Engine session owns the group—as `start_new_session=True` and group-wide TERM imply—the escalation should send `SIGKILL` to the group as well. Otherwise, direct-child-only ownership needs to be stated and enforced consistently.

4. **[P2, artifacts] Narrow two guarantees to what the implementation and tests prove.**

   - The PR body and ADR say the deadline travels as an instant, never a duration; finding 1 shows retirement still receives a duration.
   - “The child is collected anyway” / “neither charged nor dropped” is stronger than a daemon background thread whose `wait()` exceptions are swallowed and whose lifetime is not drained at daemon shutdown. Either give the reaper managed lifecycle ownership or state this as best-effort background collection.
   - The PR body says the queued-control-request bound is pinned for both a failed launch child and a stale child being retired. The stale-retirement regression does enqueue the request while `wait-ready` runs; `test_a_teardown_cannot_outlast_the_readiness_deadline` still sends its status request only after `wait-ready` returns.

### Four-axis disposition

- **Artifact Review — changes required:** the one-instant and unconditional-collection claims remain stronger than the implementation; the validation inventory overstates one launch-child regression.
- **Domain-modular architecture — P1 blocking:** deadline authority is still downgraded at the stale-retirement boundary and bypassed during pre-spawn preparation. The process-group escalation gap is an adjacent P2 lifecycle issue.
- **Standards — two hard/P2 violations:** the documented whole-attempt bound is still false on stale retirement, and an expiry during spawn is reported as a missing-token failure despite the token already being available.
- **Spec — pass:** the issue #657 path no longer returns late success, the fixed synchronous reap allowance is closed, and the out-of-slice relay ceiling is explicitly recorded. No remaining Spec-axis finding.

### Recheck validation

- Focused daemon suites: **135 passed**
- Fast non-e2e selection: **1,738 passed, 2 skipped, 499 deselected**
- Real-Godot readiness acceptance: **1 passed**
- Ruff, pyright, and `git diff --check`: passed
- Current required GitHub checks are green; the GitHub Godot e2e job is skipped
- I did not independently rerun the author's full 54-test real-engine matrix

The repair closes the previous split-clock late-success mechanism, but the caller still does not own one absolute deadline across stale retirement and all avoidable pre-spawn work. Requesting changes.

qinhaihong-red:
## Fourth re-review response — all four addressed at head `b196c829`

Baseline: base tip `3bfc9ae4`, merge base `3bfc9ae4`, reviewed head `29b22d34`, new head `b196c829`. Each mechanic was reproduced on `29b22d34` before anything changed.

| Claim | Decision | Reproduced on `29b22d34` | Change |
| --- | --- | --- | --- |
| [P1] Retirement restarts the deadline as a duration | **Fully adopt** | 0.08s of pre-wait work against a 0.05s budget: `wait()` was still given **0.05** | `close` / `_terminate` take the instant; the remainder is computed where it is used |
| [P1] Pre-spawn preparation can cross the deadline and still launch | **Fully adopt** | a 0.12s log write against a 0.05s deadline reached `Popen` **0.073s late** | The gate moved to the last interruptible point, plus a post-spawn check that names the real phase |
| [P2] The escalation misses the owned process group | **Fully adopt** | leader + descendant both ignoring SIGTERM: teardown returned in 0.053s, the **descendant stayed alive** | `SIGKILL` goes to the group, matching the `SIGTERM` before it |
| [P2] Two claims are stronger than the code, one inventory line is wrong | **Fully adopt** | — | Collection stated as best-effort; the second regression now enqueues its control request concurrently |

### 1 — the instant goes all the way down

`EngineSession.close` and `_terminate` take a deadline instead of a grace, and `_terminate` computes `deadline - now` **immediately before `proc.wait()`** — after the channel close, the poll and the signal, all of which spend from the same budget. `_ensure_session` now takes its deadline at entry, before the liveness check, and passes that same instant to `close`. There is one representation of the bound in the whole path; a caller with no clock of its own (daemon shutdown, the tests) mints one from `TERMINATE_GRACE` at the moment it starts.

The uninterruptible step can still overrun — a slow `poll()` in the probe costs 0.083s on a 0.05s budget — but nothing after it gets a revived grace: `wait()` now receives `0.0`.

### 2 — the last interruptible point, and honest phase attribution

The gate moved to immediately before `Popen`, so preparation that crosses the deadline spawns nothing and says why ("expired while preparing the launch"). After `Popen` returns there is one more check: an expiry during the spawn is reported as an expiry during the spawn. Your reading of the old path was exactly right — it fell through and blamed the harness for sending no auth token while the token was already queued on the connection. That is now a regression whose red message on `29b22d34` is that same false sentence.

### 3 — the group

Adopted as stated: the session is spawned with `start_new_session=True`, so gda owns the group, and an escalation that reaches only the leader contradicts the `SIGTERM` that preceded it. Both signals now go through one helper that targets the group and falls back to the child alone when there is no group. The regression spawns a leader with a descendant, both ignoring SIGTERM, and fails on `29b22d34` with "the descendant outlived the group kill".

### 4 — claims narrowed, inventory corrected

- The instant-never-a-duration claim is now true rather than aspirational (finding 1).
- Background collection is documented as **best-effort** in both the code and ADR-0017, with the reason it is not worth a managed lifecycle: SIGKILL cannot be caught, so the wait ends on its own in every observed case, and the cost of a rare miss is one process-table entry the daemon's own exit clears. I did not give the reaper a lifecycle — that is a guarantee nothing in the system needs.
- `test_a_teardown_cannot_outlast_the_readiness_deadline` now enqueues its control request on its own thread **while** wait-ready is being served, like the stale-retirement one, and asserts against `timeout + slack` instead of 2.0s. The PR body's claim matches both regressions now.

### Validation at `b196c829`

- Fast suite **1744 passed**; pyright 0 errors; ruff clean; `git diff --check` clean.
- Red-proof against `29b22d34`: all 4 new regressions fail there, with the messages in the table.
- Real-engine e2e, serial under the shared lock: **54 passed** across the daemon and live suites.
- `gda schema` byte-identical to `29b22d34`.

Unchanged and still open, as disclosed before: the CLI↔daemon leg reads its frames with no deadline. No caller of that leg promises one, so it costs the serve loop rather than a violated bound, and choosing a value wants its own decision.

qinhaihong-red:
## Fifth re-review — changes still required (`b196c829`)

I rechecked the latest response against the exact current head `b196c829304f436cace1d4f52f15fe9aca964e6a`, replaying only the fourth-review residues and their adjacent deadline/process-group edges.

The direct repairs are present: stale retirement now carries one absolute instant through `close()` / `_terminate()` and recomputes immediately before `wait()`; preparation that exhausts the deadline spawns nothing; a spawn overrun is attributed to the spawn rather than to a missing auth token; the both-stubborn leader/descendant case receives group-wide SIGKILL; and both teardown regressions now enqueue their control request while `wait-ready` is active. Two runtime ownership gaps and two artifact-accuracy residues remain.

1. **[P1, architecture] An expired readiness deadline can still return success after the liveness check.**

   `_ensure_session()` now creates the instant before `self._session.alive()`, but returns `_Established(..., launched=False)` immediately when `alive()` is true (`src/gda/daemon/server.py:428-432`) without checking whether the instant expired during that call.

   A deterministic `SessionHandle` probe used `timeout=0.05s`; `alive()` took `0.12s` and returned true. `wait-ready` still returned **success** `{launched: false}` after **0.125s**. That contradicts the command catalog's rule that an already-spent deadline is a refusal and the new comment that the liveness check draws from this instant.

   Recheck the instant after a true liveness result. If it is spent, return an accurate deadline refusal while retaining the still-valid cached session; do not misreport it as `engine_session_not_running` and do not tear it down merely because the caller's wait expired.

2. **[P2, lifecycle] A leader that obeys SIGTERM still lets an owned descendant escape teardown.**

   `_terminate()` decides whether to escalate solely from the direct leader's `poll()` / `wait()` (`src/gda/daemon/session.py:480-494`). The new regression covers a leader and descendant that both ignore SIGTERM. It misses the asymmetric branch: when the leader obeys the group-wide SIGTERM, `proc.wait()` succeeds and the SIGKILL branch is skipped even if another member of the owned group remains alive.

   A real-process probe produced:

   ```text
   elapsed:          0.001s
   leader:           -15 (SIGTERM)
   descendant:       ignores SIGTERM
   descendant_alive: true
   ```

   Preserve the Engine session's known PGID independently of the leader's liveness and complete group retirement even when the leader has already exited or `wait()` succeeds. Add the asymmetric regression above; the current early `poll() is not None` return needs the same ownership treatment.

3. **[P2, artifacts] The documents still describe a harder wall-clock bound than the implementation provides.**

   The PR body, ADR-0017, command catalog, Skill, and README say the timeout bounds the whole attempt and that spawn is the one uninterruptible step. On this head, a `0.12s` Session-log write under a `0.05s` timeout correctly prevents the spawn, but the refusal still returns after **0.125s**; the latest response likewise acknowledges that a slow `poll()` can overrun its instant.

   Either enforce that hard bound, or consistently describe the timeout as one admission/timed-wait budget: no renewed grace, timed waits use the remainder, and expiry prevents later launch/success decisions, while synchronous liveness, filesystem, display/path, and spawn calls can delay when expiry is observed. In particular, `_terminate()` should not say it “never returns later than deadline”; the property now proved is that it adds no fresh blocking grace after the deadline.

4. **[P2, artifacts] Best-effort collection is still presented as guaranteed elsewhere in the PR body.**

   The Bound section and ADR correctly call the daemon-thread reaper best-effort, but the regression inventory still says “the child is collected anyway,” and Validation says the regression “pins the property.” The normal-path test proves that its child is collected; it cannot guarantee collection when the daemon thread is unjoined and deliberately swallows `wait()` failures. Describe it as an ordinary-path regression and attribute cleanup after daemon exit to OS reparenting/reaping rather than saying the daemon's exit itself clears the entry.

### Four-axis disposition

- **Artifact Review — changes required:** the hard-bound wording and best-effort/guarantee contradiction remain. Terminology and ASD-STE100 prose otherwise pass; the concurrent-control inventory is now accurate.
- **Domain-modular architecture — fail:** one P1 expired-deadline success path and one P2 process-group ownership gap remain. Deadline representation, phase ownership, and signal fallback are otherwise coherent.
- **Standards — one hard/P2 plus one judgement call:** the group lifecycle is incomplete against the newly documented ownership. As a non-blocking Primitive Obsession smell, the repeatedly confused duration/instant concept still crosses layers as a bare `float`; a small `MonotonicDeadline` value type could make the distinction structural.
- **Spec — pass:** issue #657's lazy-launch bounded-wait command, lifecycle documentation, and `daemon start → wait-ready → first diag errors` acceptance path remain implemented. It does not require hard-real-time interruption of arbitrary synchronous system calls.

### Recheck validation

- New-fix regressions: **6 passed**
- Related daemon suites: **139 passed**
- Fast non-e2e suite: **1,742 passed, 2 skipped, 499 deselected**
- Real-Godot readiness acceptance: **1 passed**
- Ruff check/format, pyright, and `git diff --check`: passed
- Current required GitHub checks are green; the GitHub Godot e2e job is skipped
- I did not independently rerun the author's full 54-test real-engine matrix

The four reported mechanics are substantially improved, but the readiness boundary can still certify success after its instant and the Engine-session owner can still leave a group member running. Requesting changes.

qinhaihong-red:
## Fifth re-review response — head `dc04d62b`: three adopted, one declined with evidence

Baseline: base tip `3bfc9ae4`, merge base `3bfc9ae4`, reviewed head `b196c829`, new head `dc04d62b`.

| --- | --- | --- | --- |
| [P1] An expired deadline can still return success after `alive()` | **Partially adopt** — doc fixed, refusal declined | `alive()` is one `waitpid(WNOHANG)`; the probe's 0.12s comes from a fake. The postcondition ("subsequent live reads serve") **holds** when it returns true | The overstated rule is corrected (finding 3); the cached-session answer is unchanged |
| [P2] A leader that obeys `SIGTERM` lets an owned descendant escape | **Fully adopt** | Reproduced: leader `-15` in 0.001s, descendant alive | The GROUP is retired once the leader is gone, whatever ended it — including when it had already exited |
| [P2] The documents promise a harder bound than the code provides | **Fully adopt** | Your two examples, plus my own admission about a slow `poll()` | Restated as a waiting/committing budget with three provable rules |
| [P2] Best-effort collection is presented as guaranteed in the inventory | **Fully adopt** | — | Inventory and Validation reworded; reparenting attributed correctly |

### 1 — why the cached-session answer stays a success

The mechanism is real: `_ensure_session` does not recheck the instant after a true `alive()`. I am not turning that into a refusal, for three reasons.

**It would report a failure that is not one.** `wait-ready`'s documented postcondition is that subsequent live reads serve. When `alive()` returns true, that postcondition holds — the session is up and its channel is not stale. Refusing would hand an agent an error code for a session that is fine, and the sensible agent reaction to a readiness failure is to retry or tear down.

**It needs new public ABI for an unreachable path.** No registered code fits: `engine_session_not_running` is false, and `live_timeout` now means "the session was discarded". A new registered code is an ADR-0002 table row and a permanent public contract — bought for a case that requires `EngineSession.alive()`, a single `waitpid(WNOHANG)`, to outlast the caller's entire budget. Your probe reaches it with a `SessionHandle` fake whose `alive()` sleeps; the real one cannot.

**The contradiction it names is in the prose, not the code** — which is finding 3, and I have fixed it there. The catalog said "a deadline already spent is a refusal". That was too strong: what is true, and what the code does, is that a spent instant means the boundary **commits to nothing further** — it launches nothing and tears down anything it had committed to. Answering with a session that already exists commits to nothing.

Residual risk, stated plainly: if `alive()` ever becomes something that can block (a liveness probe over the channel, say), this becomes a real late-success path and the recheck you describe becomes the right fix. It is not one today.

### 2 — the group, not the leader

Adopted as diagnosed, including the branch I had missed and the early-return one. The group id is read up front, from the leader while it still holds its pid; the leader's fate no longer decides the group's, so the group is signalled `SIGKILL` once the leader is gone — by its own exit, by the `SIGTERM`, or before gda arrived. gda's own group is never a target (a guard, since the escalation now runs on the ordinary path too).

The new regression is your asymmetric case — a leader that obeys, a descendant that does not — and it is red on `b196c829` with "the descendant outlived the group retirement". The both-stubborn one stays.

One side effect worth naming: the test stand-in for an engine process now *refuses* to invent a pid. Teardown reads the pid to find the group, and a made-up one resolves to whatever real process happens to hold it — which a test would then signal. A test that lets a stand-in reach the real teardown says so with `no_engine_teardown`.

### 3 — the bound, described as what it is

Adopted in full, in the ADR, the catalog and the PR body. The timeout is a budget for **waiting and for committing to new work**, not a hard wall clock; a hard wall clock would need threads and cancellation around every synchronous call, which is not a trade this daemon should make for a readiness helper. Three rules replace it, each of which the regressions prove: no phase gets a fresh grace; every timed wait uses what remains, computed where the wait begins; once the instant is spent, nothing further is committed. A slow synchronous step delays *when* expiry is observed and never earns a new budget. `_terminate`'s docstring no longer claims it "never returns later than deadline" — it claims it starts no fresh blocking wait after it, which is what is proved.

### 4 — the inventory

"The child is collected anyway" is now "on the ordinary path the child is collected", and the Validation section says explicitly that no test can prove collection in general while the reaper is unjoined and swallows failures. The process-table entry is attributed to reparenting at daemon exit rather than to the exit itself.

### Not adopted: `MonotonicDeadline`

The duration/instant confusion was real and cost several rounds, but it is now structurally gone: there is exactly one representation in this path, every parameter that carries it is named `deadline`, and the two places that mint one do it from `TERMINATE_GRACE` / `CONNECT_TIMEOUT` at the moment they start. A wrapper type would add a class, conversions at the socket and `subprocess` boundaries (both of which want a float duration), and no guarantee the naming does not already give. If a third representation ever appears, that is the trigger.

### Validation at `dc04d62b`

- Fast suite **1745 passed**; pyright 0 errors; ruff clean; `git diff --check` clean.
- Red-proof against `b196c829`: the new group regression fails there; the pre-existing both-stubborn one passes, as expected.
- Real-engine e2e, serial under the shared lock: **54 passed** across the daemon and live suites.
- `gda schema` byte-identical to `b196c829`.

Unchanged and still open, as disclosed since the third round: the CLI↔daemon leg reads its frames with no deadline. No caller of that leg promises one.

qinhaihong-red:
## Recheck — changes requested at `dc04d62b`

Thanks for the latest fixes and detailed response. I replayed the unresolved findings and adjacent lifecycle boundaries against exact head `dc04d62bd24c8a3e5f090707dd00a826e85e37a6`. The deadline plumbing, pre-spawn expiry gate, ordinary asymmetric group teardown, and best-effort collection wording are now closed. The Spec review also passes.

Two P2 issues remain.

### [P2] Preserve owned process-group identity and wait for the group, not only its leader

`_terminate()` still derives the PGID from `proc.pid` during teardown and treats leader completion as process-group completion. Two real-process paths disconfirm the intended group-retirement contract:

1. A session leader spawned a SIGTERM-ignoring descendant and exited normally. The production ordering then called `Popen.poll()` (as `EngineSession.alive()` does), which reaped the leader before teardown. `os.getpgid(proc.pid)` could no longer recover the group, and the descendant remained alive:

   ```text
   leader_returncode=0, recovered_group_at_teardown=null, descendant_alive=true
   ```

2. A leader and descendant shared a `start_new_session=True` group. The leader obeyed SIGTERM; the descendant's SIGTERM handler needed 0.20s to finish cleanup. With a 1.0s absolute deadline, `_terminate()` returned after 0.001s and the cleanup marker was absent because it sent SIGKILL as soon as the leader exited:

   ```text
   elapsed=0.001, leader_returncode=-15, descendant_cleanup_completed=false
   ```

This contradicts the documented rule that the leader's fate does not decide the group's fate and that escalation uses the remaining deadline.

Smallest owner-safe fix:

- Capture and validate the owned PGID immediately after `Popen(start_new_session=True)` succeeds, including failed-launch teardown paths, and carry it with the Engine-session owner. Do not rediscover ownership from a leader that may already have been reaped or whose PID may have been reused.
- After group SIGTERM, use the same absolute deadline to wait for the *group* to become empty. Reap the leader independently. Send group SIGKILL only when the group still exists at the deadline.
- Add production-order regressions for `poll()/reaped leader + stubborn descendant` and `leader obeys + descendant completes cleanup within the remaining deadline`.

### [P2] Synchronize the public timeout contract with the narrowed ADR semantics

ADR-0017 and the command catalog now correctly describe `--timeout` as a budget for waiting and committing new work, not as a hard wall-clock bound. Other public surfaces still advertise the stronger guarantee:

- the PR body says the 60-second client-side bound stays above the whole operation;
- rendered help still says "wait, bounded";
- rendered schema says timeout bounds the whole readiness handshake and that the 60-second live-channel bound covers the whole request round trip;
- the Skill, README, and translations retain unqualified bounded-wait wording.

Exact-head probes still show the distinction: a `.05s` timeout plus a `.12s` synchronous log write returns at approximately `.125s`, while a `.12s` liveness check can succeed at approximately `.125s`. The live client also performs `read_message()` without an absolute request deadline, so the socket's 60 seconds is an inactivity/per-receive timeout rather than a whole-round-trip ceiling.

Please either synchronize every public surface with the ADR's wait/commit-budget definition and describe the 60-second timeout accurately, or implement the harder bounds currently advertised.

### Minor wording

The phrase "introduce threads and cancellation to pretend otherwise" is argumentative. A neutral observable formulation would be clearer, for example: "A hard wall-clock limit would require cancellation around synchronous calls. This change does not add that mechanism."

Validation on the exact head:

- focused regressions: `7 passed`
- related daemon suites: `140 passed`
- fast suite: `1743 passed, 2 skipped, 499 deselected`
- Ruff and Pyright: pass
- real Godot acceptance: `1 passed`
- required CI checks: green; Godot e2e is skipped

qinhaihong-red:
## Recheck response — both P2s and the wording fixed at head `847b8bd0`

Baseline: base tip `3bfc9ae4`, merge base `3bfc9ae4`, reviewed head `dc04d62b`, new head `847b8bd0`. Both mechanics were reproduced first.

| Claim | Decision | Reproduced on `dc04d62b` | Change |
| --- | --- | --- | --- |
| [P2] Group identity is rediscovered from a leader that may be reaped; the leader's completion is treated as the group's | **Fully adopt** | `leader_returncode=0, recovered_group=None, descendant_alive=true`; and a descendant's 0.2s `SIGTERM` cleanup cut short at 0.001s with 1.0s left | The owner captures the group at spawn and keeps it; teardown waits for the GROUP to empty within the same deadline |
| [P2] Public surfaces still advertise a harder bound | **Fully adopt — by implementing the harder one where it was cheap** | The client's 60s was a per-recv timeout, as you said | The live client now uses an absolute round-trip instant; the one description that implied a hard wall clock is corrected |
| Minor: argumentative phrasing | **Fully adopt** | — | Replaced with your formulation |

### 1 — the session owns its group

Both directions of your finding were right, and the second one was a defect I introduced in the previous round.

**Identity.** `_own_group()` is called once, immediately after `Popen(start_new_session=True)`, and the id is carried on `EngineSession` (and through `launch_session`'s failure teardowns). Teardown is told the group; it no longer guesses. Your sequence is now a regression: a leader that exits at once leaving a stubborn descendant, `alive()` polled until it reaps the leader, `_own_group()` asserted to return `None` at that point — and `close()` still retires the descendant.

**Completion.** Teardown waits for the group to empty, polling within the same deadline, and escalates only a group still standing when the deadline arrives. That fixes both edges at once: a descendant handling `SIGTERM` gets to finish (the second regression asserts its cleanup marker exists), and one that ignores `SIGTERM` still cannot outlive the session (the two earlier group regressions stay green). The leader is reaped by the same poll, so its `returncode` stays honest.

`_terminate` keeps a best-effort fallback for a caller that has no owner to ask — the tests, and daemon shutdown — and says so.

### 2 — the client's ceiling, implemented rather than qualified

You were right that `LIVE_REQUEST_TIMEOUT` was a per-receive inactivity timeout: one `settimeout` and a `read_message` that loops `recv`. Since `read_frame` already takes an absolute deadline, making the claim true cost three lines, so I did that instead of weakening the wording — a trickled reply now yields `live_timeout` inside the ceiling, with a regression that holds it to it.

That also closes the residual I had been disclosing since the third round, for the client half. The daemon's own read of a CLI request is still deadline-free; no caller of that leg promises a bound, so it costs the serve loop rather than a published guarantee. Unchanged, and still disclosed.

On the remaining surfaces: with the client fixed, the only statement that overreached was `wait-ready`'s own timeout description, which said it bounds "the launch's whole readiness handshake". It now says what the budget covers — every wait draws from one instant, and none is renewed. That is the **single** intended `gda schema` change in this slice beyond the new command itself; every other command's entry is byte-identical. The help's "wait, bounded" and the Skill/README "bounded by `--timeout`" are left as they are: under the ADR's definition the *waits* are bounded, which is what those sentences say.

### 3 — wording

Taken verbatim: "A hard wall-clock limit would require cancellation around synchronous calls. This change does not add that mechanism." You are right that the previous phrasing argued rather than described.

### Validation at `847b8bd0`

- Fast suite **1748 passed**; pyright 0 errors; ruff clean; `git diff --check` clean.
- Red-proof against `dc04d62b`: all 3 new regressions fail there — "the descendant outlived the group retirement", "the descendant's cleanup was cut short", and the CLI held past its ceiling.
- Real-engine e2e, serial under the shared lock: **59 passed** across the daemon and live suites, `screen` included this round since the client's request leg changed.
- `gda schema`: one intended entry changed (`daemon wait-ready` input description), verified by diffing every command entry against the previous head.

PR #732, opened 2026-08-24
feat(gda): complete activation gestures and guard the viewport-enter notify (#647, #652)

### LINKED ISSUE #647: Avoid duplicate viewport-enter warnings during injected mouse input
## Background

Repeated `gda input mouse-click` events in a windowed daemon session complete successfully, but
the harness reports one engine warning after the viewport has already received its first mouse
event. This pollutes `gda diag errors` during a product playtest even though the game has no runtime
error.

Observed with:

- gda 0.9.0
- Godot 4.6.3 stable
- macOS
- a windowed daemon session

## Reproduction

```bash
gda daemon start --windowed --project /path/to/project --json
gda input mouse-click 576 517 --project /path/to/project --json
gda input mouse-click 576 517 --project /path/to/project --json
gda diag errors --project /path/to/project --json
```

The click results are successful. `diag errors` contains:

```text
The Viewport was previously notified that the mouse is in its area.
There is no need to notify it at this time.
```

The call stack points to:

```text
_prepare_mouse_input (res://addons/gda_harness/gda_harness.gd:732)
_push_mouse_click (res://addons/gda_harness/gda_harness.gd:791)
```

The same warning is produced for each later click in `gda input sequence`.

## Expected

Repeated injected mouse events must not call `Viewport.notify_mouse_entered()` after the viewport
is already in the entered state. A successful mouse automation path should not add harness-owned
warnings to `gda diag errors`.

## Impact

The gameplay input works, but a consumer cannot use an empty `diag errors` result as clean runtime
evidence after a multi-click playtest. The warning is owned by the development harness, not by the
tested game.

### LINKED ISSUE #652: input: mouse-click and single-frame key injection report success without completing the activation
## Dogfooding source

- **GDA-DF-004** (2026-08-08) — a successful `gda input mouse-click` left a `Button` pressed and never emitted `pressed`; an explicit press+release `input sequence` activated it.
- **GDA-DF-016** (2026-08-09) — a click at the center of a visible enabled `Button` returned structured success while the controller phase did not change.
- **GDA-DF-034** (2026-08-11) — a key press and release contained in one immediate frame reported success but did not advance the focused UI state; separating press / process frames / release worked repeatably.

## What to build

Injected-input commands must perform the complete activation gesture their name implies:

- Make `mouse-click` synthesize the full gesture — initial move, press, release — across frames. The alternative (document it as press-only) is rejected by this issue: GDA-DF-004's feedback records that the command name and the `kind: mouse_click` result imply a complete click.
- Provide a high-level action/key **tap**. Definition: press, hold for N process frames, release, then M process frames; N and M configurable with safe defaults.
- Report the injected event phases in the structured result, plus before/after focus or handled-state evidence where the engine exposes it.

## Acceptance criteria

- [ ] `mouse-click` on a standard enabled `Button` in an `Engine session` emits `pressed` and the observable app state changes; the injected gesture includes the initial move.
- [ ] A single tap advances a focus-driven UI exactly once, repeatably, under the definition above.
- [ ] The structured result reports the injected event phases (and before/after focus where available).
- [ ] Help and `--schema` state the minimum frame semantics Godot needs for UI activation.
- [ ] Live regressions cover a standard Button click and a focus-navigation tap.

## Blocked by

None - can start immediately.

qinhaihong-red:
## W4 slice: input activation gestures (#647 + #652)

One slice for both issues because they share the harness mouse path: the incomplete click (#652) and the viewport-enter warning it triggers on every repeat (#647).

Both defects were **reproduced against a live engine on the base tip `b6e9760c` before any change**:

- Two "successful" `gda input mouse-click` calls on a standard enabled `Button` left `pressed_count == 0`, `down_count == 1`, `up_count == 0` — the press was injected, the release never was, the button stayed held forever and `pressed` never fired (GDA-DF-004).
- The second click put the exact #647 warning into `gda diag errors`, callstack `_prepare_mouse_input` → `_push_mouse_click`.
- A clean-state probe also answered a design question: a **same-frame** press+release pair fully activates a `Button` (pressed/down/up each +1), so a sequence `mouse_click` event can honor its documented "whole click at one clock offset" contract without splitting frames.

### What changed

**`mouse-click` is now the complete gesture** (#652). The op runs on the #223 multi-frame base: the initial move at window frame 0, the press at 1, the release at 2. The result reports the injected `phases` and the focused Control before/after the gesture (`focus_before` / `focus_after`) — the activation evidence the engine exposes. A sequence `mouse_click` event now pushes press **and** release (same frame); the press-only `_push_mouse_click` helper is gone.

**New `gda input tap`** (#652, GDA-DF-034): the press-hold-release of exactly one key or one InputMap action — press at frame 0, hold `--hold-frames` (default 2, ≥ 1) process frames, release, then `--settle-frames` (default 2) more frames so the game observes the release before the op returns. Godot does not advance a focused UI when the pair lands on one immediate frame, and `hold_frames >= 1` encodes that floor. Exactly-one-target, the per-family fields (modifiers = key only, strength = action only, refused by name per the GDA-DF-037 lesson), and the window ceiling (`hold + settle + 1 <= 600`) are all model-side (ADR-0015). The two live-only failures stay deferred to the harness (`live_invalid_key`, `live_unknown_action`).

**The viewport-enter notify is edge-guarded** (#647). The engine keeps `gui.mouse_in_viewport` private, but the root `Window` emits `mouse_entered`/`mouse_exited` on the same notifications that flip it (engine source: `viewport.cpp` NOTIFICATION_VP_MOUSE_ENTER/EXIT; `window.cpp` emits on both), so the harness mirrors the state via two signal connections (inside the daemon-launched gate, ADR-0018 inertness untouched) and notifies only on a real edge. This also covers a windowed session where the OS mouse genuinely enters (no notify at all) or leaves (the next injection re-arms).

### Public-surface impact

`gda schema` diff against base, verified entry-by-entry: **added** `input tap`; **changed** `input mouse-click` (params + result carry the gesture contract); `input mouse-move` and `input sequence` change **descriptions only** (the shared result model was split, the `mouse_click` event kind's docstring now states the same-frame fact). Docs synced: command catalog, SKILL.md (surface table + a "use the gesture commands for activation" guidance block), all four READMEs (+ i18n re-stamp). The #652 acceptance "help and `--schema` state the minimum frame semantics" lands in the params docstrings and command help, pinned by two help tests.

### Validation

- Fast suite **1762 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: the input suite **15 passed** (5 new: complete-gesture click with `pressed`/`down`/`up` counters and the initial move observed; repeated mouse ops leave `diag errors` free of the harness-owned warning; a sequence `mouse_click` activates a Button; a key tap advances a 3-button focus UI **exactly once, twice in a row** (A→B then B→C, via the op's own focus read-back); an action tap produces exactly one press edge and one release edge, twice in a row). Adjacent suites re-run: harness-install / game / daemon / diag **35 passed**.
- **Red-proof on base `b6e9760c`**: all 10 new/changed unit tests and all 5 new e2e regressions fail there, each with the failure the fix removes (`KeyError: 'phases'`; the #647 warning present; `pressed_count 0 == 1`; `unknown_command` for tap). One note: the `hold-frames 0` usage test originally passed on base for the wrong reason (unknown command is also exit 2); it now also asserts the error names `--hold-frames`, and is red on base.

Closes #647
Closes #652

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,9 +581,10 @@ edge-triggered controls.
 | Command | What it does |
 | ------- | ------------ |
 | `input key` | Inject a key event (with modifiers). |
-| `input mouse-click` | Inject a mouse click at `(x, y)`. |
+| `input mouse-click` | Inject a complete click gesture (move, press, release) at `(x, y)`. |
 | `input mouse-move` | Inject mouse motion to `(x, y)`. |
 | `input action` | Press/release a mapped input action. |
+| `input tap` | Tap one key or action: press, hold, release across frames. |
 | `input sequence` | Inject a multi-frame event timeline. |
 
 Mouse events report the injected viewport coordinate through `event.position`.
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=8cb5ae07a2d875db3554b7840b6fc90efdca5fa1bd84da220a2de393f6871949 -->
+<!-- gda-readme-i18n: source=README.md sha256=9412ce784d9839d437daa861f8b097a788f8e401dbb36130c2180b323b78350f -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -595,9 +595,10 @@ variables de script getter-only/no-op o controles edge-triggered.
 | Comando | Qué hace |
 | ------- | ------------ |
 | `input key` | Inyecta un evento de tecla (con modificadores). |
-| `input mouse-click` | Inyecta un clic de ratón en `(x, y)`. |
+| `input mouse-click` | Inyecta el gesto de clic completo (movimiento, pulsación, liberación) en `(x, y)`. |
 | `input mouse-move` | Inyecta un movimiento de ratón hacia `(x, y)`. |
 | `input action` | Presiona/suelta una acción de entrada mapeada. |
+| `input tap` | Toca una tecla o acción: pulsa, mantiene y suelta a lo largo de varios frames. |
 | `input sequence` | Inyecta una línea de tiempo de eventos de varios frames. |
 
 Los eventos de ratón informan la coordenada inyectada del viewport mediante
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=8cb5ae07a2d875db3554b7840b6fc90efdca5fa1bd84da220a2de393f6871949 -->
+<!-- gda-readme-i18n: source=README.md sha256=9412ce784d9839d437daa861f8b097a788f8e401dbb36130c2180b323b78350f -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -590,9 +590,10 @@ Live の `game set --property position` は `node set` と同じ `Control` ポ�
 | コマンド | 機能 |
 | ------- | ------------ |
 | `input key` | キーイベントを(修飾キー付きで)注入します。 |
-| `input mouse-click` | `(x, y)` の位置にマウスクリックを注入します。 |
+| `input mouse-click` | `(x, y)` の位置に完全なクリックジェスチャ(移動、押下、解放)を注入します。 |
 | `input mouse-move` | `(x, y)` へのマウス移動を注入します。 |
 | `input action` | マッピング済みの入力アクションを押下/解放します。 |
+| `input tap` | キーまたはアクションを 1 回タップします(押下、保持、解放を複数フレームで実行)。 |
 | `input sequence` | 複数フレームにわたるイベントのタイムラインを注入します。 |
 
 マウスイベントは、注入されたビューポート座標を `event.position` で報告します。
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -565,7 +565,7 @@ Live `game set --property position` 遵循与 `node set` 相同的 `Control` 策
 | `input mouse-click` | 在 `(x, y)` 处注入完整的点击手势(移动、按下、释放)。 |
 | `input mouse-move` | 将鼠标移动到 `(x, y)`。 |
 | `input action` | 按下/释放一个已映射的输入动作。 |
-| `input tap` | 轻按一个按键或动作:跨帧完成按下、保持、释放。 |
+| `input tap` | 轻按一个按键或动作：跨帧完成按下、保持、释放。 |
 | `input sequence` | 注入一条跨多帧的事件时间线。 |
 
 鼠标事件会在 `event.position` 中携带注入的视口坐标。Godot 在 daemon 会话中可能让
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=8cb5ae07a2d875db3554b7840b6fc90efdca5fa1bd84da220a2de393f6871949 -->
+<!-- gda-readme-i18n: source=README.md sha256=9412ce784d9839d437daa861f8b097a788f8e401dbb36130c2180b323b78350f -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -562,9 +562,10 @@ Live `game set --property position` 遵循与 `node set` 相同的 `Control` 策
 | 命令 | 作用 |
 | ------- | ------------ |
 | `input key` | 注入一个按键事件（带修饰键）。 |
-| `input mouse-click` | 在 `(x, y)` 处注入一次鼠标点击。 |
+| `input mouse-click` | 在 `(x, y)` 处注入完整的点击手势(移动、按下、释放)。 |
 | `input mouse-move` | 将鼠标移动到 `(x, y)`。 |
 | `input action` | 按下/释放一个已映射的输入动作。 |
+| `input tap` | 轻按一个按键或动作:跨帧完成按下、保持、释放。 |
 | `input sequence` | 注入一条跨多帧的事件时间线。 |
 
 鼠标事件会在 `event.position` 中携带注入的视口坐标。Godot 在 daemon 会话中可能让
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -764,12 +764,34 @@ headless is unaffected (4.4+, cross-platform).
   (ADR-0019).
 - **`input` (input simulation):** runtime input injection into the running game
   (shipped, #221). Single-frame ops `input key <KEY> [--modifiers …] [--released]`,
-  `input mouse-click <x> <y> [--button left|right|middle] [--double]` /
   `input mouse-move <x> <y>`, and `input action <NAME> [--release] [--strength F]`
-  each inject one event at a frame boundary (ADR-0020); the multi-frame
+  each inject one event at a frame boundary (ADR-0020). The **activation
+  gestures** (#652) are multi-frame: `input mouse-click <x> <y> [--button
+  left|right|middle] [--double]` injects the COMPLETE click its name implies —
+  the initial move, the press, and the release, one per process frame across a
+  3-frame window — because Godot's UI activates on the release (a bare press
+  never emits a default `Button`'s `pressed` and leaves it held down,
+  GDA-DF-004); and `input tap (--key K [--modifiers …] | --action NAME
+  [--strength F]) [--hold-frames N] [--settle-frames M]` performs the complete
+  press-hold-release of one key or one InputMap action — press at window frame
+  0, release after N (default 2, at least 1) process frames, then M (default 2)
+  settle frames so the game observes the release before the op returns — because
+  a press/release pair contained in one immediate frame reports success without
+  advancing a focused UI (GDA-DF-034). Both gesture results report the injected
+  `phases` (each phase's window frame) and the focused Control's runtime path
+  before/after the gesture (`focus_before` / `focus_after`, null when nothing
+  holds focus) — the activation evidence the engine exposes. Repeated injected
+  mouse events add no harness-owned warnings to `diag errors` (#647): the
+  harness notifies the viewport's mouse-enter state only on a real edge,
+  mirroring it from the root Window's `mouse_entered`/`mouse_exited` signals.
+  The multi-frame
   `input sequence --events <JSON>` applies a list of events across one selected
   clock and returns as one blocking payload, on the gda harness's time-windowed
-  multi-frame base (the same base `perf monitor` uses, #223). Existing sequence
+  multi-frame base (the same base `perf monitor` uses, #223). A sequence
+  `mouse_click` event is a whole click at ONE clock offset — the harness pushes
+  the press and then the release on the same frame, which fully activates a
+  default `Button` (mouse activation, unlike a focused-UI key tap, does not need
+  the pair split across frames). Existing sequence
   event `frame` offsets are explicitly the harness/process-frame clock advanced by
   the harness `_process` loop; they are preserved for compatibility and are **not**
   Godot's fixed physics frames. For deterministic simulation-duration input, use
@@ -791,7 +813,9 @@ headless is unaffected (4.4+, cross-platform).
   leave `Viewport.get_mouse_position()` and `Node2D.get_global_mouse_position()`
   stale in daemon sessions, so game code should read the injected coordinate from
   the input event. The modifier
-  set, mouse-button enum, action strength range (0..1), per-event shape, and the
+  set, mouse-button enum, action strength range (0..1), per-event shape, the
+  tap's exactly-one-target rule and window (`hold_frames + settle_frames + 1` ≤
+  the per-window ceiling), and the
   sequence's selected-clock window (`max(frame)+1` or `max(physics_frame)+1` ≤ the
   per-window ceiling, the same bound `perf monitor` enforces, #223) are bounded
   **model-side** (ADR-0015), so an
--- a/src/gda/commands/input.py
+++ b/src/gda/commands/input.py
@@ -208,10 +208,12 @@ class InputMouseMoveResult(BaseModel):
         default="mouse_move", description="The injected event kind ('mouse_move')."
     )
     position: list[float] = Field(
+        min_length=2,
+        max_length=2,
         description=(
             "The viewport position the event was injected at, as [x, y]. This "
             "mirrors event.position; engine-tracked mouse positions may remain stale."
-        )
+        ),
     )
     button: str | None = Field(
         default=None, description="Always null for a move (a historical field)."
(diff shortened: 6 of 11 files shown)


qinhaihong-red:

Reviewed exact head `fb5e0091a7543eafc30d8609c44b7ef16ea9df0e` against base `b6e9760c6422132a613ff44bf0370bfa89243c63`. The Artifact, Architecture, Standards, and Spec axes below were run independently. No cross-axis reranking has been applied.

### Artifact review — Markdown

#### Required

1. **`src/gda/skill/SKILL.md:187` gives an incomplete command.**

   The guidance says `gda input key --released`, but the public syntax is `gda input key [OPTIONS] KEY`. Running the documented command at this head exits 2 with `Missing argument 'KEY'`.

   Smallest repair: use `gda input key <KEY> --released`. A complete sentence could be: “Use `gda input key <KEY> --released` or explicit `mouse_button` sequence phases when you must construct individual press/release edges, such as a hold or drag.”

#### Minor

1. **`docs/README.zh-CN.md:568` uses inconsistent Chinese punctuation.**

   Change `轻按一个按键或动作:跨帧完成按下、保持、释放。` to `轻按一个按键或动作：跨帧完成按下、保持、释放。`

The required ASD-STE100 terminology/prose pass found no other substantive terminology or prose drift across the six changed Markdown artifacts.

#### [P2] The result models do not validate the gesture evidence they claim to provide

Locations: `src/gda/commands/input.py:223-280,420-476`.

`InputEventPhase` accepts any integer and string. `InputMouseClickResult` and `InputTapResult` also do not enforce the fixed `kind`, legal button/phase values, mutually exclusive key/action result families, `frames == hold_frames + settle_frames + 1`, or the required phase positions and order.

This was reproduced through the real success classifier: `classify_live` accepts a tap with `kind="wrong"`, both key and action, `frames=1`, and phase `{-7, "noop"}`; it also accepts a click with `kind="mouse_move"`, `button="bogus"`, and an arbitrary phase. These payloads become successful CLI/MCP results instead of `contract_violation`, so the output model does not protect #652's new activation-evidence contract.

Smallest repair:

- constrain `kind`, button, and phase with `Literal`/enums and bound frame values;
- validate target-family exclusivity, frame totals, and exact phase positions/order in the click/tap result models;
- add malformed live-sentinel regressions that must classify as `contract_violation`.

#### [P2] The action-tap strength default is still owned by the harness

Locations: `src/gda/commands/input.py:367-417`, `src/gda/harness/gda_harness.gd:1010`.

At this head:

```text
InputTapParams(action="jump").model_dump()["strength"] == None
```

The harness then converts that value to `1.0` through `_float_param(..., 1.0)`. The generated input schema consequently says `default: null`, while its description and runtime behavior promise `1.0`. Default knowledge is split across the Pydantic description, Python params, and GDScript fallback, contrary to ADR-0015's requirement that normalization and derived defaults live in the params model.

Smallest repair: normalize an omitted action-tap strength to `1.0` in `InputTapParams`, retain the harness value only as a defensive fallback, test that argv and `--params-json` both send `strength: 1.0` to the runner, and make the conditional-default schema text agree with the normalized model.

The remaining structure passes the completeness/orthogonality/extensibility review: `gda.commands.input` remains the inbound vertical slice, the harness remains the concrete live-engine adapter, source dependencies remain acyclic, and the time-windowed operation retains single-writer ownership. No new bounded context, port hierarchy, event bus, or plugin abstraction is warranted.

### Standards

No hard documented-standard violation was found. Three Fowler-baseline observations remain **judgement calls**:

1. **Mysterious Name — `src/gda/commands/input.py:197`.** `InputMouseResult` now represents only `mouse-move`; `InputMouseMoveResult` and `render_input_mouse_move` would reveal that responsibility.
2. **Primitive Obsession — `src/gda/commands/input.py:223-236`.** `phase: str` represents a closed `move | press | release` vocabulary. A `Literal` or enum would prevent protocol drift. This concern is also covered by the first architecture finding.
3. **Data Clumps / Duplicated Code — `src/gda/commands/input.py:261-279,457-475,948-965`.** Click and tap repeat the phase/focus field cluster and rendering shape. A small shared activation-evidence model/helper may give that knowledge one owner, but this should be weighed against the repository rule against speculative abstraction.

### Spec

#### [P2] #647's required regression is only partially covered

Issue #647 requires “a two-click regression asserting an empty warning stream,” and its reproduction starts the daemon with `--windowed`. `tests/test_e2e_input.py:880-897` instead starts the default daemon, adds a third mouse-move, and only filters for the specific text `previously notified`.

The current test can pass with other harness-owned warnings and does not exercise the specified windowed path. Smallest repair: start with `--windowed`, perform exactly two clicks, and assert `errors == []`.

No other Spec finding was found; the generated help/schema expose the requested frame semantics, and the key/action/click behavior is present.

### Validation

- Fast suite: `1760 passed, 2 skipped, 504 deselected`
- `tests/test_input_commands.py`: `67 passed`
- Five affected real-engine Godot 4.6.3 E2E regressions: `5 passed`
- Rendered `input tap` / `input mouse-click` help and schemas inspected directly
- `git diff --check`: pass
- Current GitHub non-Godot checks: pass; the GitHub `Godot e2e tests` job is skipped

The local real-engine runs validate the default daemon paths, but the windowed OS mouse enter/exit branch remains unverified in this review environment.

qinhaihong-red:
## Review response — all six findings addressed at head `0847e340`

Baseline: base tip `b6e9760c`, merge base `b6e9760c`, reviewed head `fb5e0091`, new head `0847e340`. Each mechanical claim was reproduced before anything changed; the four new unit regressions are red on `fb5e0091` and both empty-stream e2e twins are red on `b6e9760c`.

| --- | --- | --- | --- |
| [Required] SKILL.md documents an incomplete command | **Fully adopt** | `gda input key --released` exits 2, as you said | `gda input key <KEY> --released` |
| [Minor] zh-CN half-width colon | **Fully adopt** | 34 full-width vs 2 half-width after CJK in the file | Full-width colon; i18n stamps unchanged (the English source did not move) |
| [P2] The result models do not validate the gesture evidence | **Fully adopt** (scoped to this slice's models) | Your `classify_live` reproduction; all four malformed replies passed as successes on `fb5e0091` | Literals/enums + cross-field validators; malformed replies now classify as `contract_violation`, pinned by regressions |
| [P2] The action-tap strength default is owned by the harness | **Fully adopt** | `InputTapParams(action="jump").model_dump()["strength"] == None` on `fb5e0091` | The model normalizes an omitted action-tap strength to 1.0; both paths send it explicitly |
| Standards (3 judgement calls) | **Adopt 2, decline 1** | — | Rename done; phase Literal covered above; the shared activation-evidence model declined (reasoning below) |
| [P2] The #647 regression is partial | **Fully adopt, one premise corrected** | The old test tolerated other warnings and skipped the windowed path | Exactly two clicks, `errors == []`, as a headless twin AND a `--windowed` twin behind the display gate |

### 1 — the gesture evidence is now validated, not described

Adopted as your smallest repair, on the three models this slice introduced or split: `phase` is `Literal["move","press","release"]` with `frame >= 0`; `kind` is a Literal on the click, tap, and move results; `button` is `Literal["left","right","middle"]`. Cross-field validators pin what the fields alone cannot: a tap result carries exactly one target family (key+keycode+modifiers XOR action+strength), `frames == hold_frames + settle_frames + 1`, and exactly the phases press@0 / release@hold_frames; a click result reports exactly move@0 → press@1 → release@2. Your four tap payloads and three click payloads (plus a phases-truncated one and a move with a foreign kind) are now regressions that must classify as `contract_violation` — all red on `fb5e0091`, where each passed as a success.

Scope note, disclosed rather than hidden: the sibling result models (`InputKeyResult`, `InputActionResult`, `InputSequenceResult`) keep their loose `kind: str`. They predate this slice and tightening them is the same one-line change per model if you want it, but it changes three published schemas this PR otherwise does not touch, so I left them for a deliberate follow-up rather than widening the diff here.

### 2 — the strength default has one owner

Adopted. `InputTapParams` normalizes an omitted action-tap strength to `1.0` in its validator, so argv and `--params-json` both put an explicit `1.0` on the wire (a regression asserts both paths), and `_float_param(..., 1.0)` in the harness is now a defensive fallback rather than the owner. The declared field default stays `null` for exactly one reason, now stated in the field description: it distinguishes "omitted" from "set", which is what lets a key tap refuse `--strength` by name instead of silently ignoring it (the GDA-DF-037 lesson). A key tap continues to send `strength: null`.

### 3 — standards

- **Mysterious Name**: adopted — `InputMouseMoveResult` / `render_input_mouse_move`. This renames the mouse-move output schema's title; the diff against the reviewed head touches exactly the three ops this round hardens (`mouse-click`, `mouse-move`, `tap`), verified entry-by-entry.
- **Primitive Obsession**: adopted via finding 3 (the `phase` Literal).
- **Data Clumps / Duplicated Code**: **not adopted.** The cluster is three fields (`phases`, `focus_before`, `focus_after`) with two consumers whose invariants diverge — the click pins a fixed 3-phase sequence, the tap pins a 2-phase sequence whose release frame is another field's value — so a shared base would hold only the field declarations while each model keeps its own validator, and the render side already shares `_render_focus`. Two consumers with divergent rules is below the bar this repo has held for extracting a shared model (the same call as the Template-authoring helpers decline in bADR-0017). The trigger to revisit is a third gesture op.

### 4 — the #647 regression, with one premise corrected

Both of your repairs are adopted; one premise deserves the record: issue #647 does not contain the sentence "a two-click regression asserting an empty warning stream" — its Reproduction is two windowed clicks and its Impact clause is that "a consumer cannot use an empty `diag errors` result as clean runtime evidence". That impact clause is exactly why `errors == []` is the right assertion, so the substance of the finding stands.

The regression is now two tests with one body: **exactly two clicks, then `diag errors == []`**. The `--windowed` twin is the issue's verbatim reproduction and exercises the OS-driven mouse enter/exit branch of the signal mirror; it sits behind the repo's display gate (`require_windowed_host`), so it runs on a real display, skips on a capability refusal, and fails loudly in a confined run — your own validation note ("the windowed branch remains unverified in this review environment") is the reason the headless twin exists: it pins the defect mechanism (which is display-independent — reproduced headless before the original fix) in every environment, including this PR's CI and your review host. Both twins are red on base `b6e9760c` with the #647 warning in the stream — the windowed one being the issue's exact path failing for the issue's exact reason.

Also fixed while in the file: nothing. The third mouse-move the old test carried is gone with it; the initial-move path stays covered because every click gesture injects one.

### Validation at `0847e340`

- Fast suite **1766 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: input suite **16 passed**, the `--windowed` twin included (real display on this host).
- Red-proof: the strength-normalization and three malformed-reply regressions fail on `fb5e0091` (`[None, None] == [1.0, 1.0]`; each malformed reply accepted as success); both empty-stream e2e twins fail on `b6e9760c` with the warning present.
- `gda schema` diff vs `fb5e0091`: exactly `input mouse-click`, `input mouse-move`, `input tap` changed; no other entry moved.
- GitHub checks on `0847e340`: green (the `Godot e2e tests` job remains dispatch-only by design).

One pre-existing item observed and left alone: `docs/README.zh-CN.md:7` (`**其他语言:**`) also uses a half-width colon, but that line predates this PR and is outside its diff.

qinhaihong-red:
## Recheck — one architecture issue remains

Rechecked exact head `0847e340fe4463628541d2fcdac81d45576ae96f` against the previous findings and the latest review response. The Markdown, action-strength ownership, Standards, and #647 Spec findings are closed. One adjacent edge of the output-contract finding remains.

### [P2] The result models still accept malformed evidence they claim to validate

Locations: `src/gda/commands/input.py:261-266,469-486,518-547`.

The new validators constrain the gesture kind, phase sequence, target-family presence, and frame arithmetic, but the fields inside that evidence remain looser than their documented success contract:

- `InputMouseClickResult.position` is an unconstrained `list[float]`. With otherwise-valid click phases, `position=[]` passes `classify_live` as a successful `InputMouseClickResult`; the JSON path can therefore emit malformed success evidence, while the human renderer subsequently raises `ValueError: not enough values to unpack (expected 2, got 0)`.
- `InputTapResult` accepts an action-family success with `action=""` and `strength=2.0`. With otherwise-valid frame/phase evidence, `classify_live` again returns success even though the params contract requires a non-empty action and strength in `0..1`.

This leaves the response schema/validator weaker than both the documented result and the successful request contract, so the statement that the activation evidence is “validated, not described” is not yet complete.

Smallest repair: give click positions an exact two-number shape; mirror the successful target constraints on tap results (`key`/`action` non-empty, `strength` in `0..1`, and the applicable key-family constraints); and add malformed live-sentinel regressions proving those payloads become `contract_violation` rather than success.

### Recheck validation

- `tests/test_input_commands.py` + README i18n sync: **76 passed**
- Full real-engine input E2E suite, including the sandbox-external windowed #647 regression: **16 passed**
- `git diff --check`: pass
- Current GitHub non-Godot checks: pass; `Godot e2e tests` remains skipped

The other findings from the first review remain closed. This residual keeps the architecture axis at **changes required**.

qinhaihong-red:
## Recheck response — the residual is closed at head `2b35d580`

Baseline: base tip `b6e9760c`, merge base `b6e9760c`, rechecked head `0847e340`, new head `2b35d580`. Both of your reproductions were confirmed on `0847e340` before the change: `position=[]` classified as a successful click (and the human renderer then died on the `[x, y]` unpack), and an action-family tap with `action=""` or `strength=2.0` classified as success.

| --- | --- | --- | --- |
| [P2] The evidence FIELDS are looser than the documented success contract | **Fully adopt** | Your two payloads pass `classify_live` on `0847e340`; the new regressions are red there | The result models mirror the request-side bounds; eight malformed replies now classify as `contract_violation` |

### What the bounds are now

- **`position`** is an exact two-number pair (`min_length=2, max_length=2`) on `InputMouseClickResult` **and** `InputMouseMoveResult` — the move result had the identical gap (same renderer unpack), so the mirror covers both rather than leaving the sibling to a third round.
- **`InputTapResult`** mirrors the successful request contract: `key` and `action` non-empty, `strength` in `0..1`, `keycode >= 1` (a successful tap can never echo `KEY_NONE` — the harness refuses an unresolvable key with `live_invalid_key` before any window opens), and echoed `modifiers` are checked against the one shared vocabulary rule (`_validate_modifiers`, the same home the request uses, so the two sides cannot drift).

### Regressions

Eight new malformed replies join the contract-violation suites: your two (`action=""`, `strength=2.0`, and `position=[]` for the click), plus the adjacent edges the same mirror closes — `key=""`, `keycode=0`, a foreign modifier (`"control"`), and a one-element `position` on the move result. All are red on `0847e340`, each failing on exactly its new payload with the pre-existing payloads still passing first.

### Validation at `2b35d580`

- Fast suite **1766 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: input suite **16 passed**, the `--windowed` #647 twin included — real harness replies satisfy the new bounds, so the tightening refuses only drift, not the live engine.
- `gda schema` diff vs `0847e340`: exactly `input mouse-click`, `input mouse-move`, `input tap` changed (the new field constraints); no other entry moved.
- GitHub checks on `2b35d580`: green (the `Godot e2e tests` job remains dispatch-only by design).

PR #735, opened 2026-08-24
feat(gda): perf sample — bounded engine-monitor windows with statistics and budget verdicts (#662)

### LINKED ISSUE #662: perf: bounded multi-frame sampling with aggregate statistics and budget verdicts
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-027** (2026-08-10, reconfirmed on roughly ten packages). `perf monitors` (plural) exposes one instantaneous snapshot; five serial calls cannot establish whether values are stable, cached, or representative. Concrete evidence of the gap: two short sessions on the same package reported 120 FPS vs 444–497 FPS after a QA-side override was removed; another package's five snapshots ranged 115–120 FPS with 11.5–26.7 ms process time. Frame-time/draw-call release gates stayed open: no bounded distribution could be produced — and several packages also breached their draw-call or texture budgets outright, which sampling alone will not close.

## What to build

**Existing primitive to reuse:** `perf monitor` (singular) already samples a node property or signal per frame over a bounded window (`--frames`, ceiling `MAX_WINDOW_FRAMES = 600`, mirrored in the harness). The new capability is windowed sampling of the **engine performance monitors** with aggregate statistics — build it on the same window mechanics.

- A sampling operation over a declared window (`--frames N`, and `--seconds S` only if the harness gains a time-based window — today it counts frames), for selected monitors, returning:
  - count, min, max, mean, p50, p95 per selected monitor;
  - the raw timestamped samples;
  - a per-monitor pass/fail verdict when a budget file is supplied.
- Naming decision for triage: extend `perf monitors` with `--frames`, or add a third command beside `perf monitor`/`perf monitors` — avoid a near-homonym triple; #670 owns the general naming-consistency sweep.
- The `--frames` ceiling inherits the harness window limit; state it in help and the result.

## Acceptance criteria

- [ ] One command samples a declared frame window and returns aggregate statistics plus raw samples for the selected monitors.
- [ ] Budget evaluation returns a per-monitor pass/fail verdict.
- [ ] The existing snapshot behavior remains available.
- [ ] The window ceiling is stated in help and reflected in the result.
- [ ] A live regression samples a running scene over a bounded window.

## Blocked by

None - can start immediately. Related: #670 (command naming consistency).

qinhaihong-red:
## W4 slice: `perf sample` — bounded engine-monitor windows with statistics and budget verdicts (#662)

**The gap (GDA-DF-027, reconfirmed on ~10 packages):** `perf monitors` exposes one instantaneous snapshot; serial calls cannot establish whether values are stable, cached, or representative (120 vs 444–497 FPS across two sessions of the same package; five snapshots ranging 115–120 FPS with 11.5–26.7 ms process time), so frame-time/draw-call release gates could not close.

### What shipped

**`gda perf sample [--frames N] [--monitor NAME ...] [--budget FILE]`** — a third `perf` command beside `monitors`/`monitor` (the issue left the naming open: extending `perf monitors` with a mode-switching `--frames` would put two result shapes under one descriptor, against ADR-0023's one-command/one-output-model projections; `sample` is a distinct verb that does not worsen the existing monitor/monitors near-homonym pair).

- **Harness side** (`perf-sample`): reads every SELECTED engine monitor once per frame over the window — all monitors when none selected — on the same #223 time-windowed base `perf monitor` uses, frame-coherently (ADR-0020), and returns only the raw timestamped rows. The harness stays dumb on purpose.
- **CLI side** (a recipe command, the `screen` pattern, ADR-0023): computes count/min/max/mean/**p50**/**p95** per monitor (percentiles are nearest-rank, stated in the schema) and, with `--budget`, per-monitor pass/fail verdicts plus an overall `passed`. **A failed budget is data — the command exits 0**; agents branch on the JSON.
- **Budget file**: a JSON object of `{monitor: {stat, min?, max?}}` entries. `stat` (one of min/max/mean/p50/p95) is **required** — a defaulted statistic would let a release gate pass against a number nobody chose. Validated **before** dispatch, so a bad budget never costs a live window; a budget naming a monitor outside the sampled selection is refused with the fix named.
- **Model-side bounds (ADR-0015)**: monitor names validate against `PERF_MONITOR_NAMES`, a CLI mirror of the harness's `_perf_monitors` table held identical by a new sync test (the MAX_WINDOW_FRAMES pattern); `--frames` inherits the 1..600 per-window ceiling — stated in help and **echoed as `max_frames` in the result** (the issue's AC4).
- **Wire-reply coherence is validated** (the #732 review's lesson, applied up front): the declared window length must match the row count and every row must carry exactly the declared monitors, else the reply classifies as `contract_violation` — pinned by three malformed-reply regressions.
- The one-frame `perf monitors` snapshot and the per-node `perf monitor` timeline stay byte-identical in schema (AC3).

`--seconds` is not included: the harness window counts frames (the issue scoped a time-based window to a harness capability it does not have).

| AC | Evidence |
| --- | --- |
| One command samples a declared window, returns statistics + raw samples | e2e: 30-frame window over `fps` + `node_count`, 30 rows echoed, `count == 30`, `min <= p50 <= p95 <= max` per monitor |
| Budget evaluation returns per-monitor pass/fail | e2e budget pairs a rule that must pass (`node_count max 1e6`) with one that must fail (`fps min 1e6`): verdicts true/false, `passed: false`, exit 0 |
| Existing snapshot behavior remains available | same e2e calls `perf monitors` after the window; schema diff shows no change to it |
| Ceiling stated in help and reflected in the result | help test pins "per-window ceiling" + 600; result carries `max_frames` |
| A live regression samples a running scene over a bounded window | `tests/test_e2e_perf.py::test_perf_sample_collects_a_bounded_window_with_stats_and_verdicts` |

### Validation

- Fast suite **1779 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: perf suite **6 passed** (new window test included), harness-install suite **4 passed** (the harness bytes changed).
- **Red-proof on dev tip `03e47208`**: all 12 new unit tests (stats, budgets, bounds, malformed replies, mirror) and the new e2e regression fail there — `unknown_command` for the absent command, `ImportError` for the absent mirror, and the ceiling test additionally pins its message to `--frames` so it cannot pass by coincidence on a head without the command.
- `gda schema` diff vs `03e47208`, entry-by-entry: **added `perf sample`; nothing else moved.**
- Docs synced: command catalog, SKILL.md surface table, all four READMEs (+ i18n re-stamp).

Closes #662

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,6 +575,7 @@ edge-triggered controls.
 | ------- | ------------ |
 | `perf monitors` | Snapshot the engine's performance counters (fps, memory, nodes, …). |
 | `perf monitor` | Sample a node property or signal over a frame window (timeline). |
+| `perf sample` | Sample engine monitors over a frame window: statistics + budget verdicts. |
 
 **`input`** — input simulation
 
--- a/README.md
+++ b/README.md
@@ -573,9 +573,8 @@ edge-triggered controls.
 
 | Command | What it does |
 | ------- | ------------ |
-| `perf monitors` | Snapshot the engine's performance counters (fps, memory, nodes, …). |
+| `perf monitors` | Snapshot the engine's counters — or, with `--frames`, sample a window with statistics and budget verdicts. |
 | `perf monitor` | Sample a node property or signal over a frame window (timeline). |
-| `perf sample` | Sample engine monitors over a frame window: statistics + budget verdicts. |
 
 **`input`** — input simulation
 
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=8924756d76fa1008feb565864baeeb2500f62b7fc0dc3eb7e70d8ed577dbda28 -->
+<!-- gda-readme-i18n: source=README.md sha256=281f548e47e86237aacdb1a48124e333b1fa94287c3281d0963a5699b073b87b -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -587,9 +587,8 @@ variables de script getter-only/no-op o controles edge-triggered.
 
 | Comando | Qué hace |
 | ------- | ------------ |
-| `perf monitors` | Toma una instantánea de los contadores de rendimiento del motor (fps, memoria, nodos, …). |
+| `perf monitors` | Toma una instantánea de los contadores del motor — o, con `--frames`, muestrea una ventana con estadísticas y veredictos de presupuesto. |
 | `perf monitor` | Muestrea una propiedad o señal de nodo a lo largo de una ventana de frames (línea de tiempo). |
-| `perf sample` | Muestrea los monitores del motor a lo largo de una ventana de frames: estadísticas agregadas y veredictos de presupuesto. |
 
 **`input`** — simulación de entrada
 
--- a/docs/README.es.md
+++ b/docs/README.es.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=9412ce784d9839d437daa861f8b097a788f8e401dbb36130c2180b323b78350f -->
+<!-- gda-readme-i18n: source=README.md sha256=8924756d76fa1008feb565864baeeb2500f62b7fc0dc3eb7e70d8ed577dbda28 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -589,6 +589,7 @@ variables de script getter-only/no-op o controles edge-triggered.
 | ------- | ------------ |
 | `perf monitors` | Toma una instantánea de los contadores de rendimiento del motor (fps, memoria, nodos, …). |
 | `perf monitor` | Muestrea una propiedad o señal de nodo a lo largo de una ventana de frames (línea de tiempo). |
+| `perf sample` | Muestrea los monitores del motor a lo largo de una ventana de frames: estadísticas agregadas y veredictos de presupuesto. |
 
 **`input`** — simulación de entrada
 
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=9412ce784d9839d437daa861f8b097a788f8e401dbb36130c2180b323b78350f -->
+<!-- gda-readme-i18n: source=README.md sha256=8924756d76fa1008feb565864baeeb2500f62b7fc0dc3eb7e70d8ed577dbda28 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -584,6 +584,7 @@ Live の `game set --property position` は `node set` と同じ `Control` ポ�
 | ------- | ------------ |
 | `perf monitors` | エンジンのパフォーマンスカウンタ(fps、メモリ、ノード数など)のスナップショットを取得します。 |
 | `perf monitor` | ノードのプロパティまたはシグナルを、フレームのウィンドウ(タイムライン)にわたってサンプリングします。 |
+| `perf sample` | エンジンのパフォーマンスモニターをフレームのウィンドウにわたってサンプリングします(集計統計とバジェット判定)。 |
 
 **`input`** — 入力シミュレーション
 
--- a/docs/README.ja.md
+++ b/docs/README.ja.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=8924756d76fa1008feb565864baeeb2500f62b7fc0dc3eb7e70d8ed577dbda28 -->
+<!-- gda-readme-i18n: source=README.md sha256=281f548e47e86237aacdb1a48124e333b1fa94287c3281d0963a5699b073b87b -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -582,9 +582,8 @@ Live の `game set --property position` は `node set` と同じ `Control` ポ�
 
 | コマンド | 機能 |
 | ------- | ------------ |
-| `perf monitors` | エンジンのパフォーマンスカウンタ(fps、メモリ、ノード数など)のスナップショットを取得します。 |
+| `perf monitors` | エンジンのカウンタのスナップショットを取得します。`--frames` を付けるとフレームウィンドウをサンプリングし、集計統計とバジェット判定を出力します。 |
 | `perf monitor` | ノードのプロパティまたはシグナルを、フレームのウィンドウ(タイムライン)にわたってサンプリングします。 |
-| `perf sample` | エンジンのパフォーマンスモニターをフレームのウィンドウにわたってサンプリングします(集計統計とバジェット判定)。 |
 
 **`input`** — 入力シミュレーション
 
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=8924756d76fa1008feb565864baeeb2500f62b7fc0dc3eb7e70d8ed577dbda28 -->
+<!-- gda-readme-i18n: source=README.md sha256=281f548e47e86237aacdb1a48124e333b1fa94287c3281d0963a5699b073b87b -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -554,9 +554,8 @@ Live `game set --property position` 遵循与 `node set` 相同的 `Control` 策
 
 | 命令 | 作用 |
 | ------- | ------------ |
-| `perf monitors` | 对引擎的性能计数器拍一张快照（fps、内存、节点数等）。 |
+| `perf monitors` | 对引擎计数器拍快照——或配合 `--frames` 在一个帧窗口内采样，输出聚合统计与预算判定。 |
 | `perf monitor` | 在一个帧窗口内对某个节点属性或信号采样（时间线）。 |
-| `perf sample` | 在一个帧窗口内采样引擎性能监视器：聚合统计与预算判定。 |
 
 **`input`** — 输入模拟
 
--- a/docs/README.zh-CN.md
+++ b/docs/README.zh-CN.md
@@ -1,4 +1,4 @@
-<!-- gda-readme-i18n: source=README.md sha256=9412ce784d9839d437daa861f8b097a788f8e401dbb36130c2180b323b78350f -->
+<!-- gda-readme-i18n: source=README.md sha256=8924756d76fa1008feb565864baeeb2500f62b7fc0dc3eb7e70d8ed577dbda28 -->
 
 # godot-agent (`gda`): Godot AI Agent CLI, Skill, and MCP Server
 
@@ -556,6 +556,7 @@ Live `game set --property position` 遵循与 `node set` 相同的 `Control` 策
 | ------- | ------------ |
 | `perf monitors` | 对引擎的性能计数器拍一张快照（fps、内存、节点数等）。 |
 | `perf monitor` | 在一个帧窗口内对某个节点属性或信号采样（时间线）。 |
+| `perf sample` | 在一个帧窗口内采样引擎性能监视器：聚合统计与预算判定。 |
 
 **`input`** — 输入模拟
 
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -843,6 +843,21 @@ headless is unaffected (4.4+, cross-platform).
   node is `live_perf_node_not_found`, an absent property `live_perf_property_not_found`,
   an absent signal `live_perf_signal_not_found`; a genuinely stalled engine is caught
   by the daemon-level `live_timeout`.
+  `perf sample [--frames N] [--monitor NAME ...] [--budget FILE]` (shipped, #662) is
+  the windowed counterpart of the one-frame snapshot: the harness reads every
+  selected ENGINE monitor once per frame over the window (all monitors when no
+  `--monitor` is given) and returns the raw timestamped samples; the CLI — a
+  recipe command, the `screen` pattern (ADR-0023) — computes count/min/max/mean/
+  p50/p95 per monitor (percentiles are nearest-rank) and, with `--budget`, a
+  per-monitor pass/fail verdict plus an overall `passed`. The budget file is a
+  JSON object of `{monitor: {stat, min?, max?}}` entries (`stat` one of min, max,
+  mean, p50, p95; at least one bound), validated BEFORE dispatch so a bad budget
+  never costs a live window; a budget for a monitor outside the sampled selection
+  is refused. A failed budget is data — the command still exits 0. Monitor names
+  are bounded model-side against a CLI mirror of the harness's monitor table (a
+  sync test holds the two identical), the `--frames` bound inherits the same
+  1..600 per-window ceiling (stated in help and echoed as `max_frames` in the
+  result), and the one-frame `perf monitors` snapshot stays as it is.
 - **`diag` (diagnostics):** runtime errors of the running game (shipped, #224; callstacks #283). `gda diag errors`
   reads the running game's runtime errors as structured `{level, message, function?, file?, line?, callstack}`
   (warnings included, distinguished by `level`), with `--limit N`. `callstack` is an ordered
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -843,21 +843,29 @@ headless is unaffected (4.4+, cross-platform).
   node is `live_perf_node_not_found`, an absent property `live_perf_property_not_found`,
   an absent signal `live_perf_signal_not_found`; a genuinely stalled engine is caught
   by the daemon-level `live_timeout`.
-  `perf sample [--frames N] [--monitor NAME ...] [--budget FILE]` (shipped, #662) is
-  the windowed counterpart of the one-frame snapshot: the harness reads every
-  selected ENGINE monitor once per frame over the window (all monitors when no
-  `--monitor` is given) and returns the raw timestamped samples; the CLI — a
-  recipe command, the `screen` pattern (ADR-0023) — computes count/min/max/mean/
-  p50/p95 per monitor (percentiles are nearest-rank) and, with `--budget`, a
-  per-monitor pass/fail verdict plus an overall `passed`. The budget file is a
-  JSON object of `{monitor: {stat, min?, max?}}` entries (`stat` one of min, max,
-  mean, p50, p95; at least one bound), validated BEFORE dispatch so a bad budget
-  never costs a live window; a budget for a monitor outside the sampled selection
-  is refused. A failed budget is data — the command still exits 0. Monitor names
-  are bounded model-side against a CLI mirror of the harness's monitor table (a
-  sync test holds the two identical), the `--frames` bound inherits the same
-  1..600 per-window ceiling (stated in help and echoed as `max_frames` in the
-  result), and the one-frame `perf monitors` snapshot stays as it is.
+  `perf monitors` also has a WINDOW mode (shipped, #662; the issue's triage
+  decision put it on the existing command — no third near-homonym). With
+  `--frames N`, the harness reads every selected engine monitor once per frame
+  over the window. All monitors are read when no `--monitor` is given. The
+  harness returns only the raw timestamped samples. The CLI computes the
+  statistics — count, min, max, mean, p50, p95 per monitor; percentiles are
+  nearest-rank — because the command is a recipe (ADR-0023, the `screen`
+  pattern). With `--budget FILE`, the CLI also evaluates one pass/fail verdict
+  per budgeted monitor, plus an overall `passed`. A failed budget is data: the
+  command still exits 0. The budget file is a JSON object of
+  `{monitor: {stat, min?, max?}}` entries. `stat` is required (one of min, max,
+  mean, p50, p95) and at least one bound must be set. Admission is strict:
+  UTF-8 only, unique keys at every depth, finite numbers only. The budget is
+  validated before dispatch, so a bad file never costs a live window; a budget
+  for a monitor outside the sampled selection is refused. Monitor names are
+  bounded model-side against a CLI mirror of the harness's monitor table (a
+  sync test holds the two identical). The reply is correlated with the request:
+  a self-consistent reply for a different window or selection classifies as
+  `contract_violation`. The `--frames` bound inherits the same 1..600
+  per-window ceiling, stated in help and echoed as `max_frames` in the result.
+  The result names its mode (`kind: snapshot | window`); `--monitor` and
+  `--budget` require `--frames`, and the no-flag snapshot behavior is
+  unchanged.
 - **`diag` (diagnostics):** runtime errors of the running game (shipped, #224; callstacks #283). `gda diag errors`
   reads the running game's runtime errors as structured `{level, message, function?, file?, line?, callstack}`
   (warnings included, distinguished by `level`), with `--limit N`. `callstack` is an ordered
--- a/src/gda/commands/perf.py
+++ b/src/gda/commands/perf.py
@@ -237,7 +237,16 @@ def _json_integer(value: object) -> object:
     raise ValueError("must be a JSON integer")
 
 
-JsonInteger = Annotated[int, BeforeValidator(_json_integer)]
+# The window's frame count as ONE annotated type: the range markers come
+# BEFORE the validator in the metadata, which is the ordering pydantic can
+# translate into standard JSON Schema `minimum`/`maximum` keywords — with the
+# validator first, the emitted schema carries raw `ge`/`le` keys that standard
+# validators ignore, silently dropping the published range (#735 recheck 3).
+# At runtime the BeforeValidator still runs first, so the range applies to the
+# already-normalized integer.
+FrameWindow = Annotated[
+    int, Field(ge=1, le=MAX_WINDOW_FRAMES), BeforeValidator(_json_integer)
+]
 
 # The statistics a budget rule may gate on — the aggregate set minus ``count``,
 # which counts frames rather than measuring the monitor.
@@ -302,9 +311,7 @@ class PerfMonitorsParams(BaseModel):
 
     model_config = ConfigDict(json_schema_extra=_PERF_MONITORS_MODE_SCHEMA)
 
-    frames: Optional[JsonInteger] = Field(
-        default=None, ge=1, le=MAX_WINDOW_FRAMES, description=_FRAMES_DESC
-    )
+    frames: Optional[FrameWindow] = Field(default=None, description=_FRAMES_DESC)
     monitors: list[PerfMonitorName] = Field(
         default_factory=list,
         description=(
--- a/src/gda/commands/perf.py
+++ b/src/gda/commands/perf.py
@@ -26,11 +26,19 @@ time-windowed multi-frame harness base, ADR-0020).
 
 import json
 import math
+from enum import StrEnum
 from pathlib import Path
-from typing import Any, Callable, Literal, Optional
+from typing import Annotated, Any, Callable, Literal, Optional
 
 import typer
-from pydantic import BaseModel, ConfigDict, Field, ValidationError, model_validator
+from pydantic import (
+    BaseModel,
+    BeforeValidator,
+    ConfigDict,
+    Field,
+    ValidationError,
+    model_validator,
+)
 
 from gda import dispatch
 from gda.dispatch import dispatch_domain, dispatch_recipe, params_or_bad_parameter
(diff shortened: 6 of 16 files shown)


qinhaihong-red:

Reviewed exact head `e94fe4c432ea4614403ab11c0533f5cbdee36c7c` against base / merge-base `03e472085c03955245b5157be447fb4d1143fedf`. Artifact, Domain Architecture, Standards, and Spec were reviewed as independent axes; findings are not cross-axis reranked.

### Artifact review — Markdown

#### Required

1. **The six changed Markdown surfaces publish a command shape that #662's triage decision explicitly rejected.**

   Locations: `README.md:578`, `docs/README.es.md:592`, `docs/README.ja.md:587`, `docs/README.zh-CN.md:559`, `docs/command-catalog.md:846-860`, and `src/gda/skill/SKILL.md:174`.

   The issue body reserved the naming choice for triage. The member triage decision then resolved it as: extend `perf monitors` with the window and budget options, **no third command**, specifically to avoid the `monitor` / `monitors` / `sample` near-homonym triple. The docs accurately mirror the current implementation, but they therefore drift from the requirement authority they cite.

   Smallest repair: move the capability onto `perf monitors --frames/--monitor/--budget` and update all six surfaces. If the third command is retained, first amend the accepted #662 naming decision in the appropriate authority.

2. **The documented budget-validation guarantee is false at this head.**

   `docs/command-catalog.md:852-856` says a bad JSON budget is validated before dispatch and never costs a live window. Three counterexamples reach or escape that boundary:

   - duplicate monitor keys are silently resolved last-key-wins; a file that first sets `fps min=100` and then repeats `fps min=0` produces `passed: true` at FPS 60;
   - `min: -Infinity` is accepted and produces `passed: true`, while the public result serializes the actual bound as `"min": null`;
   - invalid UTF-8 leaks `UnicodeDecodeError` instead of the documented structured `invalid_params`.

   Smallest repair: admit strict UTF-8 JSON, reject duplicate keys and non-finite numbers, map encoding/parsing/model failures to `invalid_params`, and document the unique-key/finite-number boundary.

#### Minor

`docs/command-catalog.md:846-860` combines sampling ownership, recipe mechanics, statistics, budget evaluation, and validation into one long sentence with dense parentheses and all-caps `ENGINE` / `BEFORE`. Split it into short responsibility- and boundary-focused sentences.

#### Terminology and prose assessment

Sources: `CONTEXT.md`, ADR-0015/0017/0020/0023/0040, #662, and rendered help/schema. Reference: ASD-STE100 Simplified Technical English. I inspected all changed passages, the full current artifacts, command-name mapping, responsibility/scope/order language, and the Spanish/Japanese/Chinese mirrors. No other substantive terminology, prose, or translation drift was found.

#### [P1] Budget admission is not a closed, total rule boundary

Locations: `src/gda/commands/perf.py:276-294,429-468`.

The duplicate-key, non-finite-number, and invalid-UTF-8 probes above show that a release gate can pass against an ambiguous or unrepresentable rule, or bypass the structured failure contract. Keep ownership in the existing `_load_budgets` / `PerfBudget` boundary: reject duplicate pairs and parser constants, require finite model values (including direct construction / exponent overflow), catch decoding failures, and add no-runner regressions for top-level and nested duplicates, `NaN` / infinities, overflow, and invalid UTF-8.

#### [P2] The harness reply is not correlated with the request

Locations: `src/gda/commands/perf.py:393-403,515-525`.

`_SampleReply` validates only internal self-consistency. It does not require `reply.frames == params.frames`, reply monitors to equal the requested selection, unique declared monitor names, or frame indices to be exactly `0..frames-1`. A request for five `fps` frames can accept an internally consistent two-row `draw_calls` reply whose rows are both `frame: 0`, and then publish it as success.

Smallest repair: compare the classified reply with the expected frame count and monitor selection, require unique monitor names and an exact frame sequence, and classify each mismatch as `contract_violation`.

#### [P2] The budget path bypasses the shared path-normalization owner

Locations: `src/gda/commands/perf.py:251-260,501-503`.

`budget: Optional[str]` is converted directly to `Path`, so literal `~/budget.json` is not expanded on `--params-json` / MCP or direct argv without a shell. This violates ADR-0006's one CLI-layer path rule and ADR-0015's model-side normalization rule.

Smallest repair: use `NormalizedPath | None` and add argv plus `--params-json` literal-tilde regressions.

Architecture assessment: the overall direction remains sound — the harness owns frame-coherent raw sampling, the Python `perf` slice owns statistics and budget policy, and the descriptor/recipe owns public composition. Orthogonality passes; completeness fails on admission and request/reply boundaries; DRY partially fails where shared path and input-validity knowledge is bypassed. No new context, module, port hierarchy, or event mechanism is warranted.

### Standards

One hard documented-standard violation; no reportable Fowler-baseline smell:

- `src/gda/commands/perf.py:251,502` does not use `NormalizedPath`, violating `docs/adr/0006-project-context-and-path-resolution.md:33-40` and `docs/adr/0015-structured-params-input-abi.md:59-64`. Use the shared field type and test both public input channels.

### Spec

#### [High] The implementation violates the resolved naming requirement

The body of #662 offered two choices and explicitly reserved the decision for triage. The member triage comment chose “extend `perf monitors` ... — no third command.” At this head, `src/gda/commands/perf.py:655` leaves `perf monitors` as the optionless snapshot and line 740 registers `perf sample`. Reproduction: `gda perf monitors --frames 5 --json` returns `unknown_option`, while `gda perf --help` lists the prohibited near-homonym triple.

Move the window, selection, and budget behavior onto `perf monitors`, retaining the no-window invocation's snapshot behavior. The other acceptance criteria — statistics plus raw selected samples, per-monitor verdicts, ceiling in help/result, snapshot availability, and a live bounded-window regression — are implemented.

Standards summary: 1 finding; worst is the path-normalization violation. Spec summary: 1 finding; worst is the resolved public-command shape being implemented incorrectly.

### Validation

- Focused perf/error/descriptor suites: **48 passed**
- New real-engine bounded-window E2E: **1 passed**
- README/i18n/Skill checks: **12 passed**
- Aggregate schema diff: added `perf sample`; no existing entry changed
- `git diff --check`: pass
- Current non-Godot GitHub checks: pass
- GitHub `Godot e2e tests`: **skipped**

Not run: the complete local suite, MCP E2E, and the low-FPS 600-frame / `live_timeout` boundary.

qinhaihong-red:
## Review response — all findings adopted at head `909346e8`

Baseline: base tip / merge base `03e47208`, reviewed head `e94fe4c4`, new head `909346e8`. Every mechanical probe was reproduced on `e94fe4c4` before the fix; the reproduction outputs are quoted below.

| --- | --- | --- | --- |
| [High/Spec + Required/Md] The command shape violates #662's resolved triage decision | **Fully adopt** | The member triage comment (2026-08-15) on #662 says "extend `perf monitors` … no third command" — the slice read only the issue body and missed it; premise re-verified against the issue's comments | The window lives on `perf monitors --frames/--monitor/--budget`; `perf sample` is gone; all six Markdown surfaces updated |
| [Required/Md + P1] Budget admission is not a closed rule boundary | **Fully adopt** | On `e94fe4c4`: dup-key budget → `passed=True` (the 100 gate erased by the repeated 0); `min: -Infinity` → `passed=True` with the bound serialized `"min": null`; invalid UTF-8 → raw `UnicodeDecodeError` | Strict UTF-8, unique keys at every depth, finite numbers only (constants at parse, `1e999` at the model); 13 bad-budget regressions |
| [P2] The reply is not correlated with the request | **Fully adopt** | On `e94fe4c4`: a self-consistent 2-row `draw_calls` reply (both rows `frame: 0` — caught now on that axis too) answered a 5-frame `fps` request as SUCCESS | Recipe-side correlation (frames, selection) + reply-side exact frame sequence and unique monitor names; each mismatch is `contract_violation` |
| [P2 + Standards] The budget path bypasses `NormalizedPath` | **Fully adopt** | `~/budget.json` was opened verbatim off `--params-json` | `budget: Optional[NormalizedPath]`; literal-tilde regressions on argv and `--params-json` |
| [Minor] The catalog block is one dense sentence | **Fully adopt** | — | Rewritten as short responsibility/boundary sentences |

### 1 — the resolved naming, honored

The failure was process, not judgment: the PR argued the naming from the issue *body* ("reserved for triage") without checking the issue's *comments*, where triage had already resolved it. The window, selection, and budget behavior now live on `perf monitors`:

- No `--frames`: the original snapshot, byte-for-byte the same wire exchange (`perf-monitors`, `{}`), now wrapped as `kind: "snapshot"` with the window fields null.
- `--frames N`: the window mode — raw samples from the harness, statistics and verdicts CLI-side, exactly as before but under the surface triage chose.
- The result is a validated two-mode shape: each mode requires its own fields and forbids the other's (`kind: snapshot | window`), so a mixed payload fails output validation rather than passing through. `--monitor` / `--budget` without `--frames` are refused by name (a silently inert selection would be the GDA-DF-037 failure mode).
- The command stays one descriptor with one recipe (ADR-0023); the recipe still dispatches the sentinel ops, like `script validate`. The window wire op keeps its `perf-sample` name harness-side — it is internal ABI, and the relayed-op mirror now counts it from the one source-side constant (`PERF_SAMPLE_OP`) so no parallel table appears.
- `gda schema` vs the dev tip: **exactly one entry changed (`perf monitors`); nothing added, nothing removed.** `gda perf --help` lists `monitors` and `monitor` only.

### 2 — budget admission is a closed boundary

Adopted as specified, in the existing `_load_budgets` / `PerfBudget` owners:

- **UTF-8**: the file is read as bytes and decoded strictly; a mis-encoded file is the structured `invalid_params`, never a traceback.
- **Duplicate keys**: an `object_pairs_hook` refuses a repeated key at ANY depth — top-level (your gate-erasing case) and inside an entry alike — instead of `json.loads`' silent last-key-wins.
- **Finite numbers**: the JSON-extension constants (`Infinity`/`-Infinity`/`NaN`) are refused at parse via `parse_constant`; an exponent-overflow like `1e999` arrives as an ordinary float and is refused by the model's `isfinite` rule, which also covers direct construction.

The 13-case bad-budget regression list now includes your three probes plus nested duplicates, all three constants, and the overflow.

### 3 — request/reply correlation

Split across the two owners by what each can know: the reply model now requires rows to be exactly frames `0..N-1` in order and the declared monitors to be unique (self-consistency); the recipe — the only place that holds the params — requires `reply.frames == params.frames` and `reply.monitors ==` the requested selection (the mirrored full table for an empty one). Your exact probe (two `frame: 0` rows of `draw_calls` for a five-frame `fps` request) now dies twice over: the frame sequence fails the reply model, and the selection fails the correlation; both as `contract_violation`.

### 4 — NormalizedPath

Adopted; the field is `Optional[NormalizedPath]`, so a literal `~/budget.json` expands model-side on both public channels — pinned by a regression that runs the same budget through argv and `--params-json` under a redirected `HOME`.

### Validation at `909346e8`

- Fast suite **1780 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: perf suite **6 passed** (the bounded-window regression now drives `perf monitors --frames 30 --budget …` and asserts `kind`, the ceiling echo, the deterministic pass+fail verdict pair, and the snapshot mode on the same surface), harness-install **4 passed**.
- Reproductions on `e94fe4c4` (before the fix): `dup-key: passed=True`; `inf: passed=True, serialized min: None`; `utf8: UnicodeDecodeError leaked`; `correlation: SUCCESS with monitors=['draw_calls']`. The corresponding regressions are green at the new head and did not exist in passing form on the reviewed one (the command shape they pin was refused there).
- On the boundary you flagged as not-run: a 600-frame window at 60 fps is ~10 s against the daemon's 30 s `OP_TIMEOUT`, so the ceiling holds ~3× headroom; an engine slow enough to breach it hits the daemon-level `live_timeout` guard, the same property every #223 window has had. Not separately tested here — stated so the record carries the analysis rather than silence.
- GitHub checks on `909346e8`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Exact-head recheck — changes still required

Rechecked exact head `909346e8e6f768062d447508bd35e3aee9cb920b` after the review response. This is a residue-only replay of the previous findings plus their directly adjacent contract boundaries. Artifact/Markdown, Domain Architecture, Standards, and Spec remain independent axes.

The previous command-shape finding, the original duplicate/non-finite/UTF-8 probes, request/reply correlation, `NormalizedPath`, and catalog-prose finding are closed. The following residue remains.

### Artifact review — Markdown

#### Required: the documented strict-number boundary still accepts non-number JSON tokens

`docs/command-catalog.md:857-859` says budget admission is strict and accepts finite numbers only; generated help/schema also describes finite bounds. However, `PerfBudget.min/max` are permissive Pydantic `float` fields:

- `{"min": "10"}` is accepted as `10.0`;
- `{"min": true}` is accepted as `1.0`.

Both files proceed into the live window, so the implementation does not match the published strict JSON-number boundary. The boolean case can silently create an unintended release gate. Require the original JSON token to be a number (while still accepting integer and fractional numbers), and add no-runner regressions for strings and booleans. Weakening the documentation to promise coercion would also align the artifact, but would make this release-gate boundary less safe.

Terminology/prose replay: all six Markdown surfaces now use `perf monitors`; the old dense prose is repaired; no Markdown `perf sample` residue remains.

#### [P2] Budget decoding is still not a total structured-failure boundary

`src/gda/commands/perf.py:537-568` catches `ValueError` around `json.loads`, but a 10,000-level JSON value raises `RecursionError`. The public command then exits through a raw exception with no structured `invalid_params` envelope. Catch the decoder's recursion failure at this boundary and add a regression asserting that no live operation is called.

#### [P2] `PerfBudget` admits an impossible interval

`PerfBudget.model_validate({"stat": "p50", "min": 100, "max": 50})` succeeds. `_evaluate_budgets` then reports every observation as a performance failure, misclassifying invalid gate configuration and spending a live window. Keep ownership in `PerfBudget`: reject `min > max` before dispatch and cover it with an `invalid_params` / no-runner regression.

Architecture assessment: completeness still fails on these two admission edges. Ownership, dependencies/communication, orthogonality, DRY, and extensibility pass. The previous request/reply-correlation and path-ownership findings are closed.

### Standards

#### Hard violation: the published schema and runtime models disagree on both modes

The snapshot/window conditions in `PerfMonitorsParams` and `PerfMonitorsResult` exist only in `model_validator`, which `model_json_schema()` cannot express:

- the public input schema accepts `{"monitors":["fps"]}`, while `--params-json` / `PerfMonitorsParams` reject it because `frames` is absent;
- the public output schema accepts `{"kind":"snapshot"}` and mixed snapshot/window fields, while `PerfMonitorsResult` rejects both.

This violates ADR-0015's input same-contract guarantee and ADR-0004's machine-readable output contract; gda-mcp receives schemas wider than the actual ABI. Express both modes through schema-generating typed variants/conditionals and add these counterexamples as schema tests.

#### Hard violation: public schema prose names a removed command

`src/gda/commands/perf.py:315-316` becomes `$defs.PerfSampleFrame.description` in `gda perf monitors --schema`, where it still says ``perf sample`` window. Change the public description to ``perf monitors --frames``; the explicit internal wire operation `perf-sample` can remain. The adjacent source comments at lines 172/176 should be cleaned at the same time.

Baseline smell: the `kind` switch is repeated in `_mode_fields` and `render_perf_monitors`; this shares the first finding's root and can be reduced by typed variants.

Standards summary: 2 hard findings; the worst is the public input/output schema accepting shapes that the runtime ABI rejects.

### Spec — pass

#662's resolved naming and acceptance criteria now pass: `perf monitors` owns the window/budget options, no third `sample` command exists, snapshot mode remains available, the 1..600 ceiling appears in help/result, and a real bounded-window regression covers statistics, raw samples, and budget verdicts.

### Validation

- Focused unit/contract suites: **179 passed**
- Real Godot perf E2E suite: **6 passed**
- Ruff: pass; Pyright: 0 errors; `git diff --check`: pass
- Aggregate schema vs base: only `perf monitors` changed; no command added or removed
- Current applicable GitHub checks: pass; `Godot e2e tests`: **skipped**

qinhaihong-red:
## Recheck response — all five residues closed at head `ae367b4a`

Baseline: base tip / merge base `03e47208`, rechecked head `909346e8`, new head `ae367b4a`. Every probe was reproduced on `909346e8` before the fix: the three bad budgets **passed admission and reached the runner** (the probe's stand-in runner then blew up, which is the acceptance evidence), the deeply nested document escaped as a raw `RecursionError`, and the emitted schemas accepted both of your counterexamples (`{"monitors": ["fps"]}` on input, `{"kind": "snapshot"}` on output).

| Claim | Decision | Evidence on `909346e8` | Change |
| --- | --- | --- | --- |
| [Required/Md] Non-number JSON tokens pass the "strict numbers" boundary | **Fully adopt** — the strict direction, not the doc-weakening one | `{"min": "10"}` and `{"min": true}` both admitted | Bounds are pydantic STRICT floats: int and fractional numbers accepted, strings and booleans refused |
| [P2] Budget decoding is not a total structured-failure boundary | **Fully adopt** | 20,000-deep nesting → raw `RecursionError` | A `RecursionError` arm beside the `ValueError` one; the regression asserts `invalid_params` and zero runner calls |
| [P2] `PerfBudget` admits an impossible interval | **Fully adopt** | `min: 100, max: 50` admitted, reached the runner | `min > max` refused by the model as a misconfiguration, before any live window |
| [Hard] The published schemas are wider than the runtime ABI | **Fully adopt** | Both counterexamples validate against the emitted schemas | The mode rules are published as schema conditionals (the #669 `json_schema_extra` pattern), plus a corpus parity test |
| [Hard] Public schema prose names the removed command | **Fully adopt** | `$defs.PerfSampleFrame.description` said "perf sample" | The description and the two adjacent source comments say `perf monitors --frames`; the wire op name stays, as you allowed |

### The schema/ABI alignment, in the repo's own pattern

Rather than restructuring into discriminated-union variants, the mode rules ride `json_schema_extra` — the same mechanism #669 used for the mouse-button phase, model stays the enforcing authority, schema becomes checkable:

- **Input**: two conditionals — a non-empty `monitors` requires an integer `frames`, and a string `budget` requires one — so `{"monitors": ["fps"]}` (and the present-but-null `{"frames": null, "monitors": [...]}` variant) fail the published contract exactly as `--params-json` refuses them.
- **Output**: a `oneOf` splitting the modes — each requires its own fields with their real types and pins the other mode's fields to null — plus a paired conditional making a window's `budget` and `passed` travel together. `{"kind": "snapshot"}`, mixed-mode payloads, and a budget without its overall verdict all fail.
- **Parity, both directions**: a corpus test runs the same instances through the emitted schema and the model and requires identical verdicts (so the two cannot drift), then separately asserts the schemas REJECT the counterexamples — parity alone could hold with both sides too wide.

On the baseline smell (the `kind` switch appearing in `_mode_fields` and `render_perf_monitors`): **not adopted.** With the conditionals published via `json_schema_extra`, a typed-variant restructure would buy the deduplication of two small, adjacent branches at the cost of reshaping the public `$defs` again this round; the two switches share one authority for the rule (the model validator — the renderer only formats what validation already guaranteed, and says so in its asserts). The trigger to revisit is a third mode.

### Validation at `ae367b4a`

- Fast suite **1781 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: perf suite **6 passed**.
- The bad-budget regression list is now 17 cases (your two token cases, the impossible interval, and the deep-nesting document joined it; the nesting case additionally asserts no runner call was made).
- `gda schema` vs the dev tip `03e47208`: still exactly one entry changed (`perf monitors`); nothing added or removed.
- GitHub checks on `ae367b4a`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Second exact-head recheck — two residues remain

Rechecked exact head `ae367b4a5b16a54cfd7257cbf2f0ccecb7b381d1` after the latest response. This replay is limited to the previously open findings and their directly adjacent public-contract boundaries.

The five direct residues from the previous review are closed in production: strict budget-number admission, decoder `RecursionError`, impossible intervals, the two reported mode-schema counterexamples, and the stale public `perf sample` description. Two adjacent issues remain.

### Artifact review — Markdown: pass

`docs/command-catalog.md`'s strict finite-number claim now matches observed behavior: quoted numbers and booleans are refused; integer and fractional JSON numbers remain valid. The required terminology/prose replay found no Markdown residue.

#### [P2] The budget admission table repeatedly tests one final file

`tests/test_perf_commands.py:445-448` always writes `_budget_file(...)` to `tmp_path / "budget.json"`. The `problems` list at lines 657-690 calls that helper repeatedly **before** the loop starts. Consequently, every `_budget_file(...)` entry points to the same path after it has been overwritten by the final deep-JSON case.

The test currently exercises the missing path, the separately named invalid-UTF-8 file, and the final deep document repeatedly. It does **not** retain regression coverage for syntax/shape errors, duplicate keys, non-finite values, strict string/boolean rejection, or `min > max`; those branches could regress while this test stays green.

Smallest repair: give every table entry a unique filename, or store `(name, content/bytes)` and write each case inside the loop. Keep the per-case `invalid_params` assertion and the final `fake.calls == []` assertion.

Independent, uniquely isolated probes confirm the production owners themselves are correct: `_load_budgets` returns structured `invalid_params` for deep nesting, and `PerfBudget` rejects `min > max`, strings, and booleans while accepting integer/fractional values. The finding is about incomplete durable evidence for load-bearing admission claim `ARCH-005`, not a currently reproduced production failure.

### Standards

#### Hard violation: the public input schema and `--params-json` still disagree

The new parity corpus closes the two reported mode examples, but it does not cover the complete `PerfMonitorsParams` contract. Reproductions at this head:

- `{"frames":5,"monitors":["fpss"]}` is accepted by the emitted input schema but rejected by `PerfMonitorsParams` / `--params-json`;
- `{"frames":"5"}` is rejected by the schema but accepted by `--params-json`, coerced to `5`, and dispatched;
- `{"frames":true}` is rejected by the schema but accepted by `--params-json`, coerced to `1`, and dispatched.

This is a bidirectional violation of ADR-0015's guarantee that the MCP-facing input schema and verbatim `--params-json` format are the same contract. The last two real CLI probes reached `daemon_not_running`, proving they passed input admission; the unknown monitor returned `invalid_params` despite validating against the schema.

Smallest repair: make `frames` a strict integer, express monitor names through one schema-derived `Literal`/Enum authority, and add these three instances to the parity corpus. The already-adopted `json_schema_extra` mode conditionals do not need to be replaced merely to close this finding.

Standards summary: 1 hard finding; worst issue is the public schema and verbatim structured-input ABI accepting different objects.

### Spec — pass

#662's naming and acceptance criteria remain satisfied: no third `sample` command, snapshot and bounded-window modes remain available, help/result publish the 1..600 ceiling, and the real E2E covers statistics, raw samples, and budget verdicts.

### Validation

- Focused unit/contract suites: **180 passed**
- Real Godot perf E2E suite: **6 passed**
- Ruff: pass; Pyright: 0 errors; `git diff --check`: pass
- Aggregate schema vs base: only `perf monitors` changed; no command added or removed
- Current applicable GitHub checks: pass; `Godot e2e tests`: **skipped**

qinhaihong-red:
## Second-recheck response — both residues closed at head `9994487e`

Baseline: base tip / merge base `03e47208`, rechecked head `ae367b4a`, new head `9994487e`. The parity mismatches were reproduced on `ae367b4a` before the change:

```
{'frames': 5, 'monitors': ['fpss']}  schema: True   model: False
{'frames': '5'}                      schema: False  model: True   (coerced, dispatched)
{'frames': True}                     schema: False  model: True   (coerced, dispatched)
{'frames': 5.0}                      schema: True   model: True   (the edge both sides must keep)
```

| Claim | Decision | Change |
| --- | --- | --- |
| [Hard] The input schema and `--params-json` accept different objects | **Fully adopt** | One static monitor-name authority; `frames` admits exactly what JSON Schema's `integer` admits; the parity corpus covers all three instances plus the `5.0` edge, both directions |
| [P2] The admission table aliases one final file | **Fully adopt** | Unique file per case, plus a per-case message-fragment assertion, so each branch is proven to fire for its own reason |

### 1 — the input contract, aligned in both directions

- **Monitor names**: `PerfMonitorName` is now a static `StrEnum` and the ONE authority — the mirrored `PERF_MONITOR_NAMES` tuple derives from its members, the harness sync test keeps holding that tuple to the GDScript table, and the emitted input schema enumerates the members. `{"monitors": ["fpss"]}` now fails the published contract exactly as `--params-json` refuses it; the model validator's redundant name check is gone (the enum is the boundary).
- **`frames`**: your smallest repair said "strict integer", and a plain pydantic strict int would indeed refuse `"5"` and `true` — but it also refuses `5.0`, which JSON Schema's `integer` accepts (any number with a zero fractional part), so it would have traded the reported mismatch for a new one in the opposite direction. The field instead rides a `BeforeValidator` implementing the schema's own semantics: integers and zero-fraction floats pass, quoted numbers and booleans are refused. The corpus asserts all four instances in both directions.
- The `json_schema_extra` mode conditionals stay as adopted last round, untouched.

Scope note, disclosed: the same lax-coercion gap exists on every pre-existing integer param in the repo (`perf monitor --frames`, `input tap --hold-frames`, …) — pydantic's lax mode has always accepted `"5"` where the published schemas say `integer`. This round closes the model the finding names; a repo-wide sweep changes published behavior across many commands and deserves its own decision rather than piecemeal drift inside this slice.

### 2 — the admission table carries its own evidence now

Exactly the failure you diagnosed: the helper wrote every case to one shared `budget.json` before the loop ran, so all unnamed entries aliased the deep-nesting content and the other branches lost their regression coverage (the production owners were and are correct — your isolated probes and mine agree — the gap was durable evidence). The table is rebuilt as `(filename, content, expected message fragment)` triples: each case gets its own file, written per entry, and the loop asserts both `invalid_params` AND the fragment naming that case's own refusal ("duplicate key", "non-finite", "impossible interval", "nests too deeply", "not valid UTF-8", …). A branch that regresses now fails its own row instead of hiding behind the last file written; the final `fake.calls == []` stands.

### Validation at `9994487e`

- Fast suite **1781 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: perf suite **6 passed**.
- The 17-case admission table now proves each branch by message; the parity corpus asserts the three reported instances and the `5.0` edge in both directions.
- `gda schema` vs the dev tip `03e47208`: still exactly one entry changed (`perf monitors`); nothing added or removed.
- GitHub checks on `9994487e`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Third exact-head recheck — one Standards residue remains

Rechecked exact head `9994487e2c61fa28cad76af50d4290c004366835` after the second-recheck response. This replay remains limited to the two prior residues and their directly adjacent boundaries; Artifact/Markdown, Domain Architecture, Standards, and Spec remain independent axes.

The prior DDMA admission-evidence finding is closed, and the three reported schema/ABI token mismatches plus the `5.0` edge are closed. One adjacent public-contract regression remains.

### Artifact review — Markdown: pass

This repair has no Markdown diff. The existing `1..600` wording remains the intended contract, and the terminology/prose replay found no new Markdown residue. The implementation contradiction below is counted once under Standards.

### Domain modular architecture — pass

The prior P2 is closed. All 17 malformed-budget cases now use unique paths, materialize their own content in the execution loop, assert both `invalid_params` and a branch-specific diagnostic fragment, and retain the accumulated `fake.calls == []` boundary assertion. The production admission owners and runner seam remain correctly placed.

### Standards

#### [P1] The generated input schema no longer enforces the documented frame range

At `src/gda/commands/perf.py:305-306`, combining the new `BeforeValidator` type with the outer `Field(ge=1, le=600)` emits this public schema branch:

```json
{"type": "integer", "ge": 1, "le": 600}
```

`ge` and `le` are not JSON Schema range keywords, so standard validators ignore them. At this exact head:

- `{"frames": -1}`: emitted schema accepts; model rejects;
- `{"frames": 0}`: emitted schema accepts; model rejects;
- `{"frames": 601}`: emitted schema accepts; model and `--params-json` reject with `invalid_params`.

This leaves the ADR-0015 schema/structured-input contract mismatched even though the newly reported token cases are fixed. Emit standard `minimum: 1` and `maximum: 600` constraints on the integer branch, then add `0` and `601` to the parity corpus at `tests/test_perf_commands.py:933-975`.

Verified closed on both sides: unknown monitor `fpss`, quoted `"5"`, and boolean `true` are rejected; zero-fraction `5.0` is accepted and normalized to `5`. `PerfMonitorName` also drives the published monitor enumeration and the mirrored tuple.

### Spec — pass

Issue #662 requires the inherited ceiling to be stated in help and reflected in the result. Both remain satisfied, and CLI/model behavior still enforces `1..600`; the remaining defect is the Standards-level machine-readable schema contract above.

### Validation

- Complete `tests/test_perf_commands.py`: **30 passed**
- Ruff: pass; Pyright: 0 errors; `git diff --check`: pass
- Current applicable GitHub checks: pass; `Godot e2e tests`: skipped
- Final pre-publication head check: still `9994487e`

qinhaihong-red:
## Third-recheck response — the range residue is closed at head `ea4fb78d`

Baseline: base tip / merge base `03e47208`, rechecked head `9994487e`, new head `ea4fb78d`. Reproduced first, as reported: on `9994487e` the emitted integer branch carried `{"ge": 1, "le": 600}` — keys standard JSON Schema validators ignore — and `-1`, `0`, and `601` all validated against the published input schema while the model refused them.

| Claim | Decision | Evidence on `9994487e` | Change |
| --- | --- | --- | --- |
| [P1] The generated input schema no longer enforces the documented frame range | **Fully adopt** | `{"ge": 1, "le": 600}` emitted; `-1`/`0`/`601` schema-accepted, model-rejected | The range now emits as standard `minimum: 1` / `maximum: 600`; the parity corpus gains `0`, `-1`, and `601`, plus explicit schema-rejects assertions for the bounds |

### The mechanism, for the record

Pydantic can only translate `ge`/`le` markers into JSON Schema `minimum`/`maximum` when they attach to the integer core schema itself; with the `BeforeValidator` wrapping the type and the constraints on the OUTER `Field`, it fell back to dumping the raw marker names into the schema. The fix is one annotated type, `FrameWindow = Annotated[int, Field(ge=1, le=MAX_WINDOW_FRAMES), BeforeValidator(_json_integer)]` — the range markers before the validator in the metadata, which is the ordering pydantic translates correctly, while at runtime the validator still runs first, so the range applies to the already-normalized integer. Verified behavior is unchanged from last round on every prior instance: `5` and `5.0` pass (the float normalized to `5`), `"5"`/`true` refused, and now `0`/`-1`/`601` are refused by schema and model alike. The emitted branch at this head reads `{"maximum": 600, "minimum": 1, "type": "integer"}`.

### Validation at `ea4fb78d`

- Fast suite **1781 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: perf suite **6 passed**.
- Parity corpus: 10 input instances + the explicit reject-direction assertions now include both bounds.
- `gda schema` vs the dev tip `03e47208`: still exactly one entry changed (`perf monitors`).
- GitHub checks on `ea4fb78d`: green (the `Godot e2e tests` job remains dispatch-only by design).

PR #736, opened 2026-08-24
feat(gda): texture projection — inspectable identity for path-less resource values (#666)

### LINKED ISSUE #666: Value projection: inspectable identity for path-less resource values
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-011** (2026-08-08). `game get <TextureRect> --property texture` serialized the value as an opaque string like `():<ImageTexture#-92233...>`. Instance IDs proved two objects differed but could not prove which authored source each represented — especially limiting for `ImageTexture.create_from_image()` resources whose path is intentionally empty.

**Current behavior, stated precisely:** ADR-0035's `Value projection` shipped in the dogfooded version — a `Resource` **with** a `res://` path already projects as a reference `{type, resource_path}`. The observed opaque string is the **string fallback**, which applies to path-less Objects. The real gaps are: (a) path-less Resource values fall to the string fallback; (b) reference projections carry no dimensions.

## What to build

**This requires an ADR-0035 amendment (or superseding ADR) — the projection shapes are public ABI.** The amendment must name the rules it revises:

1. The reference projection shape `{type, resource_path}` is fixed "as public ABI" — adding `width`/`height` to it changes that shape.
2. The two Object projection kinds are discriminated **by the presence of `resource_path`** (the inline value projection deliberately excludes it). A path-less texture projection must therefore **not** emit `resource_path` (not even null) — it needs its own discriminator.
3. ADR-0035 rejected "a curated per-type field table". A `Texture2D`-specific projector (`type`, dimensions, optional digest) is a **new projection kind**, not a whitelist entry — the inline projection emits storage properties, and `Texture2D` width/height are getters while `ImageTexture`'s storage property is a large `image` payload, exactly the live-side risk the whitelist contains.

Concrete deliverable, once the amendment is recorded:
- A texture projection kind for path-less `Texture2D` values: the `type` key (ADR-0035's discriminator field name), width, height.
- A content digest as an **explicit opt-in** (a flag or a named field that stays null unless requested): computing it needs `get_image()` — a GPU-to-CPU readback inside the shared harness projection, not something to pay on every read.
- The former opaque string is retained under a named key (e.g. `object_string`) as secondary diagnostic data.
- The same projection serves headless and live reads (one value shape everywhere, per ADR-0035).

## Acceptance criteria

- [ ] A live `game get` on a path-less `ImageTexture` returns the new projection (`type`, dimensions) and does not emit `resource_path`.
- [ ] With the digest opt-in enabled, two same-class path-less textures with different content are distinguishable; without it, the digest field is null.
- [ ] The same projection serves headless reads.
- [ ] The ADR-0035 amendment naming the three revised rules lands with the change.

## Blocked by

The ADR-0035 amendment decision above.

qinhaihong-red:
## W4 slice: texture projection — inspectable identity for path-less resource values (#666)

**The gap (GDA-DF-011):** `game get` on a path-less `ImageTexture` (e.g. `create_from_image()`, whose `resource_path` is intentionally empty) fell to ADR-0035's string fallback — `():<ImageTexture#-92233…>` — an instance ID that proves two objects differ but cannot say what either shows.

### The ADR-0035 amendment (the issue's blocked-by item)

A dated amendment on ADR-0035 adds a **fourth projection kind** — the **texture projection** — and names the three revised rules the issue requires:

1. the reference shape `{type, resource_path}` stays fixed (no dimensions added to it); a `Texture2D` WITH a `res://` path still projects as a reference — only a path-less one takes the new kind;
2. the new kind carries its own discriminator — the presence of **`object_string`**, which no other object shape emits — and never emits `resource_path` (not even null), so the reference branch stays unambiguous;
3. the per-type field table stays rejected: this is a new projection *kind* (a fixed shape read off getters, like the reference kind), not a whitelist entry — the inline kind emits storage properties, and `ImageTexture`'s storage property is a large `image` payload, exactly the live-side risk the whitelist isolates.

### What shipped

**The shape**: `{"type", "width", "height", "object_string", "digest"}` for a path-less `Texture2D` — class + dimensions off cheap getters, the former `str()` form retained under `object_string` as secondary diagnostics, and `digest` null by default.

**The digest opt-in**: `game get --texture-digest` (wire param `texture_digest`, threaded through the shared projection's recursion so nested textures get it too). Computing it needs `Texture2D.get_image()` — a GPU-to-CPU readback on the live side — so no ordinary read pays it. The digest is `sha256:` over the image's dimensions, format, and raw bytes (the shape prefix keeps same-bytes textures of different shapes from colliding); an image the engine cannot read back keeps null. The flag is live-only today: a path-less texture only exists where runtime code constructed one, so headless reads carry the projection with `digest` always null — stated in the amendment.

**One projection everywhere**: the arm lives in the mirrored shared-coercion block (`operations.gd` ↔ `gda_harness.gd` byte-identical; the mirror test is untouched and green), so the same shape serves the headless reads (`node`/`resource`/`project get`, `project list`, `scene get-exports`, set echoes) and the live `game get`.

**Published, not prose-only**: `TextureProjection` joins `ReferenceProjection` / `InlineValueProjection` as a named model in every projected `value` field's `$defs`, and the shared `VALUE_PROJECTION_DESC` names the kind and its discriminator. `CONTEXT.md`'s `Value projection` term and the command catalog's projection and `game` sections are updated; SKILL.md's `game` row notes the opt-in.

| AC | Evidence |
| --- | --- |
| Live `game get` on a path-less `ImageTexture` returns `{type, dimensions}`, no `resource_path` | e2e: a 2×3 runtime texture projects `{type: ImageTexture, width: 2, height: 3, object_string, digest: null}`; `"resource_path" not in value` asserted |
| Digest opt-in distinguishes same-class different-content textures; null without it | same e2e: red vs blue textures get different `sha256:` digests, a repeat of red gets the identical digest; the plain read asserts `digest is None` |
| The same projection serves headless reads | e2e: `node get` on a scene whose script `_init` constructs a 4×5 path-less texture returns the same shape (digest null) |
| The ADR-0035 amendment naming the three revised rules lands with the change | the dated amendment block on ADR-0035 |

### Validation

- Fast suite **1783 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean; the coercion-mirror test green (both `.gd` files changed byte-identically).
- Real-engine e2e, serial: game + harness-install **6 passed**; the full node suite (103 tests, the whole headless read surface over the changed shared block) **passed**.
- **Red-proof on dev tip `977ae71c`**: the digest-flag test dies on `unknown_option`, both `$defs` tests on the missing `TextureProjection`, `tests/test_models.py` on its import, and both new e2e tests on the old string fallback.
- `gda schema` diff vs the dev tip, entry-by-entry: exactly the ten value-carrying read/set surfaces changed (`game get/set`, `node get/set`, `project get/list/set`, `resource get/set`, `scene get-exports`) — the shared `$defs` + description, plus `game get`'s new flag and its argv-binding metadata. No command added or removed.

Closes #666

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -187,9 +187,9 @@ emits** — the `get` reads (`project`/`node`/`resource get`), the value echoed 
 `node set`/`resource set`, the per-entry value of `project list` and `scene
 get-exports`, and the live `game get` read — so a value reads the same
 everywhere. Two controls keep the shared projection safe on the live side: the
-whitelist bounds the kinds that emit storage properties (inline), and the
-texture kind is safe by construction — a fixed getter shape with its one
-expensive readback behind the explicit digest opt-in (ADR-0035).
+whitelist bounds the Object classes whose storage properties the inline kind
+emits, and the texture kind is safe by construction — a fixed getter shape
+with its one expensive readback behind the explicit digest opt-in (ADR-0035).
 _Avoid_: value rendering, str dump, serialization, descriptor
 
 ### Failure reporting
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -186,8 +186,10 @@ fallback** for anything else. One projection shared across **every value gda
 emits** — the `get` reads (`project`/`node`/`resource get`), the value echoed by
 `node set`/`resource set`, the per-entry value of `project list` and `scene
 get-exports`, and the live `game get` read — so a value reads the same
-everywhere; the whitelist is the boundary that keeps the shared projection safe
-on the live side (ADR-0035).
+everywhere. Two controls keep the shared projection safe on the live side: the
+whitelist bounds the kinds that emit storage properties (inline), and the
+texture kind is safe by construction — a fixed getter shape with its one
+expensive readback behind the explicit digest opt-in (ADR-0035).
 _Avoid_: value rendering, str dump, serialization, descriptor
 
 ### Failure reporting
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -174,10 +174,13 @@ _Avoid_: console output, stdout dump
 The read-side contract for rendering a Godot value into the structured JSON a
 successful result carries. Scalars and the small fixed-shape value types pass
 through directly; a `Dictionary`/`Array` (and the packed-array family) projects
-recursively; and an `Object` is rendered by one of three **projection kinds** — a
+recursively; and an `Object` is rendered by one of four **projection kinds** — a
 **reference projection** for a `Resource` that has a `res://` path (named by type
 and path, never inlined — the read-side mirror of ADR-0033's write-side
-reference), an **inline value projection** for a whitelisted path-less value
+reference), a **texture projection** for a PATH-LESS `Texture2D` (named by type
+and dimensions, with the former `str()` form under `object_string` — its
+discriminator — and an opt-in content `digest`; ADR-0035 amendment, #666), an
+**inline value projection** for a whitelisted path-less value
 `Object` (named by type plus its projected fields), or a plain **string
 fallback** for anything else. One projection shared across **every value gda
 emits** — the `get` reads (`project`/`node`/`resource get`), the value echoed by
--- a/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
+++ b/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
@@ -138,8 +138,21 @@ get-exports`, and the live `game get` read (and any future live value-returning 
 > threads through the shared projection): computing it needs `Texture2D.get_image()`, a GPU-to-CPU
 > readback on the live side, so every ordinary read keeps it `null`. When requested it is
 > `"sha256:"` + the hex digest over the image's dimensions, format, and raw bytes; an image the
-> engine cannot read back keeps `null`. The same projection serves headless and live reads (one
+> engine cannot read back keeps `null`. The field is **required-but-nullable** — the producer
+> always emits the key. The same projection serves headless and live reads (one
 > value shape everywhere); the headless read commands do not expose the opt-in flag today — a
 > path-less texture only exists where runtime code constructed one, so on the headless side the
 > digest field is always `null`. The mirrored shared-coercion block and
 > `tests/test_harness_coercion_mirror.py` are preserved.
+>
+> Two boundary rules complete the kind. **Path-less means EMPTY**: a non-empty, non-`res://`
+> `resource_path` (a `user://` path, `take_over_path`) stays the string fallback it always was —
+> only the empty path takes this kind. **`object_string` is a reserved key**: the inline value
+> projection's exclusion list drops a whitelisted class's own storage property of that name rather
+> than copying it, so the presence-based discrimination cannot be spoofed by a custom
+> `@export var object_string`. And the live-side risk story now has **two controls, stated
+> explicitly**: the inline whitelist remains the boundary for projections that EMIT STORAGE
+> PROPERTIES (this ADR's original rule, unchanged for that kind), while the texture kind is safe by
+> CONSTRUCTION — a fixed shape read off cheap getters, with the one expensive operation
+> (`get_image()`) behind the explicit digest opt-in. The historical whitelist prose above is
+> preserved as written; it now governs the inline kind specifically.
--- a/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
+++ b/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
@@ -111,3 +111,35 @@ get-exports`, and the live `game get` read (and any future live value-returning 
   origin).** The deferred inline sub-resource work of ADR-0033 has a read-side counterpart here: a
   path-less inline value Object is projected (whitelisted), not referenced.
 - **`CONTEXT.md` gains the `Value projection` glossary term** in this same change.
+
+> **Amendment (2026-08-24, #666) — a fourth projection kind: the texture projection.**
+> Dogfooding GDA-DF-011: a path-less `Texture2D` (e.g. `ImageTexture.create_from_image()`, whose
+> `resource_path` is intentionally empty) fell to the string fallback — an instance ID that proves two
+> objects differ but cannot say what either shows. This amendment adds a **texture projection** for
+> path-less `Texture2D` values: `{"type": <Class>, "width": <int>, "height": <int>, "object_string":
+> "<the former str() form>", "digest": <"sha256:…" | null>}`. It revises three rules of this ADR,
+> named explicitly because the shapes are public ABI:
+>
+> 1. **"The reference projection shape `{type, resource_path}` is fixed as public ABI"** — unchanged
+>    in shape, but no longer the only projection a `Texture2D` can take: one WITH a `res://` path
+>    still projects as a reference (dimensions are NOT added to the reference shape); only a
+>    path-less one takes the new kind.
+> 2. **"The two Object projection kinds are discriminated by the presence of `resource_path`"** — the
+>    texture projection carries its own discriminator instead: the presence of **`object_string`**,
+>    which the other object shapes never emit. It does NOT emit `resource_path` (not even null), so
+>    the reference branch stays unambiguous.
+> 3. **"A curated per-type field table — rejected"** — still rejected. The texture projection is a
+>    **new projection kind** (like the reference kind, a fixed shape read off getters), not a
+>    whitelist entry: the inline kind emits storage properties, and `Texture2D`'s dimensions are
+>    getters while `ImageTexture`'s storage property is a large `image` payload — exactly the
+>    live-side risk the whitelist isolates.
+>
+> `digest` is an **explicit opt-in** (`game get --texture-digest`; the wire param `texture_digest`
+> threads through the shared projection): computing it needs `Texture2D.get_image()`, a GPU-to-CPU
+> readback on the live side, so every ordinary read keeps it `null`. When requested it is
+> `"sha256:"` + the hex digest over the image's dimensions, format, and raw bytes; an image the
+> engine cannot read back keeps `null`. The same projection serves headless and live reads (one
+> value shape everywhere); the headless read commands do not expose the opt-in flag today — a
+> path-less texture only exists where runtime code constructed one, so on the headless side the
+> digest field is always `null`. The mirrored shared-coercion block and
+> `tests/test_harness_coercion_mirror.py` are preserved.
--- a/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
+++ b/docs/adr/0035-value-projection-compound-variants-to-structured-json.md
@@ -128,11 +128,13 @@ get-exports`, and the live `game get` read (and any future live value-returning 
 >    texture projection carries its own discriminator instead: the presence of **`object_string`**,
 >    which the other object shapes never emit. It does NOT emit `resource_path` (not even null), so
 >    the reference branch stays unambiguous.
-> 3. **"A curated per-type field table — rejected"** — still rejected. The texture projection is a
->    **new projection kind** (like the reference kind, a fixed shape read off getters), not a
->    whitelist entry: the inline kind emits storage properties, and `Texture2D`'s dimensions are
->    getters while `ImageTexture`'s storage property is a large `image` payload — exactly the
->    live-side risk the whitelist isolates.
+> 3. **"A curated per-type field table — rejected"** — superseded **narrowly, for this one type**,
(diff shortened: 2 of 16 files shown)


qinhaihong-red:

Reviewed independently along four axes (Markdown artifact, domain-modular architecture, Standards, and Spec) at exact head `3d9a1911859d7ac2134a618701bf8d33a60a395d` against base `977ae71c25f3dfd705a467af90ff09a7d4c7bfd0` and issue #666.

### Required changes

1. **Keep the texture projection inside the specified path-less boundary.**

   `operations.gd:5494-5508` (and the mirrored harness block) first handles only `res://` references, then projects every remaining `Texture2D`; it never checks that `resource_path` is empty. A real Godot probe using `take_over_path("user://owned-texture")` therefore returned a texture projection instead of the pre-existing string fallback. This changes non-empty, non-`res://` resources outside issue #666's path-less scope. Require an empty `resource_path` in the texture arm and add a `user://` regression test.

2. **Make `object_string` an actually exclusive discriminator.**

   ADR-0035:127-130 and the public descriptions say no other Object projection emits `object_string`, but `JSONIFY_BOOKKEEPING_PROPS` does not reserve it and the inline `InputEvent` arm copies arbitrary storage properties (`operations.gd:5428-5434,5533-5545`; mirrored in `gda_harness.gd`). A real custom path-less `InputEventAction` with `@export var object_string` emitted that key while lacking `width`, `height`, and `digest`, so a presence-based consumer misclassifies it as a texture. Reserve/exclude this key from inline projections, document the collision rule, and add a custom-InputEvent regression test.

3. **Publish `digest` as required-but-nullable.**

   The runtime producer always emits `digest`, and issue #666/ADR-0035 define the fixed shape with the field present and `null` when the opt-in is absent or readback is unavailable. However, `TextureProjection.digest` has `default=None` (`models.py:679-686`), so rendered `game get --schema` omits `digest` from `TextureProjection.required` and accepts a payload with the key missing. Remove the default while retaining `str | None`, and assert both required presence and nullable value in the schema tests.

4. **Reconcile the ADR's risk-boundary decision.**

   `CONTEXT.md:189-190` and ADR-0035:62-65 still describe the whitelist as the live-side risk-isolation boundary. The amendment then adds a fixed texture projector outside the inline whitelist, while issue #666's triage decision says the rejected per-type table is “superseded narrowly” and the amendment says it remains rejected. Preserve the historical decision text, but make the amendment and glossary state the current two controls explicitly: the inline Object whitelist, and the fixed Texture2D getter shape plus explicit digest opt-in.

5. **Bump the bundled harness version.**

   `gda_harness.gd` changed, but `HARNESS_VERSION` remains `"7"` in both base and head (`harness/install.py:86-90`). This violates the file's “bumped when the bundled harness changes” rule and ADR-0018's version self-sync contract: two different harness capabilities now advertise the same version. Bump it and verify the install/sync receipt.

### Validation

- 219 related fast tests and the two targeted real-Godot live/headless E2E tests passed in the independent review; the main review also reran the focused model/schema/mirror/harness suites and both E2E paths successfully.
- The `user://`, custom-`InputEvent.object_string`, and missing-required-`digest` counterexamples were reproduced on the exact head.
- The shared coercion mirror, Ruff, Pyright, and `git diff --check` pass.
- GitHub's Python/build checks are green. The GitHub **Godot e2e tests** job is skipped, so it is not evidence for the engine paths above; those paths were run locally with the real engine.

No additional substantive ASD-STE100 terminology or prose issue was found outside the document findings above.

qinhaihong-red:
## Review response — all five findings adopted at head `16327c11`

Baseline: base tip / merge base `977ae71c`, reviewed head `3d9a1911`, new head `16327c11`. Your three engine probes were confirmed by red-proofing the new regressions on the reviewed head — each fails there with exactly the reported mechanism (the `user://` texture carrying `'object_string': '(user://own…'` into a texture projection; the custom event leaking `object_string` into its inline projection; `required` missing `digest`).

| Claim | Decision | Change |
| --- | --- | --- |
| 1. The texture arm escapes the path-less boundary (`take_over_path("user://…")`) | **Fully adopt** | The arm requires an EMPTY `resource_path`; a non-empty, non-`res://` path stays the string fallback it always was; `user://` e2e regression |
| 2. `object_string` is not actually exclusive | **Fully adopt** | The key is RESERVED: the inline projection's exclusion list drops a whitelisted class's own storage property of that name; collision rule documented in the ADR, the model, and the mirrored block; custom-`InputEvent` e2e regression |
| 3. `digest` should be required-but-nullable | **Fully adopt** | `default=None` removed (type stays `str \| None`); the schema test asserts the full `required` set AND the null branch |
| 4. The risk-boundary prose contradicts the amendment | **Fully adopt** | The amendment and the `CONTEXT.md` glossary now state the two controls explicitly — the inline whitelist bounds the kinds that emit storage properties; the texture kind is safe by construction (fixed getter shape, the one expensive readback behind the explicit opt-in) — with the historical whitelist prose preserved as written |
| 5. `HARNESS_VERSION` not bumped | **Fully adopt, with a disclosure** | Bumped to `"8"`, plus a new pin guard so the omission can never be silent again |

### The version bump, and what the record should say

Your finding is narrower than the actual drift, so the record gets the whole of it: the harness body also changed in #732 (the input gestures + the viewport-enter mirror) and #735 (the `perf-sample` op), and neither bumped the version — so THREE different harness capabilities have advertised `"7"` on the dev branch. Every pre-W4 harness change did bump (the constant walked 0→7 one slice at a time; the last was #681). Two mitigating facts, verified in `install.py` rather than assumed: the sync decision is a CONTENT compare — "a version bump (or a body change) is a content difference — re-materialize only then" — so no project ever ran a stale harness; and the dev branch is unpublished, so one bump to `"8"` honestly covers the wave's accumulated changes. The receipt path is already pinned by the existing install tests, which assert against the `HARNESS_VERSION` constant rather than a literal, so the bump rode through them without edits.

### The boundary and the reservation, in the shared block

Both fixes live inside the mirrored shared-coercion block (both `.gd` files byte-identical, mirror test green):

- the texture arm is now `value is Texture2D and String(value.resource_path).is_empty()`, with the comment naming what stays outside — `user://` and `take_over_path` were the string fallback before this PR and remain it;
- `object_string` joins the inline projection's exclusion list, documented there as a RESERVED discriminator key (a different reason than the base-bookkeeping entries, and the comment says so), in the ADR amendment ("cannot be spoofed by a custom `@export var object_string`"), and on `InlineValueProjection`'s docstring.

### A pin guard, so this cannot recur silently

Beyond the bump (a maintainer request during this round): `tests/test_harness_install.py` now pins `(PINNED_HARNESS_VERSION, PINNED_HARNESS_SHA256)` against the bundled `gda_harness.gd` bytes. The guard fails in both drift directions with its own instruction — bytes changed without a bump (the message carries the new hash and the two-step fix), or a bump without a re-pin. The root cause of the three silent omissions was that the sync path's content compare never NEEDED the version, so nothing failed; the pin turns that silence into a fast-suite red. Mutation-proofed: one appended byte turns the guard red.

### Validation at `16327c11`

- Fast suite **1784 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean; the coercion mirror green.
- Real-engine e2e, serial: game + harness-install **6 passed** (the install suite exercises the version receipt against the bumped constant); the full node suite rerun over the re-changed shared block: **105 passed** (103 + the 2 new regressions).
- Red-proof on `3d9a1911`: all three new regressions fail there with your reported mechanisms.
- `gda schema` deltas beyond the previous head: `TextureProjection.required` gains `digest`; the inline-projection description gains the reservation sentence. No command added or removed.

qinhaihong-red:
## Re-review at `16327c11` — changes still required

This was a residue-only re-review of the five original findings, pinned to previous head `3d9a1911`, current head `16327c11`, and base / merge base `977ae71c`.

### Remaining required changes

1. **Finish reconciling ADR-0035 with issue #666's pinned triage decision.**

   ADR-0035:131-135 still says the curated per-type field table is **“still rejected”**, while the issue's triage decision says that rejection is **“superseded narrowly, for this one type.”** The new two-control explanation fixes the risk-boundary part of the original finding, but it does not resolve these opposite decision statements. The new `CONTEXT.md:190` wording also says “the whitelist bounds the kinds that emit storage properties”; in the glossary, `projection kind` already means the four output shapes, while the whitelist actually bounds the Object classes whose storage properties the inline kind emits.

   Minimal fix: state that the general curated-table mechanism remains rejected while the fixed `Texture2D` projector is the triage decision's one narrow exception, and change the glossary wording to something like “the whitelist bounds the Object classes whose storage properties the inline kind emits.”

2. **Update the remaining public schema descriptions to the four-kind / two-control contract.**

   The generated `game get --schema` is internally inconsistent:

   - `NodeProperty`'s description (`models.py:720-722`) still lists only `ReferenceProjection / InlineValueProjection / str()` and omits `TextureProjection`.
   - `GameGetResult`'s description (`commands/game.py:122-127`) still describes the whitelist as the boundary that keeps the whole shared live projection safe.

   The same schema's `value` description and `$defs` correctly publish `TextureProjection`, so these two enclosing descriptions should be updated to the same four projection kinds and should scope the whitelist boundary to inline storage-property projection.

### Closed findings

- The texture arm now requires an empty `resource_path`; the real-Godot `user://` counterexample stays on the string fallback.
- `object_string` is reserved from inline producer output; the custom-`InputEvent` real-Godot regression passes and the mirrored blocks remain byte-identical.
- `digest` is required-but-nullable in the model and rendered schemas.
- `HARNESS_VERSION` is now `"8"`; install/version receipt tests pass.
- The new `(version, SHA-256)` harness pin detects both bundled-byte drift and version drift and is proportionate to the repeated missed bumps.

### Axis results and validation

- Markdown Artifact review: **changes required** — 1 remaining authority/terminology finding.
- Domain-modular architecture: **pass** — no remaining runtime ownership, dependency, or producer-orthogonality finding.
- Standards: **pass** — no remaining hard standard violation or material smell.
- Spec: **changes required** — the two public-contract residues above.

Focused exact-head validation: **10 passed**, including real-Godot path-less texture, `user://` fallback, custom `object_string`, and live digest coverage. Current Python/build CI is green; GitHub's Godot E2E job is still skipped, so the engine-path evidence above is local exact-head evidence.

qinhaihong-red:
## Recheck response — both residues closed at head `57d5406c`

Baseline: base tip / merge base `977ae71c`, rechecked head `16327c11`, new head `57d5406c`.

| Claim | Decision | Change |
| --- | --- | --- |
| 1. The amendment contradicts the pinned triage wording; the glossary says the whitelist bounds "kinds" | **Fully adopt** | Rule 3 now states both halves the triage decision pinned: the GENERAL curated-table mechanism remains rejected, and the fixed `Texture2D` projector is its ONE narrow supersession, recorded by the dated note. The glossary says the whitelist bounds the Object CLASSES whose storage properties the inline kind emits |
| 2. Two enclosing schema descriptions still carry the three-kind / whole-projection-whitelist story | **Fully adopt** | `NodeProperty` lists all four kinds; `GameGetResult` names each kind's shape and scopes the whitelist to inline storage-property projection, with the texture kind's construction-based safety stated |

Both are wording-only: no value behavior, no `$defs`, and no field shape moved — the schema deltas are exactly the two description strings the finding names (they ride into every command whose result carries a `NodeProperty`, which is the same-text-everywhere property the shared description exists for).

### Validation at `57d5406c`

- Fast suite **1784 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- The engine-path evidence is unchanged from `16327c11` (no `.gd` or behavior change in this round): node **105 passed**, game + harness-install **6 passed**, mirror green, all serial with the real engine.
- GitHub checks on `57d5406c`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Second-recheck response — the last residue is closed at head `56c420a6`

Baseline: base tip / merge base `977ae71c`, rechecked head `57d5406c`, new head `56c420a6`.

| Claim | Decision | Change |
| --- | --- | --- |
| The `VALUE_PROJECTION_DESC` authority comment still counts "two named projection models" | **Fully adopt** | "the three named projection models (ReferenceProjection / TextureProjection / InlineValueProjection)", exactly as suggested |

One line of comment prose; no schema, model, or behavior change. A sweep for the same staleness elsewhere (`two named projection`, the old two-model pairing) found no other occurrence in `src/`, `docs/`, or `CONTEXT.md`.

### Validation at `56c420a6`

- Fast suite **1784 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Engine-path evidence unchanged from `16327c11` (no `.gd` or behavior change since): node **105 passed**, game + harness-install **6 passed**, mirror green.
- GitHub checks on `56c420a6`: green.

PR #738, opened 2026-08-24
feat(gda): resource import — scoped import surface for clean-worktree loading (#668)

### LINKED ISSUE #668: Scoped import surface for clean-worktree resource loading

qinhaihong-red:
## W4 slice: `resource import` — scoped import surface for clean-worktree loading (#668)

**The gap (GDA-DF-010):** a clean worktree carries the sources and their committed `.import` sidecars but not the gitignored `.godot/` cache, so a one-shot run's `preload()` of a PNG dies with "no recognized resource loader". **The anti-goal (GDA-DF-038):** the blunt fix — a project-wide editor reimport — staged 2,148 entries of unaccounted tool noise.

### The contract revision (maintainer decision, 2026-08-24)

The scoped-import deliverable was revised on the issue ([the authoritative comment](https://github.com/aigengame/godot-agent/issues/668#issuecomment-5395962383)) per the maintainer's option-(a) decision, and [corrected on 2026-08-25](https://github.com/aigengame/godot-agent/issues/668#issuecomment-5404590394) after the re-review's engine evidence: gda's scoping is in the decision — the pass runs only when a requested asset's evidence state is `missing` or `stale` (judged from the engine's own artifacts; an `invalid` request settles `failed` without a pass, because the engine skips previously failed imports; a fully cached request starts no engine) — and in the report (everything the project-wide pass touched, classified), with the dry run covering the decidable inventory (`pass_will_also_import`, the corrected field name) and declaring the engine-decided remainder. This PR implements that corrected contract.

### The triage decisions, honored

Both were pinned on the issue before implementation and are followed: the command is **`resource import`** under the existing `resource` group (no near-synonym `asset` group), and the **importer-execution point joins CONTEXT.md's `Project-code execution surface`** — engine importer code (plus any import plugins the project registers) running over project content, within the `Trusted project` assumption (ADR-0009), no new trust axis.

### The mechanism, verified against the engine source

The engine's ONE scriptable import primitive is the project-wide `godot --headless --import` pass: a per-file reimport exists only inside the editor process (`EditorFileSystem.reimport_files`), and `--script` requires a MainLoop (`main.cpp`, verified), so no headless CLI form reaches it. **gda's scoping is therefore in the decision and the report**, stated plainly in the contract:

- **The decision**: each requested asset's EVIDENCE STATE is read from the same project artifacts the engine's own reimport test (`EditorFileSystem::_test_for_reimport`) reads — with `cached` requiring positive ARTIFACT-level evidence, and the engine-state checks (importer availability and format version, settings validity, the editor cache's expected sidecar MD5) as the declared remainder, whose direction is one-sided: it can delay a re-import until the next pass, never spend a pass the engine would not — `cached` (every check passes: the PATH-derived `.md5` receipt present — located at `<filename>-<md5 of the res:// path>`, as the engine locates it — with `source_md5`/`dest_md5` matching the bytes, any declared destinations present, `source_file` naming this asset, the UID-era format; a sidecar declaring no destinations with a matching receipt passes the same artifact checks, and the engine's own pass leaves it untouched when the declared remainder is controlled — proven natively in e2e), `missing` (no sidecar), `stale` (a check fails — the engine WOULD re-import), `invalid` (a `valid=false` or unparseable sidecar, or a `.md5` receipt outside gda's documented engine-written assignment subset — the engine deliberately SKIPS parse errors to avoid a reimport loop, and gda conservatively skips unsupported receipt syntax; the subset accepts quoted-string assignments, whitespace, `;` comments, JSON-style escapes with lone UTF-16 surrogates excluded, and repeated assignments with the engine's last-value-wins order; delete the sidecar to retry, which heals a malformed receipt too, because the pass rewrites both). The pass runs only for `missing`/`stale`; an `invalid` request settles `failed` without spending a pass, and a fully cached request runs no engine at all. Every non-cached request — missing, stale, or invalid — follows the same settlement path after the pass decision.
- **The report**: after the pass, each requested non-cached asset (missing, stale, or invalid) re-verdicts to `imported` / `not_importable` (the ENGINE decided the type needs no import — e.g. a script; gda keeps no extension table) / `failed`, and `created` lists **every file the pass created**, each classified against the explicit `cache_root`: `cache_owned` (under `res://.godot`) vs `source_adjacent` (`.import` and `.uid` sidecars — the DF-038 noise, accounted file by file, incidental project-wide imports included).
- **`--dry-run`**: the evidence states, the requested assets' sidecars-to-be, and `pass_will_also_import` — the OTHER stale assets the project-wide pass will re-import (invalid ones excluded: the engine skips them, proven byte-for-byte in e2e) — writing nothing, asserted by a full before/after tree comparison. The engine's hash-named cache files and no-sidecar/`.uid` generation stay engine-decided; the real run's `created` is the authoritative inventory. The human render carries the same prediction.
- **Failures** ride the shared `classify_launch_or_crash` prefix (`launch_timeout`, `binary_not_found`, `engine_crashed`); a nonzero pass exit is `operation_failed` with the engine's stderr.
- The result model **validates its mode field sets** (the #732 lesson applied at birth): a dry run reporting created files, a real run carrying predictions or `missing`, or a summary disagreeing with its lists all fail validation.
- **Plain `script run` is untouched** — and proven: the e2e asserts the failing pre-import run added no import artifacts.

| AC | Evidence |
| --- | --- |
| Clean project: preload fails without the import, succeeds after | e2e: `script run` exits 1 ("no recognized resource loader") → `resource import` → the same run exits 0 with `LOADED` |
| Per-asset hit/miss + machine-readable summary | unit + e2e: `assets[].status`, `summary` with per-status and per-classification counts, cross-validated by the model |
| Dry run reports the inventory, writes nothing | Met per the revised contract ([issue comment, 2026-08-24](https://github.com/aigengame/godot-agent/issues/668#issuecomment-5395962383)): the decidable inventory (requested sidecars-to-be + the project-wide `pass_will_also_import` sidecar scan), with the engine-decided remainder declared; zero-write asserted by tree comparison |
| Every created file classified | e2e: `.godot/**` all `cache_owned`; `icon.png.import` and `check.gd.uid` `source_adjacent` |
| Plain `script run` unchanged | code untouched; e2e asserts no import artifacts from the failing run; adjacent script e2e suite green |

### Validation

- Fast suite **1824 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: resource-import **9 passed** (the DF-010 heal, dry-run zero-write, classified inventory, idempotent second run, engine-decided `not_importable`, the real-engine invalid/stale/alias/skip proofs, and four native-parity proofs — the no-destination sidecar the engine's own pass leaves byte-identical with the cache remainder controlled, the malformed receipt and the lone-surrogate receipt its parse-error branch deliberately skips, and the duplicate-assignment receipt its last-value-wins order keeps current); adjacent resource + script suites **99 passed**.
- **Red-proof on dev tip `c7f6e4c0`**: 15/15 unit tests and both e2e tests fail (`unknown_command` / `ImportError`; the no-assets usage test additionally pins the `ASSETS` metavar so it cannot pass by coincidence on a head without the command).
- `gda schema` diff vs the dev tip, entry-by-entry: **added `resource import`; nothing else moved.**
- Docs synced: command catalog (a dedicated section), SKILL.md surface table, CONTEXT.md's execution-surface term, all four READMEs (+ i18n re-stamp).

Closes #668


qinhaihong-red:

Reviewed exact head `95b593ce01963ac571b86ea288248a1e4d9e00b7` against issue #668, its triage decision, the root context/ADRs, and the public help/schema. I ran the Markdown artifact, domain-modular architecture, Standards, and Spec reviews independently before adjudicating them.

### Blocking findings

1. **[P1] The operation does not satisfy the scoped-import or full-dry-run requirements.**

   Issue #668 requires importing only the requested dependencies and requires dry-run to report the full mutation inventory. `run_resource_import_operation()` predicts only requested `.import` sidecars in dry-run (`src/gda/commands/resource.py:744-756`), but a real miss invokes the unfiltered project-wide `godot --path PROJECT --import` pass (`:759-769`).

   Reproduction on this head with `icon.png` and `other.png`, requesting only `icon.png`:

   - dry-run predicted only `res://icon.png.import`;
   - the real run also created `other.png.import`, the unrequested image cache/MD5 files, and the remaining `.godot/**` inventory.

   The committed E2E contains the same smaller counterexample: dry-run predicts only `icon.png.import`, while the real pass also creates the unrequested `check.gd.uid` (`tests/test_e2e_resource_import.py:104-128`). Therefore the PR body's acceptance table must not mark the original dry-run AC as satisfied. Either implement an actually scoped mechanism with a complete plan, or first change and accept the issue contract before synchronizing the PR/docs.

2. **[P1] A failed import is reported as successfully imported.**

   `_asset_state()` treats every sidecar without `dest_files` as cached (`resource.py:659-662`), without checking `valid=false`; the post-pass code then rewrites that state to `imported` (`:800-806`).

   Exact-head real-engine reproduction: copy ordinary text to `broken.png`, then run `gda resource import res://broken.png`. Godot 4.6.3 writes `broken.png.import` with `valid=false` and no `dest_files`, but gda exits 0 with `status="imported"` and `failed=0`.

   Godot is the authority for import validity/freshness. Its decision also covers UID, importer/version/settings, source and destination MD5, and other invariants; the Python destination-exists heuristic creates a second, incomplete knowledge owner. At minimum, `valid=false` must be `failed`; the durable fix should obtain or faithfully adapt the engine-owned verdict and add real-engine invalid/stale/missing-MD5 regressions.

3. **[P2] `res://` input can escape the project boundary.**

   `_asset_res_path()` checks containment only for filesystem-form input (`resource.py:635-647`). On this head, `resource import res://../../../pyproject.toml --dry-run` accepts and reports the outside file as a project asset. This conflicts with the parameter contract, ADR-0006's single path-normalization rule, ADR-0009's Trusted-project scope, and ADR-0015's params-model authority. Reuse a shared canonical project-scoped path gate and cover root/`..`/other-scheme/symlink cases.

4. **[P2] The public execution kind contradicts the actual channel.**

   `RESOURCE_IMPORT_COMMAND` has a native `--import` recipe but leaves `kind` at its default `headless` (`resource.py:854-864`), which the taxonomy defines as the `operations.gd` sentinel pipeline (`src/gda/execution.py:19-39`). The per-command schema and aggregate manifest therefore advertise the wrong channel, and `tests/test_resource_import_commands.py:280-286` currently pins that contradiction. Add an accurate kind, or revise the taxonomy through the appropriate architecture decision.

### Required documentation corrections

- `CONTEXT.md:239-257` and the four README trust sections still say autoloads run on every `--project` operation. A cached import starts no engine; a missing import runs the editor import pass but did not execute an autoload `_init()` write-file probe. The README sections also omit project-wide importer/import-plugin execution. Document cached and missing import as distinct execution points, and separate the import pass from the scene-preflight sentence.
- `README.md:499-508`, the translated README rows, and `src/gda/skill/SKILL.md:129` still describe the whole `resource` group as `.tres` files, although `import` accepts PNG/SVG/GDScript and other project assets. Widen the group description or state the per-command scopes.

### Standards judgement call

- `make_launch` (`resource.py:704-710`) is a second launch seam with no caller, while all added tests patch the late-bound module `launch`. Prefer one seam: remove the unused parameter, or make direct injection the sole tested mechanism.

### Validation

- Local exact-head checks: **34 passed** — 15 resource-import unit tests, 2 real-Godot E2E tests, and 17 README-i18n/Skill-surface/descriptor-registry tests.
- `git diff --check`: passed.
- Additional real-engine probes reproduced the unscoped import, false-success import, and `res://` escape above.
- Current Python, ruff, and pyright CI checks are green. The GitHub Godot E2E job is skipped; green CI does not cover these counterexamples.

No other substantive terminology or prose issue was found. English technical prose was reviewed against ASD-STE100; Spanish/Japanese native-language polish was not assessed.

qinhaihong-red:
## Review response — five findings adopted at head `a8cf673b`; finding 1's scoped half needs a maintainer decision

Baseline: base tip / merge base `c7f6e4c0`, reviewed head `95b593ce`, new head `a8cf673b`. Every mechanism was reproduced first: your text-as-PNG probe reported `imported` on the reviewed head; `res://../` with an EXISTING outside file was accepted; and all nine new regressions (7 unit + 2 real-engine e2e) are red on `95b593ce`, each on its reported mechanism.

| Claim | Decision | Change |
| --- | --- | --- |
| 1. [P1] Scoped import + full dry-run inventory not satisfied | **Partially adopt now; the rest is the maintainer's call** (below) | The dry-run half ships: a project-wide sidecar scan lists `pass_also_missing`; the scoped half needs an issue-contract decision |
| 2. [P1] A failed import reported as imported | **Fully adopt** | The ENGINE's verdicts are adopted wholesale: `valid=false` is never a hit and settles `failed`; cache hits also require the engine's `.md5` receipts (missing receipt or a source hashing differently → `missing`, exactly what the engine itself would re-import) |
| 3. [P2] `res://` input escapes the project | **Fully adopt** | One canonical containment gate for BOTH input forms: resolve + `relative_to`, covering `..`, absolute-outside, and symlinks pointing out |
| 4. [P2] The published kind claims the sentinel pipeline | **Fully adopt** | `ExecutionKind.IMPORT` (the `SCRIPT_RUN` self-description precedent: routes by recipe, adds no runner branch); the aggregate/manifest pins grow to five |
| 5. Documentation corrections | **Fully adopt** | Cached and missing imports are DISTINCT execution points (cached: no engine at all; miss: the editor importer path, NO autoloads) in `CONTEXT.md` and all four README trust sections; the `resource` group description covers project assets, not only `.tres` |
| 6. Second launch seam | **Fully adopt** | The caller-less `make_launch` parameter is gone; one seam (module-global `launch`, patched as the scene/script channels are) |

### Finding 2, the durable half

"Obtain or faithfully adapt the engine-owned verdict" is what shipped: the verdicts gda reads are all ENGINE-written — the sidecar's `valid=false`, the declared `dest_files`, and the `.md5` receipts' `source_md5` (verified equal to `md5(source bytes)` against a live engine before implementing). gda adds no judgment of its own on top; where a receipt is absent it answers `missing` because that is what the engine itself would re-import. Regressions: real-engine invalid (`text-as-.png` → `failed`, stable across a second run) and stale (changed source → `missing` → re-`imported`); fake-seam invalid / stale / missing-md5.

### Finding 1 — what ships now, and the decision I cannot make for you

**The dry-run half ships.** The dry run now scans every committed sidecar in the project and reports `pass_also_missing`: the assets OUTSIDE the request whose declared cache is missing, invalid, or stale — your `icon.png`+`other.png` probe is pinned as an e2e (request `icon.png`, the scan names `res://other.png`). What remains engine-decided — assets with NO sidecar, and generated `.uid` files — is named in the contract, with the real run's `created` list as the authoritative inventory. The PR body's acceptance table is corrected accordingly (the dry-run AC is met for the decidable inventory, with the engine-decided remainder declared, not claimed).

**The scoped half needs the issue contract decided.** The mechanism evidence (verified against the engine source and a live engine): the ONE scriptable import primitive is the project-wide `--headless --import`; `EditorFileSystem.reimport_files` exists only inside the editor process, and `--script` requires a MainLoop, so no headless CLI form reaches a per-file import. The two honest paths:

- **(a) Revise the issue contract** to what the engine can do: "runs the pass only when a requested asset's cache is missing, and accounts for everything the project-wide pass touched" — which is what this PR implements and reports. My recommendation: the constraint is the engine's, not gda's, and simulating per-file scope would require the alternative below.
- **(b) A temporary-subproject mechanism**: copy the requested assets + sidecars + the import-relevant project settings into a scratch project, run the pass there, transplant `.godot/imported/*` back. It achieves literal scope but makes gda a second owner of the engine's cache-layout knowledge (hash naming, settings inheritance, uid caches) — the same second-knowledge-owner defect this review's finding 2 just had me remove — and breaks silently on engine-internal changes.

I have not changed the issue; that revision is a maintainer act. Until it lands, the PR body marks the scoped-import deliverable as **open, pending that decision**, rather than satisfied.

### Validation at `a8cf673b`

- Fast suite **1806 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: resource-import **4 passed** (heal, dry-run zero-write, invalid, stale + project-gap scan).
- Red-proof on `95b593ce`: 7 unit + 2 e2e regressions fail there (`imported`≠`failed`, `cached`≠`missing` ×2, escapes accepted, `pass_also_missing` absent, kind `headless`).
- `gda schema` vs the dev tip: still exactly one entry added (`resource import`), now `kind: "import"`.
- GitHub checks on `a8cf673b`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
The scoped-import contract question from the review is resolved: the maintainer chose option (a), and the revision is recorded on the issue — https://github.com/aigengame/godot-agent/issues/668#issuecomment-5395962383. The PR body's acceptance table now cites it; head unchanged at `a8cf673b` (the implementation already matches the revised contract). Ready for re-review.

qinhaihong-red:
## Exact-head re-review — changes required

Re-reviewed exact head `a8cf673bd3d1b2e2cc45abedfe25013398a11fb8` against issue #668, including the authoritative [2026-08-24 contract revision](https://github.com/aigengame/godot-agent/issues/668#issuecomment-5395962383). I reran the Markdown artifact, domain-modular architecture, Standards, and Spec axes independently.

The contract revision closes the old literal per-file-scope objection. The original `valid=false -> imported`, `res://../` escape, incorrect `headless` kind, and redundant `make_launch` seam are also fixed. The following residue remains.

### Artifact review — changes required

1. **The PR body and command catalog still describe the old cache and dry-run contract.**

   PR body lines 17–19 and [`docs/command-catalog.md:702-714`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/docs/command-catalog.md#L702-L714) still say that existing `dest_files` alone make a hit and that other missing assets' sidecars are wholly engine-decided. The revised contract and current implementation also use `valid`, `.md5` receipts, and `pass_also_missing`; only assets with no sidecar and generated `.uid` files belong to the declared remainder.

2. **The Project-code execution documentation remains contradictory and one translation is unsynchronized.**

   - [`CONTEXT.md:239-257`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/CONTEXT.md#L239-L257) still says autoloads run on “every `--project` op”, then identifies import as an exception.
   - [`docs/README.es.md:654-677`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/docs/README.es.md#L654-L677) retains the old universal autoload claim and omits the import execution point entirely.
   - All four README summaries still say resolving a project runs project code, although a fully cached import starts no engine. The statement needs to be conditional on the operation/channel.

3. **The bundled Skill incorrectly says scripts are ensured into the import cache.**

   [`src/gda/skill/SKILL.md:129`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/src/gda/skill/SKILL.md#L129) includes scripts among assets that `resource import` ensures into the cache, while the implementation reports scripts as `not_importable`.

Minor: PR body validation counts are stale. It says `1800 passed` and 2 resource-import E2E tests; current CI reports `1804 passed, 2 skipped, 514 deselected`, and this head contains 4 resource-import E2E tests.

Terminology/prose assessment: the resource-group scope is now correctly wider than `.tres`; no other independent English prose issue was found. Spanish/Japanese native-language polish was not assessed.

### Domain-modular architecture

1. **[P1] `ARC-738-FRESHNESS` remains open: `_asset_state()` is still not a faithful adapter of Godot's freshness verdict.**

   [`resource.py:670-719`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/src/gda/commands/resource.py#L670-L719) checks `valid=false`, destination existence, and `source_md5`, but omits `source_file`, UID, importer identity/version/settings, destination digest, and other invariants Godot owns.

   Real Godot 4.6.3 counterexample on this head:

   - import `icon.png`, then copy both the source and its `.import` sidecar to `alias2.png`;
   - gda returns `engine_pass:false`, `status:"cached"`, `created:[]`, while the copied sidecar still says `source_file="res://icon.png"` and points at icon's cache;
   - native `godot --headless --import` immediately reimports `alias2.png` and rewrites its source/destination paths.

   Therefore a false cache hit suppresses a necessary engine pass. Godot's 4.6 implementation also checks importer/version/settings and source/destination MD5 in [`EditorFileSystem::_test_for_reimport`](https://github.com/godotengine/godot/blob/4.6/editor/file_system/editor_file_system.cpp#L3270-L3376). The same incomplete verdict causes false negatives in `pass_also_missing`.

2. **[P1] `ARC-738-PASS-INVENTORY` is open: evidence state and pass policy are collapsed into one `missing` status.**

   [`_project_import_gaps()`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/src/gda/commands/resource.py#L734-L758) lists every `_asset_state(...).status == "missing"`, including `valid=false`. Exact-head real-engine counterexample: an unrelated `bad.png.import` with `valid=false` appears in `pass_also_missing`, but the subsequent project-wide pass leaves that sidecar's SHA unchanged. Godot deliberately skips automatic reimport of a prior failed import. Preserve `missing` / `invalid` / `stale` as distinct evidence states, then derive the requested-pass decision and project-pass prediction separately.

3. **[P2] `ARC-738-KIND-AUTHORITY` is open.**

   [`ExecutionKind.IMPORT`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/src/gda/execution.py#L34-L46) now describes runtime behavior correctly, but the fifth public `--schema`/manifest enum value is absent from the persistent architecture authorities. ADR-0004 and ADR-0012 still enumerate the earlier taxonomy; ADR-0031 explicitly treated `script_run` as a coordinated public ABI/schema migration. Record `IMPORT` through an ADR amendment or equivalent architecture decision and synchronize the public taxonomy.

Stable claim status: `ARC-738-SCOPE`, `ARC-738-PATH`, the behavioral part of `ARC-738-KIND`, and `ARC-738-LAUNCH-SEAM` are closed. `ARC-738-FRESHNESS`, `ARC-738-PASS-INVENTORY`, and `ARC-738-KIND-AUTHORITY` remain open.

### Standards

Four hard violations remain:

1. [`_asset_res_path()`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/src/gda/commands/resource.py#L648-L667) builds a second path-normalization/containment authority after `NormalizedPath` model construction instead of extending/reusing the ADR-0006/ADR-0015 shared project-path rule. It also changes the established treatment of symlinked project paths.
2. The fifth public execution kind is not synchronized with its ADR/public-ABI authorities.
3. The generated output schema and public documentation still describe the old destination-exists verdict and old project-wide remainder; the human renderer also omits `pass_also_missing`.
4. The authoritative `Project-code execution surface` contains the universal `--project` claim and its import exception simultaneously.

The earlier `make_launch` Speculative Generality finding is closed. No additional baseline smell remains.

### Spec

The revised scope contract is implemented, but three actionable Spec findings remain:

1. **[P1]** stale-cache detection can return a real false hit and suppress the required pass (`source_file` reproduction above);
2. **[P1]** `pass_also_missing` can claim that the pass will import an invalid asset that Godot actually skips;
3. **[P2]** [`render_resource_import()`](https://github.com/aigengame/godot-agent/blob/a8cf673bd3d1b2e2cc45abedfe25013398a11fb8/src/gda/commands/resource.py#L899-L925) never renders `pass_also_missing`, so a default non-JSON dry run does not report the revised contract's project-wide decidable inventory.

### Verification

- Exact-head targeted suite: **96 passed** (21 resource-import unit, 4 real-Godot resource-import E2E, README/Skill synchronization, and aggregate-schema tests).
- `git diff --check`: passed.
- `resource import --schema` and aggregate manifest: `kind: "import"`, with argv bindings intact.
- Current Python, ruff, and pyright CI checks are green; the GitHub Godot E2E job remains skipped.
- Worktree stayed clean; no repository files were changed by this review.

qinhaihong-red:
## Re-review response — all findings adopted at head `3327280a`

Baseline: base tip / merge base `c7f6e4c0`, re-reviewed head `a8cf673b`, new head `3327280a`. Every mechanism was reproduced before the fix — your alias probe (`source_file` naming another source → `cached`), the invalid neighbor's untouched sidecar bytes after a pass, and the escapes — and all eight new regressions (7 unit + 1 real-engine e2e) are red on `a8cf673b`, each on its reported mechanism.

| Claim | Decision | Change |
| --- | --- | --- |
| [P1] `ARC-738-FRESHNESS` | **Fully adopt** | `_asset_state` is a faithful adaptation of `EditorFileSystem::_test_for_reimport`, in the engine's own order (details below) |
| [P1] `ARC-738-PASS-INVENTORY` | **Fully adopt** | Evidence states (`cached`/`missing`/`stale`/`invalid`) are separate from settlements and from pass policy; `pass_will_also_import` lists `stale` only |
| [P2] `ARC-738-KIND-AUTHORITY` | **Fully adopt** | ADR-0004 and ADR-0012's outcome notes record both self-description-only enum growths (`script_run`, `import`) on the ADR-0031 pattern |
| Standards 1 (second path authority) | **Fully adopt** | `_asset_res_path` reuses `path_outside_project` + `project_anchored` (the ADR-0006 double reading); a legitimately symlinked-in file is ACCEPTED again, as `script run` accepts it; `res://` input is the engine's lexical namespace and refuses `..` outright |
| Standards 3 / Spec 3 (docs + renderer behind the contract) | **Fully adopt** | The catalog, schema descriptions, and the human renderer all carry the evidence-state contract; the dry-run render prints `pass_will_also_import` and an invalid-retry hint |
| Artifact 1–3 + minor | **Fully adopt** | PR body rewritten to the shipped contract with fresh counts; `CONTEXT.md`'s claim is conditional up front (no claim+exception pair); the es README's trust section (lost to an earlier failed edit — my miss) and all four summary sentences are synchronized; SKILL no longer says scripts are ensured into the cache |

### The faithful adapter, in the engine's order

`_asset_state` now reads exactly what `_test_for_reimport` reads, and answers as it answers: an unparseable or `valid=false` sidecar → `invalid` (the engine SKIPS these — never auto-retried; the contract says "delete the sidecar to retry"); `importer="keep"/"skip"` → `cached`; the pre-UID format, any missing remap/destination reference (`path*`, `files`, `dest_files`), a `source_file` naming a different source (your alias), a missing `.md5` receipt, or a `source_md5` / `dest_md5` disagreeing with the actual bytes (`dest_md5` via the engine's one-context multi-file digest, replicated from `FileAccess::get_multiple_md5`) → `stale`. The two checks that are NOT readable from the artifacts — the current importer's format version and `are_import_settings_valid` — are the adaptation's **declared remainder**: an engine or import-settings upgrade can look `cached` until any pass runs, and the contract names that instead of pretending completeness. The alias is proven both ways: the unit fixture and a real-engine e2e in which the pass rewrites the copied sidecar to `source_file="res://alias2.png"`.

### Evidence vs policy

The states are now facts, the decisions derived: the pass runs for `missing`/`stale` requests; an `invalid` request settles `failed` WITHOUT spending a pass (the engine would skip it anyway — proven in e2e by the invalid neighbor's byte-identical sidecar after a full pass); `pass_will_also_import` (renamed from `pass_also_missing` to say what it now means) lists other `stale` assets only. The result model validates the mode split: evidence states on a dry run, settlements on a real run, and the summary's seven per-status counters against the list.

### Validation at `3327280a`

- Fast suite **1808 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: resource-import **5 passed** — the DF-010 heal, dry-run zero-write, engine-decided `not_importable`, real invalid (`failed`, stable), real stale (re-import), and the combined alias-rewrite + invalid-skip proof.
- Red-proof on `a8cf673b`: 7 unit + 1 e2e regressions fail there (`cached`≠`stale` for the alias and the tampered dest, a pass spent on invalid, the prediction unrendered, the renamed field absent, the symlink-in refused, the res:// escape message from the removed gate).
- `gda schema` vs the dev tip: still exactly one entry added (`resource import`).
- GitHub checks on `3327280a`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Exact-head re-review — changes still required

Re-reviewed exact head `3327280abbc5ba910dc7312bf44cf56d7c5ebb2f` against issue #668 and its authoritative 2026-08-24 contract revision. The specific alias/source-file false hit, invalid-neighbor overprediction, human-renderer omission, execution-kind ADR gap, Project-code execution wording, and Skill wording from the previous review are closed. The following same-class residue remains.

### Markdown / public contract

1. **The PR body still contradicts itself and the shipped schema.** Its contract summary says requested `missing` / `invalid` / `stale` assets trigger the pass, while the mechanism section says `invalid` does not. It also uses both `pass_will_also_import` and `pass_also_missing` for the same public field.
2. **The “exactly/faithful” freshness claim is still stronger than the implementation.** The PR body and `docs/command-catalog.md` say the adapter reads the engine's reimport verdict exactly, but the real-engine counterexample below disproves that claim.

### Domain-modular architecture

1. **[P1] `ARC-738-FRESHNESS` remains open: the Python adapter still produces a real false cache hit.**

   At `src/gda/commands/resource.py:768-834`, a parseable sidecar containing only `uid="uid://test"` falls through to `cached`; `needs_pass` at `:929-933` therefore becomes false. On Godot 4.6.3, gda reports `status:"cached", engine_pass:false`, while a subsequent native `godot --headless --import` rewrites the sidecar and restores its importer/cache data. The adapter omits the engine's expected sidecar-MD5 check and importer-existence check; those are additional gaps beyond the two declared remainders (importer format version and import-settings validity). The same false verdict also removes affected neighbors from the project-wide prediction.

   The smallest safe repair is to treat an unproven hit conservatively as `stale`, cover the readable engine invariants (including expected sidecar MD5 and importer existence), and add real-engine altered/partial-sidecar regressions.

2. **[P1/P2] `ARC-738-PATH` is only partially closed.**

   - `resource.py:725-735`: with a relative `--project game` and a legitimate `game/link.png` symlink into the project namespace, the lexical fallback compares relative `game/link.png` with an absolute project path and raises an uncaught `ValueError` traceback.
   - `resource.py:709-735`: `user://` and `uid://` bypass the command's documented “res:// or filesystem” boundary and are rewritten as literal `res://user:/…` / `res://uid:/…` paths when such files exist.

   Normalize the project/candidate into one coordinate system through the shared path authority, and reject engine-virtual schemes other than `res://` before filesystem conversion.

Stable claim status: `ARC-738-KIND-AUTHORITY`, the invalid-neighbor policy split, the launch seam, and the Project-code execution documentation are closed. `ARC-738-FRESHNESS` remains open; `ARC-738-PATH` reopens on the adjacent boundaries above.

### Standards

1. **[P1]** The relative-project symlink case escapes the structured-error contract as a bare traceback (`resource.py:725-735`).
2. **[P1]** The public model-derived schema/help still carries the old two-state contract: `ResourceImportParams` (`:453-461`), `ResourceImportResult` (`:580-593`), and the CLI docstring (`:1108-1115`) say only `cached`/`missing` and pass-on-missing, while the implementation publishes `stale`/`invalid` too.
3. **[P2]** Unsupported `user://` / `uid://` schemes are accepted and misaddressed instead of refused (`:709-735`).
4. **[P2]** `state()` adds the only new unreasoned source suppression, `# type: ignore[arg-type]` at `:763`, contrary to ADR-0030's reasoned-suppression policy. Type the evidence-state boundary explicitly instead.

No other actionable Fowler-baseline smell remained.

### Spec

1. **[P1]** Freshness can still return a false hit and suppress the required pass (partial-sidecar reproduction above).
2. **[P1]** Requested `invalid` assets do not trigger a pass (`resource.py:929-933`, pinned by `tests/test_resource_import_commands.py:280-297`), while the authoritative issue revision explicitly says the pass runs when a requested cache is `missing, invalid, or stale`. Even if Godot currently skips a `valid=false` sidecar, the implementation cannot silently rewrite the approved contract; restore the trigger or revise the issue authority first.
3. **[P2]** The authoritative revision names the machine field `pass_also_missing`, but the schema now publishes only `pass_will_also_import` (`resource.py:621-631`). Preserve the authoritative field, or formally revise the contract and record the compatibility decision.

The old alias/source-file reproduction, invalid-neighbor prediction, and human-renderer omission are closed.

### Verification

- Exact-head targeted suite: **100 passed**, including 5 real-Godot resource-import E2E tests.
- Fast suite: **1807 passed, 2 skipped, 515 deselected** on this macOS environment.
- Additional Godot 4.6.3 probes reproduced the partial-sidecar false hit, relative-project symlink traceback, and non-`res://` virtual-scheme misaddressing.
- Ruff, frozen pyright, and `git diff --check`: passed.
- Current Python, ruff, and pyright GitHub checks are green; the GitHub Godot E2E job remains skipped.
- Worktree stayed clean; no repository files were changed by this review.

qinhaihong-red:
## Second re-review response — all findings closed at head `634b4eb4`

Baseline: base tip / merge base `c7f6e4c0`, re-reviewed head `3327280a`, new head `634b4eb4`. Your three probes were reproduced before the fix (the uid-only sidecar answering `cached, engine_pass:false`; the relative-project symlink traceback; `user://` misaddressing), and all four new regressions are red on `3327280a`.

| Claim | Decision | Change |
| --- | --- | --- |
| [P1] `ARC-738-FRESHNESS`: a parseable-but-empty sidecar is a false hit | **Fully adopt** | `cached` now requires POSITIVE evidence — keep/skip, or declared destinations proven by the `.md5` receipts; a sidecar with no importer line or no destinations is conservatively `stale`. The declared remainder grows honest: importer format version, import-settings validity, AND the editor cache's expected sidecar MD5 (an engine-private cache format, not a project artifact) |
| [P1] `ARC-738-PATH`: relative-project symlink traceback; `user://`/`uid://` misaddressed | **Fully adopt** | Both sides of the containment comparison share one coordinate system (the project is absolutized before `path_outside_project` / `project_anchored`), with a structured refusal as the total fallback; non-`res://` engine-virtual schemes are refused by name via the shared `is_engine_virtual_path` |
| [P1 Spec] The implementation silently rewrote the approved trigger set | **Fully adopt — by correcting the authority, not the code** | A dated correction on #668 (2026-08-25) records the `missing`\|`stale` trigger and the reason: the engine skips `valid=false` sidecars byte-for-byte (verified), so an `invalid`-triggered pass spends a run that cannot change the outcome |
| [P2 Spec] The authoritative field name is `pass_also_missing` | **Fully adopt — same correction** | The correction records `pass_will_also_import` as the field (the semantics are stale-only; the old name predated them; never published in a release, so no compatibility decision is owed) |
| [P1 Standards] Models/help still said two states | **Fully adopt** | `ResourceImportParams`, `ResourceImportResult`, and the CLI docstring state the four evidence states and the trigger set |
| [P2 Standards] The unreasoned `# type: ignore` | **Fully adopt** | Gone: the evidence vocabulary is a shared `AssetStatus` Literal, typed end to end |
| Markdown 1–2 (PR body contradictions; "exactly/faithful" overclaim) | **Fully adopt** | The PR body cites the correction, uses one field name, and — like the catalog and docstrings — describes the adapter as reading the same artifacts with `cached` requiring positive evidence, plus the named three-part remainder |

### On the freshness boundary, stated as the contract now states it

The conservative rule replaces case-chasing: nothing is `cached` without proof. That closes your uid-only and importer-less counterexamples (both pinned as regressions, red on the previous head) and any future "parseable but hollow" variant, because the fall-through direction is now `stale` → the pass runs → the engine decides. The remainder that stays is exactly the engine-state checks: importer format version, `are_import_settings_valid`, and the editor cache's expected sidecar MD5 — the last was implicit before and is now named, since `filesystem_cache10` is an engine-private, versioned cache file rather than a project artifact; adapting it would rebuild the second-knowledge-owner problem this thread already removed once. The consequence direction is declared too: those three can only make gda look `cached` until any pass runs, never spend a pass the engine would not.

### Validation at `634b4eb4`

- Fast suite **1813 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: resource-import **5 passed**.
- Red-proof on `3327280a`: the four new regressions fail there (`cached`≠`stale` for the uid-only and importer-less sidecars, the bare-traceback symlink case, `user://` accepted).
- Issue authority: [the 2026-08-25 correction](https://github.com/aigengame/godot-agent/issues/668#issuecomment-5404590394) is recorded on #668 under the maintainer's standing authorization for this thread's contract records; the PR body cites it.
- `gda schema` vs the dev tip: still exactly one entry added.
- GitHub checks on `634b4eb4`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Third exact-head re-review — changes still required

Re-reviewed exact head `634b4eb400b097a549d61d230a6b222bcda2750e`. I reran the Markdown artifact, domain-modular architecture, Standards, and Spec axes independently, then replayed the remaining boundary against Godot 4.6.3.

### Blocking finding

1. **[P1] `ARC-738-FRESHNESS` remains open: a declared importer is not positive evidence that the importer exists.**

   `_asset_state()` checks only whether an `importer=` line exists (`src/gda/commands/resource.py:817-824`). With intact destinations and receipts, any non-`keep`/`skip` importer name falls through to `cached` at `:887`, which suppresses the required pass at `:986` and also omits the asset from `pass_will_also_import`.

   Real-engine reproduction:

   1. Import a PNG normally with Godot 4.6.3.
   2. Change only `importer="texture"` in its sidecar to `importer="definitely_missing_importer"`.
   3. Update the expected sidecar MD5 in `.godot/editor/filesystem_cache10`, excluding that already-declared remainder from the experiment.
   4. `gda resource import res://icon.png --dry-run` reports `status:"cached"` and `engine_pass:false`.
   5. Native `godot --headless --import` immediately reimports the asset and restores `importer="texture"`.

   This is outside the declared importer-version, settings-validity, and expected-sidecar-MD5 remainder. It violates the revised per-asset hit/miss contract because a requested engine-stale asset is reported cached and prevented from triggering the pass. The implementation needs an engine-owned authority for importer availability; otherwise the authority and public contract must explicitly accept and describe this additional uncertainty instead of calling it a proven cache hit.

### Public contract and Markdown residue

1. **[P2] The generated schema still overstates cache certainty.** `ResourceImportAsset` at `resource.py:505-518` says evidence is read "exactly" like the engine and `cached` means "every engine check passes". This is emitted as public `--schema` text and contradicts both the declared remainder and the reproduction above, contrary to ADR-0004 model-driven self-description.
2. **[P2] The Markdown contract repeats the same unsupported proof claim.** `docs/command-catalog.md:703-714` presents the listed artifacts as positive cached evidence and the three-part remainder as exhaustive. The PR body mechanism bullet also lists three unreadable checks, then calls them "the two checks"; it says only previously-missing assets are re-verdict-ed, although requested stale assets follow the same settlement path.
3. **[P3] Adjacent internal prose remains stale.** `resource.py:448` says the pass runs only when a requested cache is `missing`, omitting `stale`.

### Closed since the previous re-review

- The UID-only and importer-less sidecar probes now return `stale`.
- Relative `--project` plus a symlinked-in asset is structured correctly.
- `user://` and `uid://` inputs are refused.
- The `invalid` trigger and `pass_will_also_import` naming findings are closed by the 2026-08-25 issue correction.
- The old two-state help/schema wording and the unreasoned `type: ignore` are closed.

### Verification

- Fast suite: **1811 passed, 2 skipped, 515 deselected** on this macOS environment.
- Real-Godot resource-import E2E: **5 passed**.
- Ruff, pyright, and `git diff --check`: passed.
- Current GitHub Python, ruff, and pyright checks are green; the GitHub Godot E2E job remains skipped and does not cover this counterexample.
- Worktree remained clean; no repository files were changed by this review.

Terminology and prose were checked against issue #668 and its corrections, `CONTEXT.md`, ADR-0004/0012, and the command catalog, using ASD-STE100 for the changed English technical prose. No additional substantive terminology or prose finding remained; Spanish and Japanese native-language polish was not assessed.

qinhaihong-red:
## Third re-review response — the open [P1] closed per your second option, at head `83229170`

Baseline: base tip / merge base `c7f6e4c0`, re-reviewed head `634b4eb4`, new head `83229170`. **No behavior change in this round** — the finding is closed the way your finding text offered second: the authority and the public contract now explicitly accept and describe the uncertainty, instead of calling it a proven hit.

| Claim | Decision | Change |
| --- | --- | --- |
| [P1] A declared importer is not positive evidence that it exists | **Fully adopt, via your second option** | Importer availability joins the declared engine-state remainder — recorded on the ISSUE authority (the [2026-08-25 addendum](https://github.com/aigengame/godot-agent/issues/668#issuecomment-5404751739)) and in every public description, with the remainder's one-sided direction stated |
| [P2] The schema text overstates certainty ("exactly", "every engine check passes") | **Fully adopt** | The `ResourceImportAsset` self-description states the ARTIFACT-level boundary and names the four-part remainder |
| [P2] The catalog presents the evidence as proof and the remainder as exhaustive; PR-body count/settlement errors | **Fully adopt** | "Artifact-level is the boundary, not a proof of the engine's whole verdict"; the "two checks" count and the previously-missing-only settlement sentence are corrected |
| [P3] `resource.py` module comment omits `stale` | **Fully adopt** | "missing or stale" |

### Why the second option, and not an engine-owned importer authority

The importer registry is OPEN: import plugins register names at editor startup, so the authoritative list exists only inside a running editor process — the same place the per-file reimport API lives, which this thread already established has no headless CLI entry. An offline list held by gda (hardcoded or scraped) would misjudge every legitimate plugin importer and rebuild the curated-table / second-knowledge-owner defect this review had me remove twice (the destination-exists heuristic in round 2, the temporary-subproject idea in round 1). Your own reproduction also shows the class this belongs to: step 3 had to update the editor cache's expected sidecar MD5 first — i.e., in a real editor-cached project, the engine catches an altered importer name through exactly the expected-MD5 check that is already a declared remainder; in a clean worktree the engine's `p_expected_import_md5.is_empty() → reimport` path re-imports everything regardless. Sidecar-content drift is engine-state territory end to end, so the honest contract names it rather than simulating it.

What changed to make that acceptance real rather than implicit:

- **The authority**: the issue addendum records the four-part remainder (importer availability, importer format version, import-settings validity, the editor cache's expected sidecar MD5) and its direction — it can delay a re-import until the next pass, never spend a pass the engine would not.
- **The public contract**: the schema-emitted docstring, the catalog, and the PR body all describe `cached` as positive ARTIFACT-level evidence with that named remainder; "exactly" and "every engine check passes" are gone.

### Validation at `83229170`

- Fast suite **1813 passed**; ruff check + format clean; `git diff --check` clean; the diff is docstrings/comments/Markdown only (no `.py` logic), so the engine-path evidence from `634b4eb4` stands: resource-import e2e **5 passed**, red-proofs as previously reported.
- GitHub checks on `83229170`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Fourth exact-head re-review — changes still required

Re-reviewed exact head `83229170b97d597761c129613b47841d1cf95528`. I reran the Markdown artifact, domain-modular architecture, Standards, and Spec axes independently, then replayed the previous boundary and adjacent evidence states against Godot 4.6.3.

The former unknown-importer finding is closed by the new issue addendum: importer availability is now explicitly accepted as an engine-state remainder. A distinct false-`stale` path remains.

### Blocking finding

1. **[P1] `ARC-738-FRESHNESS` remains open: a no-destination sidecar triggers a project-wide pass that Godot would not run, then settles to `failed`.**

   At `src/gda/commands/resource.py:893-898`, every non-`keep`/`skip` sidecar without `dest_files` is unconditionally classified `stale`. That drives `needs_pass` at `:992-998`; the same classification also feeds `_project_import_gaps()` at `:914-940`.

   Real-engine reproduction from a normally imported PNG:

   1. Keep the sidecar importer, UID, `source_file`, import parameters, and matching import-base `.md5` receipt.
   2. Remove only its `path=` and `dest_files=` lines.
   3. Synchronize the editor-cache expected sidecar MD5, excluding that accepted engine-state remainder.
   4. Exact-head gda dry-run reports `status:"stale"` and `engine_pass:true`.
   5. Native Godot 4.6.3 `--editor --import --quit` exits 0 and leaves the sidecar bytes and mtime unchanged: `_test_for_reimport` considers it current because the importer/UID/source receipt checks pass and there are no destinations to check.
   6. A real gda run still launches the project-wide pass and then reports `status:"failed"`, `created:[]`; the sidecar remains unchanged.

   Requesting another missing asset also falsely lists this asset in `pass_will_also_import`, even though the native pass does not re-import it.

   This state is outside the accepted false-`cached` engine-state remainder. The current vocabulary conflates two questions: whether artifacts positively prove a usable cache, and whether the engine will act during a pass. Here neither `cached` nor the current `stale`/`invalid` definitions can express the observed state. The smallest repair is to represent it as an explicit no-pass/retry state — for example, broaden the existing `invalid`/delete-sidecar contract — or otherwise separate artifact evidence from `engine_will_reimport`. Add requested, neighbor-prediction, and real-engine regressions.

### Public contract residue

1. **[P2]** `ResourceImportSummary.cached` at `resource.py:569` still means “Assets whose cache was already intact,” contradicting the newly accepted artifact-level false-`cached` remainder. This text is emitted into public `--schema` and gda-mcp through ADR-0004/0012.
2. **[P3]** The public timeout schema at `resource.py:491-492` says the project-wide pass imports every `missing` asset, omitting `stale` assets.
3. **Markdown:** `docs/command-catalog.md:703-710` calls a sidecar with no destinations `stale` and states that stale assets are what the engine would import; the reproduction disproves that claim. The live PR body still says only previously-missing assets are re-verdict-ed, although stale requests follow the same settlement path, and its fast-suite count remains `1808`.

### Closed since the previous re-review

- Importer availability is now explicitly owned by the issue addendum as one of four engine-state remainders.
- The `ResourceImportAsset` schema no longer says “exactly” or “every engine check passes.”
- The catalog now names the artifact-level boundary and four-part remainder.
- The internal pass comment now says `missing or stale`.

### Verification

- Fast suite: **1811 passed, 2 skipped, 515 deselected** on this macOS environment.
- Real-Godot resource-import E2E: **5 passed**.
- Ruff, format, pyright, and `git diff --check`: passed.
- Current GitHub Python, ruff, and pyright checks are green; the GitHub Godot E2E job remains skipped and does not cover this counterexample.
- Exact local and remote head remained `83229170`; the review worktree stayed clean.

Terminology and prose were checked against issue #668 and its addenda, `CONTEXT.md`, ADR-0004/0012, the command catalog, and Godot `_test_for_reimport`, using ASD-STE100 for the changed English technical prose. No additional substantive terminology or prose finding remained.

qinhaihong-red:
## Re-review response — all findings adopted at head `300c5755`

Baseline: base tip / merge base `c7f6e4c0`, re-reviewed head `83229170`, new head `300c5755`. Your live probe was confirmed against the engine source before the fix: `ResourceFormatImporter::get_import_base_path` derives the import base from the ASSET path — `.godot/imported/<filename>-<md5 of the res:// path>` — and `md5("res://icon_a.png")` reproduces exactly the hash your probe observed. My adapter located the receipt from the sidecar's destination filenames instead, which is why a no-destination sidecar's intact receipt was invisible to it.

| Claim | Decision | Change |
| --- | --- | --- |
| [P1] no-destination sidecar with intact receipt judged `stale` → a pass Godot would not run → false `failed` | **Fully adopt** | The receipt is now looked up at the PATH-derived base, checked regardless of declared destinations; a no-destination sidecar with a matching receipt is `cached` (the unproven→stale else-branch is removed) |
| [P2] `summary.cached` description promised more than the contract | **Fully adopt** | Now reads "Assets the artifacts show the engine's pass would leave as is (subject to the declared engine-state remainder)" |
| [P3] `timeout` description said the pass re-imports "every missing asset" | **Fully adopt** | Now "re-imports every missing or stale asset" |

### The mechanism, and where it now lives

`_asset_state` computes the receipt path exactly as the engine does — `Path(rel).name + "-" + md5(res_path) + ".md5"` under `.godot/imported/` — so the receipt check no longer depends on the sidecar declaring destinations. The verdict chain is unchanged otherwise: a missing receipt or a `source_md5`/`dest_md5` mismatch is still `stale`; declared destinations are still existence- and digest-checked; and the no-destination case with a matching receipt now lands in `cached`, which is the engine's own verdict (nothing to check, receipts pass — your probe showed the pass leaves such a sidecar untouched). The prose that contradicted this — the catalog's "no importer or destinations proves nothing" sentence and the adapter docstring — now states the path-derived lookup and the no-destination-cached case explicitly, and the PR body's settlement sentence covers all non-cached requests, not only previously-missing ones.

### Validation at `300c5755`

- New regressions, red-proofed on `83229170` against its `resource.py`: unit — no-dest sidecar + matching receipt is `cached` with `engine_pass:false` on both dry and real runs (fails there: `'stale' == 'cached'`); unit — the same state excluded from `pass_will_also_import` (fails there: `['res://icon.png'] == []`); e2e — a real import whose sidecar has its `path=`/`dest_files=` lines stripped stays `cached`, spends no pass, creates nothing (fails there on the old verdict).
- Test fixtures now write receipts at the path-derived base via a shared `_receipt_path` helper — the fixture layout matches the engine's, so the suite can no longer pass on the wrong lookup.
- Fast suite **1815 passed**; resource-import e2e, serial: **6 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- GitHub checks on `300c5755`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Fifth exact-head re-review — changes still required

Re-reviewed exact head `300c575544d811cded371a66a76f134746451efc` against base / merge-base `c7f6e4c01e25e53b05953c80b72966adc3e504f8`. I kept the Markdown artifact, domain-modular architecture, Standards, and Spec axes independent, then replayed the previous finding and adjacent receipt states against Godot 4.6.3.

The previous no-destination + matching path-derived receipt finding is closed. One adjacent receipt state still violates the accepted one-way pass guarantee.

### Blocking finding

1. **[P1] A malformed path-derived receipt is classified `stale`, so gda runs a project-wide pass that Godot deliberately skips and then reports `failed`.**

   `src/gda/commands/resource.py:893-900` reads the receipt with regexes. A malformed value such as `source_md5=[` has no `_SOURCE_MD5_LINE` match, so `_asset_state()` returns `stale`. That drives `needs_pass` at `:1011`; the same verdict feeds `_project_import_gaps()` at `:951`.

   Real-engine reproduction from a normally imported PNG:

   1. Replace its path-derived `.md5` receipt with malformed `source_md5=[`, retaining the other imported artifacts.
   2. Keep the sidecar contents and editor-cache expected MD5 aligned; change only the sidecar mtime to force `EditorFileSystem::_test_for_reimport`.
   3. Exact-head gda dry-run reports `status:"stale"` and `engine_pass:true`.
   4. Native Godot 4.6.3 logs the receipt parse error and leaves the sidecar, receipt, and destination unchanged. Its receipt parse-error branch returns `false` rather than entering a reimport loop.
   5. A real gda run still spends the pass and returns `status:"failed"`, `created:[]`.

   Requesting another missing asset also falsely lists this malformed-receipt asset in `pass_will_also_import`, although the native pass skips it.

   This contradicts issue #668's accepted guarantee that the artifact adapter may delay a reimport but must “never spend a pass the engine would not.” It also makes the public `stale` / `pass_will_also_import` descriptions false for this state. The smallest repair is to distinguish receipt parse failure from a valid missing or mismatching digest and map it to an explicit no-pass invalid/retry outcome (with an accurate receipt-repair hint), or add another explicit no-pass state. Please add requested, neighbor-inventory, and real-engine regressions.

### Standards / test evidence

1. **[P2] The new no-destination E2E does not verify the native verdict named by the test and its comments.**

   `tests/test_e2e_resource_import.py:265-299` strips `path=` / `dest_files=` after the initial import, but it neither synchronizes `.godot/editor/filesystem_cache10`'s expected sidecar MD5 nor invokes native Godot after the mutation. Both later calls are gda calls; because gda returns `cached`, neither starts Godot. On the identical unsynchronized state, a native import restores both lines because the expected sidecar MD5 changed.

   To preserve the intended regression, synchronize the expected sidecar MD5, explicitly run native import, and assert the sidecar remains byte-identical. Otherwise narrow the test name/comments and add a separate native-parity test.

No other actionable Fowler-baseline smell remained.

### Markdown artifact residue

- `docs/command-catalog.md:707-708` and the PR body call a no-destination sidecar with a matching receipt “current to the engine.” That must be bounded to gda's artifact-level checks and the declared engine-state remainder; the current E2E does not prove the absolute native claim.
- The PR body still reports resource-import E2E as **5 passed**, while the current file collects and passes **6** tests. The latest reply has the correct count, but the body is stale.

Terminology and prose were checked against issue #668 and its addenda, `CONTEXT.md`, ADR-0004/0012/0023, the command catalog, and Godot's `_test_for_reimport`, using ASD-STE100 for the changed English technical prose. No other substantive terminology or prose issue remained.

### Architecture and Spec status

- `ARC-738-FRESHNESS`: **open** — the evidence adapter does not preserve Godot's receipt parse-error control flow before the application decides to launch the external pass.
- `ARC-738-PASS-INVENTORY`: **open** — the same centralized verdict falsely propagates into the neighbor inventory.
- Module ownership, dependency direction, orthogonality, DRY, and existing extension seams remain sound; this needs a local semantic repair, not a new abstraction.
- Spec scope creep: none. The malformed-receipt case is the remaining Spec P1.

### Closed since the previous re-review

- The no-destination + matching path-derived receipt state now reports `cached`, spends no pass, and is excluded from the neighbor inventory when the accepted engine-state remainder is controlled.
- `summary.cached`, timeout, and non-cached settlement descriptions are corrected.

### Verification

- Fast suite: **1813 passed, 2 skipped, 516 deselected** on macOS.
- Real-Godot resource-import E2E: **6 passed**.
- Pyright: **0 errors**; ruff check and format: clean; `git diff --check`: clean.
- GitHub Python, ruff, pyright, package, and co-install checks are green. The GitHub Godot E2E job remains skipped and does not cover either counterexample above.
- Exact local and remote head remained `300c5755`; the review worktree stayed clean.

qinhaihong-red:
## Re-review response — all findings adopted at head `a3a681a2`

Baseline: base tip / merge base `c7f6e4c0`, re-reviewed head `300c5755`, new head `a3a681a2`. Your parse-error claim was verified against the engine source before the fix: `EditorFileSystem::_test_for_reimport` parses the receipt with `VariantParser`, and its error branch is `return false` with the comment "skip and let user attempt manual reimport to avoid reimport loop" — the same deliberate no-pass skip as a `valid=false` sidecar, and a different branch from a MISSING receipt (`return true`, re-import) and from a receipt that parses but lacks `source_md5` ("Lacks md5, so just reimport", `return true`).

| Claim | Decision | Change |
| --- | --- | --- |
| [P1] malformed receipt classified `stale` → a pass Godot deliberately skips → false `failed`, and a false `pass_will_also_import` entry | **Fully adopt** | An unparseable receipt is now `invalid`: no pass, settles `failed` without launching, excluded from the gap scan (the existing invalid paths — the gap scan lists `stale` only, `needs_pass` keys on missing/stale — carry it with no new mechanism) |
| [P2] the no-destination e2e did not verify the native verdict its name claims | **Fully adopt** | The test now controls the declared remainder — it synchronizes the editor cache's expected sidecar MD5 to the mutated bytes (the one recorded digest in `filesystem_cache10`) — then runs NATIVE `--import` and asserts the mutated sidecar stays byte-identical, before the gda assertions |
| Markdown: "current to the engine" overclaims; e2e count stale | **Fully adopt** | The catalog, adapter docstring, and PR body now bound the claim to gda's artifact-level checks, with the engine's untouched-sidecar behavior stated as remainder-controlled and now natively proven; the body's e2e count is current (7) |

### The receipt grammar, and its declared direction

The adapter no longer keys the verdict on whether a regex matched: the receipt has an explicit grammar — the one line shape the engine ever writes into it (`key="value"`) — checked before any digest comparison. A receipt outside that grammar is `invalid` (the engine's parse-error skip); a receipt inside it that lacks `source_md5` stays `stale` (the engine's "Lacks md5, so just reimport"); a missing receipt stays `stale`. Where gda's grammar is narrower than `VariantParser` (a hand-written receipt like `source_md5=123`, which the engine would parse and then re-import on mismatch), the divergence errs toward `invalid` — the no-pass direction the accepted one-way guarantee sanctions (delay is allowed; a pass the engine would not run is not) — and the docstring declares that boundary. The `invalid` prose now names both parse-error sources everywhere it appears (result model, `summary.invalid`, the human renderer's hint, the recipe, the catalog), with the repair stated accurately: delete the sidecar to retry — the one repair that is robust for both kinds, since the pass rewrites the sidecar and its receipt together.

### Validation at `a3a681a2`

- Red-proof on `300c5755` against its `resource.py`: the requested-asset regression fails there (`'stale' == 'invalid'`), the neighbor-inventory regression fails there (`['res://icon.png'] == []`), and the real-engine e2e fails there on the same mechanism after its native section passes (the engine skipped either way — the old head's gda just disagreed with it). The third new unit test (receipt lacking `source_md5` is `stale`, not `invalid`) is a boundary pin and is green on both heads by design.
- The new e2e replays your reproduction: replace the receipt with `source_md5=[`, bump the sidecar mtime so the cache's expected-MD5 check passes and `_test_for_reimport` reaches the receipt parse; native `--import` leaves the sidecar, the malformed receipt, and the destination byte-identical; gda then reports `invalid` / `engine_pass:false` on the dry run and settles `failed` with `created:[]` and no launch on the real run.
- Fast suite **1818 passed**; resource-import e2e, serial: **7 passed** (including both native-parity proofs); pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- GitHub checks on `a3a681a2`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
## Follow-up at head `13009588` — one residue found and closed by an internal review of `812dbfb0`

`812dbfb0` (the maintainer's assignment-order fix) was re-reviewed internally against the engine source and the issue #668 one-way guarantee. Its own claim holds — `_test_for_reimport`'s receipt loop overwrites on every assignment, so the last value wins, exactly as the new parser reads it — and every prior round's finding stays closed. One residue came out of the widened escape tolerance:

**Lone UTF-16 surrogate escapes.** `json.loads` accepts `"\ud800"`, but `VariantParser` rejects it (`unpaired lead/trail surrogate`, `TK_ERROR`) — the engine's deliberate parse-error skip. The subset parser therefore decoded such a receipt, mismatched the digest, classified the asset `stale`, and spent a project-wide pass the engine would not run — the same contract clause as the previous round's [P1], on a still narrower hand-crafted input.

**The fix**: a decoded value carrying any surrogate code point is outside the documented subset (`None` → `invalid`, the sanctioned no-pass direction). Paired surrogate escapes decode to one real code point on both sides and stay inside — pinned by a boundary test. The subset sentence now says "JSON-style escapes (lone UTF-16 surrogates excluded)" in the model docstring, the adapter docstring, the parser docstring, and the catalog.

**Validation at `13009588`**:
- Red-proof on `812dbfb0` against its `resource.py`: the lone-surrogate regression fails there (`'stale' == 'invalid'`); the paired-surrogate boundary pin is green on both heads by design.
- New real-engine e2e replays the divergence natively: the engine's own `--import` leaves the lone-surrogate receipt, sidecar, and destination byte-identical (the skip), and gda reports `invalid` / no pass / `failed` with `created:[]`.
- Fast suite **1824 passed**; resource-import e2e, serial: **9 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- GitHub checks on `13009588`: green (the `Godot e2e tests` job remains dispatch-only by design).

Issue #660, opened 2026-08-15
screen capture: evidence-grade receipt binding session, scene, state, and output hash


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-026** (2026-08-10, reconfirmed on four later packages) — `screen capture` returns only path, dimensions, byte count, and format, plus the opt-in inline PNG. Every package had to read the gallery state separately (`game get`), hash scene/script sources and each PNG locally, and join the facts into an explicitly unsigned `agent_observed_unsigned_cli_result`.
- **GDA-DF-031 "Runtime capture receipts do not bind scene, state, and output hash"** (2026-08-12, marked high in the notes, further reconfirmed on roughly seven packages; the document's GDA-DF numbering collides across sessions — identified by title and date) — same request from an independent session: an image can be a valid PNG while belonging to the wrong gallery state or a stale session, and downstream tests can verify the joined files but not the original capture event.

## What to build

Extend the capture result with a **receipt** — identity and hash fields that bind session, scene, frame, and output hash in one result:

- session identity, launched scene path (UID only when the project provides one — gda-authored scenes carry none, ADR-0036), engine frame or timestamp;
- SHA-256 of the output file.

The receipt fields are unconditional additions to the capture result; existing fields are unchanged.

**Dependency to resolve in this issue:** gda has no session identifier today — neither the daemon session nor `daemon status` exposes one. This issue defines it: an identifier bound to the `Engine session` lifetime, readable outside the capture result (e.g. on `daemon status`) so a receipt can be correlated.

**Predicate boundary:** the capture predicate (capture-when-property-holds) is owned by **#661**. When a predicate is supplied through #661's mechanism, the receipt echoes the observed value and the frame it was evaluated at.

## Acceptance criteria

- [ ] A capture result carries the session identifier, scene path, engine frame or timestamp, and output SHA-256.
- [ ] The same session identifier is readable on `daemon status` and stays stable for the lifetime of one `Engine session`.
- [ ] When a #661 predicate was supplied, the receipt echoes its observed value and evaluation frame.
- [ ] Existing capture consumers see only additive fields.

## Blocked by

None for the receipt fields. The predicate mechanism is #661's scope; the echo criterion lands with or after #661.

qinhaihong-red:
> *This was generated by AI during triage.*

[...]

PR #743, opened 2026-08-25
feat(gda): predicate-gated screen capture with atomic input events (#661)

### LINKED ISSUE #661: live: atomic input-and-capture or frame-predicate capture for transient animation states
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-023** (2026-08-10, reconfirmed on two later packages). Input and capture are separate live operations; CLI/daemon transport timing (observed ~4 s round trip) means a 3–8-frame VFX peak is routinely missed — both commands report success while the PNG shows an already transparent VFX or the settled idle pose. Packages had to author game-side QA galleries that freeze representative frames to make captures deterministic.

Note: `screen frames` bursts consecutive frames, but it still starts after the same round trip and cannot say which frame holds the requested state, so it does not remove the game-side fixture.

## What to build

This issue owns the **synchronization mechanism**; #660 owns the capture receipt (which echoes what this mechanism observes).

- A capture predicate over a runtime property (e.g. asset key / current animation frame), evaluated at a frame boundary (ADR-0020's frame-coherence point; state explicitly if evaluation must happen after the frame is drawn rather than at a process-frame boundary): capture completes when the predicate holds, or fails with a typed error after a bounded number of frames.
- Optionally, one atomic input-and-capture operation that injects input and applies the same predicate-gated capture without a second round trip. (Naming note for triage: a three-token command name does not fit the current `<group> <command>` CLI structure — pick a two-token name.)
- The result reports the frame index observed at capture, and the predicate's observed value when one was supplied.

## Acceptance criteria

- [ ] An agent can capture a declared transient animation state deterministically without game-side freeze fixtures.
- [ ] A predicate that never holds fails with a typed error after the stated frame bound.
- [ ] The result reports the observed frame index (and the predicate evaluation when supplied); #660's receipt echoes them.
- [ ] A live regression captures a short transient state reliably across repeated runs.

## Blocked by

None - can start immediately. Related: #660 (receipt fields; echoes this mechanism's observations).

qinhaihong-red:
## W5 slice: predicate-gated screen capture with atomic input events (#661)

**The gap (GDA-DF-023):** input and capture are separate live operations, and the ~4 s CLI/daemon round trip lands after a 3–8-frame VFX peak — both commands report success while the PNG shows an already-settled pose. Packages had to author game-side QA galleries that freeze representative frames. `screen frames` does not remove the fixture: it starts after the same round trip and cannot say which frame holds the requested state.

### What shipped

**The synchronization mechanism** (this issue's ownership; #660's receipt echoes it):

- `gda screen capture --await-node <path> --await-property <name> --await-value <json-scalar> [--await-frames N]` holds the capture **game-side**: the harness evaluates `node.property == value` once per PROCESS frame and captures at the first holding frame's boundary. The result carries a `predicate` report — the `observed` value, the absolute `engine_frame`, and the window-relative `frames_waited` — the fields #660's receipt will echo. A gated success is **correlated CLI-side** before the file is written: the report must name this request's predicate, observe a value that satisfies it, and wait inside the declared window — anything else (including an unsolicited report on a plain capture) is `contract_violation`.
- A predicate that never holds within the bound (default 60 frames, ceiling the shared 600) is the new registered **`live_predicate_unmet`**, carrying the last observed value in its message.
- `--await-events <json>` is the **atomic input-and-capture form**: input-sequence events (the same discriminated union `input sequence` takes — single authority, imported not copied) are applied at process-clock `frame` offsets in the same serialized live operation, so a transient triggered by input cannot be missed by a second round trip. **Every declared event fires before the reply** — an early predicate match or an error path never leaves a scheduled release held (proven live for key, mouse-button, and action) — and **a declared event that fails makes the whole capture that typed failure**: the capture payload is discarded, no file is written, and later events still drain.
- The predicate ceiling and total drain ceiling are intentionally distinct: an event may sit at or beyond `await_frames` and still fire, but every event offset is schema-bounded to `0..599`, so the operation drains within the shared 600-frame live-window bound. Such a late event can no longer satisfy the predicate, whose scan ends at `await_frames`. Physics-clock offsets are refused both model-side (ADR-0015) and in the Draft 2020-12 schema.
- **The model and emitted schema accept the same event set**: every scalar field of the sequence union is strict; the modifier vocabulary is a structural enum; the total-window event-offset maximum is structural; and the former model-only offset-inside-predicate-window rule remains removed. A validator/Pydantic parity corpus pins both directions, integer-for-number boundaries, unknown modifiers, the total offset boundary, and the supported beyond-predicate-ceiling case.

- **No new live op and no three-token command**: the predicate rides the existing `screen-capture` op as params, so the op table, `gda-mcp` tool list, and the live contract guards are untouched.

**Mechanism reuse:** the predicate window is the #223 multi-frame base — `_advance_window` gains an early-COMPLETE arm symmetric to its existing early-error arm (`{"complete": payload}` beside `{"error": …}`); the events reuse `_apply_sequence_event` and the offset helpers verbatim; property existence uses the same storage-property + explicit-script-variable two-step `game get` resolves with. Predicate equality is a JSON-scalar contract (numbers numerically, strings against the String rendering, bools strictly) — documented, not a Variant matcher.

**The coherence contract, verified live on both trigger paths:** each window tick EVALUATES BEFORE it injects, so the observed property is always the previously COMPLETED frame's state — exactly the frame the captured texture presents. A `_process`-driven flip is observed with its own presentation (red-pixel e2e, three repeats); a state written by an injected event's synchronous `_input` callback is observed one boundary later, together with its presentation (the same-callback e2e decodes a green Marker pixel). Declared consequences: the predicate sees frame-boundary state only (an intra-frame value is never observable — the typed unmet error, never mismatched pixels); an event's effect is observable from the next boundary; a game-side visual-after-property lag trails by that game frame. Stated in the schema descriptions, help, catalog, SKILL, and the ADR-0020 amendment; ADR-0017 records early completion plus the event-failure reply rule, and ADR-0040 records the fourth one-way group edge (`screen` → `input`).

**Self-description:** `--await-events` is the surface's first nullable-compound single-token option — exactly the case #669's registration guard was built to force — so the argv-binding `json_value` derivation now sees a compound behind `anyOf`/`oneOf`, and the guard's premise comment is updated to "keep the two detectors agreeing".

**Harness identity gate:** `HARNESS_VERSION` 8 → 10 (9 declared multiple bodies during this PR and was retired). The editable in-test history has been removed because it could not enforce history. Instead, the required Python CI job checks the Git merge base against the PR/push head: if `gda_harness.gd` bytes changed, the head must contain a strictly greater numeric `HARNESS_VERSION`. Checkout uses full history, while the current version/hash unit pin still catches accidental snapshot edits and gives the replacement hash.

### Intended public-surface diffs (enumerated)

- `gda screen capture --help`: the five new `--await-*` options.
- `gda schema`: `screen capture` gains five new input properties (`await_frames` bounded 1..600; `await_events` embedding the input-sequence event `$defs`), the `predicate` result field backed by the new `CapturePredicateReport` model, and `json_value: true` on the `--await-events` argv binding. The shared constraint publication also reaches the input group: `input key` / `input tap` emit the modifier vocabulary as an `enum` (the tap result mirrors it), and the sequence-event `$defs` gain `maximum: 599` on `frame` / `physics_frame` with matching descriptions. The complete `gda schema` diff vs the dev tip touches exactly these four commands: `input key`, `input tap`, `input sequence`, `screen capture`.
- SKILL.md `screen` row and the command catalog's `screen` section describe the predicate and the atomic form; ADR-0002's table gains the `live_predicate_unmet` row.

| AC | Evidence |
| --- | --- |
| Deterministic capture of a declared transient without game-side fixtures | e2e: `phase == 5` holds for exactly ONE process frame per 8-frame cycle and the rect is RED on that frame alone; the decoded PNG shows the matched frame's pixels, **three times in one session** — and the input-triggered path decodes the same-callback Marker pixel green |
| Never-holding predicate → typed error after the stated bound | e2e: `phase == 99 --await-frames 20` → `live_predicate_unmet`, "did not hold within 20 frames", "last observed" in the message, no file written |
| Result reports frame index + predicate evaluation | unit + e2e: `predicate.{observed, engine_frame, frames_waited}` asserted; renderer prints the predicate line |
| Live regression capturing a short transient reliably across repeated runs | the 3-iteration e2e above, plus the input-triggered form: `--await-events` injects a key press and captures `flash == 1`, a one-frame state ~3 frames later |

### Validation

- Exact delivered head: `402ecb21520fcd6a2fc82cf61eb721a015287b1f`.
- Non-E2E suite: **1850 passed, 2 skipped, 527 deselected**.
- Real-engine windowed E2E: screen **13 passed**, input **16 passed** (**29 passed** total).
- Pyright: 0 errors; repository-wide Ruff check and format: clean (**468 files**); workflow YAML parse and `git diff --check`: clean.
- Harness guard adversarial proof: historical `2270e22f` → `75afef5c` is rejected because the harness body changed while version 9 was reused; dev merge base `85d8a285` → delivered head passes with the legitimate 8 → 10 transition.
- Schema/model probes now both reject `modifiers:["hyper"]` and an event at frame 600, while both accept an event beyond the predicate ceiling when it remains inside the total ceiling.
- Existing red proofs for the original feature and earlier review rounds remain documented in the commit history; no command or live op was added.

Closes #661

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,8 @@ jobs:
         uses: actions/checkout@v6
         with:
           lfs: true
+          # The harness-version guard compares the PR/push base and head.
+          fetch-depth: 0
 
       - name: Set up Python environment
         uses: ./.github/actions/setup-python-env
@@ -162,6 +164,13 @@ jobs:
           # gda-balancing has its own required matrix and stable aggregator.
           # Keep this required root context scoped to the root project.
           sync: root
+
+      - name: Guard harness version identity
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        run: python scripts/harness_version_guard.py "$BASE_SHA" "$HEAD_SHA"
 
       - name: Run unit tests
         run: uv run --frozen pytest -m "not e2e"
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -305,6 +305,7 @@ operation, and parse codes the CLI assigns).
 | `live_invalid_key` | `live` | `classifier` | `6` | A live input key event named a key the engine could not resolve to a keycode (Phase 2, #221). |
 | `live_unknown_action` | `live` | `classifier` | `6` | A live input action targeted an action the running game's InputMap does not declare (Phase 2, #221). |
 | `live_invalid_event_spec` | `live` | `classifier` | `6` | A live input sequence event has a type the harness does not recognize (Phase 2, #221). |
+| `live_predicate_unmet` | `live` | `classifier` | `6` | A live `screen capture` predicate (`--await-*`, #661) did not hold within its declared frame bound (Phase 2). |
 | `live_display_unavailable` | `live` | `classifier` | `6` | A live `screen` capture ran on a headless engine session (the dummy DisplayServer cannot read pixels); start the daemon windowed with `gda daemon start --windowed` (Phase 2, #222). |
 | `live_unsupported_platform` | `environment` | `classifier` | `127` | Live operations require a UNIX platform (macOS/Linux); they use Unix domain sockets, unavailable here. Phase-1 headless is unaffected (Phase 2, ADR-0021). |
 | `live_windowed_unavailable` | `environment` | `classifier` | `127` | A windowed Engine session (`gda daemon start --windowed`) was requested but the host has no usable DisplayServer (no on-console GUI session / no `$DISPLAY`), so the session cannot come up; refused before spawning Godot (Phase 2, #345). |
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -231,13 +231,17 @@ that implements it (ADR-0002), not by this ADR. The agent-facing contract (typed
 models, `--json`, `--schema` gate, registered `GdaError.code`s) is identical to
 headless.
 
-> **Amendment (2026-08-25, #661):** a multi-frame live operation's declared
-> frame count is a CEILING, not the required duration. The harness window may
-> complete EARLY with a success payload once its outcome is decided (the
-> predicate capture completes on the first frame its predicate holds), provided
-> every input event the request declared has been applied before the reply —
-> an early completion never skips an accepted event. The one-shot RPC shape is
-> unchanged: still exactly one reply per request.
+> **Amendment (2026-08-25, #661; revised same day after the #743 re-review):**
+> a multi-frame live operation's declared frame count is a CEILING, not the
+> required duration. The harness window may complete EARLY with a success
+> payload once its outcome is decided (the predicate capture completes on the
+> first frame its predicate holds), provided every input event the request
+> declared has been applied before the reply — an early completion never skips
+> an accepted event, and a declared event that FAILS makes the reply that
+> typed failure even when a capture already succeeded (the capture payload is
+> discarded; later events still drain, so scheduled releases fire on the error
+> path too). The one-shot RPC shape is unchanged: still exactly one reply per
+> request.
 
 ## Considered options
 
--- a/docs/adr/0017-gda-daemon-live-execution-mechanism.md
+++ b/docs/adr/0017-gda-daemon-live-execution-mechanism.md
@@ -231,6 +231,14 @@ that implements it (ADR-0002), not by this ADR. The agent-facing contract (typed
 models, `--json`, `--schema` gate, registered `GdaError.code`s) is identical to
 headless.
 
+> **Amendment (2026-08-25, #661):** a multi-frame live operation's declared
+> frame count is a CEILING, not the required duration. The harness window may
+> complete EARLY with a success payload once its outcome is decided (the
+> predicate capture completes on the first frame its predicate holds), provided
+> every input event the request declared has been applied before the reply —
+> an early completion never skips an accepted event. The one-shot RPC shape is
+> unchanged: still exactly one reply per request.
+
 ## Considered options
 
 - **Attach to a human-opened editor via an EditorPlugin (godot-mcp-pro's model)** —
--- a/docs/adr/0020-live-state-consistency.md
+++ b/docs/adr/0020-live-state-consistency.md
@@ -60,15 +60,21 @@ engine session's runtime state:
 this Phase-2 live layer under these scoped guarantees — not for the stateless Phase-1
 headless CLI (ADR-0001). Recorded here and via a pointer on ADR-0000.
 
-> **Amendment (2026-08-25, #661):** the predicate-gated capture reads its two
-> facts at the SAME frame boundary: the awaited property as the game last
-> wrote it, and the viewport texture as the engine last presented it — both
-> belong to the frame that just completed, verified frame-by-frame against a
-> live probe. The declared limit: when a game updates a visual one frame
-> AFTER the property it gates on, the captured image trails the property by
-> that game-side frame; the contract tells the caller to gate on the visual's
-> own property when exact pixels matter. Frame coherence per ADR-0020 is
-> engine-side; a game's internal property-to-visual latency is the game's.
+> **Amendment (2026-08-25, #661; revised same day after the #743 re-review):**
+> the predicate-gated capture reads its two facts at the SAME frame boundary,
+> and each window tick EVALUATES BEFORE it injects that tick's input events —
+> so the observed property is always the state of the previously COMPLETED
+> frame, which is exactly the frame the viewport texture presents. Verified
+> live on both trigger paths: a `_process`-driven flip is observed together
+> with its own presentation, and a state written by an injected event's
+> synchronous `_input` callback is observed one boundary later, together with
+> its presentation. Declared consequences: the predicate observes
+> frame-boundary state only — a value overwritten before its frame completes
+> is never observable (the typed unmet error, never a capture of mismatched
+> pixels) — and a game that updates a visual one frame after the property it
+> gates on trails by that game-side frame; gate on the visual's own property
+> when exact pixels matter. Frame coherence per ADR-0020 is engine-side; a
+> game's internal property-to-visual latency is the game's.
 
 ## Considered options
 
--- a/docs/adr/0020-live-state-consistency.md
+++ b/docs/adr/0020-live-state-consistency.md
@@ -60,6 +60,16 @@ engine session's runtime state:
 this Phase-2 live layer under these scoped guarantees — not for the stateless Phase-1
 headless CLI (ADR-0001). Recorded here and via a pointer on ADR-0000.
 
+> **Amendment (2026-08-25, #661):** the predicate-gated capture reads its two
+> facts at the SAME frame boundary: the awaited property as the game last
+> wrote it, and the viewport texture as the engine last presented it — both
+> belong to the frame that just completed, verified frame-by-frame against a
+> live probe. The declared limit: when a game updates a visual one frame
+> AFTER the property it gates on, the captured image trails the property by
+> that game-side frame; the contract tells the caller to gate on the visual's
+> own property when exact pixels matter. Frame coherence per ADR-0020 is
+> engine-side; a game's internal property-to-visual latency is the game's.
+
 ## Considered options
 
 - **Multi-writer concurrent clients** — rejected for Phase 2: it contradicts the
--- a/docs/adr/0040-per-command-group-modules.md
+++ b/docs/adr/0040-per-command-group-modules.md
@@ -101,12 +101,18 @@ src/gda/
 5. **Dependency direction**: `cli` → `commands/*` → `dispatch` → `headless` →
    runners / `errors` / `models` → foundation. A group may import another
    group's public symbol one-way where the language genuinely shares a shape —
-   three such edges exist: `node` → `scene` for `SceneNode` and
+   four such edges exist: `node` → `scene` for `SceneNode` and
    `derive_scene_root_name` (the filename-stem default an `--instance`
    composition reuses); `shader` → `script` for the `ScriptSetMode` edit
-   interface `shader set` reuses; and `logger` → `diag` for the `SourceFrame`
+   interface `shader set` reuses; `logger` → `diag` for the `SourceFrame`
    location and the `--limit` option the two log-reading groups share (the
-   ADR-0022/0026 lineage). No reciprocal group imports. A shape **no single
+   ADR-0022/0026 lineage); and `screen` → `input` for the
+   `InputSequenceEvent` union that `screen capture --await-events` embeds
+   (#661): the event shapes are the input group's owned contract — the
+   discriminated union, its per-kind field rules, and their harness
+   application — and the predicate capture REUSES them verbatim rather than
+   forking a second event vocabulary, so the edge points at the owner and
+   stays one-way (added 2026-08-25). No reciprocal group imports. A shape **no single
    group owns** stays in the `gda.models` core rather than moving into one,
    because several groups read it (`NodeProperty`, `EngineVersion`,
    `MAX_WINDOW_FRAMES`). A shape a **single group does own** moves to that
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -894,9 +894,12 @@ headless is unaffected (4.4+, cross-platform).
   sequence` takes) INSIDE the same window at their process-clock `frame` offsets — the
   atomic input-and-capture form, so a 3–8-frame transient triggered by the input
   cannot be missed by a second CLI round trip; physics-clock offsets are refused. An
-  offset at or beyond the ceiling still fires — the reply waits for every declared
-  event — but can no longer satisfy the predicate, whose scan ends at the ceiling; the
-  published schema and the model accept exactly the same event set (ADR-0015). Every
+  offset at or beyond the PREDICATE ceiling still fires — the reply waits for every
+  declared event — but can no longer satisfy the predicate, whose scan ends at that
+  ceiling. Every event offset is nevertheless at most 599, so the TOTAL drain stays
+  within the shared 600-frame live-window ceiling; both limits and the modifier
+  vocabulary are published in the schema, and the schema/model event sets agree
+  (ADR-0015). Every
   declared event fires before the reply even when the predicate matches first, so no
   injected press is left held — and a declared event that FAILS makes the whole
   capture that typed failure (the capture payload is discarded, no file is written,
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -893,15 +893,26 @@ headless is unaffected (4.4+, cross-platform).
   additionally applies input-sequence events (the same discriminated union `input
   sequence` takes) INSIDE the same window at their process-clock `frame` offsets — the
   atomic input-and-capture form, so a 3–8-frame transient triggered by the input
-  cannot be missed by a second CLI round trip; physics-clock offsets are refused,
-  every offset must be inside the window, and every declared event fires before the
-  reply even when the predicate matches first, so no injected press is left held. The
-  predicate compares JSON scalars (numbers numerically, strings against the String
-  rendering). The coherence contract, verified frame-by-frame against a live probe:
-  the property and the pixels are read at one boundary and both belong to the frame
-  that just COMPLETED — a game that updates a visual one frame after the property it
-  gates on trails by that game-side frame, so gate on the visual's own property when
-  exact pixels matter (ADR-0020 amendment).
+  cannot be missed by a second CLI round trip; physics-clock offsets are refused and
+  every offset must be inside the window (a cross-field bound the emitted JSON Schema
+  cannot state — enforced at validation, disclosed in the field description). Every
+  declared event fires before the reply even when the predicate matches first, so no
+  injected press is left held — and a declared event that FAILS makes the whole
+  capture that typed failure (the capture payload is discarded, no file is written,
+  later events still drain). The predicate compares JSON scalars (numbers
+  numerically, strings against the String rendering). The coherence contract,
+  verified live on both trigger paths (ADR-0020 amendment): each tick EVALUATES
+  BEFORE it injects, so the observed property is always the state of the previously
+  COMPLETED frame — exactly the frame the captured texture presents. A
+  `_process`-driven flip is observed with its own presentation; a state written by an
+  injected event's synchronous callback is observed one boundary later, together with
+  its presentation. Consequences: the predicate sees frame-boundary state only (a
+  value overwritten before its frame completes is never observable — the typed unmet
+  error, never a capture of mismatched pixels); an event's effect is observable from
+  the NEXT boundary (leave one frame between the last state-changing event and the
+  ceiling); and a game that updates a visual one frame after the property it gates on
+  trails by that game-side frame — gate on the visual's own property when exact
+  pixels matter.
 - **`perf` (runtime performance monitoring):** `perf monitors` snapshots the running
   game's instantaneous Performance counters in one frame (shipped, #223); `perf
   monitor --property … --frames N` / `--signal … --frames N` collects a per-frame
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -893,9 +893,10 @@ headless is unaffected (4.4+, cross-platform).
   additionally applies input-sequence events (the same discriminated union `input
   sequence` takes) INSIDE the same window at their process-clock `frame` offsets — the
   atomic input-and-capture form, so a 3–8-frame transient triggered by the input
-  cannot be missed by a second CLI round trip; physics-clock offsets are refused and
-  every offset must be inside the window (a cross-field bound the emitted JSON Schema
-  cannot state — enforced at validation, disclosed in the field description). Every
+  cannot be missed by a second CLI round trip; physics-clock offsets are refused. An
+  offset at or beyond the ceiling still fires — the reply waits for every declared
+  event — but can no longer satisfy the predicate, whose scan ends at the ceiling; the
+  published schema and the model accept exactly the same event set (ADR-0015). Every
   declared event fires before the reply even when the predicate matches first, so no
   injected press is left held — and a declared event that FAILS makes the whole
   capture that typed failure (the capture payload is discarded, no file is written,
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -886,19 +886,22 @@ headless is unaffected (4.4+, cross-platform).
 - **`screen` / capture:** running-game viewport screenshot, multi-frame capture. The
   `--await-*` predicate (shipped, #661) holds a `screen capture` game-side until
   `node.property == value` first holds — checked once per PROCESS frame, up to
-  `--await-frames` (default 60, ceiling 600) — then captures THAT frame and reports the
-  predicate evidence (`observed` value, absolute `engine_frame`, window-relative
-  `frames_waited`); a predicate that never holds is the typed `live_predicate_unmet`
-  carrying the last observed value. `--await-events` additionally applies
-  input-sequence events (the same discriminated union `input sequence` takes) INSIDE
-  the same window at their process-clock `frame` offsets — the atomic
-  input-and-capture form, so a 3–8-frame transient triggered by the input cannot be
-  missed by a second CLI round trip; physics-clock offsets are refused, and every
-  offset must be inside the window. The predicate compares JSON scalars (numbers
-  numerically, strings against the String rendering); evaluation happens at the
-  process-frame boundary and the captured image is the most recently PRESENTED frame,
-  so a property written after that frame's draw can be one frame ahead of the pixels —
-  the documented limit.
+  `--await-frames` (default 60, ceiling 600) — then captures at that SAME frame
+  boundary and reports the predicate evidence (`observed` value, absolute
+  `engine_frame`, window-relative `frames_waited`); a predicate that never holds is
+  the typed `live_predicate_unmet` carrying the last observed value. `--await-events`
+  additionally applies input-sequence events (the same discriminated union `input
+  sequence` takes) INSIDE the same window at their process-clock `frame` offsets — the
+  atomic input-and-capture form, so a 3–8-frame transient triggered by the input
+  cannot be missed by a second CLI round trip; physics-clock offsets are refused,
+  every offset must be inside the window, and every declared event fires before the
+  reply even when the predicate matches first, so no injected press is left held. The
+  predicate compares JSON scalars (numbers numerically, strings against the String
+  rendering). The coherence contract, verified frame-by-frame against a live probe:
+  the property and the pixels are read at one boundary and both belong to the frame
+  that just COMPLETED — a game that updates a visual one frame after the property it
+  gates on trails by that game-side frame, so gate on the visual's own property when
+  exact pixels matter (ADR-0020 amendment).
 - **`perf` (runtime performance monitoring):** `perf monitors` snapshots the running
   game's instantaneous Performance counters in one frame (shipped, #223); `perf
   monitor --property … --frames N` / `--signal … --frames N` collects a per-frame
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -883,7 +883,22 @@ headless is unaffected (4.4+, cross-platform).
   spelling differs by kind — `pressed` belongs to `mouse_button` alone, an `action`
   releases with `release`, a `key` with `released` — and a field from another kind is
   refused with the spelling this kind uses instead.
-- **`screen` / capture:** running-game viewport screenshot, multi-frame capture.
+- **`screen` / capture:** running-game viewport screenshot, multi-frame capture. The
+  `--await-*` predicate (shipped, #661) holds a `screen capture` game-side until
+  `node.property == value` first holds — checked once per PROCESS frame, up to
+  `--await-frames` (default 60, ceiling 600) — then captures THAT frame and reports the
+  predicate evidence (`observed` value, absolute `engine_frame`, window-relative
+  `frames_waited`); a predicate that never holds is the typed `live_predicate_unmet`
+  carrying the last observed value. `--await-events` additionally applies
+  input-sequence events (the same discriminated union `input sequence` takes) INSIDE
+  the same window at their process-clock `frame` offsets — the atomic
+  input-and-capture form, so a 3–8-frame transient triggered by the input cannot be
+  missed by a second CLI round trip; physics-clock offsets are refused, and every
+  offset must be inside the window. The predicate compares JSON scalars (numbers
+  numerically, strings against the String rendering); evaluation happens at the
+  process-frame boundary and the captured image is the most recently PRESENTED frame,
+  so a property written after that frame's draw can be one frame ahead of the pixels —
+  the documented limit.
 - **`perf` (runtime performance monitoring):** `perf monitors` snapshots the running
   game's instantaneous Performance counters in one frame (shipped, #223); `perf
   monitor --property … --frames N` / `--signal … --frames N` collects a per-frame
--- a/scripts/harness_version_guard.py
+++ b/scripts/harness_version_guard.py
@@ -1,0 +1,120 @@
+"""Fail when bundled harness bytes change without a numeric version increase.
+
+The unit-tier hash pin catches accidental edits inside one checkout. This guard
+compares the merge base and head, which is the information a current-snapshot
+test cannot recover: when ``gda_harness.gd`` changes in the reviewed range, the
+head's HARNESS_VERSION must be numerically greater. CI supplies the PR base/head
+or the before/after commits of a push.
+
+Stdlib only, so the workflow can run it without another project dependency.
+"""
+
+from __future__ import annotations
+
+import argparse
+import re
+import subprocess
+import sys
+
+HARNESS_PATH = "src/gda/harness/gda_harness.gd"
+INSTALL_PATH = "src/gda/harness/install.py"
+_VERSION = re.compile(rb'^HARNESS_VERSION = "([0-9]+)"$', re.MULTILINE)
+
+
+class HarnessVersionGuardError(Exception):
+    """The two revisions cannot prove a valid harness version transition."""
+
+
+def _version(source: bytes, revision: str) -> int:
+    matches = _VERSION.findall(source)
+    if len(matches) != 1:
+        raise HarnessVersionGuardError(
+            f"{revision} must declare exactly one numeric HARNESS_VERSION assignment"
+        )
+    return int(matches[0])
+
+
+def check_change(
+    base_harness: bytes,
+    base_install: bytes,
+    head_harness: bytes,
+    head_install: bytes,
+) -> None:
+    """Require a numeric version increase exactly when harness bytes changed."""
+    if base_harness == head_harness:
+        return
+    base_version = _version(base_install, "base")
+    head_version = _version(head_install, "head")
+    if head_version == base_version:
+        raise HarnessVersionGuardError(
+            "bundled harness bytes changed but HARNESS_VERSION remained "
+            f"{head_version}; increase it and update the current hash pin"
+        )
+    if head_version < base_version:
+        raise HarnessVersionGuardError(
+            "HARNESS_VERSION must increase when bundled harness bytes change "
+            f"(base {base_version}, head {head_version})"
+        )
+
+
+def _git_blob(revision: str, path: str) -> bytes:
+    result = subprocess.run(
+        ["git", "show", f"{revision}:{path}"],
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    if result.returncode != 0:
+        detail = result.stderr.decode(errors="replace").strip()
+        raise HarnessVersionGuardError(
+            f"cannot read {path} at revision {revision}: {detail}"
+        )
+    return result.stdout
+
+
+def _git_merge_base(base: str, head: str) -> str:
+    result = subprocess.run(
+        ["git", "merge-base", base, head],
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    if result.returncode != 0:
+        detail = result.stderr.decode(errors="replace").strip()
+        raise HarnessVersionGuardError(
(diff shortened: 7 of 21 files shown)


qinhaihong-red:
Adversarial review at exact head `27ee69a37bfbeabf097f55ed1f20a0b1b0a44c0f` (base `85d8a285eacd2b7523f01216b6eca868cab59933`): **changes required**.

The four requested review axes are kept independent below; repeated findings are intentional.

## Markdown (`artifact-review`) — Changes required

1. **The docs conflate the predicate-observation frame with the PNG pixel frame.** `docs/command-catalog.md:889` says the command “captures THAT frame”, while lines 899–901 acknowledge that it captures the most recently presented frame and that the property can be one frame ahead of the pixels. `src/gda/skill/SKILL.md:176` and the PR acceptance table similarly overstate deterministic capture of the visual transient. A real-Godot counterexample changed a `ColorRect` from blue to red when the predicate became true: the result reported the matching property value, but the PNG pixel was still blue. Either make the rendered image correspond to the observed state, or consistently document the presentation lag and narrow the acceptance claim.

2. **The PR body says schema descriptions disclose the frame-offset limitation, but the emitted schema does not.** None of the `await_*` or predicate report descriptions explain the “most recently presented frame” behavior, while `screen capture --help` still says “captures THAT frame”. Synchronize the model-driven schema descriptions, help, catalog, Skill, and PR claim.

3. **Accepted ADR-0040 is stale.** `src/gda/commands/screen.py:29` adds a fourth command-group edge, `screen -> input`, while `docs/adr/0040-per-command-group-modules.md:101-115` still says there are three and enumerates only those three. Record the new owner-consumer edge and its rationale.

4. **Accepted ADR-0017/0020 do not describe the new completion and frame-consistency boundaries.** ADR-0017 still describes finalization only after the requested sample count, but predicate success can now complete early. ADR-0020 describes a coherent single-frame live result without accounting for current-process-frame predicate evidence paired with a previously presented viewport image. Add a scoped amendment for both boundaries.

Terminology/prose were checked against `CONTEXT.md`, issue #661, the affected accepted ADRs, and ASD-STE100. No other material terminology or prose defects were found; the ADR-0002 `live_predicate_unmet` row is consistent with the error registry.

## Architecture (`design-domain-modular-architecture`) — Changes required

### ARC-743-001 — [P1] Capture readiness prematurely completes the whole composite operation

`_begin_predicate_capture()` applies only events due at the current offset and returns `complete` as soon as the predicate matches (`gda_harness.gd:1370-1403`). `_advance_window()` then immediately clears the window (`309-340`), so accepted future events never run.

Real-Godot counterexample: action press@0, release@5, predicate matched at frame 1; the RPC returned success, but a subsequent `game get` still reported the action as pressed. This leaks input state into later live operations in the same Engine session.

Minimal direction: preserve the first matching capture payload but continue the window through all accepted events before replying, or reliably compensate all held input state. Add real-engine coverage for key, action, and mouse-button cleanup, including error paths.

Evidence state: `open`. It would be disproved if all scheduled release events execute before the reply and subsequent live reads show no held state.

### ARC-743-002 — [P2] Runtime-property resolution now has another authority and performs an unreported pre-read

`_runtime_property_exists()` (`gda_harness.gd:1414-1421`) reimplements the storage-first/script-variable-second rule already present in `game get`, while the adjacent set path has a third classification path. Its script-variable branch calls a helper that executes `node.get()` and projection before the predicate window formally samples the property, after which the sampler reads it again.

Extract a metadata-only runtime-property resolver shared by the applicable get/predicate/set paths; perform the actual value read only inside the owning use case. Verify getter invocation counts with a side-effecting getter corpus.

Evidence state: `open`. Remaining validation includes custom getters, dynamic property lists, and getter-only script variables.

### ARC-743-003 — [P2] The new peer dependency is not represented in the dependency authority

The `screen -> input` dependency is one-way, and `InputSequenceEvent` still has a clear owner in the input command group. The smallest architecture fix is to record the allowed edge in ADR-0040 and guard against `input -> screen`; duplicating the union or introducing an ownerless `common` module is not necessary.

Evidence state: `open`; it would be disproved if ADR-0040 were explicitly non-exhaustive and another authoritative dependency map already covered this edge.

No evidence requires a new bounded context, aggregate, event bus, or RPC. The existing Python -> daemon -> harness direction can remain.

## Standards (`review`) — Changes required

1. **[Hard] The emitted input JSON Schema is not equivalent to runtime validation.** A standard Draft 2020-12 validator accepts `await_node` alone, `await_value: null`, and an empty `await_events`, while `ScreenCaptureParams.model_validate()` rejects all three. This violates ADR-0015’s single-model schema contract. Publish the cross-field, non-null, and non-empty constraints and add a standard-validator/Pydantic parity corpus.

2. **[Hard] A success receipt is not validated or correlated with its request.** `_PredicateReply` permits missing fields and negative/unbounded frames, `_CaptureReply.predicate` is optional, and the recipe writes the PNG without requiring predicate evidence for a gated request or checking node/property/expected/observed/frame bounds. Exact-head probes accepted both a missing predicate report and fabricated evidence (`/root/Wrong`, wrong property/value, `engine_frame=-1`, `frames_waited=999`). Reject these as `contract_violation` before writing the output file.

3. **[Hard] Early predicate success drops accepted release events.** Independently of the architecture finding, this violates the documented live state-consistency and input-lifecycle standards. Complete or compensate the accepted input sequence before returning success.

4. **[Hard] The new command-group dependency is absent from the documented standards authority.** The Standards axis suggests moving the shared event model to a neutral lower owner; the Architecture axis instead finds `input` to be a valid owner and recommends documenting the allowed edge. These recommendations are deliberately left independent for adjudication.

No separate Fowler-style smell was found beyond the documented standard breaches above.

## Spec (`review`) — Changes required

1. **[P1] The optional atomic input-and-capture form can permanently skip future release events.** Issue #661 requires one atomic operation that injects input and applies predicate-gated capture. Returning after the predicate match while leaving part of the accepted input sequence unexecuted does not satisfy that operation.

2. **[P1] A gated request can succeed with missing or fabricated observation evidence.** Issue #661 requires the result to report the observed frame and predicate value. The current client accepts no predicate report or an unrelated, internally inconsistent one. Require and correlate the evidence, including equality and frame bounds.

3. **[P1] The live regression does not test capture of the requested transient visual state.** `tests/test_e2e_screen.py:220-289` changes a property while the `ColorRect` remains visually constant, then checks only predicate metadata and PNG magic bytes. It cannot detect the demonstrated property/pixel skew. Make the target state visually distinct, decode the image, and assert its pixels across repeated captures.

No independent scope creep was found.

## Verification

- Exact-head focused tests: `180 passed`
- Exact-head real-Godot/windowed suite: `32 passed, 41 deselected`
- `git diff --check`: passed
- Additional adversarial probes reproduced the stale visual frame, held input state, fabricated receipt acceptance, and schema/runtime mismatch
- GitHub ruff, pyright, Python/package, co-install, and release-scope checks are green
- GitHub `Godot e2e tests` is **SKIPPED**; the local real-engine passes do not invalidate the counterexamples above

Summary: **Markdown: 4 required (worst: incorrect same-frame contract); Architecture: 3 findings, worst P1; Standards: 4 Hard findings; Spec: 3 P1 findings.**

qinhaihong-red:
## Review response — all four axes closed at head `2270e22f`

Baseline: base `85d8a285`, reviewed head `27ee69a3`, new head `2270e22f`. Every mechanism was reproduced or probed against a real engine before the fix; the one finding whose premise the probe contradicted (the pixel-frame claim) is answered with the probe's data rather than a code change in the direction first suggested — details under Markdown 1.

| Claim | Decision | Change |
| --- | --- | --- |
| ARC-743-001 / Standards 3 / Spec 1 — early match drops accepted release events | **Fully adopt** | The window replies only after EVERY declared event fired; first decided outcome wins; proven live for key, mouse-button, and action holds, plus the unmet-error path |
| Standards 2 / Spec 2 — missing or fabricated receipts accepted | **Fully adopt** | CLI-side correlation before the file is written: report present iff gated, naming this request, observed satisfying the declared predicate, wait inside the window; wire model bounds the frames (ge=0); six regressions |
| Standards 1 — emitted schema ≠ runtime validation | **Fully adopt** | The trio, ceiling/events dependencies, non-empty events, and process-clock-only items are published Draft 2020-12 constraints (`allOf` if/then, `minItems`, items `not`); a validator/pydantic parity corpus pins agreement; the one value-dependent rule (offset inside the window) is pinned as the disclosed model-only residual — Draft 2020-12 cannot state a cross-field numeric relation |
| Markdown 1 / Spec 3 — "captures THAT frame" vs stale pixels | **Partially adopt** (docs + e2e adopted; the one-frame-delay direction refuted by measurement) | See below — the same-boundary contract, verified frame-by-frame, now stated everywhere; the e2e decodes pixels |
| Markdown 2 — schema descriptions silent on the frame contract | **Fully adopt** | `await_value` and `predicate.observed` descriptions, help, catalog, and SKILL all carry the same-boundary contract and its game-side limit |
| Markdown 3 / ARC-743-003 / Standards 4 — the `screen → input` edge undocumented | **Fully adopt (Architecture's direction)** | ADR-0040 records the FOURTH one-way edge with its rationale (the union stays with its owner; no neutral module, no duplication) — adjudicated over the Standards axis's move-to-neutral-owner option because the three existing edges set exactly this precedent |
| Markdown 4 — ADR-0017/0020 boundaries | **Fully adopt** | ADR-0017 amendment: the frame count is a CEILING and early completion never skips an accepted event; ADR-0020 amendment: the same-boundary coherence contract and the game-side latency limit |
| ARC-743-002 — property resolution re-implemented + pre-read | **Fully adopt** | One metadata-only rule (`_runtime_property_declared`, storage or script variable off `get_property_list()` alone) shared with `game get`'s script-variable path via the extracted `_is_script_variable`; the value is read only inside the window — proven by a getter-counting e2e: a scripted getter runs exactly `frames_waited + 1` times, no pre-read, no capture-tick re-read |

### Markdown 1, measured: the pixels and the property are coherent at the SAME boundary

Your counterexample sent me to instrument the engine rather than guess. A live probe (an autoload — the harness's exact vantage — observing a scene that flips a `ColorRect` red on the single frame `phase == 5`) prints, frame by frame: `AUTOLOAD sees phase=5 pixel=RED` — the flipped property and the flipped pixel on the SAME tick, every cycle. The mechanism: the autoload processes before the scene, so the property it reads and the presentation the texture holds both belong to the frame that just COMPLETED. I also tested the suggested direction empirically: delaying the capture by one frame returns the frame AFTER the match (blue), by two frames likewise — both delays were tried and both fail the pixel e2e that now pins the contract.

What your probe most plausibly observed is the declared limit that now stands in the contract everywhere (schema descriptions, help, catalog, SKILL, ADR-0020 amendment): when a game updates a visual one frame AFTER the property it gates on, the image trails by that game-side frame — gda cannot know a game's internal property-to-visual latency, so the contract says to gate on the visual's own property when exact pixels matter. The new e2e is exactly your requested shape: the target state is visually distinct (red on the matched frame alone), the PNG is decoded (a full five-filter PNG reader), and the matched frame's pixels are asserted three times per session. If your scene bound the visual to the gated property in the same tick and still saw stale pixels, that would refute this measurement — I'd want that scene verbatim.

### Validation at `2270e22f`

- Red-proof on `27ee69a3` against its `src/gda`: the no-report and wrong-target correlation tests and the schema parity corpus fail there, and the held-state drain e2e fails with the exact reported leak — `key_down` stays `True` after an early-match reply.
- Real-engine e2e, serial: screen **11 passed** — pixels×3, typed unmet, input-triggered atomic with release drain, the three-way (key / mouse-button / action) held-state drain, the error-path drain, and the getter-count proof; adjacent suites over the shared window base: input **16**, perf **6**, game **2**, harness-install **4**.
- Fast suite **1844 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean; harness pin re-pinned (version 9, new bytes).
- GitHub checks on `2270e22f`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
Exact-head re-review at `2270e22f5c4a488fe2af8e779b8d1d702564e495` (previous reviewed head `27ee69a37bfbeabf097f55ed1f20a0b1b0a44c0f`): **changes still required**.

I replayed the prior findings, the author response, and adjacent failure paths. The receipt-correlation gate, successful release draining, metadata-only property resolution, and the ADR-0040 dependency edge are closed. The residue below remains; repeated findings are intentionally kept under their independent review axes.

## Markdown (`artifact-review`) — Changes required

1. **The new same-boundary documentation is disproved by the atomic-input path.** `docs/command-catalog.md:886-904`, the ADR-0020 amendment at lines 63–71, command help/schema descriptions, and the PR body now say the property and pixels both belong to the same completed frame. The original real-Godot counterexample still fails at this head: one `_input` callback sets both `flash = 4` and `$Rect.color = red`; the result reports `observed = 4`, but the PNG pixel is still blue `(0, 0, 255)`. This is not a game-side one-frame property-to-visual delay—the two writes happen in the same callback. `src/gda/skill/SKILL.md:176` also still contains neither the claimed same-boundary contract nor its limitation, despite the response saying the Skill was updated.

2. **“Every declared event is applied” is not true on a late event failure.** ADR-0017:234-240 says every declared input event has been applied before the reply; the catalog and PR body say every event fires. A real-Godot probe matched early, then a scheduled unknown action returned `live_unknown_action`; the harness ignored that error, returned exit 0, and wrote the PNG. The event was attempted but not applied.

3. **The schema-parity closure claim still conflicts with ADR-0015.** The new parity test explicitly pins `event.frame >= await_frames` as schema-valid/model-invalid. There are also undisclosed reverse divergences: JSON Schema rejects `await_frames: "10"` and event `frame: "1"`, while Pydantic accepts and coerces both. An internal helper docstring is not a public disclosure, and ADR-0015 requires the emitted schema to be exactly what `--params-json` accepts.

Terminology/prose were rechecked against `CONTEXT.md`, issue #661, the affected accepted ADRs, and ASD-STE100. No additional terminology or prose findings remain.

## Architecture (`design-domain-modular-architecture`) — Changes required

### ARC-743-001 — [P1] Still open: capture readiness is separated from release execution, but not from the final operation verdict

`gda_harness.gd:1350-1352` declares that the first decided outcome wins; lines 1382–1394 retain an event error only while no capture outcome exists. Releases now drain successfully, but a later declared event can fail while the composite operation still reports success.

Real-Godot counterexample: action press `qa_probe`@0 satisfies the predicate; unknown action@5 should produce `live_unknown_action`; release `qa_probe`@6 executes; the command instead exits 0 and writes the PNG. Preserve the first capture payload while draining, but let any declared-event failure determine the final reply and prevent the file write.

Evidence state: `open`. Disconfirming condition: the same probe returns the typed event failure, writes no file, and still leaves `qa_probe` released.

### ARC-743-004 — [P1] The new state-consistency claim conflates the input-processing boundary with the presentation boundary

At `gda_harness.gd:1387-1423`, the sampler injects input, immediately reads the property, and then reads the most recently presented viewport texture. Synchronous input handling can update both the property and visual node before the predicate read, but the new pixels have not yet been presented. The exact real-engine blue-pixel counterexample above confirms this path.

Capture after the matched state has been drawn, or explicitly redefine and constrain the atomic-input contract; the current architecture does not provide the visual-frame synchronization issue #661 requires.

Evidence state: `open`. Disconfirming condition: the same-callback counterexample reliably captures the red pixel.

Resolved architecture findings: ARC-743-002 (metadata-only resolution/getter count) and ARC-743-003 (`screen -> input` authority).

## Standards (`review`) — Changes required

1. **[Hard] A late scheduled event failure is silently converted into success.** This violates ADR-0017’s typed live-failure and complete-event contract. Draining releases is fixed, but the operation must not return success when one of its declared events failed.

2. **[Hard] Input schema/runtime parity remains divergent.** `screen.py:61-108` manually mirrors cross-field rules; `tests/test_screen_commands.py:921-1018` blesses one model-only rule, while omitting the string-coercion divergences above. This conflicts with ADR-0015:46-50,145-147 (“same model”, “exactly”, “never hand-maintained twice”).

3. **[Hard] The public success schema understates the enforced receipt.** `CapturePredicateReport` (`screen.py:256-283`) still makes `expected` and `observed` optional/nullable and gives `engine_frame` and `frames_waited` no minimum. The new wire gate guarantees those values are present, non-null, correlated, and nonnegative. Therefore `--schema` advertises successful outputs the implementation cannot emit, contrary to ADR-0004’s success-output self-description contract.

The pre-write correlation gate and ADR-0040 dependency authority are closed. No separate Fowler-style smell remains.

## Spec (`review`) — Changes required

1. **[P1] Atomic input-triggered capture still returns pixels from before the matched visual state.** Issue #661 requires deterministic capture of a declared transient at ADR-0020’s frame-coherence point. The new repeated red-pixel E2E covers a periodic `_process` transition, but `tests/test_e2e_screen.py:415-445`, the atomic input-triggered case, checks only PNG magic. Add a decoded-pixel assertion for the input-triggered path and make the runtime satisfy it.

2. **[P1] Atomic input-and-capture can report success after a requested input event fails.** The declared input sequence was not fully injected, so the operation does not meet issue #661’s single atomic-operation requirement. Preserve late event errors while still draining cleanup events.

Receipt correlation and successful release draining are closed. No scope creep was found.

## Verification

- Exact snapshot fast suite: `1842 passed, 2 skipped`
- Exact snapshot screen real-Godot E2E: `11 passed`
- Two temporary real-Godot adversarial probes both failed, reproducing the stale visual frame and late-event false success
- Ruff check/format: passed; pyright: `0 errors`; `git diff --check`: passed
- GitHub Python/package, ruff, pyright, and co-install checks are green
- GitHub `Godot e2e tests` remains **SKIPPED**

Summary: **Markdown: 3 required; Architecture: 2 open P1; Standards: 3 Hard; Spec: 2 P1 (incorrect matched pixels and atomic false success).**

qinhaihong-red:
## Re-review response — all findings adopted at head `75afef5c`

Baseline: base `85d8a285`, re-reviewed head `2270e22f`, new head `75afef5c`. Your same-callback counterexample exposed the real mechanism my round-1 probe missed: the sampler injected input and then read the property IN THE SAME tick, so a synchronous `_input` write was observed before its frame had been presented. The fix is structural, not a delay constant.

| Claim | Decision | Change |
| --- | --- | --- |
| ARC-743-004 / Markdown 1 / Spec 1 — atomic path returns pre-match pixels | **Fully adopt** | Each window tick now EVALUATES BEFORE it injects: the observed property is always the previously COMPLETED frame's state — exactly the frame the texture presents. State an event writes is observed one boundary later, together with its own presentation |
| ARC-743-001 / Standards 1 / Spec 2 — late event failure becomes false success | **Fully adopt** | During drain, a declared event failure REPLACES a captured success (never an earlier error); the reply is the typed failure, the CLI writes no file, later events still drain |
| Markdown 3 / Standards 2 — parity divergences and disclosure | **Fully adopt** | Strict clock ints: `await_frames` and the sequence union's `frame`/`physics_frame` no longer coerce strings the published `integer` type rejects (the union fix also aligns `input sequence` with its own schema); the parity corpus covers both reverse cases; the one cross-field bound Draft 2020-12 cannot state is disclosed PUBLICLY — in the `await_events` field description (published schema text) and the catalog |
| Standards 3 — success schema understates the enforced receipt | **Fully adopt** | `CapturePredicateReport.expected`/`observed` are required non-null; both frame fields carry `ge=0` — the schema now advertises exactly what the correlation gate guarantees |
| Markdown 1 tail — SKILL.md missing the contract | **Fully adopt** (and my round-1 reply overstated this — the row had the flags but not the contract; my miss) | The `screen` row now states the same-completed-frame contract, the intra-frame limit, the next-boundary event rule, and the event-failure reply |
| Markdown 2 — "every declared event is applied" | **Fully adopt** | ADR-0017's amendment now says: applied before the reply, and a FAILED declared event is the reply even after a successful capture; catalog and body updated to match |

### Both disconfirming conditions, satisfied on a real engine

- **ARC-743-004**: the new e2e is your exact shape — one `_input` callback sets the property (`hit`) AND the visual (a Marker rect turns green); the decoded PNG pixel at the Marker is green `(r<80, g>200)`. Red-proof on `2270e22f`: the same test reads the stale blue `(51,153,229)` there — your counterexample, reproduced and then closed.
- **ARC-743-001**: action press@0 satisfies the predicate, an unknown action@5 fails, release@6 is still declared — the command returns the typed `live_unknown_action`, writes no file, and `act_down` reads `false` afterwards (the release drained). Red-proof on `2270e22f`: exit 0 there — the false success.

### The contract, stated once

Each tick evaluates before it injects, so the predicate observes frame-boundary state only — the state of the frame that just completed, which is exactly the frame the captured texture presents. Three declared consequences now stand in the schema descriptions, help, catalog, SKILL, and the revised ADR-0017/0020 amendments: an intra-frame value (written and overwritten inside one frame) is never observable — the typed unmet error, never a capture of mismatched pixels; an injected event's effect is observable from the NEXT boundary, so the last state-changing event needs one frame of window left; and a game that updates a visual one frame after the property it gates on still trails by that game-side frame.

### Validation at `75afef5c`

- Fast suite **1844 passed**; pyright 0 errors; ruff check + format clean; `git diff --check` clean; harness pin re-pinned.
- Real-engine e2e, serial: screen **13 passed** (process-driven pixels×3, same-callback input pixels, typed unmet, atomic with release drain, three-way held-state drain, error-path drain, late-event-failure reply, getter-count); adjacent suites: input **16**, perf **6**, game **2**, harness-install **4**.
- Red-proof on `2270e22f` against its `src/gda`: the same-callback pixel e2e (stale blue), the late-event-failure e2e (exit 0), and the parity corpus (`await_frames: "10"` — schema rejects, the lax model accepted) all fail there on the reported mechanisms.
- GitHub checks on `75afef5c`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
Exact-head second re-review at `75afef5c1d75edbca8c118c9cae14434783b2bf2` (previous head `2270e22f5c4a488fe2af8e779b8d1d702564e495`): **changes are still required**.

The two runtime P1 mechanisms from the prior re-review are now genuinely closed:

- **ARC-743-004 / Spec 1 — resolved.** The harness now evaluates before injecting each tick. The exact same-callback counterexample sets both `hit = true` and the Marker visual to green; the real-engine test decodes the matching green PNG pixel.
- **ARC-743-001 / Spec 2 — resolved.** A late declared-event failure now replaces a captured success, prevents the PNG write, and still drains later release events. The press@0 / unknown-action@5 / release@6 probe returns `live_unknown_action`, writes no file, and leaves the action released.
- The public `CapturePredicateReport` schema is also closed: `expected` / `observed` are required and non-null, and both frame fields publish `minimum: 0`.

The remaining residue is below. The four requested review axes remain independent.

## Markdown (`artifact-review`) — Changes required

1. **The PR body still overstates schema/runtime parity and the remaining residual.** The strict fixes for `await_frames` and event `frame` / `physics_frame` are effective, but they do not align the whole embedded input-event union. A standard Draft 2020-12 validator rejects each of these values while `ScreenCaptureParams.model_validate()` coerces and accepts it:

   - key `released: "false"`
   - mouse-click `x: "10"`
   - mouse-click `double: "false"`
   - action `strength: "0.5"`
   - action `release: "false"`

   In addition, `tests/test_screen_commands.py:1025-1043` still deliberately pins `event.frame == await_frames` as schema-valid/model-invalid. Therefore the PR body's statement that the parity corpus covers both directions and leaves only the disclosed residual is not accurate at this head. Either make the embedded event scalar fields strict and eliminate the cross-field acceptance difference, or narrow the closure claim to the behavior actually established.

Terminology and prose were rechecked against `CONTEXT.md`, issue #661, the affected accepted ADRs, and ASD-STE100 across the current PR body, ADR amendments, catalog, generated help/schema, and gda skill. The frame-boundary and event-failure descriptions are now coherent; no other terminology or prose finding remains.

## Architecture (`design-domain-modular-architecture`) — Changes required

### ARC-743-005 — [P2] Two different harness bodies declare the same harness version

Between `2270e22f` and this head, the bundled harness hash changed from `792cd491...` to `d01ec327...`, but `HARNESS_VERSION` and `PINNED_HARNESS_VERSION` both remained `"9"`. This conflicts with the harness-install authority in `tests/test_harness_install.py:800-836`, which explicitly requires every harness edit to bump `HARNESS_VERSION` and update both pins.

This is **not** a stale-install claim: `_materialize()` compares the full materialized content, so it still rewrites an older same-version body. The architecture defect is that two implementations report the same version identity, so the version/header can no longer identify the harness bytes it declares.

Smallest fix: bump the harness version to `10` and update both pins. Evidence state: `open`. Disconfirming condition: the new body has a distinct declared version, or the project formally redefines the version as non-identifying metadata and reconciles the install/test authorities.

ARC-743-001 and ARC-743-004 are resolved at this head; ARC-743-002 and ARC-743-003 remain resolved. No new module, bounded context, or dependency edge is needed.

## Standards (`review`) — Changes required

1. **[Hard] Input schema/runtime parity remains divergent.** The additional event-field coercions above and the deliberately retained `frame == await_frames` difference still conflict with ADR-0015's “same model”, “exactly”, and “never hand-maintained twice” requirements. Public disclosure does not make the accepted input sets equal.

2. **[Hard] The harness hash pin does not enforce its documented cross-revision version-bump invariant.** The tests did run and pass. They pass because the change updated the editable golden hash while leaving both version values at `9`; the test only checks the current equalities:

   - `HARNESS_VERSION == PINNED_HARNESS_VERSION`
   - `current_digest == PINNED_HARNESS_SHA256`

   It never checks `hash changed relative to the PR base => version changed relative to the PR base`. Thus the pin catches an accidental harness edit whose golden hash was not updated, but it cannot catch this exact “new golden hash, unchanged version” state. Bump to `10`; if the gate is intended to enforce this automatically, add a base-diff CI check or an equivalent immutable version-to-hash history.

No separate Fowler-style smell remains beyond the duplicated schema/model rule authority already covered by finding 1.

**Standards: 2 open findings, both Hard.**

## Spec (`review`) — Pass

Both prior P1 findings are resolved by real-engine evidence, and no independent scope creep was found.

**Spec: 0 open findings; worst severity: none.**

## Verification

- Exact-head real windowed Godot screen E2E: `13 passed`
- Focused unit/schema/harness tests: `171 passed`
- Non-E2E run: `1841 passed, 2 skipped`; the sole failure was the known unwritable default `uv` cache inside a wheel-build subprocess, and that exact test passed when rerun with a task-scoped writable `UV_CACHE_DIR`
- Ruff check/format, pyright, and `git diff --check`: passed
- Standard-validator/Pydantic adversarial corpus reproduced the five event-field coercion differences above
- GitHub checks are green; GitHub `Godot e2e tests` remains **SKIPPED**, while the local real-engine suite above exercised the two original runtime counterexamples

Summary: **Markdown: 1 required; Architecture: 1 P2; Standards: 2 Hard; Spec: Pass.**

qinhaihong-red:
## Second re-review response — all findings adopted at head `2752ab03`

Baseline: base `85d8a285`, re-reviewed head `75afef5c`, new head `2752ab03`.

| Claim | Decision | Change |
| --- | --- | --- |
| Markdown 1 / Standards 1 — five more union coercions; the retained cross-field difference | **Fully adopt, the eliminate branch** | Every scalar field of the sequence-event union is strict (`released`, `x`, `y`, `double`, `pressed`, `release`, `strength` — following round 2's clock ints), and the pinned variant/op constraint guard propagated the same strictness to the mirrored single-frame op params, so the whole input surface stops coercing strings its published `boolean`/`number` types reject. The former model-only offset-inside-window rule is REMOVED rather than disclosed: the drain applies an out-of-window event anyway (the reply waits for every declared event), it just cannot satisfy the predicate, whose scan ends at the ceiling. The model and the emitted schema now accept exactly the same set — no residual left to disclose |
| ARC-743-005 / Standards 2 — two bodies declared version 9 | **Fully adopt** | `HARNESS_VERSION` 9 → 10, both pins updated; version 9 is retired (it named two in-PR bodies and never merged anywhere) |
| Standards 2's deeper point — the pin cannot catch "new golden hash, unchanged version" | **Fully adopt, the in-repo equivalent** | A new append-only `PINNED_HARNESS_HISTORY` maps every declared version to its ONE body hash, with a guard asserting versions strictly increase, hashes are unique, and the current pins are the last row. A same-version body change now has to REWRITE an existing history row — the guard fails on the duplicate hash/version shape it can see, and a rewritten row is the visible tamper signal for review. Honest boundary: like any in-repo golden, the history is editable; what it adds over the mutable pins is that the edit must falsify a row that claims to be immutable, which is exactly the diff a reviewer catches. A base-diff CI job would enforce it mechanically — left out here as a workflow change beyond this slice; say the word and I will file it as a follow-up issue |

### Parity, this time measured to zero

The corpus now pins, per payload, that a Draft 2020-12 validator and `model_validate` agree: the five reported coercions (`released: "false"`, `x: "10"`, `double: "false"`, `strength: "0.5"`, `release: "false"`) are both-reject; int-for-float is both-accept (JSON `number` admits integers, strict float admits int); the former residual (`frame: 10` with `await_frames: 10`) is both-accept with its semantics documented in the field description and the catalog — an offset at or beyond the ceiling still fires but can no longer satisfy the predicate. The unmet-error e2e now proves that semantics live: the release sits at offset 12 with ceiling 10, the reply is the typed unmet error, and `key_down` reads `false` after it — the out-of-window release fired during the drain.

### Validation at `2752ab03`

- Red-proof on `75afef5c` against its `src/gda`: the parity corpus fails there on the exact reported mechanism (`released: "false"` — schema rejects, the lax model accepted).
- Fast suite **1845 passed** (the strict alignment also holds the variant/op mirror guard green — the constraint discipline propagated instead of drifting); pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: screen **13 passed** (including the out-of-window drain variant); input **16 passed** over the strictened surface.
- GitHub checks on `2752ab03`: green (the `Godot e2e tests` job remains dispatch-only by design).

qinhaihong-red:
Exact-head third re-review at `2752ab03b31935f60c3f4e96214aac64e79d1f1e` (previous head `75afef5c1d75edbca8c118c9cae14434783b2bf2`): **changes are still required**.

The five scalar-coercion counterexamples from the prior review are closed, and the harness body now has the distinct version identity `10`. The remaining residue is below; repeated findings remain intentionally separated by review axis.

## Markdown (`artifact-review`) — Changes required

1. **The documented 600-frame bound is no longer the operation's real bound.** `docs/command-catalog.md:887-899` says the predicate scans up to `--await-frames` (ceiling 600), then returns `live_predicate_unmet`, while events at or beyond that ceiling still drain. The model and schema both accept `await_frames=10` with `event.frame=1_000_000_000`; the harness waits for that offset, so the public caller reaches `live_timeout` instead of the documented bounded predicate error.

2. **The exact schema/model event-set claim is still false.** Draft 2020-12 accepts a key event with `modifiers:["hyper"]`, while Pydantic rejects it through `_validate_modifiers`. Publish the modifier vocabulary structurally or narrow the claim.

3. **The PR body overstates the harness history gate.** The new history is editable in the same change as the two current pins, so it does not mechanically enforce append-only history or make a same-version byte change fail. It also says it records every declared version on the branch lineage while omitting version 9, which this PR declared for multiple different bodies.

No additional terminology or prose issue was found.

## Architecture (`design-domain-modular-architecture`) — Changes required

### ARC-743-006 — [P1] The composite capture dropped the shared live-window ceiling

`screen` reuses the input event union, but not `InputSequenceParams`' aggregate window invariant. `_SequenceEvent.frame` has only `ge=0`; `ScreenCaptureParams` now rejects only physics-clock events. The harness then opens `max(frames, last_event + 1) + 1` frames and explicitly trusts the model to have bounded every live window.

Exact-head counterexample: both the standard schema validator and Pydantic accept an event at frame `1_000_000_000`. The request occupies the serialized live session until daemon timeout and causes session retirement. Preserve the useful new semantics that an event may sit beyond the predicate ceiling, but enforce a schema-visible TOTAL drain bound, e.g. `event.frame <= MAX_WINDOW_FRAMES - 1`.

Evidence state: `open`. It is disproved when every accepted capture request has a total drain window within the shared 600-frame ceiling. ARC-743-005 is resolved by the v10 bump.

## Standards (`review`) — Changes required

1. **[Hard] A valid `await_events` payload bypasses the repository-wide 600-frame anti-stall contract.** This converts the issue's bounded typed-error path into `live_timeout` and channel retirement.
2. **[Hard] Schema/runtime parity still diverges for the modifier vocabulary.** The changed scalar matrix otherwise agrees.
3. **[Hard] `PINNED_HARNESS_HISTORY` does not enforce the documented cross-revision version-bump invariant.** In an adversarial snapshot I changed the harness bytes without bumping v10, rewrote the current hash plus the last history row, and both guard tests passed. A merge-base comparison in CI or another genuinely immutable authority is still required for mechanical enforcement.
4. **[Soft] Stale comments remain.** `src/gda/commands/screen.py:61-72` and `tests/test_screen_commands.py:938-943` still describe the deleted model-only inside-window exception.

No additional Fowler-style smell was found.

## Spec (`review`) — Changes required

1. **[P1] The bounded typed-error acceptance criterion is regressed.** A never-holding predicate with a very late accepted event does not return `live_predicate_unmet` after its stated bound; it waits until daemon `live_timeout`. No other independent Spec regression or scope creep was found.

## Verification

- Exact-head non-E2E: `1843 passed, 2 skipped`
- Exact-head real windowed Godot: screen `13 passed`; input `16 passed`
- Ruff check/format, pyright, and `git diff --check`: passed
- Adversarial probes reproduced the unbounded accepted offset, modifier schema/model divergence, and unchanged-version harness-history bypass
- GitHub's normal checks are green; the GitHub `Godot e2e tests` job remains skipped

Summary: **Markdown: 3 required; Architecture: 1 P1; Standards: 3 Hard + 1 Soft; Spec: 1 P1.**

qinhaihong-red:
Implemented the third re-review findings at exact head `402ecb21520fcd6a2fc82cf61eb721a015287b1f` and refreshed the PR body so it no longer claims an unenforceable append-only history.

Resolution by the original review axes:

### Markdown (`artifact-review`)

- The catalog and PR body now distinguish the predicate ceiling from the total drain ceiling. Events may remain beyond `await_frames`, but their offsets are bounded to `0..599`, keeping the serialized operation within the shared 600-frame bound.
- The exact schema/model parity claim is now true for the reviewed residuals: the modifier vocabulary is emitted as an enum, and the event-offset maximum is emitted structurally.
- The editable `PINNED_HARNESS_HISTORY` claim was removed. The PR body now documents the mechanical cross-revision CI gate instead.

### Architecture (`design-domain-modular-architecture`)

- ARC-743-006 is closed: the shared sequence event authority now publishes `frame` / `physics_frame <= MAX_WINDOW_FRAMES - 1`. `screen capture` can preserve beyond-predicate-ceiling drain semantics without accepting an unbounded live-session monopoly.

### Standards (`review`)

- The total live-window bypass and modifier schema/runtime divergence are covered by executable boundary/parity tests.
- Stale model-only-exception comments were removed.
- Harness identity is now enforced across Git revisions by `scripts/harness_version_guard.py`, wired into the required **Python unit tests and package build** job with full checkout history. If `gda_harness.gd` differs between the merge base and head, the head must have a strictly greater numeric `HARNESS_VERSION`.

### Spec (`review`)

- A valid request can no longer defer the bounded predicate outcome until daemon `live_timeout` by declaring an arbitrarily late event. Frame 600 is rejected; frame 599 remains the accepted total-window boundary.

### Harness-bypass proof

The prior hash test only pinned the current snapshot, so updating the harness, current hash, and editable history together erased the evidence needed to compare revisions. The new gate restores that missing authority:

- historical `2270e22f` → `75afef5c`: **rejected** (`gda_harness.gd` changed while version 9 was reused);
- dev merge base `85d8a285` → delivered head: **accepted** (legitimate version 8 → 10 transition);
- GitHub required-job step `Guard harness version identity`: **success** in the current run.

### Validation

- non-E2E: `1850 passed, 2 skipped, 527 deselected`;
- real windowed Godot: screen `13 passed`, input `16 passed` (`29 passed` total);
- focused contract/guard modules: `156 passed`;
- Pyright: 0 errors; repository-wide Ruff check/format: clean (`468 files`); workflow YAML parse and `git diff --check`: clean;
- all GitHub required checks: green; the normal GitHub Godot E2E job is skipped by policy, with the real-engine evidence supplied locally above.

Conclusion: all findings from the exact-head third re-review are addressed; the PR is ready for an independent final re-review.

qinhaihong-red:
Exact-head fourth re-review at `402ecb21520fcd6a2fc82cf61eb721a015287b1f` (previous head `2752ab03b31935f60c3f4e96214aac64e79d1f1e`): **no remaining findings — the third re-review is closed.**

Verification of each prior finding, independently reproduced in a detached worktree at the exact head:

- **ARC-743-006 / Spec 1 / Standards 1 / Markdown 1 (unbounded drain) — closed.** `frame` / `physics_frame` carry `le = MAX_WINDOW_FRAMES - 1` on the shared event union and the emitted schema publishes `maximum: 599`. The CLI boundary test rejects frame 600 before the wire; the parity corpus both-rejects frame 600 and both-accepts a beyond-predicate-ceiling offset. The catalog now separates the predicate ceiling from the total drain ceiling. Red-proof reproduced: with `src/gda` reverted to `2752ab03` under the head's tests, the boundary test, both parity tests, and the schema-publication test fail (4 failed); all green at the head.
- **Standards 2 / Markdown 2 (modifier vocabulary parity) — closed.** `InputModifier = Literal[...]` is the one authority; the enum is emitted at all four schema sites (`input key`, `input tap` request+result echo, `KeySequenceEvent`). `modifiers: ["hyper"]` is now rejected by both the Draft 2020-12 validator and the model; no stale `unknown modifier(s)` assertion remains.
- **Standards 3 / Markdown 3 (harness history gate) — closed.** The editable `PINNED_HARNESS_HISTORY` and its test are removed; `scripts/harness_version_guard.py` runs in the required *Python unit tests and package build* job (`fetch-depth: 0`, pull_request + push-to-main; dispatch/schedule runs re-test refs already gated). Both proofs reproduced on real history: `2270e22f → 75afef5c` **rejected** (bytes changed, version 9 reused), `85d8a285 → 402ecb21` **accepted** (8 → 10). The previously demonstrated bypass no longer exists because the authority is Git history, not editable test data; missing/ambiguous version declarations fail closed (unit-tested). The regex matches `install.py:90` exactly; `pythonpath`/`extraPaths` include `scripts`.
- **Standards 4 (stale comments) — closed.** Both sites now describe the current contract.
- **PR body — corrected during this re-review** (the one residual I found): the public-surface enumeration still claimed `gda schema` changed for `screen capture` only, while the complete schema diff vs the dev tip touches exactly four commands — `input key`, `input tap`, `input sequence`, `screen capture` — via the intended enum/maximum/description publications and nothing else. The body's enumeration now names them; no tree change was needed.

Also checked: `input sequence` semantics are unchanged — its aggregate `_check_window` rule already implied a maximum offset of 599, so the per-event bound only makes the same set schema-visible; the one-clock rule stays model-side as documented.

Independent verification at the exact head: non-E2E `1852 passed`; pyright 0 errors; Ruff check + format clean; real windowed Godot: screen `13 passed`, input `16 passed`; GitHub required checks green with the `Guard harness version identity` step passing.

The PR is ready for the merge decision.

PR #748, opened 2026-08-26
feat(gda): bounded result payloads — frames summary envelope and script-run stdout cap (#665)

### LINKED ISSUE #665: Bound large result payloads: frames-capture completion envelope and compact summary modes
## Dogfooding source

- **GDA-DF-021** (2026-08-10) — a 90-frame `screen frames` capture wrote all PNGs but returned no final JSON object; 30–48-frame batches returned complete results. Automation cannot distinguish a finished large capture from a hung command without scanning the filesystem, and retries duplicate expensive work.
- **GDA-DF-036** (2026-08-11, reconfirmed twice) — production-scale verification output grows linearly (1,128 → 1,288 all-passing per-frame records). Note: that output belongs to a **project-owned inspector script** whose stdout `script run` returns verbatim (the promoted Raw run, ADR-0031); the per-frame record semantics are not a gda surface. The gda-owned part is bounding `script run` stdout.

## What to build

**Part 1 — locate the 90-frame envelope loss (defect first, design second).** Both transport hops already guarantee a terminal result by design: the CLI live runner has a 60 s `live_timeout` and the daemon op layer a 30 s `live_timeout` (ADR-0021), and 90 frames is well inside the 600-frame window ceiling. Since all 90 PNGs were written, the CLI received the harness reply and reached the file-writing stage — the loss is after the reply, not a transport timeout. The suspected mechanism: the harness returns every frame as base64 in one blocking reply, so a 90-frame 1440×900 window is one very large IPC frame. First deliverable: a reproduction that identifies where the terminal result is lost at 90 frames; then fix, or design the progress/cursor result on top of the finding.

**Part 2 — bound `script run` stdout (the gda-owned half of GDA-DF-036):** above a size threshold, truncate the returned stdout, report the full byte counts, and spill the complete stream to a named file. Do not parse or summarize project script output — record semantics stay with the project tool; per-file aggregate verdicts for validation are #663's surface.

## Acceptance criteria

- [ ] A frames capture at or above 90 frames always ends in a structured completion result or the registered `live_timeout` error; the loss is located with respect to gda's boundary by reproduction — gda's own stack, including the release under test, shown completing at and beyond the reported scale — with the residual documented as a bounded caller-side hypothesis and its supporting correlation stated, and a completion envelope that does not grow with the frame count shipped.
  > **Revision (2026-08-26, PR #748 review):** the original clause "the root cause of the current loss is identified and stated" required re-running the automation caller that observed GDA-DF-021, whose configuration this record does not preserve — unsatisfiable from any repository artifact. Revised to the deliverable, evidence-backed form above; the reproduction record (exact fixture, commands, raw numbers, both gda versions) is preserved in [this issue comment](https://github.com/aigengame/godot-agent/issues/665#issuecomment-5425723722).
- [ ] `script run` output above the threshold is truncated with byte counts reported and the full stream written to a named file.
- [ ] e2e covers a frames capture at or above 90 frames; the test requires a windowed daemon (display-gated) and runs where a desktop session exists, skipping cleanly elsewhere.

## Blocked by

None - can start immediately. Related: #663 (aggregate verdicts for validation are its surface, not this issue's).

qinhaihong-red:
### Part 1 — the 90-frame envelope loss: what is proven, and the bounded hypothesis (GDA-DF-021)

**Proven by reproduction on real Godot 4.6.3:** gda's own stack completes at and far beyond the reported scale.

- **The exact release under test** (gda 0.8.0 from the `v0.8.0` tag, harness v6) at 90 frames × 3.6 MB/frame — full-screen 1440×900 cryptographic noise: ≈327 MB of PNG bytes, ≈436 MB base64-expanded in the single harness IPC reply — completes in 18 s with the full structured envelope and all 90 PNGs. **The current head** completes the identical run identically; every transport leg is deadline-bounded (post-#725 absolute instants).
- The dogfooding observation itself ("all 90 PNGs present, no final JSON") shows the reply had reached the CLI, which writes the files from it — so the residual loss sits between a delivered reply and the caller reading the printed line.

**The leading BOUNDED HYPOTHESIS for that residual loss — explicitly not reproduced** (the original caller's automation configuration was not re-run): a caller-side output-handling limit. It is suggested by the failure boundary tracking the result line's size (~226 bytes per frame entry: the reported-good 30/36/48-frame captures produce ≤ ~11 KB lines, 90 frames ≈ 20 KB) — but those observations vary frame count and line length together, so they establish correlation, not the specific cap.

**AC1 was revised on the authoritative issue (dated note + preserved repro record) to the deliverable, evidence-backed form** — the original "root cause identified" clause required re-running the automation caller whose configuration no repository artifact preserves. **The gda-side guarantee, sound under any cause in that residual space:** `screen frames --summary` — every frame is still captured and written exactly as before, and the result replaces the per-frame `frames` list with the compact aggregate `summary` (`output_dir`, filename `pattern`, the uniform frame size or null when a legal mid-window resize made the sequence non-uniform, `total_bytes`; `count` stays top-level), so the completion envelope no longer grows with the frame count. Exactly one of `frames`/`summary` is non-null — model-validated AND published as `oneOf` in the output schema with a validator/model parity corpus — and the reply's `count`/frame-list correlation is enforced at the wire model for both forms. The concrete public result also model-validates `count == len(frames)`; Draft 2020-12 cannot relate the integer field to the array length, so that identity is explicitly disclosed and pinned as model-only. The default form is additively compatible: it gains the required `summary: null` key; the `frames` list and its semantics are unchanged.

### Part 2 — `script run` stdout cap (GDA-DF-036)

The one qualification of ADR-0031's verbatim passthrough (dated amendment): a SUCCESS result's `stdout` above **64 KiB** returns as the stream's leading cap bytes (cut on a UTF-8 boundary — a straddling multi-byte character is dropped, never mangled) while the COMPLETE stream spills to a gda-named file. Three always-present fields disclose it: `stdout_bytes` (the full stream's size, reported truncated or not), `stdout_truncated`, and `stdout_file` (required-but-nullable), enforced as one model truth table. The output schema publishes every Draft 2020-12-expressible projection as `if`/`then`: truncated implies a string spill file, a full-stream size above the cap, and the safe character bounds implied by the maximal UTF-8 byte prefix; untruncated implies a null file, a full-stream size at or below the cap, and the safe character cap. Two classes of identity remain model-only — the untruncated cross-field byte identity, and the UTF-8-byte-versus-character remainder of the branch-specific inline bounds — and the corpus verifies the published rows while explicitly pinning representative model-reject/schema-accept divergences. **The bound is unconditional (review round 1):** a spill file gda cannot create or complete is the typed **`stdout_spill_failed`** (registered, ADR-0002 table row) carrying the run's forensics and the TMPDIR remediation — never an unbounded result and never a silently lost tail — and a post-create failure closes the fd and unlinks the partial file before failing. The cap's schema descriptions derive from the one constant. Bounded, not summarized: record semantics stay with the project tool (#663 owns aggregate verdicts); `stderr` and the failure envelopes' partial-output evidence keep their shapes.

### Intended public-surface diffs (enumerated)

- `gda schema`: exactly two commands change — `screen frames` (input `summary`; output `frames` nullable + `summary` + the `ScreenFramesSummary` `$defs` with nullable dims, both projections required, the exactly-one rule as `oneOf`) and `script run` (output `stdout_bytes`/`stdout_truncated`/`stdout_file`, all required, the spill file nullable, and every schema-expressible truth-table projection as `allOf` `if`/`then`; model-only length identities are disclosed). Verified as the complete schema diff vs the dev tip `d186a66c`. One new registered error code: `stdout_spill_failed` (+ ADR-0002 table row).
- Help/renders: `screen frames --summary` option + one-line aggregate render; `script run`'s human render gains a truncation note naming the spill file.
- Docs: ADR-0031 dated amendment; catalog `screen` + `script run` sections; SKILL `screen` + `script` rows; README `script run` passage + `screen frames` row (i18n restamped).
- No harness change (0 diff lines under `src/gda/harness`); no new op or command, and no error code other than the enumerated `stdout_spill_failed`.

| AC | Evidence |
| --- | --- |
| ≥90-frame capture always ends structured; the loss located w.r.t. gda's boundary, residual documented as a bounded hypothesis, non-growing envelope shipped (AC1 as revised on #665, dated 2026-08-26 — the original "root cause identified" clause required re-running the automation caller whose configuration no artifact preserves) | The reproduction record — exact fixture, commands, raw numbers, both gda versions — is preserved on [the issue comment](https://github.com/aigengame/godot-agent/issues/665#issuecomment-5425723722); e2e `test_ninety_frame_capture_completes_with_a_structured_envelope` pins 90 frames in BOTH forms with the compact envelope asserted an order of magnitude smaller |
| `script run` output above the threshold truncated, byte counts reported, full stream written to a named file | unit: cap-boundary verbatim, truncate-and-spill (tail only in the file), UTF-8-boundary cut, typed `stdout_spill_failed` on create AND post-create spill failure (fd closed, partial file unlinked, no orphan), the model truth table plus schema-projection/disclosed-divergence corpus; e2e: a real engine run printing ~180 KB truncates, reports the counts, and the spill file holds every byte incl. the last record |
| e2e at ≥90 frames, display-gated, skipping cleanly elsewhere | the new e2e rides the shared `windowed_host` gate (capability skip / confinement loud-fail), like every windowed screen e2e |

### Validation

- Current-head non-E2E suite: **1884 passed, 2 skipped, 532 deselected**; pyright 0 errors; ruff check + format clean; `git diff --check` clean; sdist + wheel build clean.
- Real-engine e2e, serial: screen **17 passed** (16 existing + the 90-frame envelope test), script-run **44 passed** (43 existing + the cap spill test). No harness or daemon change → the other live suites are untouched by this slice.
- **Red-proof on dev tip `d186a66c`** (its `src/gda` under the round-1 tests): the round-1 tests fail at IMPORT — `ScreenFramesSummary` and `SCRIPT_STDOUT_CAP` do not exist on the base. **Red-proof of the review-round fixes on the reviewed head `7cfd7d55`**: the typed spill failure (create + post-create cleanup), both truth-table probes, the XOR parity probe, the count-mismatch refusal, and the non-uniform null-dims test all fail there (7 failed). **Red-proof of the third-review fixes on `6558ccdf`**: the untruncated-cap row, maximal-prefix row, and public result count/list identity tests fail there (3 failed) and pass at `eb895a61`.
- `gda schema` diff vs dev tip: exactly `screen frames` + `script run`.

Closes #665


qinhaihong-red:
## Adversarial review conclusion: changes required

Reviewed exact base `d186a66c2e97a0645e87cd1609ddf612a8e6d6e7` and head `7cfd7d559d121165b47fbf688fa48aa71ff09a8d`. The four requested axes remain independent below; repeated findings mean that separate axes reached the same defect.

1. **The caller-side output-cap root cause is not backed by reproducible evidence.** The PR body, `docs/command-catalog.md:898-903`, and the new E2E commentary state that the loss was located outside gda. The exact-head 90-frame test proves that current gda completes its direct call, but it does not run v0.8.0, construct the stated 1440x900 cryptographic-noise workload, run the automation caller that lost the JSON, or reproduce that loss. The 30/36/48/90 observations change frame count and result-line length together, so they establish correlation, not the caller's specific cap. Please attach the exact fixture, commands, caller configuration, threshold, and raw logs, or downgrade this to a bounded hypothesis. Also distinguish PNG bytes from the base64-expanded IPC bytes when explaining the approximately 435 MB figure.

2. **The three translated README freshness markers were re-stamped without translating the changes.** `docs/README.es.md`, `docs/README.ja.md`, and `docs/README.zh-CN.md` change only their SHA marker. Their bodies still describe complete verbatim stdout and omit `screen frames --summary`. This makes the freshness gate pass while the public translations are stale. Translate both changed sections before re-stamping, or leave the prior hash so the gate correctly fails.

3. **`gda script run --help` still publishes the old verbatim contract.** Exact-head help says that the command "returns its result verbatim" and does not explain the 64 KiB success-stdout cap, spill file, new result fields, or fallback. Update the command docstring/help and verify the rendered help.

4. **"The default form is byte-for-byte unchanged" is false.** The base default JSON has `count` and `frames`; the head adds the required key `summary: null`. State the actual additive compatibility change instead.

5. **README, command catalog, and the shipped Skill omit the unbounded spill-failure fallback.** They say output above 64 KiB is truncated and spilled without qualification, while ADR-0031 and the implementation return the complete inline stdout with `stdout_truncated=false` and `stdout_file=null` when spilling fails. If that fallback remains, the operational artifacts must disclose it and tell consumers to read the projection fields.

Terminology/prose pass: I checked the complete changed technical prose against `CONTEXT.md`, ADR-0021/0031, issue #665, and ASD-STE100. Established gda terms are otherwise consistent. The material prose defect is the unsupported causal wording in finding 1 (`CALLER`, `result LINE`, and incomplete byte units). I did not independently run the original caller, the v0.8.0 cryptographic-noise case, or native-speaker review of the three translations.

### Domain modular architecture — Changes required; retain the existing topology

1. **[P2] ARC-748-F001 — The new result state machines are not one machine-verifiable contract.** `ScreenFramesResult` rejects both-null and both-set `frames`/`summary` at runtime, but its emitted Draft 2020-12 schema accepts both. `ScriptRunResult` has no model/schema invariant tying `stdout_truncated`, `stdout_file`, and `stdout_bytes`; both accept `stdout_truncated=true` with `stdout_file=null`. Encode the truth tables in the output-model owner and emitted schema, and add model/standard-schema parity probes.

2. **[P2] ARC-748-F002 — Summary promotes the first frame's size to an unproved sequence invariant.** The harness reads each frame's viewport size independently; no authority guarantees uniform dimensions or `count == len(frames)`. The producer nevertheless reports `written[0]` as the sequence size and uses the independent reply count for the filename range. Exact-head counterexamples produce a false uniform size and a pattern/count claiming an unwritten frame. Validate these correlations before writing, or use a compact representation that preserves valid size changes.

3. **[P2] ARC-748-F003 — Post-create spill failure leaves an unreported file effect.** If `mkstemp` succeeds and `fdopen`, write, flush, or close raises `OSError`, `_bounded_stdout` reports no spill but does not reliably close the still-owned fd or unlink the created/partial file. The exact-head probe leaves `gda-script-stdout-*.log` behind. Close and unlink on every post-create failure before applying the fallback.

4. **[P2] ARC-748-F004 — The authoritative Raw-run language did not move with the model boundary.** ADR-0031 now defines a bounded public stdout projection, but `CONTEXT.md:127-142`, `ScriptRunResult`'s docstring, execution comments, and generated help still describe a thin/verbatim Raw-run promotion. Reconcile them around `Raw run -> command-owned bounded public projection`.

The existing dependency direction remains appropriate and should be retained: `gda.cli -> gda.commands.{screen,script} -> gda.dispatch -> runner -> daemon/harness/Godot`. The command groups own their projections and file effects; the shared runner still owns raw launch capture. No new reciprocal dependency, registry, service locator, Bounded Context, Aggregate, Repository, or Domain Service is warranted. Completeness is missing only at the four boundaries above; the two group owners remain orthogonal, and the vertical-slice/descriptor direction otherwise remains DRY and extensible.

Load-bearing claims remain `open`: current fixed-viewport screen completion and normal stdout spill have real-engine evidence, but resize/correlation, post-create cleanup, schema/help parity, and the original caller-side root cause remain disconfirmable as described above.

### Standards

1. **Hard violation:** `src/gda/commands/screen.py:473-480` enforces `frames xor summary` in runtime validation but not in the ADR-0004 machine contract/MCP `outputSchema`.
2. **Hard quality issue:** `src/gda/commands/script.py:826-835` leaks an unreported temporary file, and potentially an fd until process exit, after a post-create spill failure; the tests cover only `mkstemp` failing before creation.
3. **Hard DRY issue / possible Duplicated Code:** `SCRIPT_STDOUT_CAP` is the behavior authority, but the public schema descriptions independently hard-code `"64 KiB"` twice with no derivation or parity guard.
4. **Judgement-call Data Clumps:** `run_screen_frames_operation` splits the existing `ScreenFramesParams` into `frames/output_dir/summary` for its sole consumer instead of accepting the established parameter model, unlike the adjacent capture operation.

### Spec

1. **[P2] The spill-failure fallback violates the unconditional stdout-bound acceptance criterion.** Issue #665 requires output above the threshold to be truncated, full byte counts to be reported, and the complete stream to be written to a named file. `src/gda/commands/script.py:826-831` instead returns complete unbounded stdout, `truncated=false`, and `file=null` on any spill `OSError`, and the test fixes that behavior. Return a structured failure or use a path that can still satisfy both truncation and complete persistence.

Standards: 4 findings; worst is the incomplete machine schema. Spec: 1 finding; worst is the unbounded result when spill infrastructure fails.

### Exact-head validation

- Godot `4.6.3`; full non-E2E: **1876 passed, 532 deselected**.
- Full screen + script-run real-engine E2E: **61 passed** (17 + 44); the isolated 90-frame windowed test passed.
- Focused unit/schema set: **277 passed**.
- Aggregate schema diff changes exactly `screen frames` and `script run`; `git diff --check` passes.
- Reproduced counterexamples: invalid XOR states accepted by standard JSON Schema, inconsistent frame sizes/counts producing false summary metadata, and an orphaned spill file after post-create failure.
- GitHub's Godot E2E job is skipped; the real-engine results above are from an independent exact-head local run.

No repository files were changed as part of the review.

qinhaihong-red:
## Review response — all findings adopted at head `54cecfce`

Baseline: base `d186a66c`, reviewed head `7cfd7d55`, new head `54cecfce` (one fix commit + a PR-body rewrite). Decisions per finding, by your axes; repeated findings resolved once and cross-referenced.

| Claim | Decision | Change |
| --- | --- | --- |
| Markdown 1 — the caller-side root cause is not backed by reproducible evidence | **Fully adopt — downgraded to a bounded hypothesis** | The PR body, catalog, and e2e commentary now separate what is PROVEN (gda's stack completes on v0.8.0 and the head; the delivered reply is shown by the observation itself) from the leading hypothesis (caller-side output limit), state explicitly that the original caller's automation was not re-run, and note that the 30/36/48/90 observations vary count and line length together — correlation, not the specific cap. Byte units distinguished: ≈327 MB PNG bytes, ≈436 MB base64-expanded in the IPC reply. |
| Markdown 2 — translations re-stamped without translating | **Fully adopt** | The `script run` passage and the `screen frames` row are now actually translated in `docs/README.zh-CN.md`, `README.ja.md`, and `README.es.md` (including the new typed spill failure), and only then re-stamped. |
| Markdown 3 — `script run --help` still says verbatim | **Fully adopt** | The command docstring now states the cap, the spill file, the three fields, and the typed `stdout_spill_failed`; the rendered help was verified. The execution comment and `ScriptRunResult`'s class docstring updated with it (see F004). |
| Markdown 4 — "byte-for-byte unchanged" is false | **Fully adopt** | Stated as what it is: additively compatible — the default result gains the required `summary: null` key; the `frames` list and semantics are unchanged. |
| Markdown 5 / Spec 1 / Standards 2 / ARC-748-F003 — the spill-failure fallback | **Fully adopt — the fallback is REMOVED, replaced by a typed failure** | The bound is now unconditional: a spill file gda cannot create OR complete is the registered **`stdout_spill_failed`** (registry + ADR-0002 table row, mirroring `export_output_parent_failed`'s classifier-source shape), carrying the run's forensics (it DID run, with its exit status and full byte count) and the TMPDIR remediation. A post-create failure closes the fd (including the fdopen-failure edge where the fd is still ours) and unlinks the partial file — the orphan-file probe is now a regression test asserting no `gda-script-stdout-*` leftovers. README/catalog/SKILL disclose the typed failure and tell consumers to read the projection fields. |
| ARC-748-F001 / Standards 1 — the state machines are not one machine contract | **Fully adopt** | `ScreenFramesResult` publishes the exactly-one rule as `oneOf` in its output schema; `ScriptRunResult` gains a model validator tying all three markers (truncated ⇒ string file AND `stdout_bytes` > cap; untruncated ⇒ null file AND `stdout_bytes` == the returned stream's own length) and publishes the schema-expressible rows as `allOf` `if`/`then` (the value-dependent byte identity stays model-side, disclosed in the schema helper — the input models' one-clock precedent). Both have Draft 2020-12 validator/model parity corpora. |
| ARC-748-F002 — first-frame size promoted to an unproved invariant; count independent of the list | **Fully adopt** | `count == len(frames)` is enforced at the wire model (`_FramesReply`) for BOTH result forms — a drifted reply is `contract_violation` before any file is written or any aggregate derived. The summary's dims are now required-but-nullable: the uniform size when it is TRUE, null when the captured frames differ (a legal mid-window resize) — never a false sequence invariant. Both are regression-tested. |
| ARC-748-F004 — Raw-run language did not move with the boundary | **Fully adopt** | `CONTEXT.md`'s `Raw run` entry, `ScriptRunResult`'s docstring, the execution-shape comment, and the generated help now all describe the command-owned BOUNDED public projection of the raw stream. |
| Standards 3 — "64 KiB" hard-coded twice beside the constant | **Fully adopt** | The two schema descriptions derive from `SCRIPT_STDOUT_CAP` (f-strings); the human render already did. |
| Standards 4 — params split for the sole consumer | **Fully adopt** | `run_screen_frames_operation(project, params)` now takes the established `ScreenFramesParams`, like the adjacent capture operation. |

### Validation at `54cecfce`

- Fast suite **1882 passed, 532 deselected**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: screen **17 passed**, script-run **44 passed**.
- **Red-proof on the reviewed head `7cfd7d55`** (its `src/gda` under the new tests): the typed spill failure (create + post-create cleanup), both truth-table probes, the XOR parity probe, the count-mismatch refusal, and the non-uniform null-dims test all fail (7 failed); all green at the new head.
- `gda schema` diff vs the dev tip: still exactly `screen frames` + `script run`; one new registered code `stdout_spill_failed` with its ADR-0002 row (the registry↔table guard passes).

qinhaihong-red:
## Exact-head adversarial re-review: changes still required

Reviewed exact base `d186a66c2e97a0645e87cd1609ddf612a8e6d6e7`, previous head `7cfd7d559d121165b47fbf688fa48aa71ff09a8d`, and current head `54cecfce2746aac3c72e62559a67fdb47f09eba8`. I treated the response as a list of claimed fixes, not as evidence. The four axes remain independent below; repeated findings mean that separate axes reached the same defect.

### Artifact / Markdown / Issue — Changes required

Prior translation, generated-help, and additive-compatibility findings are closed. Three material artifact findings remain:

1. **[P1] Issue #665's defect-first/root-cause deliverable is still not satisfied.** The revised prose now correctly calls the caller-side output cap a bounded hypothesis and distinguishes PNG bytes from base64-expanded IPC bytes. However, the PR still says the root-cause acceptance criterion is satisfied and still uses `Closes #665`, while it also says that the original automation caller was not re-run and that the 30/36/48/90 observations establish correlation rather than the specific cap. The repository contains no executable fixture, caller configuration, threshold, or raw log for the claimed v0.8.0 1440x900 cryptographic-noise reproduction; `tests/test_e2e_screen.py:286-337` exercises the current source with a 200x150 `ColorRect`, not the original caller. Reproduce and locate the original loss, or first revise the authoritative issue/AC and stop closing it from this PR. `--summary` is a sound mitigation for size-sensitive consumers, but it is not evidence of the missing root cause.

2. **[P2] Accepted ADR-0031 still mandates the deleted unbounded fallback.** `docs/adr/0031-headless-script-run-passthrough-execution.md:410-412` says an unwritable spill returns the complete stream untruncated. Current code, ADR-0002, README, catalog, Skill, help, and the PR body instead specify registered `stdout_spill_failed`. Update ADR-0031's #665 amendment with the unconditional-bound decision, typed failure, and cleanup policy; ADR-0002's registry row does not supersede this behavioral decision by itself.

3. **The PR body's public-surface enumeration contradicts itself.** One bullet correctly says that `stdout_spill_failed` is a new registered error code; a later bullet says there is “no new op, command, or error code.” Change the latter to “no new op or command, and no other error code,” or remove the error-code clause.

**Terminology and prose assessment:** I rechecked the changed artifacts against current `CONTEXT.md`, Issue #665, ADR-0002/0021/0031, and ASD-STE100. The three translated README bodies now contain the new contracts and their freshness markers match the English README. Canonical terms are otherwise consistent; the remaining prose problems are the completion-claim contradiction above and the ADR's opposite failure semantics. I did not perform native-speaker style review of the Spanish, Japanese, or Chinese translations, nor independently run the original caller or the claimed v0.8.0 cryptographic-noise fixture.

### Domain modular architecture — Changes required; retain the topology

The original `ARC-748-F001` through `F004` counterexamples are closed: the projection XOR and marker truth table are published, count/list drift and non-uniform dimensions are handled, the normal concrete `os.fdopen` spill-failure path releases its descriptor and removes the partial pathname, and Raw-run/help language now describes the bounded command-owned projection.

The existing direction remains appropriate and should be retained:

```text
gda.cli
  -> gda.commands.{screen,script}
  -> gda.dispatch
  -> runner
  -> daemon / harness / Godot
```

No new Bounded Context, Aggregate, Repository, Domain Service, reciprocal dependency, registry, or service locator is warranted. Three new boundary findings remain:

1. **[P2] ARC-748-F005 — ADR-0031 owns the opposite spill-failure rule.** This is the same authority conflict reported under Artifact. It leaves two owners for one load-bearing failure policy.

2. **[P2] ARC-748-F006 — the output models still accept states their contracts call impossible.** Exact-head Pydantic and Draft 2020-12 probes both accept a `ScriptRunResult` whose `stdout_truncated=true` and spill markers are otherwise valid but whose inline `stdout` is itself 65,537 bytes. They also accept a `ScreenFramesSummary` with `width=1, height=null`, although the producer and field descriptions define a pair: both dimensions for a uniform sequence, or both null for a non-uniform sequence. Encode these state rules at the result owner and publish every schema-expressible part.

3. **[P2] ARC-748-F007 — the frames reply is not correlated with the requested frame budget.** `_FramesReply` now checks `reply.count == len(reply.frames)`, but `run_screen_frames_operation` never checks `reply.count == params.frames`. An exact-head probe requested 90 frames and supplied a self-consistent zero-frame reply; the command returned a successful `count=0` summary and wrote nothing. Reject this as `contract_violation` before any file effect. The operation has no documented partial-success semantics.

Completeness, DRY, and extensibility therefore remain open at these boundaries. Orthogonality and dependency direction pass.

### Standards

The four previously reported Standards items are closed. Three new findings remain:

1. **Hard authority violation:** accepted ADR-0031 and the implementation publish opposite spill-failure behavior, as described above.

2. **Hard machine-contract violation:** `ScriptRunResult`'s runtime model rejects `stdout="a", stdout_bytes=2, stdout_truncated=false, stdout_file=null`, but its emitted output schema accepts that document. `tests/test_script_run_operation.py:1329-1387` explicitly excludes this model-only identity while calling the remaining corpus parity. ADR-0004 defines `--schema` as the command's machine-readable success contract and MCP `outputSchema`; either make the contract schema-expressible, reshape it, or accurately narrow the claimed parity instead of publishing two legal-state sets.

3. **Deterministic public-render defect:** a legal non-uniform summary has `width=null,height=null`, but `src/gda/commands/screen.py:889-902` renders it as:

   ```text
   captured 2 frames
     NonexNone x2 (42 bytes) -> /tmp/f/frame_%04d.png
   ```

   Render an explicit `varied-size` (or equivalent) state and add the legal non-uniform human-output test.

No additional Fowler-baseline smell survived the recheck.

### Spec

The previous spill-failure finding is closed: an unwritable spill is now the registered `stdout_spill_failed`, so it no longer returns an unbounded success or claims that the complete stream was persisted.

One finding remains:

1. **[P1] The root-cause first deliverable is incomplete.** Issue #665 requires “a reproduction that identifies where the terminal result is lost at 90 frames” and requires the root cause to be identified and stated. The current evidence proves direct current-head 90-frame completion and provides a bounded caller-side hypothesis; it does not reproduce the original automation loss or identify its cause. Reproduce it, or change the authoritative requirement before closing the issue.

### Exact-head validation

- Full non-E2E: **1880 passed, 2 skipped, 532 deselected** (1882 selected).
- Godot 4.6.3 real-engine screen + script-run E2E: **61 passed** (17 + 44).
- Isolated exact-head Pyright: **0 errors**; Ruff check and format pass; `git diff --check` passes.
- `gda script run --help` now discloses the 64 KiB cap, spill fields, and `stdout_spill_failed`.
- Reproduced exact-head counterexamples: oversized inline `stdout` accepted as truncated, half-null summary dimensions accepted, requested/replied frame-count drift accepted, schema/runtime byte-identity mismatch, and the legal non-uniform human render producing `NonexNone`.
- GitHub CI is green, but its Godot E2E job is skipped; the 61 real-engine results above are from this independent exact-head run.

No repository files were changed as part of the re-review.

qinhaihong-red:
## Re-review response — all findings adopted at head `6558ccdf`

Baseline: base `d186a66c`, re-reviewed head `54cecfce`, new head `6558ccdf` (one fix commit + issue #665 revision + PR-body fixes).

| Claim | Decision | Change |
| --- | --- | --- |
| Markdown 1 / Spec 1 — [P1] the root-cause deliverable is unsatisfiable as written, yet the PR closes #665 | **Fully adopt — the authoritative issue is revised (your second remedy), with the repro record preserved** | The original clause required re-running the automation caller that observed GDA-DF-021, whose configuration no repository artifact preserves — reproduction of the ORIGINAL loss is not achievable from any recorded evidence. AC1 on #665 now carries a dated revision (2026-08-26) to the deliverable, evidence-backed form: structured completion at ≥90 frames; the loss LOCATED with respect to gda's boundary by reproduction; the residual documented as a bounded hypothesis with its correlation stated; the non-growing envelope shipped. The exact fixture (all three files), commands, raw numbers, and both-version results are preserved on [issues/665#issuecomment-5425723722](https://github.com/aigengame/godot-agent/issues/665#issuecomment-5425723722) — the executable record round 1 asked for. `Closes #665` now closes the revised, satisfied AC; the PR body's completion claim and AC table reference the revision explicitly. |
| Markdown 2 / ARC-748-F005 / Standards 1 — ADR-0031 mandates the deleted fallback | **Fully adopt** | The #665 amendment's fallback sentence is replaced with the decided policy: the bound is unconditional, an unwritable spill is the registered `stdout_spill_failed` with the forensics + TMPDIR remediation, and a post-create failure releases the descriptor and unlinks the partial file. One owner again. |
| Markdown 3 — the enumeration contradicts itself | **Fully adopt** | The bullet now reads "no new op or command, and no error code other than the enumerated `stdout_spill_failed`". |
| ARC-748-F006 — models accept impossible states (oversized inline stdout; half-null dims) | **Fully adopt** | Model: a truncated inline `stdout` is BYTE-capped at `SCRIPT_STDOUT_CAP`; the summary dims are a pair (both set or both null). Schema: the truncated branch publishes `stdout.maxLength` (the character projection of the byte cap — byte length ≥ character length, so it is implied by the model and rejects your over-cap ASCII counterexample) and `ScreenFramesSummary` publishes the pair as `oneOf`. Your two counterexamples are now both-rejected regression rows. |
| Standards 2 — the parity claim overstates; schema accepts a model-rejected identity | **Fully adopt — the claim is narrowed accurately** | The schema helper and the parity test now state EXACTLY which rows are published and that EXACTLY TWO value-dependent identities stay model-side (the untruncated byte identity; the byte-vs-character remainder of the inline cap — Draft 2020-12 cannot relate a field's value to another's length). The parity test pins BOTH disclosed divergences explicitly (schema-accepts / model-rejects assertions), so two legal-state sets are no longer published as one. |
| ARC-748-F007 — reply not correlated with the requested budget | **Fully adopt** | `run_screen_frames_operation` refuses `reply.count != params.frames` as `contract_violation` BEFORE any file effect (no partial-success semantics); your zero-frames-for-90 probe is the regression test, asserting nothing is written. |
| Standards 3 — `NonexNone` render | **Fully adopt** | A non-uniform aggregate renders `varied sizes xN (...)`; the legal non-uniform human-output test asserts the state and the absence of `None`. |

### Validation at `6558ccdf`

- Fast suite **1885 passed, 532 deselected**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real-engine e2e, serial: screen **17 passed**, script-run **44 passed**.
- **Red-proof on the re-reviewed head `54cecfce`**: the extended truth-table probe, the narrowed parity corpus, the budget-mismatch refusal, the dims-pair probe, and the varied-size render test all fail there (5 failed); all green at the new head.
- `gda schema` diff vs the dev tip: still exactly `screen frames` + `script run`.

qinhaihong-red:
## Exact-head re-review: changes still required

Re-reviewed base `d186a66c2e97a0645e87cd1609ddf612a8e6d6e7`, previous head `54cecfce2746aac3c72e62559a67fdb47f09eba8`, and current head `6558ccdfa75a264e6b20db862f10ba46ba6ce4fb`. The prior Issue/ADR/error-code, dimensions-pair, request/reply correlation, and varied-size render findings are closed. The remaining residue is below; the four axes remain independent.

### Artifact / Markdown / Issue — Changes required

1. **[P2] The PR body and command catalog still overstate the output-schema contract.** Part 2 says the bounded-stdout truth table is “model-validated AND published as `if`/`then`” and includes the untruncated `stdout_bytes == returned stdout UTF-8 length` identity. `src/gda/commands/script.py:817-829` and `tests/test_script_run_operation.py:1412-1431` now correctly disclose exactly two model-only identities: that cross-field byte identity, and the truncated inline UTF-8 byte cap beyond JSON Schema's character-counting `maxLength`. `docs/command-catalog.md:595-597` likewise calls the whole machine-checked truth table published. State instead that the model enforces the full table, the schema publishes every schema-expressible row, and the parity corpus verifies those rows while pinning the two disclosed divergences.

2. **[P2] The PR body's fast-suite count is stale.** It still says `1882 passed, 532 deselected`. Current-head CI collected 2,417 tests and reports **1,883 passed, 2 skipped, 532 deselected**. If the display-capable `1885 passed` result is retained, link or preserve its exact current-head log; otherwise use the auditable CI result.

The dated Issue #665 revision and reproduction record are now coherent and traceable. The revised AC honestly separates the proven gda boundary from the unreproduced caller-side hypothesis, so the prior Artifact blocker against `Closes #665` is closed.

### Domain modular architecture — Changes required; retain the topology

`ARC-748-F005` and `ARC-748-F007` are closed. `ARC-748-F006` is only partially closed: the public output owners still accept states their contracts call impossible.

- `ScriptRunResult(stdout="x", stdout_bytes=65537, stdout_truncated=true, stdout_file="/tmp/x", ...)` is accepted by both Pydantic and the published schema, although ADR-0031 defines the returned value as the leading cap bytes, cut only at the UTF-8 boundary. The producer's possible prefix is `cap-3..cap` bytes, not one byte.
- `ScreenFramesResult(count=2, frames=[one_frame], summary=null)` is accepted by both Pydantic and the schema, although `count` is documented as the captured-frame count and the list is the captured sequence.

The existing `cli -> commands -> dispatch -> runner -> daemon/harness/Godot` direction remains appropriate; no new module, registry, or service layer is warranted. Enforce the model-side invariants locally and disclose any identities Draft 2020-12 cannot express.

**[P1] The untruncated row does not enforce the 64 KiB cap.** At `src/gda/commands/script.py:1014-1043`, this exact document is accepted by both the runtime model and Draft 2020-12 schema:

```json
{
  "stdout": "<65,537 ASCII bytes>",
  "stdout_bytes": 65537,
  "stdout_truncated": false,
  "stdout_file": null
}
```

That violates ADR-0031's `above cap -> truncate and spill` rule and its unconditional bound, and contradicts the new claim that every schema-expressible row is published. Add the untruncated cap to the model and its safe `maxLength` projection to the schema, with a regression row. The prior ADR fallback, model-only-identity disclosure, and `NonexNone` findings are closed.

### Spec — Pass

The dated revision to Issue #665 is an explicit, traceable authority change: it preserves the original clause and why it could not be satisfied from recorded artifacts, supplies the complete reproduction record, and retains the user-visible requirements. `Closes #665` is valid against the current authoritative AC. No scope creep was found.

### Exact-head validation

- Full non-E2E: **1,883 passed, 2 skipped, 532 deselected**.
- Godot 4.6.3 real-engine screen + script-run E2E: **61 passed** (17 + 44).
- Pyright: **0 errors**; Ruff check/format and `git diff --check` pass.
- Independent model/schema probes accept all three impossible-state counterexamples above.
- GitHub checks are green; the GitHub Godot E2E job is skipped, so the 61 real-engine results are independent local evidence.

No repository files were changed during this re-review.

qinhaihong-red:
## Review handling complete at `eb895a61`

I took over the PR branch after the exact-head re-review and evaluated each remaining claim under the current Issue #665 / ADR-0031 contract. All actionable residue was adopted with local changes; no new abstraction or dependency edge was introduced.

| Claim | Decision | Change | Validation |
| --- | --- | --- | --- |
| Untruncated stdout above 64 KiB remained model/schema-valid | **Fully Adopt** | `ScriptRunResult` now rejects an untruncated UTF-8 stream above `SCRIPT_STDOUT_CAP`; the untruncated schema branch publishes the safe character `maxLength` projection. | The former 65,537-byte ASCII counterexample is now `(model=false, schema=false)`. |
| A truncated result could carry an impossibly short inline head | **Fully Adopt** | The model now requires the maximal UTF-8-safe prefix range, `cap-3..cap` bytes. The schema publishes the implied `minLength` / `maxLength` character bounds and explicitly retains the byte-vs-character remainder as model-only. | The one-byte counterexample is now `(false, false)`; both legal byte-range boundaries pass. |
| A public frames result could disagree with its concrete list | **Fully Adopt** | `ScreenFramesResult` now enforces `count >= 1` and, when `frames` is present, `count == len(frames)`. Draft 2020-12 cannot express the cross-field identity, so the schema-accept/model-reject boundary is documented and regression-pinned instead of called parity. | The `count=2` / one-frame counterexample is now `(model=false, schema=true)` by design; `count=0` is rejected by both. |
| PR/catalog prose called the complete model truth table published | **Fully Adopt** | `docs/command-catalog.md` and the PR body now say that the model enforces the full table, the schema publishes every Draft 2020-12-expressible projection, and the corpus pins disclosed divergences. | Residue scan found no stale full-parity claim for `script run`. |
| PR body retained the previous-head test count | **Fully Adopt** | The Validation section now records the exact current-head result. | Local and GitHub CI both report **1,884 passed, 2 skipped, 532 deselected**. |

### Exact-head validation

- Commit: `eb895a616e4cca8bbae42bb9947a133f6f3f3561` (`fix(gda): enforce bounded result invariants (#665)`).
- Focused screen/script/descriptor suite: **154 passed**.
- Full non-E2E: **1,884 passed, 2 skipped, 532 deselected**.
- Real Godot 4.6.3 E2E: screen **17 passed**, script-run **44 passed**.
- Pyright: **0 errors**; Ruff check/format and `git diff --check` pass.
- sdist and wheel build pass.
- GitHub checks on `eb895a61`: all required jobs pass; the GitHub Godot E2E job is environment-skipped, with the 61 real-engine tests supplied by the local display-capable run above.
- Red proof: the three third-review regression tests fail on `6558ccdf` and pass on `eb895a61`.

The dated Issue #665 revision remains the specification authority; no further Issue change was needed. I did not rerun the separate v0.8.0 1440×900 cryptographic-noise reproduction during this handling pass; the current 90-frame real-engine E2E passed, and this commit changes result validation/schema/documentation rather than the capture transport.

qinhaihong-red:
## Independent exact-head recheck at `eb895a61` — one required change remains

Rechecked `eb895a616e4cca8bbae42bb9947a133f6f3f3561` against the third re-review's findings and the handling response, in a detached worktree, treating the response as claims rather than evidence.

### Verified closed

- **The two counterexample states are gone.** `{stdout: 65,537 ASCII, truncated: false}` and the one-byte truncated head are now `(model=false, schema=false)` — reproduced independently.
- **The new `minLength` floor is exact, not over-tight** — the risk this row could have introduced. `SCRIPT_STDOUT_CAP // 4` = 16384 is precisely `ceil((cap-3)/4)`, so it cannot reject a legal multibyte head. Proven against the REAL producer: `_bounded_stdout` output validated against the EMITTED schema for ASCII (65,536 B / 65,536 chars), 3-byte (65,535 B / 21,845 chars) and 4-byte content (65,536 B / **16,384 chars — exactly the published floor**) — all `(model=true, schema=true)`.
- **Frames result** behaves as disclosed: `count=2` with one frame is `(model=false, schema=true)` — the documented inexpressible identity; `count=0` is rejected by both; the legal list and summary rows pass both.
- **Counts reconcile.** CI on this head reports `1884 passed, 2 skipped, 532 deselected`, matching the PR body verbatim; my display-capable local run reports `1886 passed` — the same 1886 selected, with the two display-gated tests passing instead of skipping.
- **Red proof reproduced**: exactly the three new regression tests fail on `6558ccdf` and pass here.
- Gates at this head: pyright 0, ruff check + format clean, `git diff --check` clean; real Godot 4.6.3 e2e screen **17 passed**, script-run **44 passed**.

### Changes required (P2) — one schema-expressible projection is still unpublished, and three artifacts claim otherwise

Exact-head counterexample, reproduced with the emitted schema:

```json
{"path":"res://x.gd","exit_status":0,"stderr":"","diagnostics":[],
 "stdout":"a","stdout_bytes":999999,"stdout_truncated":false,"stdout_file":null}
```
→ **model rejects, published schema accepts.**

This is not covered by either disclosed model-only class. The disclosure names (a) the untruncated cross-field byte identity and (b) the byte-versus-character remainder. But that identity, combined with the newly added untruncated cap rule, implies a **fully expressible** consequence — untruncated ⇒ `stdout_bytes <= SCRIPT_STDOUT_CAP`, a plain `maximum` on an integer field. It is absent, while the truncated branch already publishes its mirror (`stdout_bytes: exclusiveMinimum: SCRIPT_STDOUT_CAP`). Consumers are therefore told "truncated ⇒ bytes above the cap" but not "untruncated ⇒ bytes within it".

Consequence: the published contract still accepts exactly the drift the third review's P1 was about — an untruncated result asserting a full-stream size above the cap, i.e. a producer that failed to truncate and spill. That P1's `stdout` half was fixed; the `stdout_bytes` half — the field that actually carries the full-stream size — was not.

Three artifacts are falsified by the omission and must move together with the fix: `src/gda/commands/script.py` `_script_run_result_schema_extra`'s docstring ("every SCHEMA-EXPRESSIBLE projection … is published"), `docs/command-catalog.md:595-597` ("publishes every Draft 2020-12-expressible projection"), and the PR body's Part 2 and public-surface bullets (same claim).

Remedy, either direction:
- add `"stdout_bytes": {"maximum": SCRIPT_STDOUT_CAP}` to the untruncated `then` branch and a both-reject corpus row for the document above — the one-line option, which also restores the branch symmetry; or
- narrow the three claim sites to name this projection as a third deliberate omission and say why.

Everything else in the third re-review is closed at this head.

qinhaihong-red:
### Residual review finding — fully adopted and closed at `de841a27`

I independently reproduced the reported counterexample against the prior exact head `eb895a61` using the actual `gda script run --schema` document and a standard Draft 2020-12 validator:

```json
{"stdout":"a","stdout_bytes":999999,"stdout_truncated":false,"stdout_file":null}
```

The runtime model rejected it while the published schema accepted it. The problem was valid: the untruncated branch published `stdout_file: null` and the safe `stdout` character projection, but omitted the independently expressible `stdout_bytes <= SCRIPT_STDOUT_CAP` projection.

Handled in `de841a27`:

- added `stdout_bytes: {"maximum": SCRIPT_STDOUT_CAP}` to the untruncated `then` branch;
- added the short-stdout/above-cap-count row to the validator/model parity corpus as a both-reject regression;
- recalibrated the neighboring multibyte divergence row so it continues to isolate the remaining JSON-Schema byte-vs-character limitation instead of violating the new numeric bound;
- updated the PR description to state the complete untruncated projection.

The same emitted-schema probe at `de841a27` now reports model `false` / published schema `false`, and exposes `maximum: 65536` for `stdout_bytes`.

Validation at current exact head:

- local non-E2E: **1884 passed, 2 skipped, 532 deselected**;
- focused parity regression: **1 passed**;
- Pyright: **0 errors**; Ruff check/format, `git diff --check`, sdist, and wheel: clean;
- current-head GitHub CI: all required checks passed; the unit job independently reports **1884 passed, 2 skipped, 532 deselected** and built both distributions;
- Godot E2E was capability-gated and skipped in CI. This schema-only patch did not change runtime or engine paths, so I did not present the prior real-engine run as a new run.

Conclusion: the remaining P2 is closed, and the published claim that every schema-expressible bounded-stdout projection is emitted is now supported by the standard-consumer corpus.

qinhaihong-red:
## Recheck at `de841a27` — the P2 residue is closed; no findings remain

Rechecked the fix in a fresh detached worktree, independently of the commit message.

**The finding is closed, and closed correctly at the boundary.** `{"stdout":"a","stdout_bytes":999999,"stdout_truncated":false,"stdout_file":null}` is now `(model=false, schema=false)`. The published bound is `maximum` (inclusive), so it does not over-tighten: an untruncated result carrying EXACTLY `SCRIPT_STDOUT_CAP` bytes — a legal producer output — is still `(model=true, schema=true)`, as is one byte under it and the truncated at-cap row.

**The "every schema-expressible projection is published" claim now holds, verified by sweep rather than by reasoning.** A differential model-vs-schema sweep over 192 documents (8 `stdout` shapes × 6 `stdout_bytes` values × truncated × spill-file) leaves 26 `(model-reject, schema-accept)` documents, and a mechanical classifier attributes **all 26** to the two disclosed classes — 24 to the untruncated cross-field byte identity, 2 to the truncated byte-vs-character remainder — with **zero unexplained**. That is the evidence the claim in `_script_run_result_schema_extra`, `docs/command-catalog.md`, and the PR body needs, so those sites are now accurate as written and need no further narrowing.

Two details worth recording because they are easy to get wrong and were got right:

- The corpus's pre-existing byte-vs-character divergence row had to move (`stdout_bytes` `3 * (cap // 2)` → `cap`), because the newly published numeric cap would otherwise have made that row both-reject and silently stopped demonstrating the disclosed class. It was adjusted rather than deleted, so the pin still isolates the intended remainder — re-verified: that row is `(model=false, schema=true)` for the byte-vs-character reason alone.
- The new both-reject row is red-proofed: `test_result_truth_table_schema_projections_and_disclosed_divergences` fails on `eb895a61` and passes here.

**Validation at this head:** non-E2E **1886 passed, 532 deselected** (display-capable local; CI's `1884 passed, 2 skipped` is the same 1886 selected); pyright 0 errors; ruff check + format clean; real Godot 4.6.3 `script run` e2e **44 passed**; GitHub checks 8 success / 4 environment-skipped; `CLEAN` / `MERGEABLE`.

No findings remain from my side across the rounds I rechecked (`54cecfce` → `6558ccdf` → `eb895a61` → `de841a27`).

PR #749, opened 2026-08-27
feat(gda): declared read-only game call surface (#673)

### LINKED ISSUE #673: live: allowlisted read-only game call surface
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-033 "Live property reads cannot call debug state contracts"** (2026-08-13, marked low in the notes; the document's GDA-DF numbering collides across sessions — identified by title and date). `game get qa_current_state_contract` returned `live_unknown_property` — gda exposes stored-property reads but no bounded read-only method call, so evidence had to fall back to `_state_index` plus screen capture and independent hashes.

## What to build

An allowlisted read-only `game call` surface with JSON-safe arguments and results.

**Why an allowlist (framing matters):** the target project is trusted (ADR-0009) — gda already executes arbitrary project code via `script run` and script `_init` on load — so the allowlist is **not** a trust boundary. Its purposes are: (a) keep the live *read* surface free of side effects so `State consistency` (ADR-0020) still holds for reads, and (b) bound results to what the `Value projection` can render (ADR-0035; note that ADR's *projection* whitelist is a payload/traversal boundary — a separate mechanism from this method allowlist).

**Enforcement limit, stated up front:** gda cannot verify that a GDScript method has no side effects. The allowlist records the declarer's assertion; the guarantee gda provides is that no *undeclared* method is callable.

**First design decision of this issue — the declaration site.** Candidates: a project-side declaration (e.g. a config the harness reads) vs a per-invocation CLI flag. They differ in discovery, schema, and how the read-only assertion is reviewed; pick one and record why before implementing.

- Results are rendered through the `Value projection`.
- Method-existence and allowlist misses are distinguishable failures.
- No mutating calls are exposed by default.

## Acceptance criteria

- [ ] The declaration-site decision is recorded on this issue.
  > **Amendment (2026-08-27, PR #749 review):** the triage decision's substance
  > (project-side declaration, not a per-invocation flag) is implemented unchanged; two
  > subordinate clauses are amended — the declaration is read PER CALL rather than at
  > `Engine session` start (a per-class constant has nothing to read at session start for
  > a node that does not exist yet), and enumeration is DEFERRED, with the declared set
  > carried on the `live_method_not_allowlisted` message instead. Rationale and scope:
  > [this comment](https://github.com/aigengame/godot-agent/issues/673#issuecomment-5434211424).
- [ ] A method declared at the chosen site can be invoked via `gda game call` and returns a projected value.
- [ ] A non-allowlisted method and a missing method each return a distinct `Gda error code` in an `Error envelope`.
- [ ] The read-only guarantee and its enforcement limits are documented (and recorded in an ADR if the surface is added).

## Blocked by

The declaration-site decision above (design, not code).

qinhaihong-red:
## W5 slice: declared read-only `game call` surface (#673)

**The gap (GDA-DF-033):** a project exposed its debug state as a method, but `game get` reads stored properties and had no bounded way to invoke that contract. This PR adds a declared, projected live-call surface.

### Declaration-site decision (ADR-0041)

The project declares callable method names in a statically-read script constant:

```gdscript
extends Node2D

const GDA_CALLABLE := ["qa_current_state_contract"]

func qa_current_state_contract() -> Dictionary:
	return {"phase": _phase, "ready": true}
```

gda resolves the declaration from the addressed node's attached script along its base chain. An opted-in inheritance chain has **at most one declaration owner**: GDScript rejects a subclass that redeclares a base constant, while a base owner covers its subclasses and need not define every method it names. A chain with no declaration remains default-deny.

The allowlist is project-side, not a per-invocation assertion, and reading it runs no project code (`get_script_constant_map()` / `get_base_script()`). The allowlist is not a trust boundary: the project is trusted (ADR-0009). It bounds the live read surface and projected results. gda cannot verify that a declared method has no side effects; it guarantees that no undeclared method is callable.

The authoritative issue records two implementation amendments in [the dated #673 comment](https://github.com/aigengame/godot-agent/issues/673#issuecomment-5434211424): declarations are read per call, so runtime-created nodes and current scripts are covered, and a dedicated enumeration operation is deferred. A `live_method_not_allowlisted` error carries the resolved declaration meanwhile.

### What shipped

- `gda game call <node> --method NAME [--args JSON]`, `kind = LIVE`, served by the existing daemon/harness path. One method is invoked per request at a frame boundary, and the result uses the shared recursive Value projection (ADR-0035); void projects as `null`.
- Three distinguishable pre-call refusals: `live_unknown_method`, `live_method_not_allowlisted`, and `live_invalid_call_args`. Count and type checks happen before `callv`, avoiding false-success `null` results and Session-log errors.
- Live JSON produces six source Variant types: null, bool, float, String, Array, and Dictionary. Godot's live parser materializes **every JSON number as float**. The preflight table mirrors the retained `Variant::can_convert_strict` edges, including bool/float numeric conversions, String to `StringName` / `NodePath` / `Color`, null to `Object`, and Array to all ten `Packed*Array` targets. A real-engine `callv` oracle covers all 18 non-identity edges and representative refusal/identity boundaries.
- The shared params model recursively refuses `NaN` / `Infinity` and JSON integer tokens decoded as Python `int` outside ±`9007199254740991`. Finite float values do not inherit that integer-only bound; real-engine regressions pin `1e17`, `2.5e17`, and `1e300`. Standard JSON Schema cannot distinguish an exponent/fraction token from a mathematically equal integer, so the recursive schema keeps its `number` branch broad and explicitly identifies the params model as the execution authority for the integer-token bound. RFC JSON excludes `NaN` / `Infinity`, but some in-memory schema validators accept those extensions; the model/schema corpus pins that deliberate over-acceptance. Godot 4.6.3 can also parse ordinary normal small-magnitude values such as `1.2345678901234567e-300` as `0.0`, and can lose small result values during `JSON.stringify`; [#752](https://github.com/aigengame/godot-agent/issues/752) owns the cross-operation transport fix.
- Deferred, recorded rather than silent: a dedicated enumerate operation and `--texture-digest` for call results remain follow-ups.

### Intended public-surface diffs

- `gda schema`: exactly one command added, `game call`; no existing command changed, verified against merge base `878b6da4`.
- Three registered error codes with ADR-0002 rows: `live_unknown_method`, `live_method_not_allowlisted`, and `live_invalid_call_args`.
- ADR-0041 plus synchronized `CONTEXT.md`, command catalog, bundled Skill, README, and Spanish/Japanese/Chinese README translations.
- Bundled harness base→head version: **12 → 15**. Versions 13 and 14 were within-PR intermediates retired when the harness body changed again; the cross-revision guard validates the final transition.

| AC | Evidence |
| --- | --- |
| Declaration-site decision recorded | ADR-0041 and the dated issue amendment describe chain ownership, static lookup, per-call timing, and deferred enumeration. |
| Declared method invocation returns projected values | Real-engine e2e covers nested projection, args/defaults, void, a base-owned declaration invoked on a subclass node, and preserved high-range floats. |
| Missing, undeclared, and invalid args are distinct | Real-engine and unit regressions pin all three error codes and pre-call refusal, including conversion edges and typed-container refusal. |
| Limits are public and consistent | Schema/help/docs disclose the integer-token runtime bound, both standard-schema limitations, the verified high-range float examples, and the tracked small-magnitude input/result limitation. |

### Validation at `a6d97071`

- Fast/headless selection: **1899 passed, 2 skipped, 539 deselected** in 63.75 s. The two skips are display-gated; they are not reported as passes. Pyright: 0 errors/warnings; ruff check and format clean across 468 files; `git diff --check` clean.
- Full real-engine `tests/test_e2e_game.py` on Godot 4.6.3: **9 passed** in 136.11 s. This includes the complete conversion-edge oracle, observed numeric `typeof`, safe-integral boundaries, high-range floats (`1e17`, `2.5e17`, `1e300`), non-finite/session preservation, declaration inheritance, parse-error pin, and projected call surface.
- Focused schema/model/help/contract regressions: **160 passed** in 25.93 s.
- Direct Godot probes reproduced the small-float class accurately: `1e-300` and `1e-308` remain non-zero on parse; the normal value `1.2345678901234567e-300`, `DBL_MIN`, and subnormals can parse as `0.0`; `JSON.stringify` can also turn small non-zero results into `0.0`. The root path is Godot `JSON::_get_token` → `String::to_float`; #752 tracks the complete request/result transport decision.
- Independent final review found no actionable repository finding, confirmed that a decimal heuristic would be partial and can over-refuse, and required this PR-body synchronization. It approved the candidate after that external-artifact item was reserved for delivery.
- RED evidence: on `d7665623`, schema/argv/pure-`--params-json`/real-Godot regressions rejected the three high-range floats; on `c570a8f4`, the new schema/help assertions failed because the normal-value and in-memory-validator limitations were absent. Both classes pass at `a6d97071`.
- GitHub's Godot e2e job is skipped by host policy; it is not counted as runtime evidence above. All applicable CI checks at this head pass.

Closes #673


qinhaihong-red:
Adversarial review at exact head `0887030774aecd38f8b0b5aa2fcddccd63aae1c3` (base `878b6da4ead780937e7094ac4d5f6408c266ee0b`).

**Outcome: changes required.** Artifact Review, Design Domain Modular Architecture, and the Standards/Spec Review were run as independent axes. Findings repeated across axes are retained deliberately because they establish different contract failures.

## Artifact Review — Changes required

1. **[P1] Type-incompatible JSON arguments are documented as a usable call surface but are reported as a successful `null`.**

   At `src/gda/harness/gda_harness.gd:575`, only arity is validated before `callv`. On Godot 4.6.3, an allowlisted `func typed(value: int) -> int` called with `--args '["bad"]'` exits 0 with `"value": null`; `gda diag errors` then reports `Cannot convert argument 1 from String to int.` The call did not succeed, yet its result is indistinguishable from a legitimate `null`/`void` return and it pollutes the Session log. This leaves ADR-0041:104-107, the command catalog, ADR-0002, and the PR evidence incomplete. Reject definite type incompatibility before `callv` (or otherwise map it reliably to an Error envelope), add a real-engine regression, and reconcile the error-code meaning, ADR, catalog, schema/help, and PR evidence.

2. **[P1] The implementation changes the authoritative triage decision without completing the decision on issue #673.**

   The triage comment requires a project-side declaration read by the harness at Engine session start and identifies enumeration as the discovery path. ADR-0041:69-76 changes this to a per-call read and ADR-0041:109-112 defers enumeration. The ADR discloses the deviation, but the issue does not accept it; the statement that the timing clause merely described a config-file shape is not present in the authoritative brief. Record and confirm an amended issue decision (including discovery deferral and any follow-up issue), or implement the session-start read/enumeration decision before closing #673.

3. **[P2] The inheritance limitation conflicts with the decision's central “beside the method / whole class surface” reviewability rationale.**

   ADR-0041:47-67 and the README say the declaration is beside the methods and lets a reviewer see the class's whole callable surface. ADR-0041:78-85 and the real-engine test establish that a subclass cannot redeclare `GDA_CALLABLE`; exactly one class per inheritance chain can contribute. Consequently either the base knows subclass-only methods or the leaf lists inherited methods, so the declaration is not necessarily beside the implementation it asserts. Use a shape that supports per-class incremental declaration, or explicitly redefine this as an inheritance-chain declaration and reconcile the rationale, ownership, examples, README translations, and bundled Skill.

4. **[P2] The distributed Skill omits the enforcement limit, while the PR body says it includes it.**

   `src/gda/skill/SKILL.md:171` says that gda calls nothing undeclared but does not say that gda cannot verify side-effect freedom. The PR acceptance-evidence table says the Skill states both facts. The README's “Project code execution” inventory at README.md:650-683 also omits the direct project-method execution added by `game call`. Add the enforcement limit and relevant inheritance limitation to the Skill, add `game call` to the execution inventory (and translations), and correct the PR evidence.

**Terminology and prose assessment:** reviewed against `CONTEXT.md`, issue #673, ADR-0009/0020/0035, and ASD-STE100. Apart from the conflicts above (`strictly stronger`, `whole callable surface`, `beside the methods`, and the missing enforcement limit), no separate substantive terminology or prose problem was found. The three translated README additions match the English meaning and their source hashes match the exact head.

## Design Domain Modular Architecture

1. **[P1] Non-finite numbers bypass the params authority, become a 30-second transport timeout, and discard session state.**

   `src/gda/commands/game.py:256` permits arbitrary nested values, and the argv path at line 607 uses permissive Python `json.loads()`. Both Python and Pydantic accept `NaN`/`Infinity`, although they are outside the published JSON schema. Exact-head `--args '[NaN]'` waited 30 seconds and returned `live_timeout`; the daemon then marked the channel stale and relaunched the Engine session, losing prior runtime state. `GameCallParams.model_validate_json(... "args":[NaN])` also accepts the value, so `--params-json` is affected. Recursively reject every non-finite float in `GameCallParams`, cover argv and `--params-json`, and add a live regression proving immediate structured refusal with unchanged session identity/state.

2. **[P1] The call-failure owner is structurally incomplete for argument types.**

   The same real-engine counterexample as Artifact finding 1 proves that `live_invalid_call_args` owns count mismatches but not definite type incompatibility from the same method signature. Reuse the selected `get_method_list()` entry to validate compatibility before `callv`, without introducing the property string-coercion table. Either widen `live_invalid_call_args` deliberately or add a distinct LIVE code, then update the registry/harness mirror and real-engine tests.

3. **[P2] `GDA_CALLABLE` has independently editable Python and GDScript protocol copies without a required mirror guard.**

   The agent-facing schema/help copies the name in `src/gda/commands/game.py:250`, while runtime owns `GDA_CALLABLE_CONST` in `src/gda/harness/gda_harness.gd:73`. `tests/test_game_commands.py:795-806` only checks the schema's own text; it does not compare the harness constant, and this PR's GitHub Godot e2e job is skipped. Define a Python-side public constant used by schema/help and add a default-gate static mirror test against the harness constant.

The existing dependency direction remains sound: `cli composition -> commands/game.py -> dispatch/descriptor -> live runner -> daemon/session -> harness`. The vertical slice, shared value projection, error registry, and runtime communication remain orthogonal. No reverse dependency, service locator, or speculative registry was introduced.

## Review

### Standards

- **[P1] Structured-failure contract violation:** the type-incompatible `callv` path returns exit 0 plus `null`, violating ADR-0002/0017 and making failure indistinguishable from a valid return.
- **[P3, judgement call] Duplicated Code:** `tests/test_e2e_game.py:326` repeats the 16-line `run(*args)` wrapper from line 199. Extract a project/Godot-aware helper or fixture so protocol arguments, environment, and timeout do not drift.

### Spec

- **[P1]** The type-incompatible call does not invoke the declared method successfully, so it does not satisfy issue #673's JSON-safe argument/result goal or AC2 despite returning success.
- **[P2]** `NaN`/`Infinity` should be rejected at the input boundary, not surface as `live_timeout` and session loss.
- **[P2]** The session-start declaration timing is not implemented; keep it, or obtain an explicit issue-level amendment before merge.

## Validation

- Exact-head targeted unit tests: `86 passed`.
- Exact-head fast suite: `1892 passed, 2 skipped, 534 deselected` (the same count appears in GitHub CI).
- Exact-head Godot 4.6.3 `tests/test_e2e_game.py`: `4 passed` on the original matrix.
- Adversarial Godot probes reproduced both false-success-on-type-mismatch and `NaN` -> 30-second `live_timeout`.
- Aggregate schema: base 75 commands, head 76; `game call` is the only added command and existing command schemas remain unchanged.
- Ruff, format check, and `git diff --check`: clean. GitHub pyright and Python jobs are green.
- GitHub `Godot e2e tests`: **SKIPPED**, so it is not counted as runtime CI evidence.

No repository files were changed during review.

qinhaihong-red:
## Review response — all findings adopted at head `5d87b43e`

Baseline: base `878b6da4`, reviewed head `08870307`, new head `5d87b43e` (one fix commit + an issue amendment + PR-body corrections). Every runtime finding was reproduced before it was fixed, and the type-mismatch class turned out to be **wider than reported** — see the row below.

| Claim | Decision | Change |
| --- | --- | --- |
| Artifact 1 / Arch 2 / Standards P1 / Spec P1 — type-incompatible arguments reported as a successful `null` | **Fully adopt, and widened by my own reproduction** | Reproduced your counterexample exactly (`typed(value: int)` with `["bad"]` → exit 0, `value: null`, `Cannot convert argument 1 from String to int.` in the Session log). Probing the rest of the parameter space found **three more false successes you did not report**: `null` into `int`, a Dictionary into an `Object` parameter, and any JSON array into a typed `Array[int]` (the engine refuses `Array` → `Array[int]` outright, so no JSON value can ever reach such a parameter). The gate now checks each argument's TYPE before `callv`, mirroring the engine's own `Variant::can_convert_strict` (`core/variant/variant.cpp` — not exposed to GDScript) restricted to the SEVEN Variant types JSON can produce, so the mirrored table is closed and small. `live_invalid_call_args` is **deliberately widened** (registry description, ADR-0002 row, ADR-0041, catalog, help) rather than a fourth code minted: count and type are both "the arguments do not fit this signature". The e2e pins both directions — 4 engine-refused shapes AND 6 conversions the engine really performs (`true`/`3.7` into `int`, `3` into `float`, `null` into `Object`, a Dictionary into an untyped parameter) — so the gate mirrors the engine instead of over-refusing. |
| Arch 1 / Spec P2 — `NaN`/`Infinity` → 30 s `live_timeout` + session loss | **Fully adopt** | Reproduced end to end, including the state loss (session id `c56fbed0…` before, `64bb2324…` after the next call). Non-finite floats are now refused **recursively** in `GameCallParams` — the one authority both argv and `--params-json` pass through (ADR-0015) — so the refusal is structural and immediate. The live regression asserts the refusal is fast (< 10 s, not the 30 s bound), carries no `live_timeout`, and leaves the `session_id` **unchanged** across it. |
| Artifact 2 / Spec P2 — the triage decision was amended without amending the issue | **Fully adopt** | Recorded on the authoritative brief: a [dated amendment comment on #673](https://github.com/aigengame/godot-agent/issues/673#issuecomment-5434211424) covering both clauses (per-call read; enumeration deferred) with rationale and what did NOT change, plus an annotation on AC1 in the issue body pointing at it. The ADR now cites the issue amendment rather than carrying the deviation alone. The enumerate follow-up is recorded as deferred-not-rejected; say the word and I file it through `/to-issues`. |
| Artifact 3 — the inheritance limitation conflicts with the "beside the method" rationale | **Fully adopt — redefined as an inheritance-CHAIN declaration** (your second option) | Per-class incremental declaration is not available in this format (GDScript's shadowing rule), so the ADR now *defines* the constant as the chain's declaration owned by exactly one class, and states the cost plainly: one authoritative list per chain, not necessarily in the file of every method it names. The "beside the method / whole callable surface of a class" wording is gone from the ADR, README, the three translations, the catalog and the Skill; a repo-wide grep for the retired phrasings returns nothing. |
| Artifact 4 — the Skill omits the enforcement limit; README's execution inventory omits `game call`; the PR evidence overstated | **Fully adopt** | The Skill row now carries the enforcement limit verbatim ("gda CANNOT verify a declared method has no side effects … what gda guarantees is only that no undeclared method is called"), the chain rule, and all three refusal codes. README's *Project code execution* inventory gains a `game call` bullet (and all three translations gain both it and the reconciled paragraph). The PR body's AC row is corrected and now says the Skill did NOT carry the limit before this round. |
| Arch 3 — `GDA_CALLABLE` duplicated across languages with no guard | **Fully adopt** | `gda.commands.game.GDA_CALLABLE_CONST` is the Python-side authority the schema, help and prose quote, and `tests/test_live_contract_guards.py` gains a default-gate static mirror test comparing it to the harness's `GDA_CALLABLE_CONST` and asserting the published input schema names it. |
| Standards P3 — duplicated e2e `run(*args)` wrapper | **Fully adopt** | Extracted `_gda_runner(project, timeout)`; both call e2es use it, so protocol arguments, environment and timeout cannot drift. |

**Harness `HARNESS_VERSION` 13 → 14** (13 declared two bodies within this PR and is retired, per the #743/#746 precedent); pins updated.

### Validation at `5d87b43e`

- Fast suite **1897 passed, 536 deselected**; pyright 0 errors; ruff check + format clean; `git diff --check` clean.
- Real Godot 4.6.3 e2e: `tests/test_e2e_game.py` **6 passed** — the original matrix, the parse-error pin, the new argument-type matrix, and the non-finite session-preservation regression. Both matrices assert `diag errors == []`, so no refusal path leaves engine noise (the false successes did).
- **Red-proof on the reviewed head `08870307`**: the non-finite unit regression, the `GDA_CALLABLE` mirror guard, and both new e2e regressions fail there — the NaN one taking **32 s**, which is the timeout it now prevents.
- `gda schema` vs the dev tip: still exactly ONE added command (`game call`), no existing command changed.

qinhaihong-red:
Re-review at exact head `5d87b43ea1bba2364e0a29e46bb5e1d2a1a5d517` (base `878b6da4ead780937e7094ac4d5f6408c266ee0b`).

**Outcome: changes required.** The original false-success-on-type-mismatch, non-finite-number timeout/session loss, issue-level decision-amendment, and Skill/README omission findings are closed. The findings below are residue or new counterexamples introduced/discovered while replaying the fixes. The review axes remain independent, so cross-axis findings are intentionally retained.

## Artifact Review — Changes required

1. **[P2] The PR/ADR declaration-site wording still conflicts with the inheritance-chain decision.**

   The PR body still says the class declares the constant “beside the methods it describes” and that reading one class shows its “whole callable surface.” ADR-0041:34-50 likewise leads with the addressed class's own/attached script, while ADR-0041:59-64 says one class can own the whole inheritance chain and may list subclass methods that live elsewhere. This can send an adopter to redeclare the constant in a subclass and hit the documented parse error. Define the public rule consistently as: gda resolves from the attached script along the base chain; an opted-in chain has at most one declaration owner; that owner is not necessarily beside every method it names. Remove the remaining “beside methods” / “whole surface of a class” rationale from the PR body and stale source comments.

2. **[P2] The PR evidence still misreports the exact-head test result and harness-version diff.**

   The PR body and response say `1897 passed, 536 deselected`; both the exact-head local run and GitHub job `98406262395` report `1895 passed, 2 skipped, 536 deselected`. `1897` is selected, not passed. The complete base-to-head harness change is `12 -> 14`, not `12 -> 13`; version 13 is the retired within-PR intermediate. Correct the body and add an explicit correction to the response.

3. **[P2] The authoritative issue amendment's ADR link is currently broken.**

   The final link in issue #673 comment `5434211424` points to ADR-0041 on `feat/gda-dogfooding-dev`, which does not contain that file and currently returns 404. Link to the immutable current commit (or at least the PR head) so AC1's evidence is inspectable before merge.

4. **[P3] The rendered public help contains `an an argument`.**

   `src/gda/commands/game.py:638-639` renders “..., an an argument the declared parameters cannot take ...”. Remove the duplicate article.

Terminology/prose recheck otherwise passed: the enforcement limit, error names, Engine-session terminology, README inventory, bundled Skill, and three translations are now reconciled.

## Design Domain Modular Architecture — Changes required

1. **[P1] The handwritten strict-conversion mirror rejects calls Godot accepts.**

   `src/gda/harness/gda_harness.gd:668-683` claims to mirror `Variant::can_convert_strict` over the seven JSON-source Variant types, but it omits at least `String/int -> Color` and `Array -> Packed*Array`. On Godot 4.6.3, direct `callv` accepts `"red"` for a `Color` parameter and `[1,2,3]` for `PackedInt32Array`; exact-head `gda game call` rejects both as `live_invalid_call_args`. This disproves ADR-0041:127-137 and the response's “where the engine converts, gda calls” claim.

   Complete the JSON-source conversion closure and add a real-engine conformance matrix that uses direct `callv` as the oracle rather than another copy of the handwritten table. Otherwise the harness remains a second, already-incomplete authority for engine conversion knowledge.

2. **[P2] The `GDA_CALLABLE` mirror guard does not cover all public Python text.**

   The Python/harness constants and static guard are useful, but `src/gda/commands/game.py:275,603,624` still hard-code `GDA_CALLABLE`. `tests/test_live_contract_guards.py:49-73` only proves the two constants match and that the schema contains the constant somewhere. A coordinated constant rename can therefore leave model/help text publishing the old name while the guard passes. Derive the public strings from `GDA_CALLABLE_CONST`, or add an exact rendered schema/help consistency guard.

The dependency direction remains sound (`cli -> commands/game -> dispatch/live runner -> daemon/session -> harness`); no reverse dependency or new shared mutable registry was introduced.

## Review

### Standards

- **[P1] Large JSON integers silently change value across the live wire.** `src/gda/commands/game.py:245-267` rejects non-finite floats but permits Python's arbitrary-precision integers. Exact-head Godot 4.6.3 returned success while `9007199254740993` became `9007199254740992`; through `typed(value: int) -> value * 2`, the result was `18014398509481984` instead of `18014398509481986`. Larger values can become `INT64_MAX` or `null`, with empty diagnostics. Reject integers that cannot round-trip exactly through the wire, publish the supported bound, and cover argv, `--params-json`, positive/negative boundaries, and session preservation.
- **[P2]** The strict-conversion admission table over-refuses valid Color and PackedArray calls, as reproduced above.
- **[P3, judgement call]** `_gda_runner` now removes duplication between the call e2es, but `tests/test_e2e_game.py:117-137` still repeats the same binary/environment/argv/timeout wrapper instead of using the module helper. The harness comment at `gda_harness.gd:530-533` also retains the obsolete “beside the code” ownership claim.

### Spec

- **[P2] Valid JSON-source conversions are rejected.** Issue #673 requires JSON-safe arguments and AC2 requires a declared method to be invocable; the Color/PackedArray counterexamples are legitimate declared calls that gda blocks despite Godot accepting them.
- **[P2] Large valid JSON integers produce a successful but different invocation.** Silent numeric mutation violates the issue's JSON-safe argument/result requirement and AC2. Refuse values that cannot be represented losslessly by the live protocol.

The previous Spec findings for String-to-int false success, NaN/Infinity timeout/session loss, and session-start timing are closed: the current issue amendment authorizes per-call reads and deferred enumeration, and the code/ADR match it.

## Validation

- Exact-head targeted unit tests: `95 passed`.
- Exact-head fast suite: `1895 passed, 2 skipped, 536 deselected`.
- Exact-head Godot 4.6.3 `tests/test_e2e_game.py`: `6 passed`.
- Adversarial real-engine probes reproduced Color/PackedArray over-refusal and large-integer silent mutation.
- Ruff, formatting, and `git diff --check`: clean. GitHub Python/ruff/pyright jobs are green.
- GitHub `Godot e2e tests`: **SKIPPED**, so local Godot evidence remains the runtime evidence for this head.

No repository files were changed during re-review.

qinhaihong-red:
## Re-review response — all findings adopted at head `4135c8c2`

Baseline: base `878b6da4`, re-reviewed head `5d87b43e`, new head `4135c8c2`. Both P1s were reproduced before they were fixed, and the conversion one is now pinned by an oracle rather than by another table.

| Claim | Decision | Change |
| --- | --- | --- |
| Arch 1 / Standards P2 / Spec P2 — [P1] the hand-transcribed strict-conversion mirror rejects calls Godot accepts | **Fully adopt, both halves of your remedy** | Reproduced: `takes_color(["red"])` and `takes_packed_int([[1,2,3]])` — gda REFUSED, the engine CALLS. Root cause: I transcribed the per-target rows by hand and stopped at the ones I had probed. The table is now **extracted mechanically** from `Variant::can_convert_strict`, keyed by SOURCE type, over the seven JSON-producible Variants — which also caught that the `valid[]` lists are **NIL-TERMINATED** (a NIL entry is the terminator, not a legal source; `from NIL` is the early return to `OBJECT` only). It gained `int/String → Color` and `Array → all ten Packed*Array` types. And your second remedy is what keeps it honest: a real-engine **conformance matrix whose oracle is a direct `callv`** — a declared `probe_direct(method, args)` runs the call inside the game and reports whether the method body RAN — asserting gda's verdict equals the engine's across 21 (parameter type, JSON value) pairs. The harness table can no longer drift silently, because the engine is the reference. |
| Standards P1 / Spec P2 — large JSON integers silently change value | **Fully adopt** | Reproduced, and worse than reported: `9007199254740993` → `…984`, and `123456789012345678901234567890` → **`-2`**, both as successful calls. The wire's JSON parser reads every number as a double, so integers outside ±(2^53 − 1) cannot round-trip; they are refused recursively in the params model (argv and `--params-json`), the bound is published in the schema description, and the regressions cover both signs, a nested position, the `--params-json` path, session preservation, and that the boundary value `9007199254740991` still goes through **unchanged**. |
| Artifact 1 — declaration-site wording still conflicts | **Fully adopt, using your formulation** | The ADR now states the public rule once, in the decision itself: *gda resolves from the attached script along the base chain; an opted-in chain has at most one declaration owner; that owner is not necessarily beside every method it names.* The "beside the methods" / "whole callable surface of a class" rationale is gone from the ADR, the PR body, and the stale harness comment at `gda_harness.gd`; a repo-wide grep for both phrasings returns nothing. |
| Artifact 2 — evidence misreports the test count and harness diff | **Fully adopt** | The PR body now distinguishes selected/passed/skipped and names the environment (**1898 selected**; 1898 passed here, 1896 + 2 skipped on a headless host / CI), and states the base-to-head harness change as **12 → 14**, with 13 named as the retired within-PR intermediate. **Correction to my previous response:** its "1897 passed" was this host's number reported as if universal, and its "13 → 14" described the last hop rather than the base-to-head diff you review. |
| Artifact 3 — the issue amendment's ADR link 404s | **Fully adopt** | The comment on #673 now links ADR-0041 at the immutable PR-head commit (re-pinned after the final push and verified to resolve, HTTP 200), with a note explaining why the branch link could not work before merge. |
| Artifact 4 — `an an argument` in the rendered help | **Fully adopt** | Fixed. |
| Arch 2 — the `GDA_CALLABLE` guard misses hard-coded public text | **Fully adopt** | The option help is derived from `GDA_CALLABLE_CONST`; the two remaining mentions are docstrings, which cannot interpolate, so the guard now does the exact check you offered as the alternative: it renders `game call --schema` AND `--help`, extracts every `GDA_*` token in *declaration-constant position*, and asserts the set equals the constant — a rename that leaves stale published text fails it. (Scoped to that position because the shared options legitimately document `GDA_PROJECT`/`GDA_GODOT`.) |
| Standards P3 — the remaining duplicated e2e wrapper | **Fully adopt** | The texture e2e uses `_gda_runner` too; no local wrapper remains. |

### Validation at `4135c8c2`

- Fast suite **1898 selected** (`1898 passed, 538 deselected` here; 1896 + 2 skipped on a headless host/CI); pyright 0; ruff check + format clean; `git diff --check` clean; the cross-revision harness guard passes for `878b6da4 → HEAD`.
- Real Godot 4.6.3 `tests/test_e2e_game.py`: **8 passed** — adding the engine-oracle conformance matrix and the integer-bound regression.
- **Red-proof on `5d87b43e`**: both new e2e regressions and the new integer-bound unit test fail there. In the interest of not overclaiming: the rendered-text guard does NOT fail there — it is drift prevention for a future rename, not a regression for a present defect.



---

**Correction (mine, within minutes of posting):** this comment first said `1899 selected / 1899 passed`. The measured value at this head is `1898 passed, 538 deselected` on this host; I wrote the figure from memory instead of from the run, which is the exact habit your finding 2 called out. Both this comment and the PR body now carry the measured line.

qinhaihong-red:
Third review pass at exact head `4135c8c296d9e5129a81ba183d949194066ce33f` (base `878b6da4ead780937e7094ac4d5f6408c266ee0b`).

**Outcome: changes required.** The runtime fix for silent large-integer mutation is effective, String-to-Color and Array-to-PackedArray admission now match the engine, the immutable issue link resolves, and the `GDA_CALLABLE` rendered-text guard is effective. The items below are current-head residue or newly exposed test/transport boundaries. The review axes remain independent, so the schema finding is retained where it establishes different contract failures.

## Artifact Review — Changes required

1. **[P2] The inheritance-chain cardinality is still inconsistent across public artifacts.**

   ADR-0041:47-52 now correctly says an opted-in chain has **at most one** declaration owner and that the owner need not define every listed method. README.md:565, the Spanish/Japanese/Chinese translations, `docs/command-catalog.md:864-866`, `src/gda/skill/SKILL.md:171`, and the PR body still say exactly one class per chain declares. That excludes the valid default-deny case: a chain that does not opt in has zero owners. ADR-0041's heading and lines 34-36 also still lead with the class's own/attached script before the base-chain resolution rule. Use one formulation everywhere: gda resolves from the attached script along the base chain; an opted-in chain has at most one owner; that owner need not define every method it names.

2. **[P2] The PR body still contains stale current-head evidence.**

   Its public-surface list says `HARNESS_VERSION 12 -> 13`, while its validation section correctly says the base-to-head change is `12 -> 14`. It also reports the real-engine suite as 6 passed, while current head has 8 tests and independently ran as 8 passed. Update both statements and include the conformance-oracle and safe-integer regressions in the e2e inventory.

3. **[P2] The safe-integer bound is claimed as published but is absent from the public schema.**

   ADR-0041:158-162, `game.py:251`, and the latest response say the recursive `+/- (2^53 - 1)` limit is published in the params schema. Exact rendered `game call --schema` and `GameCallParams.args` at `game.py:316-326` mention only NaN/Infinity; `args.items` remains `{}`. Schema-driven callers therefore accept values the model always refuses. Publish the recursive bound in the machine contract and description, and guard schema/model boundary parity.

4. **[P3] The public help typo remains.**

   `game.py:665-666` and exact rendered help still say `an an argument`. Remove the duplicate article and pin the rendered text.

The exact-head fast-count statement is now correct, and the issue amendment's immutable ADR link returns 200.

## Design Domain Modular Architecture — Changes required

1. **[P1] The strict-conversion fix is only partial: the `int -> Color` row is unreachable on the live protocol, and its oracle case is vacuous.**

   `src/gda/daemon/protocol.py:58` serializes with plain `json.dumps`; `gda_harness.gd:229` parses with Godot `JSON.parse_string`. Godot 4.6.3 parses `16711680` as `float`. Direct `callv` with a GDScript int calls a `Color` parameter, while the parsed float does not. Consequently `JSON_ARGUMENT_CONVERSIONS`'s `TYPE_INT -> TYPE_COLOR` row never receives a live JSON argument, and exact-head `gda game call ... --args '[16711680]'` still returns `live_invalid_call_args`.

   The new `probe_direct` oracle receives its argument through the same wire first, so both sides see float and refuse; the green comparison does not exercise the advertised int row. Decide the actual live numeric domain: preserve/tag int versus float, deliberately canonicalize integral numbers before admission/call, or remove the unreachable row and document that live JSON numbers arrive as float. Assert the observed `typeof` in the oracle and explicitly cover every retained source-to-target edge.

2. **[P2] Safe-integer enforcement has one correct owner, but its schema projection is missing.**

   The recursive model validator is the correct shared authority for argv and `--params-json`; the required public projection and schema regression are absent.

The dependency direction remains sound, and the `GDA_CALLABLE` constant/rendered-text guard is closed.

## Review

### Standards

- **[P2] Machine contract mismatch:** the emitted schema accepts integers that `GameCallParams` necessarily refuses, contrary to ADR-0004/0015's schema/accepted-input alignment.
- **[P3] Public-copy defect:** rendered help still contains `an an argument`.

### Spec

- **[P2] The JSON-safe integer enforcement limit is not published to callers in schema/help, despite issue #673 AC4 and ADR-0041 saying it is.**
- **[P3] Two `--params-json` regressions test the wrong refusal.** `tests/test_game_commands.py:875-889` and `tests/test_e2e_game.py:635-652` combine positional/individual arguments with `--params-json`; both receive `usage_error` for mutually exclusive inputs, not the safe-integer validator's `invalid_params`. The implementation currently handles a pure `--params-json` request correctly, but the claimed regression path is unprotected. Invoke only `--params-json`, assert the code and range diagnostic, and cover accepted positive/negative boundaries plus session identity.

## Validation

- Exact-head fast suite: `1896 passed, 2 skipped, 538 deselected`.
- Exact-head Godot 4.6.3 suite: `8 passed`.
- Additional PackedArray element-boundary probes agreed with direct `callv`.
- Correct pure `--params-json` oversized integer returns `invalid_params`; the two committed tests' mixed invocation returns `usage_error`.
- Ruff, format check, and pyright are clean.
- GitHub's Godot e2e job is **SKIPPED**, so the current runtime evidence is local.

No repository files were changed during this review pass.

qinhaihong-red:
## Third re-review handled — `d7665623`

I evaluated the findings against the exact reviewed head `4135c8c2`, implemented the accepted residue, independently re-reviewed the final worktree, and pushed commit `d76656236aa5b37337a259b2214878ee775fdf97`.

| Finding | Decision and resolution |
| --- | --- |
| Chain cardinality and declaration-site wording | **Adopted.** All public artifacts now say that an opted-in attached-script/base chain has **at most one declaration owner**; a base owner covers subclasses and need not define every named method. Zero-owner chains remain default-deny. ADR-0041's title and the issue amendment use the same rule. |
| Stale PR evidence | **Adopted.** The PR body now records base→head harness `12 → 15`, the final test counts, six live JSON source types, and the complete real-engine evidence. |
| Safe-integral schema/help gap | **Adopted.** `GameCallParams` now emits a recursive Draft 2020-12 live-argument schema with ±`9007199254740991` on integral values. Runtime validation applies the same rule to integer-valued floats, matching JSON Schema semantics. Model/schema boundary corpus tests and rendered-help assertions pin the contract. |
| `an an argument` help typo | **Adopted.** Fixed and covered by a rendered-help regression. |
| Unreachable `TYPE_INT → Color` source row | **Partially adopted at the mechanism level; the problem is fully closed.** The daemon sends plain JSON and Godot materializes every JSON number as `TYPE_FLOAT`, so a live `TYPE_INT` source cannot exist. I removed the unreachable row and documented/asserted the actual six-source domain instead of adding a new type-tag protocol or integral canonicalization mechanism. |
| Conversion oracle coverage | **Adopted.** The real-engine matrix now covers all 18 retained non-identity source→target edges, including all ten `Array → Packed*Array` targets, plus identity/refusal boundaries. Numeric `typeof` asserts both integer-looking and fractional-form literals arrive as float. Fault injection removing `STRING → NODE_PATH` fails precisely with `gda=False / engine=True`. |
| False-positive pure `--params-json` tests | **Adopted.** Oversized-integral and adjacent non-finite cases no longer mix individual args with `--params-json`; they now assert `invalid_params`, the range/finite diagnostic, and unchanged Engine-session identity. Positive ±boundary cases are covered. |

Validation at `d7665623`:

- Fast/headless selection: `1898 passed, 2 skipped, 538 deselected` (123.39 s).
- Full real-engine `tests/test_e2e_game.py`, Godot 4.6.3: `8 passed` (98.00 s).
- Focused final regression set: `47 passed`; earlier complete artifact selection: `227 passed`.
- Pyright: 0 errors/warnings; ruff check and format clean; `git diff --check` clean.
- Harness SHA pin and base→head version-transition guard pass.
- Independent final review: the prior P2 is closed; no new findings.
- GitHub CI at the pushed head: all applicable checks pass. Godot e2e is **SKIPPED** by CI host policy and is not presented as runtime success; the local Godot 4.6.3 run above is the runtime evidence.

The PR body and [issue #673 amendment](https://github.com/aigengame/godot-agent/issues/673#issuecomment-5434211424) are reconciled; the issue's ADR link is pinned to immutable head `d7665623` and verified readable.

qinhaihong-red:
## Independent exact-head recheck at `d7665623` — one required change (P2)

Rechecked in a detached worktree, treating the response as claims. The third review's findings are closed; the change that closed the safe-integer one introduced a new over-refusal.

### Verified closed

- **The `TYPE_INT` source row removal is correct, and I confirmed its premise directly.** A real-engine probe of `JSON.parse_string('[7, 16711680, 1e+300, 2.5, 9007199254740993]')` returns `typeof == 3` (TYPE_FLOAT) for **every** element, including the integer literals — so no live JSON argument can present as TYPE_INT. The e2e now pins that same fact for `16711680` and `16711680.0` before the oracle runs, which is exactly the "assert the observed `typeof`" the review asked for.
- **The published recursive bound really validates.** `$dynamicRef`/`$dynamicAnchor` resolve under the real Draft 2020-12 validator: model and schema agree on all nine probes I ran, including `[{"deep": [2^53+1]}]` and `[[[[{"a": [2^53+1]}]]]]` (both reject) and legal nested values (both accept). `args.items` is no longer `{}`.
- Cardinality wording is now consistent — "at most one declaration owner" in the ADR (title and lead included), catalog, Skill, README and all three translations; zero-owner chains read as default-deny. The `an an argument` typo is gone and pinned by a rendered-help test. The two `--params-json` regressions now invoke `--params-json` ALONE and assert `invalid_params`.
- Counts and versions reconcile: fast suite `1900 passed, 538 deselected` on this display-capable host — the same 1900 selected as the body's headless `1898 passed, 2 skipped`; `tests/test_e2e_game.py` **8 passed** on Godot 4.6.3; base→head harness `12 → 15` with the cross-revision guard passing; pyright 0, ruff clean, `git diff --check` clean.

### Changes required (P2) — every float argument at or above 2^53 is now refused, and the published reason for refusing it is false

Widening the bound to integral floats bought schema parity at a cost that is neither intended by the finding nor disclosed. Reproduced at this head:

```text
GameCallParams(args=[1e17])    -> REFUSED   ("… when integral (the live wire reads JSON
GameCallParams(args=[2.5e17])  -> REFUSED    numbers as doubles, so a larger value
GameCallParams(args=[1e300])   -> REFUSED    arrives as a DIFFERENT value)")
```

Two problems:

1. **The stated reason is false for these inputs.** A Python float IS already a double, so it crosses the wire unchanged — proven in-engine: `JSON.parse_string('[1e+300]')` returns the value with `preserved == true` on comparison to `1e300`. The wire-precision argument holds for oversized *integer literals* (`9007199254740993` → `…992`, which is what the finding was about); it does not hold for any float.
2. **The refusal is broader than "large numbers".** Every double at or above 2^53 has no fractional part, so `is_integer()` is true for all of them: `2.5e17` is refused as readily as `1e300`. A declared method taking a `float` — a nanosecond timestamp, a large id, a physics constant — cannot be called at all, though the engine would receive the exact value.

The parity motive is explained once, in `_game_call_params_schema`'s docstring; the agent-facing surfaces (the validator message, the `args` description, ADR-0041's bullet, the catalog) carry only the wire-precision claim, and none of them mentions that the float domain above 2^53 is closed.

Either remedy closes it:

- **Keep the parity refusal, correct the record:** say in the message, the `args` description and the ADR that JSON cannot distinguish an integer from an integral float, so publishing the bound in the machine contract requires refusing both — and that a value which is already a double is refused for contract parity, not because it would change. Add a test pinning the deliberate float case, so the cost is a decision rather than a side effect.
- **Or refuse only what actually changes:** bound on `float(value) != value` (true only for Python ints outside the exactly-representable set), and disclose that the published schema is then deliberately STRICTER than the runtime — a conservative direction, since a schema-driven caller would send less than allowed, never more.

I have no preference between them; what the artifacts cannot keep is the current combination of a false justification and an undisclosed closed range. Everything else at this head is closed from my side.

qinhaihong-red:
## Fourth re-review residue — evaluated and handled

Reviewed and fixed at [`c570a8f4`](https://github.com/aigengame/godot-agent/commit/c570a8f4f2d2d5221082a566ea194742d882e983).

| Review point | Decision | Resolution / evidence |
| --- | --- | --- |
| Integer-valued floats above `2^53 - 1` were rejected together with oversized integer tokens. | **Fully adopt.** The validator conflated Python `int` with integral-valued Python `float`. | The safe-integer guard now applies only to values decoded as Python `int` (with bool handled separately). `1e17`, `2.5e17`, and `1e300` are accepted through argv and pure `--params-json`; a Godot 4.6.3 e2e calls an allowlisted typed-float method and confirms all three values. Direct Python↔Godot binary64 probes also matched their exact bytes. Oversized integer tokens such as `9007199254740993` remain rejected before the wire. |
| The previous error rationale claimed that any larger integral value would arrive changed. | **Fully adopt.** That statement was false for representable high-range floats. | Runtime text, rendered help, ADR-0041, command catalog, and bundled Skill now describe the two reproduced failures only: non-finite numbers and oversized JSON integer tokens. They explicitly state that finite floats do not inherit the integer-token bound. |
| Extending the same bound into Draft 2020-12 schema created apparent alignment but changed the public float contract. | **Correct the alignment mechanism.** Standard JSON Schema validates mathematical numeric values and cannot distinguish `1e17` from an equal integer token, so exact lexical-kind parity is not expressible here without introducing a tagged wire format. | The recursive schema's `number` branch is broad again and discloses that the shared params model is the execution authority for the integer-token bound. Regression tests intentionally pin the schema's unavoidable over-acceptance of an oversized integer while the params model rejects it. No new tagging/protocol mechanism was added. |

Validation at `c570a8f4`:

- focused schema/model/help/contract suite: **160 passed**;
- full fast/headless selection: **1899 passed, 2 skipped, 539 deselected**;
- full real-engine `tests/test_e2e_game.py` on Godot 4.6.3: **9 passed**;
- ruff check/format, pyright, and `git diff --check`: clean;
- independent final review: no actionable findings;
- all applicable GitHub checks: passed. The GitHub Godot e2e job is skipped by host policy and is not counted as runtime evidence.

Adjacent boundary observed during the probe: Godot's JSON parser maps a subnormal such as `5e-324` to zero. This behavior was already present at the reviewed parent `d7665623`, is not widened by this fix, and is why the updated artifacts deliberately do **not** claim preservation of the entire finite binary64 range. It is outside this high-range-float residue.

qinhaihong-red:
## Recheck at `c570a8f4` — my P2 is closed and verified; one new P2 from the same probe area

Rechecked in a detached worktree, independently of the commit message.

### The P2 is closed, and the fix is the stronger of the two options

- **The over-refusal is gone and the values genuinely survive.** A differential sweep (17 corpus entries) shows `1e17`, `2.5e17`, `±1e300`, `2.5`, `3.0` and nested `{"deep": [1e300]}` accepted by BOTH the model and the published schema. A real-engine round trip through `echo_value` returns `1e+17`, `2.5e+17`, `±1e+300` and the boundary `9007199254740991` **exactly**; `9007199254740993` is refused with the corrected message ("a larger **integer** may arrive as a DIFFERENT value"). Engine error stream stayed empty.
- **The justification is now true.** `MAX_EXACT_JSON_INT`'s comment, the validator message, the `args` description, help, the ADR, the catalog and the Skill all say the bound applies to values decoded as Python `int`, and that a float already is binary64. Nothing claims the bound is a published machine constraint any more.
- **The remaining schema/model divergence is the unavoidable one, and it is pinned.** With the integer bound removed from the schema, an oversized integer is schema-accepted / model-rejected; the corpus asserts exactly that pair with a comment naming it deliberate, and the `args` description carries the reason. That is the direction I would have chosen: a schema-driven caller sends less than allowed, never more.
- Gates at this head: fast suite `1901 passed, 539 deselected` on this display-capable host (the body's headless `1899 passed, 2 skipped` is the same 1901 selected); real-engine `tests/test_e2e_game.py` **9 passed**; pyright 0; ruff check + format clean; `git diff --check` clean; GitHub checks 8 success / 4 skipped, CLEAN.

### Changes required (P2) — the disclosed float residue is understated: ordinary normal doubles arrive as `0.0`

The PR body scopes the known gap to "subnormal values such as `5e-324`". Probing the same area against Godot 4.6.3 shows the flattening is **not** limited to subnormals or to the exponent floor — it follows the literal's significant-digit count:

```text
JSON.parse_string  1e-300                     -> preserved   (is_zero=false)
                   1e-308                     -> preserved   (is_zero=false)
                   1.2345678901234567e-300    -> ZERO        (is_zero=true)   <-- ordinary normal double
                   2.2250738585072014e-308    -> ZERO        (is_zero=true)   <-- DBL_MIN, not subnormal
                   1e-320, 5e-324             -> ZERO
```

`1.2345678901234567e-300` is exactly the form Python's `json.dumps` emits for a full-precision small float, so a caller passing an ordinary computed value in that range gets a **successful call on `0.0`** — the failure mode the integer bound exists to prevent, reached without any exotic input. Two artifacts read as promising otherwise: the catalog's "Finite floats already are binary64 and do not inherit that integer bound" and the body's line naming only subnormals.

Remedies, in the order I would consider them:

1. **Correct the disclosure** to the reproduced class — small-magnitude floats whose shortest representation carries many significant digits — with these counterexamples, and file the follow-up issue; the root cause is Godot's JSON number parser, which is the engine's to fix, not this slice's.
2. **Optionally add the conservative CLI-side guard** that keeps the "never succeed on a value the caller never sent" promise symmetric: refuse a non-zero float whose decimal form places `exponent − (significant digits − 1)` below the binary64 floor. It over-refuses slightly near the floor, and it is a guard rather than a mirror of engine internals.

### One P3 note

The finiteness rule is the second inexpressible divergence (the schema's `number` branch accepts `NaN`/`Infinity` in-memory; the model refuses them), and the disclosed-divergence discipline you applied to the integer row — and to `script run`'s truth table in #748 — would pin it with one more corpus row.

Separately and out of scope, recorded so it is not lost: the RESULT path loses tiny floats too (`JSON.stringify(1e-300)` → `"0.0"`), which affects any live read returning a small float, not just `game call`.

qinhaihong-red:
## Fifth re-review handled — `a6d97071`

I reproduced the new finding on Godot 4.6.3, evaluated the proposed mechanisms separately, and pushed [`a6d97071`](https://github.com/aigengame/godot-agent/commit/a6d97071b515456eaedc8a215c16a1d0dd757b92).

| Claim | Decision | Evidence and resolution |
| --- | --- | --- |
| The PR body understated the known float residue as subnormal-only. | **Fully adopt.** | Direct `JSON.parse_string` probes confirmed that `1e-300` and `1e-308` remain non-zero, while the ordinary normal value `1.2345678901234567e-300`, `DBL_MIN`, and subnormal values can become `0.0`. The machine-schema descriptions, rendered help, ADR-0041, command catalog, bundled Skill, PR body, and #673 now state the reproduced normal/subnormal class and avoid a whole-binary64 preservation claim. |
| The result direction can lose small floats too. | **Fully adopt as a tracked cross-operation defect.** | Godot 4.6.3 `JSON.stringify` can serialize a non-zero small value such as `1e-300` as `0.0`; every live-result envelope uses this path. [Issue #752](https://github.com/aigengame/godot-agent/issues/752) records both request and result directions, engine-source evidence (`JSON::_get_token` → `String::to_float`), a differential corpus, candidate protocol policies, and acceptance criteria. |
| Add an exponent/significant-digit CLI guard now. | **Do not adopt.** | The suggested threshold is a heuristic rather than a mirror of Godot behavior. It can over-refuse preserved inputs, covers only requests, and leaves the shared result serializer unchanged. Embedding that partial engine-version rule in the params model would add entropy without closing the demonstrated defect. #752 owns an evidence-backed two-direction policy instead. |
| Pin the NaN/Infinity schema/model divergence. | **Fully adopt.** | The corpus now asserts that the current Draft 2020-12 Python validator accepts in-memory `NaN`/`±Infinity` extensions while `GameCallParams` rejects all three. Schema descriptions disclose that RFC JSON excludes them and identify the params model as the execution authority. Existing argv and pure `--params-json` regressions still prove that they never enter the live channel. |

Validation at `a6d97071`:

- focused schema/model/help/contract suite: **160 passed**;
- full fast/headless selection: **1899 passed, 2 skipped, 539 deselected**;
- full real-engine `tests/test_e2e_game.py` on Godot 4.6.3: **9 passed**;
- pyright: **0 errors, 0 warnings**; ruff check and 468-file format check: clean; `git diff --check`: clean;
- independent final review: approved with no repository finding after reserving the PR-body synchronization completed here;
- all applicable GitHub checks: passed. The GitHub Godot e2e job is skipped by host policy and is not counted as runtime evidence.

qinhaihong-red:
## Recheck at `a6d97071` — both findings closed; no findings remain from my side

Rechecked in a detached worktree against the reproduced evidence rather than the response text.

**The understated float residue is corrected, everywhere I flagged it and in the same words the probes support.** The schema's `number` description, the rendered help, the `args` description, ADR-0041, the command catalog, the bundled Skill, the `MAX_EXACT_JSON_INT` comment, the `_reject_unrepresentable` docstring, the PR body and issue #673 all now name the reproduced class — small-magnitude *normal* values (`1.2345678901234567e-300`), `DBL_MIN`, and subnormals — and none of them still reads as a whole-binary64 preservation promise. The high-range claim is correspondingly narrowed to "the reproduced values `1e17`, `2.5e17`, `1e300`", which is exactly what the e2e pins.

**Declining the CLI heuristic is the right call, and I can confirm the stated reason rather than just accept it.** My own proposed rule (`exponent − (significant digits − 1) < −307`) would refuse `1e-308`, which the engine preserves — so it does over-refuse valid input, exactly as the response says. Add that it covers only the request direction while `JSON.stringify` loses small results too, and a guard in the params model would have bought partial coverage at the cost of a second, drifting authority.

**Issue #752 is a real owner, not a parking space.** It carries the reproduced table verbatim, the engine-source root path (`JSON::_get_token` → `String::to_float` → `built_in_strtod`), both directions, candidate policies, acceptance criteria, and an explicit "do not choose a decimal heuristic without a conformance corpus" note. That is the follow-up I would have written.

**The P3 is adopted and slightly exceeded**: the corpus now pins `nan`, `inf` AND `-inf` as schema-accept / model-reject, the ADR records it as the second deliberate over-acceptance, and both the `args` and `number`-branch descriptions are asserted — including the `#752` pointer, so the disclosure cannot silently disappear.

Validation at this head, independently run: fast suite `1901 passed, 539 deselected` on this display-capable host — the same 1901 selected as the body's headless `1899 passed, 2 skipped`; real-engine `tests/test_e2e_game.py` **9 passed** on Godot 4.6.3; pyright 0; ruff check and format clean; `git diff --check` clean; GitHub checks 8 success / 4 environment-skipped; `CLEAN` / `MERGEABLE`.

Across the rounds I rechecked (`d7665623` → `c570a8f4` → `a6d97071`), everything I raised is closed. I have no remaining findings.

Issue #698, opened 2026-08-17
Script-error parser: the sentence-period strip misreads dot-terminated res:// paths


qinhaihong-red:
## Review-round source

Found during PR #693's review (wave 2). The script-error parser (`gda.script_errors`) strips a trailing period from paths it extracts out of engine sentences (the period is usually sentence punctuation). For a path that genuinely ends in a dot the strip corrupts the address: the engine's "Can't load script: res://.." parses to `res://.`, which then fails to match the canonical entry `res://..` in the verdict logic — on the base branch this produced an exit-0 phantom success for `gda script run res://..`. #693 closed the ENTRY route (all root/escape spellings are refused pre-launch), but the parser weakness itself remains for its other consumers: dependency/diagnostic paths in `script run` results, and the timeout (#655) and scene-preflight (#664) reuse of the same parser — both **now merged**, so those consumers are live, not future.

**Which module.** The defect is in `gda.script_errors` (the path-extracting regexes at `script_errors.py:92` and `:137`), NOT in `gda.engine_log`. The codebase keeps these deliberately separate — `engine_log` is "the single home of *how* the engine formats an error line" (`engine_log.py:1-8`) and extracts no paths at all. The daemon's Session-log diagnostics (`src/gda/daemon/diag.py`) import only `gda.engine_log.parse_errors`, so that channel is **unaffected** by this bug; it was wrongly listed as a consumer here.

Still live on `main` (2026-08-27):

```
>>> _CANT_LOAD.match("Can't load script: res://..").group("path")
'res://.'
>>> _FAILED_LOADING_RESOURCE.match("Failed loading resource: res://weird..").group("path")
'res://weird.'
```

> **Correction (2026-08-27, during implementation).** The second line above is a badly
> constructed example and does NOT demonstrate a defect. The engine emits
> `"Failed loading resource: %s."` — always one appended period — so
> `Failed loading resource: res://weird..` is the emission for the path `res://weird.`,
> and extracting `res://weird.` is CORRECT. Checked across path shapes against what the
> engine actually prints, the old lazy `\S+?\.?$` never corrupts a well-formed
> `Failed loading resource:` line: the smallest valid split always lands on the
> guaranteed trailing period. **Only `_CANT_LOAD` carries a live corruption bug**, because
> its sentence has no appended period to absorb the strip.
>
> The mandatory-strip change below is still correct and still worth making — it is a
> fail-closed improvement (a line missing its guaranteed period is now rejected instead of
[...]

PR #754, opened 2026-08-27
refactor(gda): single-evaluate script/shader set mode via params model

### LINKED ISSUE #713: script set / shader set: drop the manual argv pre-check in favor of the params model
## Parent

Follow-up flagged during PR #710's external review round (issue #663).

## What to build

`script set` and `shader set` both run their mode-selection rule twice on the argv route: the command calls `parse_set_mode_argv()` by hand and translates the failure, then the params model's validator runs `resolve_set_mode()` again. PR #710 removed exactly this pattern from `script validate` by constructing the params model through the shared `params_or_bad_parameter()` translation path (ADR-0015: the params model is the input-rule authority), so future model rules surface identically on argv and `--params-json`.

Apply the same unification to both `set` commands. `shader set` imports the helper from the script group, so one change covers both — verify the two surfaces stay in lockstep. Refusal wording may shift to the model-rule form; existing message pins must be updated deliberately and the change stated in the PR.

## Acceptance criteria

- [ ] The mode-selection rule runs once per invocation, owned by the params model, on both commands.
- [ ] argv and `--params-json` refuse the same selections with the same error class (usage error exit 2 / structured `invalid_params`).
- [ ] No behavioral change to accepted invocations (schema/help byte-diff or stated deltas).
- [ ] `parse_set_mode_argv` (`script.py:2365`) is dealt with explicitly. Its only two
      callers are the ones this slice removes (`script.py:2347`, `shader.py:419`), so it
      becomes dead on completion — either delete it, or keep it with a docstring that
      states its remaining role honestly. Do not leave it carrying a description of a job
      it no longer has.

## Verified premise (2026-08-27)

The double-evaluation is exactly as described: `script.py:2347` calls
`parse_set_mode_argv`, whose own docstring already points at `resolve_set_mode` as the
single rule owner (`script.py:2374`), and the params model then calls `resolve_set_mode`
again in `__init__` (`script.py:374`). `shader.py` imports BOTH helpers from the script
group (`shader.py:20`) and repeats the pair at `:419` and `:210` — so the one change does
cover both surfaces, as the body says.

## Blocked by

- ~~PR #710 landing on the dev branch~~ — **cleared**: #710 is merged and
  `params_or_bad_parameter()` is on `main`.

## Scope amendments (2026-08-28 review closure)

These amendments record two coordinator-approved scope decisions discovered during implementation and adversarial review. They are part of this issue's accepted specification rather than untracked PR-only rationale.

### Shared params-model refusal renderer

Moving the two argv commands onto the params models exposed pydantic's raw `ValidationError` dump, including caller-provided field values. Repairing the argv path is inseparable from this issue because the regression was introduced by the required routing change. The identical defect also existed on the agent-facing `--params-json` path. This issue therefore explicitly includes one shared `validation_error_message()` seam in `gda.errors` for both ADR-0015 model-construction channels: argv through `params_or_bad_parameter()`, and structured invocation through `headless.invoke()`.

This deliberately changes refusal prose for other commands that use either shared path, without changing accepted inputs, failure taxonomy, exit codes, or structured envelope shape. The separate `perf --budget-file` producer is not an ADR-0015 params-model construction path and remains excluded.

### Merge-safe rationale maintenance for #698 / PR #756

In the concurrent #698/#713 wave, `src/gda/commands/script.py` is owned by the #713 slice. PR #756 fixes the parser behavior that made one `_project_scoped_res_path` rationale passage stale. To avoid a same-file collision, the merge-order-safe rationale edit is included in #713 rather than PR #756. This is documentation maintenance only; it does not alter #698's parser behavior or expand #713's runtime behavior.

qinhaihong-red:
## Summary

`script set` and `shader set` both ran the edit-mode selection rule twice on
the argv route: the CLI body called `parse_set_mode_argv()` by hand and
translated its failure, then the params model's validator called
`resolve_set_mode()` again in `__init__` — `resolve_set_mode`'s own docstring
already named itself the single rule owner, so the second evaluation
contradicted the code's stated design. `shader set` imports both helpers from
the script group and repeats the same pair.

PR #710 removed exactly this pattern from `script validate` by constructing
the params model through the shared `params_or_bad_parameter()` translation
path (ADR-0015: the params model is the input-rule authority). This PR mirrors
that precedent for both `set` commands — no new translation path invented.

- `gda/commands/script.py`'s `set_script` now builds `ScriptSetParams` through
  `params_or_bad_parameter(ScriptSetParams, ...)` instead of calling
  `parse_set_mode_argv()` and passing a pre-resolved `mode` in.
- `gda/commands/shader.py`'s `set_shader` does the same for `ShaderSetParams`,
  keeping the two surfaces in lockstep through the existing shared seam
  (`shader.py` importing from `script.py`, ADR-0040 §5) rather than
  duplicating the fix.
- `parse_set_mode_argv` is **deleted**: its only two callers were the argv
  bodies this PR changes, so it became dead on completion. `resolve_set_mode`'s
  docstring is updated to name its two remaining callers honestly
  (`ScriptSetParams._resolve_mode`, `ShaderSetParams._resolve_mode`) instead of
  describing a CLI-wrapper caller that no longer exists.

## Behavioral parity

- The mode-selection rule now runs exactly **once** per invocation on both
  commands, on both input paths.
- argv and `--params-json` still refuse the same selections with the same
  error class: usage error / exit 2 for argv, structured `invalid_params` for
  `--params-json`.
- `gda script set --help`, `gda shader set --help`, and `gda --help` are
  **byte-identical** to base (verified after every commit). `gda schema` is
  byte-identical to base through round 3; round 4 (below) makes DELIBERATE
  prose-only changes to it, to fix stale ownership language a reviewer found:
  **three docstrings edited, four schema leaves changed** — `input.description`
  and `input.$defs.ScriptSetMode.description`, under BOTH `script set` and
  `shader set` (the `ScriptSetMode` docstring is one source edit that appears
  under each command's `$defs`, so it lands twice). No other schema entry
  changed — verified by diffing the schema as parsed JSON, not raw text, by the
  implementer and again independently by the orchestrator.

## Follow-up (round 2): `params_or_bad_parameter` leaked caller input into the argv refusal

External review of this PR caught a real regression the first commit introduced.
Unifying `script set`/`shader set` onto `params_or_bad_parameter` changed their
argv refusal text from a clean sentence to the raw `str()` of pydantic's
`ValidationError` — a dump carrying the model's class name, a
`[type=..., input_value=..., input_type=...]` tag per error, and a
`pydantic.dev` URL. Concretely:

Before (base, `script set X --search foo`):
```
Invalid value: 'search' and 'replace' must be used together.
```
After the first commit (the regression):
```
Invalid value: 1 validation error for ScriptSetParams
  Value error, 'search' and 'replace' must be used together. [type=value_error,
input_value={'path': 'X', ...}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/value_error
```
Worse than noise: `input_value=` echoes the caller's OTHER field values back into
the refusal, including large/sensitive ones — `script set X --search foo --content
'SECRET_MARKER_...'` put the secret marker inside the printed refusal.

Root cause was `params_or_bad_parameter` itself (`src/gda/dispatch.py`), the
shared argv seam behind `script set`/`shader set`/`script validate` and every
other params-model construction on the argv path — not this PR's call sites, so
the fix belongs there (`src/gda/dispatch.py` has no owner this wave). Commit
`ae15b38b` fixes it at the seam: instead of `str(exc)` on the whole
`ValidationError`, it reads each error's own `msg` (already clean pydantic text
for a built-in check) or, for a validator's raised `ValueError`
(`resolve_set_mode`, `check_validate_selection`, etc.), the original exception
out of `ctx['error']` — pydantic's own `msg` re-prefixes that with `"Value error,
"`, which this also avoids. Multiple errors join on `"; "`, each tagged with its
dotted field path when field-scoped.

This restored `script set`/`shader set`/`script validate`'s argv wording to the
base clean form and, as the shared seam, also cleaned up the same dump on every
OTHER command's argv usage error (node/input/project/scene/perf/resource/daemon/
game) — grepped `tests/` for `pydantic.dev`, `"validation error for"`,
`input_value=`, `[type=`: zero hits, so no test pinned the old dump text
anywhere in the suite.

## Follow-up (round 3): `--params-json` had the SAME defect, one layer down

Round 2 fixed only the argv (human) channel, on the reviewer's explicit
"dispatch.py:51-54" pointer. A further review measured the OTHER channel:

```
gda script set --params-json '{"path":"...","search":"foo","content":"SECRET_MARKER_..._END"}'
{"error":{"category":"operation","code":"invalid_params","message":"--params-json is not a
valid params object: 1 validation error for ScriptSetParams\n  Value error, 'search' and
'replace' must be used together. [type=value_error, input_value={'path': '/private/tmp/cl
..._MARKER_..._END'}, input_type=dict]\n    For further information visit
https://errors.pydantic.dev/2.13/v/value_error","diagnostics":""}}
```

The structured envelope's OWN `message` field — the one an agent reads — carried
the identical dump, embedded newlines, docs URL, and the caller's `--content`
echoed back inside `input_value=`. After round 2 the human channel was clean and
the machine channel was not, backwards for an agent-facing CLI; #713's AC 2
pairs the two channels on error class, and same class with wildly different
message quality doesn't honour that.

Commit `33630bc0` fixes it by REUSING round 2's extractor rather than
duplicating it: `validation_error_message()` moved out of `gda.dispatch` into
`gda.errors` — a home below BOTH channels (`gda.dispatch` imports
`gda.headless`, so the reverse import would cycle; both already import
`gda.errors` for their own failure taxonomy) — and `headless.py`'s
`--params-json` `ValidationError` handling (`invoke()`) now calls the same
function instead of `str(exc)`. The `"--params-json is not a valid params
object: "` prefix is kept (useful context, not the defect); only the detail
behind it changes.

**Scope of the parity claim.** The two ADR-0015 params-model channels —
argv (`params_or_bad_parameter`) and `--params-json` (`headless.py`'s
`invoke()`) — now report the byte-identical sentence for the identical
model-construction refusal, on `script set`/`shader set`/`script validate` and
every other command built through either path. That is what was verified; it
is NOT a claim that `invalid_params` speaks one sentence everywhere in `gda` —
see "Not touched" below for a known sibling producer outside this scope. For
`script set X --search foo` (no `--replace`):

| channel | rendering |
|---|---|
| argv | `Invalid value: 'search' and 'replace' must be used together.` |
| `--params-json` | `{"error":{"code":"invalid_params","message":"--params-json is not a valid params object: 'search' and 'replace' must be used together.", ...}}` |

Same for the secret-marker repro above: with the fix, `--params-json` now
returns `"--params-json is not a valid params object: 'search' and 'replace'
must be used together."` — no dump, no secret.

`tests/test_dispatch.py` now covers both channels: the original 5 unit tests
against `params_or_bad_parameter` (plain `ValueError`, single-error
`ValidationError` model-level and field-level, multi-error join, the
secret-leak repro), plus 3 more added in round 3 — the `--params-json` channel
pinned clean end-to-end through an actual `script set` invocation, the same
secret-leak repro over `--params-json`, and a parity test asserting the two
channels' extracted sentences are `==`. **Red-proofed both rounds** the same
way: commit the fix, `git checkout <prior-commit> -- <the changed source
files>` (never the tests), observe the new tests fail with the pre-fix dump
text, `git checkout HEAD --` to restore, confirm `git status` is clean and all
tests pass again.

## Follow-up: stale rationale for a sibling slice's parser fix

Sibling slice #698 (`fix/698-script-error-parser-dots`) fixes
`gda.script_errors`' `_CANT_LOAD` regex to stop stripping a trailing period —
the engine never appends one to `Can't load script: <path>`, so the strip was
always wrong. That makes a rationale comment in `script.py`'s
`_project_scoped_res_path` (this PR's owned file) stale once both land: it
justified refusing a root address (`res://.` / `res://..`) partly as
preventing a phantom success caused by the (buggy) parser stripping the
address's own trailing dot. Commit `c3bf1d93` reframed that passage as
history; round 4 below corrects it again — that framing turned out to be
merge-order-fragile.

Routing note: this fix landed here, not in #698's own PR (#756), on the
coordinator's explicit routing call — `commands/script.py` is owned by this
wave under the file-ownership rule for this round of slices, so #756 (which
fixed the underlying parser) was kept out of a file it does not own, and the
now-stale comment was routed to this PR instead. Recorded for the review
record: a routing decision, not scope added on this implementer's own
initiative.

## Follow-up (round 4): stale ownership prose, a merge-order-fragile comment, and test dedup

A further external adversarial review at head `33630bc0` found seven items
(commit `08b74bc5` addresses five; two are declined with reasons):

1. **[P2] Stale ownership prose (fully adopted).** Three docstrings still
   described the pre-#713 world — `script.py`'s `ScriptSetMode` enum and
   `ScriptSetParams` class, `shader.py`'s `ShaderSetParams` class — all said
   "the CLI resolves ... and stamps it on `mode`", when this PR made the
   params model the sole owner of that derivation. Left as-is, a maintainer
   reading them could restore a CLI-side check and reintroduce the double
   evaluation #713 removed. Reworded all three (plus the same claim in
   `docs/command-catalog.md:466`, "modes is selected at the CLI") so the CLI
   reads as a thin argv-to-model adapter and the params model as the
   deriving/stamping owner.

   These three docstrings ARE pydantic model/enum descriptions, so the wording
   change flows into `gda schema`'s `input.description` / `$defs` for
   `script set` and `shader set` — verified by diffing the schema as PARSED
   JSON (not raw text): exactly those two commands' entries differ from base,
   nothing else. `--help` for both commands, and `gda --help`, remain
   byte-identical to base — only `gda schema`'s prose changed, in only the
   three docstrings just listed.

2. **[P2] The `#698` comment asserted a fix not in this branch (partially
   adopted).** `_project_scoped_res_path`'s docstring said "#698 fixed the
   parser itself... so that failure mode is now history", while at the
   reviewed head `src/gda/script_errors.py:92` still holds the OLD regex —
   #698's fix is PR #756, still open. **Adopted:** reworded the paragraph to
   hold regardless of merge order — it now describes the risk hypothetically
   ("a parser that folds it as though it were punctuation WOULD miss...") and
   states explicitly that this guard's own pre-launch refusal already closes
   the gap on its own, independent of whichever PR lands first; #698/#756 are
   named without asserting either PR's landed state. **Not adopted:** moving
   the commit to #756. That routing was the coordinator's own call, not this
   implementer's — `commands/script.py` is this wave's owned file, so routing
   the fix here (rather than letting #756 touch a file it doesn't own) is what
   the file-ownership rule exists for; moving it now would recreate exactly
   the collision that rule prevents (see "Routing note" above).

3. **[P2] Local pytest count reported as if it were CI (fully adopted).** See
   Validation below — local and CI results are now labeled and quoted
   separately, with the two CI-only skips identified.

4. **[P3] Duplicate Rich/ANSI usage-error normalization (adopted).**
   `tests/test_dispatch.py`'s `_argv_usage_error_message` re-implemented the
   same ANSI/box-drawing-border stripping `tests/test_screen_commands.py`'s
   `_usage_error_message` already had. Moved the shared normalization into
   `tests/support.py` as `usage_error_text` (beside the existing `plain_text`);
   both call sites now delegate to it.

5. **[P3] Unused `monkeypatch` parameter (adopted).** Dropped from
   `test_no_caller_value_leaks_into_the_refusal` — the test never used it.

6. **[P3] Remove `_RawValueErrorModel` as speculative (NOT adopted).** The
   branch it covers is real: `params_or_bad_parameter` catches a bare
   `ValueError` SEPARATELY from `ValidationError` (`src/gda/dispatch.py`),
   and something has to pin what that clause does. Deleting the test would
   not delete the branch — it would leave it unverified, which is worse. The
   test's own docstring already answers the `# type: ignore` objection:
   "pydantic itself always wraps a validator's ValueError into
   ValidationError... this simulates the one caller shape pydantic can never
   produce" — no pydantic model can exercise this branch, so the stand-in is
   deliberate, not an oversight.

7. **["Move the shared-renderer message change to its own scoped change"]
   (NOT adopted).** Reverting `headless.py`/`dispatch.py`'s shared rendering
   to the prior global behavior would restore a message that dumps pydantic
   internals and echoes the caller's own `--content` back — that is the exact
   leak measured before this fix was requested (rounds 2/3 above), and is a
   regression this PR will not reintroduce. The transparency half of the
   concern is already met: the "This restored..." paragraph in round 2 and
   the "Scope of the parity claim" paragraph in round 3 (above) each name
   every other affected command and state precisely what was verified, so the
   broader effect is disclosed, not hidden.

## Follow-up (round 5): three P3 residues from re-review

A further re-review at head `08b74bc5` closed the round-4 findings (stale
ownership prose, validation evidence, the merge-order-fragile comment,
duplicate usage-error normalization, unused fixture) and left three P3
judgement calls. New commit: `9c7a7a68`.

1. **[P3] `_RawValueErrorModel` bypasses the declared contract (partially
   adopted).** The reviewer is right that a non-`BaseModel` stand-in needing
   `# type: ignore[arg-type]` to reach `params_or_bad_parameter`'s
   `type[P]` (`P` bound to `BaseModel`) is a real contract bypass, not just
   an odd test double — round 4's "no pydantic model can exercise this
   branch" claim was wrong. It IS reachable from a genuine `BaseModel`
   subclass:

   ```python
   class M(BaseModel):
       a: int = 1
       def __init__(self, **kw):
           raise ValueError("a plain refusal sentence.")
   M(a=1)   # -> propagates a BARE ValueError; issubclass(M, BaseModel) is True
   ```

   (`model_post_init` does NOT work — pydantic wraps that hook's raise as
   `ValidationError` too, same as a field/model validator. Overriding
   `__init__` is the one route that runs before pydantic's own validation
   machinery gets a chance to wrap anything, so the plain `ValueError`
   propagates untouched.)

   So: **not deleted.** `_RawValueErrorModel` now subclasses `BaseModel` and
   raises from an overridden `__init__` instead of faking the shape from
   outside the contract — the `# type: ignore[arg-type]` is gone, the branch
   stays verified, and the docstring is rewritten to state the verified
   in-contract route instead of the "one shape pydantic can never produce"
   framing the evidence above contradicts. Deleting the branch instead would
   turn a clean usage error into an escaping traceback for any future model
   that raises from `__init__` — a safety regression for no gain, since the
   branch is real and in-contract.

2. **[P3] Package the shared-renderer cleanup separately (NOT adopted).**
   The reviewer clarified the actual concern: packaging, not reverting — the
   round-4 reply answered the wrong question by treating the two as
   equivalent. Answering the real concern:

   - The **argv half is not separable**: this PR itself introduced the
     wording regression (round 2) by routing `script set`/`shader set`
     through the params model, so fixing it here is repair of this PR's own
     damage, not an unrelated cleanup riding along.
   - The **`--params-json` half** is an explicit specification amendment
     recorded in [issue #713](https://github.com/aigengame/godot-agent/issues/713):
     shipping a state where the human channel is clean and the machine
     channel — the one an agent actually reads — still dumps pydantic
     internals was judged worse than the wider diff, for an agent-facing CLI.
   - Splitting now costs a new PR, a rebase, and a full re-verification of an
     already twice-reviewed change; the maintainer can still choose to land
     it separately at merge time if they weigh that trade-off differently.

   No code restructured for this item.

3. **[P3] The `#698` rationale edit sits outside issue #713 (fully
   adopted).** The routing decision — `commands/script.py` is this slice's
   owned file, PR #756 fixed the parser that made the old comment false, and
   the file-ownership rule kept #756 out of a file it does not own — is now
   recorded as a stable public specification amendment in
   [issue #713](https://github.com/aigengame/godot-agent/issues/713).
   No runtime code change.

## Validation

**Local** (this machine, `uv run pytest -q -m "not e2e"`, head `9c7a7a68`):
`1909 passed, 539 deselected in 183.41s (0:03:03)` (1901 from the base suite +
8 in `tests/test_dispatch.py`; the wall-clock time is slower than round 4's
105.49s from CPU contention with an unrelated concurrent test run on this
machine, not a behavior change).

**CI**: exact-head run for `33630bc0`
([33079460351](https://github.com/aigengame/godot-agent/actions/runs/33079460351),
`Python unit tests and package build` job) reported `1907 passed, 2 skipped,
539 deselected in 75.96s (0:01:15)`. The two CI-only skips are the two
parametrized cases of
`tests/test_display_probe.py::test_a_real_sandbox_denial_is_classified_as_a_permission_problem`
— each calls `pytest.skip("needs macOS with a working sandbox-exec")` (or the
on-console-session check right after) when `sandbox-exec` or a real desktop
session isn't available, which the GitHub Actions Linux runner never has and
this local macOS dev machine does; confirmed locally (`pytest -k
test_a_real_sandbox_denial... -v` → `2 passed`, not skipped). `1907 + 2 =
1909` reconciles the two counts exactly. Round 4's head `08b74bc5` also went
green on every CI job (run
[33145717067](https://github.com/aigengame/godot-agent/actions/runs/33145717067)).
CI has not yet run against round 5's head `9c7a7a68` at the time of writing.

- `uv run ruff check .` → `All checks passed!`
- `uv run ruff format --check .` → `473 files already formatted`
- `uv run pyright` → `0 errors, 0 warnings, 0 informations`
- `gda script set --help` / `gda shader set --help` / `gda --help` byte-diffed
  against base after every commit: no diff, every time. `gda schema`
  byte-diffed against base after every commit through round 3 (no diff);
  round 4's three intentional deltas are enumerated above and were confirmed
  as the ONLY schema change by diffing the parsed JSON. Round 5 touched only
  `tests/test_dispatch.py` (no `src/` file), so this was re-diffed and
  reconfirmed unchanged — still exactly those four leaves.
- Godot e2e not run for this slice (not needed — no engine-facing behavior
  changed; confirmed via the schema/help evidence and the full unit/
  integration suite above).

Two independent adversarial reviews (code + architecture) passed with no
P1/P2 as of head `33630bc0`. Architecture confirmed the `errors.py` seam
placement against ADR-0040's dependency chain; code probed
`validation_error_message` against pydantic 2.13.4 across nested models,
list-index `loc`s, unions, built-in type errors, `AssertionError`, and a
`PydanticCustomError` with no `ctx['error']` — no crash, no information loss.
A further external adversarial review at that same head produced round 4's
seven findings above; a round-5 re-review at head `08b74bc5` found the three
P3 residues above, closed by commit `9c7a7a68`.

## Not touched (flagged, out of scope)

`gda.errors.classify_run`'s `contract_violation` failure
(`f"structured-output contract violated: {exc}"`) has the same `str(exc)`
shape, but it stringifies the ENGINE's own output-schema mismatch, not
caller-supplied input — a different sensitivity profile and outside what was
asked. Left as-is; noting it here rather than silently leaving it unmentioned.
Both reviews agreed this exclusion is defensible.

`src/gda/commands/perf.py:749-754` is a THIRD `invalid_params` producer, from
another caller-supplied value (`--budget-file`), and it still emits the same
`str(exc)` dump this PR eliminated on the other two channels — found by both
adversarial reviews. It predates this PR, is not an ADR-0015 params-model
construction path, and is out of #713's scope; the coordinator is filing it
separately rather than folding it in here.

Closes #713

## Follow-up (round 6): exact ValueError contract and stable scope authority

Re-review at head `9c7a7a68` found one inaccurate production docstring and two scope decisions whose only claimed authority was the PR/wave record. Commit `264b81d0` and the public issue amendment close all three residues:

1. **Raw `ValueError` route (fully adopted).** `params_or_bad_parameter()` no longer says a raw `ValueError` is a sentence "a validator wrote". The verified in-contract route is a custom `BaseModel.__init__` raising before pydantic wraps the exception. Field/model validators and `model_post_init` are documented as arriving as `ValidationError`.
2. **Shared renderer scope (fully adopted as specification traceability).** [Issue #713](https://github.com/aigengame/godot-agent/issues/713) now explicitly includes the shared ADR-0015 renderer for argv and `--params-json`, states the wider prose-only effect, and preserves the `perf --budget-file` exclusion.
3. **#698 rationale routing (fully adopted as specification traceability).** The same issue amendment records why the merge-safe comment maintenance belongs to this slice without changing #698 runtime behavior.

No renderer code or command behavior changed in this round.

**Round-6 local validation** (head `264b81d0`):

- `pytest -q -p no:cacheprovider tests/test_dispatch.py` → `8 passed`
- `ruff check src/gda/dispatch.py` → passed
- `ruff format --check src/gda/dispatch.py` → passed
- `pyright` → `0 errors, 0 warnings, 0 informations`
- `git diff --check` → clean

### DIFF
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -463,7 +463,9 @@ The lifecycle round-trips: `create` → `list` shows it → `delete` → `list` 
 script as **raw text** — it never compiles or loads the script, so editing one never runs
 project code (the read trust boundary of #30). It edits only a script that exists; a missing
 target is `path_not_found`, never a silent create. Exactly one of three mutually-exclusive
-modes is selected at the CLI (a missing or mixed mode is a usage error, exit 2):
+modes must be selected — derived once by the params model, identically on argv and
+`--params-json` (ADR-0015, #713); a missing or mixed mode is a usage error, exit 2, on argv,
+and structured `invalid_params` on `--params-json`:
 
 - **search-replace** — `--search <old> --replace <new>`: replace **every** literal (not regex)
   occurrence of `<old>` with `<new>`. A search string the source does not contain is refused
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -263,9 +263,13 @@ def resolve_set_mode(
     The single home of the edit-mode rule, shared by ``script set`` and ``shader
     set`` and by BOTH input paths (ADR-0015): exactly one of the three
     mutually-exclusive modes must be supplied. Raises ``ValueError`` on a
-    violation — the CLI wrapper translates it to a usage error (exit 2) for argv,
-    while the params models surface it as the structured ``invalid_params`` for
-    ``--params-json``.
+    violation. Its only callers are the two params models' ``_resolve_mode``
+    ``model_validator``s (:class:`ScriptSetParams` here, :class:`ShaderSetParams`
+    in ``gda.commands.shader``) — so the rule runs exactly ONCE per invocation,
+    on both commands (issue #713). The argv body builds that model through
+    :func:`~gda.dispatch.params_or_bad_parameter`, which turns the raised error
+    into the Click usage error (exit 2); ``--params-json`` builds the same model
+    and surfaces it as the structured ``invalid_params`` instead.
     """
     has_search = search is not None or replace is not None
     has_line_range = start_line is not None or end_line is not None
@@ -2344,12 +2348,16 @@ def set_script(
     project: Optional[str] = project_option(),
 ) -> None:
     """Edit a .gd script via search-replace, line-range, or full overwrite."""
-    mode = parse_set_mode_argv(search, replace, start_line, end_line, content)
+    # The model owns the mode-selection rule and the argv body does not restate
+    # it: the shared builder turns any model-construction failure (resolve_set_mode,
+    # via ScriptSetParams's validator) into the Click usage error (exit 2) — the
+    # same translation an argv-side pre-check used to do by hand — so the rule
+    # runs once per invocation on both input paths (ADR-0015, issue #713).
     dispatch_domain(
         SCRIPT_SET_COMMAND,
-        ScriptSetParams(
+        params_or_bad_parameter(
+            ScriptSetParams,
             path=path,
-            mode=mode,
             search=search,
             replace=replace,
             start_line=start_line,
@@ -2360,27 +2368,6 @@ def set_script(
         godot=godot,
         project=project,
     )
-
-
-def parse_set_mode_argv(
-    search: Optional[str],
-    replace: Optional[str],
-    start_line: Optional[int],
-    end_line: Optional[int],
-    content: Optional[str],
-) -> ScriptSetMode:
-    """Resolve a set command's edit mode for the argv path (issue #133).
-
-    The rule itself lives in :func:`resolve_set_mode` — the single source shared
-    with ``ScriptSetParams`` / ``ShaderSetParams`` (ADR-0015). This thin wrapper
-    translates its ``ValueError`` into a Click usage error (exit 2) so the argv
-    path keeps its usage-error ergonomics, while ``--params-json`` surfaces the
-    same rule as a structured ``invalid_params`` via the model.
-    """
-    try:
-        return resolve_set_mode(search, replace, start_line, end_line, content)
-    except ValueError as exc:
-        raise typer.BadParameter(str(exc)) from exc
 
 
 @_app.command(name="attach", cls=SCRIPT_ATTACH_COMMAND.command_class())
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -1174,17 +1174,24 @@ def _project_scoped_res_path(script: str) -> str | None:
       ``res://`` spellings) — the project is the whole addressable scope, so an
       upward escape names something the ``--project`` contract does not cover.
 
-    The last two are load-bearing for the same reason, and it is not tidiness. The
-    engine answers a root address with ``Can't load script: res://.`` (or
-    ``res://..``), whose address the error parser reads back with the sentence period
-    stripped — ``res://`` for the first, ``res://.`` for the second. Neither matches
-    the entry, so the never-ran verdict misses it and the run reports a PHANTOM
-    SUCCESS. And an escape that RESOLVES (``../outside.gd``) executes a script outside
-    the project entirely, which would widen the Project-code execution surface past
-    ADR-0009's Trusted project — the very consequence the amendment cites for keeping
-    absolute paths refused, so admitting it by the relative spelling would make that
-    reasoning false. Refusing both before the launch closes them without touching the
-    error parser's own res:// handling.
+    The last two are load-bearing, and it is not tidiness. The root-address clause
+    used to ALSO be the fix for a parser bug: the engine answers a root address
+    with ``Can't load script: res://.`` (or ``res://..``), and the pre-#698 error
+    parser read that address back with an assumed sentence period stripped —
+    ``res://`` for the first, ``res://.`` for the second. Neither matched the
+    entry, so the never-ran verdict missed it and the run reported a PHANTOM
+    SUCCESS; refusing the root address here, before any launch, closed that gap.
+    #698 fixed the parser itself to read the address back intact, so that failure
+    mode is now history — the root address stays refused for the plainer reason
+    already given above (it names a directory, not a script), not because the
+    parser still corrupts it. The escape clause is the one still load-bearing for
+    the reason it always was: an escape that RESOLVES (``../outside.gd``)
+    executes a script outside the project entirely, which would widen the
+    Project-code execution surface past ADR-0009's Trusted project — the very
+    consequence the amendment cites for keeping absolute paths refused, so
+    admitting it by the relative spelling would make that reasoning false.
+    Refusing both before the launch keeps this gate, not the error parser, as the
+    one place that decides.
 
     The escape test is on the canonical remainder's first SEGMENT, not a string
     prefix: ``res://..foo.gd`` is a legal file whose name merely starts with two dots,
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -234,10 +234,14 @@ class ScriptDeleteResult(BaseModel):
 class ScriptSetMode(str, Enum):
     """The edit mode of ``gda script set``, the single source of truth (issue #133).
 
-    The CLI resolves exactly one mode from the supplied flags (its mutual-exclusion
-    check) and stamps it here, so the operation dispatches on this explicit
-    discriminator instead of re-inferring the mode from which params are present —
-    the inference precedence can no longer drift from the CLI's exclusivity rule.
+    The params model derives exactly one mode from the supplied fields — via
+    :func:`resolve_set_mode`, run once by the model's own ``_resolve_mode``
+    validator on BOTH the argv and ``--params-json`` paths (ADR-0015, #713) —
+    and stamps it here, so the operation dispatches on this explicit
+    discriminator instead of re-inferring the mode from which params are
+    present. The CLI is a thin argv-to-model adapter; it does not re-derive the
+    mode itself, so the derivation cannot drift from the model's exclusivity
+    rule.
 
     - ``SEARCH_REPLACE`` — ``search``/``replace``: every literal (not regex)
       occurrence of ``search`` is replaced with ``replace``.
@@ -306,9 +310,10 @@ class ScriptSetParams(BaseModel):
     loads the script, so editing one can never run project code (the read trust
     boundary of issue #30). ``path`` addresses the script by its ``res://`` or
     filesystem path. The remaining params carry one of three mutually-exclusive
-    edit modes; the CLI resolves which one and stamps it on ``mode`` (issue #133),
-    so the operation dispatches on that explicit discriminator rather than
-    re-inferring it from which params are present:
+    edit modes; the model derives which one and stamps it on ``mode`` (issue
+    #133, ADR-0015) — the SAME derivation on both the argv and ``--params-json``
+    paths (#713) — so the operation dispatches on that explicit discriminator
+    rather than re-inferring it from which params are present:
 
     - **search-replace** (``mode = search_replace``) — ``search``/``replace`` both
       present: every literal (not regex) occurrence of ``search`` is replaced with
@@ -1175,17 +1180,22 @@ def _project_scoped_res_path(script: str) -> str | None:
       upward escape names something the ``--project`` contract does not cover.
 
     The last two are load-bearing, and it is not tidiness. The root-address clause
-    used to ALSO be the fix for a parser bug: the engine answers a root address
-    with ``Can't load script: res://.`` (or ``res://..``), and the pre-#698 error
-    parser read that address back with an assumed sentence period stripped —
-    ``res://`` for the first, ``res://.`` for the second. Neither matched the
-    entry, so the never-ran verdict missed it and the run reported a PHANTOM
-    SUCCESS; refusing the root address here, before any launch, closed that gap.
-    #698 fixed the parser itself to read the address back intact, so that failure
-    mode is now history — the root address stays refused for the plainer reason
-    already given above (it names a directory, not a script), not because the
-    parser still corrupts it. The escape clause is the one still load-bearing for
-    the reason it always was: an escape that RESOLVES (``../outside.gd``)
+    is ALSO belt-and-suspenders against a parser risk: the engine answers a root
+    address with ``Can't load script: res://.`` (or ``res://..``), a sentence
+    whose trailing dot is part of the ADDRESS, not punctuation. A parser that
+    folds it as though it WERE punctuation — reading ``res://.`` back as
+    ``res://``, ``res://..`` back as ``res://.`` — would miss the launched
+    entry, and the never-ran verdict would report a PHANTOM SUCCESS instead of
+    the refusal it should be. Issue #698 (its fix, PR #756) targets exactly that
+    fold in :mod:`gda.script_errors`'s ``_CANT_LOAD`` regex; this paragraph's own
+    argument does not depend on whether that PR has landed at any point in this
+    branch's history, because THIS guard already closes the gap on its own,
+    independent of the parser's fold either way: a root address is refused
+    HERE, before any launch, so it never reaches the parser at all. The root
+    address therefore stays refused for the plainer reason already given above
+    (it names a directory, not a script). The escape clause is the one still
+    load-bearing for the reason it always was: an escape that RESOLVES
+    (``../outside.gd``)
     executes a script outside the project entirely, which would widen the
     Project-code execution surface past ADR-0009's Trusted project — the very
     consequence the amendment cites for keeping absolute paths refused, so
--- a/src/gda/commands/shader.py
+++ b/src/gda/commands/shader.py
@@ -136,10 +136,11 @@ class ShaderSetParams(BaseModel):
     autoloads at engine startup (ADR-0009). ``path`` addresses the shader by its
     ``res://`` or filesystem path. The remaining params carry the SAME three
     mutually-exclusive edit modes as ``script set`` (issue #118) — the shader
-    group reuses that edit interface rather than re-deriving it. The CLI resolves
-    which mode and stamps it on ``mode`` (issue #133), so the operation dispatches
-    on that explicit discriminator rather than re-inferring it from which params
-    are present:
+    group reuses that edit interface rather than re-deriving it. The model
+    derives which mode and stamps it on ``mode`` (issue #133, ADR-0015) — the
+    SAME derivation on both the argv and ``--params-json`` paths (#713) — so the
+    operation dispatches on that explicit discriminator rather than
+    re-inferring it from which params are present:
 
     - **search-replace** (``mode = search_replace``) — ``search``/``replace`` both
       present: every literal (not regex) occurrence of ``search`` is replaced with
--- a/src/gda/commands/shader.py
+++ b/src/gda/commands/shader.py
@@ -17,8 +17,8 @@ from typing import Optional, Protocol, runtime_checkable
 import typer
 from pydantic import BaseModel, Field, model_validator
 
-from gda.commands.script import ScriptSetMode, parse_set_mode_argv, resolve_set_mode
-from gda.dispatch import dispatch_domain
+from gda.commands.script import ScriptSetMode, resolve_set_mode
+from gda.dispatch import dispatch_domain, params_or_bad_parameter
 from gda.headless import (
     HeadlessCommand,
     godot_option,
@@ -416,12 +416,15 @@ def set_shader(
     """Edit a .gdshader via search-replace, line-range, or full overwrite."""
     # shader set reuses the script set edit-mode interface (issue #115): the same
     # mutual-exclusion resolver decides the single ScriptSetMode discriminator.
-    mode = parse_set_mode_argv(search, replace, start_line, end_line, content)
+    # The model owns that rule and the argv body does not restate it: the shared
+    # builder turns any model-construction failure into the Click usage error
+    # (exit 2), so the rule runs once per invocation on both input paths
+    # (ADR-0015, issue #713).
     dispatch_domain(
         SHADER_SET_COMMAND,
-        ShaderSetParams(
+        params_or_bad_parameter(
+            ShaderSetParams,
             path=path,
-            mode=mode,
             search=search,
             replace=replace,
             start_line=start_line,
--- a/src/gda/dispatch.py
+++ b/src/gda/dispatch.py
@@ -47,11 +47,60 @@ def params_or_bad_parameter(model_cls: type[P], /, **kwargs: Any) -> P:
     usage-error ergonomics — while ``--params-json``, which builds the SAME model
     in the command class, surfaces the same rule as a structured
     ``invalid_params``. Stated here once so no argv body restates it.
+
+    A raw ``ValueError``'s ``str()`` is already the plain sentence a validator
+    wrote, so it passes through as-is. A pydantic ``ValidationError`` is
+    rendered through :func:`_validation_message` instead of its own ``str()`` —
+    which dumps the model's class name, a ``[type=..., input_value=...,
+    input_type=...]`` tag PER ERROR, and a pydantic.dev URL, and can echo back
+    an arbitrary caller value (including a large or sensitive one) inside
+    ``input_value=`` (#713 review).
     """
     try:
         return model_cls(**kwargs)
-    except (ValueError, ValidationError) as exc:
+    except ValidationError as exc:
+        raise typer.BadParameter(_validation_message(exc)) from exc
+    except ValueError as exc:
         raise typer.BadParameter(str(exc)) from exc
+
+
+def _validation_message(exc: ValidationError) -> str:
+    """Render a ``ValidationError`` as the sentence(s) its checks actually wrote.
+
+    Reads each error's own ``msg`` — already the clean, human-readable text for
+    a built-in pydantic check (a type mismatch, a missing field, an out-of-range
+    value) — rather than ``str(exc)``, which additionally dumps the model class
+    name and a ``[type=..., input_value=..., input_type=...]`` tag per error
+    (the defect: ``input_value`` echoes the caller's raw field value, which can
+    be large or sensitive, e.g. a ``script set --content`` payload).
+
+    For a model or field validator's raised ``ValueError`` (pydantic's
+    ``"value_error"`` type, e.g. :func:`~gda.commands.script.resolve_set_mode`),
+    ``msg`` is pydantic's OWN ``"Value error, "``-prefixed rendering of it; this
+    reads the original exception back out of ``ctx['error']`` instead, so the
+    message is exactly the sentence the validator wrote, unprefixed.
+
+    Each message is tagged with its field path (``loc``, dotted) when the error
+    is field-scoped; a model-level validator's error (``loc == ()``, e.g.
+    ``resolve_set_mode``'s mode-selection rule, which has no single field to
+    name) carries no tag. Multiple errors join on ``"; "`` — a command usually
+    raises exactly one, but pydantic can report several at once (e.g. two
+    independently-invalid fields), and nothing here assumes otherwise.
+    """
+    parts: list[str] = []
+    for err in exc.errors():
+        ctx = err.get("ctx")
+        if err["type"] == "value_error" and isinstance(ctx, dict) and "error" in ctx:
+            message = str(ctx["error"])
+        else:
+            message = err["msg"]
+        loc = err.get("loc") or ()
+        if loc:
+            field = ".".join(str(part) for part in loc)
+            parts.append(f"{field}: {message}")
+        else:
+            parts.append(message)
+    return "; ".join(parts)
 
 
 def make_runner(binary: Path, project: Optional[Path]) -> GodotRunner:
--- a/src/gda/dispatch.py
+++ b/src/gda/dispatch.py
@@ -19,6 +19,7 @@ from pydantic import BaseModel, ValidationError
 from gda.errors import (
     Failure,
     invalid_project_failure,
+    validation_error_message,
 )
 from gda.execution import ExecutionKind
 from gda.export_runner import ExportRunner, make_subprocess_export_runner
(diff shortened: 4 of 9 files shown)


qinhaihong-red:
Reviewed PR #754 at exact head `33630bc0b755a132ae438bf2a9f7dc2ccb84ba9c` against merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3` and issue #713. The head was re-pinned after review and did not change.

## Artifact — Changes required

- **[P2] PR body, Validation:** the body reports `1909 passed, 539 deselected`, but exact-head CI run `33079460351` reports `1907 passed, 2 skipped, 539 deselected`. Please report the CI result accurately, or identify `1909 passed` explicitly as a separate local result and explain the CI skips.

Terminology and prose were reviewed against `CONTEXT.md`, ADR-0015, ADR-0040, and ASD-STE100. The complete PR body and issue #713 were inspected; no other substantive terminology or prose issue was found. This PR changes no Markdown files.

## Architecture / DDMA — Changes required

- **[P2] `src/gda/commands/script.py:237-240,308-311`, `src/gda/commands/shader.py:137-142`, `docs/command-catalog.md:462-466`:** these passages still say the CLI selects and stamps the mode. The implementation and ADR-0015 now make the params model the sole owner for argv and `--params-json`. This stale ownership description can reintroduce the duplicate CLI rule. Please describe the CLI as a thin argv-to-model adapter and the params model as the deriving/stamping owner.
- **[P2] `src/gda/commands/script.py:1177-1187`:** this exact head says “#698 fixed the parser” and that the bug “is now history,” but `src/gda/script_errors.py:92` still contains the old optional-period regex. The fix is in separate, still-open PR #756. Please remove this unrelated commit from #754, or make the dependency/order explicit and keep the prose conditional until the fix is in the base.

The implemented dependency direction itself is sound: `shader → script` reuses the script-owned edit interface, both CLI channels build the same params models, and both route through the existing dispatch boundary. Instrumented script/shader × argv/JSON × accepted/refused probes observed exactly one resolver call in all eight cases.

## Standards

No hard documented-standard violation.

- **[P3, judgement call] `tests/test_dispatch.py:45`:** `_argv_usage_error_message` duplicates the Rich/ANSI normalization in `tests/test_screen_commands.py`; use one helper in `tests/support.py`.
- **[P3, judgement call] `tests/test_dispatch.py:104`:** `_RawValueErrorModel` is outside the helper's declared Pydantic-model contract and requires `# type: ignore[arg-type]`; remove this speculative case.
- **[P3, judgement call] `tests/test_dispatch.py:154`:** remove the unused `monkeypatch` fixture parameter.

## Spec

All functional acceptance criteria pass, including exhaustive mode-presence probing, exact-once evaluation, both refusal classes, byte-identical schema/help, and deletion of `parse_set_mode_argv`.

- **[P3] `src/gda/headless.py:433`, `src/gda/dispatch.py:65`:** the shared renderer changes refusal messages for every params-model command, while #713 only requires the set-command unification and error-class parity. Please retain the prior global rendering or move this broader public-message cleanup into its own scoped change.
- **[P3] `src/gda/commands/script.py:1177`:** the #698 parser-rationale rewrite is unrelated to #713 and complicates independent review/cherry-picking. Move it to its owning PR.

Validation: 336 focused tests passed locally; exhaustive Spec probing covered 128 invocations; base/head `gda schema`, both set-command help surfaces, and root help were byte-identical. Exact-head CI is green, with Godot E2E skipped.

qinhaihong-red:
Addressed the round-4 external adversarial review (comment on head `33630bc0b755a132ae438bf2a9f7dc2ccb84ba9c`). New commit: `08b74bc5`.

## [P2] Stale ownership prose — FULLY ADOPTED

`script.py:237-240`, `script.py:308-311`, `shader.py:137-142` all said "the CLI resolves ... and stamps it on `mode`". Reworded all three so the CLI reads as a thin argv-to-model adapter and the params model reads as the deriving/stamping owner (citing `resolve_set_mode`, ADR-0015, #713 directly). Same fix applied to `docs/command-catalog.md:466` ("modes is selected at the CLI" → derived once by the params model, identically on argv and `--params-json`).

These three are pydantic model/enum docstrings, so this DOES change `gda schema`'s `input.description`/`$defs` for `script set` and `shader set`. Verified by diffing the schema as parsed JSON (not raw text): exactly those two commands' entries differ from base, nothing else. `--help` for both commands and `gda --help` stay byte-identical to base.

## [P2] `#698` comment asserted a fix not in this branch — PARTIALLY ADOPTED

Agreed this was a real bug and mine to own (the routing call, addressed below). **Adopted:** reworded `_project_scoped_res_path`'s docstring so the sentence holds in either merge order. It no longer claims `#698`'s fix has landed; it states the risk hypothetically ("a parser that folds it as though it were punctuation WOULD miss the launched entry...") and makes explicit that this guard's own pre-launch refusal already closes the gap on its own, independent of whether PR #756 has merged — the paragraph's argument never depended on the parser's fix, which is now stated rather than implied.

**Not adopted:** moving the commit to #756. That routing is the coordinator's call, not mine: `commands/script.py` is this wave's owned file, so keeping #756 out of a file it doesn't own — and routing the now-stale-comment fix here instead — is exactly what the file-ownership rule is for. Moving it now would recreate the collision the rule exists to prevent. (This is already recorded in the PR body under "Routing note".)

## [P2] Local count reported as if it were CI — FULLY ADOPTED

Pulled the actual CI log for run `33079460351` (exact head `33630bc0`) rather than assume: `1907 passed, 2 skipped, 539 deselected in 75.96s (0:01:15)`. Identified the 2 skips: both parametrized cases of `tests/test_display_probe.py::test_a_real_sandbox_denial_is_classified_as_a_permission_problem`, which `pytest.skip()` when `sandbox-exec`/a real desktop session isn't available — true on the GitHub Actions Linux runner, false on this local macOS dev machine (confirmed: `pytest -k test_a_real_sandbox_denial... -v` → `2 passed` locally, not skipped). `1907 + 2 = 1909` reconciles exactly. The PR body's Validation section now labels local vs. CI separately and states this explanation instead of a bare local number.

## [P3] Duplicate Rich/ANSI normalization — ADOPTED

Moved the shared panel-normalization logic into `tests/support.py` as `usage_error_text` (alongside the existing `plain_text`). `tests/test_screen_commands.py`'s `_usage_error_message` and `tests/test_dispatch.py`'s `_argv_usage_error_message` both delegate to it now instead of each re-implementing ANSI/box-drawing-border stripping.

## [P3] Unused `monkeypatch` — ADOPTED

Dropped from `test_no_caller_value_leaks_into_the_refusal` (`tests/test_dispatch.py`).

## [P3] Remove `_RawValueErrorModel` as speculative — NOT ADOPTED

The branch it covers is real: `params_or_bad_parameter` catches a bare `ValueError` separately from `ValidationError`, and something has to pin what that clause does. Deleting the test wouldn't delete the branch — it would leave it unverified. The test's own docstring already answers the `# type: ignore` objection: "pydantic itself always wraps a validator's ValueError into ValidationError... this simulates the one caller shape pydantic can never produce." No pydantic `BaseModel` can exercise this branch, so the non-pydantic stand-in is deliberate.

## [P3] Move the shared-renderer message change to its own scoped PR — NOT ADOPTED

Reverting to the prior global rendering would restore a message that dumps pydantic internals and echoes the caller's own field values (e.g. `--content`) back into the refusal — that's the exact leak measured before this fix was requested, across two prior review rounds. Not reintroducing it. The transparency concern is already met: the PR body's round-2 "This restored..." paragraph and round-3 "Scope of the parity claim" paragraph both name every other affected command and state precisely what was verified, so the broader effect is disclosed, not silently expanded.

## Validation (this round)

- `uv run pytest -q -m "not e2e"` → `1909 passed, 539 deselected in 105.49s (0:01:45)`
- `uv run ruff check .` → `All checks passed!`
- `uv run ruff format --check .` → `473 files already formatted`
- `uv run pyright` → `0 errors, 0 warnings, 0 informations`
- `gda script set --help` / `gda shader set --help` / `gda --help`: byte-identical to base.
- `gda schema`: three intentional deltas (item 1 above), confirmed as the ONLY change via parsed-JSON diff.

Full mapping and evidence are also in the PR body (updated).

qinhaihong-red:
Re-reviewed the fixes at exact head `08b74bc5c945424db0da577cfc1f52027e3a4932` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

## Conclusion: Changes required, P3 residue only

The stale ownership prose, validation evidence, merge-order-fragile statement, duplicate usage-error normalization, and unused fixture findings are closed. All functional #713 acceptance criteria pass. Three judgement-call residues remain.

### [P3] Remove or explicitly support the contract-external raw `ValueError` branch

`params_or_bad_parameter` declares `type[P]` with `P` bound to `BaseModel`, and all 16 production call sites pass concrete Pydantic models. Pydantic wraps field and model validator `ValueError`s as `ValidationError`. `_RawValueErrorModel` is outside that declared contract and needs `# type: ignore[arg-type]`; it only proves the speculative bare-`ValueError` branch by bypassing the contract.

The smallest fix is to remove the bare `ValueError` branch, its docstring claim, and the synthetic test. If custom `BaseModel.__init__` exceptions are intended as a supported extension, document that contract and test it with a real `BaseModel` subclass instead.

### [P3] Keep the shared renderer cleanup as an independently scoped change

The caller-value leak is real and the shared renderer fix is correct. The remaining concern is packaging, not a request to restore the leak. Moving the cleanup into a prerequisite or separate PR preserves the fix while keeping #713 independently reviewable. The reply currently treats splitting the fix as equivalent to reverting it, which does not answer that scope concern.

### [P3] The #698 rationale maintenance remains outside issue #713

The revised text is now factually merge-order-safe, so the old correctness defect is closed. File ownership explains why the edit was routed here, but it does not by itself amend #713 or remove the independent review and cherry-pick cost. This item can also close if the coordinator explicitly records the routing decision as a specification amendment for this wave.

Independent validation: 354 focused tests passed; the instrumented script/shader by argv/JSON matrix observed exactly one mode-resolution call in every case; help is byte-identical to base; schema differs only at the four declared description leaves. Exact-head CI is green with `1907 passed, 2 skipped, 539 deselected`; Godot E2E is skipped for this non-engine-facing scope.

qinhaihong-red:
Addressed the round-5 re-review (comment on head `08b74bc5`). New commit: `9c7a7a68`.

## [P3] `_RawValueErrorModel` bypasses the declared contract — PARTIALLY ADOPTED

Agreed the bypass itself is real: a non-`BaseModel` stand-in needing `# type: ignore[arg-type]` to reach `type[P]` (`P` bound to `BaseModel`) is a genuine contract escape, not just an odd double — round 4's framing ("no pydantic model can exercise this branch") was wrong.

**Not adopted: deleting the branch.** It's reachable from a real `BaseModel` subclass:

```python
class M(BaseModel):
    a: int = 1
    def __init__(self, **kw):
        raise ValueError("a plain refusal sentence.")
M(a=1)   # -> propagates a BARE ValueError; issubclass(M, BaseModel) is True
```

`model_post_init` doesn't reach it — pydantic wraps that hook's raise as `ValidationError` too, same as a field/model validator. Overriding `__init__` runs before pydantic's own validation machinery gets a chance to wrap anything, so this is the one route where a plain `ValueError` escapes untouched. That makes the clause defensive-but-in-contract, not speculative — something has to pin what it does, and deleting the branch would leave a future model that raises from `__init__` producing an unhandled traceback instead of a clean usage error.

**Adopted: fixing the bypass itself.** `_RawValueErrorModel` now subclasses `BaseModel` and raises from an overridden `__init__`, instead of faking the shape from outside the contract. The `# type: ignore[arg-type]` is gone. Docstring rewritten to state the verified in-contract route (the old "one shape pydantic can never produce" line is exactly what the `M` example above contradicts).

## [P3] Package the shared-renderer cleanup separately — NOT ADOPTED (answered properly this time)

Agreed my round-4 reply missed the point — it treated packaging and reverting as equivalent, and they aren't. Answering the actual concern:

- The argv half isn't separable from this PR: the wording regression it fixes was introduced BY this PR (round 2), when `script set`/`shader set` got routed through the params model. Fixing it here is repair of this PR's own damage, not an unrelated cleanup riding along.
- The `--params-json` half was an explicit, already-recorded orchestrator scope call (round 3, "Scope of the parity claim"): shipping a state where the human channel is clean and the machine channel — the one an agent actually reads — still dumps pydantic internals was judged worse than the wider diff, for an agent-facing CLI.
- Splitting now costs a new PR, a rebase, and re-verifying an already twice-reviewed change end to end. The maintainer can still choose to land the cleanup separately at merge time if they weigh that trade-off differently — nothing here forecloses it.

No code restructured for this item.

## [P3] The `#698` rationale edit sits outside issue #713 — CLOSED BY RECORD

Recording the routing decision at the wave level, per your stated closing condition: `src/gda/commands/script.py` is this slice's owned file, PR #756 fixed the parser that made the old comment false, and the file-ownership rule is what kept #756 out of a file it doesn't own — so routing the comment fix here, rather than there, is the rule working as intended, not scope creep. This is recorded as a specification amendment for the wave in the wave's planning document. No code change.

## Validation (round 5)

- `uv run pytest -q -m "not e2e"` → `1909 passed, 539 deselected in 183.41s (0:03:03)`
- `uv run ruff check .` → `All checks passed!`
- `uv run ruff format --check .` → `473 files already formatted`
- `uv run pyright` → `0 errors, 0 warnings, 0 informations`
- `gda schema` re-diffed against base as parsed JSON: unchanged from round 4 — exactly the four declared leaves (`input.description` and `input.$defs.ScriptSetMode.description`, under `script set` and `shader set`). This round touched only `tests/test_dispatch.py`, no `src/` file, so no new schema delta is possible.

PR body updated with this round's mapping.

qinhaihong-red:
Re-reviewed the claimed fixes at exact head `9c7a7a68e9f7b4f1b991510f5117ae00d9db2a6f` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

## Conclusion: Changes required, P3 residue only

The replacement `_RawValueErrorModel` is now a genuine `BaseModel` subclass and closes the contract-external test-double defect. The shared renderer remains technically coherent at its single lower-level seam. Two documentation/scope residues remain.

### [P3] Synchronize the production helper contract

`src/gda/dispatch.py:52-53` still says a raw `ValueError` is "the plain sentence a validator wrote." The new test and reply establish the opposite: field/model validators and `model_post_init` are wrapped as `ValidationError`; the supported bare-`ValueError` route is a custom `BaseModel.__init__`.

Please update the production helper docstring to name that constructor route, so the declared contract and its regression test describe the same supported extension.

### [P3] Make the scope amendments independently locatable

The reply says the all-command `--params-json` renderer change was an explicit orchestrator scope decision and that the #698 rationale routing was recorded in the wave's planning document. Issue #713 is unchanged, and repository/GitHub searches found no linked planning document, URL, path, or revision containing those amendments.

Please link the existing authoritative record, or record the two scope decisions in a stable, reviewable authority. The issue is traceability, not a request to revert the renderer or recreate the file-ownership collision.

Focused exact-head tests pass and required CI is green. The head was re-pinned before publishing and was unchanged.

qinhaihong-red:
Re-review follow-up complete at exact head `264b81d08685463c93a11b579ed4c435401afd8d`.

I evaluated all three residues and adopted them as follows:

1. **Raw `ValueError` contract — fully adopted.** Commit `264b81d0` corrects `params_or_bad_parameter()`'s production docstring: the verified bare route is a custom `BaseModel.__init__` raising before pydantic wraps the exception. Field/model validators and `model_post_init` are now explicitly described as arriving as `ValidationError`.
2. **Shared renderer scope — fully adopted as specification traceability.** I retained the shared implementation because the argv half repairs this PR's own routing regression, while leaving the agent-facing `--params-json` channel on the leaking renderer would preserve the same measured defect one layer down. The public canonical scope is now amended in [issue #713](https://github.com/aigengame/godot-agent/issues/713), including its wider prose-only effect and the explicit `perf --budget-file` exclusion.
3. **#698 rationale routing — fully adopted as specification traceability.** Issue #713 now also records the file-ownership/merge-order reason for maintaining the stale `script.py` rationale in this slice without changing #698 runtime behavior.

The PR body has been updated to point to that stable issue authority instead of a wave-only planning record.

Local validation:
- `pytest -q -p no:cacheprovider tests/test_dispatch.py` → **8 passed**
- Ruff check/format → **passed**
- Pyright → **0 errors, 0 warnings, 0 informations**
- `git diff --check` → **clean**

Current exact-head CI: Ruff, Pyright, co-install, scope classification, and gda-balancing required are green; Python unit/package and the latest release-scope guard are still running/queued.

**Final CI update:** all required checks pass at exact head `264b81d08685463c93a11b579ed4c435401afd8d`; Godot E2E and out-of-scope gda-balancing jobs are skipped as expected. Merge state is `CLEAN`.

PR #755, opened 2026-08-27
test(gda): pin the ADR-0002 error-description column (#701)

### LINKED ISSUE #701: Pin the error-code registry's description column against ADR-0002 drift
## Review-round source

Found during PR #695's review rounds (wave 2). `tests/test_error_registry.py`'s registry↔ADR-0002 equality pin compares only `(category, source, exit_code)` per code — the description column is parsed out of the ADR table but never compared, so the Python registry's description and the ADR row's prose can diverge silently. Wave 2 edited descriptions in three PRs (#694, #695, #696) and each had to be verified to agree by manual inspection; the drift risk is standing and registry-wide.

## What to build

Extend the pin to the description column: compare the ADR table's description text with the Python registry's. Fix any existing divergence the new pin surfaces, in whichever artifact is wrong.

### Measured before dispatch (2026-08-27) — backticks are NOT the only systematic difference

Comparing `gda.error_codes.ERROR_CODES` against the ADR-0002 table on `main`, after
normalizing backticks, whitespace and a trailing period:

| | |
|---|---|
| codes compared | 91 (91 ADR rows, 91 registry entries — no missing rows) |
| still differ | **42** |
| — trailing provenance only | 27 |
| — **substantive content** | **15** |

So the issue's premise ("the one systematic difference observed") understates the work by
an order of magnitude, and the slice has TWO decisions in it, not one:

1. **A second normalization rule must be decided, not just implemented.** 27 of the 42 differ
   only because the ADR table carries a trailing provenance parenthetical the registry does
   not — `(Phase 2, ADR-0017 / ADR-0021)`, `(Phase 2, #225)`, `(#170)`. Either the pin strips
   trailing provenance refs (and the ADR keeps them as the human-facing record), or the
   registry adopts them. Both are defensible; picking silently inside the test is what makes
   the next reader distrust the guard.
2. **15 are real content divergences and each needs a per-code verdict.** Samples:
   - `export_path_unset` — ADR: *"neither a `--output` override nor a configured `export_path`"*;
     registry: *"targeted a preset that has no configured `export_path`"*. The registry text
     silently omits an entire producing condition.
   - `export_templates_missing` — the ADR carries the exemption *"pack needs no platform
     templates and is exempt"*; the registry does not.
   - `launch_timeout` — the ADR carries *"One command does not report it — see the 2026-08-19
     scope note"*; the registry does not.
   - `live_display_unavailable` — ADR *"start the daemon **windowed** with …"* vs registry
     *"start the daemon with …"*; the registry's remedy is incomplete as written.

   These are not formatting — they are real inaccuracies in a public-ABI reference.
   **Correction (2026-08-27, after implementation):** an earlier version of this paragraph said
   at least three of them change "what an agent reading `gda schema` learns about a failure".
   That is wrong. `ErrorCodeSpec.description` reaches NO gda output — not `gda schema`, not a
   per-command `--schema` (the `error` key carries the `GdaErrorEnvelope` shape, with no
   code-with-prose enum), and not an emitted envelope (`message` comes from the failure
   builder). Measured: `gda schema` is byte-identical before and after seven descriptions were
   rewritten, and `ERROR_CODES` is consumed only by `errors.py` (for derived code SETS) and by
   tests. The audience for these descriptions is a human, or an agent reading the source —
   which is reason enough to keep them accurate, but not the reason originally given.
   Reconciling them is the substance of this slice; the pin is what stops it recurring.

Reproduce with: `uv run python -c` over `gda.error_codes.ERROR_CODES` and the ADR table
(the measurement script is trivial and deliberately not committed — the committed artifact
is the pin itself).

## Acceptance criteria

- [ ] The registry guard fails when a description differs between `error_codes.py` and the
      ADR-0002 table, beyond the normalization rules the slice explicitly decides.
- [ ] The provenance-reference question (27 codes) is decided one way for the whole table and
      the decision is stated in the test, not implied by a regex.
- [ ] Each of the 15 substantive divergences gets a verdict recorded in the PR body — which
      artifact was wrong and why — rather than being flattened to whichever side the pin
      happened to be written against.
- [ ] The suite is green after the pin lands (any surfaced divergence reconciled).
- [ ] The normalization rule is documented in the test so a future table-format change is a conscious edit.

## Blocked by

None - can start immediately (land after the open wave-2 PRs merge to avoid churn on files they touch).

qinhaihong-red:
Closes #701

## What changed

`tests/test_error_registry.py`'s registry↔ADR-0002 equality pin parsed the table's
`Meaning` cell and threw it away, comparing only `(category, source, exit_code)`. The pin
now compares all four columns, and the divergences it surfaced are reconciled.

The comparison reuses the existing `_adr_registry()` / `_python_registry()` helpers — they
now return a 4-field `ErrorRow` and the one existing equality assertion covers the
description too. No parallel parser was added.

## Derived measurement (reproduced, not quoted)

Comparing `gda.error_codes.ERROR_CODES` against the ADR-0002 table at the fork point
(`04d3e089`):

| | |
|---|---|
| codes compared | 91 (91 ADR rows, 91 registry entries; no row missing on either side) |
| differ after markup / whitespace / trailing-period normalization | **42** |
| — trailing provenance reference only | 27 |
| — **substantive content** | **15** |

This matches the issue's pre-dispatch measurement once the provenance grammar admits
`ADR-0031 amendment` as a token (`script_failed` and `script_aborted` differ by nothing
else). With a narrower grammar the split reads 25 / 17.

### After reconciliation, at head `3d035725`

Three counts, three different populations — an earlier version of this line conflated the
first two, which is the one defect a PR about measurement accuracy cannot afford:

| Count | Value | What it is |
|---|---|---|
| differ after backtick / whitespace / trailing-period normalization | **38** | every one explained by a trailing citation |
| ADR descriptions carrying a strippable citation | **41** | a property of the ADR side alone |
| differ after the citation rule | **0** | what the pin asserts |

41 and 38 are not the same population and neither is a subset relation by luck: **3** of the
41 (`ambiguous_class_name`, `invalid_params`, `usage_error`) carry the *same* citation on the
registry side and match otherwise, so they were never in the 38. Two more registry
descriptions carry a citation (`script_aborted`, `script_failed`) but a *different* one from
their ADR row — `(#655)` vs `(ADR-0031 amendment, #655)` — so those two do stay in the 38 and
are resolved by the rule, not by already matching. That is 5 registry descriptions with a
citation, of which only 3 subtract.

These are a dated snapshot, not a contract: the pin asserts row-for-row equality, never a
count, so a legitimate new citation moves these numbers and breaks nothing.

## Decision 1 — the citation rule, and why

**A trailing parenthetical is stripped, on both sides, and not compared — but only when
it CITES at least one ADR or issue and holds nothing else.** Stated with its reasoning in
`tests/test_error_registry.py` above `_normalized_description`.

Why this and not "the registry adopts the refs":

- Saying which decision added a row is the ADR's job as a decision record. The registry's
  `description` is the code's *meaning* — the sentence read next to the code, where an
  unresolvable issue number is noise.
- Adopting them means copying 41 references into the public code descriptions and making
  "append an issue number" a standing obligation for every new code.
- Stripped on **both** sides rather than the ADR's only, because the two artifacts cite for
  different readers and 5 registry descriptions carry a pointer worth keeping where it is
  (`usage_error`, `invalid_params`, `ambiguous_class_name`, `script_failed`,
  `script_aborted`).

**Why a reference is required, not merely allowed** (from the #755 review). The first form
of this rule stripped any parenthetical built from the token set, and `Phase N` was in that
set — so `(Phase 2)` and `(Phase 1)` normalized alike and a mislabelled row would have
passed the guard silently. Phase is domain content, not a citation: `CONTEXT.md` defines
Phase 1 / Phase 2 as the order capabilities are delivered in. `Phase N` stays in the token
set so it can ride *inside* a cluster — removing it would unstrip
`(Phase 2, ADR-0017 / ADR-0021)` and turn a third of the cited rows into false divergences.

**Residual, stated exactly:** a `Phase N` riding inside a citation cluster is stripped with
it and so is not compared. For 20 of the 23 rows carrying a phase label the compared
`Category` column pins the distinction indirectly — `live` exists only in Phase 2
(ADR-0021) — but the other three are `environment`-category
(`live_unsupported_platform`, `live_windowed_unavailable`,
`live_windowed_permission_denied`), so for those a phase mislabel inside a citation cluster
is caught by neither mechanism. Left that way deliberately and not taken as a follow-up:
three docs-only rows in a field with no runtime consumer do not justify the 27 rows of
churn that closing the riding case would cost.

The rule stays deliberately narrow, and that narrowness is load-bearing: a parenthetical
that **mixes prose with a ref is not a citation** and is compared verbatim. #701 contains
exactly that case — `export_templates_missing`'s `(pack needs no platform templates and is
exempt; #170)`, a real exemption the registry had lost. It was reconciled into both
artifacts, not normalized away. Two boundary tests
(`test_description_normalization_ignores_only_citations`,
`test_a_bare_phase_label_is_compared_not_stripped`) pin both directions, so widening the
token set stays a conscious edit.

The ADR carries a short summary of the rule above its table, since that is where a human
edits a row — including the fact that a registry description may carry a citation too, so
nobody "tidies away" one of the five.

## Decision 2 — a verdict per substantive divergence

18 divergences, 18 verdicts — the original 15, plus 3 more surfaced by the tightened
citation rule. Seven fix the **registry**, nine fix the **ADR**, two are phrasing with no
correctness winner. The registry is the authoritative source (ADR-0002
says so), so the default direction is "the ADR adopts the registry" — overridden wherever
the ADR's text was the more correct or more complete one.

| Code | Verdict | Why |
|---|---|---|
| `export_path_unset` | **registry was wrong** → adopts ADR | Registry named only the configured `export_path`; `export_path_unset_failure`'s own docstring says the destination is `--output` if given, else `export_path`, and the code fires only when **both** are absent. A reader of the registry would not know the override exists. |
| `export_templates_missing` | **registry was wrong** → adopts ADR | Registry said "an export run"; `export_templates_missing_failure` skips the check for `--mode pack` entirely. The registry claimed the code can fire for a mode it never fires for, and had lost the `pack` exemption. |
| `launch_timeout` | **registry was wrong** → adopts ADR (reworded) | ADR carried "one command does not report it — see the 2026-08-19 scope note"; the registry carried no caveat. `scene preflight` reports exceeding its own bound as a `timeout` **status on a success result**, so an agent branching on this code alone never sees that command's timeout. Both sides now name the command instead of pointing at a note, so the sentence stands on its own in the registry. |
| `live_display_unavailable` | **registry was wrong** → adopts ADR | Registry's remedy read "start the daemon with `gda daemon start --windowed`", dropping *windowed* — the word that says what is being fixed. The registry's own code comment already called it "the self-revealing remediation (start the daemon windowed)". |
| `live_predicate_unmet` | **registry was wrong** → adopts ADR | Registry said "a live capture predicate"; the ADR names the command and flag (`screen capture`, `--await-*`). The code exists for exactly one surface and naming it costs nothing. (`#661` moved out of the inline parenthetical into the row's trailing provenance, where it belongs.) |
| `live_unsupported_platform` | **registry was incomplete** → adopts ADR | The ADR's "Phase-1 headless is unaffected" is the scope a reader hitting this on Windows most needs; the registry dropped it. |
| `tree_too_deep` | **registry was incomplete** → adopts ADR | The ADR explains that it shares `contract_violation`'s exit code and is told apart by `code`. That is precisely the branching advice a registry entry should carry. |
| `live_timeout` | **ADR was incomplete** → adopts registry | The registry documents that the session is discarded and runtime state does not survive. Verified in `daemon/session.py`: the timeout latches `_channel_stale` because the protocol has no request id and a late frame would be read as the *next* op's reply. A live-op consumer must know this; the ADR omitted it. |
| `live_method_not_allowlisted` | **ADR was wrong** → adopts registry | ADR said "its **class** never declared gda-callable". ADR-0041 resolves `GDA_CALLABLE` along the addressed node's **attached script's** base chain — a node's engine class never declares anything. The registry was accurate. |
| `live_invalid_call_args` | **ADR adopts registry** | Em-dash vs colon before an enumeration; no correctness difference. Default direction. |
| `live_windowed_unavailable` | **ADR adopts registry** | Word order only (`A windowed Engine session (…) was requested` vs `was requested (…)`). Default direction. |
| `live_windowed_permission_denied` | **ADR adopts registry** | Same word-order difference as its sibling; kept consistent with it. |
| `stdout_spill_failed` | **ADR adopts registry** | `A script run stdout` → `A script run's stdout`. Grammar. |
| `engine_disconnected` | **phrasing; registry adopts the ADR's em-dash** | Same content (parenthetical vs em-dash). Direction flipped from the default here for one reason: the registry's form ends in a parenthetical, and the ADR row appends `(Phase 2)` after it — adopting the registry text would stack two parentheticals at the end of the cell. |
| `script_not_found` | **phrasing; both adopt one form** | Same content, different phrasing. Reworded to `— it still exits 0, and gda reads the failure from stderr` for the same reason: the registry's parenthetical form would stack against the ADR row's `(#651)`. |
| `engine_session_not_running` | **ADR carried a redundant label** → dropped | Surfaced by the tightened rule. The bare `(Phase 2)` duplicates the row's own `Category` column, which reads `live` — a category that exists only in Phase 2 (ADR-0021) and that the pin already compares. Dropped rather than copied into the registry: a delivery-phase label is not part of a code's meaning, and 24 sibling live codes carry none. |
| `engine_disconnected` | **ADR carried a redundant label** → dropped | Same, and the same reasoning. |
| `live_timeout` | **ADR carried a redundant label** → dropped | Same, and the same reasoning. |

## On the "an agent reading `gda schema`" framing

Worth correcting for the record: `ErrorCodeSpec.description` reaches **no** gda output
today. It is not in `gda schema`, not in a per-command `--schema` (the uniform `error` key
carries the `GdaErrorEnvelope` shape, not a code enum with prose), and not in an emitted
envelope — a failure's `message` comes from the failure builder, not the registry row.
Proof: `gda schema` is **byte-identical** before and after this change, despite seven
registry descriptions being rewritten.

So the description column is documentation, in two places: the ADR table and
`error_codes.py` itself. That does not weaken the case for the pin — both are read by
humans and by agents reading the source, and #701's substantive findings are real
inaccuracies in a public-ABI reference — but the divergences were not silently degrading
`gda schema` output, and the claim should not be repeated as if they were.

## Validation

| Gate | Command | Result |
|---|---|---|
| Suite | `uv run pytest -q -m "not e2e"` | `1903 passed, 539 deselected in 111.29s (0:01:51)` |
| Lint | `uv run ruff check .` | `All checks passed!` |
| Format | `uv run ruff format --check .` | `472 files already formatted` |
| Types | `uv run pyright` | `0 errors, 0 warnings, 0 informations` |
| Schema | `gda schema` before vs after | no difference (see above) |

**Red-proof, two ways.** Perturbing one registry description (`path_not_found`: "does not
exist" -> "does not exist anywhere") fails the equality pin with a readable diff, and
passes again once reverted:

```
E         Differing items:
E         {'path_not_found': ErrorRow(..., description='A requested file does not exist')} !=
          {'path_not_found': ErrorRow(..., description='A requested file does not exist anywhere')}
1 failed in 0.52s
```

And the case the review found — writing a bare `(Phase 1)` onto a `live` ADR row — now
fails too, where before it normalized away silently:

```
E         {'engine_session_not_running': ErrorRow(..., description='...to serve the live operation (Phase 1)')} !=
          {'engine_session_not_running': ErrorRow(..., description='...to serve the live operation')}
1 failed in 0.14s
```

**Godot e2e was not run** — it is not needed for this slice: no engine-facing behavior,
no GDScript, and no emitted output changes.

## Review round (PR #755)

All four findings fully adopted, in `56126712`:

1. **P2 — a bare `(Phase N)` was stripped.** Reproduced against my own function before
   fixing. Rule tightened to require a citation; the 3 rows it surfaced reconciled with
   verdicts in the table above; boundary pinned in both directions by a new test.
2. **P2 — the ADR claimed the trailing reference "rides these rows only".** False; derived
   the real figures (41 ADR rows, 5 registry descriptions) and stated both, so nobody
   removes one of the five as noise.
3. **P3 — the ADR's stated scope ("naming the deciding ADR or issue") was narrower than the
   implementation.** Rewritten in the same edit to describe what the rule now does,
   including the bare-`Phase N` exclusion.
4. **P3 — no pointer on the authoritative side.** `error_codes.py`'s module docstring now
   says the wording is pinned and where the rule lives.

Not re-litigated: generating the table instead of guarding it. Agreed with the architecture
lens — ADR-0002 is a narrative record with the table embedded in prose, so generation buys a
marker-based partial-file generator plus a CI step and reproduces the same drift class at
"forgot to regenerate".

## Remainder

None for #701's acceptance criteria. Two judgement calls worth a reviewer's eye. The
`Phase N`-inside-a-citation residual is described under Decision 1 — including the three
`environment`-category rows it leaves genuinely uncaught. And the
`launch_timeout` row now names `scene preflight` instead of pointing at ADR-0002's
2026-08-19 scope note, so the sentence works in both artifacts. The scope note still
explains the *why* and is findable — its bullet is titled "The launch bound as a verdict,
not `launch_timeout`".

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -219,8 +219,8 @@ The `Meaning` column is pinned too (#701): it must match the registry's
 `description` once markup and wrapping are normalized. A trailing parenthetical
 that cites at least one ADR or issue — a `Phase N` label may ride along inside
 it — is a citation rather than part of the code's meaning, so it is not compared.
-41 rows here carry one; so do 5 registry descriptions, which is allowed, not a
-mistake to tidy away. A bare `(Phase N)` is *not* a citation and IS compared,
+Many rows here carry one, and some registry descriptions do too: that is allowed,
+not a mistake to tidy away. A bare `(Phase N)` is *not* a citation and IS compared,
 which is why the live rows below state their phase through the `Category` column
 instead. `tests/test_error_registry.py` is the single home of that rule.
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -217,9 +217,12 @@ source and is checked by tests. GDScript mirrors only the rows whose source is
 
 The `Meaning` column is pinned too (#701): it must match the registry's
 `description` once markup and wrapping are normalized. A trailing parenthetical
-naming the deciding ADR or issue is provenance for *this* record rather than part
-of the code's meaning, so it rides these rows only and is not compared —
-`tests/test_error_registry.py` is the single home of that rule.
+that cites at least one ADR or issue — a `Phase N` label may ride along inside
+it — is a citation rather than part of the code's meaning, so it is not compared.
+41 rows here carry one; so do 5 registry descriptions, which is allowed, not a
+mistake to tidy away. A bare `(Phase N)` is *not* a citation and IS compared,
+which is why the live rows below state their phase through the `Category` column
+instead. `tests/test_error_registry.py` is the single home of that rule.
 
 Each row carries the process `Exit Code` a shell consumer keys on. It is
 per-code, not per-category: within `environment`, `binary_not_found` exits `127`
@@ -295,9 +298,9 @@ operation, and parse codes the CLI assigns).
 | `contract_violation` | `parse` | `parser` | `5` | The process claimed success but violated the structured-output contract. |
 | `tree_too_deep` | `parse` | `classifier` | `5` | The engine emitted a valid result tree that nests past gda's recursion limit; the payload is contract-conformant, the limit is wrapper-side (shares the `parse` exit code; the `code` distinguishes it from `contract_violation`). |
 | `daemon_not_running` | `live` | `classifier` | `6` | A live command found no running gda-daemon for the project; start one with `gda daemon start` (Phase 2, ADR-0017 / ADR-0021). |
-| `engine_session_not_running` | `live` | `classifier` | `6` | The daemon is running but holds no live engine session to serve the live operation (Phase 2). |
-| `engine_disconnected` | `live` | `classifier` | `6` | The engine session disconnected before the live operation returned — the game crashed or the harness connection dropped (Phase 2). |
-| `live_timeout` | `live` | `classifier` | `6` | A live operation did not return from the engine session before the daemon's timeout. The session is discarded: its reply is no longer attributable, so the next operation relaunches it and runtime state does not survive (Phase 2). |
+| `engine_session_not_running` | `live` | `classifier` | `6` | The daemon is running but holds no live engine session to serve the live operation. |
+| `engine_disconnected` | `live` | `classifier` | `6` | The engine session disconnected before the live operation returned — the game crashed or the harness connection dropped. |
+| `live_timeout` | `live` | `classifier` | `6` | A live operation did not return from the engine session before the daemon's timeout. The session is discarded: its reply is no longer attributable, so the next operation relaunches it and runtime state does not survive. |
 | `daemon_running` | `live` | `classifier` | `6` | A daemon-lifecycle command was refused because a gda-daemon is running for the project; stop it first with `gda daemon stop` (Phase 2, #225). |
 | `daemon_already_running` | `live` | `classifier` | `6` | A `gda daemon start --scene` was refused because a gda-daemon is already running for the project; `--scene` only takes effect at start, so stop it with `gda daemon stop` then start again with `--scene` (Phase 2, #278). |
 | `live_node_not_found` | `live` | `classifier` | `6` | A live game operation's node path does not resolve to a node in the running scene tree (Phase 2, #220). |
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -215,6 +215,12 @@ the Python registry in `src/gda/error_codes.py`; the table below mirrors that
 source and is checked by tests. GDScript mirrors only the rows whose source is
 `operation`, because only those codes can be reported by headless operations.
 
+The `Meaning` column is pinned too (#701): it must match the registry's
+`description` once markup and wrapping are normalized. A trailing parenthetical
+naming the deciding ADR or issue is provenance for *this* record rather than part
+of the code's meaning, so it rides these rows only and is not compared —
+`tests/test_error_registry.py` is the single home of that rule.
+
 Each row carries the process `Exit Code` a shell consumer keys on. It is
 per-code, not per-category: within `environment`, `binary_not_found` exits `127`
 but `launch_timeout` exits `124`. The `exit_codes.py` registry defines the
@@ -224,7 +230,7 @@ operation, and parse codes the CLI assigns).
 | Code | Category | Source | Exit Code | Meaning |
 | --- | --- | --- | --- | --- |
 | `binary_not_found` | `environment` | `runner` | `127` | The Godot binary could not be launched. |
-| `launch_timeout` | `environment` | `runner` | `124` | Godot launched but did not return before the runner timeout. One command does not report it — see the 2026-08-19 scope note (#664). |
+| `launch_timeout` | `environment` | `runner` | `124` | Godot launched but did not return before the runner timeout. One command does not report it: `scene preflight` reports its own `timeout` status instead (#664). |
 | `user_data_unwritable` | `environment` | `runner` | `127` | The log or user-data placement for the launch could not be made usable, so the launch was refused. |
 | `unknown_command` | `usage` | `classifier` | `2` | gda has no such command; discover the surface with `gda schema` or `gda --help`. A recognized near miss also carries the supported invocation in the envelope's `hint`. |
 | `unknown_option` | `usage` | `classifier` | `2` | The command exists but has no such option; read its options with `--help` or its input contract with `--schema`. A recognized near miss also carries the supported invocation in the envelope's `hint`. |
@@ -265,7 +271,7 @@ operation, and parse codes the CLI assigns).
 | `no_search_match` | `operation` | `operation` | `4` | A search-replace script edit found no occurrence of the search string. |
 | `invalid_line_range` | `operation` | `operation` | `4` | A line-range script edit specified lines outside the script's bounds, or end before start. |
 | `script_compile_failed` | `operation` | `operation` | `4` | A script does not compile, so the requested work could not proceed: `script attach` refuses to bind it to a node, and `script run` reports that the entry script (or a dependency it preloads) never ran (#651). |
-| `script_not_found` | `operation` | `classifier` | `4` | A `script run` entry script does not exist in the project, so the engine never ran it — read from stderr, since the engine still exits 0 (#651). |
+| `script_not_found` | `operation` | `classifier` | `4` | A `script run` entry script does not exist in the project, so the engine never ran it — it still exits 0, and gda reads the failure from stderr (#651). |
 | `script_failed` | `operation` | `classifier` | `4` | A `script run --strict` script ran to completion and chose a non-zero exit status; strict mode maps that opted-in failure onto the uniform error envelope. Never reported without `--strict` (ADR-0031 amendment, #651). |
 | `script_aborted` | `operation` | `classifier` | `4` | A `script run` was ended early, before its `--timeout`: a script error appeared on stderr, the caller's declared `--completion-marker` did not, and the run then went silent. Reported only when `--completion-marker` is declared (ADR-0031 amendment, #655). |
 | `incompatible_script_type` | `operation` | `operation` | `4` | A script compiles but its base type is incompatible with the requested use: `script attach`'s target node type, or `script run`'s requirement that a one-shot entry script extend `SceneTree`/`MainLoop` (#651). |
@@ -276,9 +282,9 @@ operation, and parse codes the CLI assigns).
 | `export_presets_not_found` | `operation` | `operation` | `4` | The project has no export_presets.cfg, so it defines no export presets. |
 | `export_preset_not_found` | `operation` | `operation` | `4` | No export preset with the requested name exists in export_presets.cfg. |
 | `export_path_unset` | `operation` | `classifier` | `4` | An export run has no destination — neither a `--output` override nor a configured `export_path` (#170). |
-| `export_templates_missing` | `operation` | `classifier` | `4` | A release/debug export needs the export templates for the running engine version, which are not installed (pack needs no platform templates and is exempt; #170). |
+| `export_templates_missing` | `operation` | `classifier` | `4` | A release/debug export needs the export templates for the running engine version, which are not installed; `pack` needs no platform templates and is exempt (#170). |
 | `export_output_parent_failed` | `operation` | `classifier` | `4` | An export run could not create the output parent directory before native export (#402). |
-| `stdout_spill_failed` | `operation` | `classifier` | `4` | A `script run` stdout exceeded the cap but the complete-stream spill file could not be written, so the bounded result cannot be delivered (#665). |
+| `stdout_spill_failed` | `operation` | `classifier` | `4` | A `script run`'s stdout exceeded the cap but the complete-stream spill file could not be written, so the bounded result cannot be delivered (#665). |
 | `export_failed` | `operation` | `classifier` | `4` | A native Godot export run failed (the engine reported the export did not complete). |
 | `invalid_uid` | `operation` | `operation` | `4` | A requested `uid://` value is not a syntactically valid resource UID. |
 | `unknown_uid` | `operation` | `operation` | `4` | A syntactically valid resource UID is not registered in the engine's UID cache. |
@@ -291,7 +297,7 @@ operation, and parse codes the CLI assigns).
 | `daemon_not_running` | `live` | `classifier` | `6` | A live command found no running gda-daemon for the project; start one with `gda daemon start` (Phase 2, ADR-0017 / ADR-0021). |
 | `engine_session_not_running` | `live` | `classifier` | `6` | The daemon is running but holds no live engine session to serve the live operation (Phase 2). |
 | `engine_disconnected` | `live` | `classifier` | `6` | The engine session disconnected before the live operation returned — the game crashed or the harness connection dropped (Phase 2). |
-| `live_timeout` | `live` | `classifier` | `6` | A live operation did not return from the engine session before the daemon's timeout (Phase 2). |
+| `live_timeout` | `live` | `classifier` | `6` | A live operation did not return from the engine session before the daemon's timeout. The session is discarded: its reply is no longer attributable, so the next operation relaunches it and runtime state does not survive (Phase 2). |
 | `daemon_running` | `live` | `classifier` | `6` | A daemon-lifecycle command was refused because a gda-daemon is running for the project; stop it first with `gda daemon stop` (Phase 2, #225). |
 | `daemon_already_running` | `live` | `classifier` | `6` | A `gda daemon start --scene` was refused because a gda-daemon is already running for the project; `--scene` only takes effect at start, so stop it with `gda daemon stop` then start again with `--scene` (Phase 2, #278). |
 | `live_node_not_found` | `live` | `classifier` | `6` | A live game operation's node path does not resolve to a node in the running scene tree (Phase 2, #220). |
@@ -299,8 +305,8 @@ operation, and parse codes the CLI assigns).
 | `live_unknown_property` | `live` | `classifier` | `6` | A live game get or set targeted a property name the running node does not expose as an addressable runtime, storage, or attached-script property (Phase 2, #220, #422). |
 | `live_uncoercible_value` | `live` | `classifier` | `6` | A live game set value cannot be coerced to the addressed runtime property's or script variable's Godot type (Phase 2, #220, #422). |
 | `live_unknown_method` | `live` | `classifier` | `6` | A live game call named a method the addressed running node does not have (Phase 2, #673). |
-| `live_method_not_allowlisted` | `live` | `classifier` | `6` | A live game call named a method the addressed running node has but its class never declared gda-callable (Phase 2, #673, ADR-0041). |
-| `live_invalid_call_args` | `live` | `classifier` | `6` | A live game call supplied arguments the declared method cannot take — a count outside its accepted range, or a value the declared parameter type cannot convert from (Phase 2, #673). |
+| `live_method_not_allowlisted` | `live` | `classifier` | `6` | A live game call named a method the addressed running node has but its attached-script chain never declared gda-callable (Phase 2, #673, ADR-0041). |
+| `live_invalid_call_args` | `live` | `classifier` | `6` | A live game call supplied arguments the declared method cannot take: a count outside its accepted range, or a value the declared parameter type cannot convert from (Phase 2, #673). |
 | `live_log_unavailable` | `live` | `classifier` | `6` | A live engine session was launched but its diagnostics log file is missing or unreadable, so `gda diag` cannot read the running game's errors/output (Phase 2, #224). |
 | `live_scene_not_found` | `live` | `classifier` | `6` | A `gda daemon start --scene` selector did not load: the launched session ran a different scene (Godot silently falls back to main_scene for a missing/invalid path or UID), verified by the harness at launch — gda never falls back (Phase 2, #278). |
 | `live_perf_node_not_found` | `live` | `classifier` | `6` | A live perf monitor's node path does not resolve to a node in the running scene tree (Phase 2, #223). |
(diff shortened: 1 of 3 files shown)


qinhaihong-red:
## Review round handled — `56126712`

All four findings **fully adopted**. Per-finding resolution:

| # | Finding | Resolution |
|---|---|---|
| 1 (P2) | A bare `(Phase N)` is stripped; Phase is domain content, not a citation | **Fixed @56126712.** Reproduced against `_normalized_description` first (`(Phase 2)` and `(Phase 1)` did collide). A trailing parenthetical is now strippable only when it cites at least one ADR or issue — implemented as a lookahead, so `Phase \d+` stays in the token set and `(Phase 2, ADR-0017 / ADR-0021)` still strips. Avoided the trap you named: the provenance-only row count is unchanged at 41. |
| 1b | Reconcile the rows it surfaces, with per-row verdicts | **Done.** Exactly the 3 you predicted: `engine_session_not_running`, `engine_disconnected`, `live_timeout`. Verdict for all three — **the ADR carried a redundant label; dropped**. The bare `(Phase 2)` duplicates the row's own `Category` column, which reads `live`, a category that exists only in Phase 2 (ADR-0021) and that the pin already compares. Not copied into the registry instead: that would put a delivery-phase label on 3 of 27 live codes and none on the rest, and a delivery phase is not part of a code's meaning. |
| 1c | Extend the boundary test in both directions | **Done.** New `test_a_bare_phase_label_is_compared_not_stripped` pins that `(Phase 2)` survives and differs from `(Phase 1)`, and that `(Phase 2, #220)` / `(Phase 2, ADR-0021)` still strip. |
| 2 (P2) | ADR:221 "rides these rows only" is factually wrong | **Fixed @56126712.** Derived both figures rather than asserting: **41** ADR rows carry a strippable citation, and **5** registry descriptions do too (`usage_error`, `invalid_params`, `ambiguous_class_name`, `script_failed`, `script_aborted`) — all 5 named, and stated as allowed rather than a mistake to tidy away. |
| 3 (P3) | ADR:220 "naming the deciding ADR or issue" narrower than the implementation | **Fixed @56126712**, same edit as #2. The sentence now describes the actual rule, including that a `Phase N` may ride inside a citation but a bare one is compared. |
| 4 (P3) | No pointer on the declared authoritative side | **Fixed @56126712.** `error_codes.py`'s module docstring gains an "Editing a `description`" paragraph naming the pin, what it normalizes away, and where the rule lives. |

**Red-proof of finding 1:** writing a bare `(Phase 1)` onto that `live` ADR row now fails the pin (`1 failed in 0.14s`); before the fix it normalized away silently. The original perturbation red-proof still holds.

**Residual I want on the record rather than papered over:** the fix closes the *bare* `(Phase N)` case. A `Phase N` riding inside a citation cluster is still stripped with it, so it is not compared on the other 24 rows. What pins those is the `Category` column next to them — `live` is Phase-2-only by definition, and that column *is* compared. Closing the riding case too would mean either 27 registry descriptions carrying a delivery-phase label or 27 ADR rows losing one; I judged neither worth the churn against a column that already pins the distinction, but say the word and I will take it as a follow-up. This is stated in the test comment and the PR body, not left implicit.

Also **not adopted, and I agree with your reasoning**: generating the table rather than guarding it.

Gates re-run on the new tip:

| Gate | Command | Result |
|---|---|---|
| Suite | `uv run pytest -q -m "not e2e"` | `1903 passed, 539 deselected in 111.29s (0:01:51)` |
| Lint | `uv run ruff check .` | `All checks passed!` |
| Format | `uv run ruff format --check .` | `472 files already formatted` |
| Types | `uv run pyright` | `0 errors, 0 warnings, 0 informations` |
| Schema | `gda schema` before vs after | byte-identical to the pre-slice base |

Godot e2e still not run and still not needed — no engine-facing behavior, no GDScript, no emitted output change.

qinhaihong-red:
## Correction to my residual claim — `3d035725`

My review reply above overstated what the guard covers, and I am correcting it here rather than leaving it standing.

I said a `Phase N` riding inside a citation cluster is uncompared but that the `Category` column pins it, `live` being Phase-2-only. Verified against every row: **23** ADR rows carry a phase label, and **3 of them are not `live`** — `live_unsupported_platform`, `live_windowed_unavailable`, `live_windowed_permission_denied` are `environment`-category. For those three the Category column pins nothing about the phase, so a phase mislabel inside their citation cluster is caught by **neither** mechanism. My claim was right for 20 rows and wrong for 3.

**No code change.** The exposure is three docs-only rows in a field with no runtime consumer, which does not justify the 27 rows of churn that closing the riding case would cost. The residual is now stated accurately in both places it is recorded — the test comment above `_normalized_description` and this PR's Decision 1 — and it is the resting place for the riding case, not a follow-up.

Gates re-run on `3d035725`, tool output verbatim:

| Gate | Result |
|---|---|
| `uv run pytest -q -m "not e2e"` | `1903 passed, 539 deselected in 74.44s (0:01:14)` |
| `uv run ruff format --check .` | `472 files already formatted` |

_(Edited: the first version of this comment carried a wall-clock figure I had not yet measured. The table above is the actual output.)_

qinhaihong-red:
## Exact-head adversarial review — changes required

Reviewed `3d035725abf6bd6d4dbb9266454ee45ac8a9aa4a` against merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`. I re-read the PR head after review; it is unchanged.

### Artifact Review

**Changes required**

1. **[P2] Correct the post-reconciliation measurement in the PR body.**

   The PR says: “After reconciliation: 0 substantive, 41 rows differing by trailing provenance only.” Exact-head reproduction gives:

   - 91 ADR rows;
   - 41 ADR descriptions carry a strippable trailing citation;
   - 5 registry descriptions carry one;
   - after only backtick, whitespace, and trailing-period normalization, **38** rows differ;
   - after the PR's citation normalization, 0 rows differ.

   Three registry descriptions (`usage_error`, `invalid_params`, and `ambiguous_class_name`) already carry the same citation as their ADR rows, so they are among the 41 citation-bearing ADR rows but not among the 38 differing rows. Please distinguish “41 ADR rows carry a citation” from “38 rows differ only because of trailing provenance.”

2. **[P3] Avoid an unguarded exact count in the accepted ADR.**

   `docs/adr/0002-headless-structured-output-contract.md:222` fixes the current counts at 41 ADR rows and 5 registry descriptions, but the guard does not derive or assert those counts. Adding a valid citation to another ADR row changes the count while the equality test still passes because the citation is intentionally normalized away.

   The smallest fix is stable qualitative wording in the ADR, such as “many ADR rows and some registry descriptions,” while retaining snapshot measurements in the PR body; alternatively, derive and assert the counts.

**Terminology and prose assessment:** I reviewed the complete current ADR-0002, issue #701, the PR body, and all current PR comments against `CONTEXT.md`, the applicable accepted ADRs, the implementation, and ASD-STE100. No other substantive terminology or prose issue was found.

### Architecture / DDMA

**[P3] The volatile measurements have no single knowledge owner.**

`docs/adr/0002-headless-structured-output-contract.md:222` and `tests/test_error_registry.py:103-137` repeat `41`, `5`, `91`, `20 of 23`, and `27 rows`, but no generated relationship or consistency assertion protects them. This is the remaining DRY/ownership defect in an otherwise directed design: `ERROR_CODES` owns production semantics, ADR-0002 is the narrative mirror, and the test owns comparison policy. Remove volatile counts from the durable explanation or derive them in one place.

Completeness, orthogonality, dependency direction, and extension shape otherwise pass: all 91 rows and four fields are covered, no second parser was added, and production code does not depend on the test/document adapter.

### Standards

**[P3, documented-standard violation]** `RULES.md:119` requires changed facts to have one authority and repeated derived facts to be generated, referenced one-way, or consistency-tested. The unguarded snapshot counts at ADR line 222 and test lines 103–137 violate that rule. No additional Fowler baseline smell was found.

### Spec

**Pass.** Issue #701's acceptance criteria are satisfied: the complete description column is compared, provenance normalization is explicitly decided and tested, all original 15 substantive divergences have per-divergence verdicts, the three newly exposed bare-phase rows are reconciled, and exact-head CI is green. No material scope creep was found.

### Validation

- Targeted registry suite: `18 passed`
- Exact-head CI: unit/build, Ruff, Pyright, and co-install checks passed
- Base/head `gda schema`: byte-identical
- Godot E2E: skipped; proportionate for this documentation/source-description/test-only slice

qinhaihong-red:
## Review round 3 handled — `38de2738`

Reproduced every figure against `3d035725` before deciding anything. The reviewer's counts and its naming of the three subtracting rows are correct:

```
ADR rows: 91 | registry: 91
ADR descriptions carrying a strippable citation: 41
registry descriptions carrying one: 5  (ambiguous_class_name, invalid_params,
                                        script_aborted, script_failed, usage_error)
differ after BASIC normalization only: 38
differ after the citation normalization: 0
41 - 38 = 3 -> ambiguous_class_name, invalid_params, usage_error
```

One refinement the finding did not spell out, and which the corrected text now carries: **five** registry descriptions carry a citation but only **three** subtract. `script_aborted` and `script_failed` carry one on both sides too — but a *different* one (`(#655)` vs `(ADR-0031 amendment, #655)`), so they still differ under basic normalization, stay in the 38, and are resolved by the rule rather than by already matching.

| # | Finding | Verdict | Evidence, change, validation |
|---|---|---|---|
| 1 (P2) | PR body line 29 conflates two counts: "0 substantive, 41 rows differing by trailing provenance only" | **Fully adopt** | Confirmed: 38 rows differ after basic normalization, not 41 — 41 is a property of the ADR side alone. **Fixed @38de2738**: the body now carries a three-row table (38 differ after basic normalization / 41 ADR descriptions carry a citation / 0 differ after the citation rule), states that 3 of the 41 never entered the 38 and names them, and adds the five-vs-three refinement above. A PR whose subject is measurement accuracy cannot carry a conflated count in its own body. |
| 2 (P3) | `docs/adr/0002-…md:222` asserts "41 rows here carry one; so do 5 registry descriptions" — counts the guard does not hold | **Fully adopt** | Confirmed the failure mode: citing a decision on a new row falsifies that sentence while the pin stays green, because citations are normalized away by design. **Fixed @38de2738**: replaced with "Many rows here carry one, and some registry descriptions do too: that is allowed, not a mistake to tidy away." The sentence's actual job — stopping a maintainer from tidying away one of the registry citations as noise — survives intact without numbers. The exact snapshot moves to the PR body. |
| 3 (P3 / RULES.md:119) | Volatile counts duplicated with no owner | **Partially adopt** | **Adopted — the duplication.** After finding 2 the counts have one owner: the test comment, which both the ADR and this PR already name as the single home of the rule, plus a dated snapshot in the PR body. No new machinery. **Adopted — the staleness marker**, minimally: the test comment gains one clause, "Counts below are a 2026-08-28 snapshot, not a contract — only the pin itself is asserted." **Not adopted — stripping the numbers from the test comment's rationale**, for two reasons below. |

### Not adopted, with reasons

**Stripping the counts out of the test comment's rationale.** Two grounds, both evidence rather than taste. First, after finding 2 that comment *is* the single owner, so holding the numbers there is what RULES.md's one-authority rule asks for, not a violation of it. Second and decisive: the "20 of the 23 rows … the other three are ENVIRONMENT-category (`live_unsupported_platform`, `live_windowed_unavailable`, `live_windowed_permission_denied`)" precision exists *because* a vague version of that same claim concealed a real hole one round ago — I had written "the `Category` column pins it", which is true for 20 rows and false for 3, and it was replaced with the exact split for that reason. Reverting to "many rows" would undo that correction. Note the three exceptions are named as stable identifiers rather than counted, which is the half that cannot go stale.

**Deriving and asserting the counts in the test** (the alternative to finding 2). Considered and rejected. An assertion that exactly 41 rows carry a citation fails the day someone legitimately cites a decision on a new row — churn with no safety, because the count is a snapshot and the contract is row-for-row equality. The pin already asserts the thing that matters; a count assertion would only add a tripwire on correct behaviour.

### Validation on `38de2738`

Gates run before these numbers were written, output quoted verbatim:

| Gate | Result |
|---|---|
| `uv run pytest -q -m "not e2e"` | `1903 passed, 539 deselected in 70.48s (0:01:10)` |
| `uv run ruff format --check .` | `472 files already formatted` |
| `uv run ruff check .` | `All checks passed!` |

No code change in this round — the regex, the token set, and all 91 reconciled rows are untouched; the diff is three prose edits.

qinhaihong-red:
Re-reviewed the fixes at exact head `38de27388c64e3ef374198b004e4b0f553af3610` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

## Conclusion: Pass

No actionable residue remains.

- The ADR now uses qualitative wording instead of maintaining drift-prone exact registry counts.
- The remaining exact numbers are explicitly a dated, non-contractual test snapshot; the line-by-line equality assertion remains the actual contract owner.
- Independent recomputation found 91 rows, 38 basic diffs, 41 ADR citation rows, 5 registry citation rows, and 0 post-normalization diffs.
- `tests/test_error_registry.py`: 18 passed.

The required exact-head CI checks are green. Godot E2E is skipped for this scope. The head was re-pinned immediately before publishing and remained unchanged.

PR #756, opened 2026-08-27
fix(gda): stop the script-error parser eating a genuine trailing dot

### LINKED ISSUE #698: Script-error parser: the sentence-period strip misreads dot-terminated res:// paths

qinhaihong-red:
## Summary

`gda.script_errors`'s `_CANT_LOAD` and `_FAILED_LOADING_RESOURCE` regexes carried two independent defects for a `res://` address with an unusual shape.

**1. `_CANT_LOAD` corrupted a genuinely dot-terminated path.** It stripped an optional trailing period on the assumption it was always sentence punctuation:

```
>>> _CANT_LOAD.match("Can't load script: res://..").group("path")
'res://.'
```

That mismatches the canonical entry `res://..`, so an entry route reaching this parser would report a phantom success (exit 0) for a run that never happened. `#693` already closed gda's pre-launch refusal for the root/escape shapes that could trigger this for `script run`, but the parser weakness itself remained for the other live consumers: dependency/diagnostic paths in `script run` results, the timeout path (#655), and scene-preflight (#664).

**2. Both regexes rejected a path containing whitespace — found by external adversarial review of this PR, reproduced independently on real Godot 4.6.3.** `\S+` cannot match a space, so a project-relative path containing one (`res://missing file.`) failed the WHOLE line, not just the trailing character:

```
$ gda script run "missing file." --project <p> --json
{"path":"res://missing file.","exit_status":0,...,"diagnostics":[]}
```

No diagnostic was produced at all, so `entry_load_failure` found nothing and the run reported a phantom exit-0 success with empty diagnostics — the same class of defect as (1), on a different boundary. **Pre-existing, not introduced by this PR**: base `04d3e089` phantom-succeeds identically, since its original `\S+?\.?$` could not match a space either. **Masked whenever the path also carries a recognized extension** — `res://missing file.gd` still resolves correctly via `_OPEN_FAILED`'s quoted, space-safe capture (`'res://missing file.gd' resulted in error 'File not found'`) — so a check that only exercises the `.gd` shape concludes there is no bug.

Per-site fixes, verified against Godot **4.6.3-stable** (an earlier pass in this PR mis-cited a local `../godot` checkout that turned out to be a development tree 3260 commits past 4.6.3-stable; corrected to match the fixtures, which are captured against the real 4.6.3 binary), not a general longest-candidate heuristic (explicitly declined for this issue):

- `_CANT_LOAD` mirrors `main.cpp:4271` — `"Can't load script: " + script`, which **never** appends a period. Now `re.compile(r"^Can't load script: (?P<path>.+)$")` — the optional period strip is dropped, and `.+` replaces `\S+` so a space no longer fails the whole match.
- `_FAILED_LOADING_RESOURCE` mirrors `resource_loader.cpp:343` — `vformat("Failed loading resource: %s.", p_path)`, which **always** appends exactly one period. Now `re.compile(r"^Failed loading resource: (?P<path>.+)\.$")` — the strip is mandatory (fails closed on a line lacking the guaranteed period) and `.+` replaces `\S+` for the same whitespace reason. `gda.engine_log` already splits stderr into lines before either regex sees one, so `.+` cannot run past the line it belongs to.

Confirmed `resource_loader.cpp:327`'s unpunctuated `print_verbose` variant of the same sentence cannot reach this parser: it emits no `ERROR:` record, and `gda.engine_log.parse_errors` only recognizes `ERROR:`/`WARNING:`/`SCRIPT ERROR:`/`SHADER ERROR:` headers.

### A finding worth flagging

Unlike `_CANT_LOAD`, `_FAILED_LOADING_RESOURCE`'s old *optional* period strip was never actually corrupted on any well-formed captured line, for any number of trailing dots — verified both mathematically and by exhaustive check. The lazy `\S+?\.?$` quantifier always finds the smallest capture whose remainder is 0-or-1 chars; since the format string guarantees the message's last character is the appended period, that smallest split always lands exactly on the guaranteed period, regardless of how many dots `p_path` itself ends in. The mandatory strip is still correct and adopted per this issue — it makes the regex fail closed on a line that lacks the guaranteed period, and makes the invariant explicit rather than relying on an accidental-but-correct quantifier interaction — but it is not fixing a live corruption bug the way the `_CANT_LOAD` dot-strip was. (The whitespace defect above is a separate, independent bug that *does* affect both regexes equally — the mathematical argument here is about the period strip only.)

## Test plan

- [x] `uv run pytest -q -m "not e2e"` — `1913 passed, 539 deselected in 70.46s (0:01:10)`
- [x] `uv run ruff check .` — `All checks passed!`
- [x] `uv run ruff format --check .` — `472 files already formatted`
- [x] `uv run pyright` — `0 errors, 0 warnings, 0 informations`
- [x] Red-proof (dot-termination): stashed the source fix, reran the parser tests — 5 failed against pre-fix code (4 `_CANT_LOAD` dot-terminated regressions + 1 `_FAILED_LOADING_RESOURCE` fail-closed regression); popped the stash and confirmed all pass again.
- [x] Red-proof (whitespace): stashed the source fix again after adding the whitespace fixtures and the public-verdict test — 4 failed against pre-fix code (2 parser-level `_CANT_LOAD` whitespace cases, 1 `_FAILED_LOADING_RESOURCE` whitespace round-trip, and `test_a_whitespace_entry_path_is_a_failure_despite_the_zero_exit`, which reproduced the exact phantom-success shape from the live repro: `ScriptRunResult(exit_status=0, diagnostics=[])` instead of a `Failure`); popped the stash and confirmed all pass again.
- [ ] Godot e2e — not run; not needed for this parser-level slice per the dispatch brief (the fix has no other CLI-surface change; `#693` already owns the entry-route e2e coverage).

New regression fixtures for `_CANT_LOAD` (`res://..`, `res://weird..`, and `res://missing file.`) and the whitespace `_FAILED_LOADING_RESOURCE` case (`res://missing file.gd`) are captured verbatim from real `godot --headless --script "..."` runs (Godot 4.6.3, macOS), per the file's fixture convention. The dot-terminated `_FAILED_LOADING_RESOURCE` fixture is hand-authored (a real run cannot produce that exact combination — `ResourceFormatLoader::recognize_path` suffix-matches a registered extension before that sentence is reachable, and no registered extension is itself a bare `.`; probed directly and confirmed), documented as such in the test comments.

Also adds a **public-verdict** regression at the operation level (`tests/test_script_run_operation.py::test_a_whitespace_entry_path_is_a_failure_despite_the_zero_exit`) that drives `run_script_run_operation` end-to-end through its real validate → launch → classify recipe and asserts the outcome flips from a phantom `ScriptRunResult` success to the registered `script_compile_failed` `Failure` — not inferred from the regex matching alone. Merges the `_CANT_LOAD` dot-terminated parametrize matrix (previously duplicated across two test functions) into one shared list, folding the new whitespace case in rather than adding a third copy.

The ADR-0031 outcome note is updated to describe what the code now does (captures to end of line unconditionally, via `.+`) and records the whitespace hole as pre-existing rather than introduced by this PR.

## Consumers checked for a pin on the old behavior

Searched every consumer named in the issue (`gda script run`, its timeout path, scene-preflight) plus every test file referencing `Can't load script:` / `Failed loading resource:`. No test asserts the corrupted (pre-fix) parse as expected behavior. Three explanatory comments described the old `_CANT_LOAD` dot-strip corruption as the *reason* `script run` refuses root/escape paths before any launch, and went stale after that fix:

- `src/gda/commands/script.py` (~1174-1177) — inside `#713`'s ownership; left untouched, routed to that slice instead of edited here.
- `tests/test_script_run_operation.py` (two spots) and `tests/test_e2e_script_run.py` (one spot, caught in a follow-up pass) — unowned by any slice this wave, so rewritten in place to read as history (`#698 fixed the parser …`) without implying the parser still corrupts; the refusal itself is unchanged and stays load-bearing regardless, independent of the parser (ADR-0009's Trusted project).

## Update: a fourth defect, one layer above the parser (trailing whitespace)

External adversarial review of this PR found a fourth defect, reproduced independently on real Godot 4.6.3, one level up from the parser fixes above. It survives the `.+` capture fix: it is not a parsing gap, it is a canonical-identity mismatch.

```
$ gda script run "missing file. " --project <p> --json
{"path":"res://missing file. ","exit_status":0,"stdout":"...","diagnostics":[]}
   stderr contains: ERROR: Can't load script: res://missing file.
```

Godot strips trailing whitespace from the `--script` argv *before* it ever echoes the path back in a load-failure diagnostic, so the engine's address (`res://missing file.`) loses what gda's own canonical spelling keeps (`res://missing file. `, since `canonical_res_path` is purely lexical `posixpath.normpath` and does not strip it). The canonical-identity comparison the never-ran verdict depends on then misses on that one trailing character, and the run reports the same phantom success (exit 0, empty `diagnostics`) the root/escape refusals in `_project_scoped_res_path` already exist to prevent — on a path neither of them admits. `missing file.gd ` has the same shape (masked from the parser-level fixture work above by `_OPEN_FAILED`'s quoted, space-safe capture, but not from this mismatch).

`_project_scoped_res_path` (`src/gda/commands/script.py`) now refuses exactly the suffix set Godot 4.6.3 removes before reporting the path: a trailing code point at or below U+0020. It does not trim the caller's address, because that would launch a different script. It also refuses any canonical address containing a line boundary recognized by `gda.engine_log`; such a character splits the emitted address across records, so no diagnostic retains the entry identity.

Scoped to the two verified identity-loss mechanisms: trailing U+0000..U+0020 and engine-log line boundaries are refused. Leading and internal ASCII spaces remain accepted. NBSP (U+00A0) and EM SPACE (U+2003) also remain accepted because Godot preserves them; this explicitly avoids Python `str.rstrip()`, whose Unicode whitespace set is wider than the engine's.

- [x] Public-operation regressions: the no-launch matrix covers the engine-trim suffixes and all line-boundary characters recognized by Python `splitlines`; separate unit and real-engine cases prove NBSP/EM SPACE are not over-rejected.
- [x] `uv run pytest -q -m "not e2e"` — `1919 passed, 539 deselected in 152.96s (0:02:32)`
- [x] `uv run ruff check .` — `All checks passed!`
- [x] `uv run ruff format --check .` — `472 files already formatted`
- [x] `uv run pyright` — `0 errors, 0 warnings, 0 informations`
- [x] Red-proof: committed the fix (`feff4381`), reverted only `src/gda/commands/script.py` to the parent commit, reran — 4 failed, each reproducing a real phantom `ScriptRunResult` (e.g. `ScriptRunResult(path='res://tests/logic.gd\t', exit_status=0, ..., diagnostics=[])`) instead of the expected `Failure`; restored the fix and confirmed all pass again.
- [x] Manual repro against real Godot 4.6.3 (`gda script run`, both path forms, with and without a `.gd` extension) before and after the fix, plus leading/internal-whitespace non-regression.

This defect also exists on the base (`origin/feat/gda-dogfooding-dev`, unaffected by the parser fixes above), but it is the same class as the whitespace P1 this PR already closes, so it closes here too.

Closes #698

## Update: exact Godot trim set and engine-log line boundaries

The round-3 re-review found that Python `str.rstrip()` both under-specified the diagnostic boundary and over-rejected Unicode spaces. Commit `517a3071` replaces it with the exact Godot 4.6.3 trailing rule (code point <= U+0020) and reuses `gda.engine_log.lines` to reject any address that cannot remain one diagnostic record. ADR-0031 now owns the complete pre-launch refusal set and the exact terminology.

Validation at `517a3071`:

- `tests/test_script_error_parser.py tests/test_script_run_operation.py tests/test_script_run_commands.py`: **203 passed**.
- Real Godot 4.6.3 regressions for the internal-LF refusal and NBSP/EM SPACE non-overreach: **13 passed**.
- Ruff check/format: pass.
- Pyright: **0 errors, 0 warnings, 0 informations**.

This section supersedes the earlier broad "trailing whitespace" and "leading/internal whitespace" wording. The implemented contract is expressed only in terms of Godot's exact trim set and the engine-log line protocol.

### DIFF
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -99,12 +99,17 @@ status: accepted
 >
 > **What stays refused**, all decided before any launch as `invalid_path`: an **absolute** path;
 > **another engine scheme** (`user://`, `uid://` — lifting one would splice a second scheme into a
-> res:// address and send the engine after a path nobody typed); a path naming the project **root**
-> (`""`, `"."`, `"sub/.."`, and the `res://` / `res://.` spellings — a directory, not a script); and a
-> path **escaping above the root** (`".."`, `"../outside.gd"`, and their `res://` spellings). The ABI
-> edge below names "a non-`res://` **or absolute** script path"; only its first half is lifted here.
+> res:// address and send the engine after a path nobody typed); a leading `~` (a filesystem HOME
+> reference, including an unresolvable one); a path naming the project **root** (`""`, `"."`,
+> `"sub/.."`, and the `res://` / `res://.` spellings — a directory, not a script); a path **escaping
+> above the root** (`".."`, `"../outside.gd"`, and their `res://` spellings); a canonical address
+> ending in a code point at or below **U+0020**, which Godot removes before reporting the path; and a
+> canonical address containing any line boundary recognized by `gda.engine_log`'s line protocol.
+> The ABI edge below names "a non-`res://` **or absolute** script path"; only its first half is lifted
+> here.
 >
-> The last two are refused for a reason beyond tidiness, and both were **verified**. The engine
+> The root and escape shapes are refused for a reason beyond tidiness, and both were **verified**. The
+> engine
 > answers a root or escape address with `Can't load script: res://.` / `res://..`, whose address the
 > error parser reads back with the sentence period stripped — so it never matches the entry, the
 > never-ran verdict misses it, and the run reports a phantom `exit_status: 0`. Worse, an escape that
@@ -115,6 +120,15 @@ status: accepted
 > the widening is what made them reachable from the ordinary project-relative form, so they are closed
 > here. The residual parser weakness (a trailing-period strip that mis-reads `res://..`) is **not**
 > addressed here; it belongs to the error parser and is tracked separately.
+>
+> **Path-identity boundary (2026-08-28, #698 review).** The two character rules above preserve the
+> same canonical identity on both sides of the entry-load verdict. Godot 4.6.3's
+> `String::strip_edges()` removes trailing code points at or below U+0020; Python `str.rstrip()` is
+> intentionally not the rule because it additionally removes Unicode spaces such as NBSP (U+00A0)
+> and EM SPACE (U+2003), which Godot preserves. Separately, `gda.engine_log` parses engine output with
+> `str.splitlines()`. Any address containing one of those line boundaries is therefore impossible to
+> recover from one diagnostic record and is refused before launch. Leading and internal ASCII spaces,
+> and Unicode spaces that Godot preserves, remain valid project-relative characters.
 >
 > **Outcome (2026-08-27, #698):** the error parser's `_CANT_LOAD` regex (`gda.script_errors`) no
 > longer strips a genuine trailing dot. It mirrors `main.cpp:4271`, which never appends sentence
@@ -138,7 +152,8 @@ status: accepted
 >
 > The refusals this decision records stay in place regardless of either parser fix — they are
 > load-bearing for reasons beyond the parser weakness (ADR-0009's Trusted project, above) — but the
-> parser itself no longer mis-reads a dot-terminated or whitespace-containing `res://` address if a
+> parser itself no longer mis-reads a dot-terminated or horizontal-space-containing `res://` address if
+> a
 > future route ever reaches it again.
 >
 > Absolute stays refused for two **verified** reasons, not merely as deferred scope. First, the engine
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -116,16 +116,18 @@ status: accepted
 > here. The residual parser weakness (a trailing-period strip that mis-reads `res://..`) is **not**
 > addressed here; it belongs to the error parser and is tracked separately.
 >
-> **Outcome (2026-08-27, #698):** the error parser's `_CANT_LOAD` and `_FAILED_LOADING_RESOURCE`
-> regexes (`gda.script_errors`) no longer strip a genuine trailing dot. `_CANT_LOAD` mirrors
-> `main.cpp:4366`, which never appends sentence punctuation, so its strip is now unconditional
-> removal — dropped entirely. `_FAILED_LOADING_RESOURCE` mirrors `resource_loader.cpp:317`'s
-> `vformat("Failed loading resource: %s.", p_path)`, which always appends exactly one period, so its
-> strip is now mandatory rather than optional. Both are per-site fixes, not a general
-> longest-candidate heuristic. The refusals this decision records stay in place regardless — they are
-> load-bearing for reasons beyond the parser weakness (ADR-0009's Trusted project, above) — but the
-> parser itself no longer mis-reads a dot-terminated `res://` address if a future route ever reaches
-> it again.
+> **Outcome (2026-08-27, #698):** the error parser's `_CANT_LOAD` regex (`gda.script_errors`) no
+> longer strips a genuine trailing dot. It mirrors `main.cpp:4366`, which never appends sentence
+> punctuation, so any strip was always wrong; the strip is dropped entirely, so `Can't load script:
+> res://..` now round-trips to `res://..` instead of the `res://.` mis-read this paragraph describes.
+> The sibling `_FAILED_LOADING_RESOURCE` regex was tightened too (its strip is now mandatory rather
+> than optional, mirroring `resource_loader.cpp:317`'s guaranteed trailing period), but — verified —
+> its old optional strip never actually mis-read a well-formed captured line, for any dot count; that
+> change is a fail-closed robustness improvement, not a fix for observed corruption. Neither change is
+> a general longest-candidate heuristic. The refusals this decision records stay in place regardless
+> — they are load-bearing for reasons beyond the parser weakness (ADR-0009's Trusted project, above)
+> — but the parser itself no longer mis-reads a dot-terminated `res://` address if a future route
+> ever reaches it again.
 >
 > Absolute stays refused for two **verified** reasons, not merely as deferred scope. First, the engine
 > reports a failed run under the **`res://` spelling even when launched with an absolute in-project
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -117,11 +117,11 @@ status: accepted
 > addressed here; it belongs to the error parser and is tracked separately.
 >
 > **Outcome (2026-08-27, #698):** the error parser's `_CANT_LOAD` regex (`gda.script_errors`) no
-> longer strips a genuine trailing dot. It mirrors `main.cpp:4366`, which never appends sentence
+> longer strips a genuine trailing dot. It mirrors `main.cpp:4271`, which never appends sentence
 > punctuation, so any strip was always wrong; the strip is dropped entirely, so `Can't load script:
 > res://..` now round-trips to `res://..` instead of the `res://.` mis-read this paragraph describes.
 > The sibling `_FAILED_LOADING_RESOURCE` regex was tightened too (its strip is now mandatory rather
-> than optional, mirroring `resource_loader.cpp:317`'s guaranteed trailing period), but — verified —
+> than optional, mirroring `resource_loader.cpp:343`'s guaranteed trailing period), but — verified —
 > its old optional strip never actually mis-read a well-formed captured line, for any dot count; that
 > change is a fail-closed robustness improvement, not a fix for observed corruption. Neither change is
 > a general longest-candidate heuristic. The refusals this decision records stay in place regardless
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -116,6 +116,17 @@ status: accepted
 > here. The residual parser weakness (a trailing-period strip that mis-reads `res://..`) is **not**
 > addressed here; it belongs to the error parser and is tracked separately.
 >
+> **Outcome (2026-08-27, #698):** the error parser's `_CANT_LOAD` and `_FAILED_LOADING_RESOURCE`
+> regexes (`gda.script_errors`) no longer strip a genuine trailing dot. `_CANT_LOAD` mirrors
+> `main.cpp:4366`, which never appends sentence punctuation, so its strip is now unconditional
+> removal — dropped entirely. `_FAILED_LOADING_RESOURCE` mirrors `resource_loader.cpp:317`'s
+> `vformat("Failed loading resource: %s.", p_path)`, which always appends exactly one period, so its
+> strip is now mandatory rather than optional. Both are per-site fixes, not a general
+> longest-candidate heuristic. The refusals this decision records stay in place regardless — they are
+> load-bearing for reasons beyond the parser weakness (ADR-0009's Trusted project, above) — but the
+> parser itself no longer mis-reads a dot-terminated `res://` address if a future route ever reaches
+> it again.
+>
 > Absolute stays refused for two **verified** reasons, not merely as deferred scope. First, the engine
 > reports a failed run under the **`res://` spelling even when launched with an absolute in-project
 > path**, so accepting absolute without also mapping it back to `res://` would break the
--- a/docs/adr/0031-headless-script-run-passthrough-execution.md
+++ b/docs/adr/0031-headless-script-run-passthrough-execution.md
@@ -118,16 +118,28 @@ status: accepted
 >
 > **Outcome (2026-08-27, #698):** the error parser's `_CANT_LOAD` regex (`gda.script_errors`) no
 > longer strips a genuine trailing dot. It mirrors `main.cpp:4271`, which never appends sentence
-> punctuation, so any strip was always wrong; the strip is dropped entirely, so `Can't load script:
-> res://..` now round-trips to `res://..` instead of the `res://.` mis-read this paragraph describes.
-> The sibling `_FAILED_LOADING_RESOURCE` regex was tightened too (its strip is now mandatory rather
-> than optional, mirroring `resource_loader.cpp:343`'s guaranteed trailing period), but — verified —
-> its old optional strip never actually mis-read a well-formed captured line, for any dot count; that
+> punctuation, so any strip was always wrong; the strip is dropped, so `Can't load script: res://..`
+> now round-trips to `res://..` instead of the `res://.` mis-read this paragraph describes. The
+> sibling `_FAILED_LOADING_RESOURCE` regex was tightened too (its strip is now mandatory rather than
+> optional, mirroring `resource_loader.cpp:343`'s guaranteed trailing period), but — verified — its
+> old optional strip never actually mis-read a well-formed captured line, for any dot count; that
 > change is a fail-closed robustness improvement, not a fix for observed corruption. Neither change is
-> a general longest-candidate heuristic. The refusals this decision records stay in place regardless
-> — they are load-bearing for reasons beyond the parser weakness (ADR-0009's Trusted project, above)
-> — but the parser itself no longer mis-reads a dot-terminated `res://` address if a future route
-> ever reaches it again.
+> a general longest-candidate heuristic.
+>
+> Adversarial review of the #698 PR found a second, independent defect in the same two regexes: their
+> path capture was `\S+`, which cannot match a SPACE, so a project-relative path containing one
+> (`res://missing file.`) failed the WHOLE line rather than losing a character — no diagnostic at all,
+> so `entry_load_failure` found nothing and the run reported a phantom success. Pre-existing on `main`
+> before #698 (its `\S+?\.?$` could not match a space either) and masked whenever the path also
+> carried a recognized extension (`_OPEN_FAILED`'s quoted capture is space-safe), so it surfaces only
+> on an extension-less spelling. Both regexes now capture `.+` — `gda.engine_log` has already split
+> stderr into lines before either regex sees one, so `.+` cannot run past the line it belongs to — and
+> genuinely capture to end of line unconditionally, which is what this outcome note now describes.
+>
+> The refusals this decision records stay in place regardless of either parser fix — they are
+> load-bearing for reasons beyond the parser weakness (ADR-0009's Trusted project, above) — but the
+> parser itself no longer mis-reads a dot-terminated or whitespace-containing `res://` address if a
+> future route ever reaches it again.
 >
 > Absolute stays refused for two **verified** reasons, not merely as deferred scope. First, the engine
 > reports a failed run under the **`res://` spelling even when launched with an absolute in-project
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -1150,7 +1150,7 @@ def _project_scoped_res_path(script: str) -> str | None:
     :func:`canonical_res_path`, so the argv, the entry-load verdict and the reported
     path cannot diverge by input spelling.
 
-    Returns ``None`` for the five shapes that are not project-scoped script
+    Returns ``None`` for the six shapes that are not project-scoped script
     addresses. Each must be caught HERE, because each is otherwise launched:
 
     - an **absolute** path — outside the ``--project`` context (the reasons it stays
@@ -1163,6 +1163,16 @@ def _project_scoped_res_path(script: str) -> str | None:
       expand it (an unknown user, #699); a resolvable ``~/x.gd`` was already expanded
       to the absolute path refused above. So both tilde outcomes land on one refusal
       instead of one being refused and the other spliced into ``res://~user/x.gd``;
+    - a path with **trailing whitespace** (``"missing file. "``, ``"missing
+      file.gd "``) — the engine strips trailing whitespace from the ``--script``
+      argv before it ever echoes the path back in a load-failure diagnostic
+      (verified against real Godot 4.6.3), so the address it reports loses
+      whitespace gda's own canonical spelling keeps. The canonical-identity match
+      the never-ran verdict depends on then misses on that one trailing character,
+      and the run reports the same PHANTOM SUCCESS the root/escape refusals below
+      exist to prevent — on a path neither of them admits (#698). Only the
+      TRAILING case is refused: leading and internal whitespace survive the
+      engine's own normalization unchanged and stay accepted;
     - a path that names the project **root** (``""``, ``"."``, ``"sub/.."``) — it names
       a directory, not a script. An unset shell variable makes ``gda script run
       "$SCRIPT"`` exactly this;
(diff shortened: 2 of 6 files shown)


qinhaihong-red:
Adversarial review of exact head `e269800e6299a3d1d425a3acb0bbbbfd034bbd95` against merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`. I re-read the PR head after review; it is unchanged.

## Artifact Review — Changes required

**[P1] The closure claims exceed the implemented path grammar.**

The PR body says that `_CANT_LOAD` now captures to the end of the line unconditionally, and ADR-0031 lines 119–130 records that the parser no longer misreads a dot-terminated `res://` address. However, `src/gda/script_errors.py:96` still uses `\S+`, although Godot's source shape is `"Can't load script: " + script` and valid resource paths can contain spaces.

Godot 4.6.3 emits this verbatim for a valid dot-terminated address:

```text
ERROR: Can't load script: res://missing file.
   at: start (main/main.cpp:4271)
```

At this exact head, `parse_script_errors(...) == []` and `entry_load_failure(...) is None`; a real `gda script run 'missing file.'` returns process exit 0 with `exit_status: 0` and `diagnostics: []`, although the entry script never ran. The issue should not be recorded as closed until this grammar boundary is covered.

**Terminology and prose assessment:** I inspected the complete current ADR-0031, the complete PR body, and issue #698 including its open state, labels, correction, and comments. Sources were `CONTEXT.md`, ADR-0031, ADR-0009, the implementation, and issue #698; ASD-STE100 was used for word choice, technical nouns and verbs, causal/scope statements, and sentence structure. Apart from the contradicted closure claim above, I found no substantive terminology or prose defect.

## Architecture / DDMA — Changes required

**[P1] `gda.script_errors` is the correct single owner of engine sentence classification and canonical resource identity, but its ownership is incomplete for the engine's actual path grammar.**

The dependency direction remains sound:

```text
commands/script + commands/scene
                ↓
          script_errors
                ↓
            engine_log
```

There is no reverse dependency or parallel parser. However, both touched patterns retain an undocumented “no whitespace in a path” rule:

- `src/gda/script_errors.py:96`
- `src/gda/script_errors.py:152`

The official Godot 4.6.3 producers accept the complete `String` path (`"..." + script` and `%s.`), not a non-whitespace token. The first restriction produces the public phantom success demonstrated above; the second drops a valid `Failed loading resource:` diagnostic containing spaces.

The smallest fix is to consume the complete non-empty single-line remainder:

```python
_CANT_LOAD = re.compile(r"^Can't load script: (?P<path>.+)$")
_FAILED_LOADING_RESOURCE = re.compile(
    r"^Failed loading resource: (?P<path>.+)\.$"
)
```

`engine_log` already owns line splitting, so this does not introduce a cross-line heuristic. Add a real Godot 4.6.3 `res://missing file.` fixture and public-verdict regression.

Completeness fails on this boundary. Orthogonality, single-authority DRY, dependency direction, and the existing consumer extension path otherwise pass. The project defines no claim-evidence-state vocabulary, so the full-grammar closure claim remains `open`; the runtime counterexample is its current disconfirming evidence.

## Standards — 2 findings

**[P3, documented terminology violation]** `tests/test_script_error_parser.py:500` says “`#693 closed the CLI entry route`”. `CONTEXT.md:11-14` defines the component name as **gda** and explicitly avoids “the CLI” as its substitute. Use “`#693 closed the gda script run entry route`”.

**[P3, judgement call — Duplicated Code]** `tests/test_script_error_parser.py:477-507` repeats the same `("stderr", "path")` parameter matrix and `parse_script_errors(stderr)` setup in adjacent tests. Adding another boundary fixture can make the lists drift. Merge the assertions into one parameterized test or extract one shared case list.

No other non-tool-enforced Standards finding was found.

## Spec — 1 finding

**[P1] `src/gda/script_errors.py:96` still rejects a valid path containing whitespace.**

Issue #698 requires a diagnostic naming a genuinely dot-terminated path to round-trip and the canonical-identity verdict not to produce phantom success. Its examples do not restrict the valid path character set, and the specified engine shape is `"Can't load script: " + script`.

For `res://missing file.`, Godot 4.6.3 emits the exact line shown above, while `parse_script_errors(...) == []` and `entry_load_failure(...) is None`. The acceptance criterion is therefore satisfied only for the two new whitespace-free fixtures. Use a complete line-remainder capture and extend the round-trip, verdict, and public `gda script run` cases with the captured fixture.

The remaining Spec requirements pass, and I found no additional scope creep.

Standards summary: 2 findings, worst P3. Spec summary: 1 finding, worst P1.

## Validation and limits

- Exact-head targeted tests: `137 passed`
- Real Godot `4.6.3.stable.official.7d41c59c4` runs confirmed the two new dot fixtures and reproduced the whitespace counterexample
- Official `4.6.3-stable` source confirms `main.cpp:4271`, `resource_loader.cpp:343`, and the non-error `print_verbose` variant at line 327
- Exact-head CI run `33081324296`: unit/build, Ruff, Pyright, and co-install passed; Godot E2E was skipped
- I did not rerun the complete local `1909 passed` suite claimed in the PR body

qinhaihong-red:
Addressed all three findings from this review round. Fixes are at `d2f24d45`.

| # | Finding | Decision | Evidence | Change | Validation |
|---|---|---|---|---|---|
| 1 (P1) | `\S+` rejects a `res://` path containing whitespace — a live phantom success (`gda script run "missing file."` → exit 0, empty `diagnostics`) | **Fully adopted** | Reproduced independently on real Godot 4.6.3 (`godot --headless --script "res://missing file."` → `ERROR: Can't load script: res://missing file.` with **no** matching diagnostic pre-fix). Confirmed **pre-existing**: base `04d3e089`'s `\S+?\.?$` cannot match a space either — this PR did not introduce it. Confirmed the **`.gd`-masking** mechanism: with a recognized extension, `_OPEN_FAILED`'s quoted `[^']*` capture is space-safe and still resolves `SCRIPT_MISSING` correctly, so a check limited to the `.gd` shape concludes there is no bug — only the extension-less spelling falls through to `_CANT_LOAD` alone | `_CANT_LOAD` and `_FAILED_LOADING_RESOURCE` now capture `.+` instead of `\S+` (`gda.engine_log` already splits stderr into lines before either regex sees one, so `.+` cannot run past its own line); fail-closed property of the mandatory period on `_FAILED_LOADING_RESOURCE` verified to survive (a line missing the guaranteed period still returns no match) | Real captured Godot 4.6.3 fixtures for the spaced case, both regexes, at the parser level (`tests/test_script_error_parser.py`); a **public-verdict** regression at the operation level (`tests/test_script_run_operation.py::test_a_whitespace_entry_path_is_a_failure_despite_the_zero_exit`) driving `run_script_run_operation` end-to-end and asserting the outcome flips from the phantom `ScriptRunResult(exit_status=0, diagnostics=[])` to the registered `script_compile_failed` `Failure` — not inferred from the regex alone. Red-proofed: 4 new tests fail against pre-fix code (2 parser-level whitespace cases, 1 `_FAILED_LOADING_RESOURCE` whitespace round-trip, 1 public-verdict), all pass after |
| 2 (P3) | `tests/test_script_error_parser.py:500` says "the CLI entry route" — `CONTEXT.md` lists "the CLI" under `gda`'s `_Avoid_` | **Fully adopted** | `CONTEXT.md`'s `gda` glossary entry: `_Avoid_: the CLI, godot-cli` | Reworded to "gda's pre-launch entry-path refusal for the root/escape shapes" (and swept the same phrase out of the PR body itself, where it also appeared twice) | Comment-only change; covered by the same test run |
| 3 (P3) | Duplicated parameter matrix across two adjacent `_CANT_LOAD` tests | **Adopted, folded into fix 1** | The two `@pytest.mark.parametrize` decorators carried byte-identical `(stderr, path)` lists | Extracted to one shared `_CANT_LOAD_BOUNDARY_CASES` list, referenced by both tests; the new whitespace case was added there once instead of as a third copy | Same test run |

Gates on `d2f24d45`: `uv run pytest -q -m "not e2e"` → `1913 passed, 539 deselected in 70.46s (0:01:10)`; `uv run ruff check .` → `All checks passed!`; `uv run ruff format --check .` → `472 files already formatted`; `uv run pyright` → `0 errors, 0 warnings, 0 informations`. PR CI green on this head.

Summary for anyone re-checking: the P1 was **pre-existing on `main`**, not introduced by this PR's dot-termination fix — but this PR now closes it (previously it only closed the dot-termination corruption). The **`.gd`-masking** detail is why a casual re-check that only tries an extensioned path will find nothing wrong: only the extension-less spelling reaches `_CANT_LOAD` (or, for `_FAILED_LOADING_RESOURCE`, only an extension-less path skips it) with no other sentence to fall back on.

qinhaihong-red:
Re-reviewed the fixes at exact head `d2f24d45085fefc535d4c2012dbddc39aecbee31` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

The original internal-whitespace parser failure and both Standards findings are closed. One adjacent whitespace-boundary residue remains.

### [P1] Reject trailing-whitespace script paths before launch

A real Godot 4.6.3 replay of `script run "missing file. " --json` returns exit 0 with success path `res://missing file. `, while the diagnostic path is `res://missing file.` and the script never ran. `missing file.gd ` has the same failure shape. Godot normalizes away the trailing whitespace, but the entry-verdict comparison retains it, so the load failure does not match and the public operation reports phantom success.

This defect also exists in the base, but it is an adjacent case of the previous P1 whitespace grammar and public-verdict closure. The smallest safe fix is to reject a spelling with trailing whitespace in `_project_scoped_res_path` before launch and add a public-operation regression test. Do not silently `rstrip()` the path: that could execute a different script from the one the caller named.

Other replayed boundaries pass, including internal spaces, leading spaces, multiple trailing dots, the `.gd` masking variants, and fail-closed malformed log shapes. Targeted exact-head tests: 141 passed. Required CI checks are green; Godot E2E is skipped, so the residue above was verified with a separate real-engine replay.

qinhaihong-red:
Addressed the trailing-whitespace P1 from this review round. Fix is at `feff4381`.

| # | Finding | Decision | Evidence | Change | Validation |
|---|---|---|---|---|---|
| 1 (P1) | Trailing whitespace in a `script run` path phantom-succeeds: `"missing file. "` / `"missing file.gd "` return exit 0 with the entry never run, because Godot strips the trailing whitespace before it echoes the path in a diagnostic, while gda's own canonical spelling keeps it | **Fully adopted** | Reproduced independently on real Godot 4.6.3, pre- and post-fix. Pre-fix: `gda script run "missing file. "` → `ScriptRunResult(path='res://missing file. ', exit_status=0, diagnostics=[...])` — a success shape for a script that never ran. Confirmed **pre-existing on the base** (`origin/feat/gda-dogfooding-dev`), same class as the whitespace P1 this PR already closes at the parser level. Confirmed leading/internal whitespace are **not** affected by the engine's own normalization (`" missing file.gd"` and `"missing file.gd"` both correctly report `script_not_found`, never a phantom success, before and after this change) — narrowed the fix to trailing only | `_project_scoped_res_path` (`src/gda/commands/script.py`) now refuses a canonical remainder with trailing whitespace before any launch, returning the same `invalid_path` failure as its four sibling refusals (absolute / another scheme / unresolvable `~` / root / escape). **Not** an `rstrip()` — silently trimming would launch a *different* script than the one the caller named, which is worse than refusing; the code comment says so explicitly | Public-operation regression: `test_a_non_project_scoped_path_is_invalid_path_before_any_launch` gains 4 trailing-whitespace cases (both accepted forms, with and without a `.gd` extension, plus a non-space whitespace char), asserting the *observable* outcome (`invalid_path` `Failure`, no engine launch) — not the helper in isolation. A companion `test_leading_and_internal_whitespace_are_accepted` locks in that the fix does not overreach. Red-proofed: committed the fix, reverted only the source file to the parent commit, reran — 4 failed, each reproducing a real phantom `ScriptRunResult` (e.g. `ScriptRunResult(path='res://tests/logic.gd\t', exit_status=0, ..., diagnostics=[])`); restored the fix, all pass again |

Gates on `feff4381`: `uv run pytest -q -m "not e2e"` → `1919 passed, 539 deselected in 152.96s (0:02:32)`; `uv run ruff check .` → `All checks passed!`; `uv run ruff format --check .` → `472 files already formatted`; `uv run pyright` → `0 errors, 0 warnings, 0 informations`.

This defect exists on the base too — it is not introduced by this PR — but it is the same class as the whitespace P1 this PR already closes at the parser level, so it closes here rather than opening a separate issue.

qinhaihong-red:
Re-reviewed the claimed fixes at exact head `feff4381adc78fc783dd6d15903a506855a66bf9` against the previous finding and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

The original trailing ASCII-space phantom-success case is closed, but the adjacent public path boundary is still incorrect.

### [P2] Match Godot's path normalization and preserve diagnostic identity

`_project_scoped_res_path` now uses Python `str.rstrip()` as a proxy for Godot normalization. Exact-head real-engine probes against Godot 4.6.3 show two counterexamples:

- `missing\nfile.gd` is accepted. Godot prints the path over multiple log lines, the line-oriented parser records only `res://missing`, the canonical entry identity does not match, and `gda script run` still returns exit 0 even though the entry never ran.
- Trailing NBSP (`U+00A0`) and EM SPACE (`U+2003`) are preserved by Godot, but Python `rstrip()` rejects them pre-launch as `invalid_path`. The previous head correctly launched these paths and returned `script_compile_failed`.

Godot's `String::strip_edges()` trims code points at or below `U+0020`; Python's Unicode whitespace set is wider. Please encode the engine's actual trailing-normalization rule and reject line-breaking characters that cannot round-trip through the line-oriented diagnostic protocol. Add public-operation regressions for both boundaries.

### [P2] Synchronize ADR-0031

ADR-0031's pre-launch `invalid_path` list does not include the new engine-normalized trailing-suffix rule. Update the authoritative contract with the exact rule after the implementation is narrowed; avoid the overbroad term "whitespace."

Artifact/DDMA/Standards/Spec rechecks otherwise found no additional residue. Focused tests pass and required CI is green, but Godot E2E is skipped; the real-engine counterexamples above therefore remain decisive. The head was re-pinned before publishing and was unchanged.

qinhaihong-red:
Handled the round-3 re-review at exact head `feff4381`. New commit: `517a3071`.

| Claim | Decision | Change | Validation |
| --- | --- | --- | --- |
| Python `rstrip()` does not match Godot path normalization and leaves internal-linebreak phantom success | **Fully Adopt** | Replaced the Unicode-wide proxy with Godot 4.6.3's exact trailing rule (code point <= U+0020). Added an engine-log single-record gate by reusing `gda.engine_log.lines`, so every `str.splitlines()` boundary is refused before launch. NBSP and EM SPACE remain accepted. | 203 focused unit/integration tests passed; real Godot 4.6.3 boundary regressions passed 13/13. |
| ADR-0031 omitted the public refusal rule and used overbroad whitespace terminology | **Fully Adopt** | Synchronized ADR-0031 with the complete pre-launch set, the exact Godot trim semantics, the line-protocol boundary, and the preserved Unicode-space cases. Updated the PR body to supersede the earlier broad wording. | Artifact/terminology pass against `CONTEXT.md`, ADR-0031, and ASD-STE100; `git diff --check` clean. |

Additional gates: Ruff check/format pass; Pyright reports 0 errors, 0 warnings, 0 informations. The worktree is clean after the commit.

Current CI at `517a3071`: Ruff, Pyright, co-install, scope classification, and gda-balancing required pass; Python unit/package and the release scope guard are still running. Godot E2E is skipped in CI, with the local real-engine regression above supplying the missing boundary evidence.

**Final CI update:** all required checks pass at exact head `517a3071d25bc6c898cde666999e03772e867718`; Godot E2E and out-of-scope gda-balancing jobs are skipped as expected. Merge state is `CLEAN`.

Issue #714, opened 2026-08-18
sentinel and export channels: preserve captured output when a launch times out


qinhaihong-red:
## Parent

Named as follow-up work in issue #655 itself ("the discard lives in the shared Headless launch primitive, so the sentinel channel (60s default) and the export channel (600s default) are affected too — they are named follow-up work, not covered by this issue's ACs").

## What to build

#655 gave `script run` a streaming capture so a timeout preserves the partial stdout/stderr the engine already wrote. The other THREE launch channels still use the buffered capture, which discards everything on timeout: a sentinel op that hangs (60s), an export that hangs (600s), and a `resource import` that hangs all fail with an empty-evidence `launch_timeout` (the third channel and the shared-ownership points below were folded in from the 2026-08-31 scope-correction comment). Extend the timeout-capture preservation to all three channels through the shared `Headless launch` primitive's existing watch/streaming seam — the mechanism exists. The envelope itself is built in ONE place: `classify_launch_or_crash`'s `LaunchFailure.TIMEOUT` arm in `gda/errors.py`, the single home all three channels' classifiers open with — that shared arm is where the evidence-carrying envelope is shaped, not per-channel forks. The old buffered-timeout stderr wording is pinned byte-compatible by several tests; those pins move WITH this slice (the non-timeout byte-identity AC below is the boundary). `CONTEXT.md`'s `Headless launch` / `Raw run` entries state buffered-timeout facts this slice makes false — correct them in the same PR.

## Acceptance criteria

- [ ] A sentinel-channel timeout returns the captured partial output (tail-capped, cap stated) and elapsed seconds in its failure envelope.
- [ ] An export-channel timeout does the same.
- [ ] A `resource import` timeout does the same (third buffered channel, same shared classifier).
- [ ] Non-timeout behavior of all three channels is byte-identical (no envelope/schema drift for successful runs).
- [ ] Real-engine e2e arms cover one timeout per channel (all three).
- [ ] `CONTEXT.md`'s `Headless launch` / `Raw run` entries are corrected in the same PR.

## Blocked by

None - can start immediately (the streaming seam landed with #655 on the dev branch).

> **Body synchronized 2026-08-31 (PR #781 review):** the 2026-08-31 scope-correction comment's four points (shared `LaunchFailure.TIMEOUT` ownership, the third buffered channel `resource import`, deliberate timeout-pin migration, CONTEXT.md synchronization) are folded in above. The comment stands as history.

[...]

PR #757, opened 2026-08-27
test(gda): deflake the streaming-launch timeout test's fake-engine startup race

### LINKED ISSUE #728: Streaming-launch timeout test races the fake engine's startup and flakes under load
## Parent

#655 — the streaming capture strategy this test guards.

## What to build

Make `tests/test_launch.py::test_streaming_timeout_preserves_the_output_the_child_already_wrote` stop racing its own fake engine's startup.

The test asserts the #655 / GDA-DF-012 guarantee: a streaming `Headless launch` that hits its timeout still reports what the child had ALREADY written, on both streams, with no gda prose mixed in. Its fake engine is a Python-shebang script that prints one line to each stream and then sleeps past the deadline, and the launch timeout is `1.5s` — so the assertion holds only if the child's Python interpreter finishes starting and prints within `1.5s`.

That is a race, not a guarantee. Measured on a dev machine, time-to-first-stdout was:

| Conditions | p50 | p95 | max |
| --- | --- | --- | --- |
| unloaded, 60 runs | 57 ms | 60 ms | **596 ms** |
| 12-way CPU load, 40 runs | 56 ms | 61 ms | **968 ms** |

The typical value is far from the deadline, but the tail is not, and a slower, more contended CI runner has a longer one. When the tail crosses `1.5s` the deadline fires before the child writes anything, BOTH streams come back empty, and the test fails as `assert '' == 'SUITE START\n'`. That happened once in a full local suite run while validating PR #725, and the same failure shape reproduces deterministically by shrinking the deadline:

```
timeout=0.01: launch_failure=TIMEOUT  stdout=''  stderr=''
timeout=1.5 : launch_failure=TIMEOUT  stdout='SUITE START\n'  stderr='boom\n'
```

**No product defect is involved.** The streaming teardown reaps the child, joins the reader threads, and only then closes the pipes, so output the child did write is not lost. Both streams empty means the child had written nothing yet.

Fix direction, not binding: give the assertion real margin instead of a probability — a fake engine whose startup costs milliseconds rather than tens-to-hundreds of them, or a wider budget for this one timing-sensitive test. One trap worth inheriting rather than rediscovering: a shell fake engine must `exec` the command it sleeps in. Without `exec`, SIGTERM ends only the shell, the orphaned `sleep` keeps the captured pipe open, and the reader join runs to its full timeout on every single launch — a 2-minute probe became a >10-minute one.

## Acceptance criteria

- [ ] The test no longer depends on the child's interpreter startup finishing inside the launch timeout: either the margin between typical time-to-first-output and the deadline is at least two orders of magnitude, or the test synchronizes on the output instead of racing it.
- [ ] The guarantee under test is unchanged — a streaming launch that times out still returns the child's already-written stdout and stderr verbatim, and carries no gda prose in either stream.
- [ ] The new margin is measured, not asserted: the test passes repeatedly under artificial CPU load, and the pre-fix failure mode is shown to be reachable (e.g. by shrinking the deadline) so the fix is known to address it.
- [ ] No fake engine leaves an orphaned process holding a captured pipe open, so the launch teardown still reaches EOF promptly and the suite's wall clock does not regress.
- [ ] The suite keeps one shared fake-engine helper, or any divergence is stated in the helper's docstring.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Summary

`tests/test_launch.py::test_streaming_timeout_preserves_the_output_the_child_already_wrote` proves the #655/GDA-DF-012 guarantee: a streaming `Headless launch` that hits its timeout still returns the child's already-written stdout/stderr verbatim. Earlier rounds on this branch reduced — but did not remove — a real race between the launch's timeout deadline and a real child process's startup, using a cheaper fake engine and a widened timeout (1.5s -> 3.0s). A reviewer then demonstrated a working probe that removes the race outright by controlling only the clock the runner's poll loop reads, while still exercising a real child, real pipes, and the real timeout/teardown path. **This PR adopts that probe and reverses this branch's own earlier decline of the same recommendation** — see "Reversed decision" below.

**No production code changed.** `src/gda/runner.py` is untouched.

## Reversed decision: clock-controlled determinism

A prior review round on this branch proposed monkeypatching `gda.runner.time.monotonic` instead of racing a real deadline. The author declined it (commit `fb2d027f`'s reply) on three grounds: (1) `runner.py` has no injectable clock seam and adding one would be a forbidden production-code change; (2) faking the module's clock would "prove the capture/teardown logic given a clock the test controls, not that the real race resolves correctly" — treated as weakening the test; (3) the issue's own acceptance criteria already passed via the typical-margin branch, so nothing required the change.

All three turned out to be wrong, or at least not decisive:

1. No seam needed to be added. `runner.py` already does a plain `import time` and calls `time.monotonic()`; that module-level attribute lookup **is** an injectable seam, unused rather than absent.
2. The reviewer's probe (and this PR's adaptation of it) keeps the real child process, real pipes, real reader threads, and the real timeout/SIGTERM/reap/join teardown — only the poll loop's read of "how much time has passed" is faked, and only until the watch has actually observed the child's real output. The final implementation replaces `gda.runner`'s module binding with a small proxy instead of mutating the process-global stdlib `time` module; the proxy delegates `sleep` to the real function and overrides only `monotonic`. Nothing about what the test proves is weakened, and unrelated code in the pytest process continues to see the real stdlib clock.
3. The acceptance criteria's typical-margin branch passing was never the concern; the residual cold-start race (worst observed outlier 1110ms, ~2.7x headroom at 3.0s) was. Removing it outright is strictly stronger than measuring it more precisely.

## The fix

- The test now monkeypatches the `gda.runner.time` binding with a runner-local `SimpleNamespace` proxy. Its fake `monotonic` reports elapsed `0.0` until the watch's accumulated capture contains both expected lines (`"SUITE START"` in stdout, `"boom"` in stderr), then jumps past any timeout in one step. Its `sleep` is the real stdlib function, so the poll loop keeps making real, `0.05s`-spaced polls against the real child until its real output arrives, and only then does the deadline "elapse". The test pins that `time.monotonic` itself remains the original function.
- A real-clock safety ceiling (15s), independent of the fake, fails the test loudly — with whatever was captured so far — if the child's output never arrives at all. A hang becomes an assertion, not a stuck suite.
- The `timeout` argument passed to `launch()` is now inert: the loop only ever observes elapsed `0.0` or the fake's post-observation jump, so any positive value under that jump behaves identically. It is `1.0` now, a plain placeholder — the 1.5s/3.0s cold-start margin math it used to carry no longer applies to anything and is deleted rather than kept as unused history.
- `_fast_fake_engine` (the `/bin/sh` stand-in) is kept, but its docstring is rewritten: it no longer narrows a race (there is none), it just keeps the still-real startup wait — and the suite — fast rather than paying a Python interpreter's startup for no reason.

## On the previous margin/probability language — superseded, not corrected in place

The prior PR body claimed the fix "reduces flake probability by two-plus orders of magnitude" and discussed a "cold start" tail (312ms/1110ms outliers) sized against a 3.0s deadline. Both statements described the now-removed probabilistic-margin approach:

- The "two-plus orders of magnitude" phrasing was itself inaccurate for that approach — the evidence only supported a **margin** change (~54x -> ~750x typical time-to-first-output vs. deadline, i.e. ~13.9x), never an estimate of `P(first_output > deadline)`.
- "Cold start" conflated two different things — a freshly spawned interpreter process, and the first use of a newly generated script path — without saying which was measured.

This PR does not restate either number correctly, because there is no more margin or probability left to state: the guarantee is now unconditional, up to the 15s real-clock safety ceiling — which exists to fail loudly rather than hang, not to bound a race. The old measurement tables and margin math are removed from this body rather than kept as historical color; that narrative stays on the PR's review thread for anyone who wants it.

## Evidence

Target test alone, 25 consecutive runs (`uv run pytest -q tests/test_launch.py::test_streaming_timeout_preserves_the_output_the_child_already_wrote`), each a fresh process: **25/25 passed**.

| | pytest-reported test duration | full process wall clock |
| --- | --- | --- |
| min | 0.53s | 0.91s |
| max | 0.97s | 1.52s |
| avg | 0.59s | 1.03s |

For comparison, every run of the previous (race-narrowed) version paid the full `3.0s` timeout, because the fake engine never exits on its own — the deterministic version instead ends as soon as the real child's real output has actually arrived, typically well under 1s.

## Validation commands

Run BEFORE any number above was written, against exact head `b907bce7`:

- `uv run pytest -q tests/test_launch.py` -> `24 passed in 7.39s`
- `uv run pytest -q -m "not e2e"` -> `1901 passed, 539 deselected in 179.01s (0:02:59)`
- `uv run ruff check .` -> `All checks passed!`
- `uv run ruff format --check .` -> `472 files already formatted`
- `uv run pyright` -> `0 errors, 0 warnings, 0 informations`
- CI at exact head `b907bce7` ([run 33162103207](https://github.com/aigengame/godot-agent/actions/runs/33162103207)): `Python unit tests and package build` job -> `1899 passed, 2 skipped, 539 deselected in 71.15s (0:01:11)`; `tests/test_launch.py` shows 0 skips in that run (the 2-skip delta is the same pre-existing macOS-local-vs-Linux-CI environment difference noted in earlier rounds, unrelated to this change). Lint, format, and pyright CI jobs also pass at this head.
- Godot e2e: not run — not needed for this slice (test-suite-only change, no engine-facing behavior touched).

## Scope

Touches only `tests/test_launch.py`.

Closes #728

## Follow-up: isolate the fake clock from the process-global stdlib module

Re-review at head `b907bce7` found that `runner.time is time`: patching `runner.time.monotonic` therefore changed `time.monotonic` for the entire pytest process, contrary to the PR's "just that one lookup" claim. Fully adopted in commit `b8706b5b`.

The test now replaces only `gda.runner`'s module binding with a `SimpleNamespace(monotonic=fake_monotonic, sleep=time.sleep)` proxy. A regression assertion pins `time.monotonic is real_monotonic`. Red proof against the previous implementation failed exactly on that assertion; the isolated proxy then passed.

Latest local validation at head `b8706b5b`:

- target test → **1 passed**, then **10/10** fresh pytest runs passed
- `tests/test_launch.py` → **24 passed**
- Ruff check/format → **passed**
- Pyright → **0 errors, 0 warnings, 0 informations**
- `git diff --check` → **clean**

### DIFF
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -15,6 +15,7 @@ import subprocess
 import sys
 import time
 from pathlib import Path
+from types import SimpleNamespace
 
 import pytest
 
@@ -319,12 +320,11 @@ def test_streaming_timeout_preserves_the_output_the_child_already_wrote(
     # the real child's startup, narrowed only by a cheap shell engine (see
     # ``_fast_fake_engine``) — a residual race, later removed outright by
     # controlling the CLOCK the runner's poll loop reads instead of the runner's
-    # behaviour: ``gda.runner`` does a plain ``import time`` and calls
-    # ``time.monotonic()``, so patching ``runner.time.monotonic`` redirects just
-    # that one lookup. ``subprocess`` and ``threading`` instead bind their own
-    # ``_time = monotonic`` reference at import time, so the child's real
-    # ``proc.wait()`` grace period and the reader-thread ``join()`` are untouched —
-    # only the poll loop's idea of "how much time has passed" is fake.
+    # behaviour: ``gda.runner`` does a plain ``import time``, so replacing that
+    # module binding with a runner-local proxy redirects its ``monotonic`` lookup
+    # without mutating the process-global stdlib module. The proxy delegates
+    # ``sleep`` to the real function, so only the poll loop's idea of "how much
+    # time has passed" is fake.
     #
     # The fake reports elapsed 0.0 until the watch's accumulated capture actually
     # contains both lines the assertions below check, then jumps past any timeout
@@ -343,8 +343,9 @@ def test_streaming_timeout_preserves_the_output_the_child_already_wrote(
     engine = _fast_fake_engine(tmp_path, "SUITE START", "boom")
     watch = _RecordingWatch()
 
-    # Captured BEFORE patching — calling this (not ``time.monotonic``) inside the
-    # fake avoids the fake recursing into itself.
+    # Captured BEFORE replacing the runner's module binding. The stdlib module
+    # itself stays untouched, and the fake uses the saved function for its
+    # independent real-time safety ceiling.
     real_monotonic = time.monotonic
     real_deadline = real_monotonic() + 15.0  # generous real-time safety ceiling
 
@@ -358,7 +359,13 @@ def test_streaming_timeout_preserves_the_output_the_child_already_wrote(
         both_lines_seen = "SUITE START" in watch.stdout and "boom" in watch.stderr
         return 100.0 if both_lines_seen else 0.0
 
-    monkeypatch.setattr(runner.time, "monotonic", fake_monotonic)
+    monkeypatch.setattr(
+        runner,
+        "time",
+        SimpleNamespace(monotonic=fake_monotonic, sleep=time.sleep),
+    )
+    assert runner.time is not time
+    assert time.monotonic is real_monotonic
 
     result = launch(engine, [], cwd=None, timeout=1.0, watch=watch, timeout_label="X")
 
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -249,15 +249,21 @@ def _fast_fake_engine(tmp_path: Path, stdout_line: str, stderr_line: str) -> Pat
     """A ``/bin/sh`` stand-in, for the one test where startup latency IS the point (#728).
 
     Diverges from ``_fake_engine`` on purpose: that one pays a fresh Python
-    interpreter's startup (measured at #728: ~25-30ms typical, spiking past
-    900ms under load — plenty of margin for every other streaming test here,
-    which race nothing) before it writes a byte. The timeout-preserves-output
-    test needs the real child to have ALREADY written before ``launch``'s
-    timeout elapses, so shrinking that head start is the fix: a shell has no
-    interpreter to load, so it typically writes within single-digit
-    milliseconds (measured at #728: ~6-10ms, worst observed under 12-way CPU
-    load ~0.5s) — two-plus orders of magnitude below the caller's timeout
-    instead of the same order of magnitude it used to be.
+    interpreter's startup before it writes a byte — plenty of margin for
+    every other streaming test here, which race nothing. ``/bin/sh`` is an
+    interpreter too, but a far cheaper one to start than Python's, and that
+    gap is what the timeout-preserves-output test needs, since its child must
+    have ALREADY written before ``launch``'s timeout elapses. The measurements
+    and the margin they justify live in that ONE test's own comment, next to
+    the timeout value they argue for — not restated here, so there is a single
+    place to update rather than two copies that can silently drift apart (a
+    real defect an earlier #728 review round caught).
+
+    ``printf '%s\\n'``, not ``echo``: XSI ``echo`` interprets backslash
+    escapes in its operand on this platform's ``/bin/sh``, so a payload
+    containing ``\\n`` or similar would print something other than what was
+    passed in. ``printf`` with a literal ``'%s\\n'`` format leaves the
+    (``shlex.quote``-escaped, so shell-syntax-safe) argument untouched.
 
     It always ``sleep``s afterward, past any timeout this suite uses, via
     ``exec`` — not a trailing background job. Without ``exec`` SIGTERM only
@@ -269,7 +275,7 @@ def _fast_fake_engine(tmp_path: Path, stdout_line: str, stderr_line: str) -> Pat
     out = shlex.quote(stdout_line)
     err = shlex.quote(stderr_line)
     script.write_text(
-        f"#!/bin/sh\necho {out}\necho {err} 1>&2\nexec sleep 30\n",
+        f"#!/bin/sh\nprintf '%s\\n' {out}\nprintf '%s\\n' {err} 1>&2\nexec sleep 30\n",
         encoding="utf-8",
     )
     script.chmod(0o755)
@@ -307,19 +313,17 @@ def test_streaming_timeout_preserves_the_output_the_child_already_wrote(tmp_path
     #
     # THE #728 FIX: proving that means the child must ALREADY have written before
     # the timeout below elapses — a real race against a real process's startup.
-    # ``_fast_fake_engine`` (not the shared ``_fake_engine``) keeps that race from
-    # being a coin flip: no interpreter to load means the write typically lands in
-    # single-digit milliseconds, two-plus orders of magnitude inside the deadline
-    # instead of sharing its order of magnitude (see its docstring for the
-    # measurements this margin is based on, not asserted from). The timeout is
-    # also doubled from the pre-#728 1.5s to 3.0s: steady state is tight (180
-    # measured runs, ~4ms typical, max 5.1ms), but isolated FIRST-invocation
-    # outliers of 312ms and 1110ms were observed across independent
-    # measurement runs — and this test spawns the engine exactly once per
-    # suite, so that cold start IS its normal path, not a rare one the steady
-    # state can stand in for. Against the worst observed outlier, 1.5s leaves
-    # only ~1.35x headroom; 3.0s leaves ~2.7x, cheaply (the fake engine never
-    # exits on its own, so this is the test's own wall-clock cost either way).
+    # ``_fast_fake_engine`` (not the shared ``_fake_engine``) narrows that race:
+    # starting ``/bin/sh`` is far cheaper than starting a Python interpreter (see
+    # its docstring), so the write lands sooner. The timeout is ALSO doubled from
+    # the pre-#728 1.5s to 3.0s, because "sooner" still has a tail: steady state is
+    # tight (180 measured runs, ~4ms typical, max 5.1ms), but isolated
+    # FIRST-invocation outliers of 312ms and 1110ms were observed across
+    # independent measurement runs — and this test spawns the engine exactly once
+    # per suite, so that cold start IS its normal path, not a rare one the steady
+    # state can stand in for. Against the worst observed outlier, 1.5s leaves only
+    # ~1.35x headroom; 3.0s leaves ~2.7x. This is the ONE place these numbers are
+    # recorded; nothing else in this file restates them (#728 review).
     engine = _fast_fake_engine(tmp_path, "SUITE START", "boom")
 
     result = launch(
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -312,14 +312,14 @@ def test_streaming_timeout_preserves_the_output_the_child_already_wrote(tmp_path
     # single-digit milliseconds, two-plus orders of magnitude inside the deadline
     # instead of sharing its order of magnitude (see its docstring for the
     # measurements this margin is based on, not asserted from). The timeout is
-    # also doubled from the pre-#728 1.5s to 3.0s: across 40 measured runs the
-    # observed worst-case time-to-first-output was ~312ms, and that max came
-    # from an UNLOADED run (12-way CPU load's own max was only ~20ms) — the
-    # tail is not purely load-driven, so reasoning from the typical case alone
-    # would repeat the error that produced the original flake. 1.5s leaves
-    # only ~4.8x headroom over that observed tail; 3.0s leaves ~9.6x, cheaply
-    # (the fake engine never exits on its own, so this is the test's own
-    # wall-clock cost either way).
+    # also doubled from the pre-#728 1.5s to 3.0s: steady state is tight (180
+    # measured runs, ~4ms typical, max 5.1ms), but isolated FIRST-invocation
+    # outliers of 312ms and 1110ms were observed across independent
+    # measurement runs — and this test spawns the engine exactly once per
+    # suite, so that cold start IS its normal path, not a rare one the steady
+    # state can stand in for. Against the worst observed outlier, 1.5s leaves
+    # only ~1.35x headroom; 3.0s leaves ~2.7x, cheaply (the fake engine never
+    # exits on its own, so this is the test's own wall-clock cost either way).
     engine = _fast_fake_engine(tmp_path, "SUITE START", "boom")
 
     result = launch(
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -13,6 +13,7 @@ suite.
 import shlex
 import subprocess
 import sys
+import time
 from pathlib import Path
 
 import pytest
@@ -246,18 +247,18 @@ def _fake_engine(tmp_path: Path, body: str) -> Path:
 
 
 def _fast_fake_engine(tmp_path: Path, stdout_line: str, stderr_line: str) -> Path:
-    """A ``/bin/sh`` stand-in, for the one test where startup latency IS the point (#728).
+    """A ``/bin/sh`` stand-in, kept for WALL-CLOCK SPEED, not correctness (#728).
 
     Diverges from ``_fake_engine`` on purpose: that one pays a fresh Python
-    interpreter's startup before it writes a byte — plenty of margin for
-    every other streaming test here, which race nothing. ``/bin/sh`` is an
-    interpreter too, but a far cheaper one to start than Python's, and that
-    gap is what the timeout-preserves-output test needs, since its child must
-    have ALREADY written before ``launch``'s timeout elapses. The measurements
-    and the margin they justify live in that ONE test's own comment, next to
-    the timeout value they argue for — not restated here, so there is a single
-    place to update rather than two copies that can silently drift apart (a
-    real defect an earlier #728 review round caught).
+    interpreter's startup before it writes a byte. The one test that uses this
+    (``test_streaming_timeout_preserves_the_output_the_child_already_wrote``)
+    used to race that startup against a real deadline; that race is gone —
+    the test now controls the runner's clock directly, so a slower child could
+    no longer flip the result. ``/bin/sh`` stays anyway: the test still waits,
+    in real wall-clock time, for this REAL child to actually write its output
+    before the (now fake) deadline is allowed to cross, and a cheap shell keeps
+    that wait — and the suite — fast rather than paying a Python interpreter's
+    startup for no reason.
 
     ``printf '%s\\n'``, not ``echo``: XSI ``echo`` interprets backslash
     escapes in its operand on this platform's ``/bin/sh``, so a payload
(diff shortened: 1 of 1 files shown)


qinhaihong-red:
Addressed the independent adversarial review's three findings:

1. **Unescaped interpolation in `_fast_fake_engine` — fixed.** Both string arguments are now passed through `shlex.quote()` before being interpolated into the generated `/bin/sh` script, so a future caller passing a value containing a quote gets a correctly-escaped literal instead of a shell syntax error (which would have surfaced as both captured streams empty — indistinguishable from the failure mode this test exists to catch).
2. **Persist the measurement harness as a repo artifact — not adopted.** #728's acceptance criteria don't require it; the numbers already live in `_fast_fake_engine`'s docstring where a reader meets them. A committed micro-benchmark whose only job is proving a review happened would be an unowned artifact that rots. Residual risk, stated openly: the measurements can't be mechanically re-derived later without redoing the work.
3. **Timeout doubling (1.5s → 3.0s) justification — comment rewritten, value kept.** Re-measured time-to-first-output for the shell fake engine (40 runs): observed worst case ~312ms, and that max came from an *unloaded* run, not the 12-way-load run (whose own max was ~20ms) — the tail is not purely load-driven. The test's comment now cites this directly: 1.5s would leave ~4.8x headroom over the observed tail, 3.0s leaves ~9.6x. 3.0s stays.

Fixed at 3a33d224. Full gate re-run: `pytest -q -m "not e2e"` → `1901 passed, 539 deselected`; `pytest -q tests/test_launch.py` → `24 passed`; `ruff check .` → all checks passed; `ruff format --check .` → 472 files already formatted; `pyright` → 0 errors/warnings. No stray `sleep 30` process survives (`pgrep -f 'sleep 30'` → no match).

qinhaihong-red:
Correction on finding 3: the previous comment cited the friendlier of two disagreeing measurements (312ms) while my own re-measurement had actually found 1110ms — against 1.5s that's ~1.35x headroom, not ~4.8x. Fixed at 548caf0a: the comment now cites the worst observed outlier (312ms and 1110ms, both first-invocation cold starts against a tight ~4ms/max-5.1ms steady state across 180 samples), notes this test spawns the engine once per suite so the cold path is its normal path, and states the resulting headroom correctly — ~1.35x at 1.5s, ~2.7x at 3.0s, which is why 3.0s stays.

Re-ran the two requested gates:
- `uv run pytest -q tests/test_launch.py` → `24 passed in 10.60s`
- `uv run ruff format --check .` → `472 files already formatted`

Nothing else in the file changed.

qinhaihong-red:
Adversarial review of exact head `548caf0aea4b3bda25fe80b5e972b26b756f25e7` against merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

- **[P2] The completion claim does not match the implemented guarantee.** Issue #728 says this test should “stop racing” and “no longer depend” on child startup finishing before the deadline, but its alternative AC accepts a ≥100× *typical* margin. Exact head still calls this “a real race” at `tests/test_launch.py:308-322`; the corrected evidence records a 1110ms first-invocation cold-start outlier on the path this test exercises, leaving only ~2.7× headroom at 3.0s. Either synchronize the verdict on observed output, or change the issue/PR wording to accurately state that this reduces flake probability and explicitly accepts the residual cold-start risk. This does not change the separate literal Spec result below.

- **[P2] Reconcile the PR body with the final evidence.** The body still says the shell has “no interpreter to load”, attributes its worst result to 12-way load at ~0.5s, and reports `1901 passed`. `/bin/sh` is itself the script interpreter; the final author correction instead identifies 312ms and 1110ms first-invocation cold outliers and says the tail is not load-driven. Exact-head CI selected 1901 tests but reported `1899 passed, 2 skipped, 539 deselected`. Please update the body and distinguish local from CI results. This does not require committing a benchmark harness.

Terminology/prose assessment: I inspected the full issue, PR body, and replies against `CONTEXT.md`, ADR-0031, `RULES.md`, and ASD-STE100. Established terms are otherwise consistent; the substantive prose defects are the causal and completion inaccuracies above.

## Domain-modular architecture review — Changes required

- **[P2] The output-before-deadline invariant still has no deterministic owner.** `tests/test_launch.py:302-327` starts the real deadline before the child writes and asserts the streams after three seconds without a readiness barrier. Loaded repetitions passed, but a tiny deadline still reproduces TIMEOUT/124 with both streams empty, and the final cold-start evidence leaves only 2.7× worst-observed headroom. The load-bearing claim therefore remains `open`: any startup-to-first-output delay ≥3s disconfirms it. A small deterministic alternative is to retain the real pipes and process teardown, but use the watch plus a controlled monotonic clock so the clock crosses the timeout only after both expected lines have been observed.

- **[P2] The latency rationale has conflicting editable owners.** The helper docstring at `tests/test_launch.py:248-260` still records “under 12-way load ~0.5s”, while the adjacent final rationale at `:308-322` records a 1110ms first-invocation cold outlier and says cold start governs this suite path. Keep one current statement beside the helper/timeout and have the other location refer to it.

Production `runner.py` remains the sole process-lifecycle/capture owner, and `exec sleep 30` correctly makes the sleeper the direct child for termination and pipe EOF. `/bin/sh` is POSIX-only, but the pre-existing streaming tests already rely on executable shebang scripts, so this is an existing Windows-test coverage gap rather than a new regression.

## Standards

- **[P3, judgement call — Speculative Generality] `tests/test_launch.py:248-273`:** `_fast_fake_engine` accepts arbitrary output lines, but `shlex.quote()` protects shell syntax only; `echo` can still interpret operands. An exact-head probe with `r"a\cb"` emitted only `a`, not the requested line. The sole current caller uses safe literals, so current behavior is unaffected. Use `printf '%s\n' ...` or remove the unused generality and inline the two fixed payloads.

No hard documented-standard violation found.

## Spec — Pass

Issue #728 explicitly permits either synchronization or a ≥100× margin against *typical* time-to-first-output. The measured ~4–10ms typical latency against 3.0s gives 300–750×, so exact head satisfies the written alternative even though a low-probability worst-case race remains.

The unchanged assertions at `tests/test_launch.py:329-333` preserve TIMEOUT/124, exact stdout/stderr, and no gda timeout prose. Helper divergence is documented, `exec sleep 30` addresses the orphan/EOF criterion, and the PR supplies loaded repetitions plus a reachable pre-fix empty-stream case. No missing requirement, scope creep, or incorrect implementation found.

## Validation

- Immutable exact-head snapshot: `tests/test_launch.py` — `24 passed`.
- Target test: repeated local and 12-worker loaded runs passed, each near the 3.0s deadline.
- Tiny timeout: TIMEOUT/124 with empty stdout/stderr, reproducing the race boundary.
- Exact-head CI: required checks green; unit job `1899 passed, 2 skipped, 539 deselected`; Godot E2E skipped.
- Sandbox restrictions prevented direct process-table inspection; prompt completion plus the `exec`/teardown path is indirect no-orphan evidence.

qinhaihong-red:
Reply to the external adversarial review at head `548caf0a`. Fixed at `fb2d027f`.

**1. [P2] Docstring vs. timeout comment contradiction — FULLY ADOPTED.**
Confirmed: this existed because a previous review round rewrote the timeout comment (`:308-322`) to the cold-start-based numbers but left the adjacent docstring (`:249-262`) with the earlier "no interpreter to load ... two-plus orders of magnitude" claim — two editable copies of one fact, and they'd drifted. Fixed by making the timeout comment the single place the measurements and margin live; the docstring now says only *why* `/bin/sh` is used (cheaper interpreter to start than Python's) and points at the comment instead of restating numbers. Also fixed the factual error in the same pass: `/bin/sh` **is** an interpreter — what the fake engine avoids is Python's startup cost, not "an interpreter" as such. Both now read that way.

**2. [P2] PR body contradicted the final evidence — FULLY ADOPTED.**
All three defects fixed:
- Removed "no interpreter to load"; the fix section now says `/bin/sh` is a cheaper interpreter to start than Python's.
- Removed the "worst observed under 12-way CPU load ~0.5s" framing. The body now has a dedicated "Refinement — cold start vs. steady state" paragraph: three 60-run batches found a tight ~4-5ms steady state with no CPU-load correlation, and the 312ms/1110ms outliers are attributed to first-invocation cold start, not load. The "one interim flake" caveat is rewritten to say the load-correlated explanation looked plausible at the time but didn't survive the closer measurement.
- Added CI's exact-head numbers alongside local: local (now `fb2d027f`) `1901 passed, 539 deselected in 130.43s`; CI (`548caf0a`) `1899 passed, 2 skipped, 539 deselected in 76.02s`, labeled, with the 2-skip delta noted as a pre-existing macOS-vs-Linux-runner environment difference unrelated to this change (`tests/test_launch.py` itself shows 0 skips on both).

**3. [P2] Completion claim overstated what shipped — ADOPTED (wording only).**
Agreed the AC is satisfied either way (the AC's own "or" branch — ≥100x typical margin — is met at ~750x), so nothing about the implementation changes. The PR body's "The fix" section now closes with an explicit, bolded statement: this reduces flake probability by two-plus orders of magnitude against the typical case and accepts a residual, much-smaller-but-nonzero cold-start risk, documented with its numbers — not a claim that the race is eliminated.

**4. [P3] `echo` reinterprets its operand — FULLY ADOPTED.**
Verified independently: `/bin/sh -c "echo 'a\cb'"` on this host emits just `a` (XSI `echo` treats `\c` as "stop output here"), while `/bin/sh -c "printf '%s\n' 'a\cb'"` emits `a\cb` intact. `shlex.quote` only protects shell *syntax*, not operand reinterpretation by the command it's passed to. Switched both `echo` calls in `_fast_fake_engine` to `printf '%s\n'` with a literal format string and the `shlex.quote`-escaped value as the argument — same cost, correct semantics. Confirmed `test_streaming_timeout_preserves_the_output_the_child_already_wrote` still passes with the new script.

**5. [P2 Architecture] Controlled monotonic clock instead of a real deadline — CONSIDERED, DECLINED.**
Agreed the invariant (proving output survives a timeout) has no deterministic owner today, and a cold start beyond ~3s would still disconfirm it. Declining anyway, on three grounds:
- `src/gda/runner.py`'s `_spawn_streamed` calls `time.monotonic()` directly at four sites (`:792, :822, :840, :848`) with no injectable clock seam. Adding one is a production-code change, which #728's brief explicitly forbade — it told the implementer to stop and report rather than add a seam that exists only for this test, and `runner.py` is reserved for a later wave (#714/#717). The alternative, monkeypatching `time.monotonic` on the `runner` module from the test, would work without touching production code, but:
- it changes what the test proves. Today it drives a real child, real pipes, and a real deadline race; with the module's clock faked, it would prove the capture/teardown logic *given* a clock the test controls, not that the real race resolves correctly — which is a step toward exactly the "weakened test" AC 2 forbids ("the guarantee under test is unchanged").
- The review's own Spec section already concludes the written AC passes as implemented, so this isn't fixing a failing acceptance criterion — it would be trading a small, now well-characterized and documented residual risk for a real, if modest, entropy increase (either a permanent test-only production seam, or a test that no longer exercises the real timing path).

**Trigger for revisiting:** if this test flakes again after this fix (i.e., a cold-start outlier beyond ~3s is observed in practice, not just measured standalone), the clock-controlled design is the next move to try — not a further widened timeout, which would just be re-running the same "give it more time" play against a tail that keeps getting measured with the friendlier sample.

---

Gates re-run at `fb2d027f` before the numbers above were written:
- `uv run pytest -q tests/test_launch.py` → `24 passed in 10.02s`
- `uv run pytest -q -m "not e2e"` → `1901 passed, 539 deselected in 130.43s (0:02:10)`
- `uv run ruff check .` → `All checks passed!`
- `uv run ruff format --check .` → `472 files already formatted`
- `uv run pyright` → `0 errors, 0 warnings, 0 informations`

qinhaihong-red:
Re-reviewed the fixes at exact head `fb2d027f21fed41e9046e552e7803f04af1513aa` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

The interpreter wording, conflicting in-code latency owners, local-versus-CI evidence, and `echo` operand-reinterpretation findings are closed. Standards passes, and the literal alternative Spec criterion of at least a 100x typical margin is satisfied. The following Artifact and architecture residues remain.

### [P2] Reconcile the closure claim and measurement language with the probabilistic guarantee

Issue #728 still says the test must stop racing and no longer depend on startup finishing inside the timeout, while the current PR explicitly accepts a non-zero race with only about 2.7x headroom over the worst observed 1110ms first-invocation sample and still closes #728. Either make the test deterministic, or formally amend #728 so that its goal and acceptance text explicitly accept the typical-margin alternative and its residual tail risk.

The PR body also says the flake probability is reduced by two-plus orders of magnitude. The evidence only establishes a typical margin change from about 54x to 750x, which is about a 13.9x increase in margin; it does not estimate the tail probability `P(first_output > deadline)` or prove a 100x probability reduction. State the measured margin and leave the residual probability unquantified.

Finally, `cold start` is used for both every newly spawned interpreter process and the first use of a newly generated script path, while repeated new processes against a reused path are called steady state. Define the measured dimension explicitly so the 312ms/1110ms causal explanation is reproducible. The validation section should now cite exact-head CI directly rather than calling `548caf0a` the same test content after the executable `echo` to `printf` change.

### [P2] Give the output-before-timeout precondition a deterministic test owner

The test still starts a real 3.0s deadline before the child writes and has no readiness barrier. Any first-output delay at or above 3.0s recreates the empty-stream false failure.

A production clock seam is not required. An independent exact-head probe monkeypatched only `gda.runner.time.monotonic`, retained the real child, real stdout/stderr pipes, reader threads, timeout branch, SIGTERM/reap/join, and made the next clock read cross the deadline only after the watch had observed both expected lines. It returned TIMEOUT / exit 124 with both streams verbatim in about 0.6s. This controls only the incidental startup timing; the separate wall-clock test continues to own real elapsed-time behavior.

The smallest fix is a test-local stateful watch and controlled clock: keep fake elapsed below the deadline until both lines are observed, then cross it, with a saved real monotonic safety ceiling to fail if output never arrives. No `runner.py` change is needed.

Exact-head `tests/test_launch.py`: 24 passed. Required CI checks are green with `1899 passed, 2 skipped, 539 deselected`; Godot E2E is skipped because this PR changes test code only.

qinhaihong-red:
Reply to the review at exact head `fb2d027f21fed41e9046e552e7803f04af1513aa`. Fixed at `b907bce7`.

**[P2] Give the output-before-timeout precondition a deterministic test owner — ADOPTED, reversing this branch's own earlier DECLINE.**

Decision: adopted, on your probe's own construction. The previous review round (reply on `fb2d027f`) declined this exact recommendation on three grounds — no injectable clock seam without a `runner.py` change, monkeypatching the clock "weakens what the test proves", and the acceptance criteria already passed via the typical-margin branch. Your probe demonstrated all three were wrong or non-decisive: `runner.py` already does a plain `import time` + `time.monotonic()`, so the seam existed unused; the probe keeps a real child, real pipes, real reader threads, and the real timeout/SIGTERM/reap/join teardown, faking only the poll loop's elapsed-time read, so nothing about the guarantee under test is weakened; and passing the AC's easier branch was never the actual concern — the residual 1110ms-outlier race was. That decline is reversed here, and I record why, not just what, in the PR body's new "Reversed decision" section.

Evidence: independently verified against this repo's own Python 3.13 interpreter that `subprocess.Popen`/`proc.wait()` and `threading.Thread.join()` each bind their own `_time = monotonic` reference at import time (`inspect.getsource` on both modules), so patching `runner.time.monotonic` cannot perturb the SIGTERM grace period or the reader-thread join — only the four `time.monotonic()` call sites inside `_spawn_streamed` are affected.

Change (`b907bce7`, `tests/test_launch.py` only): the test now monkeypatches `gda.runner.time.monotonic`. The fake reports elapsed `0.0` until the watch's accumulated capture contains both expected lines, then jumps past any timeout in one step, so the real poll loop keeps making real `0.05s`-spaced polls against the real child until its real output has actually arrived — only then does the deadline "elapse". A real-clock safety ceiling (15s), independent of the fake, fails the test loudly with whatever was captured so far if the output never arrives at all. `_fast_fake_engine` is kept for wall-clock speed (its docstring is rewritten to say so, since it no longer narrows a race).

Validation: `uv run pytest -q tests/test_launch.py` -> `24 passed in 7.39s`; the target test alone, 25 consecutive fresh-process runs, 25/25 passed, pytest-reported duration 0.53-0.97s (avg 0.59s) — versus the previous version's fixed 3.0s per run, since the fake engine never exits on its own. Full detail and the gate lines are in the PR body.

**[P2] Reconcile the closure claim and measurement language with the probabilistic guarantee — MOOT, superseded rather than fixed in place.**

Decision: the finding is correct about the PR body as it stood, but determinism above removes the thing it was critiquing. There is no more margin or probability to state correctly, so I did not restate the "two-plus orders of magnitude" claim with a corrected number, and did not go define "cold start" more precisely — both discussions described the now-deleted probabilistic-margin approach and would be misleading if kept, even corrected, next to a test that no longer has a margin.

Change: the PR body's old "The fix" (timeout-doubling, 54x/750x margin) and "Evidence" (cold-start outlier tables) sections are removed rather than corrected, replaced by the "Reversed decision" and new "Evidence" sections describing the deterministic mechanism and the 25-run timing above. The one part of this finding that survives the determinism change independently — Validation citing `548caf0a` as "same test content" when the `echo`->`printf` change had altered the executable — is fixed on its own merits: the Validation section now cites this PR's exact head (`b907bce7`) CI run directly (`1899 passed, 2 skipped, 539 deselected in 71.15s`, `tests/test_launch.py` 0 skips, same pre-existing macOS/Linux 2-skip delta as before), not a description of an earlier commit.

Validation: PR body's Validation section, and CI run https://github.com/aigengame/godot-agent/actions/runs/33162103207 (all required checks pass at `b907bce7`).

Gates re-run at `b907bce7`, before any number in this reply or the PR body was written:
- `uv run pytest -q tests/test_launch.py` -> `24 passed in 7.39s`
- `uv run pytest -q -m "not e2e"` -> `1901 passed, 539 deselected in 179.01s (0:02:59)`
- `uv run ruff check .` -> `All checks passed!`
- `uv run ruff format --check .` -> `472 files already formatted`
- `uv run pyright` -> `0 errors, 0 warnings, 0 informations`

qinhaihong-red:
Re-reviewed the deterministic-clock fix at exact head `b907bce723a6a644e7182cdde1a845cab1a6773f` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

## Conclusion: Changes required, one P3 residue

The prior Artifact/DDMA P2 findings are closed. The test now satisfies issue #728's synchronization alternative while retaining a real child, real pipes and reader threads, the real timeout branch, and real SIGTERM/reap/join teardown. Exact-head `tests/test_launch.py` passes.

### [P3] Isolate the fake clock to the runner module binding

`monkeypatch.setattr(runner.time, "monotonic", fake_monotonic)` does not redirect only the runner's lookup. `runner.time is time` is true, so the assignment temporarily replaces the process-global standard-library `time.monotonic`; any concurrent code or plugin that calls `time.monotonic()` directly observes the fake clock. The PR body and test comment therefore overstate the isolation.

`subprocess._time` and `threading._time` are independently bound and the current Spec guarantee passes, so this is a test-isolation residue rather than a failure of the timeout proof. Please replace the `runner.time` binding itself with a local proxy exposing the real `sleep` and fake `monotonic`, or use an equivalently local seam, then update the prose to match.

Required CI is green; Godot E2E is appropriately skipped for this test-only slice. The head was re-pinned before publishing and was unchanged.

qinhaihong-red:
Re-review finding fully adopted and fixed at exact head `b8706b5b9724ed088bcdc457704f4c73c2356d5e`.

The finding was correct: `runner.time is time`, so `monkeypatch.setattr(runner.time, "monotonic", ...)` mutated the process-global stdlib module. The previous PR prose claiming that it redirected only the runner lookup was therefore false.

Commit `b8706b5b` now replaces only `gda.runner`'s `time` binding with a runner-local `SimpleNamespace` proxy:
- `monotonic` is the controlled fake;
- `sleep` delegates to the real stdlib function;
- a regression assertion pins `time.monotonic is real_monotonic`.

RED proof: that assertion failed against the previous patch because `time.monotonic` had become the fake.

Local validation:
- target test → **1 passed**, followed by **10/10** fresh pytest runs
- `tests/test_launch.py` → **24 passed**
- Ruff check/format → **passed**
- Pyright → **0 errors, 0 warnings, 0 informations**
- `git diff --check` → **clean**

The PR body has been corrected to describe the runner-local proxy, not process-global monkeypatching. Current exact-head CI: Ruff, co-install, scope classification, and gda-balancing required are green; Pyright, Python unit/package, and the latest release-scope guard are still running/queued.

**Final CI update:** all required checks pass at exact head `b8706b5b9724ed088bcdc457704f4c73c2356d5e`; Godot E2E and out-of-scope gda-balancing jobs are skipped as expected. Merge state is `CLEAN`.

PR #758, opened 2026-08-27
fix(gda): one root-cache exclusion for every res:// walk (#712)

### LINKED ISSUE #712: project walks: a nested .godot directory is wrongly excluded (three walkers still compare the entry name)
## Parent

Same defect class as the script-side hole fixed in PR #710 (issue #663).

## What to build

`scene list`'s project walk skips every directory whose entry NAME is `.godot`, not just the root engine cache. A scene under a user-authored nested `.godot` directory is silently missing from the listing.

Reproduced against a real Godot 4.6.3 project (2026-08-18, re-confirmed 2026-08-27): `res://top.tscn` + `res://nested/.godot/hidden.tscn` → `gda scene list --json` returns only `["res://top.tscn"]`.

### Scope: THREE walkers still compare the entry name, not one

PR #710 fixed exactly one of four sibling walkers. Verified on `main`:

| walker | `operations.gd` | comparison | correct? |
|---|---|---|---|
| `_collect_script_paths` | `:4312` | `child != ENGINE_CACHE_DIR` | yes (#710) |
| `_collect_scene_paths` | `:4258` | `entry != ".godot"` | **no** |
| `_collect_resource_paths` | `:3729` | `entry != ".godot"` | **no** |
| `_collect_all_file_paths` | `:3750` | `entry != ".godot"` | **no** |

The two unnamed walkers are not consumers of the scene walk — they are siblings with the
same defect — so an implementer who reads AC 3 as "check the scene walk's callers" can
close this issue with three of five affected operations still wrong. The affected ops:
`project find-references` (`:3482`), `project dependencies` (`:3509`),
`project find-unused-resources` (`:3544`), `project statistics` (`:3584`), and the
`class_name` index `_project_class_name_index` (`:3684`) that node/resource creation
resolves through.

### The divergence is already user-visible as a cross-command contradiction

On one project holding `res://top.tscn`, `res://nested/.godot/hidden.tscn` and
`res://nested/.godot/hidden.gd` (reproduced 2026-08-27):

```
gda script list         -> {"scripts":[{"path":"res://nested/.godot/hidden.gd", ...}]}
gda project statistics  -> {"total_files":2, "script_count":0, "scene_count":1, ...}
gda scene list          -> {"scenes":[{"path":"res://top.tscn", ...}]}
```

`script list` finds one script; `project statistics` reports zero scripts on the same
project. Whichever way the fix goes, these must agree afterwards — that agreement, not
the single `scene list` row, is what closing this issue means.

Fix the scene walk the way PR #710 fixed the script walk: compare the full child path against the engine-owned root cache (`ENGINE_CACHE_DIR := "res://.godot"`, introduced by #710), so only `res://.godot` stays invisible. Check existing tests/docs for pins on the old behavior and update them deliberately.

## Acceptance criteria

- [ ] A `.tscn` under a nested `.godot` directory appears in `scene list`.
- [ ] Content under the real root `res://.godot` stays invisible **for the lexical path
      comparison** (regression-pinned both ways, mirroring #710's arm) — see the amendment below
      for what this deliberately does NOT cover.
- [ ] All four walkers use the same root-cache-path comparison; `project statistics`'s
      `script_count` and `script list` agree on the project above (a cross-command
      consistency regression, not four independent per-walker tests).

## Blocked by

- PR #710 landing on the dev branch (reuses its `ENGINE_CACHE_DIR` constant and mirrors its regression pattern).

---

> ## Amendment (2026-08-28) — the guarantee is LEXICAL-PATH-ONLY
>
> Raised by an external review of PR #758 and verified on real Godot 4.6.3. The acceptance
> criterion above, as originally written, is not satisfied by a path comparison, and no reasonable
> reading of this issue's scope would have satisfied it: with `res://nested/.godot` symlinked to
> `res://.godot`, the root cache is reached through the alias and enumerated.
>
> **Measured, base versus PR #758's head:**
>
> | | base `04d3e089` | PR #758 head |
> |---|---|---|
> | `scene list` | `[]` — no leak | leaks `res://nested/.godot/root_cache.tscn` |
> | `script list` | leaks the aliased cache script | leaks (unchanged) |
>
> So the exposure is **only partly pre-existing**. PR #710 shipped it for the script walk under
> #663; #758 propagates it to the other three walkers and their consumers, because making the four
> walkers agree — which is what this issue asked for — necessarily adopted the path comparison
> everywhere. An earlier note from me on that PR called the whole thing pre-existing; that was
> right about the script walk and wrong about the rest, and this amendment is the correction.
>
> **Decision.** This issue's guarantee is the LEXICAL one: exactly the path `res://.godot` is
> excluded, in all four collectors, and that is what its regressions pin. Filesystem identity
> (aliases) and traversal termination (cycles, which today emit fabricated duplicate paths up to
> the OS path limit) are **not** in scope here and are owned by **#760**, where they can be decided
> together as one symlink policy rather than half-decided inside a comparison change.
>
> The alternative — implementing the full policy before #758 merges — was considered and declined:
> the two shapes need one decision, `#760` owns it, and bolting alias resolution onto a single
> predicate would leave the cycle behaviour untouched while looking finished.

qinhaihong-red:
## What changed

`operations.gd` holds four sibling walkers over the project's `res://` tree. PR #710 fixed
one of them — `_collect_script_paths` — to compare the full child path against
`ENGINE_CACHE_DIR` (`res://.godot`). The other three still compared the directory NAME:

| walker | before | after |
|---|---|---|
| `_collect_script_paths` | `child != ENGINE_CACHE_DIR` (#710) | `_should_descend(child)` |
| `_collect_scene_paths` | `entry != ".godot"` | `_should_descend(child)` |
| `_collect_resource_paths` | `entry != ".godot"` | `_should_descend(child)` |
| `_collect_all_file_paths` | `entry != ".godot"` | `_should_descend(child)` |

So a user-authored nested `.godot` (an addon vendoring a sample project, a fixture tree)
was invisible to `scene list`, `project find-references`, `project dependencies`,
`project find-unused-resources`, `project statistics`, and the `class_name` index that
`node add --type` / `resource create --type` resolve through — while `script list` saw it.

**One decision, one site.** Unifying only the constant would have left four independent
branches — the exact shape that let #710 fix one walker while three drifted. `_should_descend`
is the single owner of the rule, and it carries the rationale; the const keeps only the value.
A generic walker taking a `Callable` accept-predicate was considered and rejected: the four
accept-tests genuinely differ (extension match, graph-eligibility, unconditional), so it would
buy nothing beyond this helper and cost indirection.

Known consequence, already adjudicated for the script walk by #663/#710 and adopted here
for consistency: if a project vendors a sub-project that WAS opened in an editor, that
sub-project's own `.godot/` artefacts now enter the analysis universe — `project statistics`
counts and `find-unused-resources` candidates. Not `dependencies`: cache artefacts do reach
the graph walk, but `_has_outgoing_references` admits only `.tscn`/`.tres`/`.gd` as source
rows, so they cannot become dependency rows. gda cannot tell that directory from an authored
one, and the alternative — a name test — is what produced the contradiction this PR closes.
The trade lives **in `docs/command-catalog.md` beside the rule**, not only in this PR body,
so a future reader meets it instead of refiling it as a bug.

## The defect, measured (Godot 4.6.3, macOS 15.6 / darwin 25.6.0)

On a project holding `res://top.tscn`, `res://nested/.godot/hidden.tscn`,
`res://nested/.godot/hidden.gd` (`class_name NestedThing`), plus content under the real
`res://.godot`:

**Before** — one project, two answers:

```
scene list        -> {"scenes":[{"path":"res://top.tscn",...}]}
script list       -> {"scripts":[{"path":"res://nested/.godot/hidden.gd","class_name":"NestedThing",...}]}
project statistics-> {"total_files":2,"scene_count":1,"script_count":0,...}
node add res://main.tscn --type NestedThing
                  -> error invalid_node_type: "no .gd script declares class_name NestedThing"
```

**After** — `scene list` returns both scenes, `project statistics` reports
`scene_count: 2` / `script_count: 1` (agreeing with the two listings), and
`node add --type NestedThing` resolves (`script_class: "NestedThing"`). Content under the
real `res://.godot` stays out of all of them **by path**, and `--type CachedThing` (declared
inside that cache) still refuses with `invalid_node_type`.

### What the exclusion does and does not guarantee (#712 AC, qualified)

The rule is **lexical** — the child's `res://` path is compared against `res://.godot`. So:

- **The AC holds for the rule #712 asked to unify.** Root-cache content authored at a
  `res://.godot/...` path is excluded from all four collectors, pinned in both directions by
  the regression.
- **It does not hold for an ALIAS.** With `res://nested/.godot` symlinked to `res://.godot`,
  the cache is enumerated through the alias path — `scene list` and `script list` return
  `res://nested/.godot/root_cache.*` and `node add --type RootCacheThing` resolves.
  Reproduced on this head by the implementing agent, and on the base `04d3e089` by the
  orchestrating agent: `script list` returns the aliased cache script there too, because
  path comparison for that walker shipped under #663/#710. **The gap predates this PR**;
  this PR propagates an already-accepted decision to the three sibling walkers, which is
  what #712 asked for.
- **A symlink CYCLE is worse and equally pre-existing.** `sub/loop -> sub` makes the walk
  descend to the OS path limit and report **33 entries for one real file**, at fabricated
  `res://sub/loop/loop/…/a.gd` paths. Identical on head and base (verified on head here,
  on base by the orchestrator). No hang; garbage output.

Resolving filesystem targets inside `_should_descend` would settle the alias and leave the
cycle standing — half of an undecided policy, decided in a slice about entry-name-versus-path.
It is **deliberately not done here** and is tracked as its own follow-up, with both probes as
evidence. The limitation is stated in `docs/command-catalog.md` beside the rule, on
`_should_descend` itself, and in ADR-0032's amendment.

## Tests / docs that pinned the old behavior

- `tests/test_e2e_scene.py::test_scene_list_enumerates_dot_prefixed_scenes_but_skips_godot_cache`
  — pins `.hidden.tscn`, `.config/deep.tscn` and root-cache invisibility only. It never
  pinned the nested-`.godot` exclusion, so it is unchanged and still passes.
- No other test or doc asserted the old (name-based) rule. `src/gda/commands/resource.py`
  has a Python-side skip (`rel.startswith(".godot/")`) that is already root-anchored and
  correct; untouched.
- **Docs updated deliberately**: `docs/command-catalog.md` stated the exclusion rule only
  inside the `script list` passage (#663). That statement is now generalized to "the rule
  for every `res://` walk" and names the #712 divergence; it also carries the **cost** of
  the rule (a vendored sub-project's own engine cache becoming visible to the analysis
  walks) next to its empirical justification, so the trade is durable rather than
  PR-body-only. The `scene` group and the static-analysis section each get a pointer that
  **refers** to that statement without restating its substance — matching the shape of the
  pre-existing `script validate --all` pointer, so the rule stays in one place.
- **ADR-0032 synced** (`docs/adr/0032-…-static-fallback.md`): its Decision still said the
  class_name static scan walks `res://` "skipping `.godot/`" — the directory-NAME test this
  slice replaced. A dated amendment records the real boundary, what it means for the index
  (a `.gd` under a nested `.godot` is indexed, so its `class_name` resolves at all three call
  sites and can make a name `ambiguous_class_name` that was not), the accepted trade, and the
  open symlink question; a Consequences bullet points at it. The original sentence stays as
  the point-in-time record.
- Two catalog claims corrected while in there: the cost sentence no longer names
  `dependencies` (see above), and "one static scan backs all four" became **two** walks over
  different universes — `find-references`, `dependencies`, `find-unused-resources` and the
  `class_name` index share the extension-filtered walk, `project statistics` counts with the
  unfiltered one (it also sees `.import` sidecars and `project.godot`), so their totals
  cannot reconcile. What is one is the exclusion rule. The same "single scan" claim stood in
  a third place — this test module's own docstring — and was corrected there too.
- After review: the catalog now says the test is **lexical**, names the alias and cycle
  limitations with a pointer to the tracked follow-up, and scopes the rule to **the four
  `res://` collectors in `operations.gd`** rather than "every `res://` walk".
- `operations.gd` comments: the rationale moved onto `_should_descend`; the const keeps only
  the value, and no walker's comment restates the rule any more.

## Test evidence

One **cross-command consistency** regression, not four per-walker tests:
`tests/test_e2e_project_analysis.py::test_project_walks_agree_on_a_nested_dot_godot_directory`
drives all five affected operations plus the two listings on ONE fixture project —
`scene list`, `script list`, `project statistics` (counts and `by_extension`),
`node add --type` (the `class_name` index), then `dependencies`, `find-references` and
`find-unused-resources` — and asserts they agree, in both directions (nested content in,
root-cache content out). The three analysis calls run after the `node add`, so the nested
script is a real reference target: `top.tscn`'s edge to it is reported from both ends, the
nested scene is an orphan candidate, and the root cache — whose own `.gd` and `.tscn` WOULD
otherwise be source rows and candidates — is absent from all three.

**Red-proof** (pre-fix tree, this branch's test): the test failed at its first arm —

```
> assert scenes == {"res://top.tscn", "res://nested/.godot/hidden.tscn"}
E AssertionError: assert {'res://top.tscn'} == {'res://neste...s://top.tscn'}
E   Extra items in the right set: 'res://nested/.godot/hidden.tscn'
1 failed in 5.29s
```

The other arms were red-proofed by the direct CLI repro quoted above (`script_count: 0`
against `script list`'s one script; `invalid_node_type` for `NestedThing`).

## Gates (all run in this worktree; host: local macOS, darwin 25.6.0, Godot 4.6.3)

`10404d7a`, `d7ee3ce3` and `90a4fe11` are behaviourally identical — the commits after the
first two change only comments, a docstring, a doc and an ADR — so an e2e result on any of
them holds for the others.

| gate | command | result | head |
|---|---|---|---|
| unit | `uv run pytest -q -m "not e2e"` | `1901 passed, 540 deselected in 120.97s (0:02:00)` | `90a4fe11` |
| unit (fixed order) | `uv run pytest -q -m "not e2e" -p no:randomly` | `1901 passed, 540 deselected in 132.24s (0:02:12)` | `90a4fe11` |
| doc/skill sync | `uv run pytest -q tests/test_skill_surface_sync.py tests/test_skill_command.py tests/test_readme_i18n_sync.py tests/test_command_descriptor_registry.py tests/test_project.py` | `71 passed in 2.21s` | `1dbfb2cd` |
| e2e (touched file) | `uv run pytest -q -m e2e tests/test_e2e_project_analysis.py -p no:randomly` | `15 passed in 50.80s` | `90a4fe11` |
| e2e (walk consumers) | `uv run pytest -q -m e2e tests/test_e2e_project_analysis.py tests/test_e2e_scene.py tests/test_e2e_project.py tests/test_e2e_classname_resolution.py -p no:randomly` | `77 passed in 127.09s (0:02:07)` | `10404d7a` |
| e2e (script walk) | `uv run pytest -q -m e2e tests/test_e2e_script.py -p no:randomly` | `68 passed in 147.40s (0:02:27)` | `10404d7a` |
| e2e (class-index consumers) | `uv run pytest -q -m e2e tests/test_e2e_node.py tests/test_e2e_resource.py -p no:randomly` | `136 passed in 534.58s (0:08:54)` | `d7ee3ce3` |
| lint | `uv run ruff check .` | `All checks passed!` | `90a4fe11` |
| format | `uv run ruff format --check .` | `472 files already formatted` | `90a4fe11` |
| types | `uv run pyright` | `0 errors, 0 warnings, 0 informations` | `90a4fe11` |

The **class-index consumers** row was produced by the orchestrating agent, not by the
implementing agent, on head `d7ee3ce3` in the same worktree and on the same host — the
implementer had skipped it for its runtime and argued the extraction could not disturb
those paths. The argument was replaced with the run: `136 passed`, the same count as on
the first commit. Every other row was produced by the implementing agent.

One transient: an earlier random-order unit run reported `1900 passed, 540 deselected, 1 error`,
the error at `tests/test_input_commands.py::test_an_explicit_null_button_still_means_the_left_button`
(setup/teardown, not an assertion). That test passes in isolation and the full suite passed
twice afterwards in both orders. The traceback was not captured, so the cause is not proven;
the likely one is temp-directory contention between concurrently running pytest processes in
sibling worktrees, which share the `pytest-of-<user>` base. This change cannot reach that test —
it touches comments, a doc and an ADR only.

Closes #712

### DIFF
--- a/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
+++ b/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
@@ -35,6 +35,22 @@ path` index once per process run. Resolution stays a **static text scan**; the s
 instantiation (and the Node-vs-Resource base-class check) remains **split per site** — only the
 `class_name → path` resolution is unified.
 
+> **Amendment (2026-08-28, #712) — the scan's boundary is the ROOT cache PATH, not the directory
+> name.** "Skipping `.godot/`" above was implemented as a directory-NAME test, which also excluded
+> every `.godot` deeper in the tree. #712 replaced that, in all four `res://` collectors at once,
+> with a lexical comparison of the child's path against `res://.godot`, owned by a single predicate
+> (`_should_descend` in `operations.gd`) so the four cannot drift apart again — three of them
+> disagreed with the fourth, and one project answered two ways: `script list` named a `.gd` whose
+> `class_name` this index could not resolve. For this ADR the boundary is now: the index covers a
+> `.gd` under a **nested** `.godot`, so its `class_name` resolves at all three call sites, and a
+> declaration there can make a name `ambiguous_class_name` that was unambiguous before. The trade
+> is accepted deliberately — a nested `.godot` is usually authored content (an addon vendoring a
+> sample tree), and nothing in the path distinguishes it from a vendored sub-project's own engine
+> cache, whose scripts then enter the index too. The comparison being lexical, it does not resolve
+> filesystem targets: an alias that leads to `res://.godot` is walked and the cache's own scripts
+> are indexed through it. Symlink policy for the `res://` walk is undecided and tracked separately;
+> it is not decided here.
+
 **Explicit contract edges:**
 
 - **Duplicate `class_name`** (declared in more than one `.gd`) is **ambiguous and rejected** with a
@@ -87,6 +103,9 @@ gda's existing boundary); its resolution would need a different mechanism and is
 - **Consistency restored** across the three sites: they now agree on `class_name` resolution in a
   never-opened project, rather than `find-references` reporting "no such class" while a repaired
   `resource create` resolves it.
+- **Scan boundary (amended 2026-08-28, #712).** What the scan skips is the root cache **path**, not
+  every directory named `.godot` — see the Decision amendment for the boundary, the accepted
+  nested-cache trade, and the open symlink question.
 - **Cost.** A never-opened-project miss walks the whole `res://` tree and parses every `.gd`; this is
   acceptable for one-shot ops and is **not** backed by a persistent gda-owned cache — a gda-owned
   cache would re-introduce the very ".godot ownership" complexity the editor-scan option was rejected
--- a/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
+++ b/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
@@ -48,7 +48,7 @@ instantiation (and the Node-vs-Resource base-class check) remains **split per si
 > sample tree), and nothing in the path distinguishes it from a vendored sub-project's own engine
 > cache, whose scripts then enter the index too. The comparison being lexical, it does not resolve
 > filesystem targets: an alias that leads to `res://.godot` is walked and the cache's own scripts
-> are indexed through it. Symlink policy for the `res://` walk is undecided and tracked separately;
+> are indexed through it. Symlink policy for the `res://` walk is undecided and tracked in #760;
 > it is not decided here.
 
 **Explicit contract edges:**
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -70,9 +70,8 @@ they are provenance, not status markers.
 | `gda scene preflight` | Boot a scene headless and report its startup verdict |
 
 **Enumeration** (established by #54): `scene list` walks the project's `res://`
-tree under the one exclusion every `res://` walk applies — the engine's own
-`res://.godot` cache, and only that path, never a directory that merely bears
-that name (stated in full with `script list` below; #712).
+tree under the same exclusion rule `script list` states below, so the two see the
+same directories and differ only in the extension they collect (#712).
 
 **Static instance reporting** (established by #400): `scene get` reads the stored
 `SceneState` without instantiating the host scene, but an instanced node is still
@@ -458,14 +457,15 @@ agent authored. **Only that path**, not every directory named `.godot`: a nested
 authored content, and excluding it hid real scripts from the listing and let `script validate --all`
 report a valid aggregate for a project holding an invalid script (#663 review). Sometimes it is not
 authored content — a vendored sub-project checked out under `res://` and opened once in an editor
-keeps an engine cache of its own, whose import artefacts then count in `project statistics`,
-`find-unused-resources` and `dependencies`. That cost is accepted deliberately: gda cannot tell the
+keeps an engine cache of its own, whose import artefacts then count in `project statistics` and
+become `find-unused-resources` candidates. That cost is accepted deliberately: gda cannot tell the
 two apart from the directory alone, and a false-valid aggregate is the worse failure. Hidden entries
 are otherwise enumerated as promised (#54). **This is the rule for every `res://` walk** — `script
-list`, `scene list`, and the static-analysis scan that backs `project statistics`,
-`find-references`, `dependencies`, `find-unused-resources` and the `class_name` index — so one
-project cannot answer two ways. It once did: three of the four walks compared the directory NAME,
-so `script list` reported a script `project statistics` counted as zero (#712). `gda script delete`
+list`, `scene list`, and both static-analysis walks (the extension-filtered one behind
+`find-references`, `dependencies`, `find-unused-resources` and the `class_name` index, and the
+unfiltered one `project statistics` counts with) — so one project cannot answer two ways. It once
+did: three of the four walks compared the directory NAME, so `script list` reported a script
+`project statistics` counted as zero (#712). `gda script delete`
 removes a script file and reports the removed script's `class_name`/`extends` (parsed before
 deletion), so the result names the content, not just the path. Delete honors the same addressing
 boundary as the rest of the group — only a `.gd` path is removed (a non-`.gd` target is refused
@@ -800,10 +800,12 @@ These create or edit resource files (`.gdshader`, `.tres`) and so are headless.
 | `gda project dependencies` | Map scene → scene references |
 | `gda project statistics` | File/line counts, autoloads, plugins |
 
-One static scan backs all four (and the `class_name` index node/resource creation
-resolves through). It reads the project's `res://` tree under the one exclusion every
-`res://` walk applies — the engine's own `res://.godot` cache, and only that path
-(stated in full with `script list` above; #712).
+Two static walks back these, over different file universes: `find-references`,
+`dependencies`, `find-unused-resources` and the `class_name` index (which node/resource
+creation resolves through) share the extension-filtered one, while `project statistics`
+counts with an unfiltered one that also sees `.import` sidecars and `project.godot` — so
+its file total does not reconcile with the others' candidate set. What is shared is the
+directory exclusion, the rule `script list` states above (#712).
 
 ---
 
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -69,6 +69,11 @@ they are provenance, not status markers.
 | `gda scene validate` | Check statically that a scene's dependencies resolve and its scripts compile |
 | `gda scene preflight` | Boot a scene headless and report its startup verdict |
 
+**Enumeration** (established by #54): `scene list` walks the project's `res://`
+tree under the one exclusion every `res://` walk applies — the engine's own
+`res://.godot` cache, and only that path, never a directory that merely bears
+that name (stated in full with `script list` below; #712).
+
 **Static instance reporting** (established by #400): `scene get` reads the stored
 `SceneState` without instantiating the host scene, but an instanced node is still
 identifiable. Its `type` is resolved to the referenced scene's root node type
@@ -452,7 +457,11 @@ one directory — the engine's own cache at `res://.godot`, whose contents are i
 agent authored. **Only that path**, not every directory named `.godot`: a nested one is authored
 content, and excluding it hid real scripts from the listing and let `script validate --all` report
 a valid aggregate for a project holding an invalid script (#663 review). Hidden entries are
-otherwise enumerated as promised (#54). `gda script delete`
+otherwise enumerated as promised (#54). **This is the rule for every `res://` walk** — `script
+list`, `scene list`, and the static-analysis scan that backs `project statistics`,
+`find-references`, `dependencies`, `find-unused-resources` and the `class_name` index — so one
+project cannot answer two ways. It once did: three of the four walks compared the directory NAME,
+so `script list` reported a script `project statistics` counted as zero (#712). `gda script delete`
 removes a script file and reports the removed script's `class_name`/`extends` (parsed before
 deletion), so the result names the content, not just the path. Delete honors the same addressing
 boundary as the rest of the group — only a `.gd` path is removed (a non-`.gd` target is refused
@@ -786,6 +795,11 @@ These create or edit resource files (`.gdshader`, `.tres`) and so are headless.
 | `gda project find-references` | Find references to a script/class |
 | `gda project dependencies` | Map scene → scene references |
 | `gda project statistics` | File/line counts, autoloads, plugins |
+
+One static scan backs all four (and the `class_name` index node/resource creation
+resolves through). It reads the project's `res://` tree under the one exclusion every
+`res://` walk applies — the engine's own `res://.godot` cache, and only that path
+(stated in full with `script list` above; #712).
 
 ---
 
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -463,8 +463,7 @@ engine cache of its own, whose import artefacts then count in `project statistic
 apart from the directory alone, and a false-valid aggregate is the worse failure. And because the
 test is lexical it compares the path as written, so it does **not** resolve filesystem targets: a
 symlink or alias under another path that leads to `res://.godot` is walked, and the cache's contents
-are then enumerated through that path. Symlink policy for the `res://` walk is undecided — a
-tracked follow-up owns it, together with the symlink CYCLE the same walk descends until the OS path
+are then enumerated through that path. Symlink policy for the `res://` walk is undecided — #760 owns it, together with the symlink CYCLE the same walk descends until the OS path
 limit stops it. Hidden entries are otherwise enumerated as promised (#54). **This rule governs the
 four `res://` collectors in `operations.gd`** — the `script list` walk, the `scene list` walk, and
 both static-analysis walks (the extension-filtered one behind `find-references`, `dependencies`,
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -454,10 +454,14 @@ issue #30) — null when the source declares neither, so the listing names every
 Enumeration needs a project, so projectless it is refused with `project_not_found` (pass
 `--project`); an empty project is a valid, empty listing, not an error. The walk excludes exactly
 one directory — the engine's own cache at `res://.godot`, whose contents are import artefacts no
-agent authored. **Only that path**, not every directory named `.godot`: a nested one is authored
-content, and excluding it hid real scripts from the listing and let `script validate --all` report
-a valid aggregate for a project holding an invalid script (#663 review). Hidden entries are
-otherwise enumerated as promised (#54). **This is the rule for every `res://` walk** — `script
+agent authored. **Only that path**, not every directory named `.godot`: a nested one is usually
+authored content, and excluding it hid real scripts from the listing and let `script validate --all`
+report a valid aggregate for a project holding an invalid script (#663 review). Sometimes it is not
+authored content — a vendored sub-project checked out under `res://` and opened once in an editor
+keeps an engine cache of its own, whose import artefacts then count in `project statistics`,
+`find-unused-resources` and `dependencies`. That cost is accepted deliberately: gda cannot tell the
+two apart from the directory alone, and a false-valid aggregate is the worse failure. Hidden entries
+are otherwise enumerated as promised (#54). **This is the rule for every `res://` walk** — `script
 list`, `scene list`, and the static-analysis scan that backs `project statistics`,
 `find-references`, `dependencies`, `find-unused-resources` and the `class_name` index — so one
 project cannot answer two ways. It once did: three of the four walks compared the directory NAME,
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -453,19 +453,25 @@ issue #30) — null when the source declares neither, so the listing names every
 Enumeration needs a project, so projectless it is refused with `project_not_found` (pass
 `--project`); an empty project is a valid, empty listing, not an error. The walk excludes exactly
 one directory — the engine's own cache at `res://.godot`, whose contents are import artefacts no
-agent authored. **Only that path**, not every directory named `.godot`: a nested one is usually
-authored content, and excluding it hid real scripts from the listing and let `script validate --all`
-report a valid aggregate for a project holding an invalid script (#663 review). Sometimes it is not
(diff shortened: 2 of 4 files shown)


qinhaihong-red:
Reviewed PR #758 at exact head `d7ee3ce3e7fbe081c1fa5c356c31dace3f9dde42` against merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`. I re-read the PR head after review; it is unchanged. Changes are required.

## Artifact Review

**[P2] `docs/command-catalog.md:454-466` and the PR body overstate the exclusion guarantee.**

They say the walks exclude exactly the engine cache at `res://.godot` and that its contents remain invisible. On Godot 4.6.3, a project with `res://nested/.godot` symlinked to `res://.godot` returned the root-cache scene and script as:

- `res://nested/.godot/root_cache.tscn`
- `res://nested/.godot/root_cache.gd`

`project statistics` counted both, and `node add --type RootCacheThing` resolved the cache script. `realpath` confirms that the lexical root and alias are the same directory. The phrase “every `res://` walk” is also broader than the four enumeration/static-analysis collectors governed here. Please preserve the intended statement by making the implementation target-aware, then scope the text to those four collectors.

**Terminology and prose assessment:** I inspected the complete current command catalog, the complete PR body, and issue #712 including its state, labels, milestone, and comments, using `CONTEXT.md`, ADR-0006/0009/0032, and ASD-STE100. No other substantive terminology or prose issue was found.

## Architecture / DDMA

**[P2] `src/gda/ops/operations.gd:3725` centralizes only the lexical predicate `child != ENGINE_CACHE_DIR`; it does not own the actual invariant “this directory is not the project's engine cache.”**

All four collectors call the helper, which is good DRY and dependency direction, but a symlink alias admits the same cache into scene/script listings, statistics, the reference graph, unused-resource analysis, and the `class_name` index. Resolve and compare filesystem targets, or otherwise reject aliases to the root cache, while continuing to admit an ordinary authored nested `.godot`. Add one cross-consumer alias regression.

Completeness and the root-boundary invariant fail on this case. Orthogonality, lexical single ownership, dependency direction, and the existing extension path otherwise pass. Claim evidence state is `open` because this project defines no stronger evidence-state vocabulary: ordinary nested-directory inclusion has representative passing evidence; root-cache invisibility has the counterexample above.

## Standards

**[P2, documented-standard violation]** `RULES.md:121-122` requires architecture/design changes to synchronize affected ADRs, but `docs/adr/0032-project-local-class-name-resolution-static-fallback.md:30-33` still defines the static scan as walking `res://` while “skipping `.godot/`.” It does not record the new root-only boundary or accepted nested-cache trade. Amend ADR-0032's Decision/Consequences so the architecture authority agrees with the implementation and catalog.

No Fowler-baseline smell was found; `_should_descend` is a justified single rule owner rather than a Middle Man.

## Spec

**[P2] Issue #712 requires: “Content under the real root `res://.godot` stays invisible (regression-pinned both ways, mirroring #710's arm).”**

The symlink probe above violates that acceptance criterion. `tests/test_e2e_project_analysis.py:562` and `:619` only reject result paths beginning with `res://.godot/`, so an alias to the same cache passes the test. Make the shared rule compare resolved directory identity and add the alias arm.

No other missing requirement, scope creep, or incorrect implementation was found.

Standards summary: 1 finding, worst P2. Spec summary: 1 finding, worst P2.

## Validation and limits

- Exact-head new E2E plus two adjacent direct-root/hidden E2Es: `3 passed`
- Real Godot 4.6.3 symlink probe reproduced the root-cache leak across listings, statistics, analysis consumers, and class resolution
- `git diff --check`: passed
- Exact-head CI run `33082762417`: unit/build, Ruff, Pyright, and co-install passed; Godot E2E was skipped
- I did not run the long full E2E suite or expand the probe to symlink cycles/external targets because the root-cache alias already falsifies the acceptance criterion

qinhaihong-red:
## Review response — head `90a4fe11`

Thank you for the symlink probe; the mechanism is real and reproduces here. Below is each
finding with the decision, the evidence it rests on, and what changed. Summary: the alias
defect is confirmed and **pre-existing**, a second and worse symptom of the same undecided
policy was found, so the code fix is declined in this slice and tracked as its own issue —
while every prose claim that overstated the guarantee is corrected, and ADR-0032 (which
still described the rule this slice replaced) is synced.

### Evidence, on head AND base

**Alias — `res://nested/.godot` symlinked to `res://.godot`.**

On this head (`90a4fe11`, real Godot 4.6.3, local macOS), reproduced by the implementing
agent:

```
scene list  -> {"scenes":[{"path":"res://nested/.godot/root_cache.tscn",...}]}
script list -> {"scripts":[{"path":"res://nested/.godot/root_cache.gd","class_name":"RootCacheThing",...}]}
node add res://main.tscn --type RootCacheThing -> {"script_class":"RootCacheThing",...}
```

On the base `04d3e089`, reproduced by the orchestrating agent: `script list` returns the
aliased cache script **identically**. Path comparison for that walker shipped in #710 under
#663; this PR propagates that already-accepted decision to the three sibling walkers, which
is exactly what #712 asked for. `scene list` did not leak on the base only because the
directory-name test excluded it — the bug this PR was filed to fix.

**Cycle — `sub/loop -> sub`, one real file `sub/a.gd`.**

On this head (implementing agent) and on the base (orchestrating agent), identically:

```
script list -> 33 entries for 1 real file
deepest path: res://sub/loop/loop/loop/…/loop/a.gd   (174 chars)
```

No hang — the OS path limit stops it — but the output is fabricated paths. No test anywhere
covers it.

### 1. Resolve filesystem targets in `_should_descend` — **not adopted in this PR** (tracked)

Not because the problem is unreal, and not because it is infeasible: `DirAccess.is_link()` /
`read_link()` are bound (`dir_access.cpp:674-675`).

It is declined because symlink handling in the `res://` walk has never been decided, it
produces at least the two distinct defects above, and realpath comparison inside this
predicate would fix the alias while leaving the cycle exactly as it is — half a policy,
decided in a slice whose subject is entry-name-versus-path. Deciding one rule in two places
is how the four walkers drifted apart in the first place; that is the failure mode this PR
exists to remove, and it would be an odd way to close it. Both probes go to the follow-up
issue as its evidence.

### 2. ADR-0032 out of sync — **adopted**

Correct, and it is this slice's responsibility: ADR-0032's Decision described the scan as
walking `res://` "skipping `.godot/`", which is the directory-NAME test this PR replaced.

`docs/adr/0032-…-static-fallback.md` now carries a dated amendment after that paragraph
stating the real boundary (the root cache **path**, compared lexically, owned by one
predicate across four collectors), what it means for this ADR specifically — a `.gd` under a
nested `.godot` is indexed, so its `class_name` resolves at all three call sites and a
declaration there can make a name `ambiguous_class_name` that was unambiguous before — the
accepted trade, and the open symlink question. A Consequences bullet points at it. The
original sentence is left standing as the point-in-time record, per the corpus convention.

### 3. The catalog overstated the guarantee — **adopted (prose only)**

Both counts were right.

- The rule is now stated as **lexical**: the child's `res://` path is compared against that
  one path, and the text says outright that it does not resolve filesystem targets, so an
  alias leading to `res://.godot` is walked and the cache is enumerated through it. The cycle
  is named in the same breath. Both point at the tracked follow-up.
- "every `res://` walk" is now **"the four `res://` collectors in `operations.gd`"** — the
  scope the rule actually governs.

`_should_descend` carries the same caveat at the decision site, with an explicit "do not
decide half of it in this predicate", so the next reader does not assume target-resolution
semantics. The PR body carries the same correction.

### 4. The AC "content under the real root stays invisible" — **adopted as a qualified claim**

The PR body now states, rather than glossing:

- the AC holds for the **lexical** rule #712 asked to unify — root-cache content at a
  `res://.godot/...` path is excluded from all four collectors, pinned both ways by the
  regression;
- an **alias** falsifies the literal reading of the AC;
- that gap **predates this PR** (base evidence above) and is tracked, not fixed here.

A qualified true claim beats a green checkbox — particularly for an AC whose literal reading
no shipped version of this code has ever satisfied.

### Gates on `90a4fe11`

| gate | result |
|---|---|
| `uv run pytest -q -m "not e2e"` | `1901 passed, 540 deselected in 120.97s (0:02:00)` |
| `uv run pytest -q -m "not e2e" -p no:randomly` | `1901 passed, 540 deselected in 132.24s (0:02:12)` |
| `uv run pytest -q -m e2e tests/test_e2e_project_analysis.py -p no:randomly` | `15 passed in 50.80s` |
| `uv run ruff check .` | `All checks passed!` |
| `uv run ruff format --check .` | `472 files already formatted` |
| `uv run pyright` | `0 errors, 0 warnings, 0 informations` |

One transient is recorded in the PR body: an earlier random-order unit run reported
`1900 passed, 540 deselected, 1 error` at
`tests/test_input_commands.py::test_an_explicit_null_button_still_means_the_left_button`
(setup/teardown, not an assertion). It passes in isolation, the suite passed twice afterwards
in both orders, and this change reaches only comments, a doc and an ADR. The traceback was not
captured, so the cause is stated as unproven rather than attributed.

qinhaihong-red:
Re-reviewed the fixes at exact head `941f74878ce75b125de34ced68a3f4c3ad444d3c` against the previous findings and merge-base `04d3e089b32330aeffc7da98f4824bb21873c2b3`.

The documentation and ADR findings are closed: the lexical policy, accepted limitation, cycle risk, and #760 follow-up are now stated accurately. One Spec residue remains.

### [P2] `Closes #712` still conflicts with the literal root-cache acceptance criterion

A real Godot replay at the current head still exposes root-cache content through a nested symlink alias:

- `scene list` returns `res://nested/.godot/root_cache.tscn`;
- `script list` returns the aliased cache script;
- statistics counts 3 scenes and 1 script.

The same project at base `04d3e089` does not leak the aliased cache scene and reports 2 scenes and 0 scripts, although the older script walker already leaks the script. The defect is therefore only partly pre-existing: this PR propagates the alias exposure into walkers and consumers that previously excluded it.

Issue #712 still requires content under the real root `res://.godot` to remain invisible, while this PR still closes #712. The honest disclosure and new #760 follow-up do not amend that acceptance criterion.

The rejection of a narrow realpath-only fix is reasonable because alias identity and cycles need one complete symlink policy. To close this finding, either implement that complete policy before merge, or formally amend and approve #712 as a lexical-path-only guarantee and let #760 own filesystem identity and cycle behavior.

Required exact-head CI checks are green. Godot E2E is skipped, so the behavior above was independently replayed against the real engine. The head was re-pinned immediately before publishing and remained unchanged.

qinhaihong-red:
## `Closes #712` — resolved by amending the acceptance criterion (2026-08-28)

The Spec residue on this PR is now closed, by the second of the two routes the review offered.

**Your base-versus-head measurement was right, and it corrects an earlier note of mine on this PR.** I re-ran it:

| | base `04d3e089` | this PR's head |
|---|---|---|
| `scene list` | `[]` — no leak | leaks `res://nested/.godot/root_cache.tscn` |
| `script list` | leaks the aliased cache script | leaks (unchanged) |

So the alias exposure is only partly pre-existing: PR #710 shipped it for the script walk under #663, and this PR propagates it to the other three walkers because making the four agree — the whole ask of #712 — necessarily adopted the path comparison everywhere. I had earlier told the implementer the gap predated this PR; that was right about the script walk and wrong about the rest.

**Issue #712 has been formally amended.** Its guarantee is now stated as the LEXICAL one — exactly the path `res://.godot` is excluded, in all four collectors, which is what this PR's regressions pin. Filesystem identity (aliases) and traversal termination (cycles, which today emit fabricated duplicate paths to the OS path limit) are explicitly out of scope there and owned by **#760**, together, as one symlink policy. The amendment carries the measurement above and names the correction.

The alternative — implementing the full policy before this merges — stays declined for the reason the review itself accepted: the two shapes need one decision, and resolving aliases in a single predicate would leave the cycle behaviour untouched while looking finished.

`Closes #712` is therefore accurate against the amended criterion rather than against a criterion no path comparison could satisfy.

Issue #760, opened 2026-08-28
res:// walk has no symlink policy: an alias re-admits the engine cache, a cycle emits fabricated paths


qinhaihong-red:
## Parent

Surfaced by an external adversarial review of PR #758 (issue #712), and deliberately NOT fixed
there — see that PR's reply for why deciding half of this policy inside a lexical-comparison slice
was declined.

## What to build

The `res://` walk in `src/gda/ops/operations.gd` has **no symlink policy**. It compares paths
lexically and descends into whatever the directory listing calls a directory. Two distinct defects
follow, both reproduced on real Godot 4.6.3 and both present on the pre-#712 base as well:

**1. An alias re-admits the excluded engine cache.** With `res://nested/.godot` symlinked to
`res://.godot`:

```
scene list  -> res://nested/.godot/root_cache.tscn
script list -> res://nested/.godot/root_cache.gd   (class_name RootCacheThing)
node add --type RootCacheThing  -> resolves against the cache script
```

`realpath` confirms the alias and the root cache are the same directory. This falsifies the literal
reading of #712's acceptance criterion "content under the real root `res://.godot` stays invisible".

**2. A cycle produces fabricated duplicate paths.** With `sub/loop` symlinked to `sub`, holding one
real `a.gd`:

```
script list -> 33 entries, 1 distinct file
               deepest: res://sub/loop/loop/loop/…/a.gd  (174 chars)
```

The walk descends until the OS path limit stops it. No hang, but every entry past the first is a
path that does not name a distinct file. Nothing in the suite covers this.

Decide the policy once, for all four collectors, and implement it in the one place that now owns
the descent decision (`_should_descend`). `DirAccess.is_link()` and `read_link()` are bound and
available (`core/io/dir_access.cpp:667-668`), so target resolution is feasible; whether to resolve,
to refuse links outright, or to bound depth is the decision this issue exists to take.

## Acceptance criteria

- [ ] A recorded decision covering BOTH shapes — cache aliasing and cycles — with its rationale,
      including what happens to a legitimate symlink that points at ordinary authored content
      (a vendored checkout reached through a link is a real workflow; do not break it silently).
- [ ] Content under the real root `res://.godot` stays invisible when reached through an alias, in
      all four collectors and the `class_name` index.
- [ ] A symlink cycle terminates without emitting fabricated paths, with a structured outcome
      rather than a truncation artefact of the OS path limit.
[...]

Issue #759, opened 2026-08-27
perf --budget: an invalid_params failure leaks pydantic internals and caller content


qinhaihong-red:
## Parent

Found by two independent adversarial reviews (code and architecture lenses) of PR #754, which
delivered #713. Left deliberately out of that PR's scope; its body records the exclusion.

## What to build

`perf`'s `--budget` loader reports an invalid budget entry by interpolating a pydantic
`ValidationError` directly into the failure message. The result is an `invalid_params` envelope
whose `message` — the field an agent reads — carries pydantic's internal dump instead of the rule
that was broken: the model class name, a `[type=..., input_value=..., input_type=...]` block that
echoes the caller's own budget-file content back, embedded newlines inside the JSON string, and a
`https://errors.pydantic.dev/...` URL.

PR #754 fixed exactly this defect class for the two ADR-0015 params-model channels — argv and
`--params-json` — by adding a public extractor in the shared failure taxonomy that renders a
`ValidationError` as the sentence the validator itself wrote. That extractor is public and already
used by two layers. This is the third producer of the same public code from another
caller-supplied value, and it predates #713.

Route this producer through the same extractor so one `Gda error code` does not speak two
languages depending on which surface produced it.

## Acceptance criteria

- [ ] An invalid `--budget` entry produces an `invalid_params` envelope whose `message` states
      the validator's own sentence, with no pydantic class name, no `[type=`, no `input_value=`,
      and no `pydantic.dev` URL.
- [ ] The caller's own budget-file content does not appear in the failure message.
- [ ] The existing extractor is reused, not re-implemented or copied.
- [ ] A regression pins the clean shape and the absence of the leaked markers, red-proofed against
      the pre-fix code.
- [ ] A sweep records whether any further `ValidationError`-to-public-failure producer remains, and
      names each one it deliberately leaves alone with the reason. (Known deliberate exclusion:
      `classify_run`'s `contract_violation` branch validates the ENGINE's output against gda's
      result schema, not caller input — a different origin and a different failure consequence.)

## Blocked by

- PR #754 merging into the dev branch (it introduces the extractor this reuses).

> **Correction (2026-08-31, PR #776 review):** the public option is `--budget` (the JSON artifact it names is still a budget file); this body originally wrote `--budget-file`, a token the CLI has never defined. All occurrences above are corrected in place.

PR #761, opened 2026-08-28
fix(gda): dogfooding phase-2 P1 — parser, walk, and input-authority fixes (dev→main promotion)

qinhaihong-red:
## What

The **phase-2 P1** promotion from `feat/gda-dogfooding-dev` to `main`, per the milestone branch model: slices land on the dev branch one squashed commit each; this PR promotes the wave via **merge commit** (never squash — squashing would collapse the per-issue `fix:`/`test:`/`refactor:` commits release-please derives the changelog from).

P1 was the first phase-2 wave: five independent slices, dispatched in parallel to isolated worktrees, merged in the order below. Every slice went through independent adversarial review on up to three lenses (code / document / architecture), and each was re-verified in the main loop against its own gates before merge.

## Merged slices

| Slice | Issue | dev commit | Slice PR |
| --- | --- | --- | --- |
| pin the ADR-0002 error-description column and reconcile the drift it surfaces | #701 | `fa441c91` | #755 (3 review rounds) |
| script-error parser: dot-terminated paths, whitespace grammar, engine-exact trailing rule | #698 | `78d17f19` | #756 (4 review rounds) |
| one root-cache exclusion for every `res://` walk | #712 | `8d445e63` | #758 (5 review rounds) |
| `script set` / `shader set`: one mode-selection authority | #713 | `55b85291` | #754 (5 review rounds) |
| deflake the streaming-launch timeout test | #728 | `708578d0` | #757 (4 review rounds) |

### What #701 shipped

The registry↔ADR-0002 equality pin covered `(category, source, exit_code)` but silently ignored the `description` column, so the two artifacts had drifted for 91 codes with nobody noticing.

- Measured at the fork point: **38 of 91 rows differed** after backtick/whitespace/period normalization; **41 ADR descriptions carry a trailing citation**, three of which match their registry counterpart exactly and so never entered the 38. All three populations are now stated separately rather than conflated.
- The pin extends to the description column, with one **decided** normalization rule: a trailing parenthetical is excluded from comparison only when it contains at least one real `ADR-NNNN`/`#N` reference. A bare `(Phase N)` is domain content (`CONTEXT.md` defines Phase 1/2 as a delivery concept), so it **is** compared — an earlier form of the rule let `(Phase 2)` and `(Phase 1)` normalize alike, which is the exact drift class this issue exists to catch.
- **18 substantive divergences** reconciled with a per-code verdict: 7 fixed the registry (e.g. `export_path_unset` had lost the `--output` producing condition; `live_display_unavailable` had lost the word "windowed"), 9 fixed the ADR (e.g. `live_method_not_allowlisted` said "its class" where ADR-0041 resolves along the attached-script chain), 2 were phrasing with no correctness winner.
- Residual stated exactly rather than papered over: a `Phase N` riding **inside** a citation cluster stays uncompared; for 20 of 23 phase-labelled rows the compared `Category` column pins the distinction indirectly, and the other three (`live_unsupported_platform`, `live_windowed_unavailable`, `live_windowed_permission_denied`) are `environment`-category, so for those it is caught by neither mechanism.

### What #698 shipped

Three defects in the script-error parser, each closing a path to **phantom success** — a run reporting `exit_status: 0` for a script that never executed.

- **The dot-terminated address.** `_CANT_LOAD` stripped an optional trailing period, but `main/main.cpp:4271` emits `"Can't load script: " + script` and never appends one — so `res://..` parsed back as `res://.`, missed the canonical-identity match, and the never-ran verdict was lost. The strip is dropped. `_FAILED_LOADING_RESOURCE` mirrors `resource_loader.cpp:343`, which **always** appends exactly one period, so its strip became mandatory — fail-closed hardening, **not** a fix for observed corruption (the old lazy quantifier never mis-read a well-formed line; the issue's original example was mis-constructed and was corrected).
- **The whitespace grammar.** `\S+` could not match a path containing a space, so `gda script run "missing file."` returned a success shape while the engine printed `Can't load script:`. Both patterns now capture the full line remainder. The `.gd` spelling masks this — with an extension the engine additionally emits a quoted `Attempt to open script '…'` that survives whitespace — so testing only that shape shows no bug.
- **The trailing-normalization boundary.** A first fix used Python `str.rstrip()` as a proxy; Godot's `String::strip_edges()` trims code points `<= U+0020` (`ustring.cpp:4073`), while Python's Unicode set is wider. The guard now encodes the engine's actual rule — trailing NBSP and EM SPACE reach the engine as before, trailing space/tab are refused pre-launch, and a path containing a line break is refused because it cannot round-trip through the line-oriented diagnostic protocol. Refusal, never `rstrip()`: silently trimming would launch a *different* script than the caller named. ADR-0031 records the complete pre-launch refusal set.

### What #712 shipped

`gda` answered two ways about one project. PR #710 had fixed `_collect_script_paths` to compare the full path against the root cache; the other three walkers still compared the directory **name**, so a user-authored nested `.godot` was invisible to them and visible to it.

- Reproduced as a cross-command contradiction: on one project, `script list` returned a script that `project statistics` counted as `script_count: 0`, and `node add --type NestedThing` refused a `class_name` the listing had just reported.
- All four walkers now route through a single `_should_descend()` — the PR's own review noted that unifying the *constant* while leaving four independent branches is the shape that produced the bug, so the **decision** has one owner too.
- The rule is documented as **lexical** in the command catalog, with its accepted cost (a vendored sub-project's own cache becomes visible to the analysis walks) stated beside it rather than left in a PR body, and ADR-0032 synchronized.
- **Issue #712's acceptance criterion was formally amended**, not silently reinterpreted: a symlink alias re-admits the root cache, which no path comparison can prevent. Measured base-vs-head, that exposure is only partly pre-existing. The guarantee is now stated as lexical-path-only, with **#760** owning filesystem identity and cycle termination together as one symlink policy.

### What #713 shipped

`script set` and `shader set` evaluated their mode-selection rule twice — a hand-rolled argv pre-check plus the params model's own validator — contradicting ADR-0015, which makes the model the input-rule authority.

- Both commands now build their params model through `params_or_bad_parameter()`; `parse_set_mode_argv` is deleted. The rule runs exactly once per invocation, verified across an instrumented script/shader × argv/JSON matrix.
- That routing exposed a real regression the slice then repaired at its seam: `str()` on a pydantic `ValidationError` dumps the model class name, a `[type=…, input_value=…]` tag per error, a pydantic.dev URL — and **echoes the caller's own value back**, including a `--content` payload. A shared `validation_error_message()` in `gda.errors` now renders the validator's own sentence, used by **both** ADR-0015 channels, so argv and `--params-json` report the identical sentence for the identical refusal.
- Public surface: `--help` byte-identical; `gda schema` differs at exactly four description leaves under `script set`/`shader set`, all deliberate.

### What #728 shipped

The streaming-launch timeout test raced its own fake engine's startup against a real deadline, with an observed tail that crossed it.

- The race is now **removed, not narrowed**: the test replaces `gda.runner`'s `time` binding with a runner-local proxy whose clock crosses the deadline only after the watch has observed both expected lines. The real child, real pipes, real reader threads, the real timeout branch and real SIGTERM/reap/join all still run; only the incidental startup timing is controlled, and a real-clock safety ceiling turns a hang into a loud assertion.
- The proxy replaces the module **binding**, not the module: an earlier form mutated process-global `time.monotonic`, which the test now pins against (`time.monotonic is real_monotonic`).
- Runtime fell from a fixed 3.05 s per run to ~0.6 s, and the timeout constant lost its purpose and its obsolete cold-start margin arithmetic.

## Follow-ups filed

- **#759** — `perf --budget-file`'s `invalid_params` still renders a `ValidationError` via `str(exc)`, leaking pydantic internals and caller content. Third producer of the defect class #713 fixed for the two ADR-0015 channels; a one-line reuse of the new extractor.
- **#760** — the `res://` walk has no symlink policy: an alias re-admits the engine cache, and a cycle emits fabricated duplicate paths to the OS path limit. Both reproduced on 4.6.3, both present pre-#712.

## Validation

- Every slice re-verified in the main loop against its own gates before merge, not on the implementer's report alone.
- **Integration check before merging:** all five branches merged onto a scratch tree and the full suite run there — `1941 passed, 543 deselected`, ruff and pyright clean. Two cross-slice file overlaps (`docs/command-catalog.md`, `src/gda/commands/script.py`) were trial-merged and content-checked in both directions, since a silent keep-both merge is the hazard this wave's parallel shape creates.
- After #755 landed, the remaining four were re-merged onto the new dev tip and **#701's new description pin re-run** (`18 passed`) — a pin installed mid-wave has to be re-checked against the slices that follow it.
- Full CI with `run-e2e=true` dispatched on this branch tip: run [33171526977](https://github.com/aigengame/godot-agent/actions/runs/33171526977).

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -215,6 +215,15 @@ the Python registry in `src/gda/error_codes.py`; the table below mirrors that
 source and is checked by tests. GDScript mirrors only the rows whose source is
 `operation`, because only those codes can be reported by headless operations.
 
+The `Meaning` column is pinned too (#701): it must match the registry's
+`description` once markup and wrapping are normalized. A trailing parenthetical
+that cites at least one ADR or issue — a `Phase N` label may ride along inside
+it — is a citation rather than part of the code's meaning, so it is not compared.
+Many rows here carry one, and some registry descriptions do too: that is allowed,
+not a mistake to tidy away. A bare `(Phase N)` is *not* a citation and IS compared,
+which is why the live rows below state their phase through the `Category` column
+instead. `tests/test_error_registry.py` is the single home of that rule.
+
 Each row carries the process `Exit Code` a shell consumer keys on. It is
 per-code, not per-category: within `environment`, `binary_not_found` exits `127`
 but `launch_timeout` exits `124`. The `exit_codes.py` registry defines the
@@ -224,7 +233,7 @@ operation, and parse codes the CLI assigns).
 | Code | Category | Source | Exit Code | Meaning |
 | --- | --- | --- | --- | --- |
 | `binary_not_found` | `environment` | `runner` | `127` | The Godot binary could not be launched. |
-| `launch_timeout` | `environment` | `runner` | `124` | Godot launched but did not return before the runner timeout. One command does not report it — see the 2026-08-19 scope note (#664). |
+| `launch_timeout` | `environment` | `runner` | `124` | Godot launched but did not return before the runner timeout. One command does not report it: `scene preflight` reports its own `timeout` status instead (#664). |
 | `user_data_unwritable` | `environment` | `runner` | `127` | The log or user-data placement for the launch could not be made usable, so the launch was refused. |
 | `unknown_command` | `usage` | `classifier` | `2` | gda has no such command; discover the surface with `gda schema` or `gda --help`. A recognized near miss also carries the supported invocation in the envelope's `hint`. |
 | `unknown_option` | `usage` | `classifier` | `2` | The command exists but has no such option; read its options with `--help` or its input contract with `--schema`. A recognized near miss also carries the supported invocation in the envelope's `hint`. |
@@ -265,7 +274,7 @@ operation, and parse codes the CLI assigns).
 | `no_search_match` | `operation` | `operation` | `4` | A search-replace script edit found no occurrence of the search string. |
 | `invalid_line_range` | `operation` | `operation` | `4` | A line-range script edit specified lines outside the script's bounds, or end before start. |
 | `script_compile_failed` | `operation` | `operation` | `4` | A script does not compile, so the requested work could not proceed: `script attach` refuses to bind it to a node, and `script run` reports that the entry script (or a dependency it preloads) never ran (#651). |
-| `script_not_found` | `operation` | `classifier` | `4` | A `script run` entry script does not exist in the project, so the engine never ran it — read from stderr, since the engine still exits 0 (#651). |
+| `script_not_found` | `operation` | `classifier` | `4` | A `script run` entry script does not exist in the project, so the engine never ran it — it still exits 0, and gda reads the failure from stderr (#651). |
 | `script_failed` | `operation` | `classifier` | `4` | A `script run --strict` script ran to completion and chose a non-zero exit status; strict mode maps that opted-in failure onto the uniform error envelope. Never reported without `--strict` (ADR-0031 amendment, #651). |
 | `script_aborted` | `operation` | `classifier` | `4` | A `script run` was ended early, before its `--timeout`: a script error appeared on stderr, the caller's declared `--completion-marker` did not, and the run then went silent. Reported only when `--completion-marker` is declared (ADR-0031 amendment, #655). |
 | `incompatible_script_type` | `operation` | `operation` | `4` | A script compiles but its base type is incompatible with the requested use: `script attach`'s target node type, or `script run`'s requirement that a one-shot entry script extend `SceneTree`/`MainLoop` (#651). |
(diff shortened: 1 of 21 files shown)


qinhaihong-red:
## Wave-level full e2e: green

Dispatched on this branch tip `708578d0` with `run-e2e=true` — run [33171526977](https://github.com/aigengame/godot-agent/actions/runs/33171526977). **All 15 jobs succeeded**, including the Godot e2e job that PR CI always skips.

| Suite | Result |
| --- | --- |
| `gda` Godot e2e | `524 passed, 19 skipped, 1941 deselected in 918.83s` |
| Panda Adventure engine + e2e | `66 passed, 3 skipped, 485 deselected in 272.13s` |
| Python unit tests and package build | success |
| Type check (pyright) · Lint and format (ruff) · Both products co-install | success |
| gda-balancing (required, inventory, wheel smoke, 6 shards) | success |

This is the evidence the slice PRs could not supply: `Godot e2e tests` is schedule/dispatch-only, so every slice merged with that tier skipped. Only #712 in this wave is engine-facing, but the wave runs it whole — the 524-passed gda suite exercises the four `res://` walkers, the class-name index, and the script-run entry paths that #698's parser and pre-launch guards feed.

Ready for review. Merging this PR needs the two repo settings temporarily widened — the branch ruleset is squash-only and `allow_merge_commit` is off — because squashing here would collapse the five per-issue commits release-please derives the changelog from.

Issue #763, opened 2026-08-28
One containment answer for a res:// target, instead of three


qinhaihong-red:
## Parent

Candidate 1 of the architecture review of PR #761 (phase-2 P1 promotion). The defect that exposed
the shallowness is tracked separately as #762; this issue is the deepening that #762 leaves open.

## What to build

Three commands answer one domain question — "does this target belong to the resolved project" —
and each answers it its own way. Measured on the P1 head, on the single input `res://../outside.gd`:

| command | verdict | decided by |
|---|---|---|
| `script run` | refused | its own gate |
| `resource import` | refused | its own `..`-in-parts check |
| `script validate` | **accepted** | the declared ADR-0006 containment authority |

The two commands that behave correctly are protected by their own duplicated logic, not by the
authority. That is the shallowness: the authority's interface promises to answer containment, and
two callers found it insufficient and wrote their own answer instead of deepening it.

The gates also disagree with each other on a second input, which nothing in either suite pins:
`res://foo/../bar.gd` (net-inside, contains a literal `..`) is **accepted** by the script gate,
which canonicalises first, and **rejected** by the resource gate, which inspects path parts before
any collapsing.

After #762 makes the authority answer `res://` containment correctly, collapse the duplicated
answers onto it: one primitive decides inside / escape / other scheme, and the command gates keep
only what is genuinely theirs. For `script run` that remainder is real and must stay — its
engine-normalized trailing-suffix rule and its engine-log line-protocol rule exist to keep the
canonical identity matchable against what the engine echoes back, which is a verdict-matching
concern, not a containment one.

## This is a design decision, not a wire-up (added 2026-08-28 after an architecture review of PR #766)

The original framing above — "collapse the duplicated answers onto it" — understates the work. PR
#766 fixed the authority, and an architecture review of the result found the three gates are not
three copies of one rule; they are **three different rules**, two of which actively disagree:

- **`resource import`** rejects ANY literal `..` segment (`".." in rel.parts`), even one that
  collapses back inside. **The fixed authority** canonicalises first and rejects only a NET-upward
  remainder. On `res://foo/../bar.gd` they return opposite verdicts today — pinned deliberately by
  `tests/test_project.py::test_a_res_path_that_collapses_back_inside_is_not_outside`, which names
  this issue. Converging them means DECIDING which rule is correct, not merging two spellings of one.
- **`script run`**'s gate is not a containment predicate at all. It refuses other engine schemes
[...]

PR #766, opened 2026-08-28
fix(gda): close the res://../ containment bypass in path_outside_project

### LINKED ISSUE #762: script validate: a res://../ spelling bypasses #658's outside-the-project refusal
## Parent

Found during an architecture review of PR #761 (phase-2 P1 promotion), while tracing how many
modules answer "is this target inside the resolved project".

## What to build

**#658's containment refusal is bypassable by spelling the target as `res://../…`.** The same file,
by two spellings, gets opposite verdicts from the same command:

```
gda script validate /abs/path/outside.gd --project <p> --json
  -> {"error":{"code":"project_not_found","message":"... is outside the resolved Godot project ..."}}

gda script validate "res://../outside.gd" --project <p> --json
  -> {"valid":true,"scripts":[{"path":"res://../outside.gd","valid":true,...}],"project_root":"<p>"}
```

Reproduced on real Godot 4.6.3 against a scratch project holding no such file, with `outside.gd`
one directory above the project root.

**Root cause.** `path_outside_project` (`src/gda/project.py`) is the declared ADR-0006 containment
authority, and it returns "inside" for EVERY `res://` string:

```python
path_outside_project("res://../outside.gd", project)  ->  None   # None == inside
```

Its docstring states the assumption plainly — *"An engine-virtual path is inside by construction: it
addresses the project the engine was launched with, and gda can make no filesystem statement about
one."* That holds for a well-formed `res://a/b.gd`. It does not hold for a path that LEXICALLY
escapes the namespace: `res://../outside.gd` names a file the engine resolves outside the project,
and nothing checks for that before the short-circuit.

**Why only this command is affected.** Three commands answer the containment question, and each
answers it its own way — the one that trusts the authority is the one that is wrong:

| command | verdict on `res://../outside.gd` | decided by |
|---|---|---|
| `script run` | refused (`invalid_path`) | its own `_project_scoped_res_path` |
| `resource import` | refused (`invalid_params`, "escapes the res:// namespace") | its own `..`-in-parts check in `_asset_res_path` |
| `script validate` | **accepted** | `path_outside_project` — the declared authority |

The two protected commands are protected by duplication, not by the authority.

**Consequence.** Per `CONTEXT.md`'s `Project-code execution surface`, `script validate` compiles the
target, which runs its static initializers — so this admits code outside the `Trusted project`
through a refusal that exists precisely to prevent it. It is not a new capability class (the
absolute spelling is already refused, so the intent is settled); it is that refusal not holding.

## Acceptance criteria

- [ ] `path_outside_project` decides containment for a `res://` spelling instead of short-circuiting:
      a path that stays inside the namespace after normalization is inside, one that escapes is not.
      Its docstring's "inside by construction" reasoning is corrected in the same change — it is the
      reason the hole exists.
- [ ] `gda script validate "res://../outside.gd"` refuses, with the same code the absolute spelling
      already returns, and a regression pins both spellings together so they cannot drift apart again.
- [ ] The other two commands keep their current verdicts, and the PR states whether their private
      checks now become redundant — if they do, say so rather than leaving three answers in place.
- [ ] A sweep records every remaining site that answers the containment question, so the next reader
      learns whether it is one authority or still three.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Summary

- `path_outside_project` (the ADR-0006 containment authority, `src/gda/project.py`) short-circuited **every** `res://` string as inside the project. Its docstring's own reasoning — "an engine-virtual path is inside by construction... gda can make no filesystem statement about one" — is true for a well-formed `res://a/b.gd` but false for one that lexically escapes the namespace: `res://../outside.gd` still names a file above the project root after Godot's own `.`/`..` collapsing, and nothing checked for that.
- `script validate` trusts this authority directly (`_script_validate_recipe`), so the same file got opposite verdicts by spelling: refused as `/abs/path/outside.gd`, accepted as `res://../outside.gd`. Per `CONTEXT.md`'s Project-code execution surface, `validate` compiles the target, so the bypass admitted code outside the `Trusted project` through a refusal that exists precisely to prevent that.
- Fix, part 1: a `res://` path is now canonicalized through the shared `gda.script_errors.canonical_res_path` and refused only when it still climbs above the namespace root after that collapsing (`res://..`, `res://../outside.gd`, `res://../../x`).
- Fix, part 2 (round-2 review): **the canonicalizer itself did not match the engine on separators.** Godot folds `\` to `/` across a `res://` address *before* it collapses anything (`String::simplify_path`, `core/string/ustring.cpp:4192`, `4.6-stable-3260-g070dc9897e`), so `res://..\outside.gd` **is** the escaping `res://../outside.gd` to the engine. The fold now lives in `canonical_res_path`, where the engine puts it, so all three consumers get the engine's own identity. Applied only to a `res://` string: `\` is a legal POSIX filename character, and the fold is the engine's rule, not the filesystem's.
- A path that lexically contains `..` but collapses back inside (`res://foo/../bar.gd` → `res://bar.gd`) stays accepted — it is net-inside the namespace, exactly as Godot itself would canonicalize the address. A filename that merely starts with two dots (`res://..foo.gd`) is unaffected (checked by path segment, not string prefix). `user://` and `uid://` are untouched: gda still cannot make a filesystem statement about either.

## Engine parity of `canonical_res_path`

The round-2 review asked that the claim "canonicalized the way the engine canonicalizes one" be true or the gap be stated. `canonical_res_path`'s docstring now audits `String::simplify_path` (`core/string/ustring.cpp:4149-4233`) step by step. Reproduced: the scheme/"drive" extraction (narrowly, for an exact `res://` prefix); the `\` → `/` fold (4192); the repeated-`//` collapse and `split("/", false)` (4193-4201); the `.`/`..` collapse **with the leading-`..` strip disabled for `res://`** (4204-4221, the engine's own `FIXME`-marked exception), which is what lets an escape survive to be detected at all.

**One stated gap**, in the join (4223-4232): when every segment collapses away the engine yields the bare `res://`, `posixpath.normpath` yields `.`, so `res://a/..` canonicalizes here to `res://.`. Both spellings name the project root and every consumer already treats the pair alike — `script run`'s gate tests the pair explicitly (`_ROOT_REMAINDERS = {"", "."}`) and a `.` is not an upward escape — so the divergence is unobservable. Closing it would churn a deliberate accommodation in another command's gate, which #763 owns reconciling.

## Real-engine evidence (Godot 4.6.3, `4.6-stable-3260-g070dc9897e`)

All three spellings are one address to the engine, and it loads the file one directory above the project:

```
spelling=res://../outside.gd
  simplify_path -> res://../outside.gd
  load          -> (res://../outside.gd):<GDScript#-9223372010883643985>
spelling=res://..\outside.gd
  simplify_path -> res://../outside.gd
  load          -> (res://../outside.gd):<GDScript#-9223372010883643985>
spelling=res://a\..\..\outside.gd
  simplify_path -> res://../outside.gd
  load          -> (res://../outside.gd):<GDScript#-9223372010883643985>
```

The separator bypass was **wider than `script validate`**. At the previous head (`e6e8cd0b`), against the real engine:

| command | previous head | this head |
|---|---|---|
| `gda script validate 'res://..\outside.gd'` | `valid=true`, exit 0 | `project_not_found`, exit 4 |
| `gda script run '..\outside.gd'` | exit 0, stdout `<<<OUTSIDE-SCRIPT-RAN>>>` — **it executed a script outside the project** | `invalid_path`, exit 4, no launch |

`script run`'s gate lifts a project-relative path onto a `res://` address before canonicalizing, so the un-folded backslash reached the engine as a real traversal. That is exactly the ADR-0009 `Trusted project` widening the gate exists to prevent, and it is closed by the same one-line fold because all three gates share the canonicalizer.

## Scope discipline (per #762)

Three commands answer "is this target inside the resolved project", each its own way:

| command | `res://../outside.gd` | `res://..\outside.gd` | decided by |
|---|---|---|---|
| `script validate` | now refused (was accepted — the bug) | now refused (was accepted) | `gda.project.path_outside_project` (fixed here) |
| `script run` | refused | now refused (was **launched**) | its own `_project_scoped_res_path` (`src/gda/commands/script.py`) — its refusal *rules* are unchanged; it shares `canonical_res_path`, so the fold fixed its separator blind spot too |
| `resource import` | refused | refused on POSIX, by the later `is_file()` check rather than by its `..` gate | its own `..`-in-parts check in `_asset_res_path` (`src/gda/commands/resource.py`) — unchanged |

Neither `script run`'s nor `resource import`'s private gate becomes redundant by this fix: neither ever delegated the `res://` case to `path_outside_project` — that's exactly the "three answers" problem #763 (blocked on this PR) is scoped to unify. This PR makes the authority answer correctly, fixes the one caller (`script validate`) that trusted it, and corrects the shared canonicalizer that all three read.

`res://foo/../bar.gd` (net-inside, contains a literal `..`): the script-validate gate **accepts** it (canonicalizes first, matching Godot's own address identity), while `resource import`'s gate **rejects** it (its parts-based check refuses any literal `..` regardless of net effect). That divergence already existed before this PR and is left in place — reconciling it is explicitly #763's job, not this one's.

**Disclosed residual risk (unchanged by this PR, for #763):** `_asset_res_path` splits with `PurePosixPath`, so `res://..\x.png` is one segment and its `..` gate does not fire. On POSIX the request is still refused, one step later, by `source.is_file()` on the literal-backslash name; a Windows-native `Path` join would treat `\` as a separator and could reach the parent directory. Not reproducible on this platform, so it is reported rather than changed — `resource import` is a different command with its own gate, and #763 owns unifying them.

## Test plan

- [x] `tests/test_script_error_parser.py` — the canonicalization matrix gains `res://a\b.gd`, `res://..\outside.gd`, `res://a\..\..\outside.gd` (all folding to the engine's spelling), `res://\a.gd` (pins that the fold runs **before** the leading-slash strip, as the engine runs it before its own split), and `/abs/we\ird.gd` (a filesystem path keeps its backslashes).
- [x] `tests/test_project.py` pins the authority directly: `res://../outside.gd` and `res://..` are outside (with the resolved location reported); the new `test_a_backslash_spelled_res_escape_is_outside` covers `res://..\outside.gd` and the mixed `res://a\..\..\outside.gd`; `test_a_backslash_in_a_filesystem_path_is_not_a_separator_on_posix` pins that `..\outside.gd` as a *filesystem* path stays inside **on POSIX**, where `\` is a legal filename character — native Windows `pathlib` reads the same string as a parent escape and gda reports it outside, correctly by that platform's rule, so the test carries the repository's `os.name != "posix"` skip. The platform-independent half of that boundary — that the fold is the engine's `res://` rule and never rewrites another string — is pinned where it is already owned, in `tests/test_script_error_parser.py`'s `canonical_res_path` matrix, which now carries `a\b.gd` and `..\outside.gd` one row away from their `res://` twins; `res://foo/../bar.gd` and `res://..foo.gd` stay inside; `user://../outside.gd` / `uid://../abc123` stay inside (untouched).
- [x] `tests/test_script_commands.py` — `test_script_validate_refuses_a_res_dotdot_escape_the_same_as_the_absolute_spelling` is now parametrized over the slash, backslash and mixed spellings, each paired **in one test** with the absolute path of the same file: identical `project_not_found` verdict, identical reported location, neither reaching the engine.
- [x] `tests/test_script_run_operation.py` — the `_project_scoped_res_path` refusal matrix gains `res://..\outside.gd`, `..\outside.gd` and `res://a\..\..\outside.gd`, the spellings that previously launched.
- [x] Existing `test_script_validate_does_not_refuse_a_res_path` (well-formed `res://deck.gd`) still passes; its comment cross-references the new escape test.
- [x] `test_engine_virtual_paths_are_never_outside` renamed to `test_well_formed_engine_virtual_paths_are_not_outside`, its comment qualified — the old name asserted an absolute that the sibling escape test disproves (round-2 Standards finding).
- [x] Red-proofed: with the committed head's `src/` restored over the new tests, **10 cases failed** — 4 canonicalizer cases, `test_a_backslash_spelled_res_escape_is_outside`, 2 CLI-parity cases, 3 `script run` gate cases (the last reporting `ScriptRunResult(path='res://a\\..\\..\\outside.gd', exit_status=0)`, i.e. the pre-fix build launched the escape). With the fix restored, all 246 tests in those four files pass.
- [x] Real-engine severity check run both ways (table above), plus a legitimate in-project `gda script run inside.gd` still succeeding (exit 0).

Gates (verbatim, at `0c926509`):
```
$ uv run pytest -q -m "not e2e"
1958 passed, 543 deselected in 55.49s

$ uv run ruff check .
All checks passed!

$ uv run ruff format --check .
473 files already formatted

$ uv run pyright
0 errors, 0 warnings, 0 informations

$ uv run pytest -q -m e2e -rs tests/test_e2e_script.py tests/test_e2e_script_run.py tests/test_e2e_project.py
141 passed in 160.48s (0:02:40)

$ uv run pytest -q -m e2e -rs tests/test_e2e_scene_preflight.py tests/test_e2e_scene_validate.py tests/test_e2e_scene_security.py
30 passed in 22.88s
```

The second e2e batch is there because `canonical_res_path` is shared with the script-error parser, which the scene preflight also reads.

## Sweep (AC bullet 4)

Every remaining site that answers "is this target inside the resolved project", confirmed by grep across `src/gda`:

1. `gda.project.path_outside_project` — the declared ADR-0006 authority (fixed here).
2. `gda.commands.script._project_scoped_res_path` — `script run`'s own gate, independent rules, unchanged; shares `canonical_res_path`, so it inherits the separator fold.
3. `gda.commands.resource._asset_res_path` — `resource import`'s own gate: reuses the authority for filesystem paths, but has its own inline check for `res://` inputs, unchanged.

Still three answers, not one — #763 is the tracked follow-up to unify them.

Closes #762

### DIFF
--- a/src/gda/project.py
+++ b/src/gda/project.py
@@ -22,6 +22,8 @@ import os
 from collections.abc import Mapping
 from pathlib import Path
 
+from gda.script_errors import canonical_res_path
+
 GDA_PROJECT_ENV = "GDA_PROJECT"
 
 # The file Godot uses to mark a directory as a project root.
@@ -44,7 +46,10 @@ def is_engine_virtual_path(path: str) -> bool:
     anywhere in the string. Owned here, in the project-resolution module, and
     read by both callers of the rule: :func:`gda.models.normalize_path` (which
     passes such a path through unexpanded) and :func:`path_outside_project`
-    (which can make no filesystem statement about one).
+    (which still makes no FILESYSTEM statement about a well-formed one, but for
+    a ``res://`` spelling specifically checks it for a lexical escape of the
+    project namespace, #762 — a ``user://``/``uid://`` spelling stays inside by
+    construction, unchanged).
     """
     return path.startswith(ENGINE_VIRTUAL_PREFIXES)
 
@@ -102,6 +107,25 @@ def project_anchored(path: str, project: Path) -> Path:
     return _expand_user(project) / target
 
 
+def _res_escape_remainder(path: str) -> str | None:
+    """The canonical remainder of a ``res://`` address when it escapes upward, else ``None`` (#762).
+
+    Canonicalizes ``path`` the way Godot canonicalizes a ``res://`` address before
+    it resolves or reports one (:func:`gda.script_errors.canonical_res_path`
+    collapses ``.``/``..`` segments), then reads what is left: an exact ``..`` or a
+    leading ``../`` means the address is still climbing above the namespace root
+    AFTER that collapsing, which is genuinely outside. Anything else is not an
+    escape, including a ``res://foo/../bar.gd`` spelling — it collapses to
+    ``res://bar.gd``, net-inside — and a filename that merely STARTS with two dots
+    (``res://..foo.gd`` names a real file, not a traversal; the test is the first
+    PATH SEGMENT, not a string prefix).
+    """
+    remainder = canonical_res_path(path)[len("res://") :]
+    if remainder == ".." or remainder.startswith("../"):
+        return remainder
+    return None
+
+
 def path_outside_project(path: str, project: Path) -> Path | None:
     """The location of ``path`` when it falls OUTSIDE ``project``.
 
@@ -110,9 +134,19 @@ def path_outside_project(path: str, project: Path) -> Path | None:
     project, and otherwise the target's real location, so the caller can name
     *where* it actually is in its diagnostic.
 
-    An engine-virtual path is inside by construction: it addresses the project
-    the engine was launched with, and gda can make no filesystem statement
-    about one.
+    A ``res://`` address is checked LEXICALLY against the namespace it names,
+    rather than trusted by construction: :func:`_res_escape_remainder`
+    canonicalizes it the way the engine does and refuses one that still steps
+    above the namespace root after that collapsing (``res://../outside.gd``).
+    gda still makes no FILESYSTEM statement about a well-formed ``res://``
+    path — it is the engine's own launch ``--path`` that owns resolving one,
+    and this check never touches the filesystem to decide — but a spelling
+    that lexically escapes the namespace is not well-formed, and admitting it
+    here defeated the very refusal this function exists to make (#762). The
+    other two engine-virtual schemes are unaffected and stay inside by
+    construction: ``user://`` addresses the engine's own data directory, not a
+    child of the project tree, and ``uid://`` is an opaque identifier with no
+    path structure to escape through.
 
     A filesystem path is first anchored the way the engine will address it
     (:func:`project_anchored`), then read in up to two ways — it belongs to the
@@ -149,7 +183,12 @@ def path_outside_project(path: str, project: Path) -> Path | None:
     location it would occupy rather than raising.
     """
     if is_engine_virtual_path(path):
-        return None
+        if not path.startswith("res://"):
+            return None
+        escape = _res_escape_remainder(path)
+        if escape is None:
+            return None
+        return (_expand_user(project) / escape).resolve()
     root = _expand_user(project)
     candidate = project_anchored(path, project)
     location = candidate.resolve()
--- a/src/gda/project.py
+++ b/src/gda/project.py
@@ -111,12 +111,16 @@ def _res_escape_remainder(path: str) -> str | None:
     """The canonical remainder of a ``res://`` address when it escapes upward, else ``None`` (#762).
 
     Canonicalizes ``path`` the way Godot canonicalizes a ``res://`` address before
-    it resolves or reports one (:func:`gda.script_errors.canonical_res_path`
-    collapses ``.``/``..`` segments), then reads what is left: an exact ``..`` or a
-    leading ``../`` means the address is still climbing above the namespace root
-    AFTER that collapsing, which is genuinely outside. Anything else is not an
-    escape, including a ``res://foo/../bar.gd`` spelling — it collapses to
-    ``res://bar.gd``, net-inside — and a filename that merely STARTS with two dots
+    it resolves or reports one — :func:`gda.script_errors.canonical_res_path` folds
+    ``\\`` to ``/`` and collapses ``.``/``..`` segments, and its docstring is where
+    that correspondence to ``String::simplify_path`` is audited step by step — then
+    reads what is left: an exact ``..`` or a leading ``../`` means the address is
+    still climbing above the namespace root AFTER that collapsing, which is
+    genuinely outside. The fold is what makes ``res://..\\outside.gd`` reach this
+    test as the escape the engine treats it as, rather than as an in-project
+    filename (PR #766 review). Anything else is not an escape, including a
+    ``res://foo/../bar.gd`` spelling — it collapses to ``res://bar.gd``,
+    net-inside — and a filename that merely STARTS with two dots
     (``res://..foo.gd`` names a real file, not a traversal; the test is the first
     PATH SEGMENT, not a string prefix).
     """
--- a/src/gda/script_errors.py
+++ b/src/gda/script_errors.py
@@ -173,18 +173,49 @@ def canonical_res_path(path: str) -> str:
     engine's spelling against the caller's raw one missed the match and let a
     failed run report success.
 
-    Collapses ``.``/``..`` segments and duplicate slashes on the path part only
-    (posixpath semantics), leaving the ``res://`` scheme intact. Purely lexical —
-    no filesystem access — so it is safe on a path that does not exist, which is
-    exactly the missing-entry-script case. A non-``res://`` string is returned
-    unchanged: this normalizes an address, it does not validate one.
+    Purely lexical — no filesystem access — so it is safe on a path that does not
+    exist, which is exactly the missing-entry-script case. A non-``res://`` string
+    is returned unchanged: this normalizes an address, it does not validate one.
+
+    **Against ``String::simplify_path``** (``core/string/ustring.cpp:4149-4233``,
+    Godot ``4.6-stable-3260-g070dc9897e``), the engine function every ``res://``
+    address passes through before the engine resolves or reports it
+    (``ProjectSettings::localize_path``, ``core/config/project_settings.cpp:158``).
+    Which of its steps this reproduces, in the engine's own order:
+
+    - **scheme extraction** (4153-4168: first ``://`` whose prefix is all ASCII
+      alphanumerics becomes the "drive") — reproduced NARROWLY, for an exact
+      ``res://`` prefix only. The engine's other two drive branches (network share,
+      Windows ``C:``) are deliberately NOT reproduced: they are unreachable once the
+      scheme branch matched, and a non-``res://`` string leaves here untouched anyway.
+    - **``\\`` → ``/`` across the whole remainder** (4192) — reproduced, and it must
+      run BEFORE the leading-slash strip below, exactly as the engine runs it before
+      its own empty-segment split: ``res://\\a.gd`` folds to ``res://a.gd``, which is
+      no longer possible once the strip has already passed over a backslash. Without
+      this step ``res://..\\outside.gd`` read as an ordinary in-project filename
+      while the engine loaded the file one directory ABOVE the project and reported
+      it back as ``res://../outside.gd`` (#762).
+    - **repeated-``//`` collapse and ``split("/", false)``** (4193-4201) — reproduced
+      by the leading-slash strip plus ``posixpath.normpath``, which collapses runs of
+      separators and drops a trailing one. The strip is what covers POSIX's one
+      divergence: it gives exactly two leading slashes a special meaning (``//a``
+      stays ``//a``), so ``res:////a.gd`` would otherwise stay uncanonicalized.
+    - **``.``/``..`` collapse with the leading-``..`` strip DISABLED for ``res://``**
+      (4204-4221) — reproduced: ``normpath`` on a RELATIVE remainder keeps a leading
+      ``..`` for the same reason the engine keeps it, and that is what lets a caller
+      of this function see an escape at all rather than have it silently swallowed.
+
+    One stated gap, in the join (4223-4232): when every segment collapses away the
+    engine yields the bare ``res://`` while ``normpath`` yields ``.``, so
+    ``res://a/..`` canonicalizes here to ``res://.``. Both spellings name the project
+    root and every consumer already treats the pair alike — ``script run``'s gate
+    tests the pair explicitly (``_ROOT_REMAINDERS``) and a ``.`` is not an upward
+    escape — so closing it would only churn a deliberate accommodation that #763
+    owns reconciling.
     """
     if not path.startswith(_RES_PREFIX):
         return path
-    # Strip leading slashes before normpath: POSIX gives exactly two leading
-    # slashes a special meaning ("//a" stays "//a"), which would leave a
-    # `res:////a.gd` spelling uncanonicalized.
-    remainder = path[len(_RES_PREFIX) :].lstrip("/")
+    remainder = path[len(_RES_PREFIX) :].replace("\\", "/").lstrip("/")
     if not remainder:
         return _RES_PREFIX
     # normpath("") is ".", so the empty case is handled above rather than here.
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -92,10 +92,12 @@ def test_path_in_a_sibling_tree_is_outside_and_reports_its_location(tmp_path):
     assert path_outside_project(str(script), proj) == script.resolve()
 
 
-def test_engine_virtual_paths_are_never_outside(tmp_path):
+def test_well_formed_engine_virtual_paths_are_not_outside(tmp_path):
     # res:// (and its user:// / uid:// siblings) address the project the engine
-    # was launched with, so they are inside by construction — gda makes no
-    # filesystem statement about them (ADR-0006).
+    # was launched with, so gda makes no filesystem statement about them
+    # (ADR-0006). WELL-FORMED is the qualifier the name carries: a res:// spelling
+    # that lexically escapes the namespace is not inside by construction and IS
+    # refused — see test_a_res_dotdot_escape_is_outside below (#762).
     proj = _make_project(tmp_path / "game")
 
     assert path_outside_project("res://hero.gd", proj) is None
@@ -120,6 +122,36 @@ def test_a_res_dotdot_escape_is_outside(tmp_path):
     )
     # The bare escape, with no filename after it, is refused the same way.
     assert path_outside_project("res://..", proj) == tmp_path.resolve()
+
+
+def test_a_backslash_spelled_res_escape_is_outside(tmp_path):
+    # The separator bypass (PR #766 round-2 review): Godot folds `\` to `/` across
+    # a res:// address before it collapses anything (ustring.cpp:4192), so
+    # `res://..\outside.gd` IS the escaping address to the engine — a real 4.6.3
+    # run loads the file one directory above the project and reports it back as
+    # `res://../outside.gd`. Reading the backslash as a filename character let this
+    # spelling through the very check the slash spelling above is refused by.
+    proj = _make_project(tmp_path / "game")
+
+    assert (
+        path_outside_project("res://..\\outside.gd", proj)
+        == (tmp_path / "outside.gd").resolve()
+    )
+    # Mixed separators collapse together, exactly as the engine collapses them
+    # after the fold: `a\..\..\outside.gd` is `a/../../outside.gd`, net one level up.
+    assert (
+        path_outside_project("res://a\\..\\..\\outside.gd", proj)
+        == (tmp_path / "outside.gd").resolve()
+    )
+
+
+def test_a_backslash_in_a_filesystem_path_is_not_a_separator(tmp_path):
+    # The fold is the ENGINE's res:// rule, not the filesystem's: `\` is a legal
+    # POSIX filename character, so an ordinary path keeps it and is anchored under
+    # the project as the single-segment name it is.
+    proj = _make_project(tmp_path / "game")
+
+    assert path_outside_project("..\\outside.gd", proj) is None
 
 
 def test_a_res_path_that_collapses_back_inside_is_not_outside(tmp_path):
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -103,6 +103,56 @@ def test_engine_virtual_paths_are_never_outside(tmp_path):
     assert path_outside_project("uid://abc123", proj) is None
 
 
+# --- a res:// spelling that LEXICALLY escapes the namespace (#762) ------------
+
+
+def test_a_res_dotdot_escape_is_outside(tmp_path):
+    # #762: `path_outside_project` used to short-circuit EVERY res:// string as
+    # inside, so `res://../outside.gd` bypassed containment even though it names
+    # a file one directory above the project root — exactly what the absolute
+    # spelling of the same file is already refused for. The reported location
+    # lets a caller compare the two spellings' refusals directly.
+    proj = _make_project(tmp_path / "game")
+
+    assert (
+        path_outside_project("res://../outside.gd", proj)
+        == (tmp_path / "outside.gd").resolve()
+    )
+    # The bare escape, with no filename after it, is refused the same way.
+    assert path_outside_project("res://..", proj) == tmp_path.resolve()
+
+
+def test_a_res_path_that_collapses_back_inside_is_not_outside(tmp_path):
+    # A `..` a lexical collapse cancels out stays inside: `res://foo/../bar.gd`
+    # normalizes to `res://bar.gd`, net-inside the namespace, exactly as Godot
+    # itself would canonicalize the address. `resource import`'s own res://
+    # gate (`_asset_res_path`) is stricter — it refuses ANY literal `..`
+    # component regardless of net effect — and #763 tracks reconciling the two;
+    # this authority's rule is decided here.
+    proj = _make_project(tmp_path / "game")
+
+    assert path_outside_project("res://foo/../bar.gd", proj) is None
+
+
+def test_a_res_path_starting_with_two_dots_is_not_an_escape(tmp_path):
+    # `res://..foo.gd` names a real file whose name merely starts with two dots,
+    # not a traversal — the escape test is the first PATH SEGMENT, not a string
+    # prefix.
+    proj = _make_project(tmp_path / "game")
+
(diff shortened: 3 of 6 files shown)


qinhaihong-red:
Reviewed exact head `e6e8cd0b9cdcca71cbb82fc5e754f7daf436d996` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2` (`git diff b616e7f...e6e8cd0`).

## Artifact / Markdown

**Changes required — 1 finding.**

### [P1] Correct the claim that the new check uses the engine's canonicalization

Location: PR body, Summary bullet 3 and the corresponding test-plan claims.

The body says the path is “canonicalized the way the engine canonicalizes one.” That is not true for separators. Godot 4.6.3 `String::simplify_path()` converts ``\\`` to `/` before it collapses `.` and `..`; the reused `canonical_res_path` applies `posixpath.normpath` without that conversion. At this head, `res://..\\outside.gd` passes `path_outside_project`, and the real CLI reports `valid=true` for the outside script that its absolute spelling refuses. This makes the stated scope and completion evidence materially misleading.

Smallest correction: make the implementation and regressions cover Godot-equivalent separator spellings, then update the body to name that coverage. If this is intentionally deferred, the body must state the remaining bypass and cannot claim that #762's normalization boundary is closed.

**Terminology and prose assessment.** I reviewed the complete PR body and issue #762, including status/labels, against issue #658, follow-up #763, `CONTEXT.md` (“Trusted project”, “Project-code execution surface”), ADR-0006, ADR-0009, the changed source/tests, and current CI. ASD-STE100 was the writing reference; I checked technical nouns/verbs, alternate names, causal/scope/order wording, and every requirement/procedure passage. The project terms are otherwise consistent and the sweep names the three relevant gates accurately. No separate substantive terminology or prose issue was found. The claimed local red-proof run was not independently reconstructed.

## Architecture / DDMA

**Changes required — 1 finding.**

### [P1] The containment authority and the engine still disagree on resource-path identity

Location: `src/gda/project.py:110-125`, consuming `src/gda/script_errors.py:167-191`.

Observed flow: `commands/script._script_validate_recipe` delegates containment to `project.path_outside_project`, which delegates `res://` identity to `script_errors.canonical_res_path`. The ownership direction is already acknowledged for later repair in #763; that relocation alone need not block this PR. The blocking defect is the contract mismatch: the diagnostic parser's POSIX-only helper is now used as the project-containment knowledge source, but it does not implement Godot's separator normalization.

Exact-head evidence:

```text
canonical_res_path('res://..\\outside.gd') -> 'res://..\\outside.gd'
path_outside_project(..., project)         -> None
gda script validate 'res://..\\outside.gd' -> valid=true
gda script validate /absolute/outside.gd   -> project_not_found (exit 4)
```

A direct Godot 4.6.3 run of the backslash spelling loaded the same one-level-outside test script and printed its marker. Thus the responsibility set is incomplete at the boundary even though the source graph is acyclic: one engine identity still has opposite containment verdicts. The slash, dot, repeated `..`, duplicate-slash, filename-prefix, missing-path, absolute-path, root, prefix-collision, and existing symlink cases otherwise retain owners and directed calls.

Smallest repair: make the shared canonicalizer match Godot's `\\`-to-`/` step before segment collapse, and add pure authority plus real/CLI parity regressions for backslash and mixed-separator escapes. Keep #763 as the place to relocate the primitive and reconcile the three command policies.

Structural claim `DDMA-766-C1` — “`canonical_res_path` is a Godot-equivalent resource identity suitable for containment”: evidence state **open** (the project defines no stronger claim-state vocabulary). Supporting evidence is the passing slash matrix; the observed backslash execution is the disconfirming condition. Validation remains open for a Windows-native run, although the Godot 4.6.3 conversion and the macOS real-engine reproduction already establish the present mismatch.

Completeness: fail on separator normalization. Orthogonality/DRY: the authority is reused, but it reuses an incomplete contract from the wrong responsibility area; relocation is tracked in #763. Extensibility: no new abstraction is needed. Retain the current architecture after the narrow contract repair.

## Standards

**Changes required — 1 judgment-call finding; no hard documented-standard violation.**

### Possible Mysterious Name: the retained test now states an absolute that the new tests disprove

Location: `tests/test_project.py:95-103`.

The retained name `test_engine_virtual_paths_are_never_outside` and its “inside by construction” comment conflict with the newly added `test_a_res_dotdot_escape_is_outside` at lines 109-122. Rename it to a qualified form such as `test_well_formed_engine_virtual_paths_are_not_outside` and qualify the comment. This is a Fowler-smell judgment call, not a tooling-enforced or hard repository violation. No other actionable baseline smell was found.

## Spec

**Changes required — 1 finding.**

### [P1] Apply Godot's separator normalization before deciding containment

Issue #762 requires: “`path_outside_project` decides containment for a `res://` spelling … a path that stays inside the namespace after normalization is inside, one that escapes is not.” It also says that the invalid spelling “names a file the engine resolves outside the project.” This is a behavioral requirement after engine normalization, not a restriction to the example's `/` spelling.

Because `res://..\\outside.gd` is accepted by the head check and normalized by the real engine to the escaping address, the first acceptance criterion is partial and the same absolute-vs-virtual contradiction remains. Normalize backslashes before `posixpath.normpath`, then pair that spelling and mixed-separator variants with the absolute file in authority and CLI regressions. The `/` spelling, disclosure/sweep, and adjacent-command requirements otherwise match #762; no unrelated scope creep was found.

## Validation and limits

- Immutable exact-head archive: 75 changed-area tests passed (`tests/test_project.py` and `tests/test_script_commands.py`); changed-file Ruff check and format check passed.
- Real Godot 4.6.3: the backslash spelling loaded the outside script; exact-head `gda script validate` returned `valid=true`, while the absolute spelling returned `project_not_found`.
- Current exact-head CI: seven checks pass; four jobs are skipped, including Godot e2e. The unit/build log reports 1,944 passed, 2 skipped, and 543 deselected.
- Not performed: full local non-e2e suite, Windows-native execution, full user/uid scheme runtime matrix, or reconstruction of the PR body's pre-fix red-proof procedure.

Axis summary: Artifact 1 (normalization overclaim); Architecture 1 (containment/engine identity mismatch); Standards 1 judgment call (stale absolute test name); Spec 1 (normalization AC incomplete).

qinhaihong-red:
## Round-2 review handled — all four findings adopted, at `0c926509`

The four axes name one defect from four sides: the shared `res://` canonicalizer did not fold `\` to `/`, so a backslash-spelled escape read as an in-project filename. Fixed at the canonicalizer, not at the containment helper, because that is where the engine puts the step.

| **Architecture P1** — the containment authority and the engine disagree on resource-path identity; smallest repair is the `\`→`/` fold in the shared canonicalizer plus authority and CLI parity regressions | **Fully adopt**, exactly as suggested | Reproduced independently on Godot 4.6.3 (`4.6-stable-3260-g070dc9897e`): `res://../outside.gd`, `res://..\outside.gd` and `res://a\..\..\outside.gd` all `simplify_path` to `res://../outside.gd` and `load` the file above the project | `src/gda/script_errors.py` `canonical_res_path` folds `\`→`/` before the leading-slash strip — `0c926509` | Red-proofed (10 failures on the previous head), 246 tests green in the four touched files, full non-e2e 1958 passed, 171 targeted e2e passed, PR CI green |
| **Spec P1** — apply the engine's separator normalization before deciding containment; pair the spellings with the absolute file in authority and CLI regressions | **Fully adopt** | AC 1 of #762 is behavioural after engine normalization, and the backslash spelling was the same file with the opposite verdict | `tests/test_project.py` (`test_a_backslash_spelled_res_escape_is_outside`, plus `test_a_backslash_in_a_filesystem_path_is_not_a_separator`); `tests/test_script_commands.py` parity test parametrized over slash / backslash / mixed, each paired with the absolute spelling — `0c926509` | Both spellings return `project_not_found` exit 4 with the identical resolved location, neither reaches the engine |
| **Artifact P1** — the body's "canonicalized the way the engine canonicalizes one" is not true for separators | **Fully adopt** | The claim was false at the previous head | Body rewritten: the fold is named, a new "Engine parity of `canonical_res_path`" section audits `String::simplify_path` step by step, and the real-engine before/after table replaces the earlier prose — `0c926509` | The docstring carries the same audit, so the code and the body cannot drift |
| **Standards** — `test_engine_virtual_paths_are_never_outside` asserts an absolute its sibling disproves | **Fully adopt** | Cheap and correct | Renamed to `test_well_formed_engine_virtual_paths_are_not_outside`, comment qualified with the cross-reference — `0c926509` | Pure rename; passes unchanged on the pre-fix source, so it carries no red-proof of its own |

### The bypass was wider than the review found

`script run`'s gate lifts a project-relative path onto a `res://` address before canonicalizing, so it shared the blind spot. Against the real engine at the previous head `e6e8cd0b`:

```
$ gda script run '..\outside.gd' --project <proj> --json
{"path":"res://..\\outside.gd","exit_status":0,
 "stdout":"...\n<<<OUTSIDE-SCRIPT-RAN>>>\n", ...}
exit=0
```

It **executed** a script one directory above the project — the ADR-0009 `Trusted project` widening that gate exists to prevent, not merely a wrong `valid=true`. At `0c926509` the same command returns `invalid_path`, exit 4, no launch, and `tests/test_script_run_operation.py`'s refusal matrix gains `res://..\outside.gd`, `..\outside.gd`, `res://a\..\..\outside.gd` (all three red on the previous head, one of them reporting `ScriptRunResult(..., exit_status=0)`). A legitimate `gda script run inside.gd` still exits 0.

This is why the fold went into `canonical_res_path` rather than into `_res_escape_remainder`: the engine applies it to every `res://` address, so one change gives all three consumers — the containment authority, the parser's entry matching, and `script run`'s acceptance gate — the engine's own identity. Scoped to `res://` strings only: `\` is a legal POSIX filename character, so a filesystem path keeps its backslashes (pinned by `test_a_backslash_in_a_filesystem_path_is_not_a_separator` and a `/abs/we\ird.gd` case in the canonicalizer matrix).

### `DDMA-766-C1`

The disconfirming condition you recorded is now closed at the canonicalizer, with the observed-execution evidence above added to it. The claim's supporting evidence set is the slash matrix plus the backslash/mixed matrix, re-derived against a live 4.6.3 engine. **A Windows-native run is still not performed** — the fold is platform-independent string handling, but I am not claiming Windows validation I do not have.

### Two things deliberately not changed, both disclosed

1. **One stated parity gap, not closed.** In `simplify_path`'s join (`ustring.cpp:4223-4232`) a fully collapsing address yields the bare `res://`, while `posixpath.normpath` yields `.`, so `res://a/..` canonicalizes here to `res://.`. It is unobservable: both name the project root, `script run`'s gate tests the pair explicitly (`_ROOT_REMAINDERS = {"", "."}`), and `.` is not an upward escape. Closing it would obsolete that deliberate accommodation in another command's gate, which #763 owns reconciling. The gap is now stated in the docstring and in the PR body rather than papered over.
2. **`resource import`'s gate, unchanged, with a disclosed Windows-only risk.** `_asset_res_path` splits with `PurePosixPath`, so `res://..\x.png` is one segment and its `..` check does not fire — verified: it returns the path unchanged. On POSIX the request is still refused one step later by `source.is_file()` on the literal-backslash name (verified `False`), so there is no reachable bypass here. A Windows-native `Path` join would treat `\` as a separator; I cannot reproduce that on this platform, and `resource import` is a different command with its own gate, so it is reported for #763 rather than changed on speculation.

### Validation at `0c926509`

```
uv run pytest -q -m "not e2e"          1958 passed, 543 deselected in 55.49s
uv run ruff check .                    All checks passed!
uv run ruff format --check .           473 files already formatted
uv run pyright                         0 errors, 0 warnings, 0 informations
e2e: script / script_run / project     141 passed in 160.48s
e2e: scene preflight/validate/security  30 passed in 22.88s
```

The second e2e batch is included because `canonical_res_path` is shared with the script-error parser, which the scene preflight also reads. PR CI is green on this head (unit/build, ruff, pyright, co-install, scope guards); Godot e2e stays schedule/dispatch-only, so those two batches are local evidence.

qinhaihong-red:

Reviewed exact head `0c926509e694369b19ac01dfd9f7fb733307622b` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review and the author's reply.

The original containment finding is closed. The shared canonical authority now normalizes backslashes before containment and diagnostic identity checks; exact-head CLI probes reject backslash and mixed-separator escapes before Godot launches.

### Artifact / Markdown — Pass

The PR body now accurately states the `/`, `\\`, mixed-separator, root-collapse, and unverified-Windows boundaries. The terminology and prose pass used the project glossary, issues #762/#763, relevant ADRs, and ASD-STE100; no substantive artifact issue remains.

Containment, diagnostic identity, and `script run` share the same canonical identity owner. The longer-term primitive placement and resource-import policy remain explicitly tracked by #763 and do not block this slice.

- `tests/test_project.py:148-154` unconditionally names and asserts that a backslash in a filesystem path is not a separator. That premise is POSIX-only. The repository supports Windows headless operation, and native Windows `pathlib.Path` treats `..\\outside.gd` as a parent-directory escape. Make this containment test POSIX-only or assert platform-specific outcomes; a separate cross-platform test may still prove that the `res://` canonicalizer does not rewrite non-virtual strings.

### Spec — Changes required

- The same unconditional test encodes the wrong Windows behavior for the containment contract. Production behavior for the original `res://` escape is fixed; the remaining issue is the new platform-independent test claim.

### Validation

- 246 changed/adjacent tests, Ruff, and format checks passed.
- Exact-head CLI rejected `res://..\\outside.gd` and a mixed-separator escape with exit 4 / `project_not_found` before engine launch.
- Godot 4.6.3 independently simplified the spelling to `res://../outside.gd` and could load the outside file, confirming the production guard now uses the correct identity.
- Required CI checks passed; hosted Godot E2E remains skipped. Windows-native behavior, the full local non-E2E suite, and the Windows resource-import risk tracked by #763 were not run.

qinhaihong-red:

Head is now `ed114840`.

### Standards / Spec — the unconditional POSIX premise. **Adopted.**

Confirmed by reading: `test_a_backslash_in_a_filesystem_path_is_not_a_separator` asserted `path_outside_project("..\outside.gd", proj) is None` with no platform qualification. On native Windows `pathlib` reads that string as a parent escape and `path_outside_project` reports it OUTSIDE — correctly, by that platform's own rule. The test encoded the wrong contract for a platform the repository supports.

Both of your suggested treatments were taken, because the original test was asserting two different things under one name:

- **The platform-independent invariant it was reaching for** now stands alone and runs everywhere — `test_the_fold_does_not_reach_a_non_res_string` asserts that `canonical_res_path` leaves a non-`res://` string untouched, which is the actual cross-platform claim ("the fold is the engine's `res://` rule, not the filesystem's").
- **The POSIX consequence** keeps the containment assertion, renamed `..._on_posix` and gated by the repository's established `pytest.mark.skipif(os.name != "posix", ...)` idiom. The Windows behaviour is named in the comment rather than left for a Windows runner to discover.

Red-proof: making `canonical_res_path` fold backslashes in non-`res://` strings fails the new invariant test and nothing else (`1 failed, 25 passed`); restoring it returns `26 passed`.

Production behaviour is unchanged — `res://` containment was already correct on both platforms, and no `src/` file is touched by this commit.

### Validation

`1959 passed, 543 deselected` (non-e2e), ruff check / format clean, pyright 0 errors. `Closes #762` intact. The Windows resource-import risk stays tracked on #763, where it is now recorded with the measured POSIX boundary.

qinhaihong-red:

Reviewed exact head `ed1148404c8c45d16ad1ff1695d43b3a38d111f3` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review, the author's reply, and the incremental change since `0c926509`.

The previous platform defect is closed. The tests now separate the cross-platform canonical-string invariant from the POSIX filesystem consequence, use the repository's established POSIX skip idiom, and explicitly explain Windows behavior. Production canonicalization and containment code did not change.

Two current residues remain:

1. **Artifact / Markdown:** the PR body Test plan still names the deleted `test_a_backslash_in_a_filesystem_path_is_not_a_separator` and still says, without a platform qualifier, that `..\outside.gd` “stays inside.” Update it to name the current split tests and state that the containment result is POSIX-specific.
2. **Standards — duplicated test ownership:** `tests/test_project.py:150-156` imports `canonical_res_path` to retest the non-`res://` backslash-preservation branch already owned by the parameter matrix in `tests/test_script_error_parser.py:254-263`. Move the two regression spellings into that existing canonicalizer matrix, remove the duplicate test/import, and retain the POSIX containment consequence test in `test_project.py`.

Axis result: **Artifact 1 / Standards 1 — Changes required; Architecture / DDMA and Spec — Pass.** Issue #763 remains the appropriate owner for broader canonicalizer/resource-import policy work.

Validation: 247 changed/adjacent tests passed; Ruff and format passed; exact-head CLI probes rejected backslash and mixed-separator escapes before engine launch; Godot 4.6.3 confirmed the same canonical identity. All non-skipped CI checks succeeded; hosted Godot E2E remains skipped. Windows-native and Windows resource-import behavior were not executed.

qinhaihong-red:
## Round-4 reply — both findings adopted

Head is now `ff20b394`.

### Standards — duplicated test ownership. **Adopted, and you named the right owner.**

The round-3 split was correct in substance and wrong in placement: the `canonical_res_path` matrix in `tests/test_script_error_parser.py` already owns the non-`res://` preservation branch — it is why `relative.gd` and `/abs/we\ird.gd` are pinned there. Putting a second test of the same branch in `tests/test_project.py` gave one fact two homes.

The two regression spellings moved into that matrix, deliberately placed one row away from their `res://` twins, so the contrast the round-3 test was drawing is now visible in a single place:

```
("res://a\\b.gd",        "res://a/b.gd"),
("res://..\\outside.gd", "res://../outside.gd"),
...
("a\\b.gd",              "a\\b.gd"),
("..\\outside.gd",       "..\\outside.gd"),
```

`test_project.py` keeps only the POSIX containment consequence and no longer imports `canonical_res_path`.

Red-proof re-run in the new home: letting the fold reach non-`res://` strings fails both relocated rows plus the pre-existing `/abs/we\ird.gd` row (`3 failed, 53 passed`); restoring gives `56 passed`. Coverage is preserved, not just relocated.

### Artifact — stale Test plan. **Adopted.**

The PR body named the deleted test and stated the containment result with no platform qualifier. It now names `test_a_backslash_in_a_filesystem_path_is_not_a_separator_on_posix`, says the result is POSIX-specific and why (`\` is a legal filename character there, while native Windows `pathlib` reads the same string as a parent escape and gda reports it outside — correctly, by that platform's rule), and points at the canonicalizer matrix as the home of the platform-independent half.

### Validation

`1960 passed, 543 deselected` (non-e2e), ruff check / format clean, pyright 0 errors. No `src/` file is touched by this commit. `Closes #762` intact; #763 still owns the broader canonicalizer/resource-import policy work.

Issue #752, opened 2026-08-27
bug(gda): prevent tiny-float loss in live JSON transport


qinhaihong-red:
## Problem

The live protocol transports request and result frames as JSON. On Godot 4.6.3, the engine can silently change non-zero finite `double` values near the lower end of the range to `0.0`.

This is not limited to IEEE-754 subnormal values. The result depends on the decimal literal shape and significant-digit count:

| JSON literal | Godot `JSON.parse_string` result |
| --- | --- |
| `1e-300` | non-zero |
| `1e-308` | non-zero |
| `1.2345678901234567e-300` | `0.0` |
| `2.2250738585072014e-308` (`DBL_MIN`) | `0.0` |
| `1e-320` | `0.0` |
| `5e-324` | `0.0` |

Python's `json.dumps(1.2345678901234567e-300)` emits the failing full-precision form. A `game call` can therefore report success after invoking the declared method with `0.0`, although the caller supplied a non-zero ordinary normal `double`.

The result direction is also affected. Godot 4.6.3 `JSON.stringify(1e-300)` emits `0.0`, so any live read that returns a sufficiently small float can lose it before the daemon receives the result. This is wider than `game call`.

## Reproduction and root-cause evidence

- Godot version: `4.6.3.stable.official.7d41c59c4`.
- Input path: `JSON::_get_token` calls `String::to_float` (`core/io/json.cpp`), which uses Godot's `built_in_strtod` implementation (`core/string/ustring.cpp`).
- Result path: the harness uses Godot `JSON.stringify` for result envelopes.
- PR #749 reproduces the behavior while reviewing the live-call numeric contract.

## Desired outcome

Choose and implement one evidence-backed live-number transport policy that prevents silent success on a changed numeric value in both directions, or explicitly scopes any unavoidable limitation. Candidate approaches include:

- use a Godot version whose JSON parser/stringifier preserves the required binary64 domain, with a pinned conformance matrix;
- add a NARROW, value-level tagged representation for the values a corpus proves are lossy —
  not a renegotiated wire format. **ADR-0021 explicitly considered and REJECTED a bespoke
  daemon↔harness wire format** ("reusing the ADR-0002 sentinel is what lets the entire
  classify/parse/emit pipeline be reused unchanged", Considered options). A protocol-level
  change would overturn that decision and needs a dated ADR-0021 amendment — it is not
  "a harness version bump";
- add a proven, conservative boundary guard for affected values while retaining an honest public contract.

[...]

Issue #721, opened 2026-08-19
scene validate: recurse into instanced sub-scenes for a composed verdict


qinhaihong-red:
## Parent

Follow-up from #664 (delivered by PR #720); the underlying dogfooding note is GDA-DF-040.

## What to build

`scene validate` checks only the scene file it is given: a `parent.tscn` that instances a broken child scene returns `valid: true` while the child's script and texture are missing (reproduced with a real engine during #720's review). GDA-DF-040's own case is a composed encounter scene, so an admission flow that gates on `scene validate` alone still infers a weaker static result than the scene's real load behavior. `scene preflight` does catch the broken child dynamically (`started: false`), so the pair covers the gap today — this issue closes it statically.

Extend `scene validate` so a scene's instanced sub-scenes are validated as part of the parent's verdict: resolve each `[ext_resource]` of type `PackedScene`, validate it with the same checks, and attribute each problem to the sub-scene file it was found in (so a problem inside `child.tscn` is not misread as a problem in `parent.tscn`). Decide and document the cycle/depth behavior (a scene cycle must terminate with a clear diagnostic, not hang) and whether an already-visited sub-scene is re-reported per instancing site or once per file.

## Acceptance criteria

- [ ] `scene validate parent.tscn` reports the missing dependencies of an instanced child scene, attributed to the child file.
- [ ] A sub-scene cycle terminates with a structured diagnostic instead of recursing forever.
- [ ] A known-good composed scene still validates `valid: true`.
- [ ] e2e coverage on a composed known-bad and known-good scene pair.

## Blocked by

None - can start immediately.

qinhaihong-red:
> *This was generated by AI during triage.*

## Premise re-reproduced on `main` (2026-08-27, pre-dispatch review)

Confirmed against a real Godot 4.6.3 headless run, on a minimal pair — `child.tscn` carrying `[ext_resource type="Script" path="res://missing_script.gd"]`, `parent.tscn` instancing it:

```
gda scene validate child.tscn   -> {"valid":false,"problems":[{"kind":"missing_resource",
                                    "path":"res://missing_script.gd","type":"Script", ...}]}
gda scene validate parent.tscn  -> {"valid":true,"problems":[]}
```

Two implementation facts worth having before the slice starts:

[...]

Issue #283, opened 2026-06-24
feat(diag): error callstacks in `diag errors`


qinhaihong-red:
## What to build

Enrich `gda diag errors` with the full **call stack** for each runtime error, so an agent sees *where* an error originated, not just the top `file:line`.

Each error record gains a `callstack` of ordered `{function, file, line}` frames. Source the backtrace from the session log when the engine emits one (script errors carry an `at:` chain) and/or a harness error hook capturing `get_stack()` at the error site — the slice picks the most reliable source and documents it. Read-only observability, in scope per ADR-0025; does **not** tap the remote-debug protocol.

## Acceptance criteria

- [ ] `gda diag errors --json` error records carry a `callstack` (ordered frames `{function, file, line}`) when available.
- [ ] When no backtrace is available, `callstack` is empty (not an error); the existing single-frame `{function, file, line}` fields are unchanged.
- [ ] `--schema` gate passes; the error model change is reflected.
- [ ] e2e (real Godot 4.6+): a deliberately-thrown runtime error reports a multi-frame callstack.

## Blocked by

- ADR-0025 (PR #279) — establishes read-only observability as in scope. Independent of the `gda logger` slices (A/B).

PR #767, opened 2026-08-28
feat(gda): recognize a project-raised push_error as a script diagnostic (#722)

### LINKED ISSUE #722: script errors: recognize a project-raised push_error as a structured diagnostic (#651 parser widening)
## Parent

Follow-up from #664 (delivered by PR #720); the parser is owned by #651. The underlying dogfooding note is GDA-DF-030.

## What to build

Recognize a project-raised `push_error()` as a structured script diagnostic in the #651 script-error parser.

Today a scene that reports an application invariant with `push_error` in `_ready` produces an engine `ERROR:` record whose only script attribution is its GDScript backtrace — outside #651's closed recognition set — so `scene preflight` reports `started: true` with empty `diagnostics` (reproduced with a real engine during #720's review). Assertions and genuine runtime errors ARE caught. `push_error` is the most common way a Godot project reports exactly the invariant violation GDA-DF-030 describes.

This is a deliberate contract decision, not a bug fix: widening the recognition set changes what
`script run` reports too, since the parser has one owner and the launch-backed channels share it.

> **Correction (2026-08-28, during implementation): `script validate` is NOT one of them.** It has
> its own parser — `SCRIPT ERROR:`-prefixed lines plus a `GDScript::reload (…:<line>)` frame, at
> `src/gda/commands/script.py:1779-1781` — entirely separate from `gda.script_errors`. A
> `push_error` record is an `ERROR:` line and never reaches it. The channels this widening actually
> moves are `scene preflight` (the point of the issue) and `script run`'s success-path
> `diagnostics`. Verified in the implementing PR, not assumed. The decision needs: whether a `push_error` record becomes a new `ScriptErrorKind` or maps onto an existing one, how its backtrace-only attribution is represented, and whether any current consumer depends on these records being absent.

## Verified preconditions (2026-08-27 review)

Premise reproduced on a real Godot 4.6.3 headless run — a `push_error()` reachable from
`_ready` prints:

```
ERROR: probe: invariant violated
   at: push_error (core/variant/variant_utility.cpp:1024)
   GDScript backtrace (most recent call first):
       [0] _inner (res://main.gd:7)
```

Three implementation facts the body did not carry:

1. **Key recognition on the `at:` frame function, which is literally `push_error`**
   (`core/variant/variant_utility.cpp:1017`). Keying on "any `ERROR:` that has a backtrace"
   would over-match: `ScriptServer::capture_script_backtraces`
   (`core/object/script_language.cpp:565`) attaches a backtrace to ANY error raised while
   GDScript is on the call stack, including engine-side `ERR_PRINT`s a script triggered
   indirectly (a bad `get_node()`, say).
2. **`src/gda/engine_log.py:45` carries a comment that is wrong for the build gda runs** —
   "push_error / warnings carry no backtrace". `godot --headless` is a `target=editor`
   build, which compiles `DEBUG_ENABLED` unconditionally, and `gdscript.cpp:2871-2875`
   sets `track_call_stack = true` under `DEBUG_ENABLED` regardless of the project setting.
   Correct that comment as part of this slice.
3. **`_classify()` discards the record on the `ERROR:` branch** — `_engine_error(message)`
   receives only the message string, not `function`/`callstack` (`script_errors.py:348-357`).
   Threading the record through is a small precondition of point 1, not a blocker.

**Doc drift this slice creates:** `docs/command-catalog.md:135` currently asserts as settled
fact that "`push_error()` is not among them". That sentence becomes false the moment this
ships, and it lives in a different file/paragraph than the #716/#717 docs slice — so the
correction must ride in THIS PR, not be left for that one to notice.

## Acceptance criteria

- [ ] A `push_error` raised during `scene preflight` appears as a structured diagnostic with script attribution from its backtrace.
- [ ] The widened recognition is decided and recorded once, in the parser's own contract (#651 lineage), with the `script run` impact stated.
- [ ] Existing recognized kinds (assert, runtime error) are unchanged, pinned by the existing tests.
- [ ] e2e on a real engine: a `push_error`-reporting scene yields the diagnostic through `scene preflight`.
- [ ] `docs/command-catalog.md:135` and the stale `engine_log.py:45` comment are corrected in the same PR.

## Blocked by

None - can start immediately (needs a triage decision on the contract change before an agent picks it up).

qinhaihong-red:
## What

Recognize a project-raised `push_error()` as a structured script diagnostic in the #651 parser.

Before this, a scene reporting its own invariant violation from `_ready` came back from
`scene preflight` as `started: true` with **empty** diagnostics — the GDA-DF-030 phantom-clean
start, in the shape a Godot project most often writes it. Reproduced on Godot 4.6.3 against the
merge base, and pinned by the new e2e (its red-proof below is literally `assert True is False`).

Closes #722

## The contract decision

**A new `ScriptErrorKind`, `push_error`** — not a mapping onto `runtime_error` or `assert`.

- The two are **different claims about the program**. A `runtime_error` is the engine saying it
  could not perform an operation; a `push_error` record is the PROJECT saying it does not like
  what it found. Folding them together makes `runtime_error` ambiguous for every current
  consumer, and the distinction is unrecoverable afterwards.
- **Additive**: every existing recognized record is byte-identical, pinned by the existing tests.
  No verdict, error code, or exit status changes anywhere.

**Recognition keys on the `at:` frame function**, which the engine fixes as `push_error`
(`VariantUtilityFunctions::push_error` → `ERR_PRINT`, `core/variant/variant_utility.cpp:1017`) —
never on the message, which is entirely the project's own prose.

Explicitly **not** "any `ERROR:` that carries a backtrace":
`ScriptServer::capture_script_backtraces` attaches one to ANY error raised while GDScript is on
the stack. A script's bad `get_node()` prints `ERROR: Node not found: … / at: get_node
(scene/main/node.cpp:1961)` with a full backtrace, and that is the ENGINE's failure, not the
project's report. Both shapes are pinned by verbatim Godot 4.6.3 captures, and the wrong key is
mutation-proofed red at both the unit and the real-engine level.

A project method named `push_error` cannot reach the branch either: an error raised from GDScript
goes through `ERR_HANDLER_SCRIPT` (`gdscript_vm.cpp:529`/`3988`) and prints as `SCRIPT ERROR`,
which `_classify` has already routed elsewhere. (The builtin is not even shadowable from GDScript
— verified on 4.6.3 — but the level split is the load-bearing reason.)

**Backtrace-only attribution, reported as such.** `path`/`line` come from the most recent
`res://` frame of the engine's backtrace — the call site the engine itself named — canonicalized
like every other address. No backtrace, or an all-engine one → both `null`. Nothing is
synthesized; the `at:` line's C++ location never leaks out.

**Two placements this kind does NOT get**, both recorded where they live:

- **Not in `_ENTRY_FAILURE_PRECEDENCE`** — it proves the script RAN. The lockstep test
  `test_every_never_ran_kind_is_an_entry_failure_candidate` is updated deliberately (it fails by
  design otherwise), and now excludes three kinds with a stated reason each.
- **Does not arm `script run`'s completion-marker abort** (`_entry_attributable`). The watch's
  premise is that something interrupted the run: a runtime error abandons the function it was
  raised in, which is what makes a following silence evidence. A `push_error` interrupts nothing
  — execution continues at the next statement — so a script that reports an invariant and then
  computes quietly is alive **by construction**, and aborting it would break the projects that
  use `push_error` as ordinary logging.

The module docstring now also records **the rule the closed set is widened by**, once, since
widening it changes what every launch-backed channel reports.

## Cross-channel impact

| Channel | Impact |
| --- | --- |
| `gda scene preflight` | The point of the change: `diagnostics` gains the entry, so `started` is `false` for a scene that reports its own invariant violation. |
| `gda script run` | Success-path `diagnostics` gains entries (and so does the timeout/abort prose via `_render_captured_errors`). **A project that uses `push_error` as ordinary logging will now see advisory diagnostics on runs that still succeed — that is intended**, the field is advisory. The verdict, `_ENTRY_FAILURE_CODES` mapping, exit status and abort behaviour are all unchanged. |
| `gda script validate` | **Unaffected** — verified, not assumed. It has its OWN parser (`_SCRIPT_ERROR_LINE` + the `GDScript::reload` frame, `commands/script.py`), which only reads `SCRIPT ERROR:` records; a `push_error` is an `ERROR:`. The issue body's premise that it shares the parser is not accurate. Its e2e passes unchanged (20 passed). |
| `gda diag errors` / `gda logger tail` (daemon) | Unaffected — they read `engine_log.parse_errors` directly, never `parse_script_errors`. |

**No current consumer depended on these records being absent**: the full non-e2e suite (1949
tests) passes untouched, and the daemon/harness `gda_log` path does not reach this parser at all.

## Doc drift corrected in the same PR

- `docs/command-catalog.md` — "`push_error()` is not among them" was false the moment this ships.
  The preflight paragraph now states the widened set, the frame-not-message key, and why a
  backtrace alone does not qualify a record; the `script run` paragraph states the advisory
  diagnostics and the abort carve-out.
- **`ScenePreflightResult.diagnostics`' own published schema description** carried the same
  assertion — corrected (this one is `--schema` output, not a comment).
- `engine_log.py`'s "push_error / warnings carry no backtrace" — wrong for the build gda runs.
  `godot --headless` is a `target=editor` build, so `DEBUG_ENABLED` is compiled in and
  `gdscript.cpp` forces `track_call_stack = true` regardless of
  `debug/settings/gdscript/always_track_call_stacks`. Corrected there, in `diag.py`'s two
  published descriptions carrying the same claim, and in the two test comments that repeated it.
- **"frame `[0]` equals the `at:` location" — false, and found by the round-2 review.** The
  clause retained beside the comment above states it unconditionally in four places, one of them
  published to agents through `--schema`. #722's own captured record disproves it for exactly the
  class this slice started recognizing: a `push_error`'s `at:` is `push_error
  (core/variant/variant_utility.cpp:1024)` while frame `[0]` is the `.gd` line that called it.
  `at:` is where the error was RAISED (the raising C++ site's `__FUNCTION__`/`__FILE__`/
  `__LINE__`); the backtrace is where the SCRIPT was. They coincide only when GDScript itself
  raised the error — the runtime-error class #283 measured the claim on. Corrected in the
  `engine_log` module docstring, the `FRAME_LINE` contract comment, `parse_errors`'s docstring,
  and the `DiagError.callstack` schema description, and pinned by two new regressions built from
  one real Godot 4.6.3 capture holding both shapes.

## Verification

Gates, verbatim:

```
1951 passed, 545 deselected in 55.04s                # uv run pytest -q -m "not e2e"
All checks passed!                                   # uv run ruff check .
473 files already formatted                          # uv run ruff format --check .
0 errors, 0 warnings, 0 informations                 # uv run pyright
```

Real-engine e2e (Godot 4.6.3, macOS), targeted:

```
59 passed in 59.72s                                  # test_e2e_scene_preflight.py + test_e2e_script_run.py
14 passed in 139.20s (0:02:19)                       # tests/test_e2e_scene_validate.py
20 passed, 48 deselected in 253.50s (0:04:13)        # tests/test_e2e_script.py -k validate
```

(The last two are from the round-1 head; nothing since then touches `scene validate` or
`script validate`.)

### Red-proof (every new guard, from committed state)

Source reverted to the merge base `b616e7fb`, tests kept:

- 6 of 9 new unit guards fail (`AttributeError: ... has no attribute 'PUSH_ERROR'`,
  `IndexError: list index out of range`).
- The new e2e fails with `assert True is False` on `started` — the defect itself, reproduced.

Three guards cannot go red against the base (the record is simply dropped there), so they are
mutation-proofed against the plausible wrong implementations instead:

- key on backtrace presence instead of the `at:` frame → the over-match guard fails at both
  levels, reporting the engine's `Node not found` as `kind: push_error`;
- add `PUSH_ERROR` to `_entry_attributable` → the abort guard fails
  (`[False, True, True, True] == [False, False, False, False]`).

The two round-2 regressions are red-proofed the same way, by injecting the discredited
assumption:

- synthesize frame `[0]` from the `at:` line in `parse_errors` → the parser guard fails
  (`{'function': 'push_error'} != {'function': '_inner'}`) and **nothing else does** — all 16
  other parser tests still pass, so no pre-existing test covered this;
- cross-fill `DiagError.callstack[0]` from the top fields → the model guard fails
  (`'core/variant/variant_utility.cpp' == 'res://main.gd'`), again as the sole failure.

## Deliberately out of scope

The second widening candidate recorded on #722 (`Resource file not found:` →
`script_compile_failed` where `script_not_found` is accurate) is **not** taken here. Admitting
that sentence does not by itself fix it: it is the resource layer's generic message for ANY
resource, so it classifies as `RESOURCE_LOAD_FAILED`, whose `_ENTRY_FAILURE_CODES` row already
maps to `script_compile_failed`. Getting `script_not_found` means deciding whether a generic
resource sentence may carry a script-specific verdict — a change to an error CODE on a failure
path, not an advisory diagnostic on a success path, with a different blast radius and no AC or
e2e here. The widening RULE this PR records is what a follow-up should apply to it.

## Notes for integration

**This slice carries only its own catalog contract.** An earlier revision placed #721's and
#752's `docs/command-catalog.md` paragraphs here, to keep one owner per hotspot file for the
wave. The orchestrator reversed that decision after the round-2 review: a slice must not publish
a capability it does not implement, and both siblings are now changing under their own review
rounds, so the routed text was a stale cross-branch read. Commit `4af472ea` removes it — #721
(PR #769) and #752 (PR #770) each take their own catalog contract back, and neither needs to
rebase onto this branch.

What remains in `docs/command-catalog.md` here is #722's own contract, in two paragraphs the
issue and its triage decision name explicitly: the `scene preflight` recognition set (correcting
the now-false "`push_error()` is not among them"), and the `script run` completion-marker
paragraph stating this widening's cross-channel impact.

### DIFF
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -134,9 +134,16 @@ they answer *different* ones:
   engine's error stream. Read `started`: true only when the scene reached `_ready` AND nothing was
   recognized on stderr, which is the distinction the dogfooding note asks for (GDA-DF-030 —
   static validation passed while the first live launch rejected every assembly). Recognition is
-  #651's closed set of engine failure sentences (a runtime error, a failed assertion, a script
-  that could not load, a script binding the engine refused), so project prose written with
-  `push_error()` is not among them.
+  #651's closed set: the engine's own failure sentences (a runtime error, a failed assertion, a
+  script that could not load, a script binding the engine refused) **and a project-raised
+  `push_error()`** (#722), which is the most common way a Godot project reports exactly the
+  invariant violation GDA-DF-030 describes. That one is recognized by its `at:` frame — which
+  the engine fixes as `push_error` — never by its message, which is the project's own prose; its
+  `kind` is `push_error` and its `path`/`line` are the call site named in the engine's GDScript
+  backtrace, or null when it attached none. Everything else the engine prints stays
+  unrecognized: a backtrace alone does not qualify a record, since the engine attaches one to
+  any error raised while GDScript is on the stack, including engine-side failures a script only
+  triggered indirectly.
 
 `scene validate` takes a `.tscn` specifically, and refuses a binary `.scn` with `invalid_path`:
 its dependency set comes from the scene's own TEXT (which is also what attributes each dependency
@@ -626,7 +633,11 @@ alive and did not finish) — so a slow suite is distinguishable from a hang.
 `--completion-marker <line>` declares a liveness contract — the script prints that line when
 its work is done — and a run that hit a recognized error attributable to the entry script, has
 not printed the marker, and then goes silent on both streams is ended in seconds and reported
-as `script_aborted` with the captured error and phase `aborted_on_error`. The script executes in full, within
+as `script_aborted` with the captured error and phase `aborted_on_error`. A `push_error` never
+arms that abort even though it is recognized (#722): it interrupts nothing — execution
+continues at the next statement — so a script that reports an invariant and then computes
+quietly is alive by construction. It does appear in the run's `diagnostics`, which are advisory:
+a project that uses `push_error` as ordinary logging sees entries on runs that still succeed. The script executes in full, within
 the trusted-project assumption (ADR-0009).
 
 ### `project`
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -66,7 +66,7 @@ they are provenance, not status markers.
 | `gda scene get` | Read a scene's structured tree from its file on disk |
 | `gda scene list` | Enumerate scenes in the project |
 | `gda scene get-exports` | List `@export` properties declared by a scene's nodes |
-| `gda scene validate` | Check statically that a scene and the sub-scenes it instances resolve their dependencies and compile their scripts |
+| `gda scene validate` | Check statically that a scene's dependencies resolve and its scripts compile |
 | `gda scene preflight` | Boot a scene headless and report its startup verdict |
 
 **Enumeration** (established by #54): `scene list` walks the project's `res://`
@@ -115,46 +115,18 @@ they answer *different* ones:
   it binds — the referenced `.gd` files and the embedded `[sub_resource type="GDScript"]`s the
   dependency walk never sees — and checks each script's native base against the node carrying it,
   reporting one problem per file: `missing_resource` (the file is gone), `unloadable_resource`
-  (present on disk but no `ResourceLoader` opens it — typically an asset that was never
-  imported, which a non-editor engine cannot load at all), `script_compile_failed` (the same
-  verdict `script validate` gives that file; an embedded script is named by its `::id`
-  sub-resource path), `incompatible_script` (a binding the engine would refuse — a script on a
-  node outside its native base, the same rule `node script attach` enforces asked
-  statically, or a value bound to a `script` slot that is not a script at all),
-  `cyclic_instance`, or `instance_depth_exceeded` (the last two defined below with the composed
-  walk that finds them). Each problem carries the declared `type` and the node paths referencing
-  it, read from the file's own text because the engine drops an unresolvable reference from what
-  it loads. A path declared twice is one problem with both nodes listed. The verdict is
-  **staged**: unresolved dependencies suppress the load, so the compile and binding problems
-  only the loaded scene can reveal appear after the dependencies are repaired and validate is
-  rerun — the problem list is complete for the stage it reached, not across both stages at once.
-
-  The verdict is **composed**: the scenes a scene instances are validated with it, because a
-  parent whose instanced child is broken is broken too and the parent's own walk cannot see that
-  — `res://child.tscn` resolves, opens, and hands back a usable `PackedScene` with whatever is
-  missing inside it substituted by null. Every problem therefore carries `scene`, the file it
-  was found in, and its `path` and `nodes` are read against THAT file, not against the scene the
-  command was given. The walk descends into instanced `.tscn` files only (a binary `.scn`
-  carries no dependency text — the same reason the command refuses one at the top level),
-  validates each file ONCE however many sites instance it, and reports an instancing cycle
-  rather than following it.
-
-  `cyclic_instance` — a scene instances one that already instances it. Godot refuses a cyclic
-  scene inclusion, so the composition cannot load. Reported against the file that declares the
-  closing edge (`scene`), with `path` naming the ancestor being re-instanced and `nodes` the
-  instancing sites. The walk stops at that edge, so whatever lies beyond it is unreported until
-  the cycle is broken.
-
-  `instance_depth_exceeded` — not a defect in the project but a limit of gda: the walk follows
-  at most 16 levels of instanced sub-scenes below the validated scene, and this edge is past
-  that bound, so the scene named by `path` and everything it instances were UNCHECKED, not
-  judged sound. `scene` names the file holding the declining edge. The verdict is still
-  `valid: false` — a gate must not answer "sound" about what it did not look at — and
-  validating the named scene directly gives it a verdict of its own. The bound covers gda's own
-  walk only: the engine loads the whole chain regardless, and an extremely deep chain overflows
-  the engine's own loader and ends the run with no verdict at all, which this bound does not
-  change.
-
+  (present on disk but no `ResourceLoader` opens it — typically an asset that was never imported,
+  which a non-editor engine cannot load at all), `script_compile_failed` (the same verdict
+  `script validate` gives that file; an embedded script is named by its `::id` sub-resource path)
+  or `incompatible_script` (a binding the engine would refuse — a script on a node outside its
+  native base, the same rule `node script attach` enforces asked statically, or a value bound to
+  a `script` slot that is not a script at all). Each problem carries the declared `type` and
+  the node paths referencing it, read from the file's own text because the engine drops an
+  unresolvable reference from what it loads. A path declared twice is one problem with both nodes
+  listed. The verdict is **staged**: unresolved dependencies suppress the load, so the compile
+  and binding problems only the loaded scene can reveal appear after the dependencies are
+  repaired and validate is rerun — the problem list is complete for the stage it reached, not
+  across both stages at once.
 - `gda scene preflight PATH` is **dynamic**. It instantiates the scene, adds it under a one-shot
   engine's tree root — which runs its `_ready` and the project's autoloads — keeps it alive for
   `--frames` idle frames so startup work landing after `_ready` still prints, and reports
@@ -665,8 +637,8 @@ as `script_aborted` with the captured error and phase `aborted_on_error`. A `pus
 arms that abort even though it is recognized (#722): it interrupts nothing — execution
 continues at the next statement — so a script that reports an invariant and then computes
 quietly is alive by construction. It does appear in the run's `diagnostics`, which are advisory:
-a project that uses `push_error` as ordinary logging sees entries on runs that still succeed. The script executes in full, within
-the trusted-project assumption (ADR-0009).
+a project that uses `push_error` as ordinary logging sees entries on runs that still succeed.
+The script executes in full, within the trusted-project assumption (ADR-0009).
 
 ### `project`
 
@@ -919,7 +891,7 @@ headless is unaffected (4.4+, cross-platform).
   a Dictionary into an `Object` parameter, null into `int`, and any JSON array into a
   typed `Array[int]` are refused with the reason. Without the check `callv` pushes an
   engine error, returns null and writes to the Session log — a failure that would read
-  as a successful null. Three reproduced unsafe argument classes are refused earlier
+  as a successful null. Two reproduced unsafe argument classes are refused earlier
   still, in the params model both invocation paths share (recursively, so a nested
   value counts):
   non-finite numbers (`NaN`/`Infinity`, which JSON has no literals for but Python's
@@ -929,17 +901,11 @@ headless is unaffected (4.4+, cross-platform).
   and a larger integer can arrive as a DIFFERENT value (a call that then succeeds on
   something the caller never sent). Finite floats already are binary64 and do not
   inherit that integer bound; real-engine tests pin the reproduced high-range values
-  `1e17`, `2.5e17`, and `1e300` unchanged.
-  The third class is the small-float one #752 decided: Godot's `built_in_strtod` applies a power
-  of ten it computes as a double, so a literal whose applied exponent is −309 or below is
-  divided by `inf` and arrives as `0.0` — `DBL_MIN`, every subnormal, and normal values such as
-  `1.2345678901234567e-300`. No decimal spelling avoids it (the significand is an integer and
-  the largest finite power the engine's table builds is `1e308`), so those values are REFUSED
-  rather than delivered as a zero the caller never sent; a real-engine corpus
-  (`tests/live_number_corpus.py`) pins the boundary with two same-magnitude values whose only
-  difference is significant-digit count. What crosses can still arrive 1–2 ULP away — disclosed,
-  not refused, since refusing it would reject ordinary game values.
-  Standard JSON Schema cannot distinguish
+  `1e17`, `2.5e17`, and `1e300` unchanged. This is not a full-range preservation
+  guarantee: Godot 4.6.3 parses some small-magnitude normal values, including
+  `1.2345678901234567e-300` and `DBL_MIN`, as `0.0`, and its `JSON.stringify` can
+  also lose small live-result values. [Issue #752](https://github.com/aigengame/godot-agent/issues/752)
+  owns that cross-operation transport defect. Standard JSON Schema cannot distinguish
   an exponent-form float from the equal mathematical integer, so its recursive number
   branch stays broad and discloses that the params model enforces the integer-token
   bound at execution. RFC JSON excludes `NaN` and `Infinity`; some in-memory schema
@@ -953,14 +919,6 @@ headless is unaffected (4.4+, cross-platform).
   chain has at most one declaration owner (a base owner covers its subclasses and need
   not define every method it names); a project that declares in both fails to parse with a message naming the
   member — loud, never a silently wrong allowlist.
-
-  Every live reply is framed with Godot's full-precision JSON writer (#752), so a float a live
-  read returns — `game get`, a `game call` return value, `perf` monitors, a `perf monitor`
-  timeline — reaches the caller as the exact binary64 the running game holds. The default writer
-  it replaced rendered floats fixed-point at 32 decimals, flattening every value below about
-  `1e-32.6` to `0.0` and rounding ordinary values to ~15 significant digits. One residual: a
-  negative zero reads back as `0.0`, decided by the engine before the writer's precision
-  argument applies.
 - **`input` (input simulation):** runtime input injection into the running game
   (shipped, #221). Single-frame ops `input key <KEY> [--modifiers …] [--released]`,
   `input mouse-move <x> <y>`, and `input action <NAME> [--release] [--strength F]`
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -115,19 +115,19 @@ they answer *different* ones:
   it binds — the referenced `.gd` files and the embedded `[sub_resource type="GDScript"]`s the
   dependency walk never sees — and checks each script's native base against the node carrying it,
   reporting one problem per file: `missing_resource` (the file is gone), `unloadable_resource`
-  (present on disk but no `ResourceLoader` opens it — typically an asset that was never imported,
-  which a non-editor engine cannot load at all), `script_compile_failed` (the same verdict
-  `script validate` gives that file; an embedded script is named by its `::id` sub-resource path),
-  `incompatible_script` (a binding the engine would refuse — a script on a node outside its
-  native base, the same rule `node script attach` enforces asked statically, or a value bound to
-  a `script` slot that is not a script at all), or `cyclic_instance` (defined below with the
-  composed walk that finds it). Each problem carries the declared `type` and
-  the node paths referencing it, read from the file's own text because the engine drops an
-  unresolvable reference from what it loads. A path declared twice is one problem with both nodes
-  listed. The verdict is **staged**: unresolved dependencies suppress the load, so the compile
-  and binding problems only the loaded scene can reveal appear after the dependencies are
-  repaired and validate is rerun — the problem list is complete for the stage it reached, not
-  across both stages at once.
+  (present on disk but no `ResourceLoader` opens it — typically an asset that was never
+  imported, which a non-editor engine cannot load at all), `script_compile_failed` (the same
+  verdict `script validate` gives that file; an embedded script is named by its `::id`
+  sub-resource path), `incompatible_script` (a binding the engine would refuse — a script on a
+  node outside its native base, the same rule `node script attach` enforces asked
+  statically, or a value bound to a `script` slot that is not a script at all),
+  `cyclic_instance`, or `instance_depth_exceeded` (the last two defined below with the composed
+  walk that finds them). Each problem carries the declared `type` and the node paths referencing
+  it, read from the file's own text because the engine drops an unresolvable reference from what
+  it loads. A path declared twice is one problem with both nodes listed. The verdict is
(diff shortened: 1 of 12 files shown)


qinhaihong-red:
Reviewed exact head `cc6fd7c396730be61ac5bf9c2c1d6aabb478c31c` against merge base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`.

## Artifact / Markdown

**Changes required (2 findings).**

1. `docs/command-catalog.md:69,117-156` publishes #721's composed-scene contract before this exact head implements it: it promises recursive sub-scene validation, a required `scene` field, and the `cyclic_instance` / `instance_depth_exceeded` kinds. At this head, `SceneProblem` has no `scene` field, `SceneProblemKind` has only the original four kinds, and no composed walk exists. This makes the catalog unusable as a description of the current command. Keep the #722-owned catalog change here, but move/defer this block until the #721 implementation and its schema/tests land with it.

2. `docs/command-catalog.md:922-963` similarly publishes #752 before it exists at this head: it says small floats are refused, cites an absent `tests/live_number_corpus.py`, and promises full-precision live replies while the harness still uses default `JSON.stringify`. The routed text has also drifted from #770's current evidence: line 940 says residual request drift is “1–2 ULP”, while #770's current corpus records 31- and 105-double gaps in the mantissa-cap band and explicitly rejects the “1–2 ULP” description. A caller following this catalog can still silently deliver `0.0` or rely on precision the current binary does not provide. Defer the block to the implementation-owning change and use #770's current measured wording.

**Terminology and prose assessment.** I used `CONTEXT.md`, ADR-0002, ADR-0031, ADR-0040, issue #722 (body and all comments), and the current #769/#770 artifacts as terminology/authority sources, with ASD-STE100 Simplified Technical English as the writing reference. I inspected the complete PR body, complete issue, the full changed Markdown at this head, every changed requirement/decision/procedure passage, and the causal, scope, ordering, and term relationships. Finding 2 is the substantive accuracy/prose failure; I found no other substantive terminology or prose issue. I did not run a general external-link checker.

## Architecture / DDMA

**Changes required (1 finding).**

The #722 code topology itself is sound: `engine_log.parse_errors` owns engine-log framing, `script_errors.parse_script_errors` owns the closed semantic recognition set and public kind, and the `scene` / `script` command modules consume that authority downward. `_entry_attributable` correctly remains the script-run-specific abort policy, and the new kind does not enter entry-failure precedence or arm the completion-marker watch. The resulting source graph is directed and the new knowledge has one code owner.

The sibling catalog routing, however, creates an unencoded horizontal integration dependency. This head owns #721/#752's public catalog claims without owning either capability; PRs #769 and #770 still both target the same `b616e7fb...` base rather than carrying an implementation dependency on this branch. The branch is therefore structurally incomplete when merged on its own: removing either sibling leaves a claimed capability with no implementation owner. Keep only #722's catalog delta here, and let each implementing slice co-land its own catalog contract after serial integration (or place code and catalog together in a dedicated integration change).

Load-bearing claim states are `open` because the project defines no stronger claim-evidence vocabulary: `DDMA-767-1` (single parser owner) is supported by import tracing and exact-head tests and would be disproved by another launch-backed semantic parser; `DDMA-767-2` (additive run semantics) is supported by precedence/abort tracing and the real-engine probe and would be disproved by a changed verdict/abort; `DDMA-767-3` (sibling routing is complete) is disproved at this head by the missing owners above. Completeness, orthogonality, DRY, and extensibility otherwise pass for the #722 implementation.

## Standards

**Changes required (2 hard documented-standard violations; 0 baseline smells).**

1. `docs/command-catalog.md:69,117-156` says “The verdict is **composed**,” claims every problem carries `scene`, and publishes `cyclic_instance` / `instance_depth_exceeded`, but none exists in the exact-head code or schema. This violates `RULES.md:112,119,121-123`, which requires affected artifacts to stay synchronized and repeated facts to remain constrained by one authority. Remove/defer this hunk or include the #721 implementation before documenting it.

2. `docs/command-catalog.md:922-963` says small floats “are REFUSED,” cites `tests/live_number_corpus.py`, and promises that every live reply uses the full-precision writer. The cited test is absent, the exact-head params model rejects only non-finite floats and oversized integers, and the harness uses default `JSON.stringify`. This independently violates the same synchronization/authority rules. Remove/defer these claims or co-land #752's implementation and evidence.

No baseline smell is actionable: the #722 implementation keeps framing in `engine_log.py`, semantic classification in `script_errors.py`, and channel policy in the command owners; its multi-file footprint is the repository-required vertical-slice contract/test/documentation synchronization, not Shotgun Surgery.

## Spec

**Changes required (3 findings).**

1. **Partial/wrong required documentation fix.** Issue #722 says, “Correct that comment as part of this slice.” `src/gda/engine_log.py:18-21,86-98` still says backtrace frame `[0]` equals the `at:` location, and `src/gda/commands/diag.py:100-105` publishes the same claim in `--schema`. The issue's own captured `push_error` record disproves it: `at:` is `push_error (core/...:1024)`, while frame `[0]` is `_inner (res://main.gd:7)`. Correct both contracts and add a `push_error` framing/model regression that proves the locations differ.

2. **Scope creep publishes unimplemented #721 behavior.** The issue asks to “Recognize a project-raised `push_error()` as a structured script diagnostic in the #651 script-error parser.” The catalog additionally promises composed scene validation, a `scene` field, and two new problem kinds, while the exact-head implementation has none of them. Remove/defer those paragraphs or land them only with #721's implementation.

3. **Scope creep publishes unimplemented #752 behavior.** The same quoted issue scope does not include a live-number transport change. The catalog says small floats are refused and live replies use full-precision JSON, while the exact-head params model/harness still implement the old behavior. Remove/defer those paragraphs until #752's implementation lands.

The remaining #722 acceptance criteria are implemented: the distinct additive kind, backtrace-only attribution, unchanged existing kinds, `scene preflight` real-engine coverage, `script run` diagnostics without abort/precedence changes, and the deliberate exclusion of the Resource-file candidate.

**Validation performed.** Confirmed the merge base and non-empty 11-file diff; inspected full exact-head files and adjacent consumers; read issue #722 and current sibling artifacts; rendered `scene preflight`, `script run`, and `diag errors` schemas; ran `git diff --check`; ran 247 relevant unit tests (`247 passed`); ran the two new Godot 4.6.3 preflight e2e tests (`2 passed`); and ran real-engine `script run` probes showing a structured `push_error` diagnostic and a 4-second silent post-`push_error` run that was not aborted by a declared completion marker.

**CI and skips.** The exact-head CI run completed successfully for lint/format, pyright, unit tests/package build, co-install, and required aggregation. The GitHub Godot e2e job was skipped; the targeted local real-engine checks above supply narrower evidence, not a replacement for the full e2e suite.

**Not performed.** I did not rerun the full non-e2e suite, all four PR-body e2e groups, rendered/manual UI checks, or other supported Godot versions. Green CI is not treated as proof of the documented runtime contracts.

Review-axis summary: Standards — 2 findings, worst issue is publishing unimplemented public catalog contracts; Spec — 3 findings, worst issues are the false backtrace-location contract and the two out-of-scope prepublished sibling contracts.

qinhaihong-red:
## Round-2 review handled — head `5709c2f3`

All four required findings reproduced against the reviewed head `71d50f47` before anything was changed. Two are fully adopted; the P2's *classification* half is adopted with a narrower solution than one of its two options, and the option not taken is argued below rather than skipped.

| Claim | Decision | Evidence on `71d50f47` | Change | Validation |
| **Spec P1 / DDMA P1** — repeat `daemon start` leaks `PermissionError` and leaves residue | **Fully Adopt** | Reproduced twice. (a) `daemon_pid=999`, no harness, project root `0o555` → `PermissionError: [Errno 13] ... 'proj/addons'`, raw. (b) `daemon_pid=999`, stale harness, `project.godot` `0o444` → raw `PermissionError` **and the stale body overwritten** | One transactional helper `_install_harness_transactionally` (capture → install → classify → roll back → re-raise); all three mutating call sites routed through it: start-with-no-daemon, the #225 self-sync, `daemon install` (`5709c2f3`, `src/gda/commands/daemon.py`) | 3 new regressions, red-proofed by routing the arm back to a bare `install_harness`. Byte identity asserted: the stale harness comes back verbatim and `installed_harness_version` is still `stale-old` |
| **Standards 3 (judgement call)** — duplicated install orchestration | **Fully Adopt** | The snapshot guard, install guard, conversion, rollback and re-raise appeared at `803-815` and `910-922` | Subsumed by the extraction above — with a third call site it stopped being a taste question | `ruff`, `pyright`, full non-e2e suite |
| **Standards 1 / Spec P2 / DDMA P2** — `except OSError` is wider than the code's pinned meaning | **Partially Adopt** | Reproduced: injected `ENOSPC` → `harness_install_permission_denied`; a real `addons`-is-a-file → `NotADirectoryError` → same code | Classification keyed on `errno` via `_is_filesystem_refusal`: `EACCES`, `EPERM`, `EROFS`. An `OSError` with no `errno` answers `False` | 5 negative-boundary regressions (`ENOSPC` on start and install, `ENOSPC` still rolling back, real `ENOTDIR`, capture-site `EIO`) + 1 pinning `EROFS` inside the set. All red-proofed by widening the predicate to `return True` / dropping `EROFS` |
| ↳ *sub-claim*: "map other filesystem failures separately" | **Do Not Adopt** | — | No second error code registered | Argued below; residual risk stated |
| **Artifact 1 / Standards 2** — ADR-0018 states the superseded contract | **Fully Adopt** | `0018-...md:118-129` still read "An exception still propagates to the caller unchanged" | Dated **Outcome (2026-08-29, #700)** note APPENDED; the #654 prose is left as the point-in-time record per this repo's ADR convention (cf. the 2026-06-23 note superseded by the #670 one) | The note distinguishes all four requested cases: typed refusal, exceptions that still propagate, rollback semantics, and the pre-snapshot no-write case — plus the widening to three install paths |
| **Artifact 2** — description asserts coverage the implementation lacks | **Fully Adopt** | Both halves of the claim reproduced (see rows 1 and 3) | Classifier and coverage corrected FIRST, then `src/gda/error_codes.py` and the ADR-0002 row rewritten to state exactly what it now does, including an explicit "a filesystem failure that is NOT a refusal … does not carry this code". PR body rewritten to match | `tests/test_error_registry.py` (#701's field-for-field drift pin) + `tests/test_skill_surface_sync.py` pass. `SKILL.md` needs no edit — its exit-127 row lists codes, not meanings |

### On the two judgement points

**`EROFS` is in the refusal set, deliberately.** It is the only member Python does not raise `PermissionError` for. #700's acceptance sentence is "a project whose tree **rejects writes**", and a read-only mount is that same fact with the same remediation — refused by the mount rather than the mode bits. Justification is written into the code next to the frozenset, into the registry comment, and into the ADR note, and pinned by `test_a_read_only_filesystem_is_reported_as_a_refusal`.

**No second code for non-refusal filesystem failures.** The finding offered "map other filesystem failures separately **or** preserve their existing behavior"; preserving is the half taken. Falling through restores the exact pre-#700 behaviour for `ENOSPC` / `EIO` / `ENOTDIR` / `FileNotFoundError` — the same `HarnessSnapshot` rollback, then propagate — so it is not a regression relative to this PR's base, and #700 scopes itself to "a filesystem-permission failure". A generic code would cost an `error_codes.py` row, an ADR-0002 row, registry and skill sync, and a permanent public promise, for a failure mode that has never been dogfooded here and that no caller has asked to branch on. **Residual risk, stated plainly:** an `ENOSPC` or `EIO` during the harness install still reaches the caller as a traceback. If that ever shows up in dogfooding it is a separate issue with its own code, not a speculative row added now.

### Behaviour changes worth flagging

- The repeat-start arm now takes a `HarnessSnapshot` before its self-sync. `capture()` reads `project.godot` plus the harness artifacts — a slightly wider read than the bare `install_harness` did (it also stats the `.gd.uid` sidecar), so a project where that sidecar is unreadable but everything else is fine now returns the typed refusal instead of succeeding. That is the same property the other two install paths already had; the alternative is leaving the arm un-transacted, which is the defect.
- The repeat-start arm's return type is unchanged (`DaemonStartResult | Failure` — it already returned `daemon_already_running`).

### Validation on `5709c2f3`

```
uv run python -m pytest tests/ -q -m "not e2e"   1954 passed, 543 deselected in 53.99s
uv run ruff check .                              All checks passed!
uv run ruff format --check .                     473 files already formatted
uv run pyright                                   0 errors, 0 warnings, 0 informations
```

`tests/test_daemon_lifecycle_payload.py` 73 passed; `tests/test_error_registry.py` + `tests/test_skill_surface_sync.py` 25 passed. Public CLI probe on a `0o555` project root returns the envelope at exit `127` in both JSON and human mode.

**Not verified:** no real read-only mounted filesystem was exercised (`EROFS` is injected, since a unit test cannot mount one), and Godot e2e was not run — no engine path is touched by this change. Round 1's dangling-symlink hypothesis was not revisited; it is rejected upstream as `project_not_found`.

qinhaihong-red:

Reviewed exact head `b573fc3ff9eb298257774ce2b16bba7bab203f88` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review and the author's reply.

The previous findings are closed: the #721/#752 catalog material has been removed from this branch, the stale ULP statement is gone, and the `at:` location is now correctly distinguished from backtrace frame `[0]` in both implementation prose and the public diagnostic schema.

### Artifact / Markdown — Pass

The current PR body, issue #722, command catalog, schema descriptions, and changed prose are consistent. The terminology and prose pass used `CONTEXT.md`, the applicable ADRs, and ASD-STE100; no substantive artifact residue remains.

Framing, semantic classification, and channel policy remain owned by `engine_log`, `script_errors`, and the command layer respectively. The sibling-PR documentation dependency has been removed.

- `tests/support.py:614` still says, without qualification, `# a warning has no backtrace`. The corrected contract and real Godot captures establish that `push_warning` raised while GDScript is on the stack can carry a GDScript backtrace. This adjacent fixture comment is now false and violates the repository's artifact-synchronization rule. Qualify it as this bare/synthetic fixture having no backtrace, or model a script-raised warning with its actual backtrace.

### Spec — Pass

The #722 behavior and scope are implemented without remaining Spec residue.

### Validation

- 338 related unit tests passed in an immutable exact-head snapshot.
- `gda diag errors --schema` renders the corrected location contract.
- `git diff --check` and all non-skipped required CI checks passed.
- Hosted Godot E2E remains skipped; this follow-up did not rerun the full non-E2E suite or current-head real-Godot E2E.

qinhaihong-red:

Head is now `6b78ae5c`.

### Standards — `tests/support.py`'s falsified fixture comment. **Adopted.**

You are right, and it is the same stale-absolute class as the `engine_log.py:45` comment issue #722 named: the shared engine-log fixture annotated its warning record `# a warning has no backtrace`, while this slice's own finding is that the engine attaches a GDScript backtrace to anything raised while GDScript is on the call stack — `push_warning` included.

Took the first of your two options: the comment now says that THIS bare synthetic record has none, and states the rule it is not an instance of. The fixture keeps its empty `callstack`, which is what a warning raised outside any script actually looks like — modelling a script-raised warning here would change a fixture several suites read for other reasons, for no gain over saying what it is.

`grep` confirms the stale claim appears nowhere else in `src/` or `tests/`.

The neighbouring comment at `tests/support.py:601-602` was left alone deliberately: it is explicitly scoped to "A runtime GDScript error", the one class where frame `[0]` and the `at:` location genuinely coincide, so it is not a second instance of the same defect.

### Validation

`1951 passed, 545 deselected` (non-e2e), ruff check / format clean, pyright 0 errors; 168 diag/engine-log/model tests green on the changed fixture. Comment-only — no production file is touched. `Closes #722` intact.

qinhaihong-red:
## Follow-up review — Pass

Reviewed exact head `6b78ae5c7b9407235f75d8f4d1b9a83164b61a04` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review, the author's reply, and the incremental change since `b573fc3f`.

The sole previous residue is closed. `tests/support.py:614-618` now limits the empty-callstack statement to this synthetic warning fixture, explains that it is raised without GDScript on the stack, and explicitly preserves the rule that script-called `push_warning` carries a backtrace.

- **Artifact / Markdown:** Pass — 0 findings.
- **Architecture / DDMA:** Pass — 0 findings.
- **Standards:** Pass — 0 findings.
- **Spec:** Pass — 0 findings.

Validation: 170 related tests passed; the public diagnostic schema still distinguishes the engine `at:` location from GDScript frame `[0]`; incremental/full `git diff --check` passed; all non-skipped CI checks succeeded. Hosted Godot E2E remains skipped, and I did not rerun the complete local non-E2E suite or other Godot versions.

**Current residue: none.**

PR #768, opened 2026-08-28
fix(gda): structured envelope for a filesystem-restricted harness install

### LINKED ISSUE #700: daemon start: a filesystem-restricted sandbox yields a raw PermissionError traceback, not an Error envelope
## Review-round source

Found during PR #694's review (wave 2); pre-existing, untouched by that PR. Under a filesystem-restricted sandbox — the literal GDA-DF-029 environment — `gda daemon start --windowed --project … --json` printed a raw Python traceback (`PermissionError: [Errno 1] Operation not permitted: 'proj/addons'`) instead of a structured envelope: the harness install's filesystem writes fail before any typed refusal path is reached. Automation gets an unparseable crash exactly where the restricted profile needs a typed answer.

## What to build

Classify a filesystem-permission failure during the daemon-start harness install into a structured Error envelope (an existing ENVIRONMENT code if one fits semantically, else a registered new one), with diagnostics naming the path the write failed on. The wave-2 snapshot transaction (#654's `HarnessSnapshot`) already guarantees restoration on failure — this issue is about the failure's SHAPE, not its cleanup.

## Acceptance criteria

- [ ] `daemon start` in a project whose tree rejects writes returns a structured envelope (registered code, diagnostics naming the failing path) — no traceback, JSON and human modes both.
- [ ] The snapshot restoration still runs on this path (porcelain-clean tracked fixture).
- [ ] A regression covers the restricted-write shape (chmod-based, permissions restored in
      teardown), reproducing the ORIGINAL mechanism — a failed write under `addons/` in
      `install.py`'s `_materialize`, not only the existing read-only `project.godot` fixture,
      which exercises the later config-write step.
- [ ] The existing tests that pin the CURRENT behaviour are updated deliberately: four
      `pytest.raises(PermissionError)` assertions in `tests/test_daemon_lifecycle_payload.py`
      hold "the exception still surfaces to the caller exactly as before" as today's contract.
      This slice changes that contract, so those assertions move to the structured-failure
      form in the same change — it is not an additive test.

## Blocked by

None - can start immediately. Related: GDA-DF-029 (dogfooding source), #654 (the restore transaction).

qinhaihong-red:
## Summary

- `gda daemon start` / `gda daemon install` write the harness under `res://addons` and rewrite `project.godot` before any engine exists (ADR-0018). Under a filesystem-restricted sandbox (GDA-DF-029) that write raised a raw `PermissionError` traceback instead of a structured `Error envelope` (#700).
- A filesystem REFUSAL during the install — `errno` `EACCES`, `EPERM` or `EROFS`, not every `OSError` (see round 2 below) — is classified into a new ENVIRONMENT-category code, `harness_install_permission_denied` (registered in `src/gda/error_codes.py` and `docs/adr/0002-headless-structured-output-contract.md`, exit `127`), naming the refused path and running the same `HarnessSnapshot` rollback the exception arm already performs. Every other exception from the install, and anything the spawn/readiness step raises, keeps its prior raw-raise behaviour unchanged (`test_start_rolls_back_when_the_spawn_itself_raises` and the two rollback-note tests are untouched).
- All THREE mutating paths run one transaction: `daemon start` with no daemon, `daemon start` against a running daemon (the #225 self-sync), and `daemon install` (see round 2 below).
- `docs/adr/0018-...md` carries a dated #700 outcome note recording the new failure shape.
- `src/gda/skill/SKILL.md`'s exit-127 row gained the new code (pinned against the registry by `test_skill_exit_127_row_names_every_registered_127_code`).

## Why the four existing pins moved

Per the pre-dispatch review, this is not an additive-test slice: four `pytest.raises(PermissionError)` assertions in `tests/test_daemon_lifecycle_payload.py` pinned "the exception still surfaces to the caller exactly as before" as the OLD contract. This change makes that false by design, so all four move to the structured-failure form in this same change:

- `test_failed_install_restores_a_project_whose_config_cannot_be_written` — now asserts `Failure` with `code == "harness_install_permission_denied"`, `"project.godot"` in `message`, `"rolled back"` + the harness filename in `diagnostics`.
- `test_failed_install_keeps_a_pre_existing_harness_untouched` — now asserts the same `Failure`/code, with the rest of the test's byte-identity assertions (stale body restored, pre-existing dirs kept) unchanged.
- `test_failed_install_leaves_a_tracked_project_porcelain_clean` — now asserts the `Failure`/code before checking `git status --porcelain` is still empty.
- `test_a_failed_install_restores_the_project` (the `daemon install` twin) — now asserts `Failure` with `code == "harness_install_permission_denied"` and the path in `message`.

## New regression: the ORIGINAL mechanism

Per review guidance, the existing `read_only_project_godot` fixture only exercises the LATER config-write step of `install_harness`, not the ORIGINAL #700 mechanism — a failed write under `addons/` in `install.py`'s `_materialize` (the reported repro was `PermissionError: ... 'proj/addons'`).

Added a new `unwritable_project_root` fixture (chmod `0o555` on the project root, permissions restored in teardown) that blocks `_materialize`'s `mkdir(parents=True)` for `res://addons` itself — the FIRST write `install_harness` performs, before `project.godot` is ever touched. Two new tests use it:

- `test_failed_start_reports_the_original_materialize_write_denial` (`daemon start`)
- `test_failed_install_reports_the_original_materialize_write_denial` (`daemon install`)

Both assert the structured envelope, that `"addons"` is named in `message`/`diagnostics`, and that nothing was created / `project.godot` stayed byte-identical (the mkdir failed before anything could be written).

## Registry ownership (P2 wave)

Minted a new code because no existing ENVIRONMENT code fits semantically — `binary_not_found` / `launch_timeout` / `user_data_unwritable` are all about the Godot launch/log path, not a Python-side pre-launch project access. Edited `src/gda/error_codes.py` and the ADR-0002 table together (the `#701` description-drift pin compares them field-for-field, citation-normalized). See "Recheck" below for why it's named `_permission_denied`, not `_unwritable`.

## Recheck: two P2s from an independent review, both reproduced live

A second review verified the slice with live filesystem repros rather than trusting the PR body, and found two gaps — both reproduced against a real `gda` invocation before fixing:

**1. The capture-site crash — the same defect #700 exists to fix, from a line the first pass never reached.**

```
$ chmod 0000 project.godot
$ gda daemon install --project <p> --json
PermissionError: [Errno 13] Permission denied: '.../project.godot'   (raw traceback)
```

`HarnessSnapshot.capture()` reads `project.godot` and the harness artifacts to build the pre-install rollback snapshot, and that call sat **before** the `try:` in both `run_daemon_start_operation` and `run_daemon_install_operation` — outside the guarded region entirely. A `chmod 0000 project.godot` (denying read, not merely write — `read_only_project_godot`'s `0o444` stays readable) still crashed with a raw traceback. Fixed by wrapping the `capture()` call in its own `except OSError`. Since capture failing means nothing has been written yet, this arm needs no `HarnessSnapshot` rollback — it reports directly.

**2. The mislabel this uncovers: a read failure reported as "could not write".**

`install_harness()` calls `_read_config(project_godot)` — a READ — inside the same guarded call that also performs two writes (`_materialize`, `_write_config`). The original wording ("the harness install could not **write** to the project filesystem") would mislabel a read denial reached through that path, sending whoever reads it to fix the wrong permission bit. This was unreachable before fixing (1) — `chmod 0000` always hit the unguarded `capture()` crash first — so fixing them together was necessary, not optional.

**Decision: one code, operation-neutral message.** Both fixes share one registered code, renamed from `harness_install_unwritable` to `harness_install_permission_denied` (updated across `error_codes.py`, the ADR-0002 table, and `SKILL.md`'s exit-127 row together). The `install_harness()`-site message is now deliberately neutral ("could not **access** the project filesystem") since that call site genuinely cannot tell read from write apart. The **new** capture-site failure, by contrast, knows for certain it's a read (capture only ever reads) and says so precisely — see `_harness_snapshot_unreadable_failure` vs `_harness_install_permission_denied_failure` in `src/gda/commands/daemon.py`.

**What stayed unguarded, deliberately.** `daemon uninstall` is untouched (it doesn't capture a snapshot and was never in scope). The spawn/readiness step inside `run_daemon_start_operation`'s second `try` still re-raises on any exception unchanged — a decision from the original slice, not collateral of this recheck (`test_start_rolls_back_when_the_spawn_itself_raises` still pins it).

**New regressions**, reproducing through the public operation, chmod-based, permissions restored in teardown:

- `test_failed_start_reports_an_unreadable_project_before_any_write` (`daemon start`)
- `test_failed_install_reports_an_unreadable_project_before_any_write` (`daemon install`)

Both use a new `unreadable_project_godot` fixture (`chmod 0000`, the reviewer's literal repro shape) and assert: `Failure` with `code == "harness_install_permission_denied"`, `"read"` in `message` and `"write"` **not** in `message`, `"nothing was written"` in `diagnostics`, and that nothing was materialized. Also added `"write" not in message` assertions to the two existing write-site tests that already check `message`, pinning that the neutral wording applies there too even though that site genuinely is a write failure.

## Gates (all run before writing any number here, on the final committed state)

Re-run on `5709c2f3` (the round-2 head):

```
$ uv run python -m pytest tests/ -q -m "not e2e"
1954 passed, 543 deselected in 53.99s

$ uv run ruff check .
All checks passed!

$ uv run ruff format --check .
473 files already formatted

$ uv run pyright
0 errors, 0 warnings, 0 informations
```

Plus a public CLI probe on a `0o555` project root: `gda daemon install --project <p> --json` returns the envelope at exit `127` in both JSON and human mode. Godot e2e is unaffected by this change (no engine path is touched) and is schedule/dispatch-only on PRs.

## Red-proof

Two rounds, each: commit first, then revert only the production files back to the prior commit while keeping the updated/new tests, confirm red, restore the fix, confirm green again.

- Round 1 (SHA `7aead56c`): the four moved pins + two `_materialize` regressions — all six failed with the raw `PermissionError` escaping.
- Round 3 (SHA `5709c2f3`, review round 2): the nine tests in the table above, each against a surgically reverted guard — routing the repeat-start arm back to a bare `install_harness` reddened its three; widening `_is_filesystem_refusal` back to `return True` reddened the five negative-boundary tests; dropping `EROFS` from the set reddened the ninth. Restored; 73 pass in `tests/test_daemon_lifecycle_payload.py`.
- Round 2 (SHA `71d50f47`, this recheck): eight target tests (the six above plus the two new `unreadable_project` regressions) against the pre-recheck commit — the two new tests failed with the raw `PermissionError` escaping unguarded (proving they catch the capture-site crash), the other six failed on the old `harness_install_unwritable` code name (proving the rename is pinned). Restored; all eight pass again.

## Review round 2: transaction coverage and classifier scope

An adversarial review of `71d50f47` filed two required findings per axis. Both reproduced against that exact head; both are fixed in `5709c2f3`.

**P1 — the repeat-start self-sync was outside the transaction.** `daemon start` against an already-running daemon still self-syncs the harness (#225), and `src/gda/commands/daemon.py:719` called `install_harness(project)` bare — no snapshot, no classifier, no rollback. Reproduced two ways on `71d50f47`:

| Setup (`daemon_pid` faked to 999) | `71d50f47` | `5709c2f3` |
| --- | --- | --- |
| No harness yet, project root `0o555` | `PermissionError: [Errno 13] ... 'proj/addons'` (raw) | `Failure` / `harness_install_permission_denied`, `addons/` never created |
| Stale harness, `project.godot` `0o444` | `PermissionError`, **stale body overwritten** | `Failure` / same code, harness byte-identical to the stale bytes |

#700's acceptance sentence is "`daemon start` in a project whose tree rejects writes returns a structured envelope", and this arm IS a `daemon start`. Fixed by extracting ONE transactional helper, `_install_harness_transactionally` — capture → install → classify a refusal → roll back → re-raise anything else — and routing all three mutating call sites through it. The start-with-no-daemon arm keeps the snapshot the helper returns, because its transaction stays open across the spawn and the readiness wait. This also settles the reviewer's Standards judgement call (Duplicated Code): with three call sites the extraction is no longer a matter of taste.

**P2 — `except OSError` was wider than the code's pinned meaning.** `FileNotFoundError`, `NotADirectoryError`, `ENOSPC` and `EIO` are all `OSError`, and on `71d50f47` all of them returned `harness_install_permission_denied` at exit 127. `GdaError.code` is public ABI (ADR-0002), and telling an agent "a sandbox denied access" for a full disk sends it to escape a restriction instead of freeing space.

Fixed by keying classification on `errno` through `_is_filesystem_refusal`: `EACCES`, `EPERM`, `EROFS`. `EROFS` is included deliberately and is the only judgement in the set — Python raises a plain `OSError` for it, not `PermissionError`, but #700's sentence is "a project whose tree **rejects writes**", and a read-only mount is that same fact with the same remediation, refused by the mount instead of the mode bits. An `OSError` with no `errno` answers `False`, the safe direction.

**Not adopted: a second error code for non-refusal filesystem failures.** The reviewer left the mapping open ("map other filesystem failures separately or preserve their existing behavior"). Preserving is the right half. Falling through restores the exact PRE-#700 behaviour for those cases — the same `HarnessSnapshot` rollback and then propagate once a snapshot exists, and direct propagation when `HarnessSnapshot.capture` itself is what failed, which has written nothing to roll back — so it is not a regression, and #700 asks specifically for "a filesystem-permission failure". A generic code would cost an `error_codes.py` row, an ADR-0002 row, registry + skill sync, and a permanent public promise, to serve a failure mode nothing has dogfooded and no caller has asked to branch on. Residual risk stated plainly: an `ENOSPC` or `EIO` during the install still reaches the caller as a traceback, exactly as it did before this PR.

**ADR-0018 sync.** The #654 note's "An exception still propagates to the caller unchanged" is preserved as the point-in-time record; a dated **Outcome (2026-08-29, #700)** note follows it, distinguishing the four things the reviewer asked for — the typed refusal, the exceptions that still propagate, unchanged rollback semantics, and the pre-snapshot case where nothing was written — plus the widening to all three install paths.

**Public description aligned.** `src/gda/error_codes.py`'s registered description and the ADR-0002 row (compared field-for-field by #701's drift pin) now state exactly what the classifier does: refused access only, naming the refused path, rollback where the filesystem allows it, and an explicit "a filesystem failure that is NOT a refusal ... does not carry this code". `SKILL.md`'s exit-127 row needs no change (it lists codes, not meanings).

**Nine new regressions**, each red-proofed by reverting only the guard under test:

| Test | Guard it pins |
| --- | --- |
| `test_already_running_start_reports_a_write_denied_self_sync` | repeat start, denied `mkdir addons` |
| `test_already_running_start_rolls_a_denied_self_sync_back` | repeat start rollback → byte-identical stale harness |
| `test_already_running_start_reports_an_unreadable_project_before_any_write` | repeat start, capture-site refusal |
| `test_a_full_disk_is_not_reported_as_a_permission_denial_on_start` | `ENOSPC` propagates (start) |
| `test_a_full_disk_is_not_reported_as_a_permission_denial_on_install` | `ENOSPC` propagates (install) |
| `test_a_full_disk_still_rolls_the_half_finished_install_back` | the fall-through arm still rolls back |
| `test_a_non_directory_path_is_not_reported_as_a_permission_denial` | real `ENOTDIR` (`addons` is a file) propagates |
| `test_a_non_refusal_snapshot_failure_still_propagates` | capture-site `EIO` propagates |
| `test_a_read_only_filesystem_is_reported_as_a_refusal` | `EROFS` stays IN the set |

## Sibling note

Nothing here touches `src/gda/script_errors.py`, `docs/command-catalog.md`, `ops/operations.gd`, `src/gda/project.py`, the harness/daemon protocol, or `harness/install.py`'s `HARNESS_VERSION` (#752's file) — no coordination needed with #722 / #721 / #762 / #752 for this slice.

Closes #700


qinhaihong-red:
Addressed both P2s from the recheck review, at `71d50f47`.

**1. Capture-site crash (`chmod 0000 project.godot`).** Confirmed the live repro (`gda daemon install --project <p> --json` → raw `PermissionError` traceback). `HarnessSnapshot.capture()` sat outside the guarded `try:` in both `run_daemon_start_operation` and `run_daemon_install_operation` — fixed by wrapping it in its own `except OSError`, reporting directly with no rollback needed (capture failing means nothing was written yet).

**2. Read mislabeled as write.** Correct that this was unreachable before fixing (1) — confirmed the same reasoning (`chmod 0000` always hits the unguarded `capture()` crash first, so a static permission profile can never reach `_read_config`'s read past it). Fixed together as instructed: renamed the registered code from `harness_install_unwritable` to `harness_install_permission_denied` and reworded the `install_harness()`-site message to be operation-neutral ("could not access", never "could not write") — that call site genuinely can't tell read from write apart (`_materialize` writes, `_read_config` reads, `_write_config` writes, all inside one guarded call). The new capture-site failure, by contrast, knows for certain it's a read and says so precisely.

Also confirmed the dangling-symlink correction: reproduced that it's rejected upstream as `project_not_found` before reaching this code, so no separate handling was needed for it.

Scope held to the install path: `daemon uninstall` untouched, and the spawn/readiness step's raw-raise behaviour is unchanged (`test_start_rolls_back_when_the_spawn_itself_raises` still pins it).

New regressions (chmod-based, permissions restored in teardown, reproducing through the public operation): `test_failed_start_reports_an_unreadable_project_before_any_write` and `test_failed_install_reports_an_unreadable_project_before_any_write`, using a new `unreadable_project_godot` (`chmod 0000`) fixture — plus `"write" not in message` assertions added to the existing write-site tests, pinning the neutral wording there too.

Red-proofed both rounds (commit → revert production files only → confirm red → restore → confirm green); see the PR body's "Red-proof" section for both SHAs.

Gates on the final state:
```
$ uv run pytest -q -m "not e2e"
1945 passed, 543 deselected in 206.15s (0:03:26)

$ uv run ruff check .
All checks passed!

$ uv run ruff format --check .
473 files already formatted

$ uv run pyright
0 errors, 0 warnings, 0 informations
```

CI green on `71d50f47`.

qinhaihong-red:
Reviewed exact head `71d50f477b739ca5a8e23f0c96d1699623d7b8c7` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2` (PR #768).

## Artifact / Markdown

**Changes required — 2 required findings.**

1. **ADR-0018 still states the superseded exception contract.** `docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md:118-129` says that the transactional install runs inside one guarded region and that “An exception still propagates to the caller unchanged.” This PR deliberately converts permission-related snapshot/install failures into `Failure`. Readers therefore cannot tell which exceptions remain raw or which paths return the new envelope. Add a dated #700 outcome that distinguishes permission refusal, unrelated exceptions, rollback, and the pre-snapshot no-write case.

2. **The new public code description and PR body assert semantics and coverage that the implementation does not provide.** ADR-0002:326 and `src/gda/error_codes.py:938-971` define `harness_install_permission_denied` as a sandbox-denied read/write from `gda daemon start` or `gda daemon install`. Exact-head probes instead showed (a) `NotADirectoryError`, `IsADirectoryError`, and `ENOSPC` reported under that code because both guards catch every `OSError`, and (b) a repeat `daemon start` with an existing daemon still raised raw `PermissionError`. Because `GdaError.code` is public ABI, the prose currently teaches agents an incorrect branch and remediation. Correct the classifier/coverage, then align the registry row, ADR table, and PR summary with the resulting precise contract.

**Terminology and prose assessment.** Sources: `CONTEXT.md` (`gda`, `gda-daemon`, `gda harness`, `Error envelope`, `Gda error code`, `Classifier error code`), issue #700, ADR-0002, ADR-0018, and ADR-0028. Writing reference: ASD-STE100 Simplified Technical English. I read the full current PR body, the full current issue, the complete changed ADR-0002, and the complete packaged `gda` skill; I checked all changed prose and the relevant requirement/decision passages for established nouns, alternate names, scope, causality, ordering, and concrete wording. Established terminology is otherwise consistent, and I found no separate substantive word-choice problem beyond the two factual scope contradictions above. I did not run an automated link or controlled-language checker.

## Architecture / DDMA

**Changes required — 2 required findings.**

Authority and observed flow: the Typer daemon command is the inbound adapter; `run_daemon_start_operation` / `run_daemon_install_operation` own lifecycle coordination; `HarnessSnapshot` and `install_harness` own project-filesystem mutation/rollback; `src/gda/error_codes.py` owns code/category/exit/description; `make_failure` projects that authority into the structured envelope. The source dependency direction remains acyclic.

1. **[P1] The harness-install boundary is incomplete and has more than one behavioral owner.** `src/gda/commands/daemon.py:697-734` self-syncs an already-running daemon through a bare `install_harness(project)`, bypassing the snapshot, rollback, and new classifier used at lines 803-815. With `daemon_pid=999`, a stale harness, and read-only `project.godot`, exact head raised raw `PermissionError` and left the stale harness overwritten with v15. This violates completeness (one valid `daemon start` path has no error-envelope/transaction owner), DRY, and the extension seam. Route every start/install mutation through one application helper that captures, installs, classifies only the intended refusal, rolls back, and re-raises unrelated failures; then add a repeat-start regression. Claim `DDMA-768-1` evidence state: `open` (the repository defines no stronger claim-state vocabulary); current evidence disconfirms full-path coverage. It is confirmed only when the repeat-start denial returns the same envelope and restores byte identity.

2. **[P2] The permission code does not own a permission-only predicate.** Lines 803-812 and 910-919 translate all `OSError` instances, while the registry owner at `error_codes.py:962-971` promises a denied read/write. Real parent-as-file and target-as-directory probes, plus an injected `ENOSPC`, all produced `harness_install_permission_denied` at exit 127. This makes a stable domain fact depend on an overbroad adapter exception class and can tell automation to escape a sandbox for a path-shape, capacity, or I/O defect. Centralize an errno predicate for permission/refusal cases (`EACCES`/`EPERM` and an explicitly justified `EROFS` policy); preserve or separately classify other `OSError`s. Claim `DDMA-768-2` evidence state: `open`; current counterexamples disconfirm semantic exclusivity. Confirm it with negative boundary tests for `ENOSPC` and non-directory paths.

Structural assessment: completeness, orthogonality, DRY, and extension-path checks fail at the duplicated install orchestration above; responsibility placement and dependency direction otherwise remain consistent with the current single-context architecture. No new DDD building block is warranted.

## Standards

**Changes required — 3 findings.**

1. **Hard — `src/gda/commands/daemon.py:803-812,910-919`: wrong public error classification.** Both hunks catch every `OSError` and emit `harness_install_permission_denied`. `FileNotFoundError`, `ENOSPC`, and `EIO` are also `OSError`; they are not permission denials. This violates ADR-0002’s rule that `GdaError.code` is public ABI with a pinned meaning (`docs/adr/0002-headless-structured-output-contract.md:213-225,326`). Narrow classification to permission-related errors (including any explicitly justified `EROFS` policy), preserve or accurately classify other failures, and add unrelated-`OSError` regressions for snapshot and install paths.

2. **Hard — ADR-0018’s authoritative lifecycle record contradicts the implementation.** `docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md:118-129` still says, “An exception still propagates to the caller unchanged,” while `daemon.py:803-815,910-922` converts installation `OSError`s into `Failure`. `RULES.md` requires behavior/design changes to synchronize affected ADRs and related artifacts. Add a superseding ADR-0018 outcome that records the typed permission-refusal behavior and rollback semantics.

3. **Judgement call — possible Duplicated Code, `daemon.py:803-815` and `:910-922`.** The snapshot guard, install guard, conversion, rollback annotation, and re-raise are repeated across start/install. Extract one transactional harness-install helper; that also gives the errno classification one authoritative location. This is a Fowler-smell judgement, not a hard violation.

## Spec

1. **[P1] Repeat-start still leaks `PermissionError`.** Spec: “`daemon start` in a project whose tree rejects writes returns a structured envelope … no traceback, JSON and human modes both.” At `src/gda/commands/daemon.py:697-720`, the already-running-daemon branch calls `install_harness(project)` outside both `HarnessSnapshot` and the new exception classifier. An exact-head probe with `daemon_pid=999` and a denied `addons` write re-raised raw `PermissionError`; a stale/missing harness can also remain partially changed. Route this self-sync through the same snapshot/install/rollback helper and add a restricted-write repeat-start regression.

2. **[P2] Non-permission filesystem failures are falsely classified as permission denial.** Spec: “Classify a filesystem-permission failure … into a structured Error envelope.” At `src/gda/commands/daemon.py:803-812` and `907-919`, every `OSError` is converted to `harness_install_permission_denied`. An exact-head `ENOSPC` probe returned that code and claimed denied access, causing automation to recommend escaping the sandbox instead of fixing storage. Catch only permission-denial errnos; map other filesystem failures separately or preserve their existing behavior. Add an `ENOSPC` boundary test.

Passing spec evidence: the normal no-daemon path produced `harness_install_permission_denied`, exit 127, and an `addons` diagnostic in both plain and `--json` CLI invocations. The original chmod-under-`addons` tests, rollback/porcelain test, all four migrated `pytest.raises(PermissionError)` pins, and registry checks passed in the exact-head archive.

**Validation and limits.** The fixed objects resolved; the mandated three-dot diff was non-empty (5 files, 2 commits). In `/tmp/pr768-review.Jx041X`, 8 focused #700 regressions passed, then `tests/test_error_registry.py`, `tests/test_skill_surface_sync.py`, and the full `tests/test_daemon_lifecycle_payload.py` passed (`89 passed`). I rendered both daemon schemas/help surfaces and ran public plain/JSON permission-denial probes. Adversarial probes covered read/write denial, parent-as-file, target-as-directory, `ENOSPC`, and the repeat-start branch. On the reviewed head, all non-skipped CI checks pass; `Godot e2e tests` is skipped (the gda-balancing shard jobs are also skipped as out of scope). I did not rerun the full local non-e2e suite, start a real daemon/Engine session, exercise a read-only mounted filesystem, or independently repeat the dangling-symlink case. Green CI does not cover the two counterexamples above.

Axis summary: Artifact 2 required; Architecture 2 required; Standards 2 hard + 1 judgement call; Spec 2 required. Worst within each axis: stale/overclaimed public documentation; incomplete repeat-start transaction; wrong public-code classification; repeat-start raw traceback and residue.

qinhaihong-red:
## Round-2 review handled — head `5709c2f3`

All four required findings reproduced against the reviewed head `71d50f47` before anything was changed. Two are fully adopted; the P2's *classification* half is adopted with a narrower solution than one of its two options, and the option not taken is argued below rather than skipped.

| Claim | Decision | Evidence on `71d50f47` | Change | Validation |
| **Spec P1 / DDMA P1** — repeat `daemon start` leaks `PermissionError` and leaves residue | **Fully Adopt** | Reproduced twice. (a) `daemon_pid=999`, no harness, project root `0o555` → `PermissionError: [Errno 13] ... 'proj/addons'`, raw. (b) `daemon_pid=999`, stale harness, `project.godot` `0o444` → raw `PermissionError` **and the stale body overwritten** | One transactional helper `_install_harness_transactionally` (capture → install → classify → roll back → re-raise); all three mutating call sites routed through it: start-with-no-daemon, the #225 self-sync, `daemon install` (`5709c2f3`, `src/gda/commands/daemon.py`) | 3 new regressions, red-proofed by routing the arm back to a bare `install_harness`. Byte identity asserted: the stale harness comes back verbatim and `installed_harness_version` is still `stale-old` |
| **Standards 3 (judgement call)** — duplicated install orchestration | **Fully Adopt** | The snapshot guard, install guard, conversion, rollback and re-raise appeared at `803-815` and `910-922` | Subsumed by the extraction above — with a third call site it stopped being a taste question | `ruff`, `pyright`, full non-e2e suite |
| **Standards 1 / Spec P2 / DDMA P2** — `except OSError` is wider than the code's pinned meaning | **Partially Adopt** | Reproduced: injected `ENOSPC` → `harness_install_permission_denied`; a real `addons`-is-a-file → `NotADirectoryError` → same code | Classification keyed on `errno` via `_is_filesystem_refusal`: `EACCES`, `EPERM`, `EROFS`. An `OSError` with no `errno` answers `False` | 5 negative-boundary regressions (`ENOSPC` on start and install, `ENOSPC` still rolling back, real `ENOTDIR`, capture-site `EIO`) + 1 pinning `EROFS` inside the set. All red-proofed by widening the predicate to `return True` / dropping `EROFS` |
| ↳ *sub-claim*: "map other filesystem failures separately" | **Do Not Adopt** | — | No second error code registered | Argued below; residual risk stated |
| **Artifact 1 / Standards 2** — ADR-0018 states the superseded contract | **Fully Adopt** | `0018-...md:118-129` still read "An exception still propagates to the caller unchanged" | Dated **Outcome (2026-08-29, #700)** note APPENDED; the #654 prose is left as the point-in-time record per this repo's ADR convention (cf. the 2026-06-23 note superseded by the #670 one) | The note distinguishes all four requested cases: typed refusal, exceptions that still propagate, rollback semantics, and the pre-snapshot no-write case — plus the widening to three install paths |
| **Artifact 2** — description asserts coverage the implementation lacks | **Fully Adopt** | Both halves of the claim reproduced (see rows 1 and 3) | Classifier and coverage corrected FIRST, then `src/gda/error_codes.py` and the ADR-0002 row rewritten to state exactly what it now does, including an explicit "a filesystem failure that is NOT a refusal … does not carry this code". PR body rewritten to match | `tests/test_error_registry.py` (#701's field-for-field drift pin) + `tests/test_skill_surface_sync.py` pass. `SKILL.md` needs no edit — its exit-127 row lists codes, not meanings |

### On the two judgement points

**`EROFS` is in the refusal set, deliberately.** It is the only member Python does not raise `PermissionError` for. #700's acceptance sentence is "a project whose tree **rejects writes**", and a read-only mount is that same fact with the same remediation — refused by the mount rather than the mode bits. Justification is written into the code next to the frozenset, into the registry comment, and into the ADR note, and pinned by `test_a_read_only_filesystem_is_reported_as_a_refusal`.

**No second code for non-refusal filesystem failures.** The finding offered "map other filesystem failures separately **or** preserve their existing behavior"; preserving is the half taken. Falling through restores the exact pre-#700 behaviour for `ENOSPC` / `EIO` / `ENOTDIR` / `FileNotFoundError` — the same `HarnessSnapshot` rollback, then propagate — so it is not a regression relative to this PR's base, and #700 scopes itself to "a filesystem-permission failure". A generic code would cost an `error_codes.py` row, an ADR-0002 row, registry and skill sync, and a permanent public promise, for a failure mode that has never been dogfooded here and that no caller has asked to branch on. **Residual risk, stated plainly:** an `ENOSPC` or `EIO` during the harness install still reaches the caller as a traceback. If that ever shows up in dogfooding it is a separate issue with its own code, not a speculative row added now.

### Behaviour changes worth flagging

- The repeat-start arm now takes a `HarnessSnapshot` before its self-sync. `capture()` reads `project.godot` plus the harness artifacts — a slightly wider read than the bare `install_harness` did (it also stats the `.gd.uid` sidecar), so a project where that sidecar is unreadable but everything else is fine now returns the typed refusal instead of succeeding. That is the same property the other two install paths already had; the alternative is leaving the arm un-transacted, which is the defect.
- The repeat-start arm's return type is unchanged (`DaemonStartResult | Failure` — it already returned `daemon_already_running`).

### Validation on `5709c2f3`

```
uv run python -m pytest tests/ -q -m "not e2e"   1954 passed, 543 deselected in 53.99s
uv run ruff check .                              All checks passed!
uv run ruff format --check .                     473 files already formatted
uv run pyright                                   0 errors, 0 warnings, 0 informations
```

`tests/test_daemon_lifecycle_payload.py` 73 passed; `tests/test_error_registry.py` + `tests/test_skill_surface_sync.py` 25 passed. Public CLI probe on a `0o555` project root returns the envelope at exit `127` in both JSON and human mode.

**Not verified:** no real read-only mounted filesystem was exercised (`EROFS` is injected, since a unit test cannot mount one), and Godot e2e was not run — no engine path is touched by this change. Round 1's dangling-symlink hypothesis was not revisited; it is rejected upstream as `project_not_found`.

qinhaihong-red:

Reviewed exact head `5709c2f3b5fdb6559064a0769d68bf6099e7f7fe` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review and the author's reply.

The previous behavioral and structural findings are closed: initial start, repeat-start, and explicit install use one transactional helper; permission classification is limited to `EACCES`, `EPERM`, and `EROFS`; unrelated filesystem errors propagate; and ADR-0018 now records the lifecycle accurately.

- ADR-0002 and the public error-registry description state that every non-refusal filesystem failure propagates “after the same rollback.” That is not true when `HarnessSnapshot.capture` itself fails: no snapshot exists and no rollback runs. ADR-0018 correctly distinguishes this pre-snapshot case. Change the public description to say that propagation follows rollback **when a snapshot already exists**.

The terminology and prose pass used the project glossary, ADR-0002, ADR-0018, issue #700, and ASD-STE100. No other substantive prose issue remains.

The lifecycle command owner, unified transaction helper, filesystem adapter, and error registry now have distinct responsibilities and directed dependencies.

- The same public registry prose contradicts the actual capture boundary and ADR-0018, violating the repository's artifact-synchronization requirement. The previous overbroad-`OSError`, stale-ADR, and duplicated-orchestration findings are closed.

### Spec — Pass

Issue #700's structured refusal, rollback/porcelain, original chmod mechanism, and deliberate migration requirements are satisfied.

### Validation

- 98 lifecycle/registry/Skill tests and 9 focused errno/rollback boundary tests passed.
- Direct probes confirmed typed `EACCES`/`EPERM`/`EROFS`, propagation of `ENOSPC`/`EIO`/`ENOTDIR`/`EISDIR`/missing-errno errors, and byte-identical rollback after a half-install `ENOSPC`.
- Plain and JSON permission refusals returned exit 127 without a traceback.
- Required CI checks passed; hosted Godot E2E remains skipped. A real read-only mount, real Engine session, dangling symlink, Windows, and the full non-E2E suite were not tested.

qinhaihong-red:

Head is now `4fb6643f`.

### Artifact / Standards — "after the same rollback" is untrue at the capture point. **Adopted.**

Confirmed. The description is carried identically by `src/gda/error_codes.py` and ADR-0002's table (pinned together by #701), and both said a non-refusal filesystem failure "propagates as before, after the same rollback". That holds at two of the three guarded points. At the third — `HarnessSnapshot.capture` itself — nothing has been written and no snapshot exists, so the exception propagates with no rollback at all. ADR-0018's #700 outcome already distinguishes the pre-snapshot case; the code's own public prose did not.

Both artifacts now read: "it propagates as before: after the same rollback when a snapshot exists, and directly when the failure came from the pre-install snapshot read itself, which has written nothing to roll back."

The adjacent sentence about the refusal path was left as it stands — "any partial write is rolled back where the filesystem still allows it" is already hedged correctly for a capture failure, which has no partial write.

Red-proof of the guard, not just the text: reverting the ADR half alone fails `test_adr_registry_matches_python_authoritative_registry` and nothing else (`1 failed, 17 passed`); restoring it returns `18 passed`. So the two copies are genuinely pinned to each other, and this correction could not have landed on one side only.

### Validation

`1954 passed, 543 deselected` (non-e2e), ruff check / format clean, pyright 0 errors; `test_error_registry.py` + `test_skill_surface_sync.py` green. Documentation only — no behaviour change, and the classifier, transaction helper, and errno predicate are untouched. `Closes #700` intact.

qinhaihong-red:

Reviewed exact head `4fb6643fdd5193e46c27eccefcd33ae6d48fe319` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review, the author's reply, and the incremental change since `5709c2f3`.

The target public surfaces are fixed: ADR-0002 and the `ErrorCodeSpec` description now correctly distinguish rollback when a snapshot exists from direct propagation when capture fails before a snapshot exists.

However, the same lifecycle rule remains inconsistent in adjacent maintained prose:

- `docs/adr/0018-gda-harness-lifecycle-and-concurrent-editor-boundary.md:193-197` says every other exception propagates only after rollback, contradicting its own pre-snapshot exception at `:204-208`.
- `src/gda/commands/daemon.py:574-582,594-602,719-722` still describes all non-refusals as rollback-then-propagate, while `:724-729` directly re-raises a capture failure before a snapshot exists.
- `src/gda/error_codes.py:964-972` retains the same blanket comment immediately above the corrected public description.
- The PR body's Round-2 “Not adopted” paragraph still says these cases use “the same `HarnessSnapshot` rollback, then propagate.”

An exact-head capture-site `EIO` probe confirms the actual boundary: the original exception propagates, `rollback_called=false`, and no `addons/` residue exists.

Smallest correction: scope every rollback-then-propagate statement to failures raised **after capture succeeds**, and explicitly state that a capture-site failure propagates directly because no snapshot exists.

Axis result: **Artifact 1 / DDMA 1 / Standards 1 — Changes required; Spec — Pass.** These are three projections of one remaining authority/synchronization defect; the runtime transaction behavior itself is correct.

Validation: 98 lifecycle/registry/Skill tests passed; the registry/ADR pin and 25 sync tests passed; restricted-write plain/JSON probes returned the expected structured exit-127 envelope without traceback; Ruff, format, and `git diff --check` passed. All non-skipped CI checks succeeded; hosted Godot E2E remains skipped.

qinhaihong-red:
## Round-4 reply — one finding in four places, adopted

Head is now `4dba9929`.

### The finding is right, and its shape is the interesting part

Round 3 corrected the two artifacts that are *pinned to each other* — the `ErrorCodeSpec` description and ADR-0002's row — and stopped there. The same rule is restated in four places nothing pins, and every one of them kept the blanket form. Your list was accurate, including the one I find hardest to defend: **ADR-0018's own #700 outcome contradicted itself four lines apart**, saying "the rollback runs first either way" at `:193-197` and then "a NON-refusal failure there propagates with no rollback attempt" at `:204-208`.

Confirmed by reading the arm itself — a non-refusal at `HarnessSnapshot.capture` re-raises before any rollback machinery is reached, so your exact-head `EIO` probe (`rollback_called=false`, no `addons/` residue) is what the code says too.

All five scoped, plus the PR body:

| Location | Now |
|---|---|
| ADR-0018 `:193-197` | "Where a snapshot EXISTS the rollback runs first … where the failure was `HarnessSnapshot.capture` itself there is no snapshot and nothing was written, so the exception propagates directly (spelled out below)" |
| `daemon.py` errno-set comment | "rollback then propagate once a snapshot exists, and propagate directly when the capture itself is what failed" |
| `_is_filesystem_refusal` docstring | "the pre-#700 propagating behaviour (with the rollback, once there is a snapshot to roll back)" |
| `_install_harness_transactionally` docstring | the two arms named separately, with why the capture arm needs none |
| `error_codes.py` comment above the description | same scoping as the description it introduces |
| PR body round-2 "not adopted" paragraph | same |

`grep` confirms no blanket "rollback, then propagate" or "rollback runs first either way" survives in `src/` or `docs/`.

### What this cost, recorded honestly

Round 3 fixed the copies a test guards and missed the copies it does not. The #701 drift pin ties the registry description to the ADR row and nothing else, so five neighbouring restatements of the same rule drifted freely — which is the defect this repository's own DRY clause exists to prevent, found in the code that documents it.

### Validation

`1954 passed, 543 deselected` (non-e2e), ruff check / format clean, pyright 0 errors. Prose only — the transaction, the errno predicate and the classifier are untouched, so the runtime behaviour you re-verified is unchanged. `Closes #700` intact.

Issue #61, opened 2026-06-11
Autoloads execute on every --project operation: read ops are not zero-execution


qinhaihong-red:
## What was found

#30 established that *reading* a scene must never execute project code, and `scene get` / `node list` honor it at the operation level: they walk `SceneState` without instantiating anything.

But the guarantee does not hold at the **process** level. The runner passes `--path <project>` for any operation invoked with `--project`, and `operations.gd` is a `SceneTree` main loop — so the engine registers and constructs the project's **autoload singletons** during startup, before our `_initialize` runs. If the project defines autoloads, their `_init` (and presumably `_ready`) execute on **every** `gda` operation, including read-only ones.

So today the real trust boundary is: **pointing `gda` at a project with `--project` executes that project's autoload code**, regardless of which command runs. For agents operating on untrusted projects (downloaded assets, cloned third-party repos), this is arbitrary code execution from a command that looks read-only.

## Evidence (verified on Godot 4.6, 2026-06-10)

Project with `[autoload] Spy="*res://auto.gd"` where `auto.gd`'s `_init` prints a marker to stderr:

```
$ gda scene get res://spied.tscn --project <proj> --json
AUTOLOAD: _init executed        <-- project code ran during a read op
gda: running operation: scene-get
{"scene_path": ...}             (exit 0)
```

The marker appears *before* the operation's own diagnostic — the autoload is constructed during engine startup, so the operation script cannot prevent or undo it.

## What to decide / build

This needs a posture decision before any implementation:

1. **Investigate** whether the engine offers a way to suppress autoloads for one-shot headless runs (CLI flag, setting override, non-SceneTree main loop that still serves our operations). Note: a `MainLoop` that is not a `SceneTree` reportedly skips autoloads, but our ops rely on SceneTree semantics — feasibility unknown.
2. If suppression is feasible: gate it behind the runner so the #30 guarantee becomes process-wide for reads (and arguably for mutations too — autoloads are never needed by our file-level operations).
3. If not feasible: formally accept and document the trust boundary (ADR amendment + README `--project` warning), so the "reads execute no project code" claim is scoped honestly to *scene scripts*, not *all project code*.

## Acceptance criteria

- [ ] A decision is recorded (ADR or ADR amendment) on whether `gda` suppresses autoloads in headless one-shot runs or documents them as inside the trust boundary.
- [ ] If suppression is chosen: an e2e test proves a project with a side-effecting autoload produces no side effect on `scene get` / `node list`.
[...]

PR #769, opened 2026-08-28
feat(gda): compose scene validate over instanced sub-scenes (#721)

### LINKED ISSUE #721: scene validate: recurse into instanced sub-scenes for a composed verdict

qinhaihong-red:
## What

`scene validate` answered only about the file it was given. A `parent.tscn` that
instances a broken `child.tscn` returned `valid: true` while the child's script was
missing — reproduced again on Godot 4.6.3 before this change:

```
gda scene validate res://child.tscn   -> valid:false, missing_resource on res://missing_script.gd
gda scene validate res://parent.tscn  -> valid:true,  problems:[]
```

The parent's own walk cannot see the break: Godot resolves `res://child.tscn`, opens it,
and hands back a usable `PackedScene` with the unresolvable reference substituted by null.

The walk now descends into every scene a file REFERENCES — normally the ones it instances,
but the boundary is a union of the scene extension and a declared `type="PackedScene"`, see
**Boundaries** below — and validates each with the same two-stage check (dependencies, then
the binding scan of the loaded scene). Every problem carries a
new `scene` key — the file it was found in — so a child's missing script is never read as
the parent's, and each problem's `nodes` resolve against the right tree.

```
$ gda scene validate res://parent.tscn
invalid res://parent.tscn (1 problem)
  project: /tmp/repro
  missing_resource: res://missing_script.gd (Script)
    in res://child.tscn
    the referenced file does not exist
    nodes: .
```

## Design decisions

**Recurse the scene TEXT, not the engine's stderr.** The engine does already report the
child's failure during the parent's load (four `ERROR:` records). Harvesting them would be
cheaper, but it keys the verdict on engine message prose — exactly the coupling
`gda.script_errors` exists to contain — and it is not even available in the case that needs
it: the load is deliberately skipped whenever the text walk already found problems. The
text walk is also what already produces the `path` / `type` / `nodes` triple a problem
carries, so recursion reuses the shape instead of parsing prose back into it.

**Termination is the `answered` set; depth is bounded separately.** A scene file is validated
ONCE however many sites instance it, which bounds the walk by the finite number of distinct
reachable scenes. Reporting per instancing site instead would multiply one broken file into
N problems — the opposite of the per-file rule the dependency walk already applies to a path
declared twice. (Round 3 split that one set in two: `answered`, which keeps the file's own
verdict to once, and the depth record, which keeps the reachable set order-free — see
**Round-3 review**. Round 4 renamed the second to `reached_depth` and gave the walk's edge
record an explicit outcome — see **Round-4 review** for the audit of all five fields.)

> **Corrected in review.** This section first said a depth cap "would only fire on a chain of
> hundreds of distinct nested scenes, so it was not added". Hundreds is exactly where the
> problem is. Terminating is not the same as finishing in time, and the third commit adds the
> bound — see **Depth bound** below for the measurements and for what the bound does *not* do.

**A cycle is reported, not merely survived.** `chain` holds the ancestors of the current
descent; a reference back into it becomes the new `cyclic_instance` problem kind, attributed
to the file that declares the closing edge with `path` naming the ancestor it re-instances.
`answered` alone would have stopped the walk silently and answered `valid: true` for a
composition Godot refuses to load. Checked before `answered`, since every ancestor is also
answered for. The outcome is TERMINAL and outranks a depth deferral left on the same edge by
a deeper route (round 4), so the bound can never hide a cycle.

**Boundaries — a UNION of the scene extension and the declared type.** A reference is
followed when the path it resolves to ends in `.tscn`/`.scn` **or** its `[ext_resource]` line
declares `type="PackedScene"`. Neither trigger alone is enough.

The declared type can never be a *filter*: Godot's text loader starts a load for EVERY
`[ext_resource]` line before it parses a single node, and passes that line's `type` to
`ResourceLoader` only as a HINT — the format handler is chosen by extension and accepts every
type (`ResourceLoaderText::load`, `ResourceFormatLoaderText::handles_type`). Measured on 4.6.3,
a parent declaring `[ext_resource type="Resource" path="res://child.tscn"]` and never instancing
it emits the same three errors for the child's missing script as a parent that instances it, so
selecting on `type="PackedScene"` would MISS those. But the extension alone misses the opposite
case: `ResourceSaver` writes a PackedScene into a plain `.res` (`ResourceFormatSaverBinary`
accepts `res` for any resource; the text saver refuses, so `.tres` is not a form a PackedScene
can be saved in), and such a child was skipped before any check ran (round 3). So the declared
type is an extra TRIGGER, never a filter.

A `.tscn` is read and composed into the verdict; anything else that loads as a `PackedScene` is
REPORTED (`unreadable_sub_scene`) because only the `[gd_scene]` text carries a dependency set gda
can read — the same reason the top-level op refuses such a file as its own target. A referenced
scene that does not LOAD gets no problem of its own: the referencing file's dependency walk
already named it (`missing_resource` / `unloadable_resource`) with the node that references it,
and `ResourceLoader.load(...) is PackedScene` is the discriminator that keeps the two apart. What
stays outside: a PackedScene stored under a non-scene extension **and** declared as some other
type. gda writes no such file, and gda does not load every reference to find out — that would
load every texture and audio file a scene names.

## Depth bound

Review rated the deep-chain behaviour P1 and attributed an `engine_crashed` (signal 11) to
the missing cap. Measuring the base corrects the attribution. Godot 4.6.3, a straight chain
of N scenes each instancing the next, two runs each on a quiet machine:

| N | base `b616e7fb` | composed, unbounded | composed, bounded |
| --- | --- | --- | --- |
| 100 | 2s `valid` | 5–7s `valid` | 3–4s |
| 300 | 2s `valid` | 38–47s `valid` | 5–6s |
| 1200 | `engine_crashed` | `engine_crashed` | `engine_crashed` |

**The crash is pre-existing and is NOT fixed here.** At N=1200 the engine's own loader
overflows its stack on the *pre-#721* code, where no gda recursion exists at all —
`_op_scene_validate` already loads the top scene and the engine walks the chain itself. It
arrives through that single load, which no bound in gda's walk touches. Nothing in this PR
should be read as fixing it.

**The regression is real and it was mine.** The base is flat (one load, engine-internal
recursion); the composed walk added a per-file pass per level. That term is superlinear and
at N=300 lands close enough to the 60s launch ceiling to cross it into `launch_timeout` under
machine load — which is how both the reviewer and I first saw it.

The bound is therefore on gda's own work, sized **16**: the depth `_packed_scene_root_type`
already refuses past *on the very same axis* (it walks a scene root's instancing chain), and
the value `JSONIFY_MAX_DEPTH` uses. Real compositions nest a handful of levels.

Reaching it is reported, not silently accepted: **`instance_depth_exceeded`**, built through
the same edge-problem shape as `cyclic_instance` (both now go through one
`_sub_scene_edge_problem` builder). It is the only kind that reports a limit of *gda* rather
than a defect of the project, and it still yields `valid: false` — a gate must not answer
"sound" about a subtree it never looked at, so *not established* is reported as invalid, and
the message says `UNCHECKED` in as many words.

## Contract

- `SceneProblem.scene` is a new REQUIRED string on the result, present on every problem
  including the validated scene's own. `SceneProblemKind` gains `cyclic_instance`,
  `instance_depth_exceeded` and `unreadable_sub_scene`.
- Every reported path is CANONICAL (`simplify_path()`, the engine's own normalization), so
  `res://leaf.tscn` and `res://./leaf.tscn` are one file everywhere — in the problem list, in
  the walk's own sets, and in the dependency walk's per-file rule. That includes the result's
  own `path`, which echoes the canonical spelling rather than the caller's (round 3); the field
  now carries a description saying so.
- `--schema` and `gda schema` project both from the models, so they update with no separate
  edit. The cross-language enum mirror test pins `cyclic_instance` against `operations.gd`.

## Verification

Gates, from the committed head:

```
1944 passed, 570 deselected in 52.05s                pytest -q -m "not e2e"
All checks passed!                                   ruff check .
473 files already formatted                          ruff format --check .
0 errors, 0 warnings, 0 informations                 uv run pyright
41 passed in 26.97s                                  pytest -q -m e2e tests/test_e2e_scene_validate.py
44 passed in 49.06s                                  pytest -q -m e2e  (the other four scene suites)
```

Round 2 additionally ran 190 e2e tests across node, project, project_analysis, resource and
classname_resolution — the blast radius of the canonical-path change, since
`_normalize_ext_resource_path` also feeds `scene get`'s instance markers, the id-restoring
re-save and the project analysis walks. Round 3 did not repeat them: `_canonical_resource_path`
is a pure extraction of the branch those callers already took, and every other change is inside
the composed walk. Round 4 did not repeat them either, for the same reason: it renames one walk
key and restructures the walk's own edge record, and nothing outside `_collect_sub_scene_problems`
and its two new helpers reads either.

Every new guard was red-proofed against this commit, by reverting only that guard in the
committed source and running the test that covers it:

| Guard removed | Test | Failure |
| --- | --- | --- |
| the sub-scene descent | `test_an_instanced_broken_child_makes_the_parent_invalid` | `assert True is False` |
| the `cyclic_instance` branch (leaving `visited` to terminate) | both cycle tests | `assert True is False` ×2 |
| the per-file `visited` skip | `test_a_sound_composed_scene_is_still_valid_and_each_file_checked_once` | `Left contains one more item` (the diamond's leaf reported twice) |
| the `scene` attribution stamp | `test_an_instanced_broken_child_makes_the_parent_invalid` | `contract_violation` (exit 5) |
| the renderer's `in <scene>` line | `test_human_output_names_the_sub_scene_a_problem_was_found_in` | line mismatch |
| that line's `!= validated.path` condition | `test_human_output_stays_silent_about_the_scene_the_command_was_given` | `'    in ' not in ...` |
| the new kind's spelling (drifted) | `test_scene_problem_kinds_mirror_the_operations_gd_consts` | set mismatch |
| the depth bound | `test_past_the_depth_bound_the_walk_stops_and_says_so`, `test_the_bound_does_not_hide_a_break_above_it` | `problems: []` / missing entry |
| the bound's edge (`>` → `>=`, fires one level early) | `test_a_chain_at_the_depth_bound_is_still_fully_validated` | an `instance_depth_exceeded` on a chain inside the bound |
| `simplify_path()` on the `res://` branch | `test_two_spellings_of_one_sub_scene_are_one_file` | two problems for one file |
| the depth deferral (report on crossing) | `test_the_depth_verdict_does_not_depend_on_declaration_order` | `[deep-first]` fails, `[direct-first]` passes |
| the deferral flush (defer becomes drop) | `test_a_target_no_route_reaches_in_bound_is_still_reported` | `assert True is False` |
| the `unreadable_sub_scene` report | `test_a_binary_sub_scene_is_reported_unchecked_not_silently_skipped` | `assert True is False` |
| the `is PackedScene` discriminator | `test_a_binary_sub_scene_that_does_not_load_is_still_reported_once` | one problem too many |
| the scene-edge predicate, back to `.tscn` only | `test_a_binary_sub_scene_is_reported_unchecked_not_silently_skipped` | `assert True is False` |
| the extension selector, replaced by `type == "PackedScene"` | `test_a_sub_scene_referenced_as_data_still_breaks_its_owner` | `assert True is False` |
| `instance_depth_exceeded`'s spelling (drifted) | `test_scene_problem_kinds_mirror_the_operations_gd_consts` | set mismatch |
| the declared-type trigger (extension only) | `test_a_packed_scene_saved_as_a_res_is_reported_not_skipped` | `assert True is False` |
| minimum-depth re-expansion (back to one answered set) | `test_a_sound_graph_converging_above_the_bound_is_valid_in_both_orders`, `test_a_break_below_the_convergence_is_found_in_both_orders` | `[deep-first]` fails on both, `[direct-first]` passes on both |
| the root's canonicalization | `test_an_aliased_root_is_the_same_file_as_the_reference_back_to_it` | the alias reports the root's one missing script twice |
| the per-target node merge (back to one id) | `test_an_edge_problem_lists_every_id_that_names_the_target` | `nodes: ["InnerA"]`, `InnerB` lost |
| the walk-scoped edge record (back to per-descent) | `test_a_cycle_below_a_re_expanded_scene_is_reported_once` | the same cycle reported twice |
| the outcome promotion (any record suppresses the cycle again) | `test_a_deferred_depth_edge_does_not_suppress_a_later_cycle` | `[deep-first]` fails (`valid: true`, `problems: []`), `[direct-first]` passes |
| the terminal record (a cycle no longer settles its edge) | `test_a_cycle_below_a_re_expanded_scene_is_reported_once` | `Left contains one more item` (the same cycle twice) |

## Execution surface

`CONTEXT.md`'s `Project-code execution surface` enumerates the POINTS at which one `gda`
run triggers the project's own code, and `scene validate` was never among them — a missing
point rather than a widened one, even before this branch. The second commit names it: the
narrow, compile-only point it is, with nothing instantiated, plus the extent change this
slice makes — the set of compiled scripts grows from the validated scene's own to every
script reachable through the scenes it references. The execution KIND is unchanged, so the
entry's closing ADR-0009 sentence is untouched. `tests/test_e2e_scene_security.py` stays as
it is: it is the #61/#62 characterization suite (autoloads, instantiation `_init`), not an
enumeration of every point in that entry — it has no case for `script run`, `resource
import`, `game call` or `preflight` either.

## Round-2 review

Four findings handled; see the reply comment for the decision table and the measurements.
`docs/command-catalog.md`'s `scene validate` row and composed-verdict block are now IN this
slice (routing reversed after round 2), so the catalog describes the boundary that actually
ships.

## Round-3 review

Four verified residues, all four adopted; see the reply comment for the decision table. Three
of them were one shape — a rule keyed on a stand-in for the thing it is a property of — so the
fix is the class rather than the case.

| Residue, reproduced at `97fb061c` | What changed |
| --- | --- |
| A PackedScene saved as `.res` was skipped before any check: the parent answered `valid: true, problems: []` while the engine reported the child's parse error | `_is_sub_scene_edge` owns the edge rule as the union above |
| The 16-level bound was on the FIRST route walked, so a graph converging at an ANCESTOR of a deep target answered `valid: false` deep-first and `valid: true` direct-first | the walk records the depth each file was expanded at and expands it again from a shorter route; the edge record moved onto the walk with it |
| `res://./root.tscn` seeded a key no child's reference back to `res://root.tscn` could match, so the root was answered for twice under two spellings | `_canonical_resource_path` is the shared identity; the op canonicalizes the target before anything else and the result echoes that spelling |
| An edge problem read its nodes from ONE `[ext_resource]` id while deduplicating by target, so a second id naming the same file lost its nodes | the nodes are merged per target, the rule `_scene_dependency_problems` already applied |

Cost, measured on this machine after the change: a straight chain is unaffected (the bound
still cuts it at 16 levels — 1.2s at N=100, 1.5s at N=300), and a synthetic worst case for the
re-expansion — a 16-scene chain whose root also references every link, declared shallowest-last
so every level improves — runs 0.70s against 0.61s at `97fb061c`. The repeated half is reading
and parsing the text; `_scene_own_problems` (the load and the script compiles) stays behind
`answered` and runs once per file whatever the graph.

## Round-4 review

One P1 and two artifact residues, all three adopted. Head `d9ead2de`.

The P1 is round 3's shape a fourth time: one record standing for two states. A depth
deferral is PROVISIONAL — it stands only while nothing reaches its target — and a cycle is
TERMINAL, but `reported_edges` was a single kind-agnostic boolean over both. On
`root → d1 … d15 → s → t` plus `root → t → s` (every file sound on its own, the only defect
the `s ↔ t` cycle) the deep route's deferral on `s → t` therefore suppressed the
`cyclic_instance` the short route proved on that same edge, and the deferral was then
dropped because `t` had been reached. Reproduced on Godot 4.6.3 before the fix:

| declaration order | `f1ce7043` | `d9ead2de` |
| --- | --- | --- |
| deep-first | `valid: true`, `problems: []` | `valid: false`, one `cyclic_instance` (`res://s.tscn` → `res://t.tscn`) |
| direct-first | `valid: false`, one `cyclic_instance` | the same |

**An edge now carries an OUTCOME, and the promotion rule is written down.**
`SCENE_EDGE_DEPTH_PENDING` is provisional and holds the finding it would publish;
`SCENE_EDGE_REPORTED` is terminal. Provisional → terminal is allowed and WITHDRAWS the
pending finding; every other transition is refused, which is what keeps one edge to one
problem. Promotion loses nothing: a cycle target is an ancestor of the current descent, so
it was reached inside the bound — exactly the condition under which the flush drops a
pending finding anyway. `deferred_depth` goes with it (a pending finding now lives on the
edge it belongs to), so the walk carries five fields rather than six.

**The state audit.** Each remaining field, the one question it answers, and what it does not:

| Field | Answers | Does NOT answer |
| --- | --- | --- |
| `problems` | what has been published so far | anything about the traversal |
| `answered` | has this file's OWN verdict been produced? | anything about its references — a file is answered before its subtree is walked, and stays answered when a later route walks that subtree again |
| `reached_depth` | what is the smallest depth this file was reached at INSIDE the bound, and (by carrying a key at all) was it reached inside the bound? | whether its own problems were produced — that is `answered` — and which routes reached it: only the best one is kept |
| `chain` | which files are ancestors of the descent currently under way? Its SIZE is the current edge's depth — the same fact counted, not a second one | what the walk has seen: it shrinks again on the way back up |
| `edges` | what outcome has this edge — declaring file, then target — been settled with, and what finding is still pending for it? | anything about the target's own verdict |

`expanded` is renamed `reached_depth` as part of that audit: a missing, unreadable or
non-scene target is recorded there at depth 0 and is never expanded, so the old name promised
a fact the field did not carry. The rename is mechanical; the comments now state each field's
question and its non-question at the field itself.

**The two residues.** The order-independence claim on `README.md`, the three translations,
`SKILL.md`, the schema/CLI prose and the catalog is EARNED rather than weakened: both
declaration orders of the graph above are real-engine regressions, and the eight earlier
regression cases stay green. What was missing rather than wrong is the promotion rule, so the
catalog's `cyclic_instance` bullet and the `SceneProblemKind` source comment now state it (a
cycle outranks the bound on the same edge). This body's "each instanced `.tscn`" wording is
corrected to the reference-union boundary that ships — `CONTEXT.md` already said "the scenes
it references".

**Measured, disclosed, not fixed.** Which EDGE of a cycle is named still depends on
declaration order when the cycle can be entered at more than one of its own members:
`root → s → t → s` names `t → s`, `root → t → s → t` names `s → t` (Godot 4.6.3, unchanged
between `f1ce7043` and `d9ead2de`). Both answers are truthful — each names an edge that
closes the cycle, and breaking either breaks it — and the verdict, the kind and the count are
identical. An order-free ATTRIBUTION needs a canonical representative for a cycle (its edge
set out of `chain`, then a deterministic pick), which is new machinery that still would not
settle overlapping cycles, so the property is documented at `SceneProblemKind.CYCLIC_INSTANCE`
instead.

Closes #721


qinhaihong-red:
## Depth finding: adopted, with the attribution corrected

Thanks — the finding is real and the fix is in at `ad9240a7`. But the base measurement changes what it is, so recording both parts.

**The `engine_crashed` (signal 11) is pre-existing.** I reproduced it, then ran the same fixture against the pre-#721 base:

| N | base `b616e7fb` | composed, unbounded | composed, bounded |
| --- | --- | --- | --- |
| 100 | 2s `valid` | 5–7s `valid` | 3–4s |
| 300 | 2s `valid` | 38–47s `valid` | 5–6s |
| 1200 | `engine_crashed` | `engine_crashed` | `engine_crashed` |

(Godot 4.6.3, straight chain of N scenes each instancing the next, two runs each on a quiet machine.)

At N=1200 the base dies the same way, with no gda recursion in that code at all: `_op_scene_validate` already loads the top scene and the engine walks the instancing chain internally. The overflow is the engine's, it arrives through that single load, and no cap inside gda's walk can prevent it. I have not fixed it and the PR does not claim to.

**The regression is real and it is mine — the usable range shrank.** The base is flat at 2s across N=100 and N=300 because it does one load; the composed walk added a per-file pass per level, which is superlinear. At N=300 that lands at 38–47s, close enough to the 60s launch ceiling that it crosses into `launch_timeout` under machine load — which is how you saw it, and how I first saw it too before measuring on a quiet box.

**So the bound is on gda's own work, sized 16.** `_packed_scene_root_type` already refuses past 16 on the very same axis — it walks a scene root's instancing chain — and `JSONIFY_MAX_DEPTH` uses the same value for value recursion. Real compositions nest a handful of levels; 16 leaves headroom and removes the superlinear term (5–6s at N=300).

Reaching the bound is **reported**, in the `cyclic_instance` shape: `instance_depth_exceeded`, naming the file that holds the declining edge and the subtree that was not checked, with both kinds now built by one `_sub_scene_edge_problem` helper. It is the only problem kind that reports a limit of gda rather than a defect of the project, and it still yields `valid: false` — a gate must not answer "sound" about a subtree it never looked at, so *not established* is reported as invalid, and the message says `UNCHECKED` outright.

Also corrected, rather than left standing: my original note said a depth cap "would only fire on a chain of hundreds of distinct nested scenes, so it was not added". Hundreds is exactly where the problem is. That sentence is fixed in the PR body, and the walk's own docstring now separates *terminating* (the `visited` set) from *finishing in time* (the depth bound), because conflating them is what produced the wrong call.

**Pinned by three new e2e tests**, each red-proofed against the committed head by reverting only its guard:

| Guard removed | Test | Failure |
| --- | --- | --- |
| the depth check | `test_past_the_depth_bound_the_walk_stops_and_says_so`, `test_the_bound_does_not_hide_a_break_above_it` | `problems: []` / entry missing |
| the bound's edge (`>` → `>=`) | `test_a_chain_at_the_depth_bound_is_still_fully_validated` | fires on a chain inside the bound |
| the kind's spelling | `test_scene_problem_kinds_mirror_the_operations_gd_consts` | set mismatch |

## `docs/command-catalog.md`: not actioned here, deliberately

The catalog is #722's file this wave, and PR #767 has already landed the composed-verdict paragraphs, the `cyclic_instance` definition and the table row. The `instance_depth_exceeded` kind and the depth sentence are being routed the same way rather than edited from this branch.

Gates on `ad9240a7`:

```
1944 passed, 551 deselected in 224.22s (0:03:44)   uv run pytest -q -m "not e2e"
All checks passed!                                 uv run ruff check .
473 files already formatted                        uv run ruff format --check .
0 errors, 0 warnings, 0 informations               uv run pyright
22 passed in 58.63s                                uv run pytest -q -m e2e tests/test_e2e_scene_validate.py
16 passed in 44.41s                                uv run pytest -q -m e2e tests/test_e2e_scene_security.py tests/test_e2e_scene_preflight.py
```

qinhaihong-red:
Reviewed exact head `ad9240a7101bdb4a822e805b1c97eaa29f27b689` against merge base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`.

## Artifact / Markdown

**Changes required** — 1 required finding.

1. **The public composed-verdict description is broader than the implemented boundary.** `README.md:391,404-412`, the corresponding passages in `docs/README.es.md`, `docs/README.ja.md`, and `docs/README.zh-CN.md`, and `src/gda/skill/SKILL.md:125` tell an agent that the scenes a scene instances are checked. `SceneValidateResult` similarly says the verdict covers every sub-scene. The narrower `.tscn`-only boundary appears in the PR body and CLI help, but not consistently on those public surfaces. A real-engine probe against this head saved a scripted child as binary `child.scn`, then broke its script: a `.tscn` parent that instances that declared `PackedScene` returned `valid: true, problems: []` while Godot stderr contained the child's parse error. Conversely, a `.tscn` referenced as ordinary `Resource` metadata was traversed and made its owner invalid. This can make an admission flow treat an incomplete or unrelated check as a composed verdict. Align the implementation with one authoritative instancing/PackedScene boundary; if binary children cannot receive the same checks, return a structured “not established” problem instead of `valid: true`, and state the boundary on every public surface. The routed `docs/command-catalog.md` text currently exists only on open PR #767 at `cc6fd7c3` and repeats the broad claim, so correct it there rather than crediting it to this diff.

**Terminology and prose assessment.** I used issue #721, `CONTEXT.md` (especially `Headless operation`, `Project-code execution surface`, `Gda error code`, and `Error envelope`), ADR-0002/0009/0010/0040, the PR body/reply, and the current public command descriptions as terminology authorities. ASD-STE100 Simplified Technical English (also required by `RULES.md`) was the writing reference. I inspected the complete PR/issue text and comments, all changed Markdown/README/Skill passages at the pinned head, the translated counterparts, and the changed schema/help prose for word choice, term stability, scope, causality, ordering, and actionable procedures. Apart from the scope finding above, I found no substantive terminology or prose defect. The translations received a semantic comparison, not native-speaker proofreading.

## Architecture / DDMA

**Changes required** — 2 structural findings.

Observed ownership otherwise remains directed: `commands/scene.py` owns the public models/schema, renderer, descriptor, and CLI help; `operations.gd` owns the engine-side dependency/binding and composition walk; dispatch validates the sentinel payload and stamps `project_root`. This preserves the ADR-0040 vertical group slice and introduces no reverse source dependency or second registry.

1. **`DDMA-769-EDGE` — no single projection owns the semantic instancing edge.** `_ext_resource_entries_from_text` already parses `type`, and `_scene_ext_resource_nodes_by_id` sees all property references, but `_collect_sub_scene_problems` selects solely on a `.tscn` suffix (`operations.gd:788-791`). Exact-head probes disconfirm both sides of the claimed boundary: an ordinary `Resource` `.tscn` is traversed, while an instanced binary `PackedScene` is skipped. The traversal therefore owns neither “declared PackedScene” nor “actually instanced scene,” and the execution-surface expansion can include scripts outside the described composition. **Claim evidence state: `open` (disconfirmed by the probes).** Create one scene-instance-edge projection with target identity and instancing node paths, use it for traversal/cycle/depth attribution, and give unsupported PackedScene forms an explicit verdict.

2. **`DDMA-769-ID` — `visited` and `chain` do not use a stable scene-file identity.** `_normalize_ext_resource_path` returns `res://` paths unchanged (`operations.gd:6130-6133`), then the walk keys `visited`, `chain`, and edge reports by those strings. A parent that instances the same broken file as `res://leaf.tscn` and `res://./leaf.tscn` returned two identical `missing_resource` problems, with different `scene` spellings. This violates the per-file knowledge owner, can misattribute cycle-closing edges, and lets lexical aliases evade the cost bound. **Claim evidence state: `open` (the “once per file” claim is disconfirmed).** Give the graph walk one canonical identity function and key all traversal state by it while preserving a chosen display path in the public problem.

With those exceptions, the completeness, dependency direction, cross-language enum mirror, and descriptor-derived schema/render checks passed; no new DDD building block is warranted.

## Standards

**Changes required** — 2 findings; no additional Fowler-baseline smells.

1. **Hard documented-standard violation — instance edges are over-approximated.** `operations.gd:788-791` descends through every `.tscn` external resource without establishing that it is an instanced edge; ordinary property references are included by the shared node-reference scan. A PackedScene stored as data—or a `.tscn` declared merely as `Resource`—can therefore make its owner invalid or create a false `cyclic_instance`, although `README.md:391,404-410` and `commands/scene.py:1231-1237` promise traversal through instanced scenes. This violates `RULES.md:86` (concept and behavior consistency). Build one normalized instance-edge projection and add a real-engine non-instanced-resource regression.

2. **Hard documented-standard violation — the depth verdict depends on declaration order.** At `operations.gd:802-809`, DFS can first reach a leaf through a path deeper than 16 and append `instance_depth_exceeded`, then validate the same leaf through a direct root edge while retaining the stale “UNCHECKED” problem. Reversing the root declarations lets `visited` suppress that finding. The exact-head cross-boundary-diamond probe returned `valid: false` in deep-first order and `valid: true` in direct-first order. This contradicts `commands/scene.py:353-362,468-472` and `RULES.md:86`. Track minimum reachable depth/use breadth-first traversal, or retire a pending depth problem when an in-bound route validates the target; add the diamond as E2E coverage.

The apparent multi-file change amplification is required contract/document/test synchronization under `RULES.md:121-123`, so it is not reported as Shotgun Surgery.

## Spec

**Changes required** — 1 finding.

1. **[P1] The traversal uses a filename suffix instead of the specified instanced-`PackedScene` boundary.** Issue #721 requires: “resolve each `[ext_resource]` of type `PackedScene`, validate it with the same checks.” At `operations.gd:788-791`, the walk follows every path ending in `.tscn` and skips every other path. The exact-head real-engine counterexamples are: (a) the broken binary `child.scn` instance returned a clean composed verdict, and (b) a non-instanced `[ext_resource type="Resource" path="res://child.tscn"]` used only as metadata was traversed and incorrectly made the parent invalid. This both misses an instanced child and validates resources outside the requested composition graph. Select actual instancing/PackedScene edges and handle every referenced form; if a binary child cannot receive the same static checks, return a structured “not established” diagnostic. No other Spec finding was found.

**Validation performed.** From an immutable archive of the pinned head: `tests/test_e2e_scene_validate.py` passed 22/22 against Godot 4.6.3; the targeted CLI, README-i18n-sync, and Skill-surface tests passed 24/24. The emitted `scene validate` Draft 2020-12 schema passed `jsonschema` validation, requires `SceneProblem.scene`, and contains both new enum values. I also ran the three failing adversarial probes described above. Current GitHub CI is green for the required Python/lint/type/package checks; the GitHub `Godot e2e tests` job is skipped, so the local real-engine evidence is separate. I verified that `docs/command-catalog.md` is not in this diff and is currently routed to open PR #767 (`cc6fd7c3`), not yet landed in the base.

**Not rerun / remaining checks.** I did not rerun the full 1,944-test non-E2E suite, ruff, or pyright locally because their current exact-head CI jobs are green; I did not repeat the PR body's N=100/300/1200 timing/crash measurements, test Windows, or perform native-speaker Spanish/Japanese proofreading. Green CI alone was not treated as proof of the traversal contract.

qinhaihong-red:
## Round-2 review: decisions, evidence, commits

Head `97fb061c` (was `ad9240a7`, merge base `b616e7fb`). Every claim below was
re-measured against Godot 4.6.3 on this machine before it was acted on. Two commits:
`a00a00bc` (the two correctness fixes) and `97fb061c` (the new kind and the boundary
statement on every surface).

| Spec P1 (a) / DDMA-769-EDGE / Standards 1, first half: an instanced binary `.scn` child is skipped, so a parent with a broken binary child answers `valid: true` | **Fully adopt** | Reproduced: a `.tscn` parent instancing a `.scn` child whose script has a syntax error, and one whose script cannot bind its node, both returned `valid: true, problems: []` at `ad9240a7` while the engine's own load of the same parent reported the break | New problem kind `unreadable_sub_scene` (`97fb061c`) | `test_a_binary_sub_scene_is_reported_unchecked_not_silently_skipped`; red-proofed ×2 |
| Spec P1 (b) / DDMA-769-EDGE / Standards 1, second half: following a `.tscn` declared `type="Resource"` is an over-reach and can make an owner falsely invalid | **Do not adopt** — the premise is disconfirmed | Measured, see below | Behaviour retained; the reasoning is now IN the code and on every public surface (`97fb061c`) | `test_a_sub_scene_referenced_as_data_still_breaks_its_owner`, red-proofed with the reviewer's own rule |
| Standards 2: the depth verdict depends on declaration order | **Fully adopt**, with the deferral rather than min-depth tracking or BFS | Reproduced: the cross-boundary diamond returned `valid: false` deep-first and `valid: true` direct-first at `ad9240a7` | Depth findings deferred, emitted only for targets nothing else answered for (`a00a00bc`) | `test_the_depth_verdict_does_not_depend_on_declaration_order` (both orders) + `test_a_target_no_route_reaches_in_bound_is_still_reported`; red-proofed ×2 |
| DDMA-769-ID: `visited`/`chain`/attribution do not use a stable file identity | **Fully adopt**, at one place rather than as a walk-local canonicalizer | Reproduced: `res://leaf.tscn` + `res://./leaf.tscn` in one parent produced two identical `missing_resource` entries under two `scene` spellings | `_normalize_ext_resource_path` now applies `simplify_path()` to absolute references too (`a00a00bc`) | `test_two_spellings_of_one_sub_scene_are_one_file`; red-proofed; 190 adjacent e2e tests re-run for the blast radius |
| Markdown 1: the public composed-verdict description is broader than the implemented boundary | **Fully adopt** | Confirmed on `README.md`, the three translations, `SKILL.md`, the schema descriptions | One boundary sentence on all of them plus `CONTEXT.md`, the CLI help and the catalog (`97fb061c`) | `test_readme_i18n_sync.py` green (markers restamped, translations rewritten, not weakened) |
| Markdown 1, closing note: correct the catalog on #767 instead of crediting it here | **Do not adopt** — routing was reversed by the maintainer after this review | — | The `scene validate` row and the composed-verdict block are in this diff, written against what actually ships | Catalog now names all three edge kinds and the file-not-type boundary |

### Counterexample (b), measured

The reviewer's second counterexample does not hold, and the measurement is worth
publishing because it also decides what the boundary sentence should say.

**The declared `type` is not a selector — the engine ignores it for this purpose.**
`ResourceLoaderText::load()` starts a load for *every* `[ext_resource]` line before it
parses a single node, and passes that line's `type` to `ResourceLoader::_load_start`
only as a hint. `ResourceFormatLoaderText::handles_type()` returns `true` for every
type, so `recognize_path` selects the handler by **extension**.

Measured on Godot 4.6.3, two parents in separate processes, `child.tscn` carrying a
missing script:

```
parent_instance.tscn   [ext_resource type="PackedScene" path="res://child.tscn"]  + instance=…
parent_resource.tscn   [ext_resource type="Resource"    path="res://child.tscn"]  + metadata only
```

Both emit the identical three errors:

```
ERROR: Attempt to open script 'res://missing_script.gd' resulted in error 'File not found'.
ERROR: Failed loading resource: res://missing_script.gd.
ERROR: res://child.tscn:6 - Parse Error: [ext_resource] referenced non-existent resource at: res://missing_script.gd.
```

So a reference to a broken scene breaks the referencing scene whatever the line calls
it, and `type="PackedScene"` selection would *miss* those cases. It also has no honest
fallback for the hand-written case the reviewer's rule would have to cover: an
`[ext_resource]` with no `type` at all is `ERR_FILE_CORRUPT` to the engine — such a file
does not load, so there is nothing to fall back to. **Issue #721's "resolve each
`[ext_resource]` of type `PackedScene`" is read as intent, not as a selector**, and the
extension-based rule is kept — now stated as such rather than left implicit.

**The cyclic sub-question, also measured.** A mutual *data* reference between two
`.tscn` files produces the same engine refusal as a mutual instance reference
(`res://cyc_b.tscn:6 - Parse Error: [ext_resource] referenced non-existent resource at:
res://cyc_a.tscn`), so reporting `cyclic_instance` for such an edge is not a false
positive. The same measurement corrected the message, which is a finding neither of us
had: the composition does **not** fail to load. The root loads and instantiates; the
engine drops the closing reference and the nodes it would have contributed vanish
(`Node './A' was modified from inside an instance, but it has vanished`). The message
said "the composition cannot load"; it now says what was measured.

### Choices inside the adopted findings

- **`unreadable_sub_scene` rather than narrowing the docs.** Silence there reproduces
  exactly the GDA-DF-030 failure this command exists to prevent, and it would be
  inconsistent with `instance_depth_exceeded`, whose whole justification is that a gate
  must not answer "sound" about what it did not look at. It yields `valid: false`
  meaning *not established*, with `UNCHECKED` in the message.
- **It is NOT raised for a binary child that fails to load.** Measured: such a child is
  already reported by the referencing file's own dependency walk as
  `unloadable_resource`, so the walk keeps the existing "no second view of one finding"
  rule. `ResourceLoader.load(...) is PackedScene` is the discriminator, red-proofed.
- **Deferral over minimum-reachable-depth tracking or BFS.** Both alternatives change
  the traversal itself — BFS also loses the parent-then-child reading order the problem
  list is documented to have — while the deferral is three lines and is order-free in
  *both* directions: a target another route answered for produces no entry, and a target
  nothing answered for always produces one. The depth entries consequently appear last
  in `problems`, which is now part of the field's description.
- **One identity function, not a walk-local one.** Putting `simplify_path()` in
  `_normalize_ext_resource_path` fixes the walk *and* makes `_scene_dependency_problems`
  honour the per-file rule its own docstring already promised. It is also smaller than
  adding a second canonicalization. The reported `path` is therefore canonical rather
  than as-declared; that is stated in the schema description.
- **Kind names kept.** `cyclic_instance` / `instance_depth_exceeded` name the case that
  occurs in practice; renaming them to "reference" across eight surfaces buys nothing
  behavioural, and the messages now carry the nuance.

### Validation

```
1944 passed, 558 deselected in 53.33s     pytest -q -m "not e2e"
All checks passed!                        ruff check .
473 files already formatted               ruff format --check .
0 errors, 0 warnings, 0 informations      uv run pyright
29 passed in 19.65s                       pytest -q -m e2e tests/test_e2e_scene_validate.py
73 passed in 69.69s                       pytest -q -m e2e  (the five scene suites)
190 passed in 436.12s                     pytest -q -m e2e  (node, project, project_analysis, resource, classname_resolution)
```

The 190-test row is the blast radius of the canonical-path change:
`_normalize_ext_resource_path` also feeds `scene get`'s instance markers, the
id-restoring re-save, and the project analysis walks.

Seven new red proofs, each reverting one guard in the committed source — see the PR
body's table for the failure each produced.

### Not verified, and residual risk

- Windows, and native-speaker proofreading of the Spanish and Japanese passages (they
  received a semantic rewrite against the new English, and the freshness markers were
  restamped).
- The N=100/300/1200 timing measurements were not repeated; the deferral changes when a
  depth finding is emitted, not how much work the walk does.
- A PackedScene saved under `.tres`/`.res` is outside the walk. That is now stated on
  the public surfaces rather than left implicit; no evidence was found that anything
  authors one, and gda never does.
- `unreadable_sub_scene` makes a project that deliberately keeps binary `.scn` children
  permanently `valid: false`. That is the same answer the top-level op already gives for
  a `.scn` target, so it is consistent rather than new — but it is a behaviour change for
  such a project, and re-saving as `.tscn` is the only way inside the verdict.

qinhaihong-red:

Reviewed exact head `97fb061c8023e4197cc94474a080a65ce556e782` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review and the author's reply.

The catalog-ownership finding is closed, and the new implementation improves `.tscn` alias handling, reports binary `.scn` files explicitly, and fixes the tested target-convergence diamond. Four verified residues remain.

1. README, translations, the bundled Skill, and CLI prose describe referenced sub-scenes more broadly than the implementation. A valid `PackedScene` saved as `.res` is still silently skipped and can produce a false clean verdict.
2. The claim that a depth problem means no route reaches the target within 16 levels is false for ancestor-convergence graphs; declaration order still changes the verdict.
3. The once-per-file/canonical claim is false when the root is supplied through an alias such as `res://./root.tscn`.
4. The “every referencing node” claim is false when multiple ext-resource IDs name the same canonical target. The `SceneProblem.scene` description also still says “instanced sub-scene,” although the implemented boundary includes ordinary eagerly loaded `.tscn` references.

Terminology and prose were rechecked against issue #721, `CONTEXT.md`, the applicable ADRs, current schema/help, and ASD-STE100. No additional prose finding was found; Spanish and Japanese text did not receive native-speaker review.

### Architecture / DDMA — Changes required

- `DDMA-769-EDGE`: filename suffixes do not completely own `PackedScene` identity; a loadable `.res` PackedScene is omitted.
- `DDMA-769-DEPTH`: `visited` conflates “the file was answered” with “the subgraph was expanded at minimum depth,” leaving ancestor convergence order-dependent.
- `DDMA-769-ID`: the root does not pass through the canonical identity owner used for child references.
- `DDMA-769-SITES`: edge problems deduplicate by canonical target, but node attribution is read only from the first ext-resource ID.

The four verified behaviors above are hard consistency/completeness violations: `.res` PackedScenes are missed, depth depends on declaration order, root aliases bypass once-per-file identity, and duplicate IDs lose referencing nodes. No separate Fowler-baseline smell remains.

### Spec — Changes required

1. **P1:** Issue #721 requires resolving each `PackedScene`; a `.res` PackedScene with a broken child script still returns `valid: true` while Godot reports the parse error.
2. **P1:** A known-good graph returns different verdicts for deep-first and direct-first declaration order when paths converge at an ancestor.
3. **P2:** `res://./root.tscn` can report the same root problem under raw and canonical identities.
4. **P2:** A once-per-file cycle problem does not merge all referencing sites across equivalent ext-resource IDs.

### Validation

- Godot 4.6.3: 29 scene-validation E2E tests passed; 24 targeted CLI/i18n/Skill tests passed.
- Draft 2020-12 schema, Ruff, format, and `git diff --check` passed.
- Four exact-head adversarial probes reproduced the residues: `.res` PackedScene false-clean, ancestor-convergence order dependence, aliased-root duplication, and duplicate-ID node loss.
- Required CI checks passed; hosted Godot E2E remains skipped. The full non-E2E suite, Windows, and other Godot versions were not rerun.

qinhaihong-red:
## Round-3 review: decisions, evidence, commit

Head `78d9e3d6` (was `97fb061c`, merge base `b616e7fb`). All four residues are real
— each was reproduced against Godot 4.6.3 on this machine before it was touched — and
all four are adopted. One commit, `78d9e3d6`.

Three of the four were the same defect twice removed: a rule keyed on a STAND-IN for
the thing it is a property of. The file extension stood in for "is a PackedScene", one
visited set stood in for "answered AND expanded at minimum depth", and the root
bypassed the identity its children go through. Round 2 fixed the tested instance of
the depth problem; this round fixes the class.

| Spec P1 / `DDMA-769-EDGE` / Markdown 1: a valid `PackedScene` saved as `.res` is silently skipped and can produce a false clean verdict | **Fully adopt** | Reproduced: `ResourceSaver.save(packed, "res://hero_pack.res")` returns OK (the text saver refuses `.tres` — `ERR_FILE_UNRECOGNIZED`, `ResourceFormatSaverText::get_recognized_extensions` gives `tscn` for a PackedScene); a parent instancing that `.res` after its script was broken answered `valid: true, problems: []`, while the `.tscn` twin answered `script_compile_failed` and the engine's own load of the parent printed the parse error | `_is_sub_scene_edge` — a UNION: scene extension OR `type="PackedScene"` | `test_a_packed_scene_saved_as_a_res_is_reported_not_skipped`, `test_a_res_declared_as_something_else_is_still_not_a_scene_edge`; red-proofed |
| Spec P1 / `DDMA-769-DEPTH` / Standards: a known-good graph returns different verdicts for deep-first and direct-first order when paths converge at an ANCESTOR | **Fully adopt**, with minimum-depth tracking rather than a second deferral | Reproduced: `root → d1…d15 → m → t` plus `root → m`, the whole graph sound — `valid: false` with `instance_depth_exceeded` on `res://t.tscn` deep-first, `valid: true` direct-first | `expanded` records the depth each file's references were walked at; a strictly shorter route walks it again | `test_a_sound_graph_converging_above_the_bound_is_valid_in_both_orders`, `test_a_break_below_the_convergence_is_found_in_both_orders` (both parametrized on order), `test_a_subtree_only_the_deep_route_reaches_is_still_reported`; round 2's diamond kept; red-proofed |
| P2 / `DDMA-769-ID`: `res://./root.tscn` reports the same root problem under raw and canonical identities | **Fully adopt** | Reproduced: a self-referencing root with a missing script answered ONE `missing_resource` under the canonical spelling and TWO under the alias, with two `scene` spellings | `_canonical_resource_path` split out of `_normalize_ext_resource_path`; the op canonicalizes before any check | `test_an_aliased_root_is_the_same_file_as_the_reference_back_to_it` (asserts the two spellings give byte-identical results); red-proofed |
| P2 / `DDMA-769-SITES`: an edge problem does not merge referencing sites across equivalent ext-resource ids | **Fully adopt** | Reproduced: a scene declaring `res://self.tscn` and `res://./self.tscn` under two ids, instanced by `InnerA` and `InnerB`, reported `nodes: ["InnerA"]` | nodes merged per canonical target, the rule `_scene_dependency_problems` already applied | `test_an_edge_problem_lists_every_id_that_names_the_target`; red-proofed |
| Markdown 1 / the `SceneProblem.scene` wording: public prose is broader than the implementation, and still says "instanced sub-scene" | **Fully adopt** | Confirmed on `README.md`, the three translations, `SKILL.md`, the CLI help, the schema descriptions and `docs/command-catalog.md` | all of them restated against the union boundary; `SceneProblem.scene` now says "a sub-scene reached from it — normally one it instances, but the walk follows every reference to a scene file"; the result's `path` gains a description naming the canonical echo | `test_readme_i18n_sync.py` green (translations rewritten, markers re-stamped, gate not weakened) |

### The two judgment calls inside the adopted findings

**The declared type is a TRIGGER, not a FILTER — and that does not contradict round 2.**
Round 2 rejected `type="PackedScene"` as a *selector* because it would EXCLUDE real
edges: the engine's text loader starts a load for every `[ext_resource]` line before it
parses a node and passes `type` only as a hint, so a `.tscn` declared `type="Resource"`
breaks its owner exactly as an instanced one does. Using the same field to INCLUDE an
edge is the opposite direction and costs nothing — `_ext_resource_entries_from_text`
already carries it. Both measurements point the same way: widen the trigger, never
narrow it. `ResourceLoader.load(...) is PackedScene` stays as the discriminator, so a
`.res` that does not load at all is still left to the dependency walk's
`unloadable_resource`, and no non-scene reference is loaded to find out what it is.

**Minimum depth rather than weakening the claim.** BFS was rejected again (it loses the
parent-then-child reading order the problem list documents), and so was a second
deferral (it settles a target, not a target's ancestor). Recording the depth each file
was expanded at is what makes the reachable SET order-free, which is the property the
published claim needs. The file's own problems stay behind `answered` and are produced
once, so the expensive half — `_scene_own_problems`, the load and the script compiles —
is still once per file; only reading and parsing the text repeats, and a file's recorded
depth strictly decreases each time inside a bound of 16. Two consequences worth naming:
the edge record moved from the descent onto the walk (a cycle met on two expansions is
otherwise published twice — its own test), and `problems` is now documented as "in the
order the walk first answers for each file" rather than plain depth-first.

**The residual is stated, not implied.** A PackedScene stored under a non-scene
extension AND declared as some other type is still outside the walk. That sentence is
now on `README.md`, the three translations, `SKILL.md`, the CLI help, the schema and the
catalog.

### The reported `path`

The result echoes the CANONICAL spelling, not the caller's — `gda scene validate
res://./main.tscn` answers `"path": "res://main.tscn"`. `SceneProblem.path` was already
canonical after round 2, so this makes one spelling across `path`, every problem's
`scene` and every problem's `path`, and it is what lets a reader compare them. It is
stated in the new `SceneValidateResult.path` description and in the catalog.

### Validation

```
1944 passed, 568 deselected in 53.41s     pytest -q -m "not e2e"
All checks passed!                        ruff check .
473 files already formatted               ruff format --check .
0 errors, 0 warnings, 0 informations      uv run pyright
39 passed in 26.11s                       pytest -q -m e2e tests/test_e2e_scene_validate.py
44 passed in 53.97s                       pytest -q -m e2e  (the other four scene suites)
```

Seven new e2e cases; five new guards, each red-proofed by reverting only that guard in
the committed source:

| Guard removed | Test | Failure |
| --- | --- | --- |
| the declared-type trigger (extension only) | `test_a_packed_scene_saved_as_a_res_is_reported_not_skipped` | `assert True is False` |
| minimum-depth re-expansion (back to one answered set) | the two order-parametrized convergence tests | `[deep-first]` fails on both, `[direct-first]` passes on both |
| the root's canonicalization | `test_an_aliased_root_is_the_same_file_as_the_reference_back_to_it` | the alias reports the root's one missing script twice |
| the per-target node merge (back to one id) | `test_an_edge_problem_lists_every_id_that_names_the_target` | `nodes: ["InnerA"]`, `InnerB` lost |
| the walk-scoped edge record (back to per-descent) | `test_a_cycle_below_a_re_expanded_scene_is_reported_once` | the same cycle reported twice |

Cost, measured after the change: a straight chain is unaffected (the bound still cuts it
at 16 levels — 1.2s at N=100, 1.5s at N=300), and a synthetic worst case for the
re-expansion — a 16-scene chain whose root also references every link, declared
shallowest-last so every level improves — runs 0.70s against 0.61s at `97fb061c`.

### Not verified, and residual risk

- Windows, and native-speaker proofreading of the Spanish and Japanese passages (they
  received a semantic rewrite against the new English and the markers were re-stamped).
- The round-2 190-test blast-radius e2e sweep was not repeated: `_canonical_resource_path`
  is a pure extraction of the branch `_normalize_ext_resource_path` already took for an
  absolute reference, and every other change is inside the composed walk.
- A PackedScene under a non-scene extension AND declared as some other type stays outside
  the walk — deliberately, and now stated on every public surface.
- The result's `path` echoing the canonical rather than the caller's spelling is a visible
  change for a caller that passed a non-canonical path and compared the echo by string.

qinhaihong-red:
Follow-up on the reply above: head is now `f1ce7043`. `78d9e3d6` carries the four fixes; `f1ce7043` is comments only — `SCENE_INSTANCE_MAX_DEPTH` said "instanced sub-scenes" where the walk follows every reference to a scene file and did not state that the bound is on the shortest route, and `_normalize_ext_resource_path` still named the walk's retired `visited` set. No behaviour change; gates re-run green (1944 non-e2e, 39 scene-validate e2e, ruff, format, pyright).

qinhaihong-red:

Reviewed exact head `f1ce7043c4c2c97e79b3e197d7329de5dd1ac75c` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review, the author's replies, and the incremental changes since `97fb061c`.

The four previous findings are closed: declared-`PackedScene` `.res` edges load correctly, acyclic ancestor convergence is re-expanded at minimum depth, the root is canonicalized before traversal, and duplicate ext-resource IDs preserve all referencing nodes.

### [P1] A deferred depth edge can suppress a later real cycle

`reported_edges` uses one kind-agnostic Boolean for both a provisional depth edge and a terminal cycle (`src/gda/ops/operations.gd:864-895,986-1004`). These states are not mutually exclusive across routes.

For this graph:

```text
root → d1 … d15 → s → t
root → t → s
```

deep-first traversal records `s → t` as deferred depth. The short route later makes the same edge a real cycle, but the existing record suppresses `cyclic_instance`; the deferred depth entry is then dropped because `t` was expanded. Real Godot 4.6.3 produced:

- deep-first: `valid:true, problems:[]`
- direct-first: `valid:false`, one `cyclic_instance`

The verdict is therefore still declaration-order dependent and a cyclic composition can be reported false-clean. Track deferred-depth and reported-cycle states independently, or use an explicit edge outcome that can be promoted from provisional depth to terminal cycle. Add both declaration orders as real-Godot regressions.

### Artifact residue

1. README, translations, catalog/schema/CLI prose, and the PR body broadly claim that the verdict does not depend on declaration order; the counterexample above disproves that claim.
2. The PR body still describes the boundary as each instanced `.tscn` / scenes it instances, which is narrower than the implemented reference-union boundary and the current project terminology.

Axis result: **Artifact 2 / DDMA 1 / Standards 1 / Spec 1 — Changes required on all four axes.** The architecture/standards/spec findings are the same underlying edge-state defect, not additional unrelated behavior failures.

Validation: the eight previous regression cases passed; 39 committed scene-validation E2E tests passed; the new adversarial graph reproduced the false-clean result; `1942 passed, 2 skipped, 568 deselected` in non-E2E; targeted CLI/i18n/Skill checks, schema validation, Ruff, format, and `git diff --check` passed. All non-skipped CI checks succeeded; hosted Godot E2E remains skipped.

qinhaihong-red:
## Round-4 review: decisions, evidence, commit

Head `d9ead2def222a19b2be0b36a69eab09bbfb273e7` (was `f1ce7043`, merge base `b616e7fb`).
One commit, `d9ead2de`. All three findings reproduced against Godot 4.6.3 on this machine
before anything was touched, and all three are adopted.

| **P1** — a deferred depth edge suppresses a later real cycle, so a cyclic composition is reported false-clean and the verdict depends on declaration order | **Fully adopt** | Reproduced on `root → d1 … d15 → s → t` plus `root → t → s`, every file sound on its own: `deep-first → valid: true, problems: []`; `direct-first → valid: false, one cyclic_instance`. The deep route defers `s → t` past the bound and stamps `reported_edges[s][t]`; the short route re-expands `s` at depth 2, finds the real cycle, and is suppressed by that same boolean; the deferral is then dropped because `t` was expanded | An edge carries an explicit **outcome** with one promotion rule (below); `deferred_depth` is absorbed into it | `test_a_deferred_depth_edge_does_not_suppress_a_later_cycle`, parametrized on both orders, real-engine; red-proofed |
| **Artifact 1** — the order-independence claim on README, the three translations, the catalog, the schema/CLI prose and this PR body is disproved by that counterexample | **Fully adopt, by earning the claim** | The claim is a property of the walk, and the P1 was the only case that broke it; the public sentences attribute it to the shortest-route bound, which stays true | No sentence weakened. What was MISSING rather than wrong is the promotion rule, so the catalog's `cyclic_instance` bullet and the `SceneProblemKind.CYCLIC_INSTANCE` source comment state it: a cycle outranks the bound on the same edge | Both declaration orders are e2e regressions; the eight earlier regression cases stay green; `test_readme_i18n_sync.py` green |
| **Artifact 2** — the PR body describes the boundary as "each instanced `.tscn`" / "the scenes it instances", narrower than the reference-union boundary that ships | **Fully adopt** | Confirmed in the body's **What** and **Execution surface** sections; `CONTEXT.md` already said "the scenes it references" | Body corrected to the union boundary (extension **or** declared `type="PackedScene"`) and to the shipped terminology; a **Round-4 review** section carries the audit below | `Closes #721` still present; body matches what ships |

### The fix: an outcome, not a fourth boolean

`SCENE_EDGE_DEPTH_PENDING` is **provisional** and carries the finding it would publish.
`SCENE_EDGE_REPORTED` is **terminal**: a problem about that edge is already in the result.

- provisional → terminal is **allowed**, and **withdraws** the pending finding;
- every other transition is refused, which is what keeps one edge to one problem.

Promotion cannot lose a finding, and the reason is an invariant rather than a coincidence: a
cycle target is by definition an ancestor of the current descent, `chain ⊆ reached_depth`,
so that target was reached inside the bound — which is exactly the condition under which the
flush drops a pending finding anyway. The same invariant makes the refused
terminal → provisional transition safe.

The "one problem per target per declaring file" property is now expressed as *the edge is
settled*, not as a separate flag, so `deferred_depth` is gone: a pending finding lives on the
edge it belongs to, and the flush reads the records.

### The per-state audit

| Field | The one question it answers | What it does NOT answer |
| --- | --- | --- |
| `problems` | what has been published so far | anything about the traversal |
| `answered` | has this file's OWN verdict been produced (validated, or reported as one the walk cannot read)? | anything about its references — a file is answered before its subtree is walked, and stays answered when a later route walks that subtree again |
| `reached_depth` | what is the smallest depth this file was reached at INSIDE the bound, and — by carrying a key at all — was it reached inside the bound? | whether its own problems were produced (`answered` does), and which routes reached it: only the best one is kept |
| `chain` | which files are ancestors of the descent currently under way? Its SIZE is the current edge's depth — the same fact counted, not a second one | what the walk has seen: it shrinks again on the way back up |
| `edges` | what outcome has this edge — declaring file, then target — been settled with, and what finding is still pending for it? | anything about the target's own verdict |

One field did still answer two questions, and it is fixed: `expanded` was read both as "the
depth this file's references were walked at" and as "was this file reached inside the bound"
(the flush's test), while a missing, unreadable or non-scene target is recorded there at
depth 0 and is never expanded. It is renamed **`reached_depth`**, which is the single fact
both readers actually want; the comments now state each field's question and its non-question
at the field itself.

### Validation

```
1944 passed, 570 deselected in 52.05s     pytest -q -m "not e2e"
All checks passed!                        ruff check .
473 files already formatted               ruff format --check .
0 errors, 0 warnings, 0 informations      uv run pyright
41 passed in 27.44s                       pytest -q -m e2e tests/test_e2e_scene_validate.py
44 passed in 49.06s                       pytest -q -m e2e  (the other four scene suites)
```

All non-skipped PR checks pass on `d9ead2de`; hosted Godot e2e remains dispatch-only.
The round-2 190-test blast-radius sweep was not repeated: this change renames one walk key
and restructures the walk's own edge record, and nothing outside `_collect_sub_scene_problems`
and its two new helpers reads either.

Both new guards were red-proofed against the committed source:

| Guard removed | Test | Failure |
| --- | --- | --- |
| the outcome promotion (any record suppresses the cycle again) | `test_a_deferred_depth_edge_does_not_suppress_a_later_cycle` | `[deep-first]` fails with `valid: true, problems: []`, `[direct-first]` passes — the reviewer's exact measurement |
| the terminal record (a cycle no longer settles its edge) | `test_a_cycle_below_a_re_expanded_scene_is_reported_once` | `Left contains one more item` (the same cycle published twice) |

### Measured, disclosed, not fixed

Which **edge** of a cycle is named still depends on declaration order when the cycle can be
entered at more than one of its own members: `root → s → t → s` names `t → s`, and
`root → t → s → t` names `s → t` (Godot 4.6.3, unchanged between `f1ce7043` and `d9ead2de` —
this is not a regression of the fix). Both answers are truthful: each names an edge that
closes the cycle, and breaking either breaks it. The verdict, the kind, the count and the
target set are identical, so the published order-independence claim is not affected.

Making the ATTRIBUTION order-free would need a canonical representative for each cycle (take
its edge set out of `chain`, then pick deterministically, then re-read the canonical
declaring file's text for `nodes`) — new machinery that still would not settle overlapping
cycles, where different entry points genuinely find different cycles. Documented at
`SceneProblemKind.CYCLIC_INSTANCE` instead; happy to file it as a follow-up issue if you
would rather have it tracked.

Not verified: Windows, and native-speaker proofreading of the earlier Spanish and Japanese
passages (untouched this round).

PR #770, opened 2026-08-28
fix(gda): decide the live float transport policy in both directions

### LINKED ISSUE #752: bug(gda): prevent tiny-float loss in live JSON transport

qinhaihong-red:
Closes #752

> **Round 4 (adversarial review) — what changed.** Round 3's derived walk still asked a proxy question: it treated "reachable under a described parent" as "disclosed", when the fact a disclosure is a property of is WHICH WRITER produced the value. `perf monitors` computes its window `mean` CLI-side and copies each budget bound out of the caller's own budget file — numbers no Godot writer ever touches — and they inherited the engine writer's sentence, so a real daemon returned `{"value": 1.0, "min": -0.0, "max": -0.0}` under a published claim that a negative zero reads back as `0.0`. There are now TWO published sentences, each naming its writer in its first words, and the guard MEASURES which one a field owes: a recipe-less live command's floats are engine-written by the dispatch branch itself, and a recipe-bearing one is probed — its own recipe driven with a reply whose floats are sentinels, to see which fields they reach. Plus four synchronization residues: this body, and three test comments/names that outran their guards.

> **Round 3 (adversarial review) — what changed.** Round 2's fix had moved the request rule off `game call` onto every LIVE params model — `ExecutionKind.LIVE`, which is itself a proxy: the three ops the daemon answers ITSELF never reach Godot's parser, so `daemon wait-ready --timeout 5e-324` was refused with a parser explanation for a number Python consumes. The rule now binds to the daemon-to-harness LEG (`gda.models.RelayedLiveParams`), and its guard fails BOTH ways. The result contract had the mirror problem — its coverage was a hand-picked list of six — so the required set is now DERIVED by a walk over the live result models, which caught nine more field descriptions across seven commands, including one no review named. Four published wordings were corrected.

> **Round 2 (adversarial review) — what changed.** The request-side policy was applied by ONE params model, so every other live numeric ingress reached the wire unchecked; it is now the wire's rule, applied by a base every LIVE params model inherits, red-proofed against a real daemon on the exact commands the review reproduced. The result-side contract is published in the **machine schema**, not only in help. The corpus's result row was numerically wrong (41/37/18 → **41/15/40**), re-derived on the engine and now DERIVED from a three-state corpus column rather than transcribed. The repeated corpus counts lost their duplicate surfaces, and `docs/command-catalog.md`'s block is in this PR.

## Summary

The live legs carry JSON, and **Godot 4.6.3 changes floats in both directions — differently**. A real-engine differential corpus, built before any policy was chosen (this issue's binding sequencing), separated the two failures and settled the policy:

- **Result direction: fixed.** The harness now writes every reply with Godot's full-precision JSON writer. Exact on **5500/5500** sweep values, where the default writer was exact on 1822.
- **Request direction: bounded, at every RELAYED ingress.** A float whose wire literal Godot's parser reads as `0.0` is **refused** — the corpus shows *no decimal literal at all* can deliver it, so refusing is the only honest option short of not sending a JSON number. The rule belongs to the daemon→harness **leg**, so it is applied by `gda.models.RelayedLiveParams`, the base every relayed params model inherits, over `model_dump()`: nested values, sequence events and every future numeric field are covered by construction, on both the argv and `--params-json` paths. The three ops the daemon answers itself (`diag errors`, `logger tail`, `daemon wait-ready`) are deliberately outside it — their numbers never meet that parser. Values the parser *can* read still arrive changed in their low-order bits: **disclosed, not refused**.
- **Disclosure widened** past `game call --args` to **every float-bearing live reply** (AC 3) — in rendered help, in the **machine schema** (`--schema`), in the bundled Skill, and in the command catalog. *Which* replies those are is derived from the result models, not listed: 12 commands and 25 float-bearing fields, down to a capture receipt's predicate echo and a log record's structured fields.

The wire format is unchanged — same length-prefixed JSON, same ADR-0002 sentinel, same JSON numbers — so **ADR-0021 needed no amendment**. `HARNESS_VERSION` 15 → 16.

## The differential corpus

Godot `4.6.3.stable.official.7d41c59c4`. Every value is carried to the engine as its **IEEE-754 bytes**, never as a decimal literal, so the only literal in the measurement is the one the daemon's own `json.dumps` would write. Committed as `tests/live_number_corpus.py`, re-derived from a real engine by `tests/test_e2e_live_number_transport.py`.

### The committed corpus (96 rows: normal, `DBL_MIN`, subnormal, both signs, short-form and full-precision literals, the flattening boundary at its exact step, and a named row per drift band)

| direction | writer/parser | exact | changed | → `0.0` |
| request | `JSON.parse_string` | 56/96 | 22 | **18** |
| result | default `JSON.stringify` | 41/96 | 15 (incl. `-0.0`) | **40** |
| result | `full_precision = true` | **95/96** | 1 (`-0.0` → `0.0`) | 0 |

**Corrected in round 2, and the correction is now structural.** The result row was published as 41/37/18. Re-derived on the same engine, it is **41/15/40**: the 18 was the REQUEST direction's flattening count, borrowed for a row it does not describe. The corpus recorded the default writer as one boolean, so the three-way split could not be derived from it — it is now a three-state column, and `tests/live_number_corpus.PARTITIONS` computes **all three rows above** from the table. The unit tier asserts the derived splits and the engine tier re-derives the same three rows from a running Godot, so a hand-edited count cannot survive in either place.

Full precision is exact on every row **except** the negative zero — the engine's `JSON::_stringify` returns `"0.0"` for anything equal to zero before the precision argument is consulted (`json.cpp:92-95`), so that residual is architecturally forced and is stated wherever the count is.

Request-direction rows that arrive as `0.0` (each ± its sign): `1.0000000000000002e-300`, `1.2345678901234568e-300`, `3.2956212316547955e-293`, `DBL_MIN` `2.2250738585072014e-308`, max subnormal `2.225073858507201e-308`, `1e-309`, `1e-310`, `1e-320`, `5e-324`.

Rows a magnitude cutoff would have **wrongly** refused, all preserved by the engine: `1e-300`, `1e-305`, `1e-307`, `1e-308`, `2.209278197011611e-293`.

Result-direction examples under the **default** writer: `1e-300` → `0.0`, `1e-33` → `0.0`, `3.141592653589793` → `3.14159265358979`, `0.3333333333333333` → `0.333333333333333`.

### Point-in-time sweep (5500 random values) — measurement, not a committed artifact

The percentages below were measured once during #752 against this engine build. They are **not** committed as a test: a 5500-value random pass is slow, and seeding it to make it deterministic would turn a measurement into a snapshot — the opposite of what the corpus is for. Every load-bearing claim is instead anchored by named corpus rows the e2e re-derives. Read these as colour.

Request direction — Python `json.dumps` literal into `JSON.parse_string`:

| class | n | exact | different non-zero | → `0.0` |
| uniformly random binary64 | 3000 | 1523 (50.8%) | 1402 (46.7%) | 75 (2.5%) |
| uniform ±10⁴ | 1500 | 1368 (91.2%) | **132 (8.8%)** | 0 |
| uniform (−1, 1) | 500 | 428 (85.6%) | 72 (14.4%) | 0 |
| mantissa × 10^[−30, 30] | 500 | 394 (78.8%) | 106 (21.2%) | 0 |

Largest magnitude ever flattened to `0.0`: **10^−292.62**.

Result direction over the same 5500 values:

| writer | exact | different non-zero | → `0.0` |
| --- | --- | --- | --- |
| default `JSON.stringify` | 1822 (33.1%) | 2361 (42.9%) | **1317 (23.9%)** |
| `full_precision = true` | **5500 (100%)** | 0 | 0 |

Largest magnitude the **default** writer flattened to `0.0`: **10^−32.62**.

`gda.live_numbers.wire_flattens_to_zero` agreed with the real engine on every row of both — the committed corpus (re-checked by the e2e on each run) and all 5500 sweep values — with zero mismatches.

### The residual is bigger than "1–2 ULP", and that is a correction to my first version

Anchoring the residual in the corpus falsified the claim I had shipped. "1–2 ULP" held only for literals that fit the parser's **18-digit mantissa cap**. Python writes *fixed* notation for `|v| >= 1e-4`, so between `1e-4` and `1e-2` the leading zeros after the point spend that budget and the low-order digits are dropped outright. Measured on the engine, from a deterministic candidate list:

| value sent | arrives | doubles away |
| --- | --- | --- |
| `13.591409142295225` | `13.591409142295223` | 1 |
| `1250.3538761287377` | `1250.3538761287375` | 1 |
| `90333.33333333333` (16 digits) | `90333.33333333331` | 1 |
| `3.141592653589793e-291` | `3.1415926535897923e-291` | 2 |
| `0.0012345678901234567` | `0.00123456789012345` | **31** |
| `0.00014285714285714284` | `0.00014285714285714` | **105** |

All six are corpus rows now, each carrying a `request_ulp_gap` the e2e re-derives from the engine. Every surface that said "1–2 ULP" — the authority module, the `--args` help, the args field description, the schema `number` branch, the Skill — now says what was measured.

## Why these two mechanisms, and why the policy follows from them

**Result (`JSON::_stringify` → `String::num`).** The default path formats **fixed-point** (`%.*lf`) with at most `MAX_DECIMALS` (32) decimals, which is why everything below ~`1e-32.6` becomes `0.0` and why ordinary values lose their 16th–17th significant digit. `full_precision` routes through `String::num_scientific` (grisu2, shortest round-tripping form) instead. **One argument, complete fix.**

**Request (`built_in_strtod`, `core/string/ustring.cpp`).** The parser reads the digits into an integer `fraction` and multiplies/divides by a power of ten it builds **as a double** from a table of `10^2^i`. `10^309` is not finite, so an applied exponent ≤ −309 divides by `inf`. This is **not re-spellable**: `fraction` is an integer ≥ 1 and the largest finite power the table builds is `10^308`, so the smallest non-zero value any decimal literal can produce is ~`1e-308`.

So the request direction is refuse-or-lie, and the ULP drift is inherent to the same two-rounding computation.

### Alternatives, and why they lost

- **A tagged value-level representation for the lossy values** (candidate 3, as narrowed). It would preserve the 18 corpus rows exactly — but ordinary values *also* arrive changed, by one ULP and — in the mantissa-cap band — by tens of doubles. Tagging only the zeroing class buys an incoherent contract (exact `5e-324`, silently truncated `0.0012345678901234567`), while tagging *every* number is precisely the "not a JSON number on the wire" change ADR-0021 rejected. **Not shipped, and not needed for AC 2**, which is about `0.0`.
- **Pinning a Godot version whose parser conforms.** `built_in_strtod` and the fixed-point `String::num` are still present on today's `master` (checked in the engine checkout), so there is no version to pin to.
- **A decimal exponent / significant-digit heuristic.** Explicitly warned against without a corpus. What shipped is not one: `godot_applied_decimal_exponent` recomputes the engine's own `decPt`/`mantSize`/`exp` intermediates from the literal gda is about to write, and the corpus is what proves it.

**Does the corpus force the protocol change the issue said to stop for?** Only if the requirement were bit-exact request transport for *every* binary64 value. AC 2 is about silent `0.0`, which is fully met here. Reported, not folded in.

## What changed

- **`src/gda/live_numbers.py`** (new leaf authority, no `gda` imports): `MAX_EXACT_JSON_INT` moves here, joined by `GODOT_STRTOD_MAX_POWER`, `godot_applied_decimal_exponent`, `wire_flattens_to_zero`, and the result direction's TWO published sentences — `LIVE_ENGINE_PRECISION` and `LIVE_DERIVED_PRECISION` (round 4). One place the guard, the help, the schema and the Skill read.
- **`src/gda/harness/gda_harness.gd`**: one writer, `_json`, using `full_precision`. Every reply, window envelope, predicate message and `gda_log` record goes through it. Nothing inside the `operations.gd`-mirrored shared coercion block was touched. `HARNESS_VERSION` 15 → 16, pins updated.
- **`src/gda/models.py`**: `RelayedLiveParams`, the base every RELAYED live params model inherits (round 2 introduced it as `LiveParams`; round 3 rebound it). One `model_validator` runs the scan over `model_dump()`, so the refusal is the LEG's rule rather than one command's — nested values and recipe-built wire dicts included. Placed on the params model, not the runner, because ADR-0015 makes it the one authority both input paths build: the refusal is identical on argv and `--params-json`, keeps `game call`'s existing code / exit status / message, and is decided before any daemon exists, so an input mistake stays an input error rather than a live-channel failure. A descriptor walk over the live Typer tree partitions by `gda.daemon.server.DAEMON_SERVED_OPS` and fails both a relayed command registered without the base and a daemon-served one carrying it.
- **`src/gda/commands/game.py`**: the recursive scan MOVED to `gda.live_numbers.find_unrepresentable` (messages and `args[0]['deep'][0]` paths unchanged); `game call --args` help + schema description restated as the decided contract; `game get` gains the result guarantee.
- **The live argv bodies** (round 2): `input mouse-move`, `input mouse-click`, `diag errors` and `logger tail` built their params model directly, which was harmless while no model-level rule could fire. With one, a raw pydantic `ValidationError` escaped — echoing the caller's value and a pydantic.dev URL, #713's defect. All four now route through `params_or_bad_parameter`.
- **The live result schemas**: both sentences are production constants, CONCATENATED into the live result models' field descriptions, so `--schema` (and the gda-mcp wire schemas derived from the same models, ADR-0004) publishes full precision, the negative-zero residual, and — where a reply carries a number gda made — that the residual is not that number's. Deliberately NOT on the shared headless `NodeProperty`: it serves `node get` / `resource get` too, where #771 leaves the default writer in place, and a test asserts neither promise lands on it.
- **Which replies must publish it is DERIVED, not listed** (round 3). `tests/test_live_contract_guards.py` walks each LIVE result model for every field that can carry a float — recursing into nested models, resolving the annotations pydantic leaves as `ForwardRef` until first use, and collecting the sentences on the field and its ancestors (subtree coverage is what keeps the promise off `NodeProperty` while `GameGetResult.properties` carries it). Round 2's hand-picked six left `game set`'s readback, `input mouse-click` / `mouse-move` positions, `input action` / `tap` strengths, `logger tail`'s structured fields and the `perf monitors` window aggregates silent; the walk also found `ScreenCaptureResult.receipt.observed`, which no review named.
- **WHICH sentence a field owes is MEASURED, not declared** (round 4). Coverage alone still let a subtree inherit the wrong writer's claim, so the guard now establishes each float-bearing field's writer and fails a field publishing the other one. Two mechanisms, no third proxy: a live command with **no recipe** is fulfilled by `classify_live(raw, request, cmd.output_model)` — gda.dispatch's own branch, where the CLI constructs no result, so every float in it came out of the harness's JSON (that branch is exercised on `game get` with two values the default writer would have lost); a command **with a recipe** assembles its result CLI-side, so a **probe** drives that descriptor's own recipe with a reply whose floats are sentinels and reads which fields the sentinels reach. A recipe-bearing float-bearing live command missing from `PROVENANCE_PROBES` FAILS rather than defaulting to the engine's sentence, and a discrimination floor pins the four shapes the measurement exists to separate.
- **`CONTEXT.md`**: the `Value projection` entry closed on "so a value reads the same everywhere", which this slice makes false. The shared projection still governs a value's SHAPE everywhere — that is what the entry is for — so the claim is bounded to shape, and the live/headless numeric-fidelity split is named beside it with #771 as its owner.
- **`src/gda/skill/SKILL.md`**: the request-side refusal and both halves of the result-side guarantee.
- **ADR-0041**: a dated `> Outcome` note; the limitation it disclosed twice is now decided. It carries NO corpus counts — it points at `gda.live_numbers`, the one authority, rather than restating figures an ADR cannot keep current. ADR-0021 untouched.
- **Daemon framing**: no change needed — `json.dumps`/`json.loads` are already exact in both directions; the loss was entirely engine-side.

## Red-proof (each from the committed head, reverted after)

| guard | perturbation | observed failure |
| --- | --- | --- |
| harness `_json` full precision | `JSON.stringify(value)` | live e2e: `assert 0.0 == 1e-300` on `game get --property tiny` through a real daemon |
| `game call` underflow refusal | delete the third class | 17 unit failures; live e2e: `game call --args '[5e-324]'` **exits 0** with `{"value":0.0}` — the defect, reproduced |
| engine model boundary | `GODOT_STRTOD_MAX_POWER` 308 → 309 | 10 unit rows + the engine-agreement e2e (`wire_flattens_to_zero(3.2956212316547955e-293)` is False) |
| harness version pin | `HARNESS_VERSION` back to `"15"` | `test_bundled_harness_bytes_are_pinned_to_the_declared_version`; `scripts/harness_version_guard.py b616e7fb HEAD` passes as committed |
| disclosure consistency | drop the sentence from `perf monitor` / `game rect`; weaken it in `SKILL.md` | both consistency tests fail, naming the surface |
| recorded ULP gap | change `mantissa cap ~1.4e-4` from 105 to 1 | unit `assert 1 == 105`; engine tier fails the same row against a live engine |
| quoted corpus counts | delete one corpus row | `assert 95 == 96` — the counts live only in `live_numbers.py` and the catalog, and both prose copies are read by the guard (the harness comment and ADR-0041 point at the authority instead of restating them) |
| **cross-operation refusal** (r2) | `InputMouseMoveParams(BaseModel)` | 4 unit failures (the descriptor walk names the command); **live e2e reproduces the review's defect verbatim**: `input mouse-move 5e-324 1` exits 0 with `{"position":[0.0,1.0]}` |
| **the admission scan** (r2) | `find_unrepresentable` returns `None` | 36 unit failures across every ingress and every corpus row |
| **argv routing** (r2) | revert `input mouse-move` to direct construction | 2 failures — the refusal escapes as a raw pydantic `ValidationError` instead of a usage error |
| **published partition** (r2) | flip one corpus row's `default_stringify` | unit `assert (41,15,40) == (41,16,39)`; the **engine tier** fails the same row against a live Godot |
| **authority prose** (r2) | restore the old `41/37/18` wording in `live_numbers.py` | the prose guard reads the docstring and fails — it now checks the text, not only the corpus |
| **schema publication** (r2) | drop the constant from both `perf monitor` carriers | `test_...reaches_the_machine_schema[perf monitor]` fails |
| **headless non-promise** (r2) | add the constant to the shared `NodeProperty` | `test_the_headless_property_shape_makes_no_live_precision_promise` fails |
| **the one docstring wording** (r2) | reword one of the docstrings | the consistency test fails, naming the command |
| **the relayed partition, forwards** (r3) | `InputMouseMoveParams(BaseModel)` | `these RELAYED live commands' params models do not inherit … ['input mouse-move']` |
| **the relayed partition, backwards** (r3) | `DaemonWaitReadyParams(RelayedLiveParams)` | `these DAEMON-SERVED commands' params models inherit … ['daemon wait-ready']`, plus the accept-side test reproducing the round-3 defect verbatim: `timeout float value 5e-324 cannot cross the live wire` |
| **the derived result coverage** (r3) | drop the constant from `InputActionResult.strength` | the walk names it (`{'input action': ['InputActionResult.strength']}`) and the schema check for that command fails with it |
| **the walk's own vacuity** (r3) | remove its `model_rebuild()` + `ForwardRef` assertion | the floor fails at 11 of 12 commands — `screen capture` silently drops out, which is the failure mode the assertion exists for |
| **the derived help coverage** (r3) | delete `logger tail`'s help paragraph | `test_the_precision_contract_reaches_the_rendered_help[logger tail]` |
| **the writer binding, on a computed value** (r4) | `PerfSampleStats.mean` publishes the ENGINE sentence | `{'perf monitors': [('PerfMonitorsResult.stats.mean', 'gda')]}` |
| **the writer binding, on an echoed bound** (r4) | `PerfBudgetVerdict.min` publishes the ENGINE sentence | `{'perf monitors': [('PerfMonitorsResult.budget.min', 'gda')]}` |
| **the round-4 defect itself** (r4) | restore the blanket engine sentence on the `stats` PARENT | the same failure, naming `stats.mean` — the subtree inheritance that shipped is now caught |
| **an unmeasured recipe** (r4) | delete `perf monitors` from `PROVENANCE_PROBES` | `perf monitors assembles its result CLI-side … add a probe to PROVENANCE_PROBES so the sentence it must publish is MEASURED` |
| **a probe that cannot discriminate** (r4) | make the probe's four samples identical, so the mean equals one | `[('PerfMonitorsResult.stats.mean', 'engine'), ('PerfMonitorsResult.budget.value', 'engine')]` |
| **the derived sentence's coverage** (r4) | delete the derived paragraph from `perf monitors`' help | `test_only_the_replies_that_carry_a_gda_number_publish_the_derived_contract`: `perf monitors` |
| **the derived sentence's BOUND** (r4) | paste it into `game get`'s help, which returns no gda number | the same test, the other direction: `game get` |
| **the negative zero itself** (r4) | `min=rule.min + 0.0` in `_evaluate_budgets` | `assert [1.0, 1.0] == [-1.0, -1.0]` — the sign the review observed, gone |

The boundary red-proof found a real corpus gap: moving the boundary changed no verdict, because no row sat on the step. Fixed by `1e-309` and, more sharply, by two **same-magnitude normal** values whose only difference is digit count — measured on the engine, `2.209278197011611e-293` survives while the **larger** `3.2956212316547955e-293` becomes `0.0`. That pair is what refutes a magnitude cutoff.

## Gates

```
All checks passed!                                    # ruff check .
477 files already formatted                           # ruff format --check .
0 errors, 0 warnings, 0 informations                  # uv run pyright
2193 passed, 548 deselected in 53.43s                 # pytest -q -m "not e2e"
harness version transition is valid                   # scripts/harness_version_guard.py b616e7fb HEAD
```

Live e2e against the real engine (Godot 4.6.3.stable.official.7d41c59c4) + a real `gda-daemon`, run serially, on the round-4 head:

```
 5 passed in  7.63s   # test_e2e_live_number_transport.py (corpus + live path + the ingress coverage)
15 passed in 33.55s   # test_e2e_perf.py test_e2e_game.py
19 passed in 33.79s   # test_e2e_screen.py test_e2e_mcp_live.py
```

The round-3 head's runs are kept for the record:

```
 5 passed in  7.87s   # test_e2e_live_number_transport.py (corpus + live path + the ingress coverage)
19 passed in 39.50s   # test_e2e_game.py test_e2e_perf.py test_e2e_diag.py test_e2e_logger.py
44 passed in 75.97s   # test_e2e_input.py test_e2e_daemon.py
19 passed in 91.84s   # test_e2e_screen.py test_e2e_mcp_live.py
```

The input/daemon group failed once on its first run with *"the engine started but the harness did not connect within 25s"* — the same session-launch contention reported in round 2 — and passed 44/44 on an immediate re-run of the identical pair. Reported, not smoothed over; the quoted line is the clean run.

**The boundary itself, probed against a real daemon** (a throwaway project, `XDG_RUNTIME_DIR` short):

```
gda daemon wait-ready --timeout 5e-324   -> engine_session_not_running
                                            ("the readiness deadline expired…")
gda daemon wait-ready --timeout 25       -> engine_session_not_running (this probe
                                            project defines no main scene)
gda input mouse-move 5e-324 1            -> usage error: "x float value 5e-324
                                            cannot cross the live wire…"
gda input mouse-move --params-json …     -> invalid_params, the same sentence
gda logger tail --params-json
    '{"limit": 9007199254740995}'        -> engine_session_not_running (accepted,
                                            then failed for the real reason)
```

The first line is the round-3 defect, gone: the budget now reaches the daemon and expires, instead of being refused with a claim about a parser it never meets. The third, fourth and fifth show the partition holding in both directions.

The round-1 numbers below are kept for the record:

```
13 passed in 105.27s (0:01:45)    # test_e2e_live_number_transport.py + test_e2e_game.py
38 passed in 169.98s (0:02:49)    # test_e2e_perf.py + test_e2e_daemon.py + test_e2e_harness_install.py
39 passed in 223.41s (0:03:43)    # test_e2e_input.py + test_e2e_screen.py + test_e2e_diag.py + test_e2e_logger.py + test_e2e_mcp_live.py
```

**Two transient e2e failures were observed on this machine and are reported rather than smoothed over.** Neither is quoted above; both groups are quoted from clean runs.

- The daemon group failed once with `game set` reporting *"the engine started but the harness did not connect within 25s"*, then passed on a re-run of the identical three suites, with `test_e2e_daemon.py` passing alone in between (28 passed). Known session-launch contention.
- The windowed group failed 8 tests once, its output naming the windowed display pre-flight ("refused *after* the pre-flight probe passed — the daemon re-checks"), then passed 39/39 on two subsequent runs of the identical set. The display gate reports windowed available on this machine, so the probe was momentarily unavailable during that run.

Both are environmental: the branch changes nothing in `src/gda/display.py` or `src/gda/daemon/`, and its `input.py` / `screen.py` diffs touch no runtime behaviour on those paths — `input.py` re-bases its six params models on `RelayedLiveParams`, routes four argv bodies through `params_or_bad_parameter`, and appends the result contract to four field descriptions; `screen.py` appends it to three. Neither touches display probing or session launch.

## Adjacent findings — reported, not folded in

1. **The headless result path has the identical defect — now tracked as #771.** `ops/operations.gd` (slice #721's file) frames results with the default `JSON.stringify`. Reproduced on this branch: `gda scene get-exports` on a node with `@export var tiny := 1e-300` / `precise := 3.141592653589793` reports `"value":0.0` and `"value":3.14159265358979`. The same one-argument change fixes every headless read (`node get`, `scene get-exports`, `project list`, `resource get`, …).
2. **`String.to_float()` — the `node set` / `game set --value` coercion path** — is the same parser reached through a string rather than a JSON number, so it carries the flattening AND the mantissa-cap truncation. Over the 5500-value sweep: 3713 exact (67.5%), 1712 changed, 75 flattened to `0.0`. It lives inside the shared coercion block mirrored into `operations.gd`, so it needs an owner that can edit both.

## Cross-artifact ownership

`docs/command-catalog.md`'s live-number text is **in this PR** (the wave's one-owner routing for that file was reversed in round 2). It is written from this PR's own measured corpus, sits at the live section's level rather than inside `game call` — a cross-operation rule does not belong to one command — and the stale "#752 owns that cross-operation transport defect" sentence is gone, because this PR delivers it. `CONTEXT.md`'s `Value projection` change is likewise applied here.


qinhaihong-red:
Reviewed exact head `72a4e0b01b80cbc07d2ffa0eb6ba86003a0906ac` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2` (`git diff base...head`; seven commits). The four review axes below are independent.

# Artifact / Markdown

**Changes required — 3 required changes.**

1. **The PR body's committed-corpus result table is numerically wrong.** Under **The committed corpus**, the default `JSON.stringify` row says `41/96` exact, `37` changed non-zero, and `18` became `0.0`. I re-derived all 96 rows from Godot `4.6.3.stable.official.7d41c59c4`, using the PR's IEEE-754-byte probe. The actual partition is **41 exact, 15 changed without flattening (including the `-0.0` residual), and 40 non-zero inputs flattened to `0.0`**. The request row is `56/22/18`, and full precision is `95/1/0`, as stated. The wrong default split materially misstates the defect that justifies the writer change. Correct the table and its column wording, and derive or test these published counts from the engine rows so the table cannot drift from the corpus.

2. **The artifacts describe a request-wide policy that the implementation applies only to `game call`.** The PR summary says “Request direction: bounded”; ADR-0041's outcome says a lossy float “is REFUSED by the params model”; `gda.live_numbers` calls itself the live wire's single authority. In code, only `GameCallParams` calls `wire_flattens_to_zero`. Real exact-head daemon probes silently accepted other live JSON-number inputs: `input mouse-move 5e-324 1` returned success with `position:[0.0,1.0]`, and `input action ui_accept --strength 5e-324` returned success with `strength:0.0`; nested sequence input also succeeded. Until the implementation covers every live numeric ingress, these artifacts must not present the request direction as globally bounded. Prefer completing the shared policy; if scope is intentionally narrower, state the remaining surfaces and do not close #752 as a cross-operation fix.

3. **The command-catalog acceptance criterion is not delivered, and the named sibling still carries a falsified bound.** Exact head #770 leaves `docs/command-catalog.md` saying #752 still “owns that cross-operation transport defect.” The PR body says #767 placed the replacement and that the two can land independently. However, current PR #767 head `cc6fd7c396730be61ac5bf9c2c1d6aabb478c31c` says carried requests “can still arrive 1–2 ULP away,” while this PR's measured rows reach 31 and 105 doubles. Therefore neither #770 alone nor the currently named sibling satisfies #752's catalog-consistency requirement. Put the corrected catalog text in this PR, or block closure/merge on a corrected #767 and update the ownership statement. Do not treat sibling content as already delivered.

## Terminology and prose assessment

I inspected the complete PR body and issue #752 (body, labels, status, and comments), the changed `CONTEXT.md` Value projection entry, accepted ADR-0041, the bundled gda skill, all changed public help/schema descriptions, issue #771 as a scope boundary, and the current #767 catalog paragraph. Terminology sources were `CONTEXT.md`, `CONTEXT-MAP.md`, ADR-0002/0015/0017/0021/0023/0024/0040/0041, and the issue texts. I applied ASD-STE100 Simplified Technical English checks for stable technical nouns, concrete word choice, sentence structure, and accurate causal, scope, dependency, and ordering statements. The three required changes above are the substantive accuracy/scope defects. I found no additional substantive terminology or prose defect. I did not reproduce the uncommitted 5,500-value sweep, so its point-in-time percentages remain unverified.

# Architecture / DDMA

**Changes required — 2 structural findings.**

Authority and observed structure: the accepted architecture keeps command input/output models and descriptors in `commands/*` (ADR-0015/0023/0040), relays LIVE requests through `dispatch -> daemon.protocol -> gda harness` (ADR-0017/0021), and derives schema/MCP surfaces from those contracts (ADR-0004/0012). The new import-free `gda.live_numbers` leaf is an appropriate knowledge owner, and harness `_json` is an appropriate single result writer. Dependencies remain acyclic. The defects are incomplete application and misplaced public-contract ownership.

1. **[P1] No application owner enforces the live request-number domain across commands.** `wire_flattens_to_zero` owns the engine fact, but `game.py` is its only production consumer. `screen capture.await_value`, nested `await_events`, and the input command/sequence x/y/strength fields bypass it before `daemon.protocol.write_message`. The real daemon counterexamples above prove this is not theoretical. This fails completeness (required live ingresses have no policy owner), DRY (a wire invariant is treated as a command rule), and extensibility (every new LIVE numeric field must remember a hidden `game`-local guard). Keep `live_numbers` as the leaf fact owner, but invoke one recursive admission policy at the shared LIVE application/dispatch boundary before any daemon write; command models can add narrower rules without redefining the wire invariant. Cover top-level and nested values, safe integers, booleans, non-finite values, both invocation paths, and every LIVE command family.

2. **[P2] Result precision is owned in Typer help docstrings, outside the machine-contract path.** Five command docstrings repeat the guarantee and a test compares them with a test-owned sentence, but the output models/descriptors that produce `--schema` do not carry it. As a result, `game get --schema` describes only projection shape, and `perf monitor --schema` says only “value as JSON” / “arguments as JSON”; schema-driven and MCP consumers cannot discover full precision or the negative-zero residual. Give the live result contract one production authority (for example, a shared description in `live_numbers` referenced by live-specific output schema descriptions/customization), and derive or test help, schema, Skill, and catalog projections against it. Do not put the claim on the shared headless `NodeProperty` description while #771 leaves headless fidelity different.

Load-bearing claims use `open` because the project defines no claim-evidence-state vocabulary:

- `ARC-770-001` — all lossy daemon-to-harness request floats are refused. **State:** open. **Evidence:** disconfirmed by successful exact-head input probes. **Disconfirming condition:** already met by any live numeric input outside `GameCallParams`. **Needed validation:** a descriptor-wide nested numeric corpus across every LIVE input model and real daemon paths.
- `ARC-770-002` — the result policy is consistent across public channels. **State:** open. **Evidence:** harness writer and rendered help pass, but emitted schemas omit the contract and the catalog is stale. **Disconfirming condition:** already met by the current schema/catalog text. **Needed validation:** schema/help/Skill/catalog consistency plus representative game, perf/rect, screen-predicate, handshake/error, and gda_log replies.

The existing topology can be retained; these findings require completing the shared boundaries, not a new context or layer.

# Standards

**Changes required — 2 hard violations and 1 judgement call.**

1. **Hard — test text normalization.** `tests/test_game_commands.py:884-888`, `tests/test_live_numbers.py:239-243`, and `tests/test_e2e_live_number_transport.py:102-111` bypass `tests/support.py:20-25,40-54`, which requires help/usage-text assertions to use the shared normalization. The new hunks use raw `rendered.stdout` or duplicate narrower regexes such as `re.sub(r"\x1b\[[0-9;]*m", ...)`. Reuse `plain_text`; extract a shared text-level Rich-panel normalizer usable by both `CliRunner.Result` and `CompletedProcess`. This duplication is also a **Duplicated Code** judgement call.

2. **Hard — unconstrained repeated facts.** `RULES.md:117-123` requires one authority and generation, reference, or consistency tests for repeated expressions. `src/gda/live_numbers.py:25-29` states `41 of 96` and `95 of 96`; `src/gda/harness/gda_harness.gd:1819-1822` repeats both, and ADR-0041:10-13 repeats `95 of 96`. `tests/test_live_numbers.py:94-107` says it guards these copies but only recomputes corpus counts; it never reads the duplicated prose. Remove counts from secondary surfaces and reference the corpus/authority, or add an actual synchronization guard.

3. **Judgement call — Data Clumps.** `tests/live_number_corpus.py:34` encodes five travelling fields as `tuple[str, float, bool, bool, int | None]`, then both test modules repeatedly destructure positional slots (`for label, value, zeroes, _, _ ...`). Introduce a small named record such as `LiveNumberCase` and access fields by name.

# Spec

1. **[P1] The request policy is not cross-operation.** Issue #752 requires: “Choose and implement one evidence-backed live-number transport policy that prevents silent success on a changed numeric value in both directions” and “A successful live operation never reports a value that silently became `0.0` unless the public contract explicitly accepts that projection loss.” The refusal exists only in `GameCallParams` (`src/gda/commands/game.py:350-440`). Other live JSON-number inputs remain accepted: mouse coordinates and action strengths (`src/gda/commands/input.py:133-180,318-324,782-904`) and `screen capture.await_value` (`src/gda/commands/screen.py:144-155`). Exact-head validation showed `write_message` emits those values; Godot parses them through the corpus-proven lossy parser and successfully echoes zero. Apply the policy at every live numeric ingress, including nested sequence/await events.

2. **[P2] The selected result policy is absent from machine schemas.** The issue requires: “The selected policy is consistent across daemon framing, harness parsing/stringifying, machine schema, rendered help, command catalog, and the bundled Skill.” The new wording is only in Typer command docstrings (`src/gda/commands/game.py:633-636`; analogous perf/screen hunks), which rendered help consumes. Exact-head `gda game get --schema` still describes `NodeProperty.value` only by projection shape (`src/gda/models.py:730-733`), while `gda perf monitor --schema` says merely “The property's value as JSON” / “arguments as JSON” (`src/gda/commands/perf.py:133-148`); neither publishes full-precision behavior or the negative-zero residual. Thus schema-driven clients cannot discover the required contract.

## Validation and limits

- Exact-head local: `279 passed` for live-number, game-command, and harness-install tests; `94 passed, 4 deselected` for Draft 2020-12 schema checks; `4 passed` in `test_e2e_live_number_transport.py` against real Godot 4.6.3 and a real daemon.
- Real-engine corpus: request `56/22/18`; default result `41/15/40`; full-precision result `95/1/0`. This covered the same-magnitude boundary pair, `DBL_MIN`, subnormals, `1e-300`, pi, and negative zero. Nested `game call` refusal passed.
- Other real daemon paths: the input counterexamples above; `perf monitor`, `game rect`, and `logger tail`/`gda_log` reply paths served. Harness version guard passed for exact base/head. Changed Python files passed ruff check/format; `git diff --check` passed.
- Current CI: every non-skipped check succeeds. Hosted **Godot e2e is skipped**, so it is not evidence for engine behavior.
- Not run: the full local non-e2e suite, the uncommitted 5,500-value sweep, a windowed screen-capture probe, or an exhaustive real-daemon matrix of every LIVE numeric field and error/handshake path. Static inspection found no remaining harness `JSON.stringify` call outside the single `_json(..., full_precision=true)` writer.

Axis summary: Artifact 3 required changes; Architecture 2; Standards 3 (worst: hard); Spec 2 (worst: P1). This was a defensive input-robustness and public-contract review only.

qinhaihong-red:
# Round-2 review handled — head `f2495423629244f6c73bcc3f3c6db78fbe58da65`

Base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`, five new commits on `72a4e0b0`. Every finding was verified against the reviewed head before it was acted on; one suggested placement is **partially adopted**, with the reasoning below.

| Spec P1 / Arch P1 | The request policy is not cross-operation — only `GameCallParams` calls the guard | **Fully adopt** (problem), **partially adopt** (placement — see below) | Reproduced: with the guard removed, `input mouse-move 5e-324 1` exits 0 with `{"position":[0.0,1.0]}` against a real daemon | `gda.models.LiveParams`, inherited by all 18 LIVE params models; the scan moves to `gda.live_numbers.find_unrepresentable` — `01226a32` | New e2e drives every reproduced ingress through a real daemon; a descriptor walk over the live Typer tree fails a LIVE command without the base |
| Spec P2 / Arch P2 | The result contract is owned in Typer docstrings, absent from `--schema` | **Fully adopt** | `game get --schema` described only projection shape; `perf monitor --schema` said "value as JSON" | `LIVE_RESULT_PRECISION` in `gda.live_numbers` is the production authority; the six live result models CONCATENATE it into the field descriptions `--schema` publishes; the five docstring copies are pinned against that constant — `01226a32` | `test_..._reaches_the_machine_schema` over all six commands; a second test asserts the promise stays OFF the shared headless `NodeProperty` (#771) |
| Artifact 1 | The published corpus table is numerically wrong | **Fully adopt** | Re-derived on `4.6.3.stable.official.7d41c59c4` with the committed IEEE-754 probe: default `JSON.stringify` is **41 exact / 15 changed / 40 → `0.0`**, not 41/37/18. The 18 was the REQUEST row's flattening count | Corpus records a three-state `default_stringify`; `PARTITIONS` DERIVES all three published rows; PR body corrected — `0b18992c` | Unit tier asserts the derived splits; the engine tier re-derives the same three rows from a running Godot each run |
| Artifact 2 | Artifacts claim a request-wide policy the code does not apply | **Fully adopt** | Same reproduction as Spec P1 | The implementation was completed rather than the claim narrowed — `01226a32` | As above |
| Artifact 3 | The catalog AC is undelivered and the named sibling carries a falsified bound | **Fully adopt** | Head still said #752 "owns that cross-operation transport defect" | The block is in THIS PR, written from this PR's own corpus, placed at the live section's level (a cross-operation rule is not `game call`'s); the ownership sentence is gone — `a9e71ce5` | Sibling content not read or relied on |
| Standards 1 (hard) | Three test modules hand-roll ANSI/box stripping | **Fully adopt** | `tests/support` already owned it for usage errors | `panel_text` extracted; `usage_error_text` built on it; all three copies deleted — `0b18992c` | Full suite green |
| Standards 2 (hard) | `41 of 96` / `95 of 96` repeated on three surfaces under a guard that never read the prose | **Fully adopt — both remedies** | The guard only recomputed the corpus | Counts REMOVED from `gda_harness.gd` and ADR-0041 (they point at the authority); the authority's own docstring is now READ by a test that derives the expected strings from the corpus — `a9e71ce5` | Red-proof: restoring the old `41/37/18` wording fails the prose guard |
| Standards 3 (judgement) | Data clump: the 5-tuple destructured positionally in two modules | **Fully adopt** | — | `LiveNumberCase` NamedTuple; both modules read fields by name — `0b18992c` | Full suite green |
| Standards 2, follow-through | The new catalog block reintroduced the counts as a third unowned copy | **Self-caught** | Same DRY clause as Standards 2 | The prose guard now reads `docs/command-catalog.md` beside `live_numbers.__doc__`, deriving both from `PARTITIONS` — `f2495423` | Red-proof: `preserved 41 of the 96` → 42 fails |

## The one placement I did not take, and why

The review asked for the admission policy "at the shared LIVE application/dispatch boundary before any daemon write". I kept the recursive scan as the reviewer specified — one function, one owner, nested/safe-integer/bool/non-finite coverage, both input paths — but applied it at the **params-model boundary** (`gda.models.LiveParams`) rather than in `dispatch` or `DaemonRunner`:

- **ADR-0015 makes the params model the one authority both the argv and `--params-json` paths build.** A validator there refuses identically on both. A check in `_emit`/`DaemonRunner` runs *after* model construction, so `--params-json` could no longer report `invalid_params` for it.
- **It preserves what the review required to preserve.** `game call`'s refusal keeps its code, exit status and message verbatim, including the `args[0]['deep'][0]` path; the existing tests still exercise the real rule rather than a vestige. There is no second, narrower model-level copy — the group-local guard was deleted, not kept.
- **A refused number is an input error, not a live-channel failure.** At the runner it would have to become a synthesized live envelope with `EXIT_LIVE`, giving two different public contracts for one defect. It also would not fire without a running daemon; it now does.
- **Coverage is by construction, not by memory.** The scan reads `model_dump()`, so nested params models (sequence events, a predicate's awaited value) and the recipes' wire dicts — built from these same fields — are covered, and `tests/test_live_contract_guards.py` fails a LIVE descriptor whose input model does not inherit the base.

**Residual risk, stated plainly:** the scan sees the model's fields, not the literal wire dict. A recipe that INVENTED a number from something other than its params would not be covered. None does today — the three live recipes derive their wire numbers from model fields plus one constant frame ceiling — and any new numeric *field* is covered automatically, but a future recipe could reintroduce the gap without a test failing.

## Found while fixing, also handled

- **Four live argv bodies bypassed `params_or_bad_parameter`** (`input mouse-move`, `input mouse-click`, `diag errors`, `logger tail`). Harmless while no model-level rule could fire; with one, the refusal escaped as a raw pydantic `ValidationError` — echoing the caller's value and a pydantic.dev URL, the exact defect #713 fixed elsewhere. All four now route through it (`01226a32`).
- **`game call` claimed "Every float a live reply RETURNS is exact"** in both its help and its schema number branch, while the same schema now publishes the negative-zero residual. The absolute form is gone (`34bf7004`).

## Validation

- `ruff check` / `ruff format --check` clean; `uv run pyright` **0 errors**; `pytest -q -m "not e2e"` **2169 passed, 548 deselected**; `scripts/harness_version_guard.py b616e7fb HEAD` valid.
- Real Godot `4.6.3.stable.official.7d41c59c4` + a real `gda-daemon`, serial: live-number transport **5 passed**; game/perf/diag/logger **19**; input/daemon **44**; screen/mcp-live **19**; harness-install **4**. No flakes this round.
- **Red-proofs (8, each reverted).** Cross-operation refusal removed → the e2e reproduces the review's defect verbatim (`{"position":[0.0,1.0]}`, exit 0); admission scan neutered → 36 unit failures; argv routing reverted → the raw `ValidationError` escapes; one corpus row's verdict flipped → the derived partition AND the live-engine tier both fail; the authority's prose reverted to `41/37/18` → the prose guard fails; both `perf monitor` schema carriers stripped → the schema test fails; the promise added to the shared `NodeProperty` → the headless non-promise test fails; one docstring reworded → the consistency test fails, naming the command.

## Not verified

- The **5,500-value sweep** percentages are still a point-in-time measurement, not a committed artifact, and are labelled as colour in the body. Every load-bearing number is a corpus row the e2e re-derives.
- **Hosted CI runs no Godot e2e**, so the engine evidence above is local only.
- **Out of scope, observed not fixed:** `input mouse-move`'s echoed `position` is the parsed double, while the injected `Vector2` is float32 — so a value below the float32 range is echoed faithfully but injected as `0.0`. That is the engine's destination type, not JSON transport, and widening #752 to every Godot type's precision is a different decision. Say the word and I will file it.

qinhaihong-red:

Reviewed exact head `f2495423629244f6c73bcc3f3c6db78fbe58da65` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review and the author's reply.

The previous harness-relayed request gaps are closed: mouse movement, action strength, nested sequences, and screen-await values now refuse numbers that Godot would silently flatten. Corpus partitions are derived as request `56/22/18`, default result `41/15/40`, and full precision `95/1/0`. Catalog ownership, the stale 1–2 ULP statement, Rich normalization, repeated count ownership, and the positional corpus data clump are also closed.

1. The “all LIVE requests” claim is too broad. `daemon wait-ready` is daemon-served and its timeout never crosses the Godot parser, yet `LiveParams` rejects `5e-324` with a Godot-loss explanation.
2. Result-contract discoverability is still not cross-operation. Machine schema, MCP, help, and catalog omit float-bearing results such as `game set` readback, input positions/strengths, and logger structured fields.
3. Current prose still contains factual drift: the corpus column named “changed non-zero” includes `-0.0 → 0.0`; the PR body still says ADR/harness contain counts that were removed; and the catalog says “Results are exact” immediately before documenting the negative-zero residual. A “see below” reference also points upward.

Terminology and prose were rechecked across the complete PR body, issue #752/#771, ADR-0041, catalog, `CONTEXT.md`, the bundled Skill, rendered help/schema, and ASD-STE100. No additional substantive terminology finding remains.

### Architecture / DDMA — Changes required

- `ARC-770-001`: `ExecutionKind.LIVE` is being used as a proxy for “relayed through the Godot harness,” but `DAEMON_SERVED_OPS` defines a distinct path. Numeric admission must bind to the actual daemon-to-harness boundary so daemon-served parameters are not over-refused.
- `ARC-770-002`: the result contract is projected manually onto six command schemas/docstrings rather than derived or guarded across all float-bearing live replies. The `live_numbers` and harness `_json` authorities are otherwise appropriately placed and dependencies remain acyclic.

### Standards — Pass

Shared Rich normalization, corpus-count authority, and the named corpus record are proportionate and synchronized; no current documented-standard violation or actionable Fowler smell remains.

### Spec — Changes required

1. `DaemonWaitReadyParams` incorrectly rejects a valid, finite timeout that Python consumes directly and never sends to Godot.
2. The full-precision/negative-zero policy is still missing from some float-bearing output schemas, MCP tools, help, and catalog entries, so the selected cross-operation result policy remains only partially discoverable.

### Validation

- 413 focused unit/schema/MCP/harness tests passed.
- Godot 4.6.3 + daemon: 5 live-number E2E and 3 perf/timeline/logger reply tests passed.
- All 18 LIVE public schemas passed Draft 2020-12 validation.
- Actual MCP `tools/list` and rendered help confirmed that `game get` carries the precision contract while `game set`, input commands, and `logger tail` do not.
- Direct model/daemon-handler comparison reproduced the `wait-ready` over-refusal.
- Ruff, format, `git diff --check`, harness-version transition, and all non-skipped required CI checks passed. Hosted Godot E2E remains skipped; the full non-E2E suite, 5,500-value sweep, windowed capture, and every error/handshake path were not run.

qinhaihong-red:
# Round-3 review handled — head `ab267f68ca792ee14a37dd34ebb9c2eb6c7922b9`

Base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`, two new commits on `f2495423`. Every finding was reproduced against the reviewed head before it was acted on. One suggested *placement* is **partially adopted**; everything else is fully adopted.

| Spec 1 / Artifact 1 / `ARC-770-001` | `ExecutionKind.LIVE` is a proxy for "relayed"; `DaemonWaitReadyParams` rejects a finite timeout Python consumes | **Fully adopt** | Reproduced on the reviewed head: `gda --json daemon wait-ready --timeout 5e-324` refused with the Godot-parser sentence, for a value `DaemonServer._handle_wait_ready` reads in Python and never relays | The base is `gda.models.RelayedLiveParams`, inherited by the 15 relayed params models only; `diag errors`, `logger tail` and `daemon wait-ready` return to `BaseModel`, each with a comment saying why — `d742b59f` | The guard now partitions the live Typer tree by `DAEMON_SERVED_OPS` and fails **both** ways; a new unit test builds each daemon-served model with the refused value and drives the same value through the CLI; real-daemon probe below |
| Spec 2 / `ARC-770-002` | The result contract is projected manually onto six commands rather than derived or guarded across all float-bearing live replies | **Fully adopt** (the coverage), **partially adopt** (the help mechanism — see below) | Reproduced: `game set --schema`, `input mouse-move --schema`, `logger tail --schema` published no precision contract; their help said nothing either | A walk over each LIVE result model reports every float-bearing field; the guard fails on any that discloses nothing, and the schema + help checks run over exactly that derived set. Nine field descriptions and seven docstrings added — `d742b59f` | 12 commands / 25 fields covered; red-proofed by removing one field's constant, one command's help paragraph, and the walk's own `ForwardRef` resolution |
| Artifact 3a | The corpus column "changed non-zero" includes the `-0.0 → 0.0` row | **Fully adopt** | The `-0.0` row is `default_stringify="changed"`, and its result IS zero | Header is now `changed`; the corpus docstring defines the column as the writer's *not-flattened* class and says the negative zero sits in it — `ab267f68` + PR body | The count guard already derives the row from `PARTITIONS`; the wording is now consistent with it |
| Artifact 3b | The PR body says ADR-0041 and the harness carry counts that were removed | **Fully adopt** | `a9e71ce5` removed them; both now point at `gda.live_numbers` | The red-proof row and the ADR bullet corrected in the PR body | The prose guard reads `live_numbers.__doc__` and the catalog — the only two surfaces that quote counts |
| Artifact 3c | The catalog says "Results are exact" immediately before documenting the negative-zero residual | **Fully adopt** | Same bullet, two sentences apart | "Results carry full precision, with one residual" — `ab267f68` | — |
| Artifact 3d | A "see below" reference points upward | **Fully adopt** | The block sits ~100 lines above `game call`'s paragraph | "see **Live number transport** above" — `ab267f68` | — |

## The one placement I did not take, and why

The direction was to **derive** the result contract rather than repeat it. It is derived on the machine surface: `LIVE_RESULT_PRECISION` is a production constant concatenated into the models' field descriptions, which is what `--schema` and the gda-mcp wire schemas publish (ADR-0004) — no copies. **Rendered help still carries pinned copies** (12 docstrings), and I did not move that into `gda.headless.schema_command_class`, where one hook could have appended the paragraph to every float-bearing live command's `help` (and, through `gda.surface`, to the MCP tool description):

- **Blast radius.** That hook would run the model walk at command-registration time, inside the class every command is built from. The walk calls `model_rebuild()` and refuses an unresolved annotation — today a model that cannot rebuild breaks *its own* command; from that hook it would break `typer.main.get_command(app)`, i.e. `gda --help`, `gda info`, everything. Trading a per-command failure for a CLI-wide one to remove prose duplication is the wrong side of that exchange.
- **The review's own bar is met.** `ARC-770-002` asks for the contract to be "derived **or guarded** across all float-bearing live replies". The *set* is derived — from the models, by the walk — and both surfaces are checked over it. What is duplicated is one sentence, pinned to the single production string, in the one place Typer cannot interpolate a constant.

**Residual risk, stated plainly:** the walk keys on the annotation. `Any` counts as float-bearing (the projected-value fields are `Any`), `int` does not (Godot stringifies integers through `itos`, exactly). A live reply that returned a number inside a **string** field — the projection's `str()` fallback — would not be asked to disclose anything, correctly for JSON transport but worth knowing. Round 2's residual also stands: a recipe that INVENTED a wire number from something other than its params fields would escape the request-side scan; none does today.

## Found while fixing, also handled

- **The walk found a field no review named:** `ScreenCaptureResult.receipt.observed`, the predicate echo on the capture receipt. Its sibling `predicate.observed` was disclosed; the receipt's copy was not.
- **`perf monitors`' window aggregates** (`stats.min/max/mean/p50/p95`, `budget.value/min/max`) are computed by gda over the sampled values, not read off the wire. They are disclosed with wording that says exactly that, so the contract is not claimed for a number Godot never wrote.
- **`ScreenCaptureResult.predicate` / `receipt` are `ForwardRef`s** until the model is first used, so a naive walk over `model_fields` silently skips them. The walk rebuilds and asserts, and the vacuity floor catches the degradation (red-proof below).

## Validation

- `ruff check` / `ruff format --check` clean; `uv run pyright` **0 errors**; `pytest -q -m "not e2e"` **2190 passed, 548 deselected**; `scripts/harness_version_guard.py b616e7fb HEAD` valid. PR CI: every non-skipped check green.
- Real Godot `4.6.3.stable.official.7d41c59c4` + a real `gda-daemon`, serial: live-number transport **5**, game/perf/diag/logger **19**, input/daemon **44**, screen/mcp-live **19**. The input/daemon pair failed once on its first run with the round-2 session-launch contention ("the harness did not connect within 25s") and passed 44/44 on an immediate re-run of the identical pair — reported, not smoothed over.
- **Real-daemon boundary probe** (throwaway project, short `XDG_RUNTIME_DIR`): `daemon wait-ready --timeout 5e-324` now reaches the daemon and expires (`engine_session_not_running`, "the readiness deadline expired…") instead of being refused; `input mouse-move 5e-324 1` still refuses on argv, `--params-json` still reports `invalid_params` with the same sentence; `logger tail --params-json '{"limit": 9007199254740995}'` is accepted and then fails for the real reason. Full transcript in the PR body.
- **Red-proofs (5, each reverted).** Relayed model loses the base → `['input mouse-move']`; daemon-served model gains it → `['daemon wait-ready']` **plus** the accept-side test reproducing the round-3 defect verbatim; one field's constant removed → the walk names it and that command's schema check fails; the walk's `model_rebuild()`/`ForwardRef` assertion removed → the vacuity floor fails at 11 of 12, with `screen capture` silently dropping out; one command's help paragraph deleted → the derived help check fails naming it.

## Not verified

- **Hosted CI runs no Godot e2e**, so the engine evidence above is local only.
- The **5,500-value sweep** percentages remain a point-in-time measurement, labelled as colour in the body; every load-bearing number is a corpus row the e2e re-derives.
- The probe project defines no main scene, so `daemon wait-ready --timeout 25` fails on that rather than on the budget — it is evidence about the refusal, not about a successful launch. The e2e suites above are what exercise real sessions.

qinhaihong-red:

Reviewed exact head `ab267f68ca792ee14a37dd34ebb9c2eb6c7922b9` against base `b616e7fbaf782a44df4a1b24614d7ad02a7446f2`. This is a residue-only recheck of the previous review, the author's Round-3 reply, and the two incremental commits since `f2495423`.

The previous request-boundary defect is closed: `RelayedLiveParams` now applies only to harness-relayed commands, while daemon-served models accept their Python-consumed values. A real daemon received `wait-ready --timeout 5e-324`; relayed mouse-move values remain correctly refused. Coverage also now discovers 12 float-bearing result commands / 25 fields across schema, rendered help, and MCP.

### [P2] Result disclosure confuses field type with writer provenance

The derived walk treats a float-bearing parent description as covering its full subtree. As a result, `perf monitors` statistics and budget bounds inherit the Godot-writer/full-precision/negative-zero sentence even though these values are computed or copied in Python (`src/gda/commands/perf.py:550,562,672,777`).

A real daemon run returned:

```json
{"value": 1.0, "min": -0.0, "max": -0.0}
```

This contradicts the inherited statement that Godot's writer necessarily reads negative zero back as `0.0`. Distinguish harness-origin samples from Python-derived aggregates/bounds, or use a shared disclosure that is accurate for both sources. The guard must verify semantic provenance, not only the presence of a description.

### Additional current residue

1. The PR body still says the branch does not change `commands/input.py` and changes `screen.py` only by six docstring lines; that is stale relative to the current head.
2. `tests/test_live_numbers.py:5-8` still says every live command refuses these values, omitting the daemon-served exception.
3. `tests/live_number_corpus.py:7-10,180-184` claims `PARTITIONS` derives an ADR quote and that tests validate ADR-0041 counts. ADR-0041 deliberately does not contain the counts, and the guard actually reads only `gda.live_numbers.__doc__` and the command catalog.
4. `test_every_live_...` at `tests/test_live_numbers.py:259,306` only covers relayed ingresses; rename it to `test_every_relayed_live_...` so the new boundary is explicit.

Axis result: **Artifact 2 / DDMA 1 / Standards 3 / Spec 1 — Changes required on all four axes.** The Artifact/DDMA/Spec result findings describe the same provenance defect; the remaining Standards findings are synchronization/naming residues.

Validation: 612 focused tests and five real-Godot transport E2E tests passed; 54 input/output/error schemas for all 18 LIVE commands passed Draft 2020-12 validation; actual schema/help/MCP surfaces contain the discovered disclosures; Ruff, format, and `git diff --check` passed. All non-skipped CI checks succeeded; hosted Godot E2E remains skipped.

qinhaihong-red:
## Round 4 — handled. Head `07e89039`

The P2 is real and is the same shape you named: a rule bound to a proxy for the third time. `RelayedLiveParams` fixed `ExecutionKind.LIVE`; the derived walk fixed the hand-picked list; but "reachable under a described parent" is still not "written by the engine". Your repro is exact — `perf monitors --budget` echoes the bounds straight out of the caller's budget file (`perf.py:_load_budgets` → `_evaluate_budgets`), so `-0.0` reaches the result under a published claim that it cannot.

**The fix does not add a fourth proxy: the disclosure now NAMES its writer, and the guard MEASURES which writer a field has.** Two production sentences, each opening with its provenance — `LIVE_ENGINE_PRECISION` ("A value the engine reports crosses the wire at full binary64 precision …") and `LIVE_DERIVED_PRECISION` ("A value gda derives CLI-side in Python … that writer's negative-zero residual does not apply, so a `-0.0` stays `-0.0`").

| [P2] Result disclosure confuses field type with writer provenance | Fully Adopt | `perf.py:_stats_over` computes `mean`; `_evaluate_budgets` copies `rule.min`/`rule.max` from the budget file — neither meets Godot's writer. Reproduced at recipe level: a budget declaring `-0.0` bounds yields a verdict whose bounds are still `-0.0` | Two writer-named sentences + per-field disclosure in `perf.py`; the two container descriptions stop blanketing a mixed subtree — `5a8e77dd`. Guard rebuilt to establish and require the field's writer — `865fdce8` | 2193 unit; 8 new red-proofs below; 5 live-number e2e + 34 live e2e against real Godot 4.6.3 + a real daemon |
| Residue 1 — PR body stale on `input.py` / `screen.py` | Fully Adopt | `git diff b616e7fb..HEAD` shows `input.py` +60/-… (six params models re-based on `RelayedLiveParams`, four argv bodies routed through `params_or_bad_parameter`, four field descriptions) and `screen.py` three field descriptions | PR body edited: the environmental-flake paragraph now states what those diffs actually are and why they do not touch display probing or session launch | Body re-read; `Closes #752` intact |
| Residue 2 — `test_live_numbers.py:5-8` omits the daemon-served exception | Fully Adopt | The module's own last block (`DAEMON_SERVED_INGRESSES`) asserts the opposite of what the docstring claimed | Docstring says RELAYED, names the three daemon-served ops, and points at the block that pins them — `514f34d0` | — |
| Residue 3 — corpus claims an ADR quote the ADR does not carry | Fully Adopt | ADR-0041's outcome note ends "which this note deliberately does not restate"; `test_the_authority_prose_quotes_the_derived_counts` reads `gda.live_numbers.__doc__` and `docs/command-catalog.md` | Both comments now describe the two surfaces the guard actually reads, and say ADR-0041 and the harness comment carry no count — `514f34d0` | — |
| Residue 4 — `test_every_live_…` covers only relayed ingresses | Fully Adopt | Both parametrized lists hold relayed commands only | Renamed to `test_every_relayed_live_argv_ingress_…` / `…_params_json_ingress_…`, with their lists (`RELAYED_ARGV_INGRESSES`, `RELAYED_PARAMS_JSON_INGRESSES`) — `514f34d0` | — |
| Same over-claim one surface further out (found while fixing, not raised) | Adopted | `SKILL.md` said "the one residual is that a negative zero reads back as `0.0`" for every live reply; `CONTEXT.md` said "a live reply is framed with Godot's full-precision JSON writer" | Both scoped to the engine's writer, with the gda-side class named — `5a8e77dd`, `07e89039`; the Skill pin test now requires both halves | Unit |

### The provenance classification you asked for

**The model walk cannot tell the two apart, and I did not pretend otherwise.** A field's writer is a property of the code path that FILLS it, which no annotation carries. Two mechanisms replace the guess, neither of them a declaration:

1. **The dispatch branch** (10 of the 12 commands). A LIVE command with no `recipe` is fulfilled by `classify_live(raw, request, cmd.output_model)` — `gda.dispatch` builds no result, so every float in it was parsed out of the JSON the harness wrote. That is the mechanism itself, not a correlate of it, and `test_the_classify_path_hands_back_the_engines_own_floats` drives it on `game get` with `1e-300` and `3.141592653589793` (both lost by the DEFAULT writer) to show nothing CLI-side re-renders the value.
2. **A sentinel probe** (the 2 recipe-bearing commands). `perf monitors` and `screen capture` assemble their results CLI-side, so `PROVENANCE_PROBES` drives *that descriptor's own recipe* with a reply whose floats are sentinels and reads which fields the sentinels reach. The window probe's four samples are distinct and their mean is not one of them, so a computed aggregate cannot be mistaken for a selected sample. A recipe-bearing float-bearing live command **missing** from that map fails the guard rather than defaulting to the engine's sentence.

25 float-bearing fields, 12 commands:

**Engine-written — 21 fields** (harness `_json`, full precision): `game get` `properties.value`; `game rect` `position`, `size`; `game set` `value`; `game call` `value`; `logger tail` `records.fields`; `perf monitors` `monitors.value`, `stats.min`, `stats.max`, `stats.p50`, `stats.p95`, `samples.values`; `perf monitor` `samples.value`, `emissions.args`; `input mouse-click` `position`; `input mouse-move` `position`; `input action` `strength`; `input tap` `strength`; `screen capture` `predicate.expected`, `predicate.observed`, `receipt.observed`.

Two are worth stating precisely rather than lumping:

- `logger tail` is daemon-served, but its `records.fields` floats are literals the harness's `_json` wrote into the session log; the daemon only `json.loads`/`json.dumps` them, which is exact. Engine.
- `stats.min`/`max`/`p50`/`p95` are SELECTED by Python from the window's rows. Python creates no number there — each is a value the engine wrote — so the engine's sentence, including its negative-zero clause, is true of them.

**gda-derived — 4 fields**, all under `perf monitors`' window mode: `stats.mean` (`sum(ordered) / len(ordered)`), and `budget.value` / `budget.min` / `budget.max`. `budget.value` is measured derived because gda produces the field's content; with `stat: min` the number it puts there happens to be an engine sample, which is why the derived sentence reads "from what the engine reported, **or** from input you supplied" — true in both modes, and its negative-zero clause is a statement about gda's serializer, not a promise that a `-0.0` will appear.

### Red-proof (each from the committed head, reverted after)

| guard | perturbation | observed failure |
| --- | --- | --- |
| the writer binding, on a computed value | `PerfSampleStats.mean` publishes the ENGINE sentence | `{'perf monitors': [('PerfMonitorsResult.stats.mean', 'gda')]}` |
| the writer binding, on an echoed bound | `PerfBudgetVerdict.min` publishes the ENGINE sentence | `{'perf monitors': [('PerfMonitorsResult.budget.min', 'gda')]}` |
| **the round-4 defect itself** | restore the blanket engine sentence on the `stats` PARENT | the same failure, naming `stats.mean` — the exact subtree inheritance that shipped |
| an unmeasured recipe | delete `perf monitors` from `PROVENANCE_PROBES` | `perf monitors assembles its result CLI-side … add a probe to PROVENANCE_PROBES so the sentence it must publish is MEASURED` |
| a probe that cannot discriminate | make the four samples identical, so the mean equals one of them | `[('PerfMonitorsResult.stats.mean', 'engine'), ('PerfMonitorsResult.budget.value', 'engine')]` |
| the derived sentence's coverage | delete the derived paragraph from `perf monitors`' help | `test_only_the_replies_that_carry_a_gda_number_publish_the_derived_contract`: `perf monitors` |
| the derived sentence's BOUND | paste it into `game get`'s help, which returns no gda number | the same test, the other direction: `game get` |
| the negative zero itself | `min=rule.min + 0.0` in `_evaluate_budgets` | `assert [1.0, 1.0] == [-1.0, -1.0]` — the sign you observed, gone |

The last one is the DoD probe pair: a gda-produced field carrying an accurate disclosure AND still carrying `-0.0`, beside an engine-produced field carrying the full-precision claim and the value verbatim.

### Scope decisions, stated rather than buried

- **I did not probe the 10 recipe-less commands.** Their provenance follows from the dispatch branch that fills them, and one probe of that branch is worth ten fixtures of the same shape. Adding a canned reply per command would have moved the maintenance list from `PROVENANCE_PROBES` into a bigger one without measuring anything new.
- **I did not add an e2e for the `-0.0` bound.** Those bounds never cross the wire — the budget file is read CLI-side and the verdict is assembled CLI-side — so a real daemon adds no leg to the evidence. Your daemon run is the field observation; the recipe-level test is the regression.
- **The subtree rule is kept, not removed.** It is what keeps both promises off the shared headless `NodeProperty` (`node get` / `resource get` still have #771's default writer). What changed is that inherited coverage is now checked against the measured writer, so a parent blanketing a mixed subtree is a failure instead of a pass.

### Gates

```
All checks passed!                                    # ruff check .
477 files already formatted                           # ruff format --check .
0 errors, 0 warnings, 0 informations                  # uv run pyright
2193 passed, 548 deselected in 53.87s                 # pytest -q -m "not e2e"
harness version transition is valid                   # scripts/harness_version_guard.py b616e7fb HEAD
```

Live e2e on this head, real Godot `4.6.3.stable.official.7d41c59c4` + a real `gda-daemon`, run serially:

```
 5 passed in  7.63s   # test_e2e_live_number_transport.py
15 passed in 33.55s   # test_e2e_perf.py test_e2e_game.py
19 passed in 33.79s   # test_e2e_screen.py test_e2e_mcp_live.py
```

No flakes this round; all three groups are first-run results.

### Residual risk

`budget.value` is the one field whose number can come from either writer depending on the gated `stat`. It is disclosed as gda's, which is true of how the field is produced and whose serializer emits it, and the sentence covers both origins — but a reader who wants to know whether *that particular* number originated in the engine has to look at `stat`. Splitting the field by mode would buy a mode-dependent schema for a distinction with no observable consequence (both are exact, and neither can be `-0.0` in practice), so I left it.

Issue #775, opened 2026-08-31
One reader for scene text: adapt the four [ext_resource] rescans and three [node] scans onto the canonical entries reader


qinhaihong-red:
## Parent

Found by the `/improve-codebase-architecture` review of promotion PR #773. Companion to #764, which owns the `DirAccess` folder-walk scaffolding — same file, different seam: #764 consolidates how the project TREE is walked, this consolidates how scene TEXT is read.

## What to build

Make the canonical scene-text reader the ONLY reader. `_ext_resource_entries_from_text` (path + id + type, canonicalized, the `id=`-vs-`uid=` substring trap handled) already serves six call sites, including the #721 composed walk and the save-side id restore. Beside it, four functions still recognize an `[ext_resource]` line with their own scan — the dependencies/find-references harvesters, the script-typed filter, and the save-side id substitution — and three more run their own `[node ` header scan. Three of the rescans still extract with the naive quoted-attr helper, which is safe today only because no other ext_resource attribute happens to end in `path=`; the safety lesson the canonical reader learned was never propagated.

Adapt the seven rescans onto the canonical reader — growing it a node-header sibling where one is needed — and delete the rescans. Behavior-preserving by intent; the one behavioral delta this refactor is ALLOWED to carry is the canonical identity the predecessor issue already established for the analysis graph (landing this after it keeps that delta zero).

This makes the predecessor's defect class extinct: with one reader there is no shallow scan left for the next alias or attribute-extraction bug to hide in.

## Acceptance criteria

- [ ] Exactly one implementation recognizes an `[ext_resource]` line; the independent rescans are deleted or reduced to calls through the canonical reader.
- [ ] Node-header recognition likewise has one owner.
- [ ] The WHOLE-NAME attribute rule is applied by the shared reader for path-bearing attributes (the naive quoted-attr substring extraction is the defect it replaces), pinned by a regression that fails on a naive extraction. *(Reworded 2026-09-02 — the original sentence said the substring rule is applied, the opposite of the narrative's intent.)*
- [ ] Existing suites green: scene validate e2e, project-analysis tests, save-side id-restore tests.
- [ ] No public contract changes — schemas, help, and catalog byte-identical.
- [ ] `_resolve_ref_path` and `_normalize_ext_resource_path` — bodies identical but for a parameter name since #774 (PR #777 explicitly deferred the merge here) — converge to one owner, or the PR records why two survive.

## Blocked by

[...]

PR #777, opened 2026-08-31
fix(gda): one canonical identity for the project reference graph (#774)

### LINKED ISSUE #774: project analysis keys its reference graph on raw spellings: find-unused-resources advises deleting an instanced scene
## Parent

Found by the `/improve-codebase-architecture` review of promotion PR #773. Sibling of #721's walk fix: the composed `scene validate` walk canonicalizes every reference through the engine's own `simplify_path` (its `_canonical_resource_path` adapter) precisely because alias spellings double-counted a file — and the project-analysis commands never inherited that identity. Outside the scope of both #763 (Python CLI containment gates) and #764 (the `DirAccess` folder-walk scaffolding).

## What to build

Give the project-analysis group — `gda project dependencies`, `gda project find-references`, `gda project find-unused-resources` — the same canonical resource identity the composed scene-validate walk already uses, on BOTH sides of every comparison: the reference entries harvested from scene and script text, and the caller's query string.

Today the graph is keyed on raw spellings. `_ext_resource_paths` appends the literal `path="..."` attribute untouched, `_resolve_ref_path` returns an already-prefixed reference unchanged, and `find-references` seeds `target_paths` with the query verbatim. The walk's deep adapter (`_canonical_resource_path`, a direct `simplify_path()` call — it cannot drift from the engine) is one call away and unused here.

**Reproduced on the PR #773 head** (Godot 4.6.3; a parent scene instances `res://leaf.tscn` under the spelling `res://sub/../leaf.tscn`):

```
gda project dependencies       → parent depends_on "res://sub/../leaf.tscn"   (a graph node that names no on-disk file)
gda project find-references res://leaf.tscn → references: []                  (the reference is real; the query never matches)
gda project find-unused-resources → unused: ["res://leaf.tscn"]              (advises deleting a scene that is instanced)
```

The last line is wrong advice with a destructive follow-up: an agent that trusts it deletes a live dependency and the project stops loading.

Scope boundary: this issue is the IDENTITY fix only — canonicalize at the harvest and query points. Consolidating the duplicated scene-text scans themselves is the follow-up issue and must not ride in here.

## Acceptance criteria

- [ ] `project dependencies` keys every reference canonically: the aliased declaration above reports `depends_on: res://leaf.tscn`.
- [ ] `find-references` matches regardless of spelling on either side — aliased declaration with canonical query, and canonical declaration with aliased query.
- [ ] `find-unused-resources` does not list a resource that is referenced under any alias spelling.
- [ ] Script-side references (`preload`/`load` with an already-prefixed aliased address) receive the same identity.
- [ ] Each fixed case is red-proofed: the regression fails on the pre-fix code.
- [ ] Real-engine e2e covers the three commands against one aliased-declaration project.

## Blocked by

None - can start immediately.

qinhaihong-red:
Closes #774

## What was wrong

The three project-analysis graph reads keyed their shared reference graph on the **raw spelling** each declaration used, while the walk that supplies the other side of every comparison builds its paths by joining directory entries — so the two sides could never meet on an aliased declaration.

Reproduced on this PR's base (`9122e03e`, Godot 4.6.3), a parent scene instancing `res://leaf.tscn` under the spelling `res://sub/../leaf.tscn`:

```
project dependencies       → parent depends_on "res://sub/../leaf.tscn"   (a node no file on disk answers to)
find-references res://leaf.tscn → references: []                          (the reference is real; the query never matches)
find-unused-resources      → unused: ["res://leaf.tscn"]                  (advises deleting a scene that is instanced)
```

## What changed

Every path that enters the graph — and the caller's query — now passes through `_canonical_resource_path`, the deep adapter the composed `scene validate` walk already uses (a direct `simplify_path()`, so it cannot drift from the engine). Both sides of every comparison carry one identity.

The harvest and query sites, all in `src/gda/ops/operations.gd`:

| site | was |
| --- | --- |
| `_ext_resource_paths` | appended the literal `path="…"` attribute |
| `_resolve_ref_path` | returned an already-prefixed reference unchanged |
| `_collect_references_from`'s `[ext_resource]` scan | compared the declared spelling |
| `_op_project_find_references`' `target_paths` seeding | used the caller's query verbatim |
| `_main_scene_path` / `_project_autoloads` | returned the literal `project.godot` value |

**The last row was not in the reported repro but carries the same defect on the project's entry points.** Reproduced separately: an aliased `run/main_scene="res://sub/../parent.tscn"` and an aliased autoload made `find-unused-resources` report the **main scene and the autoload scene themselves** as deletable, and `find-references res://parent.tscn` return empty. Same graph, same one-call fix, same destructive advice — so it is fixed here and red-proofed with the rest. The `*` enable marker comes off an autoload value **before** the fold: `simplify_path` does not recognize a scheme behind it and folds `*res://a/../b.gd` to the broken `*res:/b.gd` (measured on 4.6.3).

Two deliberate non-changes:

- The echoed `target` keeps the caller's own spelling — only the matching is canonical, and a `class_name` target is no path at all.
- `project statistics` reads the autoloads through the same accessor, so the paths it lists are canonical too. Identical output for any normally-spelled project; documented in the catalog.

**Scope boundary honored.** Identity only — the duplicated scene-text scans themselves are untouched; that is #775.

## Published contract

`docs/command-catalog.md`'s static-analysis section said nothing about spelling identity. It now states the alias-invariance as a contract: one graph node per file whatever a declaration spells it, matching in both directions, and the `target`-echo/`statistics` caveats. `SKILL.md` needed nothing — its `project` row is a bare command list, and the README rows are one-line descriptions that make no spelling claim.

## Stale comment corrected

The section note near `_op_project_find_references` named a `_scan_project` function that does not exist and claimed "All four share one reference graph". Both are now true: there are **two** walks (`_collect_resource_paths` and `_collect_all_file_paths`), and the graph is shared by **three** reads — `statistics` is not a graph read. The consistency claim itself is what this PR makes true again.

## Validation

Gates green on `a1297a33`:

- `uv run pytest -m "not e2e"` → **2238 passed**
- `uv run ruff check .` → clean · `uv run ruff format --check .` → clean · `uv run pyright` → **0 errors**

Real-engine e2e (Godot 4.6.3, macOS), 5 new tests in `tests/test_e2e_project_analysis.py` against one `alias_project` fixture whose references all name files the engine really resolves — only the spelling is aliased:

| test | covers |
| --- | --- |
| `test_dependencies_keys_an_aliased_declaration_canonically` | aliased `[ext_resource]` **and** aliased `preload` report the resolved path; no graph node names a `..` component |
| `test_find_references_matches_an_aliased_declaration_from_a_canonical_query` | direction 1, scene + script |
| `test_find_references_matches_a_canonical_declaration_from_an_aliased_query` | direction 2 |
| `test_find_references_matches_an_aliased_project_level_entry` | `project.godot` main scene + autoload |
| `test_find_unused_resources_never_lists_an_aliased_reference` | the instanced scene and both entry points are not listed; the real orphan still is; the unused ⇔ empty-find-references criterion still holds |

**Red-proofed**: with `src/gda/ops/operations.gd` reverted to the pre-fix source and the tests unchanged, all 5 fail. Pre-fix `find-unused-resources` returned `['res://canonical.tscn', 'res://hud.tscn', 'res://leaf.tscn', 'res://parent.tscn', 'res://sub/orphan.tres']` — the instanced scene, the autoload scene and the main scene all offered for deletion.

Regression sweep, run serially: `test_e2e_project_analysis` (20), plus `test_e2e_project`, `test_e2e_script`, `test_e2e_classname_resolution`, `test_e2e_scene_validate`, `test_e2e_scene` → **191 e2e passed**. That set covers every consumer of the five changed functions, including the script attach/save missing-dependency guard that shares `_resolve_ref_path`.

### DIFF
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -902,6 +902,21 @@ traversal itself: the two walks are one recursive walker asked two different acc
 questions. A shared traversal is not a shared universe — `project statistics` keeps
 counting the sidecars and the project file the other walk will never see.
 
+**One graph node per file, whatever a declaration spells it (#774).** A `res://` address has
+many lexical spellings for one file, and the three graph reads key on the file the engine
+resolves rather than on the spelling: every harvested path — an `[ext_resource]` line, a
+`preload()`/`load()`/`extends` argument, and `project.godot`'s own main scene and autoload
+entries — is folded through the engine's `simplify_path`, and so is the `find-references`
+target. `res://sub/../leaf.tscn` and `res://leaf.tscn` are therefore ONE node:
+`dependencies` reports the resolved path, `find-references` matches whichever side is
+aliased (aliased declaration with a canonical query, and the reverse), and
+`find-unused-resources` never reports a resource that something references under an alias.
+`project statistics` reads `project.godot` through the same accessor, so the autoload paths
+it lists are canonical too. The echoed `target` keeps the caller's own spelling — only the
+matching is canonical, and a `class_name` target is no path at all. Keying on the raw
+spellings broke all three at once: `find-unused-resources` listed an instanced scene, which
+is wrong advice with a destructive follow-up.
+
 ---
 
 ## Phase 2 — live domain commands (served by `gda-daemon`)
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -902,20 +902,31 @@ traversal itself: the two walks are one recursive walker asked two different acc
 questions. A shared traversal is not a shared universe — `project statistics` keeps
 counting the sidecars and the project file the other walk will never see.
 
-**One graph node per file, whatever a declaration spells it (#774).** A `res://` address has
-many lexical spellings for one file, and the three graph reads key on the file the engine
-resolves rather than on the spelling: every harvested path — an `[ext_resource]` line, a
-`preload()`/`load()`/`extends` argument, and `project.godot`'s own main scene and autoload
-entries — is folded through the engine's `simplify_path`, and so is the `find-references`
-target. `res://sub/../leaf.tscn` and `res://leaf.tscn` are therefore ONE node:
-`dependencies` reports the resolved path, `find-references` matches whichever side is
-aliased (aliased declaration with a canonical query, and the reverse), and
-`find-unused-resources` never reports a resource that something references under an alias.
-`project statistics` reads `project.godot` through the same accessor, so the autoload paths
-it lists are canonical too. The echoed `target` keeps the caller's own spelling — only the
-matching is canonical, and a `class_name` target is no path at all. Keying on the raw
-spellings broke all three at once: `find-unused-resources` listed an instanced scene, which
-is wrong advice with a destructive follow-up.
+**One graph node per file, whatever a declaration spells it (#774).** One file has many
+spellings, and the three graph reads key on the file the engine resolves rather than on the
+spelling. Two spellings reach the same file, and each needs its own fold:
+
+- an **alias** of an absolute address — `res://sub/../leaf.tscn` for `res://leaf.tscn` —
+  which the engine's `simplify_path` folds;
+- a **relative** address, which the engine resolves against the directory of the file that
+  DECLARES it. `path="../shared/leaf.tscn"` in `res://scenes/main.tscn` loads
+  `res://shared/leaf.tscn`, and `preload("../shared/x.gd")` in `res://scripts/user.gd` loads
+  `res://shared/x.gd`. Such a path is anchored to that directory first, then simplified.
+
+Every harvested path is folded this way — an `[ext_resource]` line and a
+`preload()`/`load()`/`extends` argument by the rule above; `project.godot`'s main scene and
+autoload entries, and the `find-references` target, are absolute already and are only
+simplified. So `dependencies` reports the resolved path (never a bare `../…` that names no
+file), `find-references` matches whichever side is aliased or relative (a folded declaration
+with a canonical query, and the reverse), and `find-unused-resources` never reports a
+resource that something references under another spelling. `project statistics` reads
+`project.godot` through the same accessor, so the autoload paths it lists are canonical too.
+
+The echoed `target` keeps the caller's own spelling, and each reference's `context` keeps the
+line as written — only the matching is folded, so an agent still sees the text it must edit.
+A `class_name` target is no path at all. Keying on the raw spelling broke all three reads at
+once: `find-unused-resources` listed an instanced scene, which is wrong advice with a
+destructive follow-up.
 
 ---
 
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -3944,18 +3944,30 @@ func _write_text_file(path: String, source: String, noun: String) -> bool:
 # would return empty — the same graph, one truth. statistics is not on that graph:
 # it only counts what its own scan reaches.
 #
-# ONE graph needs ONE identity per file. Every path that enters it — harvested
-# from an [ext_resource] line, from a preload/load/extends argument, or from
-# project.godot's main scene and autoloads — AND the caller's find-references
-# query pass through _canonical_resource_path, the engine's own simplify_path. A
-# file is one node however a declaration spells it: res://leaf.tscn and
-# res://sub/../leaf.tscn are the same node, not two. Keying on the raw spellings
-# broke all three reads at once (#774): dependencies named a node no file on disk
-# answers to, find-references for the canonical path missed the aliased
-# declaration, and find-unused-resources therefore advised DELETING a scene the
-# project instances — wrong advice with a destructive follow-up. The other side
-# of every comparison is canonical by construction: the walk builds each path by
-# joining directory entries, never by echoing a declaration.
+# ONE graph needs ONE identity per file, and that identity is the path the ENGINE
+# resolves a declaration to — not the string the declaration spells. Two spellings
+# reach the same file:
+#
+# - an ALIAS of an absolute address (res://leaf.tscn and res://sub/../leaf.tscn),
+#   folded by _canonical_resource_path, the engine's own simplify_path;
+# - a RELATIVE address, which the engine resolves against the DECLARING file's
+#   own directory — `path="../shared/leaf.tscn"` in res://scenes/main.tscn loads
+#   res://shared/leaf.tscn (measured on 4.6.3), and `preload("../shared/x.gd")`
+#   in res://scripts/user.gd loads res://shared/x.gd (measured likewise).
+#
+# So every path that enters the graph is folded by the owner that knows its base
+# directory: _normalize_ext_resource_path for an [ext_resource] line,
+# _resolve_ref_path for a preload/load/extends argument. Only two entrants have no
+# declaring file to anchor to and are absolute by construction — project.godot's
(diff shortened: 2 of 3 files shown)


qinhaihong-red:
## Adversarial review — exact head a1297a33

Reviewed issue #774 through its implementation PR #777, pinned to base 9122e03e and head a1297a333eb603207812546af6d4009ffc49d944. The four axes below were run independently; their findings have not been merged or re-ranked.

### Artifact review — Pass

The command catalog correctly defines lexical res:// alias canonicalization, caller-spelling preservation, and canonical graph identity. The issue and changed prose describe the intended contract accurately.

**Terminology and prose assessment:** Sources were root CONTEXT.md, command catalog, ADR-0032, issue #774, and the exact canonical-path authority. I reviewed the complete current command catalog, all changed prose, issue/PR text, and every claim about identity, declaration spelling, query output, and scan scope for term stability, boundedness, causal/scope relations, usability, and ASD-STE100-style English. No substantive artifact defect was found. Full claimed suites and Linux/Windows validation were not independently run; hosted Godot E2E is skipped.

### Domain modular architecture — Changes requested

**[P1] Route relative ext_resource paths through the existing base-aware graph-identity owner.**

The existing authority is _normalize_ext_resource_path(ref_path, base_dir) at src/gda/ops/operations.gd:6510-6513: it resolves a reference relative to its declaring file and then canonicalizes it. The two changed scene-text ingresses bypass that owner and call only _canonical_resource_path(ref), at lines 4375-4377 and 4436-4441.

For a valid scenes/main.tscn declaration path="../shared/leaf.tscn", that creates a graph node named ../shared/leaf.tscn instead of the walked identity res://shared/leaf.tscn. The shared graph is therefore incomplete at one ingress, and its three consumers disagree. This is an owner/completeness/DRY defect, not a request for a new abstraction.

Smallest repair: call _normalize_ext_resource_path(ref, path.get_base_dir()) at both sites and add one nested relative-ext-resource regression across all three consumers. Claim ARCH-777-CANONICAL-GRAPH-IDENTITY remains open; the counterexample meets its disconfirming condition. UID/symlink identity remains outside #774.

### Standards — Changes requested

**[Hard] Relative ext_resource references bypass the file's normalization authority.**

This is a RULES.md:119 single-authority violation and, as a judgement label on the same defect, Fowler Duplicated Code: two graph paths repeat the same incomplete transform instead of using the existing base-aware owner. The smallest repair is the same two-call substitution above plus real-engine regression coverage.

### Spec — Changes required

Issue #774 requires the same canonical identity on both sides of every comparison, including references harvested from scene/script text. It also requires find-references to match regardless of spelling and find-unused-resources not to list a referenced resource under an alias spelling. This valid relative scene reference violates all three requirements.

### Reproduction and validation

Using Godot 4.6.3 and a project whose main scene declares ../shared/leaf.tscn:

- scene get resolves the instance as res://shared/leaf.tscn;
- project dependencies emits ../shared/leaf.tscn;
- project find-references res://shared/leaf.tscn returns no references;
- project find-unused-resources wrongly returns res://shared/leaf.tscn.

The existing 20 relevant E2E tests pass, but they cover prefixed aliases/canonical declarations, not this engine-valid relative spelling. Ruff check/format, Pyright, and executed CI checks are green; hosted Godot E2E is skipped.

qinhaihong-red:
## Review handled — new head `e017f54e`

Baseline: reviewed head `a1297a33`, base tip `9122e03e`. One finding, reproduced on the exact head before any change, then fixed.

| **[P1]** Relative `[ext_resource]` paths bypass the base-aware graph-identity owner (all four axes: architecture / standards / spec) | **Fully adopt**, exactly the smallest repair proposed | Reproduced on Godot 4.6.3 — see below | `e017f54e` | 21/21 project-analysis e2e; red-proofed |
| Merge `_resolve_ref_path` with `_normalize_ext_resource_path` (identical bodies) | **Not adopted here** — stays #775 | The reviewer scoped it out too; the previous commit already declared identity-only scope | none | — |

### Reproduction on the exact head

A project whose `res://scenes/main.tscn` declares `path="../shared/leaf.tscn"`, with **no uid on either scene**, so the engine must resolve by path alone:

| Read | Head `a1297a33` | Engine truth |
| --- | --- | --- |
| `scene get res://scenes/main.tscn` | `instance_path: res://shared/leaf.tscn`, `resolved` | — |
| `scene validate res://parent.tscn` | `valid: true` | — |
| `project dependencies` | `depends_on: '../shared/leaf.tscn'` | names no graph node |
| `project find-references res://shared/leaf.tscn` | `[]` | should name `main.tscn` |
| `project find-unused-resources` | `['res://shared/leaf.tscn']` | instanced — not deletable |

`scene validate` passing while the graph reads fail is the defect in one line: the two halves of `gda` disagreed because only one used the base-aware owner.

A second fixture made it sharper still — a **main scene** instancing `scenes/nested.tscn` (relative, no `..`), which instances `../shared/deep.tscn`. Pre-fix `find-unused-resources` returned **both**, advising deletion of the project's own main-scene child.

### Fix

`_normalize_ext_resource_path(ref, path.get_base_dir())` at both ingresses — `_ext_resource_paths` and `_collect_references_from`'s inline scan. No new abstraction; the existing owner (`operations.gd:6510`, already used by the scene-validate walk) now serves the harvest side and the match side alike. All three reads now agree with the engine on both fixtures, and `find-unused-resources` returns `[]`.

### Script-side ingress — measured, no hole

You asked whether `_resolve_ref_path` (`preload`/`load`) has the same gap. Measured, not inferred:

- Engine: `preload("../shared/helper.gd")` from `res://scripts/user.gd` loads `res://shared/helper.gd` — `script run` exit 0, printed `helper-ok`.
- `gda` already agrees: `dependencies` reported `res://shared/helper.gd` (kind `preload`) and `find-references res://shared/helper.gd` matched `res://scripts/user.gd`.

`_resolve_ref_path` already joins `base_dir` for a relative argument, so the hole was **scene-text only**. Left unchanged.

### Regression + red proof

Extended the existing `alias_project` fixture with both relative flavors — `scenes/nested.tscn` (from the root, no `..`) and `../shared/deep.tscn` (from `res://scenes/`) — asserted across all three consumers, plus a new `test_find_references_matches_a_relative_declaration`. The `dependencies` guard gained an `all(startswith("res://"))` arm, because the prefix-less spelling carries no `..` for the existing guard to catch.

Red proof: reverting **only** `operations.gd` to `a1297a33` and keeping the new tests fails exactly the three consumers —

```
FAILED test_dependencies_keys_an_aliased_declaration_canonically
FAILED test_find_references_matches_a_relative_declaration
FAILED test_find_unused_resources_never_lists_an_aliased_reference
3 failed, 3 passed
```

— while the pre-existing alias arms stay green, so the new coverage is specific to this defect.

### Docs

The catalog paragraph claimed every harvested path "is folded through the engine's `simplify_path`", which was the incomplete rule. It now states both folds (alias → simplify; relative → anchor to the declaring file's directory, then simplify), names which entrants are absolute by construction, and records that each reference's `context` keeps the line as written. Same correction in the `operations.gd` section note.

### Validation

- `pytest -m "not e2e"` — 2238 passed
- `ruff check` / `ruff format --check` — clean; `pyright` — 0 errors
- Targeted e2e (local, Godot 4.6.3): `project_analysis` + `project` + `scene_validate` + `scene` + `classname_resolution` — **124 passed**

**Residual risk / not verified:** validated on macOS + Godot 4.6.3 only; no Linux or Windows run. UID and symlink identity remain out of scope for #774, as you noted. The duplicated function body stays open as #775.

Issue #772, opened 2026-08-28
node set / game set --value: String.to_float() silently changes or zeroes a written float


qinhaihong-red:
## Parent

Found by the #752 slice while building its differential corpus, reported rather than folded in.

## What to build

The WRITE side of the same number problem. `String.to_float()` — the coercion `node set` and
`game set --value` use to turn a caller's textual value into a Godot float — is the engine's
`built_in_strtod` reached through a string, and it loses values the same way the request path did.

Measured against a real engine over #752's 5500-value random sweep: **67.5% exact, 1712 changed,
75 flattened to `0.0`.** A caller writes a value and the project stores a different one, silently.

#752 decided the LIVE argument path by refusing what the parser cannot represent — a value whose
applied decimal exponent is −309 or below is divided by an infinite power of ten and arrives as
`0.0`, and no decimal spelling avoids it. The same reasoning applies here, but the decision is not
automatic: this path coerces from a string rather than parsing a JSON number, it serves a different
command set, and refusing on a write is a different contract from refusing on a call argument.

The coercion block is **mirrored between `gda_harness.gd` and `operations.gd`**, so this needs an
owner who can edit both and keep them in lockstep — that mirroring is itself part of the problem
to weigh.

## Acceptance criteria

- [ ] A decision, recorded, on what a write does with a value `String.to_float()` cannot represent:
      refuse it the way #752 refuses a live call argument, or accept it with the loss disclosed.
      State why the answer matches or differs from #752's, rather than inheriting it silently.
- [ ] The two mirrored coercion sites agree by construction after the change, or the PR states why
      they must stay separate.
- [ ] Reuse `tests/live_number_corpus.py` for the write direction instead of writing a second corpus.
- [ ] A real-engine regression: a value in the affected class round-trips through `node set` then a
      read, and the result is either exact or a refusal — never a silently different number.

## Blocked by

- #752 (its corpus and its refusal precedent are the inputs).

PR #780, opened 2026-08-31
fix(gda): report headless float reads at full binary64 precision (#771)

### LINKED ISSUE #771: Headless value reads flatten small floats and round ordinary ones — reporting, not storage
## Parent

Found by the #752 slice while fixing the LIVE result path, reported rather than folded in.
Independently reproduced on `main` by the orchestrator.

## What to build

Every headless read that projects a float loses precision, and small values are silently reported
as `0.0`. Reproduced on `main` (no #752 code involved), on a node with
`@export var tiny := 1e-300` and `@export var precise := 3.141592653589793`:

```
gda scene get-exports e.tscn --project <p> --json
  {"name":"tiny",    "type":"float", "value":0.0}
  {"name":"precise", "type":"float", "value":3.14159265358979}
```

The caller is handed a number the project does not hold, with nothing marking it as approximate.

**Root cause, same as #752's result half.** `operations.gd` frames its result with the DEFAULT
`JSON.stringify`, which renders a float through Godot's fixed-point path capped at 32 decimals —
so everything below roughly `1e-32.6` flattens to `0.0` and ordinary values round to about 15
significant digits. #752 fixed exactly this on the live side by framing with the engine's
full-precision writer instead; the headless side was out of that slice's scope and still carries it.

The fix is the same one argument, applied where `operations.gd` frames its reply. It reaches every
headless read at once — `node get`, `scene get-exports`, `project list`, `resource get`, and any
other `Value projection` carrying a float — so it is a contract change to the headless result ABI
and needs its evidence, not a quiet edit.

Measured scale, from #752's corpus against a real engine: over a 5500-value random sweep the
default writer returned 33.1% exact, 23.9% flattened to `0.0`, and 42.9% otherwise changed; the
full-precision writer returned 5500/5500 exact.

## Acceptance criteria

- [ ] A headless read reports a float as the exact binary64 the project holds, for a normal value,
      a `DBL_MIN`-scale value, and a subnormal, on both signs.
- [ ] The known residual is stated rather than hidden: a negative zero reads back as `0.0`, decided
      by the engine before the writer's precision argument applies.
- [ ] The corpus #752 committed (`tests/live_number_corpus.py`) is reused for the headless direction
      rather than a second corpus being written — one authority for what the engine does with a number.
- [ ] Every surface that documents the live-side guarantee is checked for whether it now also holds
      headless, and the ones that do are updated in the same change: command catalog, `CONTEXT.md`'s
      `Value projection` entry, the bundled Skill.
- [ ] Red-proofed: reverting the writer argument reproduces `0.0` for a small value through a real
      headless run, not only in a unit test.

## Scope is wider than `scene get-exports`, and it is REPORTING, not storage

Widened 2026-08-28 after PR #767 traced the blast radius and the orchestrator verified it.

**The stored value is correct. Only the projection into JSON is lossy.** After
`gda project set physics/2d/default_gravity --value 3.141592653589793`, the file holds
`2d/default_gravity=3.141592653589793` exactly — while the command's own JSON reply says
`3.14159265358979`. Nobody's project data is being corrupted; callers are being shown a number that
is not the one gda holds. Say this plainly wherever this issue is summarised, or the next reader
will assume the write is lossy and escalate accordingly.

**`project set` breaks its own round-trip claim**, and a caller meets it before `scene get-exports`:

```
gda project set physics/2d/default_gravity --value 3.141592653589793 --json
  -> {"type":"float","value":3.14159265358979}      # the reply already differs from the argument
gda project get physics/2d/default_gravity --json
  -> {"type":"float","value":3.14159265358979}
gda project set physics/2d/default_gravity --value 1e-300 --json
  -> {"type":"float","value":0.0}
```

**Documentation claims this falsifies** — listed for repair in the same change, and deliberately not
rewritten by the slices that found them, because this issue owns the fact:

| Where | The claim |
|---|---|
| `docs/command-catalog.md` (`project set`) | "reports the coerced value in the same JSON projection `project get` uses, so a `set` round-trips through a `get`" |
| `docs/command-catalog.md` (`project list`) | "Each entry reuses the same `{setting, type, value}` projection … so a listed entry round-trips through `project get`" |
| `docs/command-catalog.md` (`scene get-exports`) | "`value` is the export's current value in the same JSON projection `node get` uses" |
| `docs/command-catalog.md` (`project get`) | "its JSON projection, the same projection `node get` reports for a node property" |
| `CONTEXT.md` (`Value projection`) | "…numeric FIDELITY is not yet uniform" — #752 (PR #770, `3a8606cd`) BOUNDED the old claim and named this issue as owner of the headless half; when this issue changes the headless writer, the entry's fidelity outcome must be updated in the same change (read side becomes uniform; the write-side split remains #772's). *(Corrected 2026-08-31, PR #780 review: an earlier strikethrough said "nothing left for this issue to do there", contradicting the AC above.)* |

The four catalog sentences AND the glossary row's fidelity outcome are the doc scope. Note the ORDER a caller meets them: `project set` comes
first and makes the strongest promise ("a `set` round-trips through a `get`"), and it is false in
both directions — `set --value 3.141592653589793` returns `3.14159265358979` from the `set` itself
and again from the `get`, and `1e-300` returns `0.0` from both. Repair that one first.

The glossary line that was the sharpest conflict was only BOUNDED by #752 (the row above): the fidelity outcome still changes with this issue's writer fix, in the same PR.

## Blocked by

- #752 (it establishes the writer choice, the corpus and the disclosure wording this reuses).

qinhaihong-red:
## What this fixes

Every headless read that projects a float reported a number the project does not hold.
`ops/operations.gd` framed its reply with the DEFAULT `JSON.stringify`, which renders a
float fixed-point with at most 32 decimals: values below ~`1e-32.6` came back as `0.0`
and ordinary ones were rounded to ~15 significant digits — with nothing marking the
result approximate.

**It is REPORTING, not storage.** After `gda project set physics/2d/default_gravity
--value 3.141592653589793` the file held `2d/default_gravity=3.141592653589793` exactly,
while the command's own JSON reply said `3.14159265358979`. No project data was ever
corrupted; callers were shown the wrong number.

## The change

One argument, where the payload frames its reply — the same one #752 applied to the live
harness, on the same engine function (`String::num_scientific` instead of `String::num`).
`operations.gd` gains a `_json()` helper mirroring `gda_harness.gd`'s, used by both
`_succeed` and `_fail`, so the fix reaches `node get`, `scene get-exports`,
`project get` / `project list`, `resource get` and every `set` echo at once.

```
before: gda project set physics/2d/default_gravity --value 3.141592653589793 --json
        -> {"type":"float","value":3.14159265358979}
        gda project set physics/2d/default_gravity --value 1e-300 --json
        -> {"type":"float","value":0.0}

after:  -> {"type":"float","value":3.141592653589793}
        -> {"type":"float","value":1e-300}
```

**One residual is disclosed, not hidden**, and it is the live side's residual too: a
NEGATIVE ZERO reads back as `0.0`, which `JSON::_stringify` decides (it returns `"0.0"`
for anything equal to zero) before the precision argument applies.

**The write half is out of scope.** What a `--value` STRING becomes through the engine's
own parser (`String.to_float`) is #772; every claim in this PR is read-side, and the docs
say so where a round-trip is promised.

## Evidence

`tests/test_e2e_headless_number_reads.py` (new, real engine):

- All 96 rows of `tests/live_number_corpus.py` are placed in a project as **IEEE-754
  bytes** and restored in `_init` — the only placement that is not itself a decimal
  literal, and the only one that can hold `DBL_MIN` or a subnormal at all, since no
  literal can deliver those through this engine's parser. Each is read back through
  `scene get-exports` and compared bit for bit, and the `full_precision` partition is
  re-derived from the reply rather than transcribed.
- `node get` and `scene get-exports` must report one value, not two.
- The negative-zero residual is pinned as the ONLY row the reply changes.
- The issue's own `project set` -> `project get` -> `project list` round-trip, for both
  literals it quotes.

**Red-proof.** Reverting the writer argument fails 3 of the 4 new tests and reproduces
the defect through a real engine:

```
project set --value 3.141592653589793 -> {"type":"float","value":3.14159265358979}
project set --value 1e-300            -> {"type":"float","value":0.0}
scene get-exports (bit-exact corpus)  -> [0.0, 0.0, -0.0, 3.14159265358979, 0.0]
```

The new `_json` drift guard in `tests/test_harness_coercion_mirror.py` was red-proofed the
same way.

**Corpus: reused, not forked.** The headless direction needs no column and gets none — the
writer under test is the same engine function #752 measured, so a second recording would be
a second opinion about one function. The verdict helpers (`value_bits`, `stringify_outcome`,
`tally_outcome`) moved beside the corpus so both e2e tiers read one definition of "exact",
and `PARTITIONS` keeps deriving every published count from the rows.

**Gates:** pytest non-e2e 2239 passed; ruff check / ruff format --check clean; pyright 0
errors. Real-engine e2e run locally: the 4 new tests, plus `test_e2e_live_number_transport`
(5), `node` / `scene_exports` / `project` / `resource` / `object_property` (176), and the
remaining headless suites — scene, preflight, security, validate, script, script_run,
op_robustness, params_json, project_analysis, resource_import, asset_file,
classname_resolution, info, user_data (282). All green.

## Published contracts, in the same change

- `docs/command-catalog.md`: a new **"Number reporting"** paragraph under `node` owns the
  fact (read half guaranteed, write half explicitly not), and the four round-trip sentences
  the issue lists are repaired — `project set` first, since it makes the strongest promise
  and a caller meets it before the others. The Phase-2 bullet that declared headless reads
  uncovered is rewritten.
- `CONTEXT.md` (`Value projection`): the fidelity split is corrected precisely — the read
  side is now uniform across both channels with one shared residual, while the WRITE sides
  still differ (the live wire refuses; headless coerces with no refusal, #772).
- `src/gda/skill/SKILL.md`: the headless section gains the counterpart of the live
  paragraph, including the un-bounded `--value` warning.
- `gda.live_numbers` (the authority) now speaks for both channels, and records the decision
  the issue asked for: `LIVE_ENGINE_PRECISION` is **not** shared onto headless fields. Its
  sentence names the live WIRE — a leg a headless reply never crosses — and it is attached
  per-field by a guard that walks the LIVE result models; sharing the string would put a
  wire claim on a headless read, and a headless twin would be the near-duplicate that module
  exists to prevent. Publishing it in `--schema` needs the same derived discipline (a walk
  over the headless float-bearing fields), which is a slice of its own.

Two stale comments in `tests/test_live_contract_guards.py` that justified a live-only claim
by "#771 leaves headless fidelity different" are corrected to the reason that still holds.

## Follow-up (not in this PR)

Publishing the headless precision contract per-field in `--help` / `--schema`, guarded by a
derived walk over headless float-bearing fields, the way the live side does.

Closes #771

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -188,14 +188,18 @@ fallback** for anything else. One projection shared across **every value gda
 emits** — the `get` reads (`project`/`node`/`resource get`), the value echoed by
 `node set`/`resource set`, the per-entry value of `project list` and `scene
 get-exports`, and the live `game get` read — so a value reads the same
-everywhere. That sameness is of SHAPE; numeric FIDELITY is not yet uniform.
-The harness frames every reply with Godot's full-precision JSON writer, so a
-LIVE projected float crosses exactly — the one residual being that a negative
-zero reads back as `0.0` — while the headless writer still flattens small
-floats to `0.0` and rounds ordinary ones (#771). The write-side mirror on the
-live wire is a refusal: a float Godot's parser would read as `0.0` is rejected
-before a request is relayed to the harness, no decimal literal being able to
-deliver it (#752). Two controls keep the shared projection safe on the live
+everywhere. That sameness covers numeric FIDELITY as well as shape: both of
+gda's engine-side payloads frame their reply with Godot's full-precision JSON
+writer — the harness since #752, the headless operations payload since #771 —
+so a projected float is the exact binary64 the subject holds, on either
+channel, with one shared residual the engine decides before the writer is
+consulted: a negative zero reads back as `0.0`. The WRITE sides still differ.
+On the live wire it is a refusal: a float Godot's parser would read as `0.0` is
+rejected before a request is relayed to the harness, no decimal literal being
+able to deliver it (#752). Headless there is no refusal — a `--value` string is
+coerced by that same parser, which still drops the low digits of a
+many-digit literal and reads a `DBL_MIN`-scale one as `0.0` (#772). Two
+controls keep the shared projection safe on the live
 side: the whitelist bounds the Object classes whose storage properties the
 inline kind emits, and the texture kind is safe by construction — a fixed
 getter shape with its one expensive readback behind the explicit digest opt-in
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -89,10 +89,10 @@ address by (`.` for the root) — the `@export` properties the node's attached s
 each a `{name, type, hint, hint_string, value}` entry, where `type` is the export's declared
 Godot type name, `hint`/`hint_string` are the Godot `PropertyHint` enum value and its companion
 string the `@export` annotation produced, and `value` is the export's current value in the same
-JSON projection `node get` uses (its default on a freshly-loaded node) — the property-value
-introspection is reused from `node get` (#55), not re-implemented. An export is detected by its
-usage flags (`PROPERTY_USAGE_SCRIPT_VARIABLE` + `PROPERTY_USAGE_EDITOR`) on the script's own
-property list, so a node's inherited engine properties and a script's plain (non-`@export`)
+JSON projection `node get` uses (its default on a freshly-loaded node), at the same full binary64
+precision (#771) — the property-value introspection is reused from `node get` (#55), not
+re-implemented. An export is detected by its usage flags
+(`PROPERTY_USAGE_SCRIPT_VARIABLE` + `PROPERTY_USAGE_EDITOR`) on the script's own property list, so a node's inherited engine properties and a script's plain (non-`@export`)
 variables are excluded. A node with no script, or whose script declares no exports, is omitted;
 a scene with no exported variables anywhere is a valid, empty listing (`nodes == []`). Like
 `scene get`, get-exports reuses the shared load-failure ladder — a missing file is
@@ -404,6 +404,30 @@ callers can use a follow-up `game get` for domain-specific side effects. Its fai
 LIVE-category `live_unknown_property` / `live_uncoercible_value` (the harness reports them in-band,
 mapped by the live classifier — see
 [Phase 2 — live domain commands](#phase-2--live-domain-commands-served-by-gda-daemon)).
+
+**Number reporting** (#771) — the READ half, shared by every headless command that projects a
+value. `ops/operations.gd` frames each reply with Godot's **full-precision** JSON writer, so a
+projected float is the exact binary64 the project holds: `node get`, `scene get-exports`,
+`project get` / `project list`, `resource get`, and the value each `set` echoes all report
+`1e-300` as `1e-300` and `3.141592653589793` as `3.141592653589793`. They did not before #771 —
+the default writer formats fixed-point with at most 32 decimals, so it reported everything below
+about `1e-32.6` as `0.0` and rounded ordinary values to ~15 significant digits, handing the caller
+a number the project does not hold with nothing marking it approximate. The stored value was never
+the problem: `project.godot` held `3.141592653589793` while the reply said `3.14159265358979`.
+**One residual is disclosed rather than fixed**, and it is the same one the live replies carry: a
+NEGATIVE ZERO reads back as `0.0`, which the engine decides (`JSON::_stringify` returns `"0.0"`
+for anything equal to zero) before the precision argument applies. It is the same engine writer
+the live harness uses, so it is measured against the same corpus and the same partition —
+`gda.live_numbers` is the authority, and `tests/test_e2e_headless_number_reads.py` re-derives the
+verdict from a real engine.
+
+What a `--value` **string** becomes on the way IN is the other half and is **not** covered here:
+the ops coerce it with the engine's own parser (`String.to_float`), which drops the low decimal
+digits of a full-precision literal between `1e-4` and `1e-2` and reads a `DBL_MIN`-scale or
+subnormal literal as `0.0`. Unlike the live wire, headless does not refuse those — see
+[#772](https://github.com/aigengame/godot-agent/issues/772). So a `set` and a following `get`
+agree with each other on the value the project holds; that is what this catalog's round-trip
+claims mean, not that every literal survives coercion.
 
 **Structural edits** (established by #56): three commands restructure the node tree within a
 scene file, each a `load → locate → restructure → pack → save` round-trip that reuses the
@@ -730,7 +754,8 @@ valid result.
 **Reading and writing settings** (established by #111): `gda project get SECTION/KEY` reads one
 setting by its full `section/key` name (e.g. `application/config/name`) and reports it as a
 `{setting, type, value}` triple — `type` is the setting's declared Godot type name and `value` its
-JSON projection, the **same projection** `node get` reports for a node property. Compound values
+JSON projection, the **same projection** `node get` reports for a node property, at the same full
+binary64 precision (#771; see "Number reporting" under [`node`](#node)). Compound values
 arrive **structured** (ADR-0035): a `Dictionary` projects to a JSON object, an `Array` or packed
 array to a JSON array, and an embedded `Object` renders as a **reference projection**
 (`{type, resource_path}` for a Resource with a `res://` path), a **texture projection**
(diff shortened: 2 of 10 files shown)


qinhaihong-red:
## Adversarial review — exact head 8d9afad9

Reviewed issue #771 through its implementation PR #780, pinned to base 9122e03e and head 8d9afad9307fe7f9c1963087771cfdc0078194b8. The four axes below were run independently; their findings have not been merged or re-ranked.

**Resolve the contradictory CONTEXT.md requirement inside issue #771.**

One acceptance criterion requires updating every surface that documents the live-side guarantee. A later struck-table CONTEXT.md row instead says, "Nothing left for this issue to do there." Base CONTEXT.md:191-198 explicitly records headless fidelity as lossy and assigns its repair to #771; exact-head CONTEXT.md:191-201 necessarily changes that outcome to uniform read-side binary64 fidelity while retaining different write-side rules. The implementation and PR artifact are correct, but the open ready-for-agent issue gives mutually exclusive completion instructions.

Smallest repair: revise the struck-table note to say that #752 bounded the old claim, but #771 still must update the entry's fidelity outcome when the headless writer changes.

**Terminology and prose assessment:** Sources were CONTEXT.md's Value projection, ADR-0021/ADR-0035, command catalog, gda.live_numbers, issues #752/#771/#772, and Godot's JSON implementation. I reviewed complete exact-head CONTEXT.md, command catalog, bundled Skill, issue/PR text, and every changed precision, projection, residual, and read/write-boundary statement for term stability, authority ownership, causal/scope distinctions, numeric-contract accuracy, usability, and ASD-STE100-style English. No second substantive defect was found. The claimed complete test suites and Linux/Windows validation were not independently run; hosted Godot E2E is skipped.

The change reuses the shared engine-writer boundary and number corpus rather than creating a second headless precision model. Engine-produced and CLI-derived number ownership remains explicit, and the read/write asymmetry stays documented. Claim ARCH-780-SHARED-ENGINE-WRITER remains open; no competing writer owner was introduced.

### Standards — Pass

No hard or judgement-call finding. Public prose, bundled Skill, guard tests, and real-engine tests remain synchronized with the changed writer boundary. No supported Fowler smell remained. Ruff check/format and Pyright passed.

### Spec — Pass

The exact head implements issue #771's corrected behavioral requirement: headless reads use the full-precision writer on the applicable surfaces, retain the engine-owned negative-zero residual, and leave input parsing to #772.

### Validation

- 41 targeted fast tests passed.
- 4 local real-Godot precision E2E tests passed.
- Shared corpus/contract-guard suite: 263 passed.
- Godot source confirms the full-precision scientific path and the exact-zero special case.
- Executed CI checks are green; hosted Godot E2E is skipped.

qinhaihong-red:
## Review reply — the one finding adopted

No code change (Architecture / Standards / Spec all Pass; the finding was inside issue #771's own text).

**Artifact (contradictory CONTEXT.md instruction in the issue):** adopted. The struck-through table row said "Nothing left for this issue to do there", contradicting the AC that requires updating every surface where the guarantee changed. The row is rewritten: #752 BOUNDED the old claim and named #771 as owner of the headless half; when the headless writer changes, the entry's fidelity outcome must change in the same PR (read side uniform; write side stays #772's) — which is exactly what this PR did to `CONTEXT.md:191-201`. A dated correction note marks the edit. PR head unchanged at `8d9afad9`.

qinhaihong-red:
## Re-review residue — exact head 8d9afad9

The previous Artifact finding is only partially closed.

Issue #771's CONTEXT.md table row now correctly says that #771 must update the Value projection fidelity outcome. However, the immediately following prose still says:

- "The four catalog sentences are what remains."
- "The glossary line ... is already handled; see the struck row above."

The referenced row is no longer struck and now explicitly requires the CONTEXT.md update, as does the acceptance criterion. The issue therefore still gives conflicting completion guidance even though this PR correctly updated CONTEXT.md.

Smallest repair: state that the four catalog sentences and the glossary fidelity outcome are in scope, and replace the stale "struck row/already handled" sentence with the narrower fact that #752 only bounded the old claim.

No code, Architecture, Standards, or Spec residue was found at this head.

qinhaihong-red:
## Re-review residue reply — adopted

The two stale sentences following the corrected table row in issue #771 are rewritten: the doc scope now reads "The four catalog sentences AND the glossary row's fidelity outcome", and the "already handled; see the struck row above" sentence now states the narrower fact — #752 only BOUNDED the old claim; the fidelity outcome changes with this issue's writer fix, in the same PR (which is what this PR did). The issue no longer gives conflicting completion guidance. PR head unchanged at `8d9afad9` — no code was implicated.

Issue #716, opened 2026-08-18
Decision: should a timeout-captured error stream re-verdict to the #651 entry-load codes?


qinhaihong-red:
## What to decide

When `script run` times out, the captured stderr may contain errors that would satisfy #651's entry-load verdicts (`script_missing`, `not_a_main_loop`, ...). #655 deliberately does NOT re-verdict: the captured lines stay advisory diagnostics under the `launch_timeout` envelope, because ADR-0031 records that shape as reaching a failure "by another route" and narrowing it is a separate decision. This issue is that decision.

Arguments for re-verdicting: an agent branching on `Gda error code` gets the precise cause instead of a generic timeout. Against: the timeout is also true (both facts hold); re-verdicting couples the timeout path to the entry-load classifier and can misattribute when the error is stale or partial in a truncated stream.

## Acceptance criteria

- [ ] A recorded decision (issue comment or ADR note): re-verdict, keep advisory-only, or a scoped middle form.
- [ ] If behavior changes, a follow-up implementation issue with its own ACs.

## Blocked by

None - decision can be taken immediately.

---

> **Scope note (2026-08-31, P4 pre-dispatch review):** this decision was framed when only `script run` preserved a timeout's captured stream. Since #714 (P3), the sentinel, export, and `resource import` channels ALSO carry captured stderr under `launch_timeout` — so "should a captured error stream re-verdict the timeout" now has FOUR launch-backed channels, not one. Decide once for all of them, and place the dated note where all four are governed: beside `launch_timeout`'s ADR-0002 registry row (ADR-0031 covers only `script run`; its file now ends with #714's own `> Outcome (2026-08-31)` block — write after it, not before). The decision remains docs-only unless it re-verdicts.

qinhaihong-red:
> *This was generated by AI during triage.*

## Triage decision (2026-08-27): keep the captured stream ADVISORY — do not re-verdict

**Category:** `enhancement` (contract decision). **State:** `ready-for-agent` — the remaining work is recording the decision durably.

### The decision

A `script run` that gda ended at its timeout keeps reporting `launch_timeout`. Errors recognized in the captured stderr stay **advisory diagnostics** under that envelope; they do not replace the code with a #651 entry-load verdict.

### Why

- **Only one of the two facts is gda's own verdict.** gda knows it ended the run; the entry-load cause is *inferred* from bytes the child happened to have written. Re-verdicting would report an inference with the authority of an observation.
[...]

PR #791, opened 2026-08-31
feat(gda): report the elapsed clock and reached ceiling on a preflight timeout (#787)

### LINKED ISSUE #787: scene preflight: a timeout verdict drops the elapsed/ceiling evidence the launch already measured
## Parent

Found by the `/improve-codebase-architecture` review of promotion PR #786. The gap is one #714 itself exposed by raising the standard: the `Headless launch` measures elapsed wall clock and carries the reached ceiling on EVERY run, and after #714 four of the five timeout surfaces publish that evidence — the three `launch_timeout` channels and `script run`'s envelope. `scene preflight` is the fifth: its `status: timeout` verdict reads neither value (verified on the PR #786 head — zero reads in the command module; its own unit fixtures even construct the values and never assert them).

## What to build

A `Startup preflight` that ends at the caller's frame ceiling reports the same two numeric facts every other timeout surface reports: how long the run actually took, and what bound it hit. Today an agent reading `status: timeout` cannot distinguish a run that was one second over a tight ceiling from an hour-long hang — the GDA-DF-032 ambiguity, reintroduced for exactly this one channel.

This is NOT a re-verdicting change: the timeout status, its `started: false`, and the diagnostics stay as they are. The verdict gains evidence fields read off the launch result the command already holds. The only real coupling with #716 (P4 slot 1) is the shared TIMEOUT PROSE in the catalog and Skill — #716's triaged decision is docs-only and concerns `script run`'s verdict code, not this verdict object (corrected 2026-08-31; an earlier sentence here claimed a shared model). Coordinate sentences, not schemas.

## Acceptance criteria

- [ ] The preflight timeout verdict carries the elapsed wall clock and the ceiling that was reached, named consistently with the other timeout surfaces' wording.
- [ ] The machine schema and rendered output describe both fields; catalog and Skill state the evidence parity where they describe the preflight timeout.
- [ ] A real-engine e2e pins a genuinely slow scene: the verdict's elapsed is at least the ceiling, and the ceiling equals what the caller configured.
- [ ] Red-proofed: dropping the field reads reproduces the evidence-free verdict.
- [ ] Non-timeout preflight verdicts are byte-identical (no schema drift for successful runs).

## Blocked by

None - can start immediately. Soft ordering with #716: shared timeout PROSE in catalog/Skill only — coordinate sentences, not schemas (corrected 2026-08-31).

qinhaihong-red:
## What

`gda scene preflight`'s `timeout` verdict now carries the two numbers every other
timeout surface reports: `elapsed_seconds` (how long the run actually took) beside
`timeout_seconds` (the `--timeout` ceiling it reached).

The headless launch measures both on every run it ends — `RunResult.elapsed_seconds`
and, since #714, `RunResult.timeout_bound` — and this command module read neither.
Four of the five timeout surfaces publish the pair; this was the fifth, so a reader
of `status: timeout` could not tell a run a fraction over a tight ceiling from one
stuck for an hour (the GDA-DF-032 ambiguity, for this one channel).

## How

- **Read off the run, not re-derived from the params.** `TimeoutBound` is the single
  home of "the ceiling this run reached" (#714), so the verdict reads it there. The
  fallbacks cover a hand-built `RunResult` at the injected launch seam and are
  truthful rather than a guess: the ceiling gda passed IS `params.timeout`, and a run
  gda ended AT the ceiling ran at least that long — so an unmeasured clock reports
  its lower bound instead of a `0.0` that would read as "ended instantly".
- **Not a re-verdicting change.** `status`, `started`, and `diagnostics` are
  untouched. A `timeout` is still a SUCCESSFUL operation.
- **Non-timeout verdicts are byte-identical.** The pair is OMITTED, never null (gda's
  omitted-never-null convention, cf. `gda.provenance`), through a wrap serializer that
  considers only those two keys — `project_root` stays required-but-nullable, since
  its null MEANS projectless.
- **Human channel too**: `--timeout 3.0s reached (elapsed 3.02s)` under the headline,
  in the wording the `launch_timeout` envelopes use. Every other verdict renders
  exactly the lines it always did.
- **Docs**: the field descriptions carry the contract into `gda schema`; the command
  help, `docs/command-catalog.md` and `src/gda/skill/SKILL.md` state the evidence
  parity where each describes the preflight timeout. (Only the preflight-evidence
  sentences — the shared timeout prose belongs to #716.)

## Verification

- **Red-proofed.** Dropping the two field reads reproduces the evidence-free verdict:
  4 unit tests and the new e2e fail with `KeyError: 'timeout_seconds'`. Neutering the
  omit-serializer fails the 4 invariance cases plus the pre-existing full-dict
  `test_a_clean_start_is_ready_and_started`.
- **Unit** (`tests/test_scene_preflight_commands.py`, 31 passed): the evidence pair;
  the ceiling read off the launch rather than the params; the unmeasured-run lower
  bound; the human render; and an invariance guard parametrized over `ready`,
  `ready-with-errors`, `not_ready` and `quit-after-readiness`.
- **Real engine** (`tests/test_e2e_scene_preflight.py -m e2e`, 14 passed, existing
  tests unmodified): a blocked `_ready` at a `--timeout 3` — `elapsed_seconds >= 3.0`
  and `timeout_seconds == 3.0` — plus a clean scene that carries neither key.
- **Manual real-engine run** of both channels (Godot 4.6.3, `--timeout 2`): human
  `--timeout 2.0s reached (elapsed 2.00s)`; JSON `"elapsed_seconds":2.0143…,
  "timeout_seconds":2.0`.
- **Gates**: `pytest -m "not e2e"` 2259 passed; `ruff check` / `ruff format --check`
  clean; `pyright src/gda` 0 errors.

## Residual

Trivial serial-merge ordering with #716 (P4 slot 1) on `docs/command-catalog.md` and
`src/gda/skill/SKILL.md` — adjacent sentences in the same paragraphs, no shared code.

Closes #787


qinhaihong-red:
## Independent adversarial review — PR #791

Reviewed at pinned head `5294158fa17823ebcf05e95f2c05488f50a0bf5e` against base `181da20d`, in a private detached archive (`git worktree add --detach`), never a shared worktree. Every gate, red-proof and engine run below was re-executed here rather than taken from the PR body.

### Verdicts

| Axis | Verdict |
| --- | --- |
| `artifact-review` (catalog, SKILL, docstrings, field descriptions) | **Pass** — 2× P3 |
| `design-domain-modular-architecture` | **Pass** — 1× P3 |
| `review` — Standards | **Pass** |
| `review` — Spec (issue #787 ACs) | **Pass** — all 5 met, independently verified |

No P1. No P2. Three P3s, all optional.

---

### Findings

**P3-1 — the "same wording as the `launch_timeout` envelopes" claim is half true.**

`render_scene_preflight`'s docstring (and the PR body) says the human line is "in the wording the timeout envelopes use". Actually:

- envelope (`errors.launch_timeout_failure`): `Godot scene preflight launched but did not return before the timeout of 5.0s (elapsed 5.42s).`
- preflight: `  --timeout 5.0s reached (elapsed 5.42s)`

Only the `(elapsed X.XXs)` clause is shared; the ceiling half is a different phrasing. This is not a defect — naming the flag is *better*, and it lines up with #793's caller-first "raise `--timeout`" remediation — but the provenance sentence in the source overstates. Suggest: "sharing the envelopes' `(elapsed …)` clause, naming the flag rather than the bound".

**P3-2 — the two fallbacks read different sources, so a hand-built result can render an impossible pair.**

```python
elapsed_seconds=(raw.elapsed_seconds if raw.elapsed_seconds is not None else params.timeout),
timeout_seconds=bound.seconds if bound is not None else params.timeout,
```

Answering the sourcing question directly: **the fallback cannot overstate in production.** `scene.py` passes `timeout=params.timeout` into `launch`, and the streaming loop breaks only at `elapsed >= timeout`, so `params.timeout` is a genuine lower bound. It is also dead code in production — since #714 `launch` has a single streaming strategy that always sets both `elapsed_seconds` and `timeout_bound` on a synthesized `TIMEOUT`, so both fallbacks are reachable only through the injected `make_launch` seam.

At that seam the two can disagree. A `RunResult` carrying a `timeout_bound` but no clock takes `elapsed_seconds` from `params.timeout` and `timeout_seconds` from `bound.seconds`; where those differ the pair violates the invariant the PR asserts. Concretely, the shape `test_the_ceiling_is_read_off_the_launch_rather_than_re_derived_from_the_params` builds (bound `8.0`, params `5`) with its clock removed renders `--timeout 8.0s reached (elapsed 5.00s)` — you cannot reach an 8s ceiling in 5s. One-line fix that makes the lower bound honest by construction: fall back to the same source, `bound.seconds if bound is not None else params.timeout`.

For the record, the bound≠params fixture itself is **legitimate**: identical values prove nothing about provenance, so the divergence is the only thing that makes the source of `timeout_seconds` observable, and the test comment says exactly that.

**P3-3 — ADR-0002's #664 scope note is the designated authority for this contrast and was not updated.**

`docs/adr/0002-headless-structured-output-contract.md:91` records the decision that preflight reports its bound "as a SUCCESS carrying `status: timeout` (with the captured diagnostics), not as the `launch_timeout` envelope". Nothing there becomes false, but it is now incomplete in exactly the dimension #787 changed — the verdict carries the envelope's evidence pair too. This matters because sibling #793 adds a note to the same file that ends "read that note rather than re-deriving the difference by diffing the two commands", so a reader is explicitly pointed at prose that omits the new parity. One clause on the existing bullet would close it. Ordering note: #793 edits the same file (at line ~325), so whichever lands second should carry it.

---

### Adjudication points cleared (with evidence)

**The rendered `--timeout 5.0s reached` line — REFUTED as a concern.** `gda scene preflight` really does expose the flag; from the live CLI help at this head:

```
--timeout   FLOAT   Seconds before gda ends the launch and reports the 'timeout'
                    verdict with the captured diagnostics. ...
                    [default: 30.0]
```

The line teaches a real, raisable flag. Vocabulary against slot 1 (#793) is complementary, not colliding: #793 says "raise the ceiling (`--timeout`, where the command has one)", #791 says "raise `--timeout`".

**omitted-never-null — correct on every axis I could reach.**
- `--schema`: both fields are absent from `required` (unchanged from base: `['path','started','status','project_root']`) and shaped `{"anyOf":[{"type":"number"},{"type":"null"}],"default":null}` — byte-identical in shape to the established `GodotProvenance.version` precedent the comment cites, so this follows the convention rather than inventing one.
- `model_dump(mode="python")`: the wrap serializer runs there too; keys are omitted in python mode as well as JSON. Verified, plus round-trip reconstruction and `exclude_none=True` composition.
- `project_root: null` is untouched — confirmed at runtime, not just in the comment.
- MCP: `gda.mcp.server._success_result` is a pure `json.loads` passthrough into `structured_content` and declares **no `outputSchema`**, so there is no output validation an omitted key could trip.

**Non-timeout byte-identity — verified two ways.** Real engine (Godot 4.6.3), head vs base 181da20d, same scene: JSON `{"path":"res://ok.tscn","started":true,"status":"ready","diagnostics":[],"project_root":"..."}` and human `ready res://ok.tscn` are byte-identical. Whole-`--schema` comparison: input, error and argv sections identical; among shared properties, `other keys changed? set()`.

**Catalog/SKILL serial merge with #793 — no collision.** #791 touches catalog ~line 223 and SKILL line 123; #793 touches catalog ~727/~888/~982 and SKILL 66/145/230. Disjoint hunks, clean auto-merge, complementary guidance. The PR body's "adjacent sentences in the same paragraphs" is a conservative overstatement — they are different sections.

**Red-proofs, re-run in my own archive.**
- Deleting the two field reads: `4 failed, 27 passed` — exactly the four claimed tests.
- Neutering the omit-serializer: `5 failed, 26 passed` — the 4 invariance cases plus the pre-existing `test_a_clean_start_is_ready_and_started`, as claimed.

**Gates, re-run here.** Unit `31 passed`. Real engine `pytest tests/test_e2e_scene_preflight.py -m e2e` → **14 passed** (own `GDA_USER_DATA_ROOT` per PITFALLS). Manual both-channel run at `--timeout 2`: human `--timeout 2.0s reached (elapsed 2.05s)`, JSON `"elapsed_seconds":2.0031…,"timeout_seconds":2.0`. `ruff check` / `ruff format --check` clean. `pyright src/gda` **0 errors**.

Recommendation: mergeable as-is. P3-2 is the one I would take (one line); P3-1 and P3-3 are prose.

qinhaihong-red:
## Review reply — all four axes passed; the two actionable P3s adopted

Head is now `279a5117` (one commit on `5294158f`).

- **P3-2 (fallback source fork):** adopted, your one-line shape. Both evidence fields now read the bound first when the clock is missing, so a seam-injected result can no longer render the self-contradictory `--timeout 8.0s reached (elapsed 5.00s)` pair. A comment records that production never takes either arm (the streaming launch fills both — your own dead-code analysis). 31 preflight tests green.
- **P3-3 (ADR-0002's #664 note not synced):** adopted. The note is the authoritative record of the verdict-vs-envelope comparison, and #793 now points readers at it — it carries a dated #787 sentence: the SHAPE difference stands; the numeric EVIDENCE is at parity, read off the same launch measurement.
- **P3-1 (the "borrowed wording" claim is half-true):** acknowledged as accurate; the PR body's phrasing stands corrected by your finding — no artifact carries the overclaim (it was report prose), so no edit beyond this acknowledgment.

Thanks for disproving the dispatch-side worry about `--timeout` — the flag is real, the line teaches a real knob.

qinhaihong-red:
## Full independent adversarial review — PR #791

Reviewed from first principles at pinned head `279a5117c42d106c39fd22b056dd48c9db4f7cd1` against `feat/gda-dogfooding-dev@181da20d2ea418977600bd73d2958dc4308e8de6`. Existing reviews and author responses were excluded from the review evidence and scope.

| Axis | Verdict |
| --- | --- |
| Standards | **Pass** |
| Spec (#787) | **Pass** |

`docs/command-catalog.md:223-230` and `src/gda/skill/SKILL.md:126` claim that the evidence pair distinguishes a scene that is merely slow from one that is stuck. It does not. A `_ready` that never returns and a healthy frame window that overruns the same ceiling can both report essentially the same `elapsed_seconds` and `timeout_seconds`; the pair quantifies the elapsed time and configured/reached ceiling, but does not encode the cause.

This matters because the prose directs remediation from evidence that cannot support that diagnosis. Keep the useful, observable claim from #787: the pair distinguishes different duration/ceiling situations and helps decide whether to retry with a larger bound. Do not claim it distinguishes the two causes without diagnostics, progress evidence, or a changed bound.

Pass. `Headless launch` / `RunResult` remains the owner of timing facts; the `scene` command only projects existing measurements into the Startup preflight result and renderer. No duplicate timing authority, reverse dependency, or speculative extension was introduced.

### Standards

Pass. No non-tool-enforced repository-standard violation or actionable Fowler smell was found in the complete diff.

### Spec

Pass. #787's timeout-only fields, schema/rendering publication, catalog/Skill coverage, real-engine regression, red proof, and non-timeout omission/invariance are implemented.

Sources: `CONTEXT.md` definitions for **Startup preflight**, **Headless launch**, and **Error envelope**; ADR-0002; issue #787; all changed documentation and field descriptions. Writing reference: ASD-STE100 Simplified Technical English. The complete PR/issue prose and changed requirements/remediation passages were inspected. The causal overclaim above is the only substantive terminology/prose problem.

Validation: `git diff --check` passed. Current hosted CI has no failed or pending required check, but the Godot e2e job is skipped; the real-engine timeout experiment was not rerun in this review.

qinhaihong-red:

New head: `bbb518592` (one docs commit on top of the reviewed `279a5117`).

| Catalog and SKILL claim the `elapsed_seconds`/`timeout_seconds` pair tells a merely-slow scene from a stuck one; it cannot — both causes report essentially the same numbers | **Fully adopt** (and widened to the same claim's other copies) | On a timeout, elapsed ≈ ceiling by construction; a stuck `_ready` and a healthy `--frames` window overrunning the same bound are indistinguishable from the pair, and the params contract says the bounds are not cross-checked | Catalog and SKILL now state the observable claim — the pair names the consumed ceiling and what raising it buys; choose a larger `--timeout` or fewer `--frames` from what the scene should do, and rerun. The SAME overclaim in this slice's own surfaces is corrected too: the published `elapsed_seconds` field description (schema text moves with it — text only, no shape change), the result-model and command docstrings, and a renderer comment. One comparison stays (`scene.py`, "a fraction over a tight ceiling vs stuck for an hour"): it contrasts different duration/ceiling situations, which the pair genuinely distinguishes | 135 focused tests green; grep sweep of `stuck`/`merely slow` leaves only the defensible different-numbers comparison and the shared `launch_timeout` message owned elsewhere |

PR #368, opened 2026-07-01
feat(ops): project-local class_name static-scan resolution (ADR-0032)

qinhaihong-red:
## What & why

Implements **#360** per **ADR-0032**. Before this, `node add --type <class_name>`, `resource create --type <class_name>`, and `project find-references <class_name>` resolved a project-local GDScript `class_name` **only** through the editor global class list (`ProjectSettings.get_global_class_list()`), populated exclusively by the Godot editor scan into `.godot/global_script_class_cache.cfg`. In a headless, **editor-never-opened** project — gda's core target — there is no `.godot/`, so a valid project-local `class_name` was unresolvable.

## Change

A single unified `class_name → script-path` resolver (`_resolve_project_class_script`) shared by all three sites, as a cache-first chain:

1. built-in engine class (`ClassDB`, base-class-checked) — **stays per-site** (only resolution is unified);
2. the editor global class list — **kept first**, so editor-opened projects are unchanged (the fallback is unobservable there);
3. a **gda-owned static scan** of the project's own `.gd` sources — invoked only when tiers 1–2 miss.

The scan **reuses** the existing res:// walker (`_collect_resource_paths`, skips `.godot/`) and the never-compiled raw-source parser (`_script_metadata`), building a `class_name → paths` index once per process. It runs **no project code** (parse only); the Project-code execution surface (ADR-0009) is unchanged.

### Contract edges
- **Duplicate `class_name`** (declared in >1 `.gd`) → new `ambiguous_class_name` error naming the conflicting paths, for all three sites — never first-file-wins.
- **Unresolved type** → keeps `invalid_node_type` / `invalid_resource_type`, message upgraded to be actionable ("no .gd script declares class_name X") and no longer implying a missing editor cache.
- `.cs` class names stay out of scope (fail as unknown type).

`ambiguous_class_name` is registered as an Operation-reported code; the GDScript↔Python mirror and the ADR-0002 table are kept in sync.

## Tests
- Fast Python tier: `uv run pytest -m "not e2e"` → **1024 passed** (incl. the error-registry mirror/drift checks).
- New e2e (`tests/test_e2e_classname_resolution.py`, through the real engine): the 4 ADR-0032 scenarios — (a) no `.godot/` resolves all three sites, (b) editor-opened cache path unchanged, (c) duplicate → `ambiguous_class_name`, (d) unknown type actionable message → **9 passed**.
- Existing class_name/type e2e regressions (node + resource) → **14 passed**; project-analysis e2e → all pass.

[...]

PR #792, opened 2026-08-31
refactor(gda): one walk for the mounted-group tree, and the nested-tree proof (#788)

### LINKED ISSUE #788: One walk for the mounted-group tree, and the nested-tree proof adopt_group_json is missing
## Parent

Found by the `/improve-codebase-architecture` review of promotion PR #786. Two facts, both verified on the PR head:

1. The registration-time walk over the mounted command groups exists TWICE — `hints.adopt`'s group walk and `adopt_group_json`'s (#683) — two near-identical shapes 190 lines apart, asymmetric in defensiveness (one refuses to clobber a group's own callback, the other has no equivalent guard for a pre-customized `cls`) and in proof (see 2). Beyond them, the LIVE click tree is walked independently four more times (one production surface walker, three test-file copies) — six tree walks, two representations, no shared callable.
2. **`adopt_group_json`'s recursion is unproven**: deleting its recursive call passes the whole suite today, because the live tree is two levels deep and every group-`--json` test exercises a first-level group. Its twin has a dedicated synthetic nested-tree test (`test_the_adoption_reaches_a_nested_sub_group`); this walker has nothing — a silent gap that a future sub-group-of-a-group would fall through.

## What to build

One shared walk primitive for the mounted-group tree, with both adopters as visitors, and the nested-tree proof the second adopter is missing. The walk owns tree-shape knowledge (registration-time representation, recursion, the skip-on-None guard); each adopter contributes only what it installs. A third cross-cutting group behavior then becomes one more visitor, not a seventh walker.

Scope boundary: the LIVE-tree walkers (the surface walker and the three test copies) are OUT — they walk a different representation for different reasons; consolidating them is a separate judgment this issue does not prejudge. The nested-tree test for `adopt_group_json` is the non-negotiable half: land it even if the primitive's shape stalls in review.

## Acceptance criteria

- [ ] A synthetic three-level tree proves `adopt_group_json` reaches a sub-group of a group (red-proofed: deleting the recursion fails THIS test and nothing else).
- [ ] One registration-time walk primitive; both adopters route through it; neither loses its current behavior (the callback-refusal guard included, now applied uniformly or its absence on the hints side explicitly reasoned).
- [ ] Public help/schema surfaces byte-identical (this is internal wiring; the group `--json` acceptance suite and the near-miss suite stay green unmodified).
- [ ] The composition root's mount-then-adopt ordering requirement is stated once, on the primitive.

## Blocked by

None - can start immediately. Serial with #685 (P4 slot 3) inside `hints.py` — merge-cost order only, either first.

qinhaihong-red:
## What

Two commits, in the order the issue puts them.

**1. The nested-tree proof (`test`).** `adopt_group_json` recurses into every mounted
sub-group, but nothing proved it: the live tree is two levels deep and every
group-`--json` test exercises a FIRST-level group, so deleting the recursive call
passed the whole suite. The new
`test_the_json_adoption_reaches_a_nested_sub_group` builds the synthetic
three-level tree its twin on the refusal side already has
(`test_the_adoption_reaches_a_nested_sub_group`) and asserts the acceptance
itself: `--json` parses between the nested group and the command it runs, and the
nested group's help advertises it.

**2. One walk (`refactor`).** `gda.headless.walk_mounted_groups` is now the single
generator over every mounted group at any depth; both adopters are visitors over
it. The walk owns the tree-shape knowledge — the `TyperInfo` registration
representation, the recursion, the skip-on-None descent, and the composition
root's mount-then-adopt ordering requirement, stated once for every visitor. A
third cross-cutting group behavior becomes one more visitor, not a seventh walker.

## Where the primitive lives, and why

`gda.headless`. `hints` already imports that module, so routing through it adds no
import edge and risks no cycle; and `headless` already owns the cross-cutting
CLI-surface wiring the whole tree shares — the shared option factories, the
ancestor-`--json` contract, and one of the two adopters. A module of its own would
have been a third file for four lines of recursion.

## The guard asymmetry, resolved

The callback refusal stays a property of the `--json` VISITOR, not of the walk it
rides: it guards the one slot that adoption writes. Applying it uniformly is not
merely undesirable, it is impossible — `adopt_group_json` runs FIRST in the
composition root and gives every group a callback, so the same check on the hints
walk fires on the entire tree. Verified by experiment: hoisting the check into the
walk makes `import gda.cli` die at `hints.adopt` on the first group (`scene`). The
hints side's reasoned absence of an equivalent `cls` guard is written where the
walk is consumed: the class is what gda *imposes* on the whole surface (a group
opting out would silently lose the structured refusal), and it replaces no
declaration — Typer leaves `cls` a default placeholder unless `add_typer` was
given one, and no gda group is.

## Scope

The LIVE-tree walkers (the surface walker and the three test copies) are untouched,
per the issue. This is internal wiring only.

## Evidence

- **Red-proof, the non-negotiable half (measured on the unchanged walk, commit 1):**
  deleting `adopt_group_json`'s recursive call → `1 failed, 2250 passed`, the one
  failure being the new test.
- **Red-proof, after the refactor:** deleting the now-shared `yield from` →
  `2 failed, 2249 passed`, exactly the two nested-tree tests and nothing else.
- **Public surface byte-identical:** the root `--help`, every group's `--help`,
  every leaf's `--schema` and the whole-surface `gda schema --json` manifest,
  captured before and after — same 1362095 bytes, md5
  `4c6963138a5b6f8821de9ddd9e417159`. The group `--json` acceptance suite and the
  near-miss suite are unmodified.
- **Gates:** `pytest -m "not e2e"` → 2251 passed; `ruff check .`, `ruff format
  --check .`, `pyright` (0 errors) all green.
- **No engine run:** this is CLI-layer wiring; nothing here spawns Godot, and no
  e2e tier is implicated.

Closes #788

### DIFF
--- a/src/gda/headless.py
+++ b/src/gda/headless.py
@@ -269,7 +269,11 @@ def walk_mounted_groups(app: typer.Typer) -> Iterator[TyperInfo]:
     Typer refuses to build such a group into a command at all. A visitor that needs the
     instance says so itself; the walk does not decide that for it.
     """
-    for group in app.registered_groups:
+    # Snapshot, so the docstring's "what is mounted at the moment it runs" is
+    # literally true: a visitor that mounts a group MID-WALK is excluded rather
+    # than lazily swept in (#792 review P3-2 — no visitor does this today; the
+    # snapshot makes the documented boundary the actual one).
+    for group in list(app.registered_groups):
         yield group
         instance = group.typer_instance
         if instance is not None:
--- a/src/gda/headless.py
+++ b/src/gda/headless.py
@@ -8,7 +8,7 @@ classification, failure output, and JSON rendering.
 """
 
 import sys
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Generic, NoReturn, Optional, TypeVar
@@ -17,6 +17,7 @@ import typer
 from pydantic import BaseModel, ValidationError
 from typer._click import Context as ClickContext
 from typer.core import TyperCommand
+from typer.models import TyperInfo
 
 from gda.binary import resolve_godot_binary
 from gda.errors import (
@@ -243,6 +244,38 @@ def _group_json(
     """
 
 
+def walk_mounted_groups(app: typer.Typer) -> Iterator[TyperInfo]:
+    """Every command group mounted below ``app``, at any depth (#788).
+
+    The ONE walk over the mounted-group tree, so a cross-cutting group behavior is
+    written as a visitor over this rather than as another copy of the recursion. Two
+    visitors today: the shared ``--json`` option (:func:`adopt_group_json`, below) and
+    the refusal class (``gda.hints.adopt``). It walks the REGISTRATION-time tree — the
+    ``TyperInfo`` records ``add_typer`` leaves behind — because that is the only
+    representation that exists before Typer builds the click tree, and so the only one
+    a composition-time adopter can install anything onto. The BUILT click tree is a
+    different walk for a different purpose — reading a finished surface
+    (``gda.surface``).
+
+    **The ordering precondition, stated here once for every visitor:** the walk reports
+    what is mounted AT THE MOMENT IT RUNS, so the composition root (``gda.cli``) mounts
+    the whole tree first and adopts afterwards. A group mounted after an adoption is
+    never visited by it and silently keeps none of what it installs.
+
+    What is yielded is the registration record, not the sub-app: it is the wider handle
+    — the group's ``name`` and ``cls`` as well as its ``typer_instance`` — and one
+    visitor writes to the record itself. Typer types that instance as optional, so the
+    RECURSION skips a group without one: it carries no groups to descend into, and
+    Typer refuses to build such a group into a command at all. A visitor that needs the
+    instance says so itself; the walk does not decide that for it.
+    """
+    for group in app.registered_groups:
+        yield group
+        instance = group.typer_instance
+        if instance is not None:
+            yield from walk_mounted_groups(instance)
+
+
 def adopt_group_json(app: typer.Typer) -> None:
     """Give every group mounted below ``app`` the shared ``--json`` option (#683).
 
@@ -252,16 +285,20 @@ def adopt_group_json(app: typer.Typer) -> None:
     a line an agent composes naturally. Applied once from the composition root,
     AFTER every group has mounted itself, for the reason ``gda.hints.adopt`` is: it
     is one property of the WHOLE surface, so a group added later inherits the option
-    by being mounted rather than by remembering to declare it. The walk RECURSES for
-    the same reason that one does — a sub-group of a group would otherwise be missed
-    silently.
+    by being mounted rather than by remembering to declare it. Reaching a sub-group of
+    a group — which would otherwise be missed silently — is the shared walk's job
+    (:func:`walk_mounted_groups`), including the ordering precondition it states.
 
     Installing the option means installing the callback that carries it, so a group
     that declares a callback of its own is refused rather than silently replaced:
-    such a group must declare the shared option on that callback itself.
+    such a group must declare the shared option on that callback itself. That refusal
+    stays a property of THIS visitor, not of the walk it rides: it guards the one slot
+    this adoption writes, and the other visitor overwrites no such slot.
     """
-    for group in app.registered_groups:
+    for group in walk_mounted_groups(app):
         instance = group.typer_instance
+        # A group registered without a sub-app has no callback to carry the option;
+        # the walk's own contract leaves that reading to each visitor.
         if instance is None:
             continue
         if instance.registered_callback is not None:
@@ -271,7 +308,6 @@ def adopt_group_json(app: typer.Typer) -> None:
                 "gda.headless.adopt_group_json replace it."
             )
         instance.callback()(_group_json)
-        adopt_group_json(instance)
 
 
 def godot_option() -> Optional[str]:
--- a/src/gda/hints.py
+++ b/src/gda/hints.py
@@ -58,7 +58,12 @@ from typer._click.exceptions import NoSuchOption
 from typer._click.globals import get_current_context
 
 from gda.errors import Failure, make_failure
-from gda.headless import emit_failure, json_in_effect, remember_argv
+from gda.headless import (
+    emit_failure,
+    json_in_effect,
+    remember_argv,
+    walk_mounted_groups,
+)
 
 # The two registered codes this module reports (ADR-0002, the `usage` category).
 UNKNOWN_COMMAND = "unknown_command"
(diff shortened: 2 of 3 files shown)


qinhaihong-red:
## Independent adversarial review — PR #792

Reviewed at pinned head `113f07dbb9a4b34444740a99b53c8b6f624056a5` against base `181da20d`, in a private detached archive (`git worktree add --detach`), never a shared worktree. Every measurement, red-proof and byte-identity claim below was re-derived here rather than taken from the PR body.

### Verdicts

| Axis | Verdict |
| --- | --- |
| `artifact-review` (docstrings, comments, issue alignment) | **Pass** — 2× P3 |
| `design-domain-modular-architecture` | **Pass** — genuine deepening |
| `review` — Standards | **Pass** |
| `review` — Spec (issue #788 ACs) | **Pass** — all 4 met, incl. the non-negotiable half |

No P1. No P2. Three P3s, all optional. The claim that this is CLI-layer only is **verified**: nothing on either changed path spawns Godot, and no e2e tier is implicated.

---

### Findings

**P3-1 — the walk's documented skip-on-None guard does not hold for the representation it describes.**

> "Typer types that instance as optional, so the RECURSION skips a group without one"

`TyperInfo.typer_instance: Optional["Typer"] = Default(None)` — the *annotation* is Optional, but the runtime default is a `DefaultPlaceholder`, never `None`. A `TyperInfo` built without an instance therefore passes `is not None`, and the walk raises:

```
AttributeError: 'DefaultPlaceholder' object has no attribute 'registered_groups'
```

Reproduced — and reproduced **identically on both base implementations**, so this is not a regression, and it is unreachable today because `Typer.add_typer(typer_instance, *, ...)` takes the instance as a required positional in typer 0.26.7. The finding is narrower: the refactor promotes this guard into "tree-shape knowledge the walk owns" and documents a safety it does not actually deliver. Cheapest honest fix is either prose (say the guard is defensive against a hand-built record that Typer's own API cannot produce) or `if instance:`, which covers both `None` and the placeholder.

**P3-2 — the mount-then-adopt precondition is stated three times, and the generator phrasing is imprecise.**

AC 4 asks for it "stated once, on the primitive". It *is* on the primitive — but `adopt_group_json` ("Applied once from the composition root, AFTER every group has mounted itself") and `hints.adopt` ("Applied once, in the composition root, AFTER every group has mounted itself") both restate it, the latter in the same sentence as claiming the walk "states that ordering requirement once for both of its visitors". Low risk, but it is precisely what the AC targets.

Separately, "the walk reports what is mounted AT THE MOMENT IT RUNS" is loose for a lazy generator. Measured: a group appended to `registered_groups` mid-walk **is** visited (`['a','b','c']`), which that sentence reads as excluded. Either snapshot — `for group in list(app.registered_groups):` — to make the sentence true, or state plainly that a visitor must not mount during the walk. (Answering the review question: the note is adequate for today's single call site; the snapshot is what would make it self-enforcing.)

**P3-3 — the hints-side `cls` reasoning rests on an unpinned premise.**

The load-bearing claim is "Typer leaves a group's `cls` a default placeholder unless `add_typer` was given one, **and no gda group is**". I verified both halves at runtime (default is `DefaultPlaceholder`; an explicit `cls=` sets a real class). Answering the question directly: when a group *does* pass `add_typer(cls=...)`, `group.cls = GdaGroup` **silently replaces** it. That is documented as deliberate imposition, and AC 2's "or its absence on the hints side explicitly reasoned" bar is met — so this is hardening, not a gap. But nothing pins the second half of the premise, and this issue's own philosophy is that an unproven invariant should become a test. A one-assert test over the live tree ("no mounted group carries a non-placeholder `cls` before adoption") would make the reasoning self-checking, exactly as the other half of this PR did for the recursion.

---

### Adjudication points cleared (with evidence)

**Visit-order parity — CONFIRMED against both replaced implementations.** Synthetic four-level, multi-branch tree (`a{a1{a1x},a2}, b{b1}, c`), base vs head, for the hints walk and for the JSON adopter's touch order:

```
-- hints --   base: a,a1,a1x,a2,b,b1,c    head: a,a1,a1x,a2,b,b1,c
-- json  --   base: a,a1,a1x,a2,b,b1,c    head: a,a1,a1x,a2,b,b1,c
```

DFS pre-order, identical. The one shape difference — base `adopt_group_json` descended only *after* installing the callback and only into groups that passed the guard, while the new walk descends unconditionally — has no reachable divergence, because a failing guard raises and stops the traversal at the same point in the same order.

**The hoisting measurement — REPRODUCED.** Moving the callback refusal into the walk makes `import gda.cli` die on the first group:

```
File ".../src/gda/cli.py", line 193, in <module>    hints.adopt(app)
File ".../src/gda/hints.py", line 411, in adopt
RuntimeError: Command group 'scene' declares its own callback.
```

Exactly as the PR states. The guard genuinely cannot be hoisted; leaving it on the visitor is correct.

**The nested-tree test pins BEHAVIOR — red-proofed in my own archive.** At commit 1 (`93d8b794`, the unchanged walk), deleting `adopt_group_json`'s recursive call gives **`1 failed, 2250 passed`**, the single failure being `test_the_json_adoption_reaches_a_nested_sub_group`. That is AC 1's "fails THIS test and nothing else", met exactly — and it confirms the gap the issue reported.

After the refactor, deleting the shared `yield from` failed three tests, not two. The third is `tests/test_launch.py::test_timeout_synthesizes_a_result_that_keeps_what_the_run_produced`, which I confirmed is the known full-suite-load timing flake and **not** caused by the mutation: it passes 3/3 in isolation on both the mutated and the pristine tree (already tracked as the #728 deflake). Discounting it, the result is the claimed two nested-tree tests — and note this is now strictly better than base, since the shared recursion is covered from both sides.

**Byte-identity — INDEPENDENTLY RE-DERIVED.** My own capture (root `--help`, all 15 groups' `--help`, every leaf's `--help` *and* `--schema`, plus `gda schema --json`, stdout and stderr, `COLUMNS=80`):

```
BASE 181da20d: 1642172 bytes  md5 725d92367a7c351b0d90dac456b9836b
HEAD 113f07db: 1642172 bytes  md5 725d92367a7c351b0d90dac456b9836b
```

Different totals from the PR's `1362095` / `4c6963…` only because my blob captures a different set (I added leaf `--help` and stderr); the conclusion matches. AC 3 met.

**Serial merge with #798 — CONFIRMED from this side, by a real trial merge onto `97feae2a`.**

- `src/gda/headless.py` auto-merges clean.
- The **only** conflict is `src/gda/hints.py:58`, one line: `from gda.headless import emit_failure, json_in_effect, remember_argv` (#798) vs `from gda.headless import ancestor_json, emit_failure, walk_mounted_groups` (this PR).
- `ancestor_json` has **zero** remaining hints-side uses after #798 — its only consumer, `_json_in_effect`, moves into `gda.headless` as `json_in_effect`. So the resolution keeps both sides and drops it:

```python
from gda.headless import (
    emit_failure,
    json_in_effect,
    remember_argv,
    walk_mounted_groups,
)
```

I applied exactly that: `import gda.cli` succeeds and the merged suite is **`2278 passed`**. Recording it here for whichever of the two lands second.

**Gates, re-run here.** `pytest -m "not e2e"` → **`2251 passed`**. `ruff check .` / `ruff format --check .` clean. `pyright src` → **0 errors**. (A first pyright run reported 1 error on `hints.py:58`; that was my environment — pyright had resolved `gda` to the installed editable base package rather than the archive. With the archive's `src` on the path it is clean. Flagging so nobody re-derives it as a defect.)

### Design note

This is a real deepening rather than a rename: one primitive owns the tree-shape knowledge, the two cross-cutting behaviors become visitors, and the asymmetry the issue complained about is resolved the honest way — one guard stays on its visitor with a reproduced impossibility proof for hoisting it, the other's absence is reasoned from verified Typer behavior. Placing it in `gda.headless` adds no import edge (`hints` already imported it) and avoids a third file for four lines of recursion; I agree with that call.

Recommendation: mergeable as-is. P3-3 is the one I would take, since it is the same "prove the invariant" move this PR's own first commit makes.

qinhaihong-red:
## Review reply — all four axes passed; one P3 adopted, two declined with reasons

Head is now `7b56f3a4` (one commit on `113f07db`).

- **P3-2 (mid-walk mount swept in vs the docstring's claim):** adopted, the snapshot form — `list(app.registered_groups)` makes "what is mounted at the moment it runs" literally true, and the comment records that no visitor mounts mid-walk today. Your measurement (`['a','b','c']`) was the evidence.
- **P3-1 (skip-on-None guard doesn't match the runtime representation):** declined as out of scope with your own evidence — both base implementations carried the same guard, the shape is unreachable through `add_typer` today, and the guard is harmless. Recorded here rather than churned.
- **P3-3 (the `cls` premise unpinned):** declined for now — the AC's "explicitly reasoned" is met at the consumption site, and a pin asserting "no gda group passes `cls=`" would be a negative test for unplanned scope (the repo's recorded stance from #368/#124). If a group ever passes `cls=`, the silent replacement is real; the reasoning comment at the site is where a future author will look.
- **The #798 merge resolution you tested to 2278 passed** is recorded for the integration step — thank you; the fix round will use it verbatim.

qinhaihong-red:
## Full independent adversarial review — PR #792

Reviewed from first principles at pinned head `7b56f3a441e04c3e56ae5cdbe6ac132b1c6403d5` against `feat/gda-dogfooding-dev@181da20d2ea418977600bd73d2958dc4308e8de6`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Artifact review | **Pass** |
| Standards | **Pass** |
| Spec (#788) | **Pass** |

### Artifact review

Pass. The complete PR body and issue accurately describe the registration-time `TyperInfo` tree, snapshot boundary, mount-then-adopt precondition, visitor-specific guards, exclusion of the built Click tree, and the synthetic nested-group proof.

Pass. `walk_mounted_groups` is the single owner of registration-tree shape. JSON adoption and near-miss refusal remain orthogonal visitors that own only their respective writes. The change correctly keeps the built Click-tree traversal separate because it serves a different representation and purpose.

### Standards

Pass. The shared walker removes duplicated recursion without introducing an actionable documented-standard violation or Fowler smell.

### Spec

Pass. The three-level `--json` proof, one registration-time walker, both adopters, callback-refusal behavior, mount-then-adopt contract, and unchanged public behavior match #788.

Sources: `CONTEXT.md`, ADR-0040, issue #788, the complete PR body, and all changed docstrings/tests. Writing reference: ASD-STE100 Simplified Technical English. Names, ownership, ordering, scope, causal claims, and duplicated-rule descriptions were inspected. No substantive terminology or prose issue was found.

Validation: `git diff --check` passed. Current hosted CI has no failed or pending required check; Godot e2e is skipped and was not needed for this internal Typer-tree refactor. The full byte comparison and unit suites were inspected but not independently rerun.

PR #793, opened 2026-08-31
fix(gda): caller-first timeout remediation, with the decisions recorded (#716, #717, #684)

### LINKED ISSUE #684: Live operations under SceneTree.suspended: document the hard limit; timeout guidance names reachable causes
## Review-round source

Found during the #656 review (PR #681, wave 1), verified against the engine source. #656 keeps the harness serving while `SceneTree.paused` is true via `PROCESS_MODE_ALWAYS`. That escape does not exist for `SceneTree.suspended`: `Node::can_process()` is `is_inside_tree() && !tree->is_suspended() && _can_process(is_paused())` (godot `scene/main/node.cpp`), so NO process mode ticks while the tree is suspended. ~~A game that uses `get_tree().suspended = true` (freeze-frame) reproduces the exact GDA-DF-013 dead-end~~ **(refuted 2026-09-02: project GDScript cannot set `suspended` — no ClassDB binding; see the premise-reconciliation note)**. A suspended tree — reachable only from an attached debugger session, which a gda-driven run does not have — would reproduce the exact GDA-DF-013 dead-end mechanism: the harness cannot tick, every live operation stalls, and input injection cannot resume the game because it is itself a harness operation. The daemon side surfaces this as a live-operation timeout, indistinguishable from a hung game.

## What to build

Documentation plus reachable-cause timeout guidance — no serving change is possible (the engine runs no node code while suspended; detection from inside the session is structurally unavailable):

- State the limitation in the skill's live-operations guidance, next to the pause paragraph #656 added: live operations survive `SceneTree.paused` but NOT `SceneTree.suspended`; a game must not suspend the tree while gda drives it (use `paused` for freeze-frame needs during agent sessions).
- ~~Name a suspended tree as a likely cause in the live-timeout failure guidance~~ **(superseded 2026-09-02:** the live-timeout guidance names the REACHABLE causes and must not name suspension, which project code cannot produce; the skill documents the debugger-only suspension mechanism instead — see the premise-reconciliation note**)**.

Out of scope: daemon-side detection or automatic recovery — the engine offers no channel for either; do not add speculative mechanisms.

## Acceptance criteria

- [ ] The skill's live section states the suspended-tree limit and the paused-vs-suspended distinction.
- [ ] ~~The live-timeout failure text names a suspended `SceneTree` among the likely causes.~~ **Superseded (2026-09-02):** the live-timeout failure text names the REACHABLE causes and does not name suspension — see the premise-reconciliation note below.
- [ ] No serving-path behavior change; any help/skill surface diff is enumerated in the PR body.

## Blocked by

- ~~#656 (PR #681) merging into the dev branch~~ — **cleared**: #656 merged in wave 1 and the
  paragraph this extends is on `main` (`src/gda/skill/SKILL.md`, beside the
  `PROCESS_MODE_ALWAYS` pause behaviour at `harness/gda_harness.gd:154-167`). Nothing blocks
  this issue.

---

> **Premise reconciliation (2026-09-02, P4 review round, PR #793):** this issue's mechanism
> premise — "A game that uses `get_tree().suspended = true` (freeze-frame)" — is refuted on
> Godot 4.6.3 source: `SceneTree::set_suspend` / `is_suspended` have no ClassDB binding, so
> `suspended` is not reachable from project GDScript at all. Its only callers are the
> debugger agent (`scene_debugger.cpp`), i.e. an attached editor/debugger session, which a
> gda-driven run does not have. The `Node::can_process()` gate (`scene/main/node.cpp`) is
> real — a suspended tree runs no node code, harness included — so the SKILL limitation
> paragraph stands and PR #793 delivers it (paused-vs-suspended distinction, the
> debugger-only suspension mechanism, `src/gda/skill/SKILL.md`). AC2 as written would put a
> cause project code cannot produce into the live-timeout failure text; the delivered text
> (`src/gda/daemon/session.py`) names the reachable causes instead and rules out `paused`.
> AC1 and AC3 stand unchanged. The premise is recorded as refuted rather than the guidance
> authored to match it — this milestone's "inference stated as fact" rule, applied to the
> issue itself.

### LINKED ISSUE #716: Decision: should a timeout-captured error stream re-verdict to the #651 entry-load codes?

### LINKED ISSUE #717: Decision: launch_timeout's environment category misreads a caller-chosen ceiling
## What to decide

`launch_timeout` is registered with category `environment` (ADR-0002). Since #655, the ceiling is frequently a deliberate, caller-configured `--timeout` — a small value chosen to fail fast, where reaching it is a normal, expected outcome of probing a slow suite, not an environment fault. An agent branching on the category is misled toward environment remedies (retry, reinstall, different host) when the correct remedy is "raise --timeout or read the captured partial output".

Options include: keep the single code and category (documented nuance only); a category change (breaking for consumers keying on it); or a distinct code for the caller-configured case. Interacts with #687's ADR-0004 structured-failure-evidence discussion — if envelopes gain typed fields, the "who chose the ceiling" fact could ride there instead.

## Message sites (verified 2026-08-27) — "reword the guidance" is not one edit

There are **three** distinct sites, and they need three different answers. An
implementer reading only the body above (framed around `script run`) would edit the
wrong one:

1. `src/gda/errors.py:666-689` — `script_run_timeout_failure`, used by `script run`.
   **Already correct**: it names the reached `--timeout`, the elapsed wall clock and the
   termination phase, and ends with "raise `--timeout` for a run that is merely slow, or
   declare `--completion-marker`". Leave it alone.
2. `src/gda/commands/resource.py:1107` — `resource import --timeout` falls straight
   through to the shared generic message. This is the site that genuinely lacks
   caller-remedy framing, and it needs its OWN wording rather than a copy of
   `script run`'s: `resource import` uses the buffered capture strategy, so unlike
   `script run` it has no partial output to point the caller at.
3. `src/gda/errors.py:220-225` — the shared `classify_launch_or_crash` message
   ("Godot launched but did not return before the timeout"). **Explicitly out of scope
   for the reword.** This message is also what every sentinel-dispatched command gets on
   its fixed, non-configurable 60s ceiling (`runner.py:35`). Only four commands expose
   `--timeout` at all — `script run`, `scene preflight`, `resource import`, `daemon`
   (`grep -n '"--timeout"' src/gda/commands/*.py`) — so telling every caller to "raise
   `--timeout`" here would name a flag most of them do not have.
   - Existing seam worth using instead of a bespoke message: `launch()` already takes a
     `timeout_label` (`runner.py:561`; `"Godot import"`, `"Godot script"`, `"Godot scene
     preflight"`, `"Godot export"`), but today it reaches only `raw.stderr`
     (`runner.py:740`) and never the failure message. Making the shared message
     label-aware is the smaller change.

**Precedent the decision should acknowledge:** `scene preflight` (`scene.py:844-849`)
already resolved this same tension the OPPOSITE way — reaching its caller-configured
`--timeout` is a success-shaped `status: "timeout"` verdict, never a `launch_timeout`
envelope. Both shapes are defensible (`script run` / `resource import` have no bounded
pass/fail state at their ceiling; preflight's is inherently frame-bounded), but if the
decision does not say so, the next reader re-opens this question by diffing the two
commands.

## Acceptance criteria

- [ ] A recorded decision, coordinated with #687's outcome.
- [ ] If the registry changes, the ADR-0002 table and the equality-pinned test move in the same change.

## Blocked by

None - decision can be taken immediately (alignment with #687 recommended).

---

> **Scope correction (2026-08-31, P4 pre-dispatch review — verified on `main @ 181da20d`):**
> #714 (P3) delivered most of what this issue asks for, and the remaining work is smaller than the body above describes:
>
> 1. **The buffered capture strategy no longer exists.** Every launch streams; a `resource import` timeout DOES carry tail-capped partial output now — point 2's premise ("no partial output to point the caller at") is retired.
> 2. **`resource.py`'s launch call already passes `timeout_label="Godot import"`** — the very line point 2 cited as falling through to a generic message is the line that labels it.
> 3. **The "smaller change" point 3 proposed — a label-aware shared message — SHIPPED.** The shared branch is now `launch_timeout_failure` (`errors.py:338-382`), one home for all three channels, already naming the label, the ceiling, the elapsed wall clock, and the captured output. Do NOT fork a per-channel message inside it.
>
> **What remains of this issue:** the environment-category decision (point 1, unchanged) and the remediation WORDING of that one shared function — caller-first ordering (raise `--timeout` before "check the binary") — plus the echo sentences in the catalog, README, and Skill. The old `errors.py:220-225` / `runner.py:561`/`:740` anchors are all stale; the two lines the plan once said to "leave alone" (`errors.py:666-689`) are now `project_not_found` factories owned by the #763+#697 slice.

qinhaihong-red:
Three timeout-guidance issues in one slice: two contract decisions on `launch_timeout` and the live-operations guidance. `#717`'s residue and `#716`'s decision are the same registry row, so they are decided and recorded together; `#684` turned out to rest on a premise the engine refutes, so its guidance is corrected rather than written as asked.

Closes #716
Closes #717
Closes #684

## Decision 1 (#716) — a timed-out run's captured stream stays ADVISORY

**Decided for all four launch-backed channels** (sentinel, export, `resource import`, `script run`), which is what #714 changed: since it, every one of them carries captured stderr under `launch_timeout`, so the question is no longer `script run`'s alone.

A recognized entry-load error inside the capture stays an advisory diagnostic; it never re-verdicts the code into a #651 verdict. Only one of the two facts is gda's own — it **observed** that it stopped waiting, and would have to **infer** the entry-load cause from bytes the child happened to have written by then. That inference is unsound by construction: the capture is tail-capped and was cut mid-flight, so a recognized line can be an earlier phase's error the run survived, or half a line — and the misattribution would be silent, the worst shape for an agent branching on `code`. The need a re-verdict was meant to serve is the cause as **data**, which #687 owns.

Both timeout messages now state the rule in one clause, so a caller reading the diagnostics does not re-verdict on gda's behalf either. It matters most on `script run`, whose diagnostics literally open with `gda: recognized script errors seen before the timeout:`.

**Recorded:** a dated `> Outcome (2026-08-31, #716 / #717)` note beside ADR-0002's `launch_timeout` registry row — the one place all four channels are governed — plus a pointer note appended after #714's `> Outcome (2026-08-31)` block in ADR-0031. No existing decision prose was rewritten.

## Decision 2 (#717) — the category STANDS; the remediation order was the defect

`launch_timeout` keeps its registered `environment` category and exit `124`. The same code also fires for a genuinely environmental hang, so no other category is true of the whole row; `category` is public ABI every consumer keys on; and a second code for the caller-configured case would double the branch surface to carry a provenance fact that does not change what to do next (that fact, if ever wanted as data, is #687's evidence object).

What was actually wrong is that the category alone sends an agent to retry / reinstall / another-host remedies for a ceiling it chose itself. So the wording in the **one** shared function (`launch_timeout_failure`) now reads caller-first:

```
Godot import launched but did not return before the timeout of 4.0s (elapsed 4.00s).
Reaching the ceiling is not by itself an engine or host fault: read the captured output
in diagnostics for how far the run got, and raise the ceiling (--timeout, where the
command exposes one) for a run that was merely slow — suspect the binary or the machine
only when the capture shows the engine never started. The capture is truncated to the
last 16384 UTF-8 bytes (16 KiB) of each stream, and any engine error in it is advisory:
the verdict here is the timeout.
```

`--timeout` is named **with its qualifier** deliberately: of that builder's three channels only `resource import` exposes the flag (the sentinel's 60s and the export's 600s are gda's own, fixed). An unqualified "raise `--timeout`" is exactly the misfire #717's point 3 warned about. No per-channel message was forked inside the function.

**Scope proof:** no registry, category, exit-code or schema change. `tests/test_error_registry.py` (the #701 drift pin) and ADR-0002's table row are **untouched and still green** — they did not need to move because neither side moved.

**Echoes** of the caller-first reading: `docs/command-catalog.md` (the `script run` paragraph, plus a `resource import` sentence), `README.md`'s exit-code table row for `124` (+ `zh-CN` / `es` / `ja` translations, restamped via `scripts/update_readme_i18n.py`), and `src/gda/skill/SKILL.md`'s exit table row for `124`.

`scene preflight`'s opposite resolution is **cited**, not re-authored: ADR-0002's existing "launch bound as a verdict" record (2026-08-19, #664) already owns it, and the new note points there so the next reader does not re-open the question by diffing the two commands.

## #684 — premise refuted; guidance corrected

The issue asks for the `SceneTree.suspended` dead-end to be documented and named as a likely `live_timeout` cause, on the premise that "a game that uses `get_tree().suspended = true` (freeze-frame) reproduces the exact GDA-DF-013 dead-end".

**A project cannot do that.** Verified two ways:

- *Real engine (Godot 4.6.3, via `gda script run`):* `has_method("set_suspend") == false`, `has_method("is_suspended") == false`, no `suspended` in `get_property_list()`, `ClassDB.class_has_method("SceneTree", "set_suspend", true) == false`, `ClassDB.class_get_property_list("SceneTree", true)` has no `suspended`. A running daemon session confirmed it too: `get_tree().suspended = true` in `_ready` produced `Invalid assignment of property or key 'suspended' with value of type 'bool' on a base object of type 'SceneTree'` in `gda logger tail`, and every live op kept serving normally.
- *Engine source (`../godot`):* `SceneTree::set_suspend` / `is_suspended` have no `ClassDB::bind_method` and no `ADD_PROPERTY`. The only callers are `SceneDebugger::_msg_suspend_changed` and `_msg_next_frame` (`scene/debugger/scene_debugger.cpp`), driven by the editor Game view's Suspend / step buttons (`editor/run/game_view_plugin.cpp`) over the remote-debugger channel — an editor-launched game, never a daemon-launched `Engine session`.

The engine mechanism the issue describes **is** real (`Node::can_process()` is `is_inside_tree() && !tree->is_suspended() && _can_process(is_paused())`, so no process mode ticks while suspended — confirmed at `scene/main/node.cpp:908`), and it is documented as a limit. But naming it as a likely `live_timeout` cause would send an agent after something its project cannot have done — the same class of misleading remediation this PR fixes for #717. So:

- the **Skill** gains a `paused vs suspended` note under the `#656` pause paragraph: what suspending would do, that it is unreachable from game code with the evidence, and that `paused` is the supported freeze;
- the **catalog** gains the same, with the engine citations, in the Phase-2 live intro;
- the **`live_timeout` message** gains the one diagnosis hint the issue asked for, aimed at the reachable cause: *"the game stopped returning to its main loop, so the gda harness cannot tick — look for a blocking loop or wait in game code. A paused SceneTree is NOT a cause: the harness serves through a pause. `gda diag errors` and `gda logger tail` still read."* Ruling out the pause is the valuable half — it is the first thing an agent watching a frozen game blames, and `#656` already made it wrong.

Only the daemon-side message changed. The CLI-side backstop in `live_runner` keeps its bare sentence: it is reached only when the *daemon* stops answering, which this cause does not produce.

`AC2` as literally written ("names a suspended SceneTree among the likely causes") is therefore **not** implemented, on evidence. Flagging it for the reviewer rather than quietly satisfying the checkbox.

## Surface diff (help / skill / message)

- `launch_timeout` message (shared builder — sentinel, export, `resource import`): reworded, caller-first, + advisory clause.
- `script run`'s own `launch_timeout` message: one appended advisory clause; ceiling / elapsed / phase / code / exit unchanged.
- `live_timeout` message (daemon relay): cause class named, pause ruled out.
- No `--help` text, no schema, no registry, no exit code, no result model changed.

## Verification

- `uv run pytest -m "not e2e" -q` → **2254 passed**; `ruff check` / `ruff format --check` clean; `pyright` → **0 errors**.
- Real-engine e2e: `tests/test_e2e_launch_timeout.py` (3 arms — wedged sentinel op, wedged export, wedged `resource import` through the full CLI) and `tests/test_e2e_script_run.py` → **50 passed**.
- Real-engine probe of the new envelope end to end (`gda resource import --timeout 4` against a project with a wedging editor plugin): `code=launch_timeout`, `category=environment`, the message above verbatim, engine banner + import progress in `diagnostics`.
- Real-engine probe of the #716 shape (`gda script run --timeout 4` on a script that prints a parse-error line then hangs): diagnostics carry `parse_error: Parse Error: …`, the verdict stays `launch_timeout`, message carries `not an entry-load failure`.
- Live path exercised on a real daemon session (`daemon start` → `wait-ready` → `game tree` → `logger tail` → `daemon stop`) during the #684 investigation.

**Red-proof** — every new guard was seen failing on the pre-change source:

| Guard | Reverted | Result |
| --- | --- | --- |
| `test_the_timeout_remediation_reads_caller_first` | shared message | FAIL |
| `test_the_timeout_message_says_a_captured_error_is_advisory` | shared message | FAIL |
| `test_a_timeout_says_its_recognized_script_errors_are_advisory` | `script run` message | FAIL |
| `test_the_live_timeout_message_names_its_cause_and_rules_out_a_pause` | `live_timeout` message | FAIL |

The e2e `_assert_caller_first_remediation` helper is applied to all three real-engine channels, so the published sentence — and its **order**, which is the contract — is pinned on a real engine, not only at the builder.


qinhaihong-red:
## Independent adversarial review — head `a5cdc1b672`, base `181da20d`

Reviewed by pinned SHA in a throwaway detached worktree (never the author's checkout). Every claim below was probed on a real Godot **4.6.3** (`4.6.3.stable.official.7d41c59c4`) and on engine source (`../godot`, `4.6-stable-3260-g070dc9897e`).

| Axis | Verdict |
| --- | --- |
| Artifact / Markdown | **Changes required** (F1 echoes, F3, F4, F5) |
| Architecture / DDMA | **Pass** (one observation) |
| Standards | **Pass** |
| Spec | **Changes required** (F1, F2) |

Gates re-run here, all green: `pytest -m "not e2e"` → **2254 passed**; `ruff check` / `ruff format --check` clean; `pyright` → **0 errors, 0 warnings**; `pytest -m e2e tests/test_e2e_launch_timeout.py` → **3 passed in 13.12s** on a real engine.

---

## Adjudications

**1. #684's premise refutation — SOUND, and the disposition is right.** Confirmed independently, two ways:

- *Engine source:* `SceneTree::_bind_methods` (46 binds) contains **no** `set_suspend` / `is_suspended` bind and no `ADD_PROPERTY` for `suspended`; only `set_pause` / `is_paused` / `ADD_PROPERTY(..., "paused", ...)`. The only writers of `set_suspend` repo-wide are `scene/debugger/scene_debugger.cpp:215,779,780`; the only other reader is `scene/debugger/runtime_node_select.cpp` — also debugger-side. `Node::can_process()` is verbatim as cited (`scene/main/node.cpp:908`).
- *Real 4.6.3, via `gda script run`:* `has_method("set_suspend")=false`, `has_method("is_suspended")=false`, `get_property_list()` has no `suspended` (but has `paused`), `ClassDB.class_has_method("SceneTree","set_suspend",true)=false` (`set_pause`=`true`), `ClassDB.class_get_property_list("SceneTree",true)` has no `suspended`.

Refusing to name an unreachable state among "likely causes" is the correct call, and flagging the AC miss openly rather than satisfying the checkbox is the right disposition. **The defect is in the substitute** — see F1.

**2. #716 — reasoning sound, placement right, composes with #796 with one overstatement.** Inference-vs-observation is the correct axis, and it is the only axis that survives the tail-cap: the capture is cut mid-flight, so a recognized line carries no proof it was terminal. Deciding once for four channels rather than per-channel is right, and the blockquote after the registry table is the closest a Markdown table permits to "beside the row". Read #796: it states the ADR-0002 note "is left untouched … this change is what makes its advisory-capture rule workable" — no contradiction. But see **F4**.

**3. #717 — scope proof verified.** `launch_timeout_failure` has **one** message with no per-label branch (the only channel-varying token is `TimeoutBound.label`), so the shared branch was not forked. The `--timeout` qualifier is factually right: of the builder's three channels only `resource import` exposes it (`resource.py:1242`); `scene preflight` intercepts `LaunchFailure.TIMEOUT` at `scene.py:1012` *before* `classify_run` and returns its `timeout` **status**, so it never reports this code; `daemon wait-ready --timeout` reports `engine_session_not_running`. `tests/test_error_registry.py` and the ADR-0002 row are absent from the diff and green.

**4. `_assert_caller_first_remediation` — verified, it does what the body claims.** It is applied to all three real-engine arms (sentinel + export through `_assert_timeout_envelope`, `resource import` inline on the CLI round trip), and it pins **order**, not just presence, via `message.index(...)` comparison; a missing sentence raises rather than passes. Independently reproduced end to end:

```
$ gda resource import res://icon.png --timeout 4 --project <wedged>   # exit 124
code=launch_timeout  category=environment
Godot import launched but did not return before the timeout of 4.0s (elapsed 4.00s). Reaching
the ceiling is not by itself an engine or host fault: read the captured output in diagnostics
for how far the run got, and raise the ceiling (--timeout, where the command exposes one) ...
```
diagnostics carried the engine banner, the import progress and `PLUGIN WEDGED`.

**5. `live_timeout` — the paused-tree exclusion is CORRECT; the named cause class is NOT.** The exclusion checks out: `gda_harness.gd` sets `process_mode = Node.PROCESS_MODE_ALWAYS` inside the daemon-launched gate, and `Node::can_process()` consults `is_suspended()` before the process mode, so a *paused* tree keeps the serving loop ticking. Ruling out the pause is the valuable half and it is true. The other half is F1.

---

## Findings

### F1 — P1: the new `live_timeout` message states one inferred cause as fact, and it is reproducibly false

`src/gda/daemon/session.py:184` now publishes, unhedged:

> the game stopped returning to its main loop, so the gda harness cannot tick — look for a blocking loop or wait in game code.

gda observed exactly one thing: 30s passed with no reply. "The game stopped returning to its main loop" is an **inference** — and this PR's own #716 half is the argument against making it ("gda … OBSERVED that it stopped waiting, and would have to INFER … the misattribution would be silent, which is the worst shape for an agent"). The two halves of this PR take opposite epistemic stances on the same kind of evidence.

It is not a theoretical gap. `OP_TIMEOUT` is a fixed 30s (`session.py:33`) while the harness's **multi-frame window base** (`_begin_window`, #223/#221/#661) accumulates up to `MAX_WINDOW_FRAMES = 600` *engine frames* with, by design, no bound of its own — `gda_harness.gd:307` says so explicitly: *"the window has no timeout of its own — the daemon-level `live_timeout` is the stalled-engine guard."* On any game ticking below `frames / 30` FPS, that guard fires on a perfectly healthy game.

**Reproduced** (real daemon, real 4.6.3, headless). Project whose `_process` does `OS.delay_msec(400)` and prints a heartbeat every 5 frames:

```
$ gda game tree            -> {"root":{"name":"Main",...}}          # session serving
$ gda perf monitor /root/Main --property z_index --frames 100
{"error":{"category":"live","code":"live_timeout","message":"the engine session did not return
 within 30s; the game stopped returning to its main loop, so the gda harness cannot tick —
 look for a blocking loop or wait in game code. ..."}}              # 09:38:31 -> 09:39:01
$ gda logger tail          -> 39 heartbeats, HEARTBEAT 5 t=1927 ... HEARTBEAT 195 t=79164
```

190 frames over 77.2s = **2.46 fps, sustained**; 39 records at one per 5 frames means 38 intervals averaging 2.03s — arithmetically there is **no gap** anywhere in the log, including the 30s that produced the timeout. The main loop never stopped. The real cause is `100 frames / 2.46 fps = 40.7s > 30s`, and the real remedy — **lower `--frames`** — is the one thing the message does not mention, while the remedy it *does* give sends an agent hunting a blocking loop that does not exist.

That is the same class of misleading remediation this PR fixes for #717 and the reason it declined #684's AC2. A second reachable class exists too: `live_numbers.py:194` / `command-catalog.md:1121` document an unparseable frame producing `live_timeout` with the game ticking normally.

Same false claim is echoed at two more sites, so a fix has three: `SKILL.md:147` ("It means the game stopped returning to its main loop") and `SKILL.md:243` ("a `live_timeout` means the game stopped returning to its main loop, not that it is paused").

Minimal fix that keeps #684's valuable half: hedge the claim to what gda actually knows and name the second class with its own remedy — e.g. *"…did not return within 30s. Most often the game stopped returning to its main loop (a blocking loop or wait in game code); a multi-frame window op (`perf monitor`/`perf monitors --frames`, `input sequence`, `screen capture --await-*`) can also outrun this bound on a slow-ticking game — lower `--frames`. A paused SceneTree is NOT a cause…"* The pause exclusion, the `diag errors` / `logger tail` pointer and the `suspend` omission all stay exactly as they are.

### F2 — P3: for two of the three shared-builder channels the caller-first remediation dead-ends

`launch_timeout_failure` tells every caller to "raise the ceiling (`--timeout`, where the command exposes one)". The qualifier is honest, but the sentinel (60s) and export (600s) channels expose no flag, so an agent that follows the sentence to its end — reads the capture, sees the engine *did* start, so declines to suspect the binary or the host — is left with no next step at all. The ADR note argues the provenance fact "does not change what to do next"; the message's own qualifier concedes that it does. `TimeoutBound` already carries `label` and `seconds` and is the natural seam, but a per-channel branch is out of scope here by #717's own instruction — so the cheapest honest fix is one more clause for the no-flag case (*"a command with no `--timeout` is bounded by gda; a run that reaches that bound is stuck rather than slow"*), or an explicit note in the ADR that the fixed-ceiling channels are knowingly left without a caller remedy.

### F3 — P3: the ADR-0031 pointer sends the reader to the wrong amendment

The new note (line 477) says *"The amendment at the top of this ADR records a `script run` whose entry never loaded reaching `launch_timeout` … **and defers whether the captured stream should narrow it**."* The first half is right (line 19, the #651 amendment). The deferral is not there — it is in the **2026-08-17 (#655)** amendment at line 370: *"the shape this decision records as failing 'by another route' keeps doing so, and narrowing it is a separate decision."* A pointer whose whole job is to stop the next reader re-deriving the question should name #655's amendment, or both.

### F4 — P3: the ADR-0002 note overstates what #687 delivers

> "#687 owns that: typed evidence on the envelope keeps the honest verdict AND delivers the precise cause."

Per the stacked PR #796, `script_errors` — the "precise cause" — ships only for `script run`'s timeout / `script_aborted` and its script failures. The shared builder's `launch_timeout` (sentinel, export, `resource import`) gets `elapsed_seconds` / `timeout_seconds` / `termination_phase` only. Since #716 is decided **for all four channels** on the promise that the data need is served elsewhere, the note should say so precisely (e.g. "…delivers the precise cause where the channel parses one — `script run` today") rather than leave a durable ADR sentence that three of the four channels do not satisfy.

### F5 — P3: two echo sites, two smaller nits

- The `124` rows in `README.md` and `SKILL.md` now carry ~55-word cells in tables whose every other row is a short clause. In `SKILL.md` especially the table is the *scan* surface ("Branch on the stable `category`/`code` and the **exit code**"); the prose belongs under it, not inside the cell. The three translations are correct and the i18n stamp is honest — I recomputed `sha256(README.md)` = `dca00ff7f1f42a591e21f18f0bcea603b60f7154d1f4d03ab7e5525f148d2a80`, matching all three headers.
- `test_daemon_server.py`'s `assert "suspend" not in error["message"].lower()` pins the **absence of a word** rather than a behaviour, so any future rewording that legitimately mentions a suspended tree (including one that cites the refutation) turns red for no defect. The positive assertions above it already encode the decision.

### Observation (no change requested)

`live_timeout` now has two materially different messages for one code — the daemon relay's new paragraph and `live_runner.py:92`'s bare `"the live operation did not return within 60s"`. The in-code justification (the CLI backstop is reached only when the *daemon* stops answering) is correct, and I am not asking for a fork or a merge here; but it is the same shape as the `live_unsupported_platform` wording split already on the backlog, and an agent branching on `code` cannot tell which producer wrote what it is reading.

qinhaihong-red:
## Review handled — head `21869a7b`, all five findings addressed

Thanks for the real-daemon counterexample on F1; it was right, and I reproduced it independently before changing anything (numbers below).

| # | Decision | Where |
| --- | --- | --- |
| F1 (P1) | Adopted, with one deliberate narrowing | `session.py:186`, `SKILL.md` ×2, catalog |
| F2 (P3) | Adopted | `errors.py:402` |
| F3 (P3) | Adopted | ADR-0031 |
| F4 (P3) | Adopted, and the first fix's own overclaim caught | ADR-0002 |
| F5 (P3) | Adopted, both halves | README + 3 translations, `SKILL.md`, `test_daemon_server.py` |

---

### F1 — hedged, and the window-op cause named with its remedy

The message now leads with what gda observed and hedges the cause it used to assert:

> the engine session did not return within 30s — gda observed the silence, not its cause. Most often the game stopped returning to its main loop, so the gda harness cannot tick: look for a blocking loop or wait in game code. A second cause leaves the loop running — a multi-frame window counts ENGINE frames against this fixed wall clock, so a slow-ticking game outruns it: ask for fewer frames (`--frames`, `--await-frames`). A paused SceneTree is NOT a cause: the harness serves through a pause. `gda diag errors` and `gda logger tail` still read — a log that kept advancing through the wait rules out a stalled loop.

The pause exclusion, the `suspend` omission and both pointers are unchanged; the last pointer gains the check that discriminates the two causes. Both echo sites are corrected (`SKILL.md:152`, `SKILL.md:250`), as is the catalog's `paused vs suspended` note.

**One narrowing, argued rather than assumed: the unparseable-frame class is NOT named to the caller.** Your finding cites it as a second reachable class, and `live_numbers.py` / catalog:1121 do document it — but both describe it in the past tense, as the justification for a refusal. Today every reproduced value class (non-finite floats, out-of-safe-range ints, floats the engine parses as `0.0`) is refused BEFORE the write by `RelayedLiveParams._admit_live_wire_numbers` (`models.py:571`) over `find_unrepresentable`, and `tests/test_live_contract_guards.py` fails both ways on that inheritance. What is left is residue with no caller remedy — and naming an unactionable class with no remedy is precisely #684's shape in a third direction. So it is recorded in the code comment as *why the message hedges rather than enumerates*, not published as something to go check. Your own suggested minimal wording names only the window-op class, so I read this as consistent with the finding rather than a departure from it; happy to add it if you disagree.

Two smaller corrections while implementing your suggestion: I dropped the op enumeration (`perf monitor --frames`, `input sequence`, …) — it omitted `screen frames`, and a flag-named class (`--frames` / `--await-frames`) is both complete and shorter — and I checked every flag I kept exists.

### F2 — the fixed-ceiling channels get a next step

Added after the qualifier, before the host suspicion: *"Where the command exposes no `--timeout` the ceiling is gda's own and cannot be raised: reduce the work or give the machine more headroom."* It deliberately stops short of your alternative wording *"stuck rather than slow"* — that is the same unearned inference F1 is about, from the same evidence.

### F3 — right amendment named

Now: *"Whether the captured stream should NARROW it is deferred a step later, in point 3 of the 2026-08-17 (#655) amendment: 'narrowing it is a separate decision'."* The #651 amendment keeps the *"failure by another route"* half, so both are named as you suggested.

### F4 — and the first attempt at it was itself an overclaim

Your finding was right, and worth flagging: the first fix I inherited replaced the overclaim with *"all four get `elapsed_seconds` / `timeout_seconds` / `termination_phase`"* — but `timeout_seconds` and `termination_phase` do not exist anywhere in this branch (`grep` over `src/`, `tests/`, `docs/` returns only that ADR line). They are #796's shape, stated in the present tense as shipped. The note now states the bound without predicting #687's field list:

> Its reach is BOUNDED, and this decision does not assume otherwise: a PARSED cause exists only where a channel parses one, which today is `script run` alone. The other three carry how the run ended — the ceiling it reached and its elapsed clock — and keep the capture in `diagnostics` as prose … What this decision needs from #687 is that the capture stops being the only evidence, not that every channel end up with the same evidence.

### F5 — both halves

The `124` cell in `SKILL.md` and in `README.md` is now one clause pointing at a **Reading a `124`** paragraph under each table, which keeps the `--timeout` qualifier, F2's new clause and the advisory rule that a pure compression would have dropped from the README. The three translations carry the same new paragraph (translated, not copied) and are restamped — `sha256(README.md)` = `6c11b5d19ab6c2261d2d6757a46b9d771ff4a74bdced94d057d6fe65332cce80`, matching all three headers.

The word-absence assertion is gone; the guard now asserts what the message should say, including the hedge and the window-op remedy.

---

### Validation

**Red-proofs** (mutate, run, restore — the guards are load-bearing, not decorative):

| Mutation | Result |
| --- | --- |
| `live_timeout` message reverted to its pre-review text | `test_the_live_timeout_message_hedges_its_causes_and_rules_out_a_pause` FAILS on the hedge |
| hedge kept, only the window-op clause removed | same test FAILS on that clause alone |
| the no-`--timeout` clause removed from `errors.py` | `test_the_timeout_remediation_reads_caller_first` FAILS |

**Real daemon, real Godot 4.6.3, headless** — your scenario reproduced independently, then the new message read off the wire:

```
positive control: perf monitor --frames 2   -> success in 0.36s (no spurious fire)
perf monitor --frames 100                   -> live_timeout at 30.4s
logger tail: 126 heartbeats, HEARTBEAT 5 t=2358 ... HEARTBEAT 630 t=254848
             2.48 fps sustained, max inter-record gap 2.06s -> no stall anywhere
             100 frames at that rate = 40.4s > the 30s bound
diag errors                                  -> {"errors":[]} (both pointers read)
```

The message came back verbatim, naming the window-outrun cause and the remedy that actually resolves it.

**Gates**: `pytest -m "not e2e"` → 2254 passed; `ruff check` / `ruff format --check` clean; `pyright` → 0 errors, 0 warnings; `pytest -m e2e tests/test_e2e_launch_timeout.py` → 3 passed.

### Observations, no change made

- **A pre-existing cold-cache flake in `test_e2e_launch_timeout.py`.** The export arm failed once on the first run of the session — `assert 'PLUGIN REACHED' in diagnostics`, with only the engine banner captured; it passed on the next three runs. Not from this change: `_assert_caller_first_remediation` runs BEFORE that content check inside `_assert_timeout_envelope` and passed in the failing run, so the new assertions were satisfied. It looks like the first run paying the import pass inside the 4s ceiling. Reporting rather than fixing — it is outside this PR's scope.
- Your `live_timeout` two-producer observation stands unaddressed by design, alongside the `live_unsupported_platform` wording split already on the backlog.

qinhaihong-red:
## Full independent adversarial review — PR #793

Reviewed from first principles at pinned head `21869a7b0c4352051560708848157f908a49f72f` against `feat/gda-dogfooding-dev@181da20d2ea418977600bd73d2958dc4308e8de6`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Standards | **Pass** |
| Spec (#716, #717, #684) | **Changes required** |

Issue #684 remains internally authoritative but contradicts the verified decision and this PR's implementation. It says project code can set `get_tree().suspended = true` and requires `live_timeout` guidance to name suspension as a likely cause. The PR instead records and implements the opposite engine/API finding: `SceneTree::set_suspend` / `is_suspended` are not available to project GDScript or ClassDB; `src/gda/daemon/session.py:184-207` therefore names reachable causes and explicitly rules out `paused`, while `src/gda/skill/SKILL.md:244-254` explains the suspended mechanism.

Do not add a false hint to satisfy stale text. Reconcile #684 with the verified Godot 4.6.3 evidence: record the refuted premise, the accepted paused/suspended guidance, and the reachable timeout causes; then update its acceptance criteria before this PR closes it.

Pass. Shared launch-timeout classification/remediation stays in `gda.errors`; daemon live-operation timeout remains owned by `EngineSession`, a distinct runtime boundary. The PR does not duplicate verdict ownership or collapse unlike timeout state machines.

### Standards

Pass. No non-tool-enforced repository-standard violation or actionable Fowler smell was found.

### Spec — required change

#716 and #717 are implemented as requested. #684 is not: its AC2 literally requires the live-timeout text to name a suspended tree, while the current head deliberately and correctly excludes it based on engine evidence. The smallest valid fix is the issue reconciliation above; authoring knowingly false runtime guidance is not an acceptable alternative.

Sources: `CONTEXT.md` definitions for **Engine session**, **Live operation**, and **State consistency**; ADR-0002/0031; issues #716/#717/#684; changed README/catalog/Skill text; relevant Godot `scene_tree.cpp` and `scene_debugger.cpp` evidence. Writing reference: ASD-STE100. All changed decision, requirement, remediation, translation, and paused/suspended passages were inspected. The stale issue premise is the only substantive artifact problem.

Validation: `git diff --check` passed. Current hosted CI has no failed or pending required check; Godot e2e is skipped. The engine-source evidence was inspected, but the Godot 4.6.3 probes and translation generator were not rerun.

qinhaihong-red:

Current head: `21869a7b` — unchanged; this round's one finding is resolved on the issue, not the branch.

| Issue #684 stays internally authoritative for the refuted premise; its AC2 literally requires naming a suspended tree in the live-timeout text, contradicting the verified decision this PR implements | **Fully adopt** (on the issue) | Godot 4.6.3 source: `SceneTree::set_suspend`/`is_suspended` carry no ClassDB binding — project GDScript cannot reach `suspended`; sole callers are the debugger agent (`scene_debugger.cpp`), which a gda-driven run does not attach | #684 amended: AC2 superseded in place, dated **premise-reconciliation note** records the refuted premise, the delivered paused/suspended guidance (`SKILL.md`), and the reachable-cause live-timeout text (`session.py`); AC1/AC3 stand | Issue body diff; no code or artifact in this PR changes, per the review's own instruction not to author a false hint |

No other axis required changes (DDMA/Standards/Spec-#716/#717 passed). The PR now closes #684 against the reconciled acceptance criteria.

qinhaihong-red:

Reviewed at `21869a7b0c4352051560708848157f908a49f72f` against base `181da20d2ea418977600bd73d2958dc4308e8de6`.

**Verdict: Changes required**

- Artifact: **Changes required**
- DDMA: Pass
- Standards: Pass
- Spec: Pass

### Required — issue #684 still gives mutually incompatible current directions

The new premise-reconciliation note and the superseded AC2 correctly record that project GDScript cannot cause `SceneTree.suspended` and that live-timeout guidance must name only reachable causes. However, the issue title still says to “name it in the timeout guidance,” the opening source still asserts that a game can reproduce the timeout with `get_tree().suspended = true`, and the active-looking “What to build” section still instructs the implementation to name a suspended tree as a likely cause.

Those passages directly contradict the verified premise and the amended acceptance criterion. A reader arriving through the title, search results, or executable work section is still directed to implement guidance the issue now says is false.

**Smallest fix:** retitle #684 and strike or rewrite the obsolete reproduction premise and the corresponding “What to build” bullet in place. Keep the dated reconciliation note as the historical record.

The PR implementation itself remains consistent with the corrected Godot 4.6.3 finding. Current required CI has no failed or pending checks.

qinhaihong-red:
## Round-3 response

Current head: `21869a7b` — unchanged; the finding is on issue #684 and is resolved there.

| #684's title, opening premise, and "What to build" bullet still actively direct the refuted guidance; only AC2 and the appended note were reconciled | **Fully adopt** | Verified: the title said "name it in the timeout guidance"; the source paragraph asserted `get_tree().suspended = true` as a game-reachable reproduction; the second build bullet still instructed naming suspension | #684 **retitled** ("…timeout guidance names reachable causes"); the reachability claim struck in place with the refutation marker (the dead-end MECHANISM kept, attributed to the debugger-only path); the build bullet superseded in place; the lead-in reworded ("reachable-cause timeout guidance"). The dated premise-reconciliation note stays as the historical record | Issue title + body diff; a reader arriving through the title or the work section now meets only the corrected directions |

The PR branch remains as reviewed.

PR #795, opened 2026-08-31
fix(gda): decide the res:// walk's symlink policy by filesystem identity (#760)

### LINKED ISSUE #760: res:// walk has no symlink policy: an alias re-admits the engine cache, a cycle emits fabricated paths

qinhaihong-red:
Closes #760

## The decision

The `res://` walk had no symlink policy: it compared paths lexically and descended into whatever
the directory listing called a directory. Two defects followed — an alias re-admitted the engine
cache the walk excludes, and a cycle was descended until the OS refused another symlink hop.

**The walk now FOLLOWS a link, as the engine does, but identifies what it reaches by FILESYSTEM
IDENTITY rather than by the spelling that reached it.** Two rules follow, and only a link pays for
them:

1. **Aliasing rule — the engine cache is excluded by identity.** No symlink alias re-admits it: not
   a directory link AT `res://.godot`, not a directory link INTO one of its subdirectories, not a
   FILE link at a file inside it, and not any of those three reached under a **second spelling** of
   a parent directory that is itself a link. `DirAccess.is_equivalent` — the engine's own
   `(st_dev, st_ino)` / `(volume, file id)` test — answers the identity question, and the path it
   is asked about is resolved **component by component**, each read against the components already
   resolved before it. `_is_in_engine_cache` is its single owner; both walk-side predicates route
   the question to it. A **hard** link is outside the rule by construction — the filesystem does
   not report one as a link — which the catalog and the code now state rather than implying "no
   alias".
2. **Cycle termination — a linked directory already on the current descent CHAIN is not
   re-entered.** `_collect_paths` carries the chain of ancestors (a new array per branch, so a
   branch cannot see a sibling's), and the cycle terminates *by rule* instead of running out at the
   OS symlink limit. The chain being per-branch, what the listing enumerates is distinct `res://`
   **paths**: a directory reachable through several link paths is reported under each. The answer is
   decided and finite — that is the guarantee — not "each real directory exactly once".

**A link to ordinary authored content is untouched by both.** A vendored checkout that physically
lives outside `res://` and is reached through a directory link inside it is walked and enumerated
as an ordinary directory, and its `class_name`s resolve at the instantiating call sites —
including a checkout that carries an engine cache of its **own**, which is that checkout's cache
and not this project's. The policy decides **where a link leads**, not whether links are allowed
(AC1's "do not break a real workflow silently").

**Enumeration is not targeting.** Being reached by this walk says what the project can *address*;
whether gda will operate on a path once it is NAMED as a target is the separate question the
containment gate answers, from the caller's own spelling. The catalog passage and ADR-0032's
amendment both say so, so the two rules read as layers rather than as opposite answers.

One consequence is kept deliberately: a FILE link at authored content is enumerated under **both**
paths, since both are real `res://` addresses the engine loads. When the linked script declares a
`class_name`, those two paths make it `ambiguous_class_name` — the same report gda already gives
any project that declares one name twice.

### Two touch points, not one

The aliasing rule lands in **two** places, per the issue's 2026-08-31 landing-site correction:
`_should_descend` gates DIRECTORY descent only, so a FILE link into the cache
(`res://alias_cache.gd` → `res://.godot/root_cache.gd`) reaches the acceptance branch without ever
passing it. The traversal therefore asks a file-side predicate too (`_should_collect`). Red-proof 3
below isolates exactly that half.

### What the review round changed (`2bd3160` → `79ac13d`)

The first commit's `_resolved_link_path` joined a **relative** link target onto the link's `res://`
spelling instead of onto the directory the kernel reads it from. Those differ as soon as the link's
own parent is a link, so the policy still decided by spelling — in **both** directions:

- `sub/deep/c -> ../../.godot` plus `link1 -> sub/deep`: `res://sub/deep/c/...` excluded,
  `res://link1/c/...` **admitted** — the same directory entry, two answers. All four collectors,
  `project statistics` and the `class_name` index leaked the root cache under the second spelling;
- `vendored/x -> ../.godot` inside a checkout outside `res://`: the lexical join produced
  `res://.godot`, so the CHECKOUT's own nested cache was taken for the project's and silently
  **hidden** — a regression against base `181da20d`, and exactly the workflow #712 decided to walk.

The path is now resolved as the kernel resolves one — component by component, the pass repeated
until it changes nothing so a target that itself names a link is resolved too. Resolving only one
more component does not suffice: a link through a link (`L2 -> link1/c`) survives it.

A **second, independent** leak is closed by a deletion: `_is_in_engine_cache`'s ancestor climb had
borrowed `SYMLINK_PROBE_MAX_STEPS` as a PATH-DEPTH bound, and exhausting it returned `false`, which
**admitted** cache content — an alias 32 levels below the cache leaked while 31 did not. Path depth
is not a symlink count; the climb now ends at the root.

## Engine evidence (measured on Godot 4.6.3.stable, source read at the `4.6.3-stable` tag)

- **Links are followed by the engine, so the walk must decide where they lead rather than whether
  to allow them.** `DirAccessUnix::get_next` `stat()`s a `DT_LNK` entry on purpose, so a linked
  directory reports `current_is_dir()` (`drivers/unix/dir_access_unix.cpp:148-183`), and
  `ResourceLoader` loads through an alias.
- **`DirAccess.is_equivalent` is the engine's own identity test** and is bound to script
  (`core/io/dir_access.cpp:681`). `DirAccessUnix::is_equivalent` compares `(st_dev, st_ino)`
  (`drivers/unix/dir_access_unix.cpp:713-729`); `DirAccessWindows::is_equivalent` compares
  `(VolumeSerialNumber, FileId)` (`drivers/windows/dir_access_windows.cpp:411-431`); both fall back
  to `DirAccess::is_equivalent`'s **string equality** when a path cannot be `stat`'d
  (`core/io/dir_access.cpp:630-632`). That fallback is why a project with no `res://.godot` at all
  keeps answering exactly as it did.
- **The path is resolved by hand because `DirAccess` cannot do it for us.**
  `DirAccessUnix::fix_path` simplifies a path LEXICALLY before every syscall
  (`drivers/unix/dir_access_unix.cpp:55-57`), so a `..` appended to a link never reaches the kernel
  and cannot walk up out of an alias — measured: `is_equivalent("res://nested/.godot/..", "res://")`
  answers `false` through an alias of the root cache. `DirAccessUnix::read_link` gives the raw
  target relative to the link's own directory (`drivers/unix/dir_access_unix.cpp:481-499`).
- **`DirAccessWindows::read_link` has a different contract**
  (`drivers/windows/dir_access_windows.cpp:444-462`): it returns an already-resolved absolute path
  from `GetFinalPathNameByHandleW`, and on failure returns the fixed input path, **never** the
  empty string. The resolver therefore also stops when a target names the path it came from, so an
  unopenable reparse point costs one probe instead of the whole bound.
- The probe bound is `32`, the **lower** of the two ceilings gda targets: `MAXSYMLINKS` is 32 on
  macOS (`MacOSX.sdk/usr/include/sys/param.h:197`) and **40** on Linux (`include/linux/namei.h`).
  It bounds symlink traversals only — the link hops in one chain and the resolution passes, each of
  which resolves at least one more link. It does **not** bound the ancestor climb, which counts
  path components.

## Tests

`tests/test_e2e_project_walk.py` carries the original fixture (directory alias AT the cache,
directory alias INTO it, two FILE aliases at cache files, a cycle, and the legitimate vendored
checkout) and gains two more, built with **relative** link targets and links nested inside linked
directories — the shapes no existing test exercised, which is why the first commit's defect was
invisible to a green suite:

- `aliased_spelling_project` — one real directory under two `res://` spellings (`link1 -> sub/deep`),
  three relative links into the cache inside it, and `L2 -> link1/c`, a link through a link;
- `vendored_cache_project` — a checkout outside `res://` carrying its own `.godot`, reached through
  `vendored -> <outside>/v` with `v/x -> ../.godot`;
- a deep-alias case pinning the ancestor climb past the old step bound.

The planted cache scripts now `extends Node` so the `class_name`-index assertion can **fail**:
extending `RefCounted`, a leak and a clean walk both answered `invalid_node_type` with the class in
the message, so that guard asserted an outcome no regression could change. Node-derived, a leak
makes `node add --type RootCacheThing` **succeed**. `tests/test_project_walk.py` keeps the two
source-level guards: the traversal must ask both predicates, and `ENGINE_CACHE_DIR` may be compared
only in the descent predicate's lexical fast path and in `_is_in_engine_cache`.

## Verification

All run locally against Godot `4.6.3.stable.official.7d41c59c4`, on head `79ac13d0`.

| Gate | Result |
| --- | --- |
| `uv run pytest -m "not e2e" -q` | 2252 passed, 598 deselected |
| `uv run ruff check .` | All checks passed |
| `uv run ruff format --check .` | 481 files already formatted |
| e2e: `test_project_walk.py` + `test_e2e_project_walk.py` | 12 passed |
| e2e: `classname_resolution`, `project`, `project_analysis` | 60 passed |
| e2e: `script`, `scene` | 91 passed |

**Cost unchanged.** `project statistics` over `examples/platformer/panda-adventure`, 3 runs each:
base `181da20d` 0.61 / 0.61 / 0.64 s, head `79ac13d0` 0.61 / 0.62 / 0.64 s, identical counts. The
component resolution runs only for entries that ARE links, of which an ordinary project has none.

### Red-proofs

Each reverts one hunk of `operations.gd` and reruns the real-engine tests.

1. **Aliasing rule reverted** (drop the `_is_in_engine_cache` guard in `_should_descend`, and
   `_should_collect` → `return true`) → `test_an_alias_cannot_re_admit_the_engine_cache` FAILS,
   leaking all five aliased spellings.
2. **Cycle rule reverted** (drop the ancestor-chain loop) →
   `test_a_symlink_cycle_terminates_without_fabricated_paths` FAILS with `res://sub/loop/leaf.gd`,
   `res://sub/loop/loop/leaf.gd`, … — the fabricated paths the issue reported.
3. **File-side touch point reverted only** (`_should_collect` → `return true`) →
   `test_an_alias_cannot_re_admit_the_engine_cache` FAILS with **exactly**
   `{res://alias_cache.gd, res://alias_cache.tscn}` while the directory aliases stay excluded, and
   the source-level guard `test_the_engine_cache_exclusion_has_one_owner` FAILS.
4. **Component resolution reverted** (`_fully_resolved_path` → `_resolved_link_path`) → both new
   tests FAIL, in opposite directions: the leak test on
   `{res://L2/*, res://link1/c/*, res://link1/ci/cached.gd, res://link1/f_alias.gd}`, and the
   vendored-cache test on `res://vendored/x/vendored_sample.gd` being absent.
5. **Ancestor climb re-bounded** (`while` → `for _step in range(SYMLINK_PROBE_MAX_STEPS)`) →
   `test_a_cache_alias_is_excluded_however_deep_below_the_cache_it_points` FAILS with
   `res://alias_deep/deep_cache.gd`.
6. **The `class_name` assertion is falsifiable.** Under red-proof 3, `node add --type
   RootCacheThing` returns `{"script_class": "RootCacheThing", …}` — no `error` key — and the
   assertion fails. With the previous `extends RefCounted` fixture the same leak returned
   `invalid_node_type: registered class_name RootCacheThing is not a Node-derived script:
   res://alias_cache.gd`, which the old assertion **accepted**.

## Docs synced (AC5, AC6)

`docs/command-catalog.md`'s exclusion passage holds the full rule and its rationale;
`_should_descend`'s comment and ADR-0032's amendment point at it. The catalog additionally records
the AC3 disposition that previously lived only here: a refused descent is reported by **omission**,
and a result field naming the skipped link was considered and declined because it would widen all
four collectors' result contracts for a diagnostic the listing already carries.

## Residual risk

- **A link gda cannot read.** A `read_link` the process is not permitted to make, or a chain longer
  than the OS resolves, stops at the furthest path resolved — so an unresolvable alias is reported
  rather than hidden. On Linux a 33-to-40-link chain is kernel-resolvable and gda stops short of it;
  accepted, since such a chain is not a shape an honest project produces.
- **Hard links.** A hard link at a file inside the cache is enumerated, because the filesystem does
  not report one as a link. Stated as scope in the catalog and in `_is_in_engine_cache`, not left
  implied by "no alias".
- Windows `is_equivalent` and `read_link` are exercised only through the engine's own
  implementations; the e2e fixtures are POSIX symlinks and CI is macOS/Linux. The Windows contract
  difference is documented at the resolver rather than assumed away.
- **Known, not addressed in this PR** (raised by the review, no behaviour impact today): the
  `_should_descend` fast path compares `ENGINE_CACHE_DIR` outside `_is_in_engine_cache`, so the
  "one owner" claim holds for links but not for the ordinary non-link case; three co-located notes
  still describe the exclusion as one directory-side rule; the source-level guard keys on the
  constant's NAME rather than the cache literal; and `project statistics`'s fourth-collector
  evidence rests on `script_count`/`scene_count` rather than `total_files`.

### DIFF
--- a/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
+++ b/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
@@ -60,10 +60,13 @@ instantiation (and the Node-vs-Resource base-class check) remains **split per si
 > the boundary is now: a `class_name` declared in a script inside the ROOT engine cache does not
 > resolve, whatever path reaches it (the #760 repro resolved `RootCacheThing` through an alias); a
 > `class_name` under a NESTED `.godot`, or in a vendored checkout reached through a link, still
-> does. A file link at an authored script is indexed under both its paths, so a `class_name` it
-> declares is `ambiguous_class_name` — the same report this ADR already gives a name declared in two
-> files. The full rule and its rationale live with the walk, in `docs/command-catalog.md`'s
-> exclusion passage.
+> resolves — for ENUMERATION and this index. Whether that same file may be NAMED as an operation's
+> target is a different question, owned by ADR-0006's addressing gate rather than by this ADR: the
+> walk decides what the project can address by filesystem identity, the gate decides what a caller
+> may operate on from the caller's own spelling. A file link at an authored script is indexed under
+> both its paths, so a `class_name` it declares is `ambiguous_class_name` — the same report this ADR
+> already gives a name declared in two files. The full rule and its rationale live with the walk, in
+> `docs/command-catalog.md`'s exclusion passage.
 
 **Explicit contract edges:**
 
--- a/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
+++ b/docs/adr/0032-project-local-class-name-resolution-static-fallback.md
@@ -46,10 +46,24 @@ instantiation (and the Node-vs-Resource base-class check) remains **split per si
 > declaration there can make a name `ambiguous_class_name` that was unambiguous before. The trade
 > is accepted deliberately — a nested `.godot` is usually authored content (an addon vendoring a
 > sample tree), and nothing in the path distinguishes it from a vendored sub-project's own engine
-> cache, whose scripts then enter the index too. The comparison being lexical, it does not resolve
-> filesystem targets: an alias that leads to `res://.godot` is walked and the cache's own scripts
-> are indexed through it. Symlink policy for the `res://` walk is undecided and tracked in #760;
-> it is not decided here.
+> cache, whose scripts then enter the index too. The comparison being lexical, it did not resolve
+> filesystem targets: an alias that leads to `res://.godot` was walked and the cache's own scripts
+> were indexed through it. Symlink policy for the `res://` walk was undecided and tracked in #760.
+
+> **Amendment (2026-08-31, #760) — the boundary is the cache's filesystem IDENTITY, not its
+> spelling.** #760 decided the symlink policy the amendment above left open, for the whole `res://`
+> walk and therefore for this index. The walk follows a link as the engine does, but identifies what
+> it reaches with the engine's own `DirAccess.is_equivalent`, so the cache is excluded however it is
+> aliased — a directory link at it, a directory link into one of its subdirectories, or a file link
+> at a file inside it — and a link that leads back up the descent chain is not re-entered, so a
+> symlink cycle no longer floods the index with paths that name one script many times. For this ADR
+> the boundary is now: a `class_name` declared in a script inside the ROOT engine cache does not
+> resolve, whatever path reaches it (the #760 repro resolved `RootCacheThing` through an alias); a
+> `class_name` under a NESTED `.godot`, or in a vendored checkout reached through a link, still
+> does. A file link at an authored script is indexed under both its paths, so a `class_name` it
+> declares is `ambiguous_class_name` — the same report this ADR already gives a name declared in two
+> files. The full rule and its rationale live with the walk, in `docs/command-catalog.md`'s
+> exclusion passage.
 
 **Explicit contract edges:**
 
@@ -103,9 +117,9 @@ gda's existing boundary); its resolution would need a different mechanism and is
 - **Consistency restored** across the three sites: they now agree on `class_name` resolution in a
   never-opened project, rather than `find-references` reporting "no such class" while a repaired
   `resource create` resolves it.
-- **Scan boundary (amended 2026-08-28, #712).** What the scan skips is the root cache **path**, not
-  every directory named `.godot` — see the Decision amendment for the boundary, the accepted
-  nested-cache trade, and the open symlink question.
+- **Scan boundary (amended 2026-08-28 #712, 2026-08-31 #760).** What the scan skips is the root
+  cache **identity**, not every directory named `.godot` and not the path as written — see the
+  Decision amendments for the boundary, the accepted nested-cache trade, and the symlink policy.
 - **Cost.** A never-opened-project miss walks the whole `res://` tree and parses every `.gd`; this is
   acceptable for one-shot ops and is **not** backed by a persistent gda-owned cache — a gda-owned
   cache would re-introduce the very ".godot ownership" complexity the editor-scan option was rejected
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -552,19 +552,16 @@ issue #30) — null when the source declares neither, so the listing names every
 Enumeration needs a project, so projectless it is refused with `project_not_found` (pass
 `--project`); an empty project is a valid, empty listing, not an error. The walk excludes exactly
 one directory — the engine's own cache at `res://.godot`, whose contents are import artefacts no
-agent authored. The test is **lexical**: the child's `res://` PATH is compared against that one
-path. Not the directory NAME, so a nested `.godot` is walked — it is usually authored content, and
+agent authored. The first test is **lexical**: the child's `res://` PATH is compared against that
+one path. Not the directory NAME, so a nested `.godot` is walked — it is usually authored content, and
 excluding it hid real scripts from the listing and let `script validate --all` report a valid
 aggregate for a project holding an invalid script (#663 review). Sometimes it is not authored
 content — a vendored sub-project checked out under `res://` and opened once in an editor keeps an
 engine cache of its own, whose import artefacts then count in `project statistics` and become
 `find-unused-resources` candidates. That cost is accepted deliberately: gda cannot tell the two
-apart from the directory alone, and a false-valid aggregate is the worse failure. And because the
-test is lexical it compares the path as written, so it does **not** resolve filesystem targets: a
-symlink or alias under another path that leads to `res://.godot` is walked, and the cache's contents
-are then enumerated through that path. Symlink policy for the `res://` walk is undecided — #760 owns it, together with the symlink CYCLE the same walk descends until the OS path
-limit stops it. Hidden entries are otherwise enumerated as promised (#54). **This rule governs the
-four `res://` collectors in `operations.gd`** — the `script list` walk, the `scene list` walk, and
+apart from the directory alone, and a false-valid aggregate is the worse failure. Hidden entries are
+otherwise enumerated as promised (#54). **This rule governs the four `res://` collectors in
+`operations.gd`** — the `script list` walk, the `scene list` walk, and
 both static-analysis walks (the extension-filtered one behind `find-references`, `dependencies`,
 `find-unused-resources` and the `class_name` index, and the unfiltered one `project statistics`
 counts with) — so one project cannot answer two ways. It once did: three of the four compared the
@@ -575,6 +572,35 @@ copied per collector, and the copies drifted a second time — on the extension 
 shared traversal plus the acceptance test it passes. What they share is the traversal and the
 exclusion rule, **not** a file universe; the two static-analysis walks below still range over
 different files.
+
+The lexical test is the walk's first question, not its only one, because a link renames what it
+points at (#760). Comparing the path as written let an alias re-admit exactly what the rule
+excludes — `res://nested/.godot` pointing at the root cache made the cache's own scripts and scenes
+visible under a second name — and let a cycle (`sub/loop` -> `sub`) be descended until the OS
+refused another symlink hop, spelling one `.gd` 33 ways with a deepest path 174 characters long.
+The walk therefore **follows a link, as the engine does** — `DirAccess` stats a link entry so a
+linked directory lists as a directory, `ResourceLoader` loads through an alias, and gda's own
+addressing gate already counts a symlinked-in file as part of the project's `res://` namespace
+(ADR-0006) — but it **identifies what it reaches by filesystem identity**, through the engine's own
+`DirAccess.is_equivalent` (`st_dev`/`st_ino` on Unix, the volume+file id on Windows), rather than by
+the spelling that reached it. Two rules follow:
+
+- the **engine cache is excluded by identity**, so no alias re-admits it: not a directory link AT
+  `res://.godot`, not one INTO a subdirectory of it, and not a FILE link at a file inside it. That
+  last shape is a second touch point — it reaches the acceptance test without passing the descent
+  decision — so both branches of the walk ask the same owner;
+- a linked directory **already on the current descent chain is not re-entered**, so a cycle ends by
+  rule once every real directory has been walked, and the listing is a decided, bounded answer
+  rather than the leftovers of an OS limit.
+
+Both rules are about **where** a link leads, not about links: a vendored checkout that physically
+lives outside `res://` and is reached through a directory link inside it is walked and enumerated
+exactly as an ordinary directory, and its scripts' `class_name`s resolve. One consequence is kept
+deliberately — a FILE link at authored content is enumerated under **both** paths, since both are
+real `res://` addresses the engine loads; when the linked script declares a `class_name`, those two
+paths make it `ambiguous_class_name` (ADR-0032), the same report gda gives any project that declares
+one name twice.
+
 `gda script delete`
 removes a script file and reports the removed script's `class_name`/`extends` (parsed before
 deletion), so the result names the content, not just the path. Delete honors the same addressing
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -580,26 +580,49 @@ visible under a second name — and let a cycle (`sub/loop` -> `sub`) be descend
 refused another symlink hop, spelling one `.gd` 33 ways with a deepest path 174 characters long.
 The walk therefore **follows a link, as the engine does** — `DirAccess` stats a link entry so a
 linked directory lists as a directory, `ResourceLoader` loads through an alias, and gda's own
-addressing gate already counts a symlinked-in file as part of the project's `res://` namespace
-(ADR-0006) — but it **identifies what it reaches by filesystem identity**, through the engine's own
+containment gate already counts a symlinked-in file as part of the project's `res://` namespace
+(the containment rule under `script validate` below, implemented in `src/gda/project.py`) — but it
+**identifies what it reaches by filesystem identity**, through the engine's own
 `DirAccess.is_equivalent` (`st_dev`/`st_ino` on Unix, the volume+file id on Windows), rather than by
 the spelling that reached it. Two rules follow:
 
-- the **engine cache is excluded by identity**, so no alias re-admits it: not a directory link AT
-  `res://.godot`, not one INTO a subdirectory of it, and not a FILE link at a file inside it. That
-  last shape is a second touch point — it reaches the acceptance test without passing the descent
-  decision — so both branches of the walk ask the same owner;
-- a linked directory **already on the current descent chain is not re-entered**, so a cycle ends by
-  rule once every real directory has been walked, and the listing is a decided, bounded answer
-  rather than the leftovers of an OS limit.
+- the **engine cache is excluded by identity**, so no symlink alias re-admits it: not a directory
+  link AT `res://.godot`, not one INTO a subdirectory of it, not a FILE link at a file inside it,
+  and not any of those three reached under a second spelling — a parent directory that is itself a
+  link. That last shape is why the walk resolves **every component** of a path rather than only its
+  last one: a link's target is read against the directory the kernel reads it from, never against
+  the spelling the walk arrived by. The file link is a second touch point — it reaches the
+  acceptance test without passing the descent decision — so both branches of the walk ask the same
+  owner. A **hard** link is outside the rule by construction, not by oversight: the filesystem does
+  not report one as a link, so a hard link at a file inside the cache is enumerated like any other
+  file. The guarantee is about symlink aliases;
+- a linked directory **already on the current descent chain is not re-entered**, so a cycle
+  terminates by rule instead of at the OS symlink limit. What the listing enumerates is distinct
+  `res://` **paths**, not distinct directories: a directory reachable through several link paths is
+  reported under each of them, and mutually linked directories multiply the spellings quickly. The
+  answer is a decided, finite one rather than the leftovers of an OS limit — that is the guarantee,
+  and it is not "each real directory exactly once".
+
+A refused descent is reported by **omission**: the cycle's paths are simply absent, and no result
+field names the link the walk declined to follow. Adding one was considered and declined — it would
+widen all four collectors' result contracts for a diagnostic the listing already carries, and none
+of the four has a place for a per-path note.
 
 Both rules are about **where** a link leads, not about links: a vendored checkout that physically
 lives outside `res://` and is reached through a directory link inside it is walked and enumerated
-exactly as an ordinary directory, and its scripts' `class_name`s resolve. One consequence is kept
-deliberately — a FILE link at authored content is enumerated under **both** paths, since both are
-real `res://` addresses the engine loads; when the linked script declares a `class_name`, those two
-paths make it `ambiguous_class_name` (ADR-0032), the same report gda gives any project that declares
-one name twice.
+exactly as an ordinary directory, and its scripts' `class_name`s resolve — including a checkout that
+carries an engine cache of its own, which is that checkout's cache and not this project's. One
+consequence is kept deliberately — a FILE link at authored content is enumerated under **both**
+paths, since both are real `res://` addresses the engine loads; when the linked script declares a
+`class_name`, those two paths make it `ambiguous_class_name` (ADR-0032), the same report gda gives
+any project that declares one name twice.
+
+**Enumeration is not targeting.** Being reached by this walk says what the project can address; it
+does not say gda will operate on the path once it is NAMED as an operation's target. That is the
+separate question the containment rule under `script validate` below answers, and the two read
+differently on purpose — this walk decides by filesystem identity, containment reads the caller's
+own spelling. So a linked-in sub-project can be enumerated, counted and indexed here while whether
+it may be named as a target is settled there.
 
 `gda script delete`
 removes a script file and reports the removed script's `class_name`/`extends` (parsed before
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -553,38 +553,37 @@ Enumeration needs a project, so projectless it is refused with `project_not_foun
 `--project`); an empty project is a valid, empty listing, not an error. The walk excludes exactly
 one directory — the engine's own cache at `res://.godot`, whose contents are import artefacts no
 agent authored. The first test is **lexical**: the child's `res://` PATH is compared against that
-one path. Not the directory NAME, so a nested `.godot` is walked — it is usually authored content, and
-excluding it hid real scripts from the listing and let `script validate --all` report a valid
+one path. Not the directory NAME, so a nested `.godot` is walked — it is usually authored content,
+and excluding it hid real scripts from the listing and let `script validate --all` report a valid
 aggregate for a project holding an invalid script (#663 review). Sometimes it is not authored
 content — a vendored sub-project checked out under `res://` and opened once in an editor keeps an
 engine cache of its own, whose import artefacts then count in `project statistics` and become
 `find-unused-resources` candidates. That cost is accepted deliberately: gda cannot tell the two
-apart from the directory alone, and a false-valid aggregate is the worse failure. Hidden entries are
-otherwise enumerated as promised (#54). **This rule governs the four `res://` collectors in
-`operations.gd`** — the `script list` walk, the `scene list` walk, and
-both static-analysis walks (the extension-filtered one behind `find-references`, `dependencies`,
-`find-unused-resources` and the `class_name` index, and the unfiltered one `project statistics`
-counts with) — so one project cannot answer two ways. It once did: three of the four compared the
-directory NAME, so `script list` reported a script `project statistics` counted as zero (#712).
-Since #764 the four also share ONE traversal. The scaffolding around the exclusion rule had been
-copied per collector, and the copies drifted a second time — on the extension test, where the
-`scene list` walk alone compared case-sensitively — so each collector is now a single line: the
-shared traversal plus the acceptance test it passes. What they share is the traversal and the
-exclusion rule, **not** a file universe; the two static-analysis walks below still range over
-different files.
+apart from the directory alone, and a false-valid aggregate is the worse failure. Hidden entries
+are otherwise enumerated as promised (#54). **This rule governs the four `res://` collectors in
+`operations.gd`** — the `script list` walk, the `scene list` walk, and both static-analysis walks
+(the extension-filtered one behind `find-references`, `dependencies`, `find-unused-resources` and
+the `class_name` index, and the unfiltered one `project statistics` counts with) — so one project
+cannot answer two ways. It once did: three of the four compared the directory NAME, so `script
+list` reported a script `project statistics` counted as zero (#712). Since #764 the four also
+share ONE traversal. The scaffolding around the exclusion rule had been copied per collector, and
+the copies drifted a second time — on the extension test, where the `scene list` walk alone
+compared case-sensitively — so each collector is now a single line: the shared traversal plus the
+acceptance test it passes. What they share is the traversal and the exclusion rule, **not** a file
+universe; the two static-analysis walks below still range over different files.
 
 The lexical test is the walk's first question, not its only one, because a link renames what it
 points at (#760). Comparing the path as written let an alias re-admit exactly what the rule
-excludes — `res://nested/.godot` pointing at the root cache made the cache's own scripts and scenes
-visible under a second name — and let a cycle (`sub/loop` -> `sub`) be descended until the OS
-refused another symlink hop, spelling one `.gd` 33 ways with a deepest path 174 characters long.
-The walk therefore **follows a link, as the engine does** — `DirAccess` stats a link entry so a
-linked directory lists as a directory, `ResourceLoader` loads through an alias, and gda's own
+excludes — `res://nested/.godot` pointing at the root cache made the cache's own scripts and
+scenes visible under a second name — and let a cycle (`sub/loop` -> `sub`) be descended until the
+OS refused another symlink hop, spelling one `.gd` 33 ways with a deepest path 174 characters
+long. The walk therefore **follows a link, as the engine does** — `DirAccess` stats a link entry
+so a linked directory lists as a directory, `ResourceLoader` loads through an alias, and gda's own
 containment gate already counts a symlinked-in file as part of the project's `res://` namespace
 (the containment rule under `script validate` below, implemented in `src/gda/project.py`) — but it
 **identifies what it reaches by filesystem identity**, through the engine's own
-`DirAccess.is_equivalent` (`st_dev`/`st_ino` on Unix, the volume+file id on Windows), rather than by
-the spelling that reached it. Two rules follow:
+`DirAccess.is_equivalent` (`st_dev`/`st_ino` on Unix, the volume+file id on Windows), rather than
+by the spelling that reached it. Two rules follow:
 
 - the **engine cache is excluded by identity**, so no symlink alias re-admits it: not a directory
   link AT `res://.godot`, not one INTO a subdirectory of it, not a FILE link at a file inside it,
@@ -600,21 +599,23 @@ the spelling that reached it. Two rules follow:
   terminates by rule instead of at the OS symlink limit. What the listing enumerates is distinct
   `res://` **paths**, not distinct directories: a directory reachable through several link paths is
   reported under each of them, and mutually linked directories multiply the spellings quickly. The
-  answer is a decided, finite one rather than the leftovers of an OS limit — that is the guarantee,
-  and it is not "each real directory exactly once".
+  answer is a decided, finite one rather than the leftovers of an OS limit — that is the
+  guarantee, and it is not "each real directory exactly once".
 
 A refused descent is reported by **omission**: the cycle's paths are simply absent, and no result
-field names the link the walk declined to follow. Adding one was considered and declined — it would
-widen all four collectors' result contracts for a diagnostic the listing already carries, and none
-of the four has a place for a per-path note.
+field names the link the walk declined to follow. Adding one was considered and declined — it
+would widen all four collectors' result contracts for a diagnostic the listing already carries,
+and none of the four has a place for a per-path note.
 
 Both rules are about **where** a link leads, not about links: a vendored checkout that physically
 lives outside `res://` and is reached through a directory link inside it is walked and enumerated
-exactly as an ordinary directory, and its scripts' `class_name`s resolve — including a checkout that
-carries an engine cache of its own, which is that checkout's cache and not this project's. One
-consequence is kept deliberately — a FILE link at authored content is enumerated under **both**
-paths, since both are real `res://` addresses the engine loads; when the linked script declares a
-`class_name`, those two paths make it `ambiguous_class_name` (ADR-0032), the same report gda gives
+exactly as an ordinary directory, and its scripts' `class_name`s resolve — including a checkout
+that carries an engine cache of its own, which is that checkout's cache and not this project's.
+One consequence is kept deliberately — content reachable under two `res://` paths is enumerated
+under **both**, since both are real addresses the engine loads. That holds for a FILE link at an
+authored script and equally for a DIRECTORY link at authored content already reachable in-tree,
+which enumerates everything below it a second time; when a script so reached declares a
+`class_name`, the two paths make it `ambiguous_class_name` (ADR-0032), the same report gda gives
 any project that declares one name twice.
 
 **Enumeration is not targeting.** Being reached by this walk says what the project can address; it
--- a/src/gda/ops/operations.gd
+++ b/src/gda/ops/operations.gd
@@ -165,11 +165,22 @@ const SCENE_STARTUP_NOT_READY := "not_ready"
 const ENGINE_CACHE_DIR := "res://.godot"
 
 # How far a res:// walk follows symlinks when it asks whether an entry is the
-# engine cache (#760): the links resolved in one chain, and the ancestors probed
-# above a resolved target. One bound serves both, and 32 is the ceiling the OS
-# itself puts on symlink resolution (MAXSYMLINKS on macOS and Linux) — a chain gda
-# gives up on is one the kernel would refuse to open anyway, so the bound cannot
-# hide a path the walk could otherwise have reached.
+# engine cache (#760): the links followed in one chain, and the passes the
+# component-by-component resolution takes to reach its fixed point. Both count
+# SYMLINK TRAVERSALS — a pass that changes the path resolved at least one more
+# link — which is the quantity the OS itself bounds, so a chain gda gives up on
+# is one the kernel would refuse to open anyway and the bound cannot hide a path
+# the walk could otherwise have reached. The value is the LOWER of the two
+# ceilings gda targets: MAXSYMLINKS is 32 on macOS (MacOSX.sdk/usr/include/
+# sys/param.h:197) and 40 on Linux (include/linux/namei.h), so a 33-to-40-link
+# chain is resolvable on Linux and gda stops short of it — accepted, because
+# such a chain is not a shape an honest project produces.
+#
+# The ancestor climb in _is_in_engine_cache is deliberately NOT bounded by this
+# constant: it counts path COMPONENTS, a quantity no kernel limits, and running
+# out of steps there returned false, which ADMITTED cache content — the opposite
+# direction from the failure MAXSYMLINKS reasons about, and a leak at 32 levels
+# below the cache (#795 review). It terminates on its own at the root instead.
 const SYMLINK_PROBE_MAX_STEPS := 32
 
 # The project-info settings (issue #111), read with a default so a project that
@@ -4283,12 +4294,17 @@ func _ambiguous_class_name_message(class_token: String, paths: Array) -> String:
 # identifies what it reached by FILESYSTEM IDENTITY, not by the spelling that
 # reached it. Two rules follow, and only a link pays for them:
 #
-#   - the engine cache is excluded by identity, so no alias re-admits it
-#     (_is_in_engine_cache);
+#   - the engine cache is excluded by identity, so no SYMLINK alias re-admits it
+#     (_is_in_engine_cache — which also states why a hard link is outside the
+#     rule);
 #   - a linked directory already on this descent CHAIN is not re-entered, so a
-#     cycle ends by rule after each real directory has been walked once, instead
-#     of running to the OS symlink limit — `sub/loop -> sub` once emitted 33
-#     spellings of one file, the deepest 174 characters long.
+#     cycle terminates by rule instead of running to the OS symlink limit —
+#     `sub/loop -> sub` once emitted 33 spellings of one file, the deepest 174
+#     characters long. The chain is per-BRANCH, so what the walk enumerates is
+#     distinct res:// PATHS: a directory reachable through several link paths is
+#     reported under each, and mutually linked directories multiply the spellings
+#     quickly. The answer is decided and finite — that is the guarantee — not
+#     "each real directory exactly once" (#795 review).
 #
 # A link to ordinary authored content — a vendored checkout reached through one —
 # is neither the cache nor an ancestor, so it is followed and enumerated exactly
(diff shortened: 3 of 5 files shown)


qinhaihong-red:
## Independent adversarial review — head `2bd31608df` (against base `181da20d`)

Reviewed in a detached archive at that exact SHA. All engine evidence measured on
`4.6.3.stable.official.7d41c59c4`, each run with its own `GDA_USER_DATA_ROOT`. Engine
citations checked against the upstream `4.6.3-stable` and `4.6-stable` tags (no local
`../godot` checkout was available in this environment — disclosed as a substituted source).

| Axis | Verdict |
| --- | --- |
| Artifact / Markdown | **Changes required** — the published guarantee is falsified; 3 wrong engine citations |
| Architecture / DDMA | **Changes required** — 3 x P2 (ownership claim, Windows `read_link` contract, ADR-0009 miscitation) |
| Standards | **Changes required** — 4 x P3; all gates green, three P1-class hazards checked and clean |
| Spec (#760) | **Changes required** — AC2 not met; AC4 partially met (P1 test gap) |

The cycle half of the policy holds up under everything I threw at it. The **aliasing** half
does not: it decides by the spelling after all, in both directions.

---

## P1 — the identity test is anchored on the link's SPELLING, so the cache leaks under one spelling and authored content is hidden under another

`_resolved_link_path` (`src/gda/ops/operations.gd:4377-4386`) joins a relative `read_link`
target onto `current.get_base_dir()` — the link's `res://` **spelling**. Its own comment says
this is "the directory holding the link, which is how the kernel reads it"
(`:4372-4373`). It is not, whenever the link's parent is **itself a directory symlink**: the
kernel reads the target from where the link really lives, gda reads it from where the walk
spelled it. Everything downstream (`_is_in_engine_cache`'s `is_equivalent` probe and its
`get_base_dir()` ancestor walk, `:4359-4368`) is then anchored on a path the kernel never
visits.

### Minimal repro — one real directory, two `res://` spellings, two different answers

```
project.godot
.godot/root_cache.gd        (class_name RootCacheThing)
.godot/root_cache.tscn
sub/deep/c   -> ../../.godot      # an ordinary RELATIVE link
link1        -> sub/deep          # a second spelling of the same real directory
```

`gda script list` at head:

```
res://link1/c/root_cache.gd   class_name RootCacheThing     <-- LEAKED
res://link1/deep_leaf.gd
res://real.gd
res://sub/deep/deep_leaf.gd                                 <-- res://sub/deep/c/... correctly excluded
```

`res://link1/c` and `res://sub/deep/c` are the **same directory entry**. The walk excludes one
and admits the other. That is precisely the property the PR names as its design principle —
"it identifies what it reaches by FILESYSTEM IDENTITY rather than by the spelling that reached
it" — falsified by construction.

Ground truth from the engine, in the same project:

```
is_link("res://link1/c")                        = true
read_link("res://link1/c")                      = ../../.godot
gda's lexical join                              = res://link1/../../.godot
is_equivalent(that join, "res://.godot")        = false      <-- the miss
is_equivalent("res://link1/c", "res://.godot")  = TRUE       <-- the truth
```

### All three aliasing shapes leak, not just one

Each is `link1 -> sub/deep` plus one ordinary relative link inside `sub/deep`:

| link inside `sub/deep` | shape | leaked at head |
| --- | --- | --- |
| `c -> ../../.godot` | directory link **AT** the cache | `res://link1/c/root_cache.gd`, `res://link1/c/root_cache.tscn` |
| `ci -> ../../.godot/imported` | directory link **INTO** it | `res://link1/ci/cached.gd`, `res://link1/ci/cached.tscn` |
| `f_alias.gd -> ../../.godot/root_cache.gd` | **FILE** link inside it | `res://link1/f_alias.gd` |

So both touch points and both sub-rules of rule 1 are affected — `_should_collect` is not a
narrower miss than `_should_descend`, they share the same broken owner.

### Blast radius: all four collectors, `statistics`, `--all`, and the `class_name` index

For the `c -> ../../.godot` case the leaked paths appear in `script list`, `scene list`,
`project dependencies`, **`project find-unused-resources`** (it nominates
`res://link1/c/root_cache.tscn` for deletion — an engine-cache artefact offered to an agent as
removable), and `project statistics`. `script validate --all` reports
`res://link1/c/root_cache.gd valid: true` in the aggregate. And the index:

```
$ gda node add res://host.tscn --parent . --name X --type RootCacheThing
invalid_node_type: registered class_name RootCacheThing is not a Node-derived script:
                   res://link1/c/root_cache.gd
```

versus the correctly-excluded control (the PR's own fixture shape):

```
invalid_node_type: not an instantiable Node class, and no .gd script declares
                   class_name RootCacheThing
```

This is the third line of #760's own repro (`node add --type RootCacheThing -> resolves against
the cache script`), still reproducing. **AC2 is not met.**

### The same bug also HIDES authored content — a regression against `181da20d`

The anchoring error cuts both ways. A vendored checkout that lives outside `res://`, carrying
its own nested `.godot` (the case #712 decided to walk and the catalog calls "usually authored
content"), reached through a link:

```
<outside>/.godot/vendored_sample.gd     # authored content in the CHECKOUT's own cache dir
<outside>/v/x  -> ../.godot
res://vendor   -> <outside>/v
```

`res://vendor/x` really is `<outside>/.godot`, **not** the project's cache. gda computes
`res://vendor/../.godot` → simplified `res://.godot` → `is_equivalent` says true (it stats the
root cache) → excluded.

```
base 181da20d : ['res://real.gd', 'res://vendor/x/vendored_sample.gd']
head 2bd31608 : ['res://real.gd']                                        <-- silently hidden
```

That is a **new** false negative introduced here, and it is exactly what AC1 forbids: "a
vendored checkout reached through a link is a real workflow; do not break it silently."

### The scope: only the LAST component is resolved

`_resolved_link_path` resolves a chain of links at the *final* component. Every component before
it stays as the walk spelled it, and `_is_in_engine_cache`'s ancestor walk then uses
`get_base_dir()` on that half-real path. So the depth of the miss grows with the number of linked
components:

```
a/b/d/c -> ../../../.godot      # an ordinary relative link
L1      -> a/b                  # linked component
L2      -> L1/d                 # a link THROUGH a link
head: res://L1/d/c/root_cache.gd and res://L2/c/root_cache.gd both leak
```

### Remedy — resolve the whole path, not the last component

I probed two fixes in a scratch copy of the head tree.

1. **Anchor a relative target on the resolved base dir** (`:4385`, `current.get_base_dir()` →
   `_resolved_link_path(dir, current.get_base_dir())`). Closes `split_*`, `through_vendor` and
   the false exclusion, but **still leaks** `res://L2/c/root_cache.gd` above — it resolves one
   more component, not all of them.
2. **Resolve every component to a fixed point** — walk the path segment by segment, resolve each
   through `_resolved_link_path`, and repeat until a pass changes nothing (bounded by
   `SYMLINK_PROBE_MAX_STEPS`), then feed that to `_is_in_engine_cache`. This one is clean on
   **every** shape I built: the three leaks, the two-linked-component case, the false exclusion,
   and all eleven shapes that were already correct — with
   `tests/test_project_walk.py tests/test_e2e_project_walk.py` still **9 passed**.

Whatever form you pick, the invariant to hold is the one the comment at `:4372-4373` already
states and the code does not: the target is read from the directory **the kernel** reaches, not
from the one the walk spelled. Note that #799 solves the same hazard on the containment side and
its rule can be reused as the precedent (see the wave-input section).

Please also add a regression fixture with a **relative** link target and a link **inside a
linked directory** — the current fixture uses only absolute `symlink_to(...)` targets and no
nested link, which is why the suite is green over all of this.

### Honesty note on severity

Base `181da20d` leaks the same shape and more (it admits *both* spellings), so the leak is an
**incomplete fix, not a regression**. The false exclusion above *is* a regression. I file the
whole thing P1 because the PR publishes an unconditional guarantee — into
`docs/command-catalog.md` and ADR-0032, which outlive the PR — that this evidence falsifies.

---

## P2 — a SECOND, independent leak: `SYMLINK_PROBE_MAX_STEPS` is reused as a path-DEPTH bound, and its stated invariant is false

`:167-173`, repeated in the PR body:

> "One bound serves both, and 32 is the ceiling the OS itself puts on symlink resolution
> (MAXSYMLINKS on macOS **and Linux**) — a chain gda gives up on is one the kernel would refuse
> to open anyway, so the bound cannot hide a path the walk could otherwise have reached."

Two things are wrong with that sentence, and one of them is exploitable.

**(a) The bound has two uses; MAXSYMLINKS licenses only one.** It bounds `_resolved_link_path`'s
link-**hop** loop (`:4379`) — fair — and `_is_in_engine_cache`'s **ancestor-climb** loop
(`:4361-4367`), which counts path *components*, a quantity no kernel limits. Exhausting the
second returns `false`, i.e. it **admits** cache content — the opposite direction from the
failure the sentence reasons about. Reproduced at head with an absolute alias at a deep cache
subdirectory (`res://alias_deep -> <abs>/.godot/d1/…/dN`):

```
depth below the cache = 31  ->  {"scripts": []}
depth below the cache = 32  ->  {"scripts": [{"path": "res://alias_deep/deep_cache.gd", ...}]}
```

and the alias re-admits that script's `class_name` into the index too. Contrived for an honest
project — but it is a real second leak, independent of the P1, and the remedy is a deletion:
**drop `range(SYMLINK_PROBE_MAX_STEPS)` from `_is_in_engine_cache`** — the loop already
terminates on `parent.is_empty() or parent == probe` (`:4365-4366`), naturally bounded by path
depth. The constant then has one use and the MAXSYMLINKS rationale becomes true as written.

**(b) `MAXSYMLINKS` is 40 on Linux, not 32.** Verified in source:
`MacOSX.sdk/usr/include/sys/param.h:197` → `32`; `torvalds/linux include/linux/namei.h:14` →
`40`. So on Linux a 33-to-40-link chain *is* kernel-resolvable while `_resolved_link_path` gives
up and returns a partially resolved path — the bound *can* hide a path the walk could otherwise
have reached. I could not demonstrate this on macOS, where the kernel refuses first. Either say
"the lower of the two ceilings" and keep 32, or name the Linux gap.

---

## P2 — the e2e `class_name`-index assertion cannot fail

`tests/test_e2e_project_walk.py::test_an_alias_cannot_re_admit_the_engine_cache` asserts:

```python
assert error["code"] == "invalid_node_type", error
assert cache_class in error["message"]
```

Both planted cache scripts are `extends RefCounted`, so a leaked resolution *also* returns
`invalid_node_type` with the class name in the message (see the two messages quoted in the P1
section — both satisfy this assertion). The guard for the fifth consumer of the walk is
therefore blind to the very regression it exists to catch.

Remedy: assert on the discriminating half — `"no .gd script declares class_name" in
error["message"]`, or make one planted cache script `extends Node` so a leak returns a
**success** the test can fail on.

---

## P2 — three of the seven engine citations are wrong, and one names a different function

Checked against upstream `4.6.3-stable` and `4.6-stable` (identical for these files; `master`
matches none of the cited numbers either).

| Cited | Where | Actual | |
| --- | --- | --- | --- |
| `dir_access_unix.cpp:148-183` `get_next` | comment + body | 148-183 | correct |
| `dir_access_unix.cpp:55-57` `fix_path` | comment + body | 55-57 | correct |
| `dir_access_unix.cpp:481-499` `read_link` | comment + body | 481-499 | correct |
| `dir_access_unix.cpp:713-729` `is_equivalent` | comment + body | 713-729 | correct |
| `core/io/dir_access.cpp:636-638` string fallback | comment `:4336` + body | **630-632** | 636-638 is `_bind_methods`'s first static binds |
| `core/io/dir_access.cpp:688` script binding | body | **681** | 688 is `_delete_temp()` in `~DirAccess` |
| `dir_access_windows.cpp:442-462` `is_equivalent` | comment `:4338` + body | **411-431** | 442-462 is `DirAccessWindows::read_link` |

The four exact `dir_access_unix.cpp` ranges make version skew an unlikely explanation. Two of
the wrong ones ship permanently in the `_is_in_engine_cache` comment, on the one function the
whole policy rests on; the Windows pointer sends a reader to `read_link`.

(Unrelated to this PR, but worth a one-line edit when #760 is next touched: the issue body's
"`DirAccess.is_link()` and `read_link()` ... (`core/io/dir_access.cpp:674-675`)" — actual
667-668.)

---

## P2 — the cycle rule's published outcome holds only for the self-cycle the test covers

Catalog: "a cycle ends by rule **once every real directory has been walked**".
`_should_descend` comment `:4288-4289`: "a cycle ends by rule after each real directory has been walked **once**".

The rule is per-branch (a fresh `chain` per descent), so it enumerates **simple paths**, not
directories. A 2-cycle already walks each real directory twice:

```
ca/loop -> ../cb ,  cb/loop -> ../ca
script list -> res://ca/a_leaf.gd, res://ca/loop/b_leaf.gd,
               res://cb/b_leaf.gd, res://cb/loop/a_leaf.gd, res://real.gd
```

and k mutually-linked directories give a factorial number of spellings. Measured at head:

| k | entries | distinct scripts | wall clock |
| --- | --- | --- | --- |
| 4 | 65 | 4 | 1.2 s |
| 5 | 326 | 5 | 1.6 s |
| 6 | 1957 | 6 | 5.3 s |
| 7 | **13 700** | 7 | **35.7 s** |

To be fair to the change: base `181da20d` on the same k=4 fixture returns **no result after
60 s**, so this is a large improvement, and the answer *is* bounded. But "bounded" here is
factorial in the number of mutually-linked directories, not "once per real directory". Remedy
is wording, not code — say the cycle terminates by rule and the listing enumerates distinct
*paths* (a directory reachable by several link paths is reported under each), and drop "once".

---

## P2 — the residual-risk disclosure misdescribes the gap it discloses

`_is_in_engine_cache` comment `:4355-4358`:

> "A deliberately nested chain of links could still spell a path whose ancestors resolve
> elsewhere; that is a `Trusted project` (ADR-0009) question, not one this predicate defends
> against — the rule exists so an honest project's aliases report honestly."

The P1 repro is not a deliberately nested chain: it is **one ordinary relative symlink inside
one ordinary directory symlink** — relative link targets are the norm in a checked-in repo, and
the directory link is the very vendored-checkout workflow this PR blesses. ADR-0009 is about a
*malicious* project; this fires on an honest one, and the reported result is the same defect
#760 exists to close. Whatever survives the P1 fix should be re-described by shape, not
assigned to ADR-0009.

---

## P3 — "no alias re-admits it" is also false for a hard link

Both predicates gate the identity test behind `dir.is_link(child)`, so a hard link — a
filesystem-identity alias by definition — is never asked:

```
ln .godot/root_cache.gd hard_cache.gd ; ln .godot/root_cache.tscn hard_cache.tscn
script list  -> res://hard_cache.gd  class_name RootCacheThing
scene  list  -> res://hard_cache.tscn
find-unused  -> res://hard_cache.tscn
```

The PR body scopes rule 1 to the three symlink shapes, so this is a claim/scope mismatch rather
than a missed decision. The `_should_descend` comment's unqualified "the engine cache is
excluded by identity, so no alias re-admits it" (`:4286-4287`) should say *symlink* alias, or
the decision to leave hard links out should be one recorded clause.

---

## P3 — the catalog promises `class_name` resolution the walk does not always give

Catalog: "a vendored checkout ... reached through a directory link inside it is walked and
enumerated exactly as an ordinary directory, and its scripts' `class_name`s **resolve**." The
adjacent sentence warns about `ambiguous_class_name` only for a **FILE** link. A **directory**
link whose target is also reachable in-tree produces the same outcome:

```
d1 -> shared , d2 -> shared
node add --type SharedLeaf
-> ambiguous_class_name: ... res://d1/shared_leaf.gd, res://d2/shared_leaf.gd,
                             res://shared/shared_leaf.gd
```

The plain duplication case is the same and equally undocumented: with `res://mirror -> res://sub`
holding one `leaf.gd`, `script list` returns **both** `res://mirror/leaf.gd` and
`res://sub/leaf.gd`, and `project statistics` reports `script_count: 2, total_files: 3` for one
`.gd`.

Literally the catalog sentence is about a checkout *outside* `res://`, so it is incomplete rather
than false — but a reader told about the file-link ambiguity and not the directory-link one will
be surprised, and the adjacent "a decided, bounded answer" invites the conclusion that one file is
spelled once. Extend the same sentence: *…and a directory link at authored content already
reachable by another path enumerates its contents under both.*

---

## Architecture / DDMA

**A1 [P2] The "one owner" claim does not hold for the ordinary case, and the new guard test
institutionalizes the exception.** `:163-164` says the decision "lives in exactly one place,
`_is_in_engine_cache`". For a **non-link** child it is never reached: `:4298-4305` answers with
`child == ENGINE_CACHE_DIR` and then `if not dir.is_link(child): return true`. The real
`res://.godot` is not a link, so line 4298 is the *only* thing excluding the cache in a normal
project — a load-bearing rule, not a fast path, and not the owner's. The two predicates
therefore compose the question differently (`_should_descend`: `child == CACHE or (is_link and
owner)`; `_should_collect` `:4327`: `is_link and owner`), which the comment at `:4319` calls
"the same question of the same owner". And `tests/test_project_walk.py:202-204` asserts
`users == {DESCEND_PREDICATE, CACHE_OWNER}` — written to *permit* the second site, so a future
third comparison inside `_should_descend` is invisible to it.
Remedy at the same syscall count: give the owner the two cheap answers
(`_is_in_engine_cache(dir, path, is_linked)` returning true on `path == ENGINE_CACHE_DIR` and
false when `not is_linked`), have both predicates call only it, and tighten the guard to
`users == {CACHE_OWNER}` so it enforces `:163-164` instead of excepting it.

**A2 [P2] `_resolved_link_path` is specified against Unix `read_link` only.** `:4349-4350` cites
only `dir_access_unix.cpp:481-499`, while the sibling comment at `:4336-4339` carefully cites
Unix *and* Windows for `is_equivalent`, and the catalog advertises the identity rule as
cross-platform. `DirAccessWindows::read_link` (`dir_access_windows.cpp:444-462`) returns
`GetFinalPathNameByHandleW` — an already fully resolved **absolute** path — and on failure
returns **`f`, never an empty string**. So on Windows the relative-target join at `:4385` is
dead, the documented "returns `path` unchanged when the target cannot be read" (`:4374`, code
`:4383-4384`) is unreachable, and an unopenable reparse point spins all 32 iterations, each a
`GetFileAttributesW` + `CreateFileW`. `DirAccessWindows::is_link` also tests
`FILE_ATTRIBUTE_REPARSE_POINT` — broader than `S_ISLNK` (junctions, cloud placeholders) — so
"only a link pays" degrades there too. The identity *outcome* still holds; the contract is what
is wrong. Remedy: cite both platforms as `_is_in_engine_cache` already does, say the function
returns the last path it resolved (not `path`) on bound exhaustion — which is also imprecise on
Unix today — and add `if target == current: return current` before `:4385` so the Windows
failure costs one probe instead of 32.

**A3 [P2] ADR-0009 is cited for a residual it does not record.** `:4354-4358` assigns the
nested-link-chain residual to "a `Trusted project` (ADR-0009) question". ADR-0009's decision,
its mechanism classification and every consequence are about **executing project code**;
CONTEXT.md's `Trusted project` entry likewise scopes it to "its autoloads and scene scripts".
Neither says gda's *results* may be wrong under an adversarially spelled directory tree. A
maintainer following the citation finds no such record. Remedy: either extend the CONTEXT.md
entry (and ADR-0009 with a dated `> Outcome` note, the repo's established pattern) to cover
filesystem layout, or drop the citation and state the residual as an accepted bound of the
probe. (See also the P2 above: the residual as *described* is not the one that actually fires.)

**A4 [P3] ADR-0006 is cited as the precedent for a symlink rule it does not contain**, in three
places: `docs/command-catalog.md:581-584` ("gda's own addressing gate already counts a
symlinked-in file as part of the project's `res://` namespace (ADR-0006)") and
`tests/test_e2e_project_walk.py:181` and `:335`. ADR-0006 has zero symlink statements; the rule
lives in `src/gda/project.py:157-183` and the catalog's containment paragraph (~:723-726). It is
also a *different* rule — the gate reads lexically and guards `..`, the walk decides by stat
identity — so the citation invites the reader to think one policy governs both. This is the
same seam the wave-input section is about.

**A5 [P3] Co-located notes still describe the exclusion as one directory-side rule** —
`:3936` ("one directory-exclusion rule (`_should_descend`)"), `:4994-4995`, `:5040-5041`, and
`_collect_paths`'s own header at `:4390`. `test_the_static_analysis_note_names_only_helpers_that_exist`
uses a subset assertion (`{...} <= mentioned`), so it cannot catch the omission. Add
`_should_collect` to the three notes and to the required set at `tests/test_project_walk.py:128-132`.

**A6 [P3] One constant now carries two limits, under a justification that covers one.**
`:167-173` justifies `32` as MAXSYMLINKS — true for `_resolved_link_path`'s hop loop (`:4379`),
false for `_is_in_engine_cache`'s **ancestor-depth** loop (`:4361-4367`), where exhausting the
bound returns `false` and therefore *admits* cache content — the opposite direction from the
failure the comment reasons about. No behavioural impact today. Split the constant, or state the
second limit's own justification and its admit-on-exhaustion direction.

**A7 [P3] The ADR-0032 amendment duplicates the mechanism, not only its ADR-local consequence.**
`docs/adr/0032…:53-65` restates `is_equivalent`, the three alias shapes and the chain rule, then
defers to the catalog for "the full rule and its rationale" — a second independently-editable
copy with no test binding them. Cut it to the boundary statement that already follows
(`:59-64`) plus the pointer.

**A8 [P3] `chain`'s default silently assumes `dir_path` is the walk root**, which nothing
enforces (`:4416-4417` states it as fact, not invariant), while `_collect_paths` accepts any
`dir_path`. And `:4284`'s "only a link pays for them" is contradicted by `:4423-4424`, where
**every** directory pays an O(depth) `chain.duplicate()` regardless of links. State the
precondition, and fix the cost sentence. *(No finding on the defaulted parameter itself: all
eight call sites pass `"res://"` literally, `dir_path` is already exposed recursion state, and
`chain` is duplicated before append so GDScript default-argument sharing is a non-issue.)*

**A9 [P3] Three load-bearing notions get no glossary home** — "descent chain", "filesystem
identity", "engine cache identity". CONTEXT.md has no term for the `res://` walk at all, while
the surrounding prose already uses *walk* / *traversal* / *scan* for related-but-distinct
things. One entry (e.g. `Res:// walk`) naming the traversal, its four collectors plus the
`class_name` index, the identity-decided exclusion and the descent-chain rule would also be the
natural home for its relationship to the addressing gate (A4). No existing CONTEXT.md term
becomes inaccurate.

**Not a finding, recorded:** #760 framed three options ("resolve, refuse links outright, or
bound depth"). The catalog rejects the first two implicitly but never states the trade against a
depth bound. Two sentences in the catalog passage, not a new ADR.

---

## Standards

Gates re-run at head (uv, task-scoped dirs): `ruff check` clean · `ruff format --check` 481
formatted · `pyright` 0/0 · `pytest -m "not e2e"` **2252 passed, 595 deselected** ·
`pytest -m e2e tests/test_e2e_project_walk.py` **5 passed**. The PR body's verification table is
accurate. Conventional Commits, `Closes #760`, `@pytest.mark.e2e` placement — all correct.

**S1 [P3] `_resolved_link_path`'s comment contradicts its code in two of three cases.**
`:4374-4376` says it "Returns `path` unchanged when it is not a link, when the target cannot be
read, and when the chain outruns the bound." Only the first is true: `:4384` and `:4386` both
return `current`, the furthest resolved path. The code's behaviour is the better one; the comment
is what is wrong. Say "returns the furthest path it could resolve".

**S2 [P3] `test_the_engine_cache_exclusion_has_one_owner` guards the constant's NAME, not the
cache path.** `tests/test_project_walk.py:194` scans for `CACHE_CONSTANT not in stripped` — the
identifier only. A new collector written `if child == "res://.godot":` passes green, which is
exactly the literal-vs-constant drift mode #712 had. `res://.godot` appears only at `:165` today,
so the tightening is free: add `assert source.count('"res://.godot"') == 1`.

**S3 [P3] `_enclosing_function` has two latent breakages.** Probed each shape: `static func` →
returns `None` and asserts with the misleading "used outside any function"; a trailing comment
naming the constant on a code line → attributed to that function, spurious failure. Multi-line
signatures (which this file has, `:1041, :1069, :1139`), nested lambdas and column-0 comments
between funcs are handled correctly. Nothing fails today. Remedy: match `^(static )?func ` and
split the line at `#` before the `:194` test.

**S4 [P3] Rewrap artifacts in the edited catalog paragraph** — `docs/command-catalog.md:556` is
104 chars and `:564` is 72 in a paragraph that otherwise wraps at 93-100. Cosmetic; no markdown
gate exists.

**Checked and clean** (worth recording, since three of these could have been P1s):

- **GDScript default-argument semantics.** Probed on 4.6.3: defaults are re-evaluated per call —
  no Python-style shared mutable default. `_collect_paths` also `duplicate()`s before appending
  (`:4423-4424`), so it would be safe either way.
- **Mid-iteration `DirAccess` use.** The predicates call `is_link` / `read_link` /
  `is_equivalent` on the *same* handle `get_next()` is walking. Probed with a 15-entry
  directory: all 15 returned, `current_is_dir()` correct after each call. No iteration-state
  corruption. (Confirmed against the source too: none of the three touch `dir_stream`.)
- **Cost.** `project statistics` over `examples/platformer/panda-adventure` (279 files, 65
  scripts, 12 scenes), 3 runs each: base 1.03 s, head 1.09 s, identical counts. The per-entry
  `lstat` is not measurable against engine startup.

---

## Spec — AC by AC

| AC | Verdict |
| --- | --- |
| AC1 recorded decision, both shapes + the legitimate symlink | **met**, with a gap (below) |
| AC2 cache invisible through an alias, four collectors + `class_name` index | **not met** — P1 and the P2 above |
| AC3 cycle terminates with a structured outcome | reading defensible; disposition half-durable |
| AC4 cross-consumer regressions for both shapes | **partially met** — see below |
| AC5 the three "tracked separately" pointers | **met** |
| AC6 ADR-0032 + catalog synced in the same change | **met** — a full-repo sweep found no un-synced site |

**AC4 [P1] Every link the fixture builds uses an ABSOLUTE target** —
`tests/test_e2e_project_walk.py:223, 224-226, 227, 228, 232, 238` are all
`symlink_to(<absolute Path>)`. `_resolved_link_path`'s relative branch (`:4385`) is therefore
never executed by any test in the suite: half of that function is untested, and it is exactly
where the P1 leak lives. Also absent: a link **inside** a linked directory, a link→link chain, a
mutual A↔B cycle across two directories, a dangling target, a sibling directory alias. Remedy:
make `nested/.godot`, `nested/imported_alias` and `sub/loop` **relative** targets, and add one
link nested under `vendored/`.

**AC2 [P2] The "all four collectors" evidence does not cover the fourth.** `_walked_paths`
(`tests/test_e2e_project_walk.py:243-255`) unions `script list`, `scene list`, `dependencies` and
`find-unused-resources` — three collectors. The unfiltered `_collect_all_file_paths` (`:4132`) is
probed only through `stats["script_count"] == 2` / `scene_count == 1` (`:282-283`), and the
fixture plants only `.gd`/`.tscn` inside the cache. A regression re-admitting the *typical* cache
content — `.ctex`, `.md5`, `uid_cache.bin`, `global_script_class_cache.cfg`, i.e. nearly all of a
real `.godot` — changes `total_files` / `by_extension` and nothing the test asserts. Remedy: add
`assert stats["total_files"] == 4` and plant one non-`.gd` file in the cache.

**AC1 [P3] The rejected options are missing from the record.** #760 named three candidate
policies — "whether to resolve, to refuse links outright, or to bound depth". The record argues
against "refuse links" (`:4294-4296`) and never mentions "bound depth". Two sentences in the
catalog passage close it. (Homing the decision in a command spec follows the #712 precedent, so
that part is not a rule breach — but a reader searching `docs/adr/` for the walk's symlink policy
finds only the `class_name`-index slice.)

**Scope creep: none.** `_should_collect` is the second touch point the landing-site correction
explicitly required; nothing else in the diff is unasked-for.

---

## AC3 adjudication — the reading is defensible; the disposition is only half-findable

"the listing IS the bounded answer" is a fair reading: AC3's operative demand is termination
"without emitting fabricated paths ... rather than a truncation artefact of the OS path limit",
and a rule-based stop satisfies that. The **positive** half is published (catalog: "the listing
is a decided, bounded answer rather than the leftovers of an OS limit") — findable. The
**negative** half is not: the declined alternative — a result field naming the skipped link —
and the reason for declining it ("expanding them was judged out of proportion") live only in the
PR body, which is not a durable artifact (RULES.md, 权威信息源 / 状态同步). "With a structured
outcome" does read as a clause beyond "without emitting fabricated paths", so this *is* a real
narrowing, and closing an AC on a narrowed reading with the reasoning in a PR body is the case
that rule exists for. One sentence in the catalog beside the bounded-answer sentence closes it:
the walk reports the cycle by **omission**, not by a result field, because widening four result
contracts was judged out of proportion. Note it also needs the qualifier from the cycle P2 above
— as written, "bounded" oversells a factorial bound.

---

## Wave input — the serial merge with #799 leaves five published claims unreconciled

Not a git conflict: #795's catalog edit is the `script list` exclusion passage (~552-605),
#799's are ~688+/~746+/~861+, and #799 does not touch `operations.gd` at all. The collision is
semantic, and whichever lands second inherits it.

1. **#795 catalog:** "a vendored checkout that physically lives outside `res://` and is reached
   through a directory link inside it is walked and enumerated exactly as an ordinary
   directory, and its scripts' `class_name`s resolve."
   **#799 catalog:** "a script under a project **nested inside** the resolved one is contained
   and still refused, because its own `res://` references mean the nested root."
   A vendored *checkout* carries a `project.godot` (the same passage calls it "a vendored
   sub-project checked out under `res://`"). After both land, gda enumerates
   `res://vendored/foo.gd` and resolves its `class_name`, and refuses `script validate` /
   `script run` / `resource import` on that same path. Neither passage mentions the other.

2. **ADR-0032's #760 amendment:** a `class_name` "in a vendored checkout reached through a
   link, still does [resolve]" — vs **ADR-0006's #799 amendment**, which refuses a target the
   resolved project does not own. The two ADRs answer the vendored-checkout question oppositely.

3. **#799 catalog:** "`--all` ... enumerates through gda's own `res://` walk, **which excludes
   only `res://.godot`** ... Closing it means changing the shared walk every collector uses."
   After #795 that walk has just been changed, and "excludes only `res://.godot`" now reads as
   the superseded lexical rule.

4. **#795** (catalog + ADR-0032): "identifies what it reaches by filesystem identity ... rather
   than by the spelling that reached it." **#799** (ADR-0006 amendment): the owner walk is
   "bounded ..., **lexical** (it reads the caller's own spelling, so a monorepo's symlinked-in
   shared directory stays inside as the engine reads it)." Two authoritative records, opposite
   adjectives for symlink handling, and nothing says they answer different questions.

5. Sharpest of all, and a ready-made precedent for the P1 fix: **#799's containment rule already
   solves the exact hazard #795 got wrong.** Its catalog text — "a file reached through a symlink
   into the project counts as inside, **except when a `..` traversal could cross that symlink,
   where only the fully resolved location decides**" — names precisely the case where a `..`
   crosses a link and says the resolved location wins. #795's walk hits the same hazard
   (`read_link` returning `../../.godot` under a linked parent) and lets the spelling win. Two
   halves of gda, one hazard, opposite answers.

**Smallest layered reconciliation — enumeration is not targeting.** Both rules are correct;
they answer different questions, and nothing says so. One clause each way, no ADR reopened:

- **ADR-0006's #799 amendment** (it already owns "what belongs to the resolved project") gains
  the layering sentence: *the `res://` walk answers what the project can address and follows
  links by identity; whether a named target may be operated on is this gate, which reads the
  caller's spelling and refuses a target a nearer `project.godot` owns. A linked-in
  sub-project is therefore enumerated, counted and indexed, but not individually targetable
  without `--project <owner>`.*
- **#795's catalog passage** gains a pointer clause: *enumeration is not targeting — see the
  ownership rule below.*
- **ADR-0032's #760 amendment** qualifies its vendored-checkout clause: *…still resolves, for
  enumeration and the `class_name` index; whether that file may be NAMED as a target is
  ADR-0006's question.*
- **#799's `--all` KNOWN GAP** says "excludes only the engine cache, by identity however it is
  spelled (#760)".

Worth noting for #799's author: with the P1 above, its KNOWN GAP is understated — `--all` can
also pull the OUTER project's own `.godot` scripts into the aggregate
(`script validate --all` reports `res://link1/c/root_cache.gd valid: true`).

---

## Red-proofs re-run independently in my own archive

- **Red-proof 3 (file-side revert only) reproduces exactly as claimed.** With
  `_should_collect` → `return true` and nothing else changed, on the PR's own fixture shape:
  leaked = exactly `{res://alias_cache.gd, res://alias_cache.tscn}`; both directory aliases
  (`res://nested/.godot/*`, `res://nested/imported_alias/*`) stayed excluded. The isolation
  claim for the second touch point holds.
- Unmodified head on the same fixture: clean (`res://sub/leaf.gd` only).
- `tests/test_project_walk.py tests/test_e2e_project_walk.py` at head: **9 passed** (34 s).

## Claims that checked out

- **The four `dir_access_unix.cpp` citations are exact** on `4.6.3-stable` (`get_next` 148-183,
  `fix_path` 55-57, `read_link` 481-499, `is_equivalent` 713-729), and the behavioural claims
  they support — the engine `stat()`s a `DT_LNK` entry so a linked directory lists as a
  directory; `fix_path` simplifies lexically so an appended `..` never reaches the kernel — are
  correct.
- **The cost claim holds for ordinary projects.** `project statistics` over
  `examples/platformer/panda-adventure` (279 files, 65 scripts, 12 scenes), 3 runs each:
  base 1.03 s, head 1.09 s, identical counts. The `is_link` lstat per entry is not measurable
  against engine startup.
- **The diamond is correctly not a cycle** (adjudication 3): two links to one legitimate
  directory are both walked; only a link back onto the current chain is refused.
- **Red-proof 3's isolation claim is exact** (see below).

## Shapes probed that came back correct

Directory alias AT the cache; alias INTO a cache subdirectory four levels deep; a link CHAIN
`chain_a -> chain_b -> chain_c -> .godot`; a relative link into the cache from a nested
directory; an alias created INSIDE an aliased directory (absolute target); a nested-deeper cache
FILE alias, absolute and relative; a dangling link (listed, unchanged from base); a link to a
file outside `res://` (listed with its `class_name`, as intended); a **diamond** — two links to
one legitimate directory is correctly NOT treated as a cycle, both branches walked; the
self-cycle `sub/loop -> sub` (terminates, leaf listed once); a 2-cycle (terminates, see the P2
above).

qinhaihong-red:
## Review handled — head `79ac13d0` (was `2bd31608`)

Every finding was re-verified on `2bd31608` before deciding. The P1 reproduces, in **both**
directions, and worse than reported: my fixture adds `L2 -> link1/c`, so the leaked set was
`{res://L2/*, res://link1/c/*, res://link1/ci/cached.gd, res://link1/f_alias.gd}` and
`node add --type RootCacheThing` answered `ambiguous_class_name` naming three leaked paths. Thank
you — the aliasing half really did still decide by spelling.

Engine evidence re-measured on `4.6.3.stable.official.7d41c59c4` against the local `4.6.3-stable`
tag; each engine run had its own `GDA_USER_DATA_ROOT`.

### Decisions

| # | Finding | Decision | Evidence | Change | Validation |
| P1 | `_resolved_link_path` anchors a relative target on the link's res:// SPELLING — leaks under one spelling, hides a vendored checkout's own cache under another | **Fully adopt** (your remedy 2) | Reproduced both directions; base `181da20d` DOES list `res://vendor/x/vendored_sample.gd`, head `2bd31608` does not — regression confirmed | `_fully_resolved_path` + `_resolve_path_segments`: per-component resolution, repeated to a fixed point | 3 new e2e tests; red-proof 4 below |
| P2 | `SYMLINK_PROBE_MAX_STEPS` reused as a path-DEPTH cap; exhaustion ADMITS | **Fully adopt** (deletion) | Reproduced: depth 31 `{}` / depth 32 `{res://alias_deep/deep_cache.gd}` | `range(...)` dropped; the climb ends at the root | red-proof 5 |
| P2 | `MAXSYMLINKS` is 40 on Linux, not 32 | **Fully adopt** | macOS `sys/param.h:197` = 32 (verified locally); Linux `include/linux/namei.h` = 40 (your source) | Constant comment names the LOWER of the two ceilings and states the Linux gap | — |
| P2 | The e2e `class_name` assertion cannot fail | **Fully adopt** | Verified: with `extends RefCounted` and a real leak, `node add` returns `invalid_node_type: registered class_name RootCacheThing is not a Node-derived script: res://alias_cache.gd` — the old assertion **passes** | Cache scripts now `extends Node`; assertion moved into `_assert_class_name_unresolvable`, which fails on the missing `error` key | red-proof 6 |
| P2 | 3 of 7 engine citations wrong | **Fully adopt** | Verified at `4.6.3-stable`: `is_equivalent` fallback **630-632**, script binding **681**, Windows `is_equivalent` **411-431** (442-462 is `read_link`) | Corrected in `_is_in_engine_cache` and in the PR body | — |
| P2 | Cycle outcome: "once every real directory has been walked" | **Fully adopt** (wording) | The chain is per-branch, so it enumerates simple paths | Catalog + `_should_descend` now say distinct `res://` **paths**, "decided and finite", and explicitly *not* "each real directory exactly once" | — |
| P2 | Residual-risk disclosure misdescribes the gap; ADR-0009 miscited (= A3) | **Fully adopt** | The shape you built is one ordinary relative link under one ordinary directory link — an honest project, not ADR-0009's malicious one | That residual no longer exists; the ADR-0009 citation is dropped and the remaining residual is stated by shape (an unreadable link, and hard links) | — |
| P3 | "no alias re-admits it" is false for a hard link | **Fully adopt** (scope) | `is_link` gates both predicates | Catalog and `_is_in_engine_cache` say **symlink** alias and state the hard-link scope as a decision | — |
| P3 | Catalog under-describes duplication (directory link, not only file link) | **Fully adopt** | My new fixture pins it: `res://sub/deep/deep_leaf.gd` **and** `res://link1/deep_leaf.gd` are both listed | Sentence extended to the directory-link case | pinned by the new e2e test |
| A2 | `_resolved_link_path` specified against Unix `read_link` only | **Fully adopt** | Verified `dir_access_windows.cpp:444-462` — returns `GetFinalPathNameByHandleW`, and `f` (never empty) on failure | Both platforms cited; contract restated as "the furthest path it resolved" (= S1); `target == current` guard added — load-bearing now, since without it an unopenable reparse point would cost 32 hops × 32 passes | — |
| A4 | ADR-0006 cited for a symlink rule it does not contain, in 3 places | **Fully adopt** | ADR-0006 has no symlink statement; the rule is `src/gda/project.py` + the catalog containment paragraph | All three re-pointed at the containment gate (`docs/command-catalog.md`, `test_e2e_project_walk.py` ×2) | — |
| A6 | One constant, two limits, one justification | **Fully adopt** | Subsumed by the P2 deletion | The constant now bounds symlink traversals only — chain hops and resolution passes, each of which resolves ≥1 link | — |
| S1 | Comment contradicts code in 2 of 3 cases | **Fully adopt** | `:4384`/`:4386` both return `current` | Rewritten as "the furthest path it resolved" | — |
| S4 | Rewrap artifacts at `:556`/`:564` | **Fully adopt** | 104 and 72 columns mid-paragraph | Three paragraphs rewrapped at 98; word-diff confirms the only prose change is the directory-link clause | — |
| AC4 | Every fixture link uses an ABSOLUTE target; no nested link | **Fully adopt** | `_resolved_link_path`'s relative branch was executed by no test | Two new fixtures with **relative** targets, a link inside a linked directory, and a link→link chain | 12 passed |
| AC3 | The declined result field lives only in the PR body | **Fully adopt** | RULES.md 权威信息源 | Recorded in the catalog beside the bounded-answer sentence, with the reason | — |
| Wave | #799 seam — five unreconciled claims | **Adopt from this side** | Your point 5 is the sharpest: #799's containment rule ("only the fully resolved location decides" when a `..` crosses a symlink) already answers the hazard this PR got wrong — cited as precedent | Catalog gains an **"Enumeration is not targeting"** paragraph pointing at the containment rule; ADR-0032's amendment qualifies its vendored-checkout clause as "for enumeration and this index; whether it may be NAMED is ADR-0006's question"; the three ADR-0006 pointers corrected | — |

Deliberately phrased so nothing here asserts **#799's** outcome: the layering says the two rules
answer different questions, not which way the ownership question is settled. #799's own two bullets
(the ADR-0006 layering sentence and the `--all` KNOWN GAP wording) are left to its fix round, as
are its artifacts. Its KNOWN GAP is no longer understated in the way you flagged — with the P1
closed, `script validate --all` no longer reports `res://link1/c/root_cache.gd valid: true`.

### Not adopted this round — with evidence, not silence

All are real; none has a behaviour impact at this head, and each is a change to a different
decision than the one this PR owns. They are recorded in the PR body's residual-risk section rather
than dropped.

| # | Finding | Why not now |
| --- | --- | --- |
| A1 | The "one owner" claim excepts the non-link fast path, and the guard institutionalizes it | Real: `child == ENGINE_CACHE_DIR` at the descent site is what excludes the cache in an ordinary project. Moving the two cheap answers into the owner changes the shape of the hot path for every entry in every project, which wants its own before/after measurement — larger than a review fix. The claim at `:163-164` is now scoped in the PR body instead of left standing unqualified. |
| A5 | Three co-located notes still say "one directory-exclusion rule"; the guard's subset assertion cannot catch it | Documentation drift with no behaviour impact; the tightening (`_should_collect` added to the required set) belongs with A1's ownership change, not split across two rounds. |
| A7 | ADR-0032's amendment duplicates the mechanism | The amendment must read standalone for someone opening the ADR; cutting it to a pointer trades one drift risk for an unreadable record. Not adopted; the duplication is one paragraph and both copies were updated together here. |
| A8 | `chain`'s default assumes the walk root; the "only a link pays" cost sentence | Both true, neither reachable: all eight call sites pass `"res://"`, and the cost sentence is about the identity probes. Wording. |
| A9 | No glossary term for the `res://` walk | Agreed it wants one — a CONTEXT.md term is a domain-model change with its own review, not a line in a fix PR. |
| S2 | The guard keys on `ENGINE_CACHE_DIR`'s NAME, not the cache literal | Correct and cheap, but it tightens A1's guard; landing it while A1's exception stands would pin the exception. |
| S3 | `_enclosing_function` breaks on `static func` and trailing comments | Latent only — verified nothing in `operations.gd` triggers either shape today. |
| AC2 | The fourth collector's evidence rests on `script_count`/`scene_count`, not `total_files` | Real coverage gap for a re-admission of *typical* cache content (`.ctex`, `.md5`, …). |
| AC1 | The rejected options ("refuse links", "bound depth") are not in the record | AC1 is met; this is completeness of the rationale, and the depth-bound option is arguably now *worse* documented than before, since this round's evidence is exactly why a depth bound admits rather than refuses. |

Also verified, not changed: the issue #760 body cites `core/io/dir_access.cpp:674-675` for
`is_link`/`read_link`; the actual bindings are **667-668**. I did not edit the issue — not
authorized in this round's scope.

### Red-proofs (each reverts one hunk, reruns the real engine)

4. `_fully_resolved_path` → `_resolved_link_path`: **both** new tests fail, in opposite directions —
   the leak test on `{res://L2/imported/cached.gd, res://L2/root_cache.gd, res://L2/root_cache.tscn,
   res://link1/c/…, …}`, and the vendored-cache test on `res://vendored/x/vendored_sample.gd` being
   absent (`AssertionError: {'res://real.gd'}`).
5. Ancestor climb re-bounded by `SYMLINK_PROBE_MAX_STEPS`: the deep-alias test fails with
   `res://alias_deep/deep_cache.gd`.
6. `_should_collect` → `return true` (your red-proof 3's revert): `node add --type RootCacheThing`
   returns `{'script_class': 'RootCacheThing', …}` with **no** `error` key and the assertion fails.
   The same leak under the previous `extends RefCounted` fixture returned `invalid_node_type … is
   not a Node-derived script`, which the old assertion accepted — your finding confirmed by
   construction.

Red-proofs 1–3 from the original round still hold unchanged.

### Gates at `79ac13d0`

`ruff check` clean · `ruff format --check` 481 formatted · `pyright` 0/0 ·
`pytest -m "not e2e"` **2252 passed, 598 deselected** · e2e `test_project_walk.py` +
`test_e2e_project_walk.py` **12 passed** (was 9) · e2e `classname_resolution` + `project` +
`project_analysis` **60 passed** · e2e `script` + `scene` **91 passed**.

**Cost re-measured on your benchmark**, `project statistics` over
`examples/platformer/panda-adventure`, 3 runs each: base `181da20d` 0.61 / 0.61 / 0.64 s, head
0.61 / 0.62 / 0.64 s, identical counts. The component resolution runs only for entries that ARE
links, and an ordinary project has none.

### Still unverified

Windows. The `read_link`/`is_equivalent` contract differences are read from the 4.6.3 source and
documented at the resolver, not executed — the fixtures are POSIX symlinks and CI is macOS/Linux.
The `target == current` guard is reasoned from that source, not measured.

qinhaihong-red:
## Full independent adversarial review — PR #795

Reviewed from first principles at pinned head `79ac13d057577e2bc4b21982ddf76726a823133b` against `feat/gda-dogfooding-dev@181da20d2ea418977600bd73d2958dc4308e8de6`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Artifact review | **Pass** |
| Standards | **Pass** |
| Spec (#760) | **Pass** |

### Artifact review

Pass. #760's “structured outcome” is satisfied by a finite successful structured listing with fabricated cycle paths omitted; the catalog explicitly records omission as the chosen wire policy and explains why no per-path result field was added. ADR-0032 and the command catalog consistently define filesystem identity, cache exclusion, authored symlinks, per-branch cycle termination, and the hard-link boundary.

Pass. Shared traversal owns enumeration; cache identity is owned once and applied by both descent and collection branches; current-chain ancestry owns cycle termination. Legitimate authored symlinks remain an explicit variation point rather than being globally rejected.

### Standards

Pass. No non-tool-enforced repository-standard violation or actionable Fowler smell was found.

### Spec

Pass. The current head records both policy decisions, excludes cache aliases across consumers, terminates cycles without fabricated paths, retains ordinary authored links, provides cross-consumer regressions, and synchronizes ADR/catalog authority.

Sources: issue #760, ADR-0032, the command catalog, `CONTEXT.md`, traversal comments, and tests. Writing reference: ASD-STE100. The complete PR/issue prose was checked for filesystem identity, symlink/hard-link terminology, aliasing, cycle termination, enumeration versus targeting, and omission semantics. No substantive terminology or prose issue was found.

Validation: `git diff --check` passed. Current hosted CI has no failed or pending current required check (one superseded scope-guard run is cancelled); Godot e2e is skipped. Cross-platform filesystem/Godot probes, including Windows, were not rerun.

Issue #685, opened 2026-08-15
Human-mode failure rendering: lay out diagnostics as real lines instead of a single escaped JSON envelope


qinhaihong-red:
## Review-round source

Found during the #651 review (PR #678, wave 1); pre-existing and surface-wide, not introduced by that PR. Without `--json`, a failed command prints the Error envelope as one JSON line with `\n`-escaped `diagnostics`. For most codes the field is short prose and the single line is tolerable. `script run --strict` (PR #678) makes the cost concrete: its `script_failed` diagnostics deliberately carry the script's own stdout and stderr under `--- script stdout ---` / `--- script stderr ---` labels — the evidence a CI operator needs — and the human rendering shows it as one escaped line, blunting exactly the use case the flag exists for.

## What to build

Improve the shared human (non-`--json`) failure rendering so the envelope reads as lines: category/code/message on a head line (or lines), then `diagnostics` printed verbatim as real lines when present. One change in the shared failure renderer, applying to every `Gda error code` — do not special-case `script run`. `--json` output stays byte-identical; exit codes unchanged.

### Verified scope (2026-08-27, pre-dispatch review)

The premise is stronger than the body states: there is **no human rendering path to
improve — there is none at all**. `emit_failure` (`src/gda/headless.py:446`) unconditionally
serializes the envelope with `model_dump_json` and exits; it takes no `json_output`
parameter and has no branch on one. The "single escaped JSON line" a human sees is not a
degraded layout, it is the JSON channel being the only channel.

So the shape of the work is: give `emit_failure` the human/JSON distinction, then render.
That distinction has to reach **five call sites** — 2 in `src/gda/dispatch.py`, 2 in
`src/gda/headless.py`, 1 in `src/gda/hints.py`.

Existing seam to reuse rather than reinvent: `gda.hints._json_in_effect`
(`src/gda/hints.py:195-214`) already answers exactly "is `--json` in effect for this
invocation", and already handles the three ways it can be true — the command's own bound
parameter, the root `--json` inherited via `gda.headless.root_json` (#671), and the literal
token in the recorded argv. Anything this slice invents in parallel to it would be a second
authority for one fact.

**Sequencing:** run this AFTER #687. #687 may add a third optional envelope key (typed
evidence), and the renderer this slice writes is what would have to lay that key out. Doing
it in the other order means writing the layout twice, and both slices touch the same
renderer.

## Acceptance criteria

[...]

PR #796, opened 2026-08-31
feat(gda): typed evidence on the shared failure envelope (#687)

### LINKED ISSUE #687: Typed failure-path evidence on the Error envelope (ADR-0004 decision)

qinhaihong-red:
Closes #687

> **Stacked PR.** The base is `fix/716-717-684-timeout-guidance` (P4 slot 1), not
> `main`, so the diff shown is only this slice's. Merge slot 1 first.

## The decision, and where it is recorded

**ADOPTED.** Recorded as a dated amendment in
[`docs/adr/0004-schema-flag-self-description.md`](docs/adr/0004-schema-flag-self-description.md)
— `Amendment (2026-08-31, #687)`, the authority for the shape, the criterion, the
boundaries and the cost. A companion `Outcome (2026-08-31, #687)` note at the end of
[ADR-0031](docs/adr/0031-headless-script-run-passthrough-execution.md) discharges the
promise its own two amendments made ("this issue's envelope adopts #687's outcome")
and points at ADR-0004 rather than restating it. ADR-0002's `Outcome (2026-08-31,
#716 / #717)` note is left untouched: it stands unchanged, and this change is what
makes its advisory-capture rule workable.

The recommendation this did **not** follow is #687's first triage comment, which
proposed an *operation-scoped* evidence object "published in that command's `error`
schema". That contradicts a recorded ADR-0004 invariant (the `error` half is one
shared schema, identical for every command) — as the issue's own pre-dispatch
correction comment already noted. The adopted shape is the corrected one.

## The envelope's post-#687 shape — for #685 (slot 3)

`GdaError` now carries **three** optional keys, all under the same rule: `probe`
(#667), `hint` (#670) and **`evidence`** (this PR). `evidence` is a
`FailureEvidence` object with five individually-optional fields — `exit_status`
(the CHILD process's status, never gda's exit code), `elapsed_seconds`,
`timeout_seconds`, `termination_phase` (a closed enum: `launched` / `output_seen` /
`aborted_on_error`) and `script_errors` (`list[ScriptError]` — the model
`ScriptRunResult.diagnostics` and `ScenePreflightResult.diagnostics` publish on
their success results. **Not** `script validate`, which publishes a different
model, `ScriptDiagnostic`: do not write one renderer over the two shapes).

The key is **omitted, never null**, on a failure that computed none, and so is every
unset field *inside* it. A human renderer must therefore treat both the key and each
field as absent-by-default, and must not assume that a present `evidence` carries
more than one field.

Three specifics, corrected after this PR's review — the first two are what a
renderer will meet most often:

- **A `--strict` failure emits `{"exit_status": 3, "script_errors": [...]}`**, and
  `script_errors` is `[]` rather than absent when the run recognized none. (An
  earlier draft of this paragraph said the envelope was exactly
  `{"evidence": {"exit_status": 3}}`. It never was.)
- **`script_errors` has three states.** Absent = this failure's channel does not
  parse stderr at all, so read `diagnostics`; `[]` = it parsed and recognized
  nothing; populated = what it recognized. A human renderer that suppresses the
  block on falsiness treats the first two alike, which is right for humans — just
  do not surface `[]` as "no errors were looked for".
- **Every `ScriptError` record carries all four keys** (`kind` / `message` / `path`
  / `line`), `path` / `line` null where the engine named neither, identically on
  the success and failure halves. The recursive `exclude_none` used to strip them
  inside `evidence`; that is the P1 this round fixed.

`category` / `code` / `message` / `diagnostics` are unchanged, and `diagnostics`
still carries the same prose it did before (the recognized-error lines and both
labelled streams) — `evidence` is additive beside it, not a replacement for it.

## What ships

| Producer | Fields |
| --- | --- |
| `launch_timeout` (shared builder → sentinel, export, `resource import`) | `elapsed_seconds`, `timeout_seconds`, `termination_phase` |
| `script run` timeout / `script_aborted` | the three above **+** `script_errors` |
| `script run --strict` → `script_failed` | `exit_status`, `script_errors` |
| `script_not_found` / `script_compile_failed` / `incompatible_script_type` | `script_errors` (the WHOLE parsed list, not just the decider) |
| every other registered code | nothing — the envelope is byte-identical to its pre-#687 form |

`TerminationPhase` moved from `gda.commands.script` to `gda.models`: it is projected
into the shared envelope and reported by every launch-backed channel, so it is a
property of the public contract rather than of one command. `_timeout_phase` and
`_render_captured_errors` moved to `gda.errors` beside the builders that use them,
so one parse of the stderr feeds both the typed key and the prose. Since the review
round, the `<kind>: <path>:<line>: <message>` form itself lives in
`gda.script_errors` as `script_error_line()` — `gda.errors` importing it from
`gda.render` inverted ADR-0040 §5's layer direction. **Slot 3 (#798): your
`_evidence_lines` must import `script_error_line` from `gda.script_errors` and drop
the `kind` prefix it composes itself.**

## Verification

- **Byte-identity regression** (the scope-defining property):
  `test_no_registered_code_grows_a_key_by_defaulting_the_optional_context` builds a
  failure for **every** registered code through the shared builder and requires the
  same four keys. Red-proved by DEFAULTING the parameter —
  `evidence: FailureEvidence | None = FailureEvidence()` in `make_failure` → that
  test red on `Extra items in the left set: 'evidence'`. (An earlier draft of this
  line credited a different mutation — a producer handing an empty
  `FailureEvidence()`. That one leaves this test green, because it never goes
  through a producer; it reddens
  `test_script_run_operation.py::test_an_entry_load_verdict_carries_the_WHOLE_parsed_list`
  instead. The property is guarded; only the attribution was wrong.)
- **Delivery** red-proved by dropping `evidence` in `make_failure` → 8 unit tests red.
- **Publication** red-proved by removing the field from the model → the
  emitted-schema guard red.
- **Real engine** (Godot 4.6.3, local): a wedged `resource import --timeout 4` →
  typed `timeout_seconds` / `elapsed_seconds` / `termination_phase == output_seen`;
  a `--strict` `quit(1)` → typed `exit_status == 1` with gda still exiting 4; a
  failed `preload` → the typed 4-record cascade (3 `parse_error` + `compile_failed`,
  the dependency named in the message and the line carried as data). The last arm
  corrected an assumption in its own first draft: the engine attributes every record
  to the ENTRY script, so the culprit is in the message, not in `path`.
- Gates: `pytest -m "not e2e"` 2268 passed; `ruff check` / `ruff format --check`;
  `uv run --frozen pyright` 0 errors; `pytest -m e2e tests/test_e2e_launch_timeout.py
  tests/test_e2e_script_run.py` 50 passed.

## Cost, disclosed rather than glossed

ADR-0004's "zero per-command maintenance" is not zero per-command *payload*: the one
shared `error` schema is repeated per command in the aggregate manifest. Publishing
`FailureEvidence` takes it from 3,719 → 7,921 bytes and `gda schema` from 675,342 →
979,618 bytes (+45%), measured at the final head (`json.dumps` default separators for
the per-entry `error` schema; the review round's description fixes moved the earlier
interim figures, and ADR-0004/ADR-0012 record the same numbers). It was reduced where it could be (`ScriptError`'s reader-facing prose
moved beside the code; only the one branching rule an agent needs stays in the schema
`description`). Removing the multiplication itself means emitting the shared envelope
once in the manifest and referencing it per entry — a change to ADR-0012's aggregate
contract, out of this issue's scope. Recorded in the ADR so the next key on this axis
is priced before it is added.

## Synced artifacts

`CONTEXT.md` (new **Failure evidence** term), `docs/command-catalog.md` (`script run`
+ `resource import`), `README.md` + the three translations (re-stamped),
`src/gda/skill/SKILL.md`.


qinhaihong-red:
## Independent adversarial review — head `7f5c943bab71edbbcead14d5eee378bb468ab953`

Reviewed as the stacked diff `a5cdc1b672...7f5c943bab` (slot 1's work beneath is not under review), in a private detached archive at that exact SHA. Godot 4.6.3, a private `GDA_USER_DATA_ROOT` per run.

**Verdicts** — Artifact/Markdown: **Changes required** · Architecture/DDMA: **Changes required** · Standards: **Changes required** · Spec: **Changes required**.
One P1, six P2, nine P3. Nothing here disputes the decision itself: **ADOPT, as one universal fixed-shape optional key on the one shared envelope, is the right call and is recorded correctly.** The findings are about what the PR says about that decision, one published-schema contradiction, and one module edge.

### What I re-derived and confirmed (so you do not re-check it)

- **Byte-identity for non-adopting failures holds. No P1 there.** 22 real CLI invocations run against both `a5cdc1b6` and `7f5c943b` and diffed byte-for-byte — `unknown_command` (with `hint`), `unknown_option`, `--schema` near-miss, `project_not_found`, `scene_not_found`, unknown project setting, `script validate` missing, `resource get` missing, unknown export preset, `script run` outside project, `scene preflight` missing, malformed `--params-json`, `game get` with no daemon, `daemon status`, `resource create` bad type, `path_not_found`, `already_exists`, `info --json`. Every one is identical. Both channels: CLI and **gda-mcp** (`_parse_error_envelope` recognizes the envelope and relays it verbatim, `evidence` included — no adapter work, as claimed). Both option paths: argv, `--params-json`, and `--json` written at the root, at the group and after the command. The only failures that changed are the intended five producers.
- **The `error` schema really is one schema.** 76 manifest entries, `distinct error schemas: 1`, at head.
- **The measured cost is right:** 3,719 → 7,391 B per entry; `gda schema` 675,342 → 938,752 B (+38.99%).
- **The prose really is byte-for-byte.** Real-engine `script run` timeout and `script_aborted` envelopes captured on both revisions and diffed with `evidence` stripped and clocks normalized: identical, message and `diagnostics`.
- **All five producers verified on a real engine**, and none can emit an empty `evidence` object: shared `launch_timeout` via a wedged `resource import --timeout 5` → exactly `{elapsed_seconds, timeout_seconds, termination_phase}`; `script run` timeout and `script_aborted` → those three plus `script_errors`; `--strict` → `exit_status` + `script_errors`; `script_not_found` (3-record list) and `script_compile_failed` (3-record cascade, the dependency named only in the message). `incompatible_script_type` did not reproduce on 4.6.3 — a `RefCounted` entry leaves the engine idling and surfaces as `launch_timeout`, exactly as `script run --help` documents. Unit-tested only; not a defect.
- **Gates reproduce exactly:** `pytest -m "not e2e"` **2268 passed**; `ruff check` / `ruff format --check` clean; `pyright` **0 errors**; `pytest -m e2e tests/test_e2e_launch_timeout.py tests/test_e2e_script_run.py` **50 passed**.
- **Two of the three red-proofs are accurate**, re-run in my archive: dropping `evidence` in `make_failure` → **exactly 8** unit tests red; removing the model field → the emitted-schema guard red. The third is not (P2-4).

---

## P1

### P1-1 — The published `error` schema now contradicts, in two places, the omitted-never-null rule the amendment is built on

Shipped in the `error` half of **all 76** manifest entries.

**(a) `FailureEvidence`'s own schema `description` is false.**

> `"Typed evidence about a failure; each field is omitted when this run has none."`

Measured on a real engine:

```
$ gda script run --project P res://strict3.gd --strict --json     # the script just quit(3)
{"error":{…,"evidence":{"exit_status":3,"script_errors":[]}}}
```

`script_exit_status_failure` (`src/gda/errors.py:871-874`) and both gda-ended builders always pass `script_errors=list(...)`, and `[]` is not `None`. So on **every** `script run` failure the field is present, `[]` when nothing was recognized. That is a third state the description denies: *absent* (this channel does not parse — the shared `launch_timeout` builder) vs *`[]`* (parsed, found none) vs *populated*. The distinction is real and useful; nothing publishes it.

**(b) `ScriptError` renders with two different key sets across the two halves of the contract**, because `exclude_none` now filters recursively into it. Same record, same run, one `res://absent.gd` load failure:

```
success  result.diagnostics[0] → {"kind":…,"message":…,"path":"res://absent.gd","line":null}
failure  error.evidence.script_errors[0] → {"kind":…,"message":…,"path":"res://absent.gd"}
```

The published field descriptions promise the first shape and are false for the second:
> `path`: "… **or null** when the engine named none." · `line`: "… **or null** when it reported none."

And `script_errors`'s own description says these are "the same entries a successful run reports as `'diagnostics'`" — they are not, keywise. #687's triage set the bar for exactly this: *"Whatever the ADR adopts must appear in the emitted schema and be held there by a corpus."* No test pins a null-`line` `ScriptError` on the failure path: the e2e asserts `preload_error["line"] == 3` (always present) and `set(err["evidence"]) <= {"exit_status","script_errors"}` — a `<=` that cannot detect a key that is there or not.

**Fix:** decide the `[]`-vs-omitted semantics and publish it; make the `ScriptError` `path`/`line` descriptions site-aware (or stop filtering the nested list); add a regression pinning a null-field `ScriptError` inside `evidence` and one pinning `script_errors: []`. Also worth guarding: `FailureEvidence()` with every field `None` serializes to `"evidence":{}` — unreachable through today's five sites, but only incidentally, and it is a fourth state outside the amendment's argument.

---

## P2

### P2-1 — The `#685` hand-off paragraph in the PR body states a wire shape gda does not emit, and names the wrong model

Two errors in the one section written as slot 3's contract.

> "`exclude_none` filters recursively, so a strict script failure emits exactly `{"evidence": {"exit_status": 3}}`"

False — see P1-1(a). The example is the single worked example in the paragraph, and it is the case #685's renderer will meet most often.

> "`script_errors` (`list[ScriptError]`, the same model `script run` **and `script validate`** already publish on their success results)"

False for `script validate`: it publishes `ScriptDiagnostic` (`src/gda/commands/script.py:485`, `ScriptValidateResult.diagnostics`). Only `ScriptRunResult.diagnostics` is `list[ScriptError]`. Telling the diagnostics-rendering slice that two commands share one model invites one renderer over two shapes.

### P2-2 — `src/gda/skill/SKILL.md` teaches a four-key record that the envelope does not carry

> "`script_errors` (the parsed `{kind, message, path, line}` list)"

Only `kind` and `message` are required, and inside `evidence` the null `path`/`line` are **omitted** (P1-1b). An agent following this writes `e["line"]` and gets a `KeyError` on the common load-failure case. This is the highest-leverage instance of P1-1 because SKILL.md is the contract an agent actually loads.

Same paragraph: "on every `script run` failure gda decided from the engine's stderr" mis-scopes the set — `script_failed` is decided from the exit status, `launch_timeout` from the clock, `script_aborted` from the silence watch, and all three carry `script_errors` anyway. The very next sentence ("Under a `launch_timeout` those errors stay ADVISORY") presumes the presence the clause excludes. ADR-0031's Outcome note repeats the same mis-scoping ("rides every failure this channel **decides from it**" over a list where two of three families are not).

### P2-3 — #687's AC "schema, **help**, and the bundled skill are synced" is not met: no help text was touched

`gda script run --help` still tells the reader to parse the message — the exact defect the amendment says it is fixing:

> `--strict` … "The envelope's **message names the exit status** and its diagnostics carry both script streams"
> `--timeout` … "reports `'launch_timeout'` with the captured output, the elapsed time and one termination phase"

No command help or command description anywhere in the manifest mentions `evidence`. Schema ✅, skill ✅, help ❌.

### P2-4 — The stated red-proof for the scope-defining regression does not discriminate

> "`test_no_registered_code_grows_a_key_by_defaulting_the_optional_context` … Red-proved by making a producer hand an empty `FailureEvidence()` instead of `None` → red"

Re-run in my archive. Mutating `script_did_not_run_failure` to `evidence=FailureEvidence()`:

```
tests/test_error_registry.py .......... 19 passed          ← the named test stays GREEN
full non-e2e suite: 1 failed, 2267 passed
  FAILED tests/test_script_run_operation.py::test_an_entry_load_verdict_carries_the_WHOLE_parsed_list
```

The registry test never goes through a producer — it calls `make_failure(spec.code, "message", "")` directly. The mutation that *does* turn it red is the one its own name describes, defaulting the parameter:

```
-    evidence: FailureEvidence | None = None,
+    evidence: FailureEvidence | None = FailureEvidence(),
→ FAILED tests/test_error_registry.py::test_no_registered_code_grows_a_key_by_defaulting_the_optional_context
   E  Extra items in the left set: 'evidence'
```

The property is genuinely guarded, and the test is load-bearing. Only the reported proof is misattributed — please restate it with the mutation that exercises it.

### P2-5 — `gda.errors` → `gda.render` inverts ADR-0040's declared dependency direction

`src/gda/errors.py:64` imports `render_script_error_location` from the module whose own first line declares it "the presentation layer". ADR-0040 §5 fixes the chain as `cli → commands/* → dispatch → headless → runners / errors / models → foundation`; `render.py` is not in it — its symbols are otherwise imported **only** by `gda.commands.*`, i.e. group altitude. The edge makes the presentation module a transitive dependency of the core.

No cycle today (`render` imports only `gda.models` and `gda.script_errors`), so this is about direction and reasons-to-change: the helper's output is now embedded in `GdaError.diagnostics`, a **wire** field, on three envelopes, while simultaneously feeding human stdout at `commands/script.py:2032` and `commands/scene.py:935`. One authority, two reasons to change; an edit made for readability silently changes the JSON. `tests/test_command_descriptor_registry.py` still classifies the function as a renderer helper.

"`errors.py` already synthesizes prose" does not settle it — everything else it renders is built from primitives inside `errors.py`.

**Fix (~10 lines):** move the location form into `gda.script_errors` as e.g. `script_error_location()`. It is a lexical projection of a type that module owns; `script_errors` imports only `gda.engine_log`; and `render.py:52-54` already concedes it is "a foundation module the presentation layer may name". Then `errors → script_errors ← render`, both downward, and `gda.render` leaves the core's closure. While there: the `f"{kind}: {location}"` composition is now spelled three times (`errors.py:388`, `script.py:2032`, `scene.py:935`).

Related and worth a two-line ADR-0040 amendment: `src/gda/models.py:24` adds a second new core edge, `gda.models → gda.script_errors` (absent at base). Defensible on cohesion, but it is now hard — `script_errors` can never use the "shared field-description constants" ADR-0040 §4 assigns to `models.py` without closing a cycle — and no import-boundary gate exists in this repo to catch either edge. The ADR-0004 amendment is exhaustive about the wire contract and silent about the module topology it changed.

### P2-6 — The "what may enter it" criterion is unenforced, and the current producer set is not recorded — while the same tree still contains the discard #687 exists to close

`src/gda/commands/scene.py:1099-1105`: `_ended_before_the_verdict` **receives** `diagnostics: list[ScriptError]` (parsed at `scene.py:1011`), uses it in the success branches, and then returns `make_failure("operation_failed", …, raw.stderr)` — dropping it. Every clause of the criterion is met: already computed on the failure path, unrecoverable without re-parsing prose, and it changes what the caller does next (which autoload called `quit()`). That is #651's discard, alive in this PR's own tree.

Three more failures name a computed number only in prose: `engine_crashed`'s signal (`errors.py:527-531`, inside `classify_launch_or_crash` — the function whose docstring argues that being "the single home is what makes the timeout evidence a property of every channel", yet only one of its branches publishes evidence), `resource import`'s child exit code (`commands/resource.py:1112-1117`), and `export run`'s (`commands/export.py:355-359`).

I do **not** think #687 had to serve these — the issue scoped to `script run` and #655's timeout envelope, and expanding here would be scope creep. What is missing is the boundary: the amendment states a forward-looking criterion with no test, type or gate behind it (`test_error_registry.py` pins only the negative case; `test_schema_command.py` pins only the field set), and records no current producer set. A reader cannot tell the four above from an oversight.

**Fix:** name the adopting producer set in the amendment (or in the `FailureEvidence` comment) as this decision's recorded boundary, and file the expansion as a follow-up. Cheapest useful gate: assert the producer set in `test_error_registry.py` alongside the existing negative case.

---

## P3

1. **`docs/adr/0004…md`: "across 72 commands" — measured 76** at both endpoints. The byte figures and the +39% are exact; only the denominator is wrong, and this PR's own new comment in `src/gda/script_errors.py` says "**~76 times**" — the ADR and the code disagree inside one commit. In the one paragraph whose whole value is that the cost was measured, a wrong denominator gives a future reader 51.6 KB/command instead of 49.0.
2. **"`ScriptError`'s reader prose moved beside the code" is a deletion, not a move, for two facts.** `grep -rn` over the whole tree at head: *"in canonical form ('.'/'..' segments and duplicate slashes collapsed)"* and *"engine-side load errors carry no script line"* and *"never a synthesized number"* — **zero hits**. The `path` case is fine in substance (`canonical_res_path`'s docstring covers it), but "canonicalized" is now undefined in the published schema; the `line` facts survive nowhere. Unmentioned entirely: the `ScriptErrorKind` **enum's whole schema `description` was deleted** (1,288 → 314 B), which reduces the published contract of `script run` and `scene preflight`'s **success** results — commands #687 does not otherwise touch.
   For the record, the mitigation is materially worth it: I re-measured with only `script_errors.py` reverted — 9,078 B/entry and 1,069,402 B aggregate, i.e. the slimming removes **33% of the growth**. Say that, rather than "reduced where it could be".
3. **The ADR-0012 follow-up is untracked.** No issue owns it (searched), and `docs/adr/0012-*.md` — which carries Outcome notes for #230/#233/#669/#601 — has none for #687, so the document whose aggregate contract now repeats a 7,391-byte schema 76 times does not know it. RULES.md's 权威信息源与状态同步 puts 后续事项 in the tracker or in a ready-to-create draft; ADR prose in a different ADR is neither. Note the payoff is measured and the seam is intact (all 76 `error` values byte-identical): hoisting to a manifest-level `$defs` + `$ref` is mechanical and lands the manifest *below* its pre-PR size.
4. **The criterion now exists in three places, two of which claim to be the only one.** `docs/adr/0004…md:215` "*stated once* so a later field is added deliberately", `src/gda/models.py:120` near-verbatim with the same "*stated once*", and `CONTEXT.md:248-252` a third rendition. This is a rule governing future amendments — the category least likely to be edited in all three places. `Near-miss hint` one entry away in CONTEXT.md has the better pattern: name the authority, do not reproduce the rule.
5. **`docs/command-catalog.md:749-754` omits `script_errors` from the two gda-ended envelopes** in an enumeration a reader treats as exhaustive — both `script_run_timeout_failure` and `script_run_aborted_failure` set it. The PR table and ADR-0031 get this right; the catalog, the feature spec, is the outlier.
6. **`src/gda/commands/script.py:1260-1264` — "Re-exported here because this module's readers reach for it under this name" is false.** There is no `__all__`; the name survives only because line 1680 uses it, and ruff F401 would delete it otherwise. ADR-0040's Considered Options explicitly rejected re-export façades ("a second import surface that can only drift") — and the drift is already here: `tests/test_classify_launch_or_crash.py:32` imports `TerminationPhase` from `gda.models`, `tests/test_script_run_operation.py:45` from `gda.commands.script`. Delete the sentence; point the test at `gda.models`. (The *placement* in `models.py` is right — ADR-0040 §4/§5.)
7. **Stale prose introduced or left by the move.** `src/gda/headless.py:538-544` — the docstring of the emit path the whole additive argument rests on — still names only `probe` (#667), now two keys behind. `src/gda/render.py:47` still says "Two groups render recognized script errors now"; there are three consumers and the third is not a group, it is `gda.errors` putting the output on the wire. The commit message says the shared builder covers "all four launch-backed channels" — `script run` reaches `launch_timeout` through its **own** builder; the PR body's table correctly says three.
8. **`CONTEXT.md` — `_Avoid_: diagnostics` bans a live field name the entry's own body uses correctly** ("the captured streams (already in `diagnostics`)"). Every other `_Avoid_` list in the file holds non-terms; the one precedent for listing a live term, `Completion marker`'s `_Avoid_: sentinel`, earns it with an explicit "**Not** the ADR-0002 … sentinel" in the body. Add that clause or drop the term. Same entry: "the axis `probe` and `hint` established" drops the relative pronoun, and "neither GDScript surface emits or reads it" names neither (ADR-0004:103 names them in the next clause; CONTEXT.md is the document read *for* the shared language).
9. **ADR-0002's forward pointer to #687 is left undispositioned.** Its `Outcome (2026-08-31, #717)` says the ceiling's provenance "if it is ever wanted as data, belongs on #687's evidence object". The amendment lists what the criterion excludes (silence window, marker, streams) but never answers #717, so a reader following the pointer finds nothing. It matters in practice: `evidence.timeout_seconds` alone does not say whether the caller can raise the ceiling — only `resource import` and `script run` expose `--timeout`; the sentinel's 60s and the export's 600s are fixed, and only the prose says so.

---

### Adjudicated, no change needed

- **One universal `FailureEvidence` over a per-command `error` schema** — correct, and the decline is recorded honestly. The quote *"the one shared `GdaErrorEnvelope` schema, identical for every command"* is verbatim from ADR-0004's own lines 53-54/271, and I confirmed the invariant empirically (76 entries, one schema). The PR's account of #687's first triage recommendation is accurate and fair — it credits the correction comment and does not revive the "four callers" argument that comment softened.
- **`TerminationPhase` in `gda.models`** — right owner per ADR-0040 §4/§5; only the comment is wrong (P3-6).
- **The single-parse refactor** (`commands/script.py:1664`, one `parse_script_errors` feeding both the typed key and the prose) — a genuine DRY improvement, and I verified the prose it produces is byte-identical to the pre-#687 output.
- **CLI-side boundary** — no change under `src/gda/ops/` or the live path; `OperationError` stays `extra="forbid"`, `LiveError` unchanged. Confirmed.
- **README i18n** — all three stamps match the new `README.md`, translations faithful; pinned by `tests/test_readme_i18n_sync.py`, which passes.
- **`f"{phase}"` → `f"{phase.value}"`** looked like a silent prose change; it is not — base typed the parameter `str` and callers passed `.value`. Verified end-to-end on a real engine before raising, and dropped.

qinhaihong-red:
## Review handled — head `837174d3`

Rebased onto slot 1's fix first (`21869a7b`; three i18n-stamp conflicts, both slices' README prose kept, re-stamped), then one fix commit. **1 P1 + 6 P2 + 9 P3: 14 adopted, 2 partially adopted, 0 declined outright** — plus one decline *inside* the P1 (the alternative you offered) and one inside P2-5. Thank you for the byte-identity re-derivation and the five-producer engine sweep; both saved a full re-verification pass.

Gates on `837174d3`: `pytest -m "not e2e"` **2273 passed** (2268 + 5 new), `ruff check` / `format --check` clean, `pyright` **0 errors**, `pytest -m e2e` on `test_e2e_script_run.py` + `test_e2e_launch_timeout.py` + `test_e2e_scene_preflight.py` **62 passed**.

---

### P1-1 — the published `error` schema contradicting itself

**Fully adopt, and I took the option you listed second-to-last: stop filtering the nested list.**

Both halves reproduced verbatim on Godot 4.6.3 before changing anything — `--strict` on a `quit(3)` really emits `"script_errors": []`, and the same `res://absent.gd` record really came back `{kind, message, path}` inside `evidence` against `{kind, message, path, line: null}` on the success result.

**(b) — the decision.** You offered two ways out: stop the recursive filter, or publish the asymmetry precisely. I checked the cheap one first, and it is cheaper than it looks — the emitted schema already marks `path` / `line` non-required at *both* sites (`required: ["kind","message"]`), so neither shape is schema-invalid and prose alone would have closed the contradiction. I still went with the code fix, on three grounds:

1. `script_errors`' own description promises "the same entries a successful run reports as `diagnostics`" — a promise about *identity*, which prose can only retract, not keep.
2. This repo has already adjudicated this class of question the other way. `CONTEXT.md`'s `Value projection`: *"One projection shared across every value gda emits — so a value reads the same everywhere."*
3. Publishing the asymmetry would have made SKILL.md's four-key teaching (P2-2) permanently wrong and forced every consumer — #798's renderer, and any agent — onto `.get()`. Fixing the shape makes that teaching *true* instead.

Scope of the fix is exactly the boundary, not the rule: the envelope's optional keys and `FailureEvidence`'s own fields still omit-when-None (that is what buys byte-identity); a nested model that is **also published on a success result** keeps its full key set. One `@field_serializer` on `FailureEvidence.script_errors` in `gda/models.py`, ~10 lines with the reasoning beside it. I deliberately did **not** put it on `ScriptError` as a `model_serializer`: a plain one enumerates the keys, so a field added later would be silently dropped from *all* output — a worse trap than the one being fixed.

Real-engine proof at head, same run, both halves:

```
failure  error.evidence.script_errors[0] → {"kind":…,"message":…,"path":"res://absent.gd","line":null}
success  result.diagnostics[0]           → {"kind":…,"message":…,"path":"res://absent.gd","line":null}
```

**(a) — three states, kept and published.** `[]` is not collapsed to `None`: the distinction is load-bearing exactly as you say. It is now in the field `description`, in SKILL.md, in the catalog and in the ADR: *absent* = this channel does not parse stderr, read `diagnostics`; *`[]`* = parsed, recognized none, itself a finding; *populated* = what it recognized. `FailureEvidence`'s docstring no longer claims omission for a field that ships `[]`.

**The fourth state** (`FailureEvidence()` → `"evidence":{}`) is now closed by a test rather than by luck — see P2-6.

**Regressions, all red-proofed (mutation reverted after each):**

| Guard | Mutation that reddens it |
| --- | --- |
| `test_a_nested_script_error_reads_the_same_on_both_halves_of_the_contract` (unit) | drop the full-shape serializer |
| `test_script_run_missing_script_is_script_not_found` (**e2e**, real engine, null-`line` record) | same |
| `test_an_empty_script_errors_list_is_published_as_a_finding_not_as_absence` (unit) | `return None if not errors …` |
| `test_script_run_strict_fails_on_an_explicit_non_zero_quit` (**e2e**) | same |

The e2e `<=` you flagged is now `==`, plus an explicit `evidence["script_errors"] == []` — so it discriminates the case it was blind to.

---

### P2-1 — the #685 hand-off paragraph

**Fully adopt.** Both errors corrected in the PR body: the worked example is now the real `{"exit_status": 3, "script_errors": [...]}` with the `[]` case called out, and `script validate` is named as publishing `ScriptDiagnostic`, **not** `ScriptError`, with an explicit "do not write one renderer over the two shapes".

**You asked me to check whether #798 was misled — it was not.** `_evidence_lines` (`render.py:114-142` on `97feae2a`) reads the **typed** `FailureEvidence` / `ScriptError` objects, not decoded JSON, so the recursive-`exclude_none` asymmetry never reached it and `render_script_error_location` already handled null `path` / `line`. Its `if evidence.script_errors:` conflates `None` and `[]`, which is the right call for a human channel (no header over nothing). One real hand-off for #798's round, added to the PR body: after P2-5 the helper is `script_error_line` in `gda.script_errors` and it now composes the `kind` prefix itself, so `_evidence_lines` needs an import change and one less f-string.

### P2-2 — SKILL.md's four-key record

**Fully adopt.** After P1-1(b) the four keys are always present, so the teaching became true rather than needing to be softened; I added that `path` / `line` are `null` where the engine named neither ("the normal case for a load error") and the three-state paragraph. The mis-scoping is fixed in both places: SKILL now says `script_errors` rides *every* `script run` failure, and ADR-0031's Outcome note now says the same and spells out **why** the set is wider than "decided from stderr" — only the point-1 verdicts are, `script_failed` is decided from the exit status and the two gda-ended runs from the clock and the silence watch.

### P2-3 — the AC's "help is synced"

**Fully adopt, and it was worse than reported.** `--timeout`'s help ended `"Reported as prose in the message, not as envelope fields (#655)."` — not merely unsynced but flatly false at head. Five sites now name `evidence`: both `--strict` help strings, both `--timeout` help strings (typer option + params-model `description`, which is the one the manifest publishes), and `script run`'s command description.

### P2-4 — the misattributed red-proof

**Fully adopt.** Reproduced your result exactly: mutating a producer to `evidence=FailureEvidence()` leaves `test_error_registry.py` at 19 passed and reddens `test_an_entry_load_verdict_carries_the_WHOLE_parsed_list` instead; the parameter default reddens the named test on `Extra items in the left set: 'evidence'`. The PR body now credits the mutation that exercises it and says plainly which draft claim was wrong.

### P2-5 — `gda.errors` → `gda.render`

**Fully adopt the fix; decline the "while there" only in its shape, not its substance.** Moved as `gda.script_errors.script_error_line()`. `script_errors` imports only `gda.engine_log`, so `errors → script_errors ← render` and both edges point downward.

On the 3× composition: rather than move the location form and leave four sites spelling `f"{kind}: {location}"`, I folded the `kind` prefix into the one function — all four consumers (and #798's fifth) wanted exactly that string, so there is one form instead of two. Prose output is byte-identical; the e2e and human-output suites pin it.

Two things I added beyond the suggestion, both because "no import-boundary gate exists" is the part that lets this recur:

- `tests/test_render.py::test_only_the_group_layer_imports_the_presentation_module` — static, over the whole source tree, asserting only `gda/commands/*` imports `gda.render`. Red-proofed by re-adding an `errors → render` import. Deliberately **narrow**: it pins the one direction this review found inverted and is not a general boundary gate, which would be its own decision.
- An **ADR-0040 outcome note** covering both edges, including the `gda.models → gda.script_errors` one you flagged: recorded as downward-and-consistent but *not free*, naming the exact consequence (`script_errors` can never use §4's shared field-description constants without closing a cycle) and why cohesion still wins.

`_HELPER_RENDERERS` in `test_command_descriptor_registry.py` no longer lists it.

### P2-6 — unenforced criterion, unrecorded producer set

**Fully adopt the boundary; decline the widening, as you recommended.** Scope unchanged — `scene.py`'s `_ended_before_the_verdict` still discards its parsed diagnostics.

- ADR-0004's amendment now names the **five adopting producers** as this decision's recorded boundary, and separately names the four facts that **meet the criterion and are still in prose** (`_ended_before_the_verdict`, `engine_crashed`'s signal, `resource import`'s and `export run`'s child exit codes) so a reader can tell the decision from an oversight.
- `test_error_registry.py::test_only_the_recorded_producers_put_evidence_on_the_envelope` asserts that set by AST over `gda/errors.py` — a sixth producer cannot join without the ADR paragraph being revisited in the same change. Red-proofed by adding one.
- `test_no_producer_can_emit_an_empty_evidence_object` closes your "worth guarding" fourth state on all five. Red-proofed by making a producer pass a bare `FailureEvidence()`.
- A dated comment at `scene.py`'s discard site names it, says every clause of the criterion is met, and points at the ADR paragraph.

---

### P3

1. **72 → 76.** Fully adopt, and re-measured at this head rather than carried over: 3,719 → **7,825** B per entry, `gda schema` 675,342 → **972,433** B (**+44%**), 76 commands. Your base figure reproduced exactly (675,342), which pinned the measurement method.
2. **"Moved" that was deletion.** Partially adopt. The three facts are restored as a comment block above `ScriptError` (canonical in `canonical_res_path`'s sense; load errors carry no line; a `push_error` line is the engine's backtrace call site, never synthesized), and the ADR now **discloses** rather than glosses: that the slimming also shrank the SUCCESS schemas of `script run` / `scene preflight`, with the numbers (`ScriptErrorKind` 1,288 → 314 B, `ScriptError` 1,962 → 1,249). One correction to your reading: the enum's description was not deleted but reduced to its one-line summary — the docstring's first line still ships. I did **not** restore the full enum `description` to the schema: its substance is already in the `kind` field description a reader gets at the same site, so restoring it would buy duplication at 76× the price. Your mitigation figure re-measured at this head: 1,103,083 B unslimmed vs 972,433 B, so the slimming removes **31%** of the growth — stated in the ADR in place of "reduced where it could be".
3. **ADR-0012 + follow-up.** Fully adopt the note; **issue not filed**, draft below.
4. **"Stated once" ×3.** Partially adopt. The claim is removed from both restatements; ADR-0004 is named as the authority in each (`gda/models.py` and `CONTEXT.md`, the latter now following the `Near-miss hint` pattern you pointed at). I kept a short form of the criterion in both rather than reducing them to bare pointers — a reader at the model wants to know what qualifies without a round trip; what they no longer get is two documents each claiming to be the only one.
5. **Catalog omission.** Fully adopt — the enumeration now says `script_errors` rides *all* of them, and carries the four-key/three-state facts.
6. **False "re-exported here".** Fully adopt: sentence deleted, comment now says it is imported because the module uses it and names `gda.models` as the one import surface (citing ADR-0040's rejected façades); `test_script_run_operation.py` now imports `TerminationPhase` from `gda.models`, like its sibling.
7. **Stale prose.** Fully adopt. `headless.py`'s `emit_failure` docstring names all three keys and states the nested boundary; `render.py:47` went with the function; the commit message's "all four launch-backed channels" is corrected to the three on the shared builder, with `script run`'s own builder named.
8. **`CONTEXT.md`.** Fully adopt all three: the entry now earns `_Avoid_: diagnostics` with an explicit **Not**-clause contrasting the two (as `Completion marker` does for `sentinel`), restores the relative pronoun, and names both GDScript surfaces (`OperationError`, `LiveError`).
9. **ADR-0002's #717 pointer.** Fully adopt — and the answer is a **decline, on the criterion**, recorded in the amendment so the pointer resolves. The ceiling's provenance fails the second clause: it is recoverable without reading anything, from the invocation the caller itself made — only `resource import` and `script run` expose a `--timeout` at all, so a caller that passed one knows the ceiling is its own. That is the same ground the silence window and the declared marker are excluded on. `timeout_seconds` still ships the VALUE, which is not recoverable that way.

---

### Follow-up issue draft (not filed)

> **Title:** Hoist the shared `error` envelope to a manifest-level `$defs` in `gda schema` (ADR-0012)
>
> **Body:** `gda schema` repeats the full `GdaErrorEnvelope` schema inside all 76 manifest entries. The repetition predates #687, but that amendment made it the dominant term: one shared change moved the dump from 675,342 to 972,433 bytes (+44%), and every future key on the failure envelope is priced at 76×.
>
> All 76 `error` values are byte-identical (verified at `837174d3`: `distinct error schemas: 1`), so hoisting the envelope to a manifest-level `$defs` and referencing it per entry is mechanical, and lands the document **below** its pre-#687 size.
>
> It changes the per-entry shape gda-mcp reads, which is ADR-0012's contract, not ADR-0004's — hence a separate issue rather than a #687 fix. Scope: the aggregate emit path, the `$ref` resolution gda-mcp's transform needs (or a decision that it resolves nothing and consumers do), and the schema round-trip tests. Not urgent: startup cost is one dump plus one parse.
>
> Recorded as an Outcome note on `docs/adr/0012-gda-mcp-tool-surface-generation.md` at `837174d3`.

---

### Verification

- **Byte-identity for non-adopting failures still holds at `837174d3`**: 12 real CLI invocations diffed against the rebase base `21869a7b` — `unknown_command` (with `hint`), `scene inspect` near-miss, `--schema` near-miss, `project_not_found`, `script validate` missing, `resource get` missing, `script run` outside project, unknown export preset, `game get` with no daemon, `daemon status`, `node add` on a missing scene, malformed `--params-json`. All identical.
- **Real engine (Godot 4.6.3, private `GDA_USER_DATA_ROOT` per run)**: the P1's decided shape verified on both halves of the same record; `--strict` `[]`; the e2e arms above.
- **Five new guards, five discriminating red-proofs**, each mutation reverted.
- **Not verified**: `incompatible_script_type` still does not reproduce on 4.6.3 (unchanged from your finding — unit-covered only). The schema-aggregation follow-up is drafted, not filed, per instruction.

Head is **`837174d3`** — #798 rebases onto it, and its `_evidence_lines` needs the `script_error_line` import change noted above.

qinhaihong-red:
## Full independent adversarial review — PR #796

Reviewed from first principles at pinned stacked head `837174d36e485599b09007cd6ebe33f17b5282f3` against `fix/716-717-684-timeout-guidance@21869a7b0c4352051560708848157f908a49f72f`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Domain modular architecture | **Changes required** |
| Standards | **Changes required** |
| Spec (#687) | **Changes required** |

### Artifact review — required changes

1. `src/gda/skill/SKILL.md:95-97` says `script_errors` appears “on every script run failure,” but failures such as `stdout_spill_failed` do not parse stderr and carry no such evidence. Pre-launch environment/addressing failures also disprove the blanket statement. Name the actual producer set, or say “the listed `script run` verdicts that parse the run's stderr.”

2. The PR body reports `gda schema` growing from 675 KB to 939 KB (+39%), while ADR-0012 records 675,342 to 972,433 bytes (+44%) for the same current head. An independent exact-head run produced **972,433 bytes**, confirming the ADR figure. Update the PR body or identify a different measurement boundary; do not leave two incompatible current cost claims.

Minor: ADR-0012 says manifest hoisting is “Tracked as a follow-up” without a durable issue identifier. Link the tracker or state that it is recorded but untracked.

### Domain modular architecture — required change (`DDMA-796-1`)

`src/gda/models.py:149-154` defines `timeout_seconds` as the ceiling the run **reached**. `src/gda/errors.py:1017-1029` says `script_aborted` did **not** reach the timeout and nevertheless populates that same field with the configured bound. One public evidence field therefore has two owners and two meanings: observed reached ceiling and unused caller configuration.

This breaks semantic DRY and makes a generic evidence consumer interpret an early abort as timeout-bound evidence. Omit `timeout_seconds` from `script_aborted`; the caller already knows its configured input and the message retains it. If the unused configured bound must be typed, give it a distinct name and invariant.

### Standards

The same field-semantic conflict is a hard violation of the repository's consistency and single-authority rules. No additional actionable Fowler smell was found.

### Spec — required changes

1. #687 explicitly requires `script_not_found`, `script_compile_failed`, `incompatible_script_type`, and `script_failed` to carry the child's numeric status plus parsed errors. `script_did_not_run_failure` (`src/gda/errors.py:801-836`) emits only `script_errors`; only `script_failed` emits `exit_status`. Either deliver the named status data (including the engine's `0` for never-ran verdicts, with semantics documented) or reconcile #687 to record why those three statuses were declined.

2. #687 says the timeout envelope's capped streams adopt the typed shape. ADR-0004 deliberately keeps them only in `diagnostics` to avoid duplicating two 16 KiB captures. That can be a valid decision, but the issue was not reconciled with it. Amend #687 to record the explicit decline/rationale (or add typed stream fields) before closing it.

Sources: `CONTEXT.md` definitions for **Error envelope** and **Failure evidence**; ADR-0004/0012/0031/0040; issue #687; all changed README/catalog/Skill and translations. Writing reference: ASD-STE100. Producer-set wording, optional/omitted/null semantics, nested `ScriptError` shape, shared-schema ownership, and cost claims were inspected. Findings are above; terminology is otherwise stable.

Validation: `git diff --check` passed; 306 focused model/script-run/schema/registry tests passed. Exact `gda schema` output was 972,433 bytes. Current hosted CI has no failed or pending required check; Godot e2e is skipped and was not rerun.

qinhaihong-red:

New head: `883b17b41` (one commit on top of the reviewed `837174d3`).

| `DDMA-796-1`: `timeout_seconds` has two owners/meanings — the field says "ceiling this run reached", `script_aborted` ships the ceiling it did NOT reach | **Fully adopt** | ADR-0004's own admission criterion already excludes the caller's own inputs (silence window, marker); an unreached `--timeout` is the same kind of fact — the factory's old comment even conceded it while shipping the value | `script_aborted` omits `timeout_seconds`; its value stays in the message. ADR-0004 records the exclusion beside the criterion; catalog/SKILL/help updated | Abort-shape test flipped FIRST and observed red at `837174d3`, green at head; 320 focused + 2298 full |
| SKILL says `script_errors` rides "every `script run` failure" | **Fully adopt** | `stdout_spill_failed` and pre-launch refusals carry none | Reworded to "the `script run` failures that parse the run's stderr —" + the existing enumeration | SKILL diff |
| PR body's 675 KB → 939 KB (+39%) conflicts with ADR-0012's 675,342 → 972,433; independent exact-head run confirms the ADR | **Fully adopt** | Reproduced: 972,433 at `837174d3` (`gda schema \| wc -c`); the body carried an interim-measurement figure | Body corrected — and re-measured at the FINAL head, since this round's description fixes move it again: **979,618 bytes (+45%)**, shared `error` schema 3,719 → **7,921** (`json.dumps` default separators). ADR-0004 and ADR-0012 record the same final numbers | Measured at head; figures now agree across all three surfaces |
| ADR-0012 "Tracked as a follow-up" names no tracker | **Fully adopt** | The `$defs` hoist has a draft awaiting user triage, no issue id yet | Reworded: "Recorded as a follow-up; not yet tracked by an issue." | ADR diff |
| Spec: #687 requires the child's numeric status on `script_not_found`/`script_compile_failed`/`incompatible_script_type` | **Partially adopt** (reconcile, not deliver) | Godot exits `0` for a missing/non-compiling entry point (`script_did_not_run_failure` docstring): typing that `0` would publish a phantom success number for a run whose failure is that the script never ran; the recorded field contract is "a failure whose verdict IS that status" | #687 amended: dated scope-reconciliation note records the decline and its ground; AC2 annotated with the delivered shape | Issue body diff |
| Spec: #687 says the timeout envelope's capped streams adopt the typed shape; ADR-0004 keeps them prose-only without the issue recording it | **Fully adopt** (reconcile) | ADR-0004's criterion: copying two 16 KiB captures into `evidence` doubles the payload to say the same thing twice; elapsed/phase ARE typed | Same #687 note records the decline; the abort narrowing above is recorded there too | Issue body diff |

Schema/help surface diff of this round's commit, enumerated: the `timeout_seconds` field description and the `--timeout` param help changed (text only, no shape change), moving `gda schema` to the figures above.

qinhaihong-red:

Reviewed at `883b17b41f3b174a13a1b5117468c26416f67ecd` against current stacked base `21869a7b0c4352051560708848157f908a49f72f`.

**Verdict: Changes required**

- Artifact: Pass
- DDMA: Pass
- Standards: **Changes required**
- Spec: Pass

The prior findings are closed: `script_aborted` now omits the unreached `timeout_seconds`; the `script_errors` producer set, schema measurements, ADR cost disclosure, and issue #687 scope decision are reconciled.

### Required — typed timing evidence still claims a causal distinction it cannot establish

The #687 additions still state that ceiling, elapsed time, and termination phase distinguish a merely slow run from a stuck run:

- `src/gda/errors.py:440-444`, with the same overclaim also at `400-404`
- `src/gda/errors.py:932-935`
- `docs/adr/0004-schema-flag-self-description.md:270-272`
- `docs/command-catalog.md:912-914`
- the corresponding test comments

Two runs can have the same elapsed time, reached ceiling, and termination phase while one is a healthy overrun and the other is wedged. Captured diagnostics or progress evidence may support a diagnosis, and rerunning with another bound may disambiguate it, but the three numbers do not establish the cause by themselves.

This also conflicts with the adjacent text in `errors.py:428-435`, which correctly refuses to infer “stuck” and states that partial capture is not a verdict.

**Smallest fix:** sweep the introduced prose and test comments to say that typed evidence quantifies the reached bound, elapsed time, and startup/progress state and supports choosing a rerun bound; explicitly state that it does not distinguish slow from stuck by itself.

Validation: 460 focused tests passed; `gda schema` regenerated to exactly 979,618 bytes; the exact-head diff passes `git diff --check`; current required CI has no failed or pending checks.

qinhaihong-red:
## Round-3 response

New head: `883b17b41` → `300a7418c` (one docs/comments commit).

| The #687 additions still claim ceiling/elapsed/phase distinguish a merely slow run from a stuck one (errors.py ×3, ADR-0004, catalog, test comments) — conflicting with the adjacent text that refuses exactly that inference | **Fully adopt** | Two runs can carry identical numbers while one is a healthy overrun and the other is wedged; the neighbouring #717 remediation paragraph already calls that inference unearned | All nine copies swept: the evidence quantifies the reached bound, the duration and how far the run got, and supports choosing the rerun bound; `launched` vs `output_seen` stays a real distinction; none of it names slow-versus-stuck by itself — the capture carries the progress. Fixed in `errors.py` (three docstring passages), ADR-0004's provenance note, the catalog's `resource import` passage, and four test comments (`test_classify_launch_or_crash` ×2, `test_e2e_launch_timeout`, `test_script_run_operation`, `test_e2e_script_run`). Historical statements about what the pre-#655 envelope LACKED stay as history | Docstrings/comments only: `gda schema` byte-identical at **979,618** (re-measured); 156 focused + 2378 chain-tip tests green; #798/#799 rebased onto the new head (clean) |

Prior rounds' closures stand as verified by this re-review.

PR #797, opened 2026-08-31
fix(gda): refuse a --value literal Godot's float parser destroys (#772)

### LINKED ISSUE #772: node set / game set --value: String.to_float() silently changes or zeroes a written float

qinhaihong-red:
Closes #772

## The decision

**Refuse what the parser destroys; disclose what it merely drifts.** A `--value`
literal that Godot's own float parser turns into `0.0` although the caller wrote no
zero, or into `NaN`, is refused as the existing `uncoercible_value` /
`live_uncoercible_value` (exit 4, target untouched). Low-order drift is accepted and
documented.

This is #752's answer, reached from #752's principle rather than inherited from it. A
write PERSISTS — into a `.tscn`, a `.tres`, `project.godot`, a running node — so gda
does not store a number the caller never sent. What differs is HOW the question is
asked, and the reason is that the two paths know the literal at different times:

- on the live wire **gda** spells the literal, so the refusal must PREDICT the outcome
  from the value (`live_numbers.wire_flattens_to_zero`);
- on a write the **caller** spells it, and the engine has already answered by the time
  the coercion returns, so the coercion OBSERVES what the parser produced.

Observing is what lets one rule cover three mechanisms instead of one, including two
that no model of the applied exponent can see. It also buys a remedy the wire cannot
offer, because the caller owns the spelling: the refusal names it.

Observing also draws the edges, both now documented. An **overflow is not refused**:
`1e400` reads as `inf`, the correctly-rounded IEEE-754 answer for a magnitude past
binary64's top and the engine's number for "larger than it can hold" — in the direction
the caller asked for, not a different number put in its place — and it is stored as
`inf`. `0e600` reads as `NaN`, which is not the value, not near it and not in its
direction; that is the substitution the rule exists to stop. What does NOT separate
them is loudness: a stored `NaN` reports as JSON `null` exactly as a stored `inf` does,
so the accepted overflow carries a residual of its own, now disclosed — `inf` is stored
but never reportable, so the echo and every later `get` read `null`. A literal **below
binary64's reach is refused anyway** (`1e-400` fails exactly as `1e-320` does, although
`0.0` is the correctly-rounded answer there — the coercion cannot separate a true
underflow from the engine's −309 cliff without modelling the parser it deliberately
asks instead).

## The measurement

`String::to_float()` is `built_in_strtod<char32_t>(get_data())`
(`core/string/ustring.cpp:2677`), and `JSON.parse_string` reaches the same function for
a number (`core/io/json.cpp:393`) — one parser, two callers, so the write direction
reuses `tests/live_number_corpus.py` rather than forking a second corpus.

Measured on **Godot 4.6.3.stable.official.7d41c59c4** over 179 literals: every corpus
row in BOTH spellings (`repr` and fixed notation), plus `WRITE_CASES`, plus the four
edge literals the contract names. The partition is not carried in this prose —
`tests/test_e2e_write_value_fidelity.py` builds that exact set and pins
`PUBLISHED_PARTITION`, so these counts are re-derived from a real engine on every e2e
run (the discipline #770 established for the read side's `live_numbers.PARTITIONS`):

| bucket | count | refused |
|---|---|---|
| exact | 109 | no |
| drifted (low-order only) | 31 | no |
| flattened to `0.0` | 37 | **yes** |
| `NaN` — `0e600` | 1 | **yes** |
| overflowed to `inf` — `1e400` | 1 | no |
| **total** | **179** | **38** |

`1e-400` sits in the flattened bucket and is refused with it, deliberately, although
`0.0` is the correctly-rounded answer there. The load-bearing claim holds over every
row: **nothing the rule accepts is a destroyed value.** Three mechanisms:

| class | literal | `to_float()` |
|---|---|---|
| applied decimal exponent ≤ −309 (#752's class) | `5e-324`, `2.2250738585072014e-308`, `3.2956212316547955e-293` | `0.0` |
| 18-digit mantissa window, leading zeros counted | `0.000000000000000001` (`1e-18` is exact) | `0.0` |
| zero mantissa × overflowed power | `0e600` | `NaN` |

The window is the engine's own: *"If the mantissa has more than 18 digits, ignore the
extras"* (`built_in_strtod`). The remedy is measured, not asserted — the two
mantissa-cap corpus rows come back **exact** as `1.2345678901234567e-3` and
`1.4285714285714284e-4`, where `json.dumps`'s fixed spelling lands 31 and 105 doubles
away.

## Red-proof (real engine, both channels)

Against the base (`181da20d`), `node set` and `game set --value` each reported success
and stored something else:

```
--- headless `node set` (RED — base 181da20d) ---
  5e-324                       -> exit 0, stored 0.0      [SILENTLY CHANGED]
  2.2250738585072014e-308      -> exit 0, stored 0.0      [SILENTLY CHANGED]
  3.2956212316547955e-293      -> exit 0, stored 0.0      [SILENTLY CHANGED]
  1e-320                       -> exit 0, stored 0.0      [SILENTLY CHANGED]
  0.000000000000000001         -> exit 0, stored 0.0      [SILENTLY CHANGED]
  0e600                        -> exit 0, stored None     [SILENTLY CHANGED]   # NaN
  1e-18                        -> exit 0, stored 1e-18    [EXACT]
--- live `game set` through a real daemon (RED — base 181da20d) ---
  ... the same six, exit 0, live value 0.0 / NaN
```

On this branch, all six are `uncoercible_value` (exit 4) headless and
`live_uncoercible_value` (exit 6) live, the target untouched, and `1e-18` still stores
exactly. `tests/test_e2e_write_value_fidelity.py` carries this as a regression in two
tiers: an ENGINE tier that judges gda's rule (written independently in Python) against
what a real Godot did to every corpus row, and a COMMAND tier that drives `node set`,
`project set` and — through a real daemon and a real session — live `game set`.

## Harness version

The predicate must run where the property's DECLARED TYPE is known (`--value 1e-320` on
a String property is an ordinary string and must not be refused), so it lives in the
mirrored coercion block of `ops/operations.gd` and `harness/gda_harness.gd`, byte-identical
(`tests/test_harness_coercion_mirror.py`). **`HARNESS_VERSION` 16 → 17**; the pinned
version/sha in `tests/test_harness_install.py` moved with it and
`scripts/harness_version_guard.py 181da20d <head>` reports a valid transition. The
review round moved both mirrors again, so 17 (never shipped) covers that change too and
only the sha pin moved.

The note the refusal carries belongs to the FLOAT coercion, not to every uncoercible
failure: `_destroyed_float_literal(raw, type)` walks exactly what `_coerce_value` walks
for the declared type, so an `int`, a `Dictionary` and a three-component `Vector2` keep
the plain message they always had.

## Published contracts in this PR

`CONTEXT.md` (Value projection: the two WRITE sides now answer alike from one
principle — a literal the parser reads as `0.0` or `NaN` although no zero was written is
refused, with the overflow boundary named so the glossary, the catalog and
`gda.live_numbers` state the same rule), `docs/command-catalog.md` (a new **Number
coercion** section under `node`, plus the `project set` and live-command
cross-references), `src/gda/skill/SKILL.md` (both the headless paragraph and the LIVE
one, where `game set --value` is outside the wire's pre-daemon refusal and has its own
in-session `live_uncoercible_value`), and the `gda.live_numbers` docstring, which stays
the single authority for the measurement.

One caveat found while measuring and recorded in the catalog because it belongs to
neither half: Godot's scene packer omits a property whose value is not "different" from
its default, comparing floats **approximately, as float32**
(`PropertyUtils::is_property_value_different` → `Math::is_equal_approx`). On
`@export var v: float = 0.0`, `--value 1e-6` echoes `1e-6` and a following `node get`
reads `0.0` — the packer's elision, not the parser's loss, and invisible to any
coercion rule. `project set` is unaffected (measured). The `node set` round-trip
sentence 120 lines above that caveat now points at it instead of claiming a round-trip
without qualification.

## Validation

- `pytest -m "not e2e"` — 2250 passed (the same `test_launch.py` streaming flake
  appeared once under full-suite load and was green on the rerun and on 8 isolated
  runs; this PR touches no launch code)
- `ruff check` / `ruff format --check` — clean; `pyright` via `uv run --frozen` —
  0 errors
- `tests/test_harness_coercion_mirror.py`, `tests/test_harness_install.py` pins,
  `scripts/harness_version_guard.py 181da20d <head>` — pass
- e2e (real Godot 4.6.3): `test_e2e_write_value_fidelity` (7, two of them new),
  `test_e2e_headless_number_reads`, `test_e2e_game`, `test_e2e_node`,
  `test_e2e_project`, `test_e2e_resource`, `test_e2e_harness_install`,
  `test_e2e_daemon` — all green

## Known gap

A number nested inside a Dictionary or Array `--value` reaches the same parser through
`JSON.parse_string` / `str_to_var` without passing this check. Named in the catalog and
in `gda.live_numbers` so the contract is not read wider than it is.


qinhaihong-red:
## Adversarial review — PR #797 at head `920d756f7f`, against base `181da20d`

Reviewed from an independent detached archive at `920d756f7f` (never the `p4-772` worktree). All
probes below ran against a real Godot `4.6.3.stable.official.7d41c59c4`; the live arm ran against a
real `gda-daemon` and a real `Engine session`.

**Verdicts** — Architecture/DDMA: **approve**. Standards: **changes requested** (P2-1).
Artifact/Markdown: **changes requested** (P2-2, P2-3). Spec (#772 ACs): **conditional** (P2-4).

---

### P2-1 — the fidelity note is attached to the generic coercion failure, so it misreports causes it did not diagnose

`_float_fidelity_note(raw)` is appended at all six `_coerce_value` failure sites
(`operations.gd:1781/1830/2986/3621`, `gda_harness.gd:721/749`) and is gated only on the raw text
containing a literal the float parser destroys — never on the failure having been the float refusal,
and never on the declared type routing to `_coerce_float`. Its own comment states the opposite
intent ("… and `""` when it failed for the ordinary reason"). Measured at head:

```
node set --property n(int)     --value 1e-320       -> uncoercible_value: "... to int for property n on node .
                                                       — Godot's own float parser reads 1e-320 as 0.0, so the
                                                       write would store a number you did not send; gda refuses
                                                       it ... Try the same value in scientific notation ..."
node set --property d(Dict)    --value 0.000000000000000001 -> same note, "... to Dictionary ..."
node set --property pos2(Vec2) --value 1,2,1e-320    -> same note; the real failure is the 3-component ARITY
game set --property iv(int)    --value 1e-320       -> live_uncoercible_value, same note (live channel too)
```

Base `181da20d` returns the clean `cannot coerce value 1e-320 to int for property n on node .` for
the first one, so this is a message regression introduced here. Three problems in one: the stated
cause is false (nothing would have been stored — the int coercion refused it at base as well), the
named remedy is unreachable (`1e-320` already *is* scientific notation, and no float spelling helps an
`int`/`Dictionary`/arity failure), and the arity case attributes a structural failure to the parser.
`test_the_refusal_is_narrow_and_names_its_remedy` only covers `--value abc` on a `float`, which is why
it passes. Gate the note on the declared type actually reaching `_coerce_float`, or have
`_coerce_value` report which sub-coercion refused.

### P2-2 — `CONTEXT.md`'s new write-side invariant is falsified by the PR's own documented edge

> The WRITE sides answer alike from one principle — a value the engine's parser cannot carry is
> REFUSED, never stored

Measured counterexample: `1e400`. The parser cannot carry it, and gda **stores** it, exit 0 —
`v = inf` in the `.tscn`, `probe/value=inf` in `project.godot`. `docs/command-catalog.md` and
`gda.live_numbers` both record that exception explicitly; `CONTEXT.md`, the shared-language authority,
states the invariant without it, so the three surfaces disagree. The predicate's actual rule states
cleanly: *a value the parser reads as `0.0` or `NaN` is REFUSED, never stored.*

### P2-3 — the "loud, not silent" argument does not separate the two edges it is used to separate

Catalog, `gda.live_numbers` and the PR body all justify accepting overflow with: `1e400` "reads as
`inf` … the reply reports as JSON `null` … so it is loud rather than the silent substitution this rule
exists to stop." Loudness does not distinguish it. Measured at base `181da20d`:

```
node set --value 0e600  -> exit 0, reply "value": null, scene file `v = nan`
node set --value 1e400  -> exit 0, reply "value": null, scene file `v = inf`
```

Both reply JSON `null`; one is refused, the other accepted. The recorded reason therefore does not
support the asymmetry. It gets weaker on `project set`: `--value 1e400` writes `probe/value=inf` and
every later `project get` reports `null` — a value gda can never report again, which is a thin reading
of "loud". State the real discriminator (a saturated magnitude is still a number in the direction the
caller asked for; `NaN` is not a number at all) or refuse both.

### P2-4 — two disclosed silent-loss paths carry no tracker item, and one falsifies a live catalog claim

Both reproduced at head, both exit 0:

```
node set --property d --value '{"a": 1e-320, "b": 0.000000000000000001}' -> "value": {"a": 0.0, "b": 0.0}
node set --property w --value 1e-6      (@export var w: float = 0.0)
     -> echo "value": 1e-6 ; property ABSENT from the .tscn ; a following node get reads 0.0
```

The first is exactly the failure this PR calls unacceptable, still live on one path. The second is the
packer elision — engine citation verified: `scene/property_utils.cpp:47`
(`!Math::is_equal_approx((float)p_a, (float)p_b)`) called from `scene/resources/packed_scene.cpp:984`
— and it falsifies `docs/command-catalog.md:330`'s unqualified "so a `set` round-trips through a
`get`", which is asserted 120 lines before the caveat that withdraws it. Prose-only disclosure leaves
both unowned: please file an issue for each (the nested-container gap is the one that belongs to this
contract), and make line 330 point at the "Number coercion" caveat.

### P3-1 — an internal issue number inside a user-facing error message

`operations.gd:6302` / `gda_harness.gd:2192` end the refusal with "… the live wire refuses that same
class (#752)". A grep at head shows this is the only `#NNN` inside a message *string* in either `.gd`
file; every other occurrence is a comment. It means nothing outside this repo. Keep it in the block
comment, drop it from the message.

### P3-2 — the published partition does not sum to its stated total and is not re-derivable from any artifact

> 179 literals: 111 exact, 30 changed, 34 flattened to `0.0`, 1 `NaN`; 36 refused by this rule

`111 + 30 + 34 + 1 = 176`. I re-derived the sweep on a real engine over a 179-literal set built the way
the body describes — the 96 corpus rows in both spellings (160 literals; the 32 fixed spellings ≥400
chars excluded exactly as the test does) plus the test's own 15 `WRITE_CASES` plus the 4 literals the
body names (`1e400`, `0.0012345678901234567`, `1.2345678901234567e-3`, `1.4285714285714284e-4`) — and
got **110 exact / 31 changed / 36 zero / 1 NaN / 1 inf; 38 refused**. The set size matches; no count
does. Two things are invisible in the four published numbers either way: the overflow bucket (which is
why they do not sum) and `1e-400`, which sits inside "36 refused" but in none of the four.

This is the one place the #770 discipline was not applied: the READ-side counts are derived from
`PARTITIONS` and pinned verbatim by `tests/test_live_numbers.py`, while these live only in the PR body
with no artifact recording the literal set. Either name the fifth bucket and where `1e-400` sits, or
drop the counts and point at `tests/test_e2e_write_value_fidelity.py`. **The load-bearing claim
reproduces exactly**: over all 179 rows, nothing the rule accepts is a destroyed value.

### P3-3 — `SKILL.md`'s live section omits the live write rule this PR adds

`SKILL.md:183` enumerates what the live channel refuses ("… refused before the request leaves … a usage
error on the argv path and `invalid_params` on `--params-json`, **decided without a running daemon**").
`game set --value` is a STRING and is outside that rule — as this PR's own test docstring says — and now
has a second, harness-side refusal with a different code, a different exit and a different decision
point: measured through a real daemon, `live_uncoercible_value`, exit 6, decided *with* a running
session. Only the headless paragraph at `SKILL.md:135` was updated.

---

### What I verified and did NOT find wrong

- **Mirror discipline.** Both shared blocks are raw byte-identical (18605 bytes each, not merely
  tab-normalized); `HARNESS_VERSION` 16→17; `PINNED_HARNESS_SHA256` matches `shasum -a 256` on the
  file; `scripts/harness_version_guard.py 181da20d 920d756f7f` → "harness version transition is valid".
  The daemon reported `harness_version: 17` on a real install.
- **Engine citations, against `../godot`.** `String::to_float()` at `core/string/ustring.cpp:2677` is
  `built_in_strtod<char32_t>(get_data())`; `core/io/json.cpp:393` is `String::to_float(&p_str[index],
  &rptr)` — one function, two callers, as claimed. The 18-digit window (`ustring.cpp:2460`, `2471`)
  counts every mantissa digit *including leading zeros* (`mantSize > 18 → fracExp = decPt - 18`), so
  `0.000000000000000001` spends the whole budget on zeros — accurate. The −309 cliff is `dblExp`
  overflowing to `inf` then `fraction /= dblExp`; `0e600` is `0.0 * inf` after `exp > maxExponent` (511).
- **The "only GDScript can compute it" claim is true as stated.**
  `godot_applied_decimal_exponent("0.000000000000000001")` returns `-17` and `("0e600")` returns `600`;
  neither reaches `wire_flattens_to_zero`'s `≤ -309` gate, so the shipped Python model sees 1 of the 3
  mechanisms. The decisive reason is the stronger one the docstring gives first — the declared type —
  and it holds: `node set --property s(String) --value 1e-320` succeeds and stores the string.
- **Red-proof reproduced in my own archive at base `181da20d`**: `5e-324`, `2.2250738585072014e-308`,
  `3.2956212316547955e-293`, `1e-320`, `0.000000000000000001` all exit 0 and write `v = 0.0`; `0e600`
  exits 0 and writes `v = nan`; `1e-18` stores exactly. At head all six are `uncoercible_value` (exit 4)
  with the scene byte-unchanged.
- **Live arm, real daemon + real session**: the four literals return `live_uncoercible_value`, the
  running node still reads `1.5` after each, `1e-18` writes exactly. `project set` shares the refusal;
  `project set --value 1e-6` is unaffected by the packer elision, as the catalog says (measured).
- **The refused class is not only "tiny"**: `--value 0000000000000000000001` — the integer 1 with 21
  leading zeros — is refused, and the engine really does read it as `0.0`. A nice case the base
  corrupted silently.
- **Numeric prose is exact**: fixed spellings land 31 and 105 doubles away, both scientific spellings
  are bit-exact, the ordinary band is 1 ULP; the `1e-6` echo is literally `1e-6` in the raw JSON.
- **Reconcile**: ADR-0041's #752 outcome note points at `gda.live_numbers` as the decided contract and
  states nothing this PR falsifies, so no ADR amendment is owed. The `fix` type matches the #752
  precedent (`a8725603`).
- **Gates** (from the archive): `pytest -m "not e2e"` → 2250 passed, no flake; `ruff check` and
  `ruff format --check` clean; `tests/test_e2e_write_value_fidelity.py -m e2e` → 5 passed. `pyright`
  could not be judged from a detached archive (no venv there; 244 cascading unresolved-import errors) —
  environment evidence, not evidence against the PR.

The decision itself is right, the observe-vs-predict asymmetry is argued honestly, and the deliberate
over-refusal of `1e-400` is recorded rather than hidden. P2-1 is the only code change I would gate the
merge on; P2-2 and P2-3 are one clause each.

qinhaihong-red:
## Review handled — head `729d2f42`

> All the substance is in `3daaecaf`; `729d2f42` on top of it is a whitespace-only
> reflow of two paragraphs this round rewrapped mid-sentence. Every SHA cited below is
> `3daaecaf`, where the behaviour was measured.

Every finding was re-verified on `920d756f` before anything changed. Six adopted, one
partially. Nothing in "What I verified and did NOT find wrong" was touched.

| # | Decision | Evidence (re-verified at `920d756f`) | Change | Validation |
|---|---|---|---|---|
| P2-1 | **Fully adopt** | Reproduced all four shapes: `int`/`1e-320`, `Dictionary`/`0.000000000000000001`, `Vector2`/`1,2,1e-320`, live `game set` on an int — each carried the note | `_destroyed_float_literal(raw, type)`; every call site passes the declared type | 2 new red-proofed e2e regressions; `test_e2e_write_value_fidelity` 7 passed |
| P2-2 | **Fully adopt** | `node set --value 1e400` → exit 0, `v = inf`; `project set` → `probe/value=inf` | `CONTEXT.md` states the rule the predicate actually implements, with the overflow boundary | catalog / `live_numbers` / `CONTEXT.md` now state one rule |
| P2-3 | **Fully adopt** | Confirmed: a stored `NaN` and a stored `inf` both read back as JSON `null` | Rationale rewritten around what is STORED, in the catalog, `live_numbers` and the PR body | — |
| P2-4 | **Partially adopt** | Both paths reproduced at head (below) | catalog:330 qualified **in this PR**; ONE issue drafted below, not filed | `node get` measured for both |
| P3-1 | **Fully adopt** | `#752` was the only `#NNN` inside a message string | Dropped from the message; the block comment keeps it | measured message text |
| P3-2 | **Partially adopt** | `111+30+34+1 = 176 ≠ 179` confirmed | Partition pinned in the e2e module and re-derived from a real engine; PR body quotes it | red-proofed against the old numbers |
| P3-3 | **Fully adopt** | `game set --value` is outside the wire rule; measured `live_uncoercible_value`, exit 6 | `SKILL.md`'s LIVE paragraph now carries the in-session refusal | — |

---

### P2-1 — the note now belongs to the refusal it diagnosed

You offered two routes. Gating on the declared type alone is not sufficient: it fixes
`int` and `Dictionary` but leaves the arity case you measured, since a three-component
`Vector2` *does* route to `_coerce_float`. So the gate is the type **and** the walk:

```gdscript
func _destroyed_float_literal(raw: String, type: int) -> String:
	var components: PackedStringArray
	match type:
		TYPE_FLOAT:   components = PackedStringArray([raw])
		TYPE_VECTOR2: components = raw.split(","); if components.size() != 2: return ""
		TYPE_COLOR:   ...  # hex form parses no float; 3 or 4 components or nothing
		_:            return ""
	for part in components:
		var literal := part.strip_edges()
		if not literal.is_valid_float(): return ""   # the ordinary failure, first
		if _float_literal_is_destroyed(literal): return literal
	return ""
```

It walks what `_coerce_value` walks, in the same order and behind the same gates, and
stops where `_coerce_float_list` stops. Measured at `3daaecaf` on Godot 4.6.3:

```
n(int)     <- 1e-320       cannot coerce value 1e-320 to int for property n on node .
d(Dict)    <- 1e-18-fixed  cannot coerce value 0.000000000000000001 to Dictionary for property d on node .
pos2(Vec2) <- 1,2,1e-320   cannot coerce value 1,2,1e-320 to Vector2 for property pos2 on node .
pos2(Vec2) <- abc,1e-320   cannot coerce value abc,1e-320 to Vector2 for property pos2 on node .
s(String)  <- 1e-320       exit 0, stores "1e-320"
v(float)   <- 1e-320       ... — Godot's own float parser reads 1e-320 as 0.0 ...   [kept]
pos2(Vec2) <- 1,1e-320     ... — Godot's own float parser reads 1e-320 as 0.0 ...   [kept, names the component]
game set n(int) <- 1e-320  cannot coerce value 1e-320 to int for script variable n on node /root/Main
```

`test_the_note_explains_only_the_refusal_it_diagnosed` (headless) and an added arm in
`test_game_set_refuses_the_same_literals_against_a_real_daemon` (live) pin all of it
with **whole-message equality**, not a substring. Red-proofed: with only the two `.gd`
files reverted to `920d756f`, both fail with the note appended.

Both mirrors moved together, so `HARNESS_VERSION` stays **17** (never shipped) and only
the sha pin moved — `scripts/harness_version_guard.py 181da20d 3daaecaf` → *harness
version transition is valid*; the byte-identity test passes because both copies moved.

### P2-2 — the invariant now states the rule the predicate implements

`1e400` is stored, so "cannot carry ⇒ refused" was false. `CONTEXT.md` now reads *a
literal the engine's parser reads as `0.0` or `NaN` although no zero was written is
REFUSED, never stored*, and names the overflow in the same sentence, so the glossary,
`docs/command-catalog.md` and `gda.live_numbers` state one rule with one boundary.

### P2-3 — the discriminator is what was STORED

You are right that loudness does not separate them, and the `project get` observation is
right too: `probe/value=inf` reads back as `null` forever. Rewritten in all three places
(catalog, `live_numbers`, PR body):

- `1e400` → `inf`: the correctly-rounded IEEE-754 answer for a magnitude past
  binary64's top — the engine's number for "larger than it can hold", in the direction
  the caller asked for, not a different number put in its place.
- `0e600` → `NaN`: not the value, not near it, not in its direction. That is the
  substitution the rule exists to stop.

The text now says explicitly that a stored `NaN` reports as `null` exactly as a stored
`inf` does, and discloses the overflow's own residual — stored but never reportable.

### P3-2 — the partition is now derived, not carried

Re-derived on a real engine over the exact 179-literal set the body describes. My
numbers agree with yours except for one row:

| bucket | count | refused |
|---|---|---|
| exact | 109 | no |
| drifted (low-order only) | 31 | no |
| flattened to `0.0` | 37 | **yes** |
| `NaN` — `0e600` | 1 | **yes** |
| overflowed to `inf` — `1e400` | 1 | no |
| **total** | **179** | **38** |

The single-row difference from your 110/31/36/1/1 is `1e-400`: `0.0` **is** its
correctly-rounded value, so calling it *exact* is defensible — but it is REFUSED, so
this table puts it in the bucket it is refused with, and says so. Refused is now
exactly `flattened + NaN`, which is the whole rule.

These are no longer prose. `tests/test_e2e_write_value_fidelity.py` builds that set
(`_corpus_spellings()` + `WRITE_CASES` + `NAMED_EDGE_LITERALS`) and pins
`PUBLISHED_PARTITION` / `PUBLISHED_REFUSED`, asserting `nan == ["0e600"]`,
`inf == ["1e400"]` and `"1e-400" in zeroed` — the #770 discipline, applied here.
Red-proof: pinned to the old published numbers the test fails on `drifted`. The
load-bearing claim is asserted over every published row and stays prominent in the body.

### P3-3 — `SKILL.md`'s live section

Added after the wire-refusal sentence: `game set --value` is outside that rule (the
value travels as a STRING), and refuses in-session with `live_uncoercible_value`
(exit 6, the running game untouched), following the same rule as headless `--value`.

---

## P2-4 — catalog:330 fixed here; ONE issue drafted, not filed

Line 330 now reads *"…so a `set` round-trips through a `get` — with one measured
exception, the scene packer's elision, recorded under "Number coercion" below…"*. Only
the packer elision qualifies it: for the accepted overflow the echo and the `get` agree
(both `null`), so that one is not a round-trip break.

I drafted the two silent-loss paths as **one** issue rather than two, because any answer
has to place a single boundary — *which silent float losses does the `--value` write
contract own?* — and answering one path without the other leaves that boundary
undefined. Split it if you disagree; the two halves are separable.

<details>
<summary>Issue draft (not filed)</summary>

**Title:** `--value` writes still store a number the caller did not send, on two paths #772's rule cannot see

**Body:**

#772 refuses a `--value` literal Godot's float parser destroys, but the rule sits in
`_coerce_float`, so it sees only a literal that reaches that function. Two paths reach
storage without it. Both measured at `3daaecaf` on Godot 4.6.3, both exit 0.

**Path A — numbers nested in a Dictionary or Array `--value`.** A container value is
parsed by `JSON.parse_string` / `str_to_var`, which reach `built_in_strtod` for each
number without passing through `_coerce_float`:

```
gda node set main.tscn --node . --property d --value '{"a": 1e-320, "b": 0.000000000000000001}'
  -> exit 0, reply "value": {"a": 0.0, "b": 0.0}, .tscn holds d = { "a": 0.0, "b": 0.0 }
```

This is exactly the failure #772 calls unacceptable, still live on one path. It is
disclosed in `docs/command-catalog.md` and `gda.live_numbers` but owned by nobody. Note
the echo here is HONEST — the reply already shows `0.0` — so the question is whether
disclosure plus an honest echo is the contract, or whether the refusal should reach
nested numbers as it reaches scalars.

**Path B — the scene packer's elision.** Godot omits a property whose value is not
"different" from its declared default, comparing floats approximately, as **float32**
(`PropertyUtils::is_property_value_different` → `Math::is_equal_approx`,
`scene/property_utils.cpp:47`, from `scene/resources/packed_scene.cpp:984`):

```
# @export var w: float = 0.0
gda node set main.tscn --node . --property w --value 1e-6
  -> exit 0, reply "value": 1e-6 ; property ABSENT from the .tscn
gda node get main.tscn --node .
  -> "w": 0.0
```

Here the echo is DISHONEST: it reports `1e-6` for a value the file does not hold. This
is the path that qualifies the `node set` round-trip claim, and `project set` is
unaffected (measured) — it is the scene packer's, not the parser's.

**Why one issue.** Both are "a `--value` write stores something the caller did not
send", and both are outside `_coerce_float`. Any answer places one boundary: does the
write contract promise the STORED value, or only an honest echo of what the engine did?
Path A can be closed inside the coercion (walk the parsed container and apply the same
predicate); path B cannot — nothing at the coercion site can see the packer — so it is
either a post-save read-back verification, a refusal keyed on the declared default, or a
documented limit. Deciding them apart risks two different answers to one question.

**Acceptance criteria (proposed).**
- A decision recorded for each path: refuse, verify after save, or document as a limit.
- Whatever is chosen, an e2e regression in `tests/test_e2e_write_value_fidelity.py`
  covering both, driven by a real engine.
- `docs/command-catalog.md`, `gda.live_numbers` and `CONTEXT.md` state the resulting
  boundary identically.

</details>

---

## Validation at `3daaecaf`

- `pytest -m "not e2e"` — **2250 passed**. The `test_launch.py` streaming test failed
  once under full-suite load and was green on the full rerun and on 8 isolated runs; it
  is the flake the body already records, and this PR touches no launch code.
- `ruff check` / `ruff format --check` — clean. `pyright` via `uv run --frozen` —
  **0 errors** (a bare `.venv/bin/pyright` resolves a different environment and reports
  cascading import errors; that is the environment, not the branch).
- `tests/test_harness_coercion_mirror.py` (byte-identity), `tests/test_harness_install.py`
  (version + sha pins), `scripts/harness_version_guard.py 181da20d 3daaecaf` — pass.
- e2e, real Godot 4.6.3: `test_e2e_write_value_fidelity` **7 passed** (two new),
  `test_e2e_headless_number_reads` + `test_e2e_game` (20 with the above),
  `test_e2e_node` + `test_e2e_project` + `test_e2e_resource` **162 passed**,
  `test_e2e_harness_install` + `test_e2e_daemon` **32 passed**.

**Residual risk.** `PUBLISHED_PARTITION` is engine-pinned: a Godot whose `built_in_strtod`
changes fails that test loudly, which is intended but is a new maintenance point tied to
the corpus. And the two paths in the draft above stay open in this PR — disclosed in the
catalog and `gda.live_numbers`, not closed.

qinhaihong-red:
## Full independent adversarial review — PR #797

Reviewed from first principles at pinned head `729d2f42a81b4e9ac363fa19e160f7270c84b195` against `feat/gda-dogfooding-dev@181da20d2ea418977600bd73d2958dc4308e8de6`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Standards | **Pass** |
| Spec (#772) | **Pass** |

### Artifact review — required changes

1. `docs/command-catalog.md:437-438` assigns exit 4 to both `uncoercible_value` and `live_uncoercible_value`. The registry defines the headless operation code at exit 4 and the live code at exit 6. State “exit 4 / 6 respectively”; the current prose can make shell and agent branching incorrect.

2. The central admission rule is grammatically and semantically inconsistent around zero. `CONTEXT.md:203-204`, `docs/command-catalog.md:437`, and `src/gda/skill/SKILL.md:135` say the parser result is refused when it becomes `0.0` or `NaN` “although/when no zero was written.” Yet the documented `0e600` case denotes zero and must be refused because the parser returns `NaN`. State the real two-clause rule everywhere: refuse `0.0` when the literal does not denote zero; refuse any `NaN` substitution.

Pass. The headless and harness coercion blocks are physically mirrored because neither runtime can load one shared module under the established topology. `tests/test_harness_coercion_mirror.py` compares the complete shared block byte-for-byte, so this is a guarded derived copy rather than two independent authorities.

### Standards

Pass. The apparent duplication is explicitly permitted by that runtime boundary and protected by the exact parity guard; it is therefore not an actionable Duplicated Code smell here. No other non-tool-enforced violation was found.

### Spec

Pass. #772's explicit refusal decision, mirrored agreement by construction, shared number corpus, and real-engine headless/live exact-or-refusal regressions are present.

Sources: `CONTEXT.md`'s number/value language; issue #772; command catalog; `gda.live_numbers`; bundled Skill; mirrored coercion comments; error/exit registries. Writing reference: ASD-STE100. The complete PR/issue prose was checked for parser/coercion terminology, prediction versus observation, exact/drifted/destroyed values, underflow/overflow/NaN boundaries, channel distinctions, and remedies. The two public-contract wording errors above are substantive.

Validation: `git diff --check` passed. Current hosted CI has no failed or pending required check; Godot e2e is skipped. The full 179-literal Godot corpus was inspected but not rerun.

qinhaihong-red:

New head: `c4d072470` (one docs commit on top of the reviewed `729d2f42`).

| Catalog assigns exit 4 to both `uncoercible_value` and `live_uncoercible_value`; the registry gives 4 (operation) and 6 (live) | **Fully adopt** | `error_codes.py`: `uncoercible_value` → `EXIT_OPERATION`, `live_uncoercible_value` → `EXIT_LIVE` | Catalog now reads "`uncoercible_value`, exit 4, headless; `live_uncoercible_value`, exit 6, live" | Catalog diff; 271 focused tests green |
| The admission rule is stated as "`0.0` or `NaN` when no zero was written", which mispredicts the documented `0e600` case (denotes zero, refused because the parser substitutes `NaN`) | **Fully adopt** | `live_numbers.py:129` already documents the correct two-clause rule; the prose surfaces collapsed it | All FIVE prose copies corrected to the two-clause rule (refuse `0.0` when the literal does not denote zero; refuse any `NaN` substitution): `CONTEXT.md`, catalog's coercion paragraph (now naming `0e600` as a second-clause refusal) and its cross-reference at the live section, and both SKILL sentences (headless `uncoercible_value` and live `live_uncoercible_value`) — two copies beyond the three the review cited, found by sweeping the phrasing | Grep sweep: no collapsed copy remains; code untouched (the implementation already follows the two-clause rule) |

qinhaihong-red:

Reviewed at `c4d07247051ae8565c2a10c6b225cce463faa790` against base `181da20d2ea418977600bd73d2958dc4308e8de6`.

**Verdict: Changes required**

- Artifact: **Changes required**
- DDMA: Pass
- Standards: Pass
- Spec: Pass

The previous exit-code error and the main two-clause number-admission wording are fixed. One contradictory contract copy remains.

### Required — the `project set` recap still treats valid zero literals as refused

`docs/command-catalog.md:831-833` says the parser “refuses the literals it would turn into `0.0` or `NaN`.” Taken literally, that includes valid zero literals. It contradicts the corrected rule at `docs/command-catalog.md:432-444` and `1070-1073`: a `0.0` result is refused only when the literal does not denote zero, while any parser substitution to `NaN` is refused.

**Smallest fix:** change the recap to say that it refuses literals the parser turns into `0.0` **when they do not denote zero**, or into `NaN`.

The implementation and the other reviewed contract copies remain consistent. The exact-head diff passes `git diff --check`; current required CI has no failed or pending checks.

qinhaihong-red:
## Round-3 response

New head: `c4d072470` → `56c9cdd70` (one docs commit).

| The `project set` recap still says the parser "refuses the literals it would turn into `0.0` or `NaN`" — taken literally it covers valid zero literals, contradicting the corrected two-clause copies | **Fully adopt** | Verified at the cited lines; this copy escaped the round-2 sweep because it carries no "no zero" phrasing to match | The recap now reads "refuses the literals it would turn into `0.0` when they do not denote zero, or into `NaN`" | Sweep for the collapsed `` `0.0` or `NaN` `` pattern across docs/src/CONTEXT/READMEs: no residual copy; CI green on the new head |

The implementation and all other contract copies stand as reviewed.

PR #798, opened 2026-08-31
feat(gda): human-mode failure rendering as real lines (#685)

### LINKED ISSUE #685: Human-mode failure rendering: lay out diagnostics as real lines instead of a single escaped JSON envelope

qinhaihong-red:
Closes #685

> **Stacked PR.** The base is `feat/687-typed-failure-evidence` (P4 slot 2), not
> `main`, so the diff shown is only this slice's. Merge slots 1 and 2 first.

## What was actually wrong

The issue's pre-dispatch review had it right, and it is stronger than "the layout is
poor": there was **no human failure channel at all**. `emit_failure` serialized the
envelope with `model_dump_json` and exited — it took no channel parameter and had no
branch on one — so a caller who never asked for JSON got it anyway, with
`\n`-escaped `diagnostics`. `script run --strict` is where that costs: its labelled
script stdout/stderr is the evidence a CI operator reads, delivered as one
unreadable line.

So the shape of the work was: give the emitter the channel, then render.

## What ships

**One renderer, `gda.render.render_failure`.** Nothing in it keys on a
`Gda error code`, so no command can grow a private failure layout, and it is
**total over the envelope** — the text replaces a JSON line that carried every key,
so dropping one would say less than what it replaced:

```
error: script_failed (operation)
script run --strict: res://suite.gd exited with status 3
evidence:
  exit status: 3
  script errors:
    push_error: res://suite.gd:6: assertion failed: expected 3, got 2

--- script stdout ---
Godot Engine v4.6.3.stable.official.7d41c59c4 - https://godotengine.org

running 3 tests
1 of 3 tests failed
--- script stderr ---
ERROR: assertion failed: expected 3, got 2
   at: push_error (core/variant/variant_utility.cpp:1024)
   GDScript backtrace (most recent call first):
       [0] _init (res://suite.gd:6)
```

Two layout rules, both stated in the renderer's docstring:

- **Short before long.** The fixed-size parts (head, message, `probe` #667, `hint`
  #670, `evidence` #687) stay together under the head line; `diagnostics` goes last
  because it is the only unbounded part — two 16 KiB tail-capped captures on a
  timeout — so it can never push the verdict off a terminal.
- **Absent parts render nothing.** No empty section, no blank tail, no trailing
  newline for `typer.echo` to double. `evidence` is treated as absent-by-default at
  BOTH levels the #687 contract requires: the key, and every field inside it — an
  `evidence` object carrying no field prints no header.

The `evidence` block follows the **model's own field declaration order**, so the two
channels enumerate the evidence identically and neither has to be kept in step by
hand. The clock is rendered at the two decimals the `launch_timeout` message already
states it in; full precision stays in JSON.

**The channel reaches all six call sites explicitly, and each is pinned.**
`emit_failure` now takes a REQUIRED keyword-only `json_output` — a new call site
cannot default its way back into the wrong channel:

| Call site | How it answers | Human-channel test |
| --- | --- | --- |
| `dispatch.py:149` `_resolve_project_or_fail` (via `_project_context`) | the flag the dispatch tail already holds | `..._on_the_dispatch_tail` |
| `dispatch.py:258` `dispatch_recipe` | same | `test_a_recipe_failure_is_refused_in_lines_too` |
| `headless.py:545` `--params-json` conflict | `json_in_effect(ctx)` | `test_the_params_json_conflict_is_refused_in_lines` |
| `headless.py:564` `--params-json` invalid JSON | same | `test_an_invalid_params_json_object_...` |
| `headless.py:788` `HeadlessCommand.run` | threaded from `emit`, for the same reason `emit` takes it: emitting is a public-channel act | `test_a_human_invocation_gets_the_lines_...` |
| `hints.py:257` `_answer` | `json_in_effect(ctx)` — the usage refusal is rendered by this same renderer, not by click's usage error | `test_a_usage_refusal_without_json_...` |

The count was five in the first body because row 3 covered two sites, and the
anchor note it came from missed `HeadlessCommand.run` (it called the private
`_fail` alias at base, invisible to the `grep`). Exhaustiveness is the evidence
offered for "a new call site cannot default its way back", so the count is now
right and each row carries the test that reds when that site reverts to
always-JSON.

**Prefactoring commit (`69985064`).** `_json_in_effect`, `RAW_ARGV_META_KEY` and the
record's writer moved from `gda.hints` to `gda.headless` as public
`json_in_effect` / `remember_argv`, beside `ancestor_json` — the reading they build
on. Reason, per the issue's 2026-08-31 anchor note: `gda.hints` imports
`gda.headless` for the failure channel, so a channel question owned by `hints` would
need that import to run backwards. No behaviour change; the three readings, their
order and every call site are identical.

## `--json` is byte-identical — the scope-defining property

Measured three ways, not asserted:

- **Registry-wide, through the real emit path.**
  `test_the_json_channel_stays_the_model_dump_for_every_registered_code` builds a
  failure for **every** registered code with all three optional keys populated and
  requires `emit_failure(..., json_output=True)` to print exactly
  `model_dump_json(exclude_none=True) + "\n"`.
- **Base-vs-head CLI byte compare** (base tree extracted with `git archive`, same
  interpreter): `binary_not_found`, `unknown_command` + hint, `unknown_option`,
  `project_not_found`, `invalid_params`, `usage_error`, all three `--json`
  spellings, and a REAL-engine `script run --strict` `script_failed` carrying
  `evidence` — identical bytes and identical exit codes in all nine.
- **`gda schema` and `gda --help`**: byte-identical to the base (938,752 bytes;
  `--help`, `scene --help`, `script run --help`, `info --help`,
  `resource import --help`). No help/schema surface diff — AC 4 answered.

## Verification

- **Red-proofs**, 8 mutations, each hitting exactly the guard it should and nothing
  else: always-JSON (the defect itself) → the human CLI guard; empty `evidence`
  prints a header → the compactness guard; `probe`/`hint` dropped → the totality
  guard; the human layout leaking into `--json` → both JSON guards; blank
  `diagnostics` opening a section → 4 layout guards; `diagnostics` re-escaped into
  one line → the motivating guard; an evidence label renamed → the evidence guard;
  the head line losing its category → 4 guards. All 9 new tests are covered.
- **Real engine** (Godot 4.6.3, local), both arms: a `--strict` `quit(3)` with a
  `push_error` → the block above, gda exiting 4; a wedged `resource import
  --timeout 4` (blocking editor plugin) → `error: launch_timeout (environment)` with
  `elapsed: 4.02s` / `timeout: 4.0s` / `termination phase: output_seen` and the
  engine's own captured stdout as real lines, exit 124.
- **Gates**: `pytest -m "not e2e"` 2277 passed; `ruff check` / `ruff format --check`;
  `uv run --frozen pyright` 0 errors. **Full e2e tier run locally**: 591 passed, 1
  skipped (the Linux-only templates arm) — every file, including the daemon/live and
  MCP suites.

## The tests this changes, and why

34 existing tests asserted the error ENVELOPE on an invocation that never asked for
it — the defect at test level. They now pass `--json`
(`test_scene_errors` 12, `test_params_json` 11, `test_info_errors` 8,
`test_e2e_info` 1, plus two `HeadlessCommand.emit` unit tests that passed
`json_output=False` and then parsed JSON). `test_no_renderer_is_orphaned` gains
`render_failure` as a known helper: a descriptor's `render` takes that command's own
RESULT model, while the error half is one schema for every command (ADR-0004), so its
rendering is bound to no descriptor by construction.

## Two decisions worth naming

- **Stream unchanged.** The human failure goes to **stdout**, where the envelope
  went. Moving it to stderr is a separate contract change, not a layout one.
- **Recognized script errors can print twice** — on the two `script run` verdicts
  whose `diagnostics` prose already renders them (`_ended_run_diagnostics`).
  Accepted rather than special-cased: for the four OTHER codes carrying
  `evidence.script_errors` the diagnostics is raw engine stderr, where the curated
  list is the summary that makes the dump readable — and suppressing it per code is
  exactly the per-command layout this renderer exists to prevent.

## Synced artifacts

`CONTEXT.md` (`Error envelope` now names both renderings, the one exit code they
share, and the one renderer that keys on no code) and `src/gda/skill/SKILL.md`
("Always pass `--json`" now states the consequence: it selects the channel for
failures too). README needs no change — its `--json` row already promised "without
it, commands print a concise human-readable rendering", a promise failures did not
keep until this slice, so no translation re-stamp.


qinhaihong-red:
## Adversarial review — PR #798 @ `97feae2ab6d0b1a9c4747b217f62f48353a695d3`

Reviewed against the stacked base `7f5c943bab` (`feat/687-typed-failure-evidence`); the chain beneath is not under review. Four axes (Artifact/Markdown, Architecture/DDMA, Standards, Spec), independent detached worktrees, real Godot 4.6.3 probes with a per-run `GDA_USER_DATA_ROOT`, plus source-level mutation testing against the full `-m "not e2e"` suite.

**Verdicts** — Artifact/Markdown: changes required. Architecture/DDMA: changes required. Standards: changes required (gates green; blocked on artifact sync). Spec: behaviour meets all 4 ACs and was reproduced on a real engine; AC3's *test-pinning* half is not met.

The scope-defining property holds. I re-derived `--json` byte-identity independently on **17 cases the PR did not pick** — `--params-json` (invalid JSON, arg conflict, `{}`, stdin `-`), live-channel with the daemon absent (`game tree`, `game get`, `daemon status`, `daemon stop`), usage errors (no args, unknown group, unknown command, unknown option), `project_not_found`, a curated near-miss carrying `hint`, `--schema`, and the trailing `--json` spelling. stdout bytes, stderr bytes and exit codes identical in all 17. The "no code branching" claim also survives construction: rendering **every one of the 92 registered codes** with identical other fields and normalising the code/category tokens yields one canonical shape, `error: <CODE> (<CAT>)\nM\n\nD`, with zero deviations.

---

### P1 — The "one renderer, keys on no `Gda error code`" invariant is false for the entire `usage` category, and `CONTEXT.md` now states it unconditionally

`CONTEXT.md:235-240` asserts: *"without that flag the same failure is RENDERED for a human instead — the code and its category on a head line, the message, each optional typed key as a labelled line … Two renderings of ONE outcome, at one exit code, from one renderer that keys on no `Gda error code`."*

`hints.py:241-247` (`_answer`) calls `emit_failure` **only** inside `if json_in_effect(ctx)`; otherwise it raises click's `UsageError`. `unknown_command` and `unknown_option` are the only two `usage`-category codes, so that whole category has a second, private human layout:

```
$ gda scene inspect            # exit 2, STDERR
Usage: gda scene [OPTIONS] COMMAND [ARGS]...
╭─ Error ─────────────────────────────────────────────╮
│ `gda scene inspect` is not a gda command. …         │
$ gda scene inspect --json     # exit 2, STDOUT
{"error":{"category":"usage","code":"unknown_command",…,"hint":"gda scene get"}}
```

No head line, no code, no category. The glossary also contradicts its own neighbour: `CONTEXT.md` `Near-miss hint` already says *"the human error carries the same correction in its message"*.

**Consequent dead code:** `render.py`'s `hint:` branch is unreachable in production. `GdaError.hint` is set in exactly one place (`hints.py:194` `Refusal.failure()`), and that value only ever flows into `emit_failure(..., json_output=True)`. The PR justified totality with *"a human failure that quietly dropped one would say less than what it replaced"* and produced a dead branch for the one key that can never arrive.

**Fix (either):** scope the `CONTEXT.md` sentence and `render.py`'s docstring to "every code that reaches the failure channel — the `usage` refusals answer through click's usage error", and drop the `hint:` branch; **or** route the human near-miss through `emit_failure(..., json_output=False)` (the sentence and exit 2 are already identical) and make the invariant true.

### P1 — AC3's second half ("a test pins the human layout") is not met: two mutation proofs

Source-level mutations, each run against the full `-m "not e2e"` suite (2277 tests) in a disposable worktree at this head:

| Mutation | Result |
| --- | --- |
| Move `diagnostics` **before** `evidence` — violates the headline "short before long" rule | **2277 passed, 0 failures** |
| `emit_failure` always emits JSON — **the exact defect #685 exists to fix** | **1 failed** (`test_a_human_invocation_gets_the_lines_and_never_the_envelope`) |
| Swap `probe`/`hint` order | 1 failed |
| Empty `evidence` renders a bare header | 1 failed |
| Trailing newline added | 5 failed |
| Head line loses its category | 4 failed |
| `evidence` block dropped entirely | 2 failed |

Two things follow. First, the PR's **primary stated layout rule is unpinned** — I confirmed the mutated renderer really does bury the verdict, printing `diagnostics` then `evidence:`. "Absent parts render nothing" *is* well pinned; "short before long" is not pinned at all. Second, **the entire feature's CLI-level behaviour rests on one test**: 4 of the 6 `emit_failure` call sites (`dispatch.py:149`, `dispatch.py:258`, `headless.py:547`, `headless.py:566`) have zero human-channel coverage — reverting them to always-JSON is invisible to the suite.

The PR's "8 red-proofs" are real and each hit the guard claimed, but the chosen mutation space omitted the order swap and the per-call-site reverts.

**Fix:** one ordering assertion (a failure carrying `evidence` *and* `diagnostics`, asserting `evidence:` precedes the diagnostics block), and one human-channel CLI test per uncovered call site.

---

### P2 — "one authority … read the same way by both channels" is inverted, and unguarded

`_evidence_lines` is a hand-written five-branch mirror of `FailureEvidence`. The JSON side is model-driven (`model_dump_json`); the human side is not. A sixth field ships on `--json` and is **silently dropped** from human output with **no test failing** — the same holds for a fourth optional `GdaError` key (there have been three in three consecutive slices: `probe` #667, `hint` #670, `evidence` #687). The claim appears in three artifacts (PR body, `render.py` docstring, `tests/test_human_failure_output.py:117-119`).

A reachable instance already exists. `script_errors` uses a truthiness test while the other four use `is not None`, so an empty list diverges. Verified live on a real timeout:

```
JSON : "evidence":{"elapsed_seconds":5.006…,"timeout_seconds":5.0,"termination_phase":"output_seen","script_errors":[]}
HUMAN: evidence:
         elapsed: 5.05s
         timeout: 5.0s
         termination phase: output_seen        <- no script-errors line
```

Here the omission is the better output, but it falsifies the stated invariant. **Fix:** assert `set(FailureEvidence.model_fields)` equals the set the renderer reads (the repo already uses this idiom at `test_render.py:92`), and state the truthiness choice as rationale.

### P2 — Undisclosed second duplication: `diagnostics` now prints twice on a terminal

`HeadlessCommand.execute` forwards the run's stderr to **stderr**; `render_failure` then prints the same bytes inside `diagnostics` on **stdout**. On the most common failure in the product:

```
STDOUT: error: binary_not_found (environment)
        Godot binary could not be launched: /nonexistent/Godot
        (blank)
        gda: Godot binary could not be launched: /nonexistent/Godot ([Errno 2] …)
STDERR: gda: Godot binary could not be launched: /nonexistent/Godot ([Errno 2] …)
```

Three near-identical sentences on an interleaved terminal. Pre-existing in substance (the base's JSON line carried the same bytes escaped), but this slice makes it adjacent human text, and "Two decisions worth naming" names only the script-errors duplication. The motivating case is unaffected — `script run --strict` captures rather than forwards. **Fix:** disclose it, or drop the stderr tee on the failure path now that `emit_failure` renders it.

### P2 — README family still documents the pre-PR behaviour (DoD-4 / state sync)

`README.md:70` promises unconditionally that a failure is *"mapped to a stable non-zero exit code plus a structured `{"error": {…}}` envelope"*; `:438-439` and `:483` say the same. After this PR that holds only under `--json`. `CONTEXT.md` and `SKILL.md` were synced; the human-facing onboarding document was not, nor its three translations (`docs/README.zh-CN.md:71`, `.es.md:73`, `.ja.md:71`).

Separately, the body's README justification is retro-fitted. `README.md:649` reads *"Emit the result as a single JSON object on stdout. **Without it, commands print a concise human-readable rendering.**"* — the second sentence's antecedent is *"the result"*; the row describes the success channel and never promised anything about failures. (The body's "no translation re-stamp" claim is **accurate as written**: the three translations carry a `sha256` of `README.md` guarded by `tests/test_readme_i18n_sync.py`, and it matches at this head — so a README edit costs 4 files plus `scripts/update_readme_i18n.py`.)

### P2 — The body says "five call sites"; there are six

`grep -n "emit_failure(" src/` at this head returns `headless.py:545`, `:564`, `:781`, `dispatch.py:149`, `:258`, `hints.py:246`. The table has five rows because row 3 covers two. The count is inherited from the issue's anchor note, which missed `HeadlessCommand.run` — it called the private `_fail` alias at base, invisible to that grep. Exhaustiveness is the evidence offered for *"a new call site cannot default its way back into the wrong channel"*, so the count should be right.

### P2 — Stale rationale in the hoisted docstring

`headless.py:161-166` justifies the hoist *"because `emit_failure` in this module now asks the same question"*. `emit_failure` does **not** ask it — it receives `json_output` as a required keyword and never calls `json_in_effect`. The in-module askers are `_SchemaCommand` at `:547` and `:566`. The layering conclusion is right; the stated reason is not.

---

### P3 findings

- **The disclosed duplicate is a triple, and the dichotomy is false.** On a timeout carrying a recognized error the same fact appears three times within ten lines — the `evidence:` block, the `gda:` diagnostics prose, and the raw engine stderr. The body frames the choice as *accept* vs *special-case per code*; a third option needs no per-code branching: remove the curated list from `_ended_run_diagnostics`, the **producer**. That changes JSON `diagnostics` bytes, so it must be a follow-up issue, not an in-slice change — but the current framing forecloses it.
- **`strip()` vs `strip("\n")` asymmetry** (`render.py`). The truth test sees all whitespace; the emitted value strips only newlines. `diagnostics="   \n   x"` renders a whitespace-only line, and a CRLF stream leaves a stray `\r` (the runner does no newline translation). `strip("\r\n")` fixes the CRLF half.
- **ANSI is stripped in the human channel but preserved in JSON.** `click.echo` strips escape sequences when stdout is not a tty, so engine output that emits colour survives `--json` (escaped in the JSON string) and is silently altered without it — a new divergence from this slice. Benign under the trusted-project assumption (ADR-0009), but the docstring says `diagnostics` is rendered "verbatim".
- **"Two 16 KiB tail-capped captures" invites a wrong inference.** That cap is a timeout-path property. A `--strict` failure is uncapped — I measured **303,342 bytes / 4,015 lines** of human output. The layout rule did its job (the verdict stayed on line 1), which is good evidence *for* the design, but the only bound named in the docstring does not apply to the motivating path.
- **No ADR amendment.** ADR-0002/0004 carry scope notes for #667/#670/#687. #685 makes the human channel a public output surface for the first time and neither ADR moves; `CONTEXT.md` now carries an ADR-shaped statement with no ADR behind it. Either amend or say why not.
- **`SKILL.md`** puts the consequence of disobeying the directive between it and the definitions it introduces, and its universal wording conflicts with `SKILL.md:82-84`, which describes usage refusals as envelopes (see P1).
- **`render.py` docstring**: *"Two of them carry rationale"* now introduces three bullets; the module docstring still scopes the file to *"rendering for `gda` results"*.
- **Emitter asymmetry**: `emit_failure(failure, *, json_output)` is keyword-only while its stated sibling `emit_result(result, json_output, render)` is positional, and `run`/`emit` order their parameters differently.

---

### Merge-order note for the integrator (not a review finding)

**#798 and #788 (`refactor/788-mounted-group-tree-walk`) conflict.** Both rewrite the same import line in `src/gda/hints.py`. Verified by trial merge at this head:

- `src/gda/headless.py` **auto-merges cleanly** (`walk_mounted_groups` and `json_in_effect`/`remember_argv` land in different regions) — so the conflict is textual only, in one place.
- `src/gda/hints.py:58` conflicts. Correct resolution keeps both and drops `ancestor_json`, whose only consumer in that file (`_json_in_effect`) moved out in this PR:

```python
from gda.headless import (
    emit_failure,
    json_in_effect,
    remember_argv,
    walk_mounted_groups,
)
```

I applied that resolution and ran the full `-m "not e2e"` suite on the merged tree: **2278 passed**. Whichever merges second takes this one-line fix.

---

### Verified clean — please do not re-litigate

- `--json` byte-identity and exit codes: 17 independent cases above, all identical.
- "Renderer keys on no code": constructed over all 92 registered codes, zero deviations.
- The hoist's import direction: `hints -> headless` is real (`emit_failure` is defined far below where a reverse import would sit); `render -> {models, script_errors}` closes without touching `errors`/`headless`/`hints`/`dispatch`, so `headless -> render` adds no cycle and follows an edge `errors.py` already had. `render.py` is the right home under ADR-0040 ("shared render helpers").
- `hints._answer` passing literal `json_output=True` is sound — `_answer` re-evaluates `json_in_effect` itself and the literal sits inside that branch.
- The `_fail` alias is fully removed; no Python-side references remain.
- The **34 changed tests: none were weakened.** 12 sampled individually; in every case the original assertion parsed JSON and is false-by-design after this change, so `--json` is the only correct repair, and no assertion was deleted. The real cost is coverage of the default path, quantified in P1 above.
- AC1, AC2, AC4 reproduced on a real engine; the body's `script run --strict` example block matches character-for-character apart from the line number.

The scope-defining property is solid and the human rendering is a genuine improvement over an escaped blob. What blocks is the gap between the invariants the PR asserts and what the code and the suite actually guarantee: the `usage` category falsifies the headline claim (P1), and the layout rule the body leads with is not pinned at all (P1).

qinhaihong-red:
## Review round handled — head `05d06cc3`

Two commits on top of the reviewed head, plus `974ec040` which landed the P1-1
disposition before this round. Everything below was reproduced before it was
changed and red-proofed after.

### P1 — the `usage` category falsified the headline invariant

**Adopted, the deep option.** `hints._answer` now calls
`emit_failure(refusal.failure(), json_output=json_in_effect(ctx))` like every other
call site, so a usage refusal is rendered by the one renderer instead of raising
click's `UsageError` (`974ec040`).

What moved and what did not, measured on the real CLI:

| | before | after |
| --- | --- | --- |
| exit code | 2 | **2** |
| `--json` stdout bytes | envelope | **identical** |
| human stream | stderr, click's panel | **stdout**, the rendered lines |
| human content | sentence only, no code/category/hint | head line + sentence + `hint:` |

The stream move is the deliberate half and is the direct consequence of the option
the review offered: `emit_failure`'s human arm is `typer.echo`, i.e. stdout, which is
where every other human failure in the product already lands. What is lost is click's
`Usage:` synopsis, whose remedy the sentence already carries. It is pinned by
`test_a_usage_refusal_without_json_is_rendered_by_the_same_one_renderer`.

The `hint:` branch is now production-reachable, not dead code:

```
$ gda scene inspect            # exit 2, stdout
error: unknown_command (usage)
`gda scene inspect` is not a gda command. Use `gda scene get` instead: `get` is the read verb in every group (ADR-0005); it reports the scene's structured node tree
hint: gda scene get

$ gda --schema                 # exit 2, stdout
error: unknown_option (usage)
`gda` has no option `--schema`. Use `gda schema` instead: `--schema` is a per-command flag; the whole-surface manifest is the `gda schema` command
hint: gda schema
```

`_refuse`'s fall-through is untouched — with no correction AND no `--json`, gda still
returns and the parser's own message (did-you-mean included) stands. That is the one
case outside the single renderer, and `CONTEXT.md` now scopes it to exactly that
condition rather than to "a parser refusal" in general, which would have read as
covering the `--json` arm too (it does not: `gda scene nosuchcmd --json` still gets a
gda envelope).

### P1 — AC3's test-pinning half

**Adopted.** Both mutations you ran now red, measured against the full
`-m "not e2e"` suite (2298 tests at this head):

| Mutation | Before | After |
| --- | --- | --- |
| `diagnostics` before `evidence` | 2277 passed, 0 failures | **1 failed** — `test_the_bounded_keys_all_precede_the_unbounded_diagnostics` |
| `emit_failure` always emits JSON | 1 failed | **8 failed**, one per call site plus the two `test_unknown_invocation` arms |

The ordering pin asserts the whole line sequence of one failure carrying `probe`,
`hint`, `evidence` and `diagnostics` at once, rather than one part in isolation as the
neighbouring tests do — that isolation is what let the order swap through.

Per-call-site coverage: one engine-free CLI test each for the four uncovered arms, and
each was red-proofed by reverting **that** site to `json_output=True` in isolation:

| Site | Case | Reverting it alone |
| --- | --- | --- |
| `dispatch.py:149` | `scene get --project <not a project>` | 1 failed |
| `dispatch.py:258` | `screen capture` against an empty `XDG_RUNTIME_DIR` | 1 failed |
| `headless.py:545` | `--params-json` + a positional arg | 1 failed |
| `headless.py:564` | `--params-json '{}'` | 1 failed |

Also red-proofed: dropping one `_evidence_lines` branch (3 failed), the head line
losing its category (10 failed), and the model growing a sixth field (1 failed).

### P2 — `_evidence_lines` inverted authority, and the `script_errors` divergence

**Adopted, and the divergence is closed rather than documented.**

The guard is in the shape that reds on both halves: a sample table asserted equal to
`set(FailureEvidence.model_fields)`, plus one render per sample so a dropped branch
cannot hide behind a neighbour's line. The docstring's claim that this is "one
authority read the same way by both channels" was, as you say, backwards — it now
states that the JSON side is model-driven, that this side is a hand-written branch per
field on purpose (each field reads differently), and that the model's field set is
therefore an authority a **test** holds it to.

On `script_errors` I did not take the "state the truthiness choice as rationale"
option, because the model publishes **three** states and truthiness collapses two of
them — `[]` is documented as "itself a finding". Your own live case is the proof: a
human got nothing where `--json` reported `[]`. Uniform `is not None` now, with the
empty list given a sentence rather than the bare header the layout rule forbids.
Verified on Godot 4.6.3, same run, both channels:

```
HUMAN:   termination phase: output_seen
         script errors: none recognized
JSON :   "termination_phase":"output_seen","script_errors":[]
```

### P2 — the duplicate `diagnostics` on a terminal

**Disclosed, not removed** — and the disclosure names why neither removal fits this
slice. Dropping `execute`'s stderr tee outright moves `--json`'s **stderr** bytes,
which is the one property the slice is scoped by. Dropping it only on the human
failure path silently loses the tee for the classifiers whose `diagnostics` is curated
prose rather than the raw stream, which is a different regression. The honest fix is
producer-side — the same place your P3 points at — and belongs to a follow-up issue,
not here.

### P2 — README family (DoD-4)

**Adopted** (`05d06cc3`). `README.md:70` now names the envelope by what it carries
(category + code) and states the two renderings; the `--json` flag row says outright
that the flag governs the failure outcome too, instead of leaning on a sentence whose
antecedent was "the result" — you were right that quoting it was retro-fitting. All
three translations carry both edits and were re-stamped with
`scripts/update_readme_i18n.py`, so the gate records a real re-translation.

`:439` and `:484` are left alone deliberately: both are about **which channel**
answers — error envelope versus success result — and that choice did not change. With
`:70` corrected they read correctly.

### P2 — "five call sites" and the stale hoist rationale

**Both adopted.** The body's table is now six rows with the line numbers and, per row,
the test that reds when that site reverts. The `json_in_effect` docstring no longer
claims `emit_failure` asks the channel question — it takes `json_output` as a required
keyword and never reads a context; the in-module askers are the two `_SchemaCommand`
refusals. The layering conclusion (the import direction) is restated as the actual
reason.

### P3

- **`strip("\n")` → `strip("\r\n")`** — adopted, with a pin. Note the pin splits on
  `"\n"` alone: `str.splitlines` treats a lone `\r` as a line boundary, so the obvious
  assertion hides exactly the character under test (my first attempt at this test was
  green against the unfixed renderer).
- **The triple, and the producer-side third option** — not foreclosed; the disclosure
  in `render_failure` now names the producer as where the fix belongs. Out of slice
  because it moves JSON `diagnostics` bytes.
- **ANSI stripped in human, preserved in JSON; the "two 16 KiB captures" bound;
  ADR amendment; `SKILL.md` ordering; the emitter's keyword-only asymmetry** — not
  taken this round. Reasons: the first two are docstring precision on a benign
  divergence under ADR-0009; the ADR question and the emitter signature are decisions
  I would rather not make unilaterally inside a review round. Flagging them for triage
  rather than silently dropping them.

### Validation

- `-m "not e2e"`: **2298 passed**. `ruff check` / `ruff format --check` / `pyright`:
  clean.
- **`--json` byte-identity re-run**, independently of the suite: 21 invocations
  (usage refusals in all four spellings, the hint-carrying near-miss, all four
  `--params-json` shapes including stdin `-`, `project_not_found` on both tails,
  live-with-no-daemon ×3, `binary_not_found`, `schema`, `version`) run in a disposable
  clone at the PR base `837174d3` and at this head. **stdout, stderr and exit code
  identical in all 21**, the sole difference being `gda version`'s own
  `revision`/`dirty` fields, which report the checkout by design.
- **Real engine (Godot 4.6.3, per-run `GDA_USER_DATA_ROOT`)**: `script run --strict`
  human + `--json` twin, a real 5s timeout carrying `script_errors: []` on both
  channels, a usage refusal, and both curated near-misses. Plus the targeted e2e
  selection (25 tests) green.

### Merge-order note

Confirmed and unchanged by this round — `src/gda/hints.py` still conflicts with #788
on the same import line, and your resolution (keep both, drop `ancestor_json`) is
still correct: `_json_in_effect` moved out of that file in this PR.

**Head: `05d06cc3`.**

qinhaihong-red:
## Full independent adversarial review — PR #798

Reviewed from first principles at pinned stacked head `05d06cc34df1670f0f7f3872bfda55989498752e` against `feat/687-typed-failure-evidence@837174d36e485599b09007cd6ebe33f17b5282f3`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Artifact review | **Pass** (one minor) |
| Domain modular architecture | **Changes required** |
| Standards | **Changes required** |
| Spec (#685) | **Pass** |

### Artifact review

Pass. Context, README/translations, Skill, and ADR consistently distinguish JSON envelopes from human rendering and preserve one failure outcome/exit code. Minor: issue #685's implementation map says five `emit_failure` call sites; the current slice has six (two dispatch, two `--params-json`, `HeadlessCommand.run`, and hints). Update the issue count/list.

### Domain modular architecture — required change (`DDMA-798-1`)

Human failure output has two external-effect owners. `src/gda/headless.py:753-754` unconditionally forwards child stderr before classification. For classifiers whose `error.diagnostics` is that same raw stderr, `src/gda/render.py:101-102` then renders it again on stdout. The human renderer therefore cannot guarantee one coherent outcome.

This was reproduced at the exact head with stdout/stderr captured separately: `Godot Engine v4.6.stable` appeared once under the human diagnostics block on stdout and again verbatim on stderr. The existing focused tests all pass because they do not assert the two streams independently.

Classify before forwarding. For human failures whose diagnostics already contain raw stderr, let `render_failure` own emission. Preserve the tee for JSON, successes, and failures with curated diagnostics; represent that producer distinction explicitly rather than infer it from error codes.

### Standards

The same duplicate output is a hard coherent-interface/DRY violation introduced by the new human channel. No separate Fowler smell was found.

### Spec

Pass. #685's real-line human layout, compact empty diagnostics, JSON byte invariance, unchanged exits/help/schema, and all actual emission sites are implemented. The duplicate cross-stream side effect is an adjacent interface defect, not a missing AC.

Sources: `CONTEXT.md` definitions for **Error envelope**, **Failure evidence**, and **Near-miss hint**; ADR-0004/0040; issue #685; README/translations; Skill; renderer contract. Writing reference: ASD-STE100. Channel terminology, totality, absent parts, ordering, stdout behavior, and call-site descriptions were inspected. No substantive prose/terminology issue was found; the stale count is minor.

Validation: `git diff --check` passed; 59 focused human-output/headless/render tests passed; the separate-stream reproduction above still demonstrated the defect. Current hosted CI has no failed or pending required check; Godot e2e is skipped and was not rerun.

qinhaihong-red:

New head: `db58e86ae` — rebased onto #796's review-fix head `883b17b41` (stacked cascade), plus one fix commit.

| `DDMA-798-1`: human failure output has two external-effect owners — the pre-classification stderr tee and the renderer both say the same bytes when a classifier's `diagnostics` IS the raw child stderr (reproduced at head) | **Fully adopt** (mechanism differs from the suggestion) | Reproduced independently: real-engine `path_not_found` run, streams captured separately | Classify-before-forward, with the producer fact carried EXPLICITLY: `execute` attaches the child's stderr to the `Failure` (`child_stderr`, never serialized) instead of printing, and `emit_failure` — the one place the channel is known — forwards it unless the human channel is about to print byte-identical `diagnostics`. Byte identity rather than a per-producer flag: a flag can drift from what a factory actually does, identity cannot; the reviewer's carve-outs fall out of it (curated/tail-capped diagnostics differ from the raw stream, so their tee — the only complete copy — survives; `--json` unconditional; successes unchanged) | Duplication test run against pre-fix sources and observed red; JSON-invariance + capped-diagnostics + success-tee pins green on both; real engine: human = one copy on stdout, empty stderr; `--json` = envelope + full tee. 2301 non-e2e green, ruff/pyright clean |
| Issue #685's implementation map says five `emit_failure` call sites; the slice has six | **Fully adopt** (on the issue) | `grep` at head: 2 dispatch tails, 2 `--params-json`, `HeadlessCommand.run`, hints | Dated call-site-count note appended to #685 | Issue body diff |

`render_failure`'s docstring paragraph that disclosed the duplication as a follow-up now records the delivered ownership rule instead.

PR #799, opened 2026-08-31
feat(gda): one containment answer, and refuse a target another project owns (#763, #697)

### LINKED ISSUE #697: script validate: derive-vs-refuse for a target under a different project.godot (ADR-0006 amendment)
## Review-round source

Created during PR #695's review (issue #658, wave 2) to keep GDA-DF-035's remainder tracked: #658 delivers the ADR-0006-conformant refusal (script outside the RESOLVED project → structured refusal, `project_root` on success), but the EXACT dogfooded invocation still reproduces on that head, in both readings:

- a project nested in a plain workspace, validated from the ancestor → projectless run, the false `res://` dependency cascade still appears, now attributable via `project_root: null`;
- an ancestor that IS a project, target inside a nested project → contained, same false cascade, `project_root: <outer>`.

Both boundaries are pinned by #695's tests as deliberate scope lines. The fix needs the decision ADR-0006 explicitly deferred: nearest-`project.godot` discovery (derive the owning project from the target) versus refusing when the target sits under a DIFFERENT `project.godot` than the resolved one.

## What to build

1. **The ADR-0006 amendment**: decide derive-vs-refuse for a target owned by another `project.godot` (including the nested-under-resolved case), against ADR-0006's recorded rejection rationale (a call spanning multiple paths could imply different roots — #663's batch mode is exactly that hazard). Record the decision as a dated amendment.
2. **The implementation** per the decision, for `script validate` (and the same containment question for `script run`, which #675 gave filesystem path forms; `gda.project.path_outside_project` is the reusable seam).
3. **Registry follow-on to consider**: a sibling code `target_outside_project` — #695
   reuses `project_not_found`, and **#663 has since shipped**, so the "second producer of
   the same class" this point predicted as the honest trigger now EXISTS rather than being
   hypothetical. `project_not_found` is produced from five distinct sites on `main`:
   `errors.py:468` (`script run`, no resolved project), `errors.py:486` (#658, target
   outside the resolved project), `errors.py:754` (explicit `--project`/`$GDA_PROJECT`
   that does not resolve), `live_runner.py:65`, and `daemon.py:421`. Decide against that
   count, not against "a fourth producer". Coordinate with #687, which is separately
   re-examining what an envelope can carry.
4. **Doc note**: README's `script validate` paragraph (and any onboarding text) should state that a `valid=false` full of `res://` dependency errors can be a project-context artefact, readable from `project_root` (the bundled skill gains this clause in #695 directly).
5. **Known consequence to weigh**: with the refusal in place there is no escape hatch to validate a standalone `.gd` while cwd is a project (`--project ""` is refused as empty); the amendment should either bless that or define the override.

## Acceptance criteria

- [ ] The ADR-0006 amendment records derive-vs-refuse (and the override question) with rationale.
- [ ] The GDA-DF-035 invocation (ancestor workspace, nested project) no longer emits the false dependency cascade — it derives the owning project or refuses per the decision.
- [ ] #663 batch semantics and #675's `script run` containment follow the same decision.
- [ ] Docs updated per point 4; the `target_outside_project` code question is decided (mint or explicitly decline).

## Blocked by

- ~~#658's PR #695 merging~~ — **cleared**: #658 is closed; the refusal seam
  (`gda.project.path_outside_project`), the `project_root` field and the boundary tests
  are on `main`.

## Reproduced on `main` (2026-08-27)

Reading 2 of the two above, on a real engine — `outer/` and `outer/inner/` each a project,
`inner/main.gd` doing `preload("res://local_dep.gd")` where `local_dep.gd` sits beside it:

```
gda script validate outer/inner/main.gd --project outer/inner
  -> {"valid":true,  ..., "project_root":".../outer/inner"}
gda script validate outer/inner/main.gd --project outer
  -> {"valid":false, "diagnostics":[{"line":2,
        "message":"Parse Error: Preload file \"res://local_dep.gd\" does not exist."}],
      "project_root":".../outer"}
```

The same file is valid or invalid depending only on which ancestor was named, and nothing
in the envelope says the second verdict is a project-context artefact except
`project_root` — which is exactly why point 4's doc note is part of the slice, not
optional polish.

### LINKED ISSUE #763: One containment answer for a res:// target, instead of three

qinhaihong-red:
Closes #763
Closes #697

The chain's final slice: one decision with its implementation, delivered together because
neither issue is complete without the other. #697 decides what "belongs to the resolved
project" means; #763 makes that one answer instead of three.

Stacked on `feat/685-human-failure-rendering` (rebased onto its `05d06cc3` after the
#687/#796/#798 chain moved) — review the six commits, not the whole diff against `main`.

## Decision 1 (#697) — refuse, do not derive

Recorded as a dated ADR-0006 amendment (2026-08-31). **A target the resolved project does
not own is refused; `--project <owner>` is the supported way to say what you mean.**

"Does not own" is two halves, one verdict:

| half | asks | applies to |
|---|---|---|
| **containment** (`path_outside_project`) | is the target inside the resolved project's tree? | `script validate`, `script run`, `resource import` |
| **ownership** (`owning_project`, new) | is the resolved project the *nearest* `project.godot` at or above the target? | `script validate`, `script run` |

Ownership is the half GDA-DF-035 exposed and #695 pinned as a deliberate scope line. A
target can sit squarely inside the resolved tree and still be owned by a nested project,
and then its own `res://` references resolve against a foreign root: `outer/inner/main.gd`
reads `valid` under `--project outer/inner` and `invalid` under `--project outer`, from a
cascade of false missing-file errors.

**Finding an owner is not deriving one.** The walk is bounded (stops at the resolved
project), lexical (reads the caller's spelling, so a monorepo's symlinked-in shared
directory stays inside as the engine reads it), and its result is *reported, never
adopted*: resolution is still flag > env > cwd, one call still has exactly one root, a
batch naming several owners still returns one verdict. ADR-0006's two objections to
derivation — undefined outside any project, several roots per call — do not apply, because
no owner simply means no refusal.

**All three commands apply ownership, for one reason each.** `script validate` and
`script run` because a target compiled or run against a foreign root resolves its *own*
`res://` references there. `resource import` because the ENGINE does the same thing:
`EditorFileSystem::_should_skip_directory` (`editor/file_system/editor_file_system.cpp:3482`)
skips a directory holding a nested `project.godot` — *"Skip if another project inside
this"* — so such an asset cannot be imported into the outer project at all. gda used to
accept the request, spend a pass and return `not_importable`, while `--dry-run` predicted a
sidecar that would never appear. (An earlier revision of this PR exempted `resource import`
on the opposite claim about the engine; the review below caught it against the source.)

Sub-decisions the issue asked to settle, all recorded in the amendment:

- **standalone `.gd` while cwd is a project: blessed, no `--no-project` flag.** The cwd is
  the last resolution step, so the answer is to not stand there, or to name the owner. A
  projectless escape hatch would add a cross-cutting flag to every command in the surface
  for a case no dogfooding round has produced. Trigger to revisit: one that does.
- **`user://` on `script validate`: decline to narrow.** Verified on Godot 4.6.3 —
  `gda script validate user://bad.gd` returns the real parse error, so it is a working
  input, not a bypass. Each command's scheme set follows from its own contract
  (`script run`: ADR-0031's one-address model; `resource import`: an asset is a `res://`
  member; `script validate`: the engine loads it and reports truthfully).
- **`target_outside_project`: minted.** #697 stated the honest trigger as a second producer
  of the class; convergence produces three. `project_not_found` was true of neither the
  condition nor the remedy — a project *was* resolved, and the fix is a *different* one.
- **known edge left open, stated:** a case-differing project spelling (`--project …/Game`
  against on-disk `…/game`) is still refused on a case-insensitive filesystem. Both
  readings compare strings and `Path.resolve()` does not canonicalize case; a real fix
  needs a `samefile`-style walk and is independent of this decision.

## Decision 2 (#763) — one containment answer

`gda.project` — ADR-0006's path authority — now owns the whole answer:
`canonical_res_path` (relocated from `gda.script_errors`, killing the inverted import;
the stderr parser is now one consumer among several), `res_escape_remainder`,
`path_outside_project`, `owning_project`.

`canonical_res_path` also gains `String::simplify_path`'s **empty join** (ustring.cpp
4223-4232), so the project root has ONE spelling, `res://` — closing the parity gap PR #766
documented and this issue owns. `script run`'s two-member root set (`{"", "."}`) goes with
it; it existed only to accommodate a primitive that gate did not own.

### Reconciliation table — spelling × command, before → after

Measured on a real Godot 4.6.3, project `game/` with `bar.gd`/`pic.png` inside and
`outside.gd`/`outside.png` one level up.

| spelling | `script validate` | `script run` | `resource import` |
|---|---|---|---|
| `res://bar.gd` | OK → OK | OK → OK | OK → OK |
| `res://foo/../bar.gd` (net-inside) | OK → OK | OK → OK | **`invalid_params` → OK** |
| `res://./bar.gd` | OK → OK | OK → OK | OK → OK |
| `res:///bar.gd` | OK → OK | OK → OK | **`invalid_params` → OK** |
| `res://foo\..\bar.gd` | OK → OK | OK → OK | **`invalid_params` → OK** |
| `res://../outside.gd` | `project_not_found` → **`target_outside_project`** | `invalid_path` → **`target_outside_project`** | `invalid_params` → **`target_outside_project`** |
| `res://..\outside.gd` | `project_not_found` → **`target_outside_project`** | `invalid_path` → **`target_outside_project`** | **admitted, then `invalid_params` "does not exist"** → **`target_outside_project`** |
| `res://a\..\..\outside.gd` | `project_not_found` → **`target_outside_project`** | `invalid_path` → **`target_outside_project`** | admitted, then "does not exist" → **`target_outside_project`** |
| `res://a/..` / `res://.` / `res://` (root) | op-level `invalid_path` (unchanged) | `invalid_path` (unchanged) | "asset `res://.` does not exist" → "asset `res://` does not exist" |

The `res://..\…` row is the Windows gap: `PurePosixPath` read it as ONE segment carrying no
`..`, so `resource import` admitted it and only a later `is_file()` check stopped it on
POSIX — on native Windows `\` IS a separator and the join reaches the parent directory. The
shared canonicalizer folds `\` first, as the engine does, so the escape is refused by the
rule on every platform. `test_the_engine_agrees_the_escape_leaves_the_project` runs Godot
directly on that spelling and reads back that it executed the file above the project.

### Every change to an admitted-input set, enumerated (#763 AC)

1. `resource import` **accepts** `res://foo/../x`, `res:///x`, `res://a//..//x`,
   `res://foo\..\x` — all net-inside, all already accepted by both script commands.
2. `resource import` **refuses** `res://..\x` and `res://a\..\..\x` as escapes (previously
   admitted; inert on POSIX, a real escape on Windows).
3. `resource import` maps the project-root spellings to `res://` instead of the bogus
   `res://.`; still refused by the existence check, now under its real address.
4. `script validate` **refuses** a target a nested `project.godot` owns (was accepted;
   `project_root` was the only signal).
5. `script validate` **refuses** a projectless target that has an owner (was validated
   projectless against gda's cwd). This is the widest change here and is deliberate: a
   projectless verdict on a file inside a project is computed in a context that file never
   runs in — `res://` resolves against gda's cwd, and the project's autoloads, `class_name`
   registry (ADR-0032) and settings are all absent — so `valid: true` from such a run is
   not evidence the script compiles in its project. ADR-0006's projectless promise is kept
   for the files it was written for: a `.gd` no `project.godot` claims is still validated by
   filesystem path. The refusal names the exact `--project` to pass.
6. `script run` **refuses** a script a nested `project.godot` owns.
7. `resource import` **refuses** an asset a nested `project.godot` owns (was accepted,
   cost an engine pass, and came back `not_importable` because the engine skips that
   directory).
8. `script validate`, `script run` and `resource import` **refuse** a target reached
   through a DIRECTORY symlink at a checkout that carries its own `project.godot` (was
   accepted; the owner is the checkout). A **file** symlink at a file inside another
   project stays accepted — the walk climbs the directories above the link, and those
   are the resolved project's. #697 recorded this edge for the amendment to settle; one
   rule settles both halves, because only a directory can carry a marker into the tree.
   A symlinked-in PLAIN library (no `project.godot`) is untouched and still inside.
9. Codes only, no verdict change: the escape refusals in the table above.

### What deliberately did NOT move (#763 AC)

`script run`'s **verdict-matching** rules stay local — the engine-`strip_edges` trailing
suffix and the engine-log line boundary keep the canonical identity matchable against what
the engine echoes back on stderr, a concern only a channel that reads that stderr has — as
does its **root-address** rule (a script address names a file, not a directory). Six of its
seven path shapes stay `invalid_path`, ADR-0031's own ABI edge; only the escape moved.
`tests/test_res_containment_consistency.py::test_the_rules_that_stay_script_run_s_own_are_stated_here`
is that statement in executable form.

## Envelope

`target_outside_project` — ADR-0002 registry + row, `operation` category, exit 4,
**classifier**-source (so NOT GDScript-mirrored: ADR-0006 keeps project resolution CLI-side,
no operation can report it; `operations.gd` is untouched). `project_not_found`'s description
loses the outside-target clause on both sides in lockstep, so the #701 drift pin stays green.

The refusal carries the mismatch's three coordinates as typed evidence (#687):
`target_location`, `project_root`, `owning_project` — each omitted, never null, when that
refusal does not know it. `script run`'s pre-launch escape has none of the three and says so
by omitting the key rather than naming a project the call may not have.

## Red-proofs

Each guard reverted in turn, on the full non-e2e suite. **Re-measured at `6c123088`** —
the counts below are this head's, not an earlier one's (#799 review P3-1):

| reverted | fails |
|---|---|
| the root-collapse parity fix | 14 (incl. `script run` LAUNCHING `res://.` / `sub/..`) |
| `resource import` back to its own `".." in parts` res:// arm (the exact pre-#763 code) | 19 |
| the ownership probe neutralised, all three commands | 17 |
| `script run`'s escape back on `invalid_path` | 17 |

And this round's own guards:

| reverted | fails |
|---|---|
| the remediation back to "Pass `--project <owner>`" alone | 4, in `test_e2e_res_containment.py` — each parses the re-issue out of the message and RUNS it |
| `resource import` reporting the project as spelled | 2 |
| the re-issue target subtracted from the RESOLVED pair instead of the lexical one | 1 |
| ownership dropped from `resource import` alone | 6 (was 1 before the cross-command row) |

## Verification

- `pytest -m "not e2e"`: **2375 passed**. `ruff check` / `ruff format --check` / `pyright`: clean.
- Registry + ADR drift pins green (both sides moved together) — including ADR-0004's
  `Amendment (2026-08-31, #687)`, whose recorded evidence-producer set the rebase brought
  under this branch and which now names this PR's two refusal builders.
- **Real-engine e2e: every module run, all green** — the new `test_e2e_res_containment.py`
  (**19**), plus `script`, `script_run`, `resource_import`, `project`, `scene*`, `node`,
  `resource`, `object_property`, `asset_file`, `classname_resolution`, `project_analysis`,
  `project_walk`, `op_robustness`, `params_json`, `info`, `headless_number_reads`,
  `launch_timeout`, `user_data`, `export*`, `mcp_*`, `harness_install`, `daemon`, `diag`,
  `logger`, `game`, `input`, `perf`, `screen`, `live_number_transport`.

## Cross-language coordination (not in this PR)

The GDScript side's `_canonical_resource_path` (`operations.gd:6561`) is
`String.simplify_path()` directly, so it already agrees with the Python primitive on every
spelling here — and now on the empty join too, which is the divergence this PR removed:
before it, `res://a/..` was `res://.` in Python and `res://` in GDScript. `operations.gd` is
untouched (that file's own consolidation is #775's open scope). One observed
asymmetry, recorded rather than changed: `script run` canonicalizes the argv it hands the
engine while `script validate` passes the caller's spelling through, so
`res://nosuchdir/../bar.gd` is a truthful engine `path_not_found` from the latter (the
engine's file access resolves `..` through the filesystem) though it is lexically
net-inside. Not a containment verdict; pinned as such in
`tests/test_e2e_res_containment.py`.

## Review round (commit 4)

An independent adversarial review of the first three commits found the convergence sound
and the ownership half leaky in four places. All are reproduced, fixed and red-proofed in
`fix(gda): close the four seams…`:

| finding | effect | fix |
|---|---|---|
| lexical-only stop | a correct call refused, with `owning_project == project_root` — impossible for this code — whenever `--project` and the target were spelled through different-but-equal paths (macOS `/tmp` → `/private/tmp`) | the stop consults both readings, as containment does |
| unbounded start | an escaping `res://` walked from *outside* the tree, so an unrelated ancestor could be named the "owner"; `""` / `.` / `sub/..` started above their own stop | ownership is asked only of a target that is a FILE inside the resolved tree |
| `res://` target location | `evidence.target_location` reported `<project>/res:/inner/main.gd` — a path that does not exist — for a `res://` target | one anchoring rule for the walk and the reported location, mirroring containment's branching |
| human channel | `render_failure` claims totality; `_evidence_lines` enumerates by hand and dropped all three new fields (its guard test enumerated by hand too) | fields rendered; the new guard reads `FailureEvidence.model_fields` |

It also refuted the `resource import` exemption against the engine source, which is why
that command now applies ownership too, and prompted the ADR to state two things it should
have: the projectless half NARROWS ADR-0006's fallback (the original sentence is about a
calling convention, not file ownership), and the `--project` it pushes callers onto WIDENS
the Project-code execution surface — a projectless validate runs no autoloads. Recorded as
a trade, not a free improvement.

## Review round 2 (commits 5-6)

A second independent adversarial review (Architecture and Standards: approve) found four
P2s and five P3s. All reproduced on the reviewed head, all handled; nothing declined
outright, one adopted by a smaller mechanism than proposed.

| finding | effect | fix |
|---|---|---|
| P2-1 | the refusal's remediation was not executable: "Pass `--project <owner>`" leaves the caller's own spelling, which anchors at the project, so the re-issue reached a file that is not there (`path_not_found`; `script_not_found` after a full engine launch) — and `evidence.target_location` is a form `script run` structurally refuses | the message states BOTH operands: the `--project` and the target respelled relative to that owner (`gda.project.owner_relative_target`, lexical on both sides — the only total reading). Red-proofed by an e2e that PARSES the re-issue out of the message and runs it, per command |
| P2-2 | `resource import` reported `evidence.project_root` as the caller spelled it, against a field whose schema promises "its resolved form" — one call answered `/tmp/…` there and `/private/tmp/…` from `script validate` | the refusals report the resolved root; the CHECKS keep reading the spelled one, as the two sibling gates already did |
| P2-3 | the symlink edge #697 recorded had been silently decided and split, and one half was written down nowhere; a unit test's comment contradicted the shipped envelope | the split is enumerated in the ADR-0006 amendment, the catalog and this body (item 8 above); the envelope reports the owner RESOLVED and the test now says so and why |
| P2-4 | wave coherence with #795: both PRs asserted what gda's `res://` walk excludes, and both rewrote the passage | the exclusion sentence is ADR-0032's to own — dropped from ADR-0006 and the catalog here; ADR-0006 carries the matching layer statement (enumeration is not targeting) that #795 carries on its side |
| P3-1 | the Red-proofs and Verification counts were stale | re-measured at this head, and one row was understated by more than the reviewer's own correction (ownership neutralised: 3 → 17) |
| P3-2 | the `model_fields` guard was only strict for string fields (2 lines of slack) | **superseded by the rebase**: #798 landed a stronger pair — a sample table asserted EQUAL to `FailureEvidence.model_fields`, plus one render per field ALONE. The weaker guard was dropped in the rebase and the three new fields joined that table |
| P3-3 | the cross-command pin covered containment only | the ownership half now drives all three gates from one spelling table, at the level that BUILDS the envelope — which is also where P2-2 lived |
| P3-4 | the ownership walk is O(depth²) syscalls | measured cost recorded in a comment with why it stands; not changed — the lockstep rewrite would have to keep two chains in step through the symlink case the double reading exists for |
| P3-5 | the `--all` follow-up was drafted in a comment, not on a persistent surface | the draft is reproduced below, verbatim; filing stays user triage |

Two things NOT adopted, with the reason:

- **A fourth evidence field for the re-issue target.** It is derivable from the two
  coordinates already published (`target_location` relative to `owning_project`), which is
  exactly the second clause of ADR-0004's criterion for what may enter that object — and
  the shared `error` schema is repeated once per command, so the criterion is what keeps
  `gda schema` from growing by 76.
- **Carrying it on the `hint` key.** ADR-0004's #670 amendment bounds `hint` to an
  invocation gda could not RESOLVE, from the curated `hints.py` table. This invocation
  resolved fine and was wrongly rooted; widening the key would overturn a recorded boundary
  for a sentence that works.

## Follow-ups (drafted, not filed — user triage)

Reproduced verbatim from the round-1 reply, so it lives somewhere durable:

> **Follow-up worth an issue** (drafted, not filed): *`script validate --all` and every gda
> `res://` walk descend into a nested project the engine skips.* `operations.gd`'s
> `_should_descend` excludes only `res://.godot`; `EditorFileSystem::_should_skip_directory`
> also skips a directory holding `project.godot` and one holding `.gdignore`. Aligning gda's
> walk with the engine's would close the `--all` cascade and make `script list` / `scene
> list` / project analysis / the import-gap listing agree with what the engine actually
> sees — one shared change, wide blast radius, wants its own slice.

One correction to make before filing: the "`_should_descend` excludes only `res://.godot`"
clause is stale as of #795, which adds cache-alias and descent-cycle exclusions by
filesystem identity. The gap itself — a nested project's directory is descended — is
unaffected by that PR and reproduces at this head.

## Known gaps, recorded not hidden

- **`script validate --all` still produces the cascade.** It enumerates through gda's own
  `res://` walk (`operations.gd`'s `_should_descend`), which does not skip a directory
  holding a nested `project.godot` — the engine's editor scan does. (What that walk DOES
  exclude is ADR-0032's to state, and #795 is changing it; this PR no longer asserts it.)
  So `--all`
  compiles nested-project scripts against the outer root and the same file gets opposite
  answers depending on the selector. Closing it means changing the walk every collector
  shares (`script list`, `scene list`, project analysis, the import-gap listing) — its own
  slice, its own blast radius. Stated in the ADR, the catalog and the bundled skill.
- **Other path-taking commands do not ask the question.** `gda scene validate
  res://inner/main.tscn --project outer` still compiles a nested project's scene against
  the outer root. The amendment bounds itself to the three commands that implement it.
- **The case-differing project spelling** (`--project …/Game` vs on-disk `…/game`) is
  unchanged; a real fix needs a `samefile`-style walk.

## Also

`PITFALLS.md` gains a verified entry found while running the e2e batches: exporting
`GDA_USER_DATA_ROOT` for a whole suite relocates the platform application-data directory,
which is also where Godot keeps export templates — so an unrelated export test reports
`templates_installed=false` on a host that has them.


qinhaihong-red:
## Review round handled (commit `915a19f9`)

An independent adversarial review of `9a10ef12..864a6235` ran against the real toolchain (Godot 4.6.3, engine source). Every finding, and what it got:

| # | finding | verdict | where |
|---|---|---|---|
| P1-1 | `owning_project`'s stop test is lexical-only, so a project and a target spelled through different-but-equal paths (`/tmp` → `/private/tmp`) refuse a correct call, reporting `owning_project == project_root` | **fixed** — the stop consults both readings, as `path_outside_project` documents it must | `src/gda/project.py`; red-proofed by `test_a_project_and_a_target_spelled_through_different_paths_have_one_owner` |
| P1-2 | `script validate --all` still produces the GDA-DF-035 cascade, and the code claimed it could not — the engine's scan skips a nested-project directory (`_should_skip_directory`), gda's own walk does not | **claim corrected; gap recorded, not closed** — closing it means changing `operations.gd`'s `_should_descend`, which every collector shares (`script list`, `scene list`, project analysis, the import-gap listing). Its own slice. Now stated in ADR-0006, `command-catalog.md` and the bundled skill | follow-up draft below |
| P1-3 | `_target_location` hands a `res://` string to filesystem anchoring, so `evidence.target_location` named `<project>/res:/inner/main.gd` | **fixed** — one anchoring rule (`gda.project._anchored_target`) serves the ownership walk and the reported location, mirroring containment's own branching | red-proofed by `test_target_location_anchors_a_res_address_in_the_namespace_it_names` + a CLI-level pin |
| P1-4 | the ADR's reason for exempting `resource import` is factually false — the engine SKIPS a nested-project directory | **fixed by adopting the opposite decision** — `resource import` now applies ownership too, on the engine's own reason. All three commands ask both halves | `_asset_res_path`; unit + e2e arms |
| P2-1 | the human channel drops all three new evidence fields; its guard test enumerates fields by hand so it could not catch that | **fixed** — fields rendered, and the new guard reads `FailureEvidence.model_fields` so the next field is caught too | `src/gda/render.py`, `test_the_human_channel_renders_every_declared_evidence_field` |
| P2-2 / P2-3 | the walk could start outside the resolved tree — an escaping `res://` climbed from the project's parent; `""`/`.`/`sub/..` started above their own stop and named the parent project | **fixed together** — ownership is asked only of a target that is a FILE inside the resolved tree; everything else is containment's, which already answered them | red-proofed by `test_ownership_says_nothing_about_a_target_outside_the_resolved_tree` |
| P2-4 | the amendment quotes ADR-0006's projectless promise as continuity it does not have; and `cd` does not escape the ownership half | **fixed** — the amendment now says the half NARROWS the fallback, states that the `--project` it pushes callers onto WIDENS the Project-code execution surface (autoloads boot), and scopes the `cd` remedy to containment | ADR-0006 |
| P3 | amendment scope should be bounded to the commands that implement it; `4203` → `4204` citation; `RES_PREFIX` vs literals; the "first `://`" wording; no engine arm for the empty-join parity; `/tmp/proj` unit paths now environment-coupled | **all taken** — including a new e2e that runs Godot on `res://a/..` and reads back `res://`, and a named guard test for the `/tmp/proj` assumption so a stray marker fails with its cause stated rather than flipping a dozen verdicts | |

Nothing was declined. Gates after: `pytest -m "not e2e"` 2343 passed, ruff + pyright clean, and the affected e2e suites re-run green against the real engine (210 tests: `res_containment`, `script`, `script_run`, `resource_import`, `resource`, `asset_file`, `project`).

**Follow-up worth an issue** (drafted, not filed): *`script validate --all` and every gda `res://` walk descend into a nested project the engine skips.* `operations.gd`'s `_should_descend` excludes only `res://.godot`; `EditorFileSystem::_should_skip_directory` also skips a directory holding `project.godot` and one holding `.gdignore`. Aligning gda's walk with the engine's would close the `--all` cascade and make `script list` / `scene list` / project analysis / the import-gap listing agree with what the engine actually sees — one shared change, wide blast radius, wants its own slice.

qinhaihong-red:
## Independent adversarial review — head `915a19f9ea` vs stacked base `97feae2ab6`

Reviewed in an own detached archive at `915a19f9ea` (`97feae2ab6...915a19f9ea`, 5 commits); the chain beneath is not under review. Probes on Godot `4.6.3.stable.official.7d41c59c4` and engine source `4.6-stable-3260-g070dc9897e`, each run with its own `GDA_USER_DATA_ROOT`. Gates re-run here: `pytest -m "not e2e"` **2343 passed**, `tests/test_e2e_res_containment.py` **15 passed**, `ruff check` / `ruff format --check` / `pyright` clean.

**Verdicts** — Architecture/DDMA: approve. Standards: approve. Artifact/Markdown: approve with changes. Spec: approve with changes.

### What held up under stress

The convergence is real and the engine agrees. `canonical_res_path` matches `String.simplify_path()` on all 11 spellings I ran through a real engine, the empty join included (`res://a/..` → `res://`, `res://.` → `res://`, `res://foo\..\bar.gd` → `res://bar.gd`, `res://..\outside.gd` → `res://../outside.gd`). Every row of the reconciliation table reproduces on all three commands. Both `/tmp` → `/private/tmp` directions are accepted (seam 1), the case-differing project spelling falls through to containment with `owning_project` correctly omitted — no `owning_project == project_root` reachable — and a symlinked-in plain library stays inside. Citations check out: `editor/file_system/editor_file_system.cpp:3482` is the `FileAccess::exists(p_path.path_join("project.godot"))` line, `ustring.cpp:4149`/join `4223-4232` are right, `operations.gd`'s `_should_descend` really does exclude only `res://.godot`, and `gda script validate user://bad.gd` on a real `user://` file returns `Parse Error: Expected parameter name.` The relocation is clean and acyclic: `gda.script_errors` now imports `gda.project` and `gda.project` imports nothing from `gda`.

Red-proofs sampled in my own archive (full non-e2e suite each): empty-join parity fix reverted → **14 red** (matches); `script run`'s escape back on `invalid_path` → **18**; seam‑1 stop back to lexical-only → **1**; seam‑2 walk-start bound removed → **2**; seam‑4 human-channel fields dropped → **2**; ownership removed from `resource import` only → **1**; from `script validate` only → **3**.

---

### P2‑1 — the refusal's remediation is not executable for the spelling that produced it

`script validate inner/main.gd --project <outer>` refuses with *"Pass `--project <outer>/inner` to work on it in the project that owns it."* Doing exactly that:

```
$ gda script validate inner/main.gd --project <outer>/inner
{"error":{"code":"path_not_found","message":"script file does not exist: inner/main.gd"}}
```

because a relative path anchors at the project (`project_anchored`, ADR-0006). `script run` is worse: the same message, and the retry costs a full engine launch (`script_not_found`), while `evidence.target_location` — the coordinate the builder's docstring calls *"exactly the pair a caller needs to … re-issue with that `--project`"* — is an absolute path that `script run` structurally refuses:

```
$ gda script run <abs>/outer/inner/main.gd --project <outer>/inner
{"error":{"code":"invalid_path","message":"script run requires a project-relative or res:// script path, got: '/…/inner/main.gd'"}}
```

The only working re-issue is `gda script run main.gd --project <outer>/inner` — a third address the refusal never names. Either name the re-issue **target** beside the project, or drop the claim that the coordinates are sufficient for a machine re-issue.

### P2‑2 — `resource import` reports `evidence.project_root` unresolved, unlike its two siblings

`FailureEvidence.project_root` is documented as *"The project gda resolved for this call, **in its resolved form** — the same value a successful result reports as `project_root`."* Same fixture, `--project` spelled `/tmp/…`:

- `resource import` → `"project_root": "/tmp/…/outer"`
- `script validate` → `"project_root": "/private/tmp/…/outer"`

`_asset_res_path` builds `project_abs` with `expanduser()` + cwd-anchoring and never `.resolve()`s it, then passes it to both `target_owned_by_another_project_failure` and `target_outside_project_failure`. That falsifies the field's own schema description for one of the three producers and the body's "one message and one pair of typed coordinates". One `.resolve()`.

### P2‑3 — the edge #697 recorded is decided, changed, and absent from the decision record

#697's first comment recorded, for this amendment's consideration: *"a symlink inside project P pointing at a file owned by another project Q is ACCEPTED and compiled against P … the link-spelled instance of the nested-under-resolved boundary this issue owns."* At this head that edge **splits**, and only one half is written down anywhere public:

- directory link to a project (`outer/addons/vendored` → `../../vendor/pkg`, which has `project.godot`) → now **refused**, `owning_project` = `/…/vendor/pkg`;
- file link to a file in a project (`outer/linked_vend.gd` → `../vendor/pkg/vend.gd`) → still **accepted**.

The refusal is a behaviour change on a recorded open edge, pinned only by `test_a_symlinked_in_PROJECT_is_owned_by_it_like_a_nested_one`. It is not in the ADR-0006 amendment (which mentions symlinks only as the *rationale* for a lexical walk), not in the body's "Every change to an admitted-input set, enumerated (#763 AC)", and not in the catalog. Also, that test asserts *"the owner it names is the path the caller typed, not the link target"* — but the builders pass `owner.resolve()`, so the envelope names the link **target**. Record the decision in the amendment, and reconcile the test's stated intent with what ships.

### P2‑4 — wave coherence with #795 (`fix/760-res-walk-symlink-policy`, head `2bd31608df`)

Two collisions the serial merge has to settle; neither PR mentions the other.

1. **Prose.** This PR asserts in ADR-0006, `docs/command-catalog.md` and the body that gda's walk *"excludes only `res://.godot`"*. #795 rewrites `_should_descend(dir, child, chain)` to also exclude symlink aliases of the engine cache (by `DirAccess.is_equivalent`) and descent-chain cycles, and rewrites the catalog's exclusion passage. Whichever lands second leaves the other's sentence false.
2. **Substance.** #795's ADR-0032 amendment blesses *"a vendored checkout reached through a link"* as ordinary enumerable content whose `class_name`s resolve into the outer index. This PR refuses every targeted op on a file in such a checkout once it carries a `project.godot`, while `--all` still validates it — reproduced here:

```
$ gda script validate --all --project <outer>
… "res://addons/vendored/vend.gd" valid:true …
      "res://inner/main.gd" valid:false  "Preload file \"res://local_dep.gd\" does not exist."
$ gda script validate addons/vendored/vend.gd --project <outer>
{"code":"target_outside_project", "owning_project":"/…/vendor/pkg"}
```

So after both land, two ADRs answer "does this directory belong to the resolved project" in opposite directions for the same directory, with no cross-reference. The scoping paragraph and the `--all` gap paragraph are the right places to say so explicitly.

### P3‑1 — the Red-proofs table and Verification counts are stale relative to this head

Measured at `915a19f9ea`: non-e2e is **2343**, not 2336 (the review-round comment already says 2343); `test_e2e_res_containment.py` is **15** tests, not 13; and *"the ownership probe removed from **both** commands → 3"* predates commit 4, which made it three commands — neutralising `owning_project` now fails **9**. The guards are more load-bearing than claimed, but the table reads as current verification and is not.

### P3‑2 — the `model_fields` guard is only strict for string fields

`test_the_human_channel_renders_every_declared_evidence_field` asserts `len(rendered[3:]) >= len(FailureEvidence.model_fields)`. The `script_errors` block contributes two lines, so the count carries two lines of slack (9 body lines vs 7 fields); a newly added `int`/`float`/enum field the renderer forgets renders nothing and still passes. Only the per-string-value loop is a real guard — which is the same shape of hole the hand-enumerated predecessor had. Assert one labelled line **per set field** instead of a `>=` on the total.

### P3‑3 — the cross-command pin covers containment only

`test_res_containment_consistency.py` drives all three gates from one spelling table, exactly as #763 asked. Ownership — the half commit 4 extended to the third command — has no such row; each gate is guarded only by its own test (1 red for `resource import`, 3 for `script validate`). One parametrized ownership row over the three gates would make the "all three converged" claim drift-proof the same way the containment table does.

### P3‑4 — the ownership walk is O(depth²) syscalls, per batch entry

The stop test calls `directory.resolve()` on every ancestor, and `resolve()` is itself O(depth). Measured (200-call mean, macOS APFS): 0.07 ms at depth 1, 1.04 ms at 20, 7.2 ms at 60, 27.3 ms at 120. Irrelevant for real layouts, but #663 makes it per path. Resolving `start` once and walking the resolved chain in lockstep removes it.

### P3‑5 — the deferred `--all` follow-up is drafted, not filed, and untracked from the surfaces that disclose it

The gap is honestly disclosed on ADR-0006, `command-catalog.md`, `SKILL.md` and four READMEs, and reproduces exactly as described (above). But none of those surfaces names a tracking id and the draft lives only in a PR comment, so a reader who hits the cascade has nothing to follow. File it and cite the number in ADR-0006's "Not closed here" bullet and the catalog paragraph.

qinhaihong-red:
## Review round 2 handled — new head `6c123088`

**This is the last fix of the chain cascade; `6c123088` is the final head.** The branch was
first rebased onto #798's fix head `05d06cc3` (the base had moved from `97feae2a`), then the
findings were handled on the rebased head. Two commits:

- `171bd4fd` — the rebase's own integration debt, below;
- `6c123088` — the review findings.

### Rebase outcome

Four conflicts, all in things the chain beneath changed under this branch:

| conflict | resolution |
|---|---|
| `src/gda/commands/script.py` imports | #796 moved `script_error_line` into `gda.script_errors`; this branch relocates `canonical_res_path` INTO `gda.project`. Both, no `gda.render` import left |
| `src/gda/render.py` `_evidence_lines` | took the base's tri-state `script_errors` branch (#798) and appended the three new fields to it |
| `tests/test_human_failure_output.py` | took the base's guard **wholesale** and dropped this branch's — see P3-2 |
| the three `README.<lang>` i18n markers | re-stamped from the merged English README, twice (once per conflicting commit) |

One thing the rebase surfaced that no conflict marked: #796's `837174d3` added
`_EVIDENCE_PRODUCERS` to `tests/test_error_registry.py`, an AST-read pin on ADR-0004's
recorded evidence-producer boundary. This branch adds two producers, so the pin went red on
the first post-rebase run. That is the pin doing its job — the ADR paragraph and the set
moved together in `171bd4fd`, and the paragraph now states why two refusals that report no
run at all still meet #687's criterion, and why `script run`'s pre-launch escape carries
nothing.

### Findings

| # | claim | decision | evidence | change | validation |
|---|---|---|---|---|---|
| P2-1 | the remediation is not executable for the spelling that produced it | **Fully adopt** | reproduced on the rebased head for all three commands: `script validate inner/main.gd --project <owner>` → `path_not_found`; `script run` → `script_not_found` after a full launch; `script run <abs target_location>` → `invalid_path` | the message names both operands — `--project <owner>` **and** the target respelled relative to it. `gda.project.owner_relative_target` computes it; the docstring's "exactly the pair a caller needs" is corrected to what it is, a derivation | `test_the_refusal_states_a_reissue_that_actually_runs` (parametrized over the three commands) PARSES the re-issue out of the message and runs it; red-proof: old wording → 4 red |
| P2-2 | `resource import` reports `evidence.project_root` unresolved | **Fully adopt** | reproduced: same fixture, `--project /tmp/…` → `resource import` `"/tmp/…"`, `script validate` `"/private/tmp/…"`, against a schema description promising "in its resolved form" | one `root = project_abs.resolve()`, reported by both refusals; the CHECKS keep the spelled value, which is the pattern `_script_validate_recipe` already used | `test_every_gate_reports_the_resolved_project_root`; red-proof → 2 red |
| P2-3 | the symlink edge is decided, changed, and absent from the record | **Fully adopt** | reproduced both halves: directory link at a project → refused, owner = link target; file link at a file in a project → accepted | the split is enumerated in the ADR-0006 amendment, the catalog and the PR body (item 8). The rule is one line and the split is what it says about two cases — only a directory can carry a marker in. The test comment was wrong about the ENVELOPE: it reports `owner.resolve()`, deliberately, so the three published coordinates share one coordinate system; the comment now says that and why, and pins the respelling as where the caller's spelling survives | `test_a_symlinked_in_PROJECT_is_owned_by_it_like_a_nested_one` (comment + new assertion), `test_the_stated_reissue_survives_a_link_spelled_owner` (e2e) |
| P2-4 | wave coherence with #795 | **Fully adopt (matching side)** | both PRs asserted what the shared `res://` walk excludes; #795 (`79ac13d0`) rewrites that walk and its catalog passage | the exclusion sentence is ADR-0032's to own — dropped from ADR-0006, the catalog and the PR body here. ADR-0006 carries the matching layer statement: enumeration is not targeting — the walk decides what the project can ADDRESS by filesystem identity, this gate decides what a caller may NAME from the caller's own spelling. Stated as the shape of the `--all` gap, not as a contradiction | prose; `grep "excludes only"` returns nothing outside the quoted follow-up draft |
| P3-1 | the counts are stale | **Fully adopt** | re-measured at `6c123088` | non-e2e **2375** (not 2336); `test_e2e_res_containment.py` **19** (not 13); ownership neutralised across all three commands **17** (not 3, and more than the reviewer's own corrected 9 — the new cross-command rows raise it); `resource import` reverted to the exact pre-#763 `res://` arm **19** (not 15) | four full-suite reverts, plus four for this round's guards; both tables in the body |
| P3-2 | the `model_fields` guard has 2 lines of slack | **Superseded by the rebase** | #798's `946c5578` landed a strictly stronger pair on the base: `_EVIDENCE_SAMPLES` asserted EQUAL to `FailureEvidence.model_fields`, plus one render per field **alone** — a forgotten branch renders nothing and reds regardless of type | this branch's weaker guard was dropped in the conflict resolution and the three new fields joined the base's sample table | `test_the_sample_table_covers_every_field_the_model_publishes` + `test_every_evidence_field_alone_reaches_the_human_block` (now 8 parametrizations) |
| P3-3 | the cross-command pin covers containment only | **Fully adopt** | each ownership gate was guarded only by its own test | three parametrized rows in `test_res_containment_consistency.py` driving all three gates from one spelling — at the level that BUILDS the envelope, since that is where P2-2 lived. `script run`'s fake launch asserts no engine is reached | red-proof: ownership dropped from `resource import` alone → **6** red, where it used to be 1 |
| P3-4 | the ownership walk is O(depth²) | **Partially adopt** — record, do not rewrite | the reviewer's numbers reproduce as a shape; a real layout is single-digit depth, where the cost is under 0.1 ms against a launch of hundreds | a comment carries the measured curve and the reason it stands: resolving `start` once and walking in lockstep has to keep two chains in step through the symlink case the double reading exists FOR, which is a correctness risk bought with microseconds | none — no behaviour change |
| P3-5 | the `--all` follow-up is drafted, not persistent | **Fully adopt (surface), decline the filing** | the draft lived only in a PR comment | it is now in the PR body's "Follow-ups" section, verbatim, with one correction noted: its "`_should_descend` excludes only `res://.godot`" clause is stale as of #795. Filing stays user triage | reproduces at this head as described |

Two mechanisms considered and **not** adopted, both because they cost more than the sentence:

- **A fourth `FailureEvidence` field for the re-issue target.** It is derivable from two
  coordinates already published (`target_location` relative to `owning_project`) — the exact
  second clause of ADR-0004's criterion for what may enter the object. The shared `error`
  schema is repeated once per command, so that criterion is what stops `gda schema` growing
  by 76 for a value a caller can subtract.
- **Carrying it on the `hint` key.** ADR-0004's #670 amendment bounds `hint` to an
  invocation gda could not RESOLVE, from the curated `hints.py` table, and says so as one of
  two explicit boundaries. This invocation resolved fine and was wrongly rooted.

### Validation at `6c123088`

- `pytest -m "not e2e"`: **2375 passed**. `ruff check` / `ruff format --check` / `pyright`: clean.
- Registry + ADR drift pins green, ADR-0004's evidence-producer pin included.
- Real engine (Godot 4.6.3, per-run `GDA_USER_DATA_ROOT`), run serially:
  `res_containment` 19, `script` 68, `script_run` 49, `resource_import` 9, `project` 26,
  `asset_file` 12, `classname_resolution` 13, `project_walk` 2 — all green.
- Every remediation in this PR's messages was executed, not only asserted.

### Residual risk

- The `--all` cascade and the other path-taking commands (`scene validate`) are unchanged
  and disclosed; the follow-up is drafted, not filed.
- The case-differing project spelling is unchanged, as recorded.
- `--project` on a projectless target widens the Project-code execution surface. Recorded in
  the amendment as a trade, not a free improvement.
- `tests/test_launch.py::test_timeout_synthesizes_a_result_that_keeps_what_the_run_produced`
  flaked once during a red-proof run and passed in every other full run at this head; it is
  #728's known deflake target, unrelated to this branch.

qinhaihong-red:
## Full independent adversarial review — PR #799

Reviewed from first principles at pinned stacked head `6c12308897fa463c13d0ae133dee3fd56a946668` against `feat/685-human-failure-rendering@05d06cc34df1670f0f7f3872bfda55989498752e`. Existing reviews and author responses were excluded.

| Axis | Verdict |
| --- | --- |
| Standards | **Pass** |
| Spec (#763, #697) | **Pass** |

ADR-0031 contradicts itself about upward escapes:

- `docs/adr/0031-headless-script-run-passthrough-execution.md:104-112` says this shape moved out of the seven `invalid_path` refusals and now reports `target_outside_project` through shared containment.
- `:119-126` immediately lists “a path escaping above the root” under what remains `invalid_path`.

That makes the public error ABI undecidable from its authority. Remove the escape item from the stale `invalid_path` list, adjust the count/root-versus-escape explanation, and retain the new `target_outside_project` rule.

Pass. `gda.project` owns canonical `res://` identity, containment, target anchoring, and nearest-project discovery; command modules call those rules without deriving replacement answers. The new target evidence fields satisfy ADR-0004's explicit admission criterion and producer-set update; a further nested object is not justified by current evidence.

### Standards

Pass. Cross-command consistency tests guard the centralized containment and ownership rules; no non-tool-enforced violation or actionable Fowler smell was found.

### Spec

Pass. #763/#697's one containment owner, cross-command parity, preserved `script run`-specific rules, enumerated admission changes, refuse-not-derive ownership decision, batch behavior, executable remediation, new code/evidence, and synchronized decision documents are implemented, apart from the internal ADR contradiction identified on the Artifact axis.

Sources: `CONTEXT.md`; ADR-0002/0004/0006/0031/0032; issues #763/#697; README/translations; catalog; Skill; `PITFALLS.md`; path/error models and tests. Writing reference: ASD-STE100. Containment versus ownership, enumeration versus targeting, lexical versus resolved paths, evidence coordinates, scheme boundaries, remediation, known gaps, and translations were inspected. The ADR contradiction above is the only substantive artifact defect.

Validation: `git diff --check` passed. Current hosted CI has no failed or pending required check, but Godot e2e is skipped. The full real-engine/cross-platform containment matrix and translation generator were not rerun.

qinhaihong-red:

New head: `648c83977` — rebased onto #798's review-fix head `db58e86ae` (stacked cascade), plus one docs commit.

| ADR-0031 contradicts itself: the amendment moves the upward escape to `target_outside_project`, while the "What stays refused" list still carries the escape under `invalid_path` — the error ABI is undecidable from its authority | **Fully adopt** | Both passages verified at the reviewed head | The escape item is removed from the stays-refused list (the six form shapes remain); the verification paragraph now names each shape's code explicitly ("the root as `invalid_path`, the escape as `target_outside_project`") | One `escaping above the root` mention remains in the file (the amendment's own moved-shape paragraph); 2378 non-e2e green on the rebased branch |

Issue #805, opened 2026-09-02
a destroyed float inside a JSON container bypasses the #772 refusal; the packer's float32 elision needs its disclosure decision


qinhaihong-red:
## Parent

Drafted during PR #797's review round (both paths reproduced there at exit 0); filed on user
triage after the P4 promotion (#801).

## What to build

Two write paths still lose a float silently; they need different treatments, and the scope
of each is part of this issue's decision record.

**1. The container path bypasses the refusal — FIX.** #772's rule refuses a scalar
`--value` literal the engine's parser destroys (`0.0` when the literal does not denote
zero, any `NaN` substitution) and disclosed low-order drift. A float INSIDE a JSON
container takes a different coercion path and skips both: measured,
`node set --property d --value '{"a": 1e-320, "b": 0.000000000000000001}'` succeeds and
stores `{"a": 0.0, "b": 0.0}` — the exact failure shape #772 calls unacceptable, alive on
one path. Extend the same admission rule to floats reached through containers — the
`uncoercible_value` message note and the doc statements cover them like scalars. There is
NO per-element traversal to hook: the container coercion is `JSON.parse_string` as the
gate plus one atomic `str_to_var(raw)`, so by the time a float exists the literal is gone.
The raw `--value` spelling is preserved as input and scanned for numeric tokens only AFTER
`JSON.parse_string` accepts it — the gate, not the scan, refuses invalid JSON and
constructor syntax. *(Reworded 2026-09-03: "read before parsing" reversed the delivered
and tested order.)* Three traps, each a
silent wrong answer if missed: a STRING value that looks like a number (`{"a": "1e-320"}`)
must NOT be refused; KEYS are strings and must not be scanned; ~~`str_to_var` accepts richer
Variant syntax than JSON, so a constructor like `Vector2(1e-320, 0)` inside an accepted
dict is reachable~~ **(refuted 2026-09-02: `JSON.parse_string` rejects constructor syntax
before `str_to_var` runs, so the edge is UNREACHABLE — measured, pinned by a test, and
stated in the catalog; see the round-2 note)** — state the chosen scanning rule in the
ADR/catalog. The block lives
inside the byte-identical mirrored region, so whatever is written is duplicated verbatim
into the harness under the existing parity guard.

**2. The packer's float32 elision — DISCLOSURE DECISION ONLY, no verification mechanism.**
The engine omits a script-exported property from the `.tscn` when its value is
[...]

PR #806, opened 2026-09-02
fix(gda): the emission seam's stderr rule reaches the preflight recipe (#803)

### LINKED ISSUE #803: scene preflight still says a raw-stderr failure twice: the emission seam's tee rule does not reach recipe channels
## Parent

Found by the `/improve-codebase-architecture` review of promotion PR #801 (the P4 wave).
Companion to #685 (PR #798), whose emission-seam rule this issue extends to the channels
that do not route through `HeadlessCommand.execute`.

## What to build

#798 decided WHERE the child-stderr forwarding decision lives: the emission seam, which
alone knows the caller's channel, forwards the child's stderr unless the human channel is
about to print the identical bytes as `diagnostics`. That rule protects only the channels
routed through `HeadlessCommand.execute`. The recipe channels each hand-roll a private
stderr policy, and one of them still has the exact duplication #798 fixed, verified live:
`gda scene preflight` tees the launch's stderr unconditionally before classifying, so
EVERY op-reported preflight refusal whose `diagnostics` is the raw stderr duplicates —
`path_not_found` is the reproduced red-proof case (stdout renders the bytes, stderr
carries them again); the engine-quit failure shares the shape but is hard to induce, so
do not red-proof on it. Its comment claims to forward "exactly as HeadlessCommand.execute
does", which has been stale since #798.

Extend the guarantee: a launch-backed recipe attaches the run's stderr to the `Failure` it
builds (the same field `execute` sets) and deletes its local tee; the emission seam it
already answers through does the rest, on both channels. `gda script run` is NOT in scope:
its no-tee design is ADR-0031's promoted Raw run (the streams ARE the result) — record that
boundary beside the rule so the next recipe author learns the whole policy from one place.

Implementation trap, named so it is not re-found: the preflight recipe has FOUR exit
paths (timeout result, ended-before-verdict result-or-Failure, classify_run Failure,
final result). Deleting the one top-of-function tee naively becomes three success tees
plus one `child_stderr` attach — funnel the returns instead, and keep the success-path
forwarding pinned.

## Acceptance criteria

- [ ] A human-mode preflight failure whose diagnostics carry the raw stderr says those bytes once, on the channel the renderer owns; `--json`'s streams stay byte-identical. Red-proof the duplication with a streams-asserted-separately test (`path_not_found` case) before the fix.
- [ ] The preflight-local unconditional tee is deleted; success-path stderr forwarding is preserved.
- [ ] The stale "exactly as HeadlessCommand.execute" comment is corrected to the emission-seam rule.
- [ ] The rule's one home also names the recorded exception (`script run`, ADR-0031) so channel policy is learnable from one place.
- [ ] Schemas, help, and catalog byte-identical — output-channel behavior only.
- [ ] ADR-0002 gains an outcome note recording the adjudicated human failure channel (see the channel-adjudication note below).

## Blocked by

None — can start immediately (assumes the P4 wave content, on `main` once PR #801 merges).

---

> **Channel adjudication (2026-09-02, user decision — deliver the record here):** the
> #685/#798 disposition is KEPT: every gda-classified failure renders for a human on
> **stdout** through the one renderer, the `usage` refusals (`unknown_command` /
> `unknown_option`) included — routing them there is what made the one-renderer invariant
> true and the `hint:` line reachable. The recorded exception is click's own parse errors
> (missing argument, invalid value), which stay click-formatted on **stderr**; `--json`
> envelopes stay on stdout, byte-identical. Deliver this as an **ADR-0002 outcome note**
> (the #798 round-1 review flagged that the human channel became a public output surface
> with no ADR behind it — this closes that too, and gives `CONTEXT.md`'s renderer sentence
> its ADR backing). Reversal was considered and declined: sending usage back to stderr
> resurrects the two-layout/dead-branch defect, and moving ALL human rendering to stderr
> buys convention compliance at the cost of a cross-channel asymmetry with `--json` plus a
> doc/test resync, with no consumer evidence demanding it.

> **Pre-dispatch review (2026-09-02):** the defect statement was widened from the engine-quit
> failure to every op-reported preflight refusal (the quit shape could not be induced live;
> `path_not_found` reproduces trivially), and the four-exit funnel trap is named. Scope
> re-confirmed complete: the preflight tee is the ONLY offending recipe
> (`grep "file=sys.stderr" src/gda/` → the defect plus one unrelated export progress notice).
> The ADR-0002 outcome note edits `docs/adr/0002` only — CONTEXT.md's renderer sentence gets
> its backing by reference, not by edit (CONTEXT.md belongs to #805 in this wave).

qinhaihong-red:
## What

`gda scene preflight` is the one launch-backed recipe that did not answer to the
child-stderr rule #798 put at the emission seam. It calls the launch primitive
directly rather than going through `HeadlessCommand.execute` — because it bifurcates
on the launch's own outcome (#664) — and so kept a private, unconditional tee at the
top of its recipe. Every op-reported refusal whose `diagnostics` IS the raw stderr
therefore said those bytes twice, across two streams.

The recipe's four post-launch exits now funnel through ONE tail in
`run_scene_preflight_operation`: a `Failure` carries the run's stderr as
`child_stderr` to `emit_failure`, which alone knows the caller's channel; a
success-shaped verdict (the final result, the timeout verdict) still tees
immediately, since it has no `diagnostics` block to duplicate and the tee is the
reader's only copy. The bifurcation itself moves into `_preflight_verdict` so the
stderr decision is made once rather than repeated at each of the four returns — the
trap the issue names.

`emit_failure`'s docstring is now the one home of the whole policy: both producers
that attach rather than tee, and the recorded exception — `gda script run`, which
tees nothing because its success result IS the promoted `Raw run` (ADR-0031), so the
child's streams are the operation's published output rather than a diagnostic copy.
The stale `exactly as HeadlessCommand.execute does` comment, wrong since #798, is
corrected to the emission-seam rule.

## Red proof (before the fix)

`tests/test_human_failure_output.py`, streams asserted separately — the defect IS
the split across the two, so a merged capture cannot see it:

```
$ uv run pytest tests/test_human_failure_output.py -q -k preflight

>       assert _PREFLIGHT_STDERR not in result.stderr
E       AssertionError: assert 'gda: runnin.../nope.tscn\n' not in 'gda: runnin.../nope.tscn\n'
E         'gda: running opera...: res://nope.tscn\n' is contained here:
E           gda: running operation: scene-preflight
E           ERROR: scene file not found: res://nope.tscn
tests/test_human_failure_output.py:609: AssertionError

FAILED ...::test_a_preflight_refusal_reaches_a_human_once_not_twice
FAILED ...::test_a_preflight_run_the_project_ended_says_its_stderr_once_too
2 failed, 3 passed, 30 deselected
```

The three that passed are the guards: `--json` keeps its unconditional tee, and both
success-shaped verdicts keep forwarding. After the fix, 5 passed.

## Real-engine before / after

Godot 4.6 on macOS, throwaway project, per-run `--user-data-root` under a fresh
`/tmp` dir.

```
$ gda scene preflight res://nope.tscn --project <proj>          # BEFORE
--- stdout ---
error: path_not_found (operation)
scene file does not exist: res://nope.tscn

gda: running operation: scene-preflight
--- stderr ---
gda: running operation: scene-preflight        <-- the duplicate
```

```
$ gda scene preflight res://nope.tscn --project <proj>          # AFTER
--- stdout ---   (byte-identical to before)
error: path_not_found (operation)
scene file does not exist: res://nope.tscn

gda: running operation: scene-preflight
--- stderr ---
(empty: 0 bytes)
```

Success path, same engine, a real `res://main.tscn`: stdout `ready res://main.tscn`,
stderr still carries `gda: running operation: scene-preflight` — the forwarding is
preserved, and pinned by a parametrized test over both success-shaped exits.

## `--json` byte-identity

Same failing invocation under `--json`, captured before and after against the real
engine and compared with `cmp`:

```
=== JSON byte-identity: stdout ===  IDENTICAL
=== JSON byte-identity: stderr ===  IDENTICAL
```

The envelope alone on stdout, the child's stderr forwarded in full on stderr —
exactly as before.

## ADR-0002 outcome note

`docs/adr/0002` gains `> **Outcome (2026-09-02, #685 / #798 / #803)**`, recording
the adjudicated human failure channel that #798's round-1 review flagged as a public
output surface with no ADR behind it:

- every gda-classified failure renders for a human on **stdout** through the one
  renderer, the `usage` refusals (`unknown_command` / `unknown_option`) included —
  routing them there is what made the one-renderer invariant true and the `hint:`
  line reachable at all;
- the recorded exception is **click's own parse errors** (missing argument, invalid
  value), which stay click-formatted on stderr — gda never classified them, so it
  has no envelope to render; likewise where gda has no correction and no `--json`
  was asked, it declines to answer and the parser's message stands;
- **`--json` is untouched** — the envelope goes to stdout byte for byte;
- **reversal considered and declined**: sending usage back to stderr resurrects the
  two-layout / dead-`hint` defect, and moving all human rendering to stderr buys
  convention compliance at the cost of a cross-channel asymmetry with `--json` plus
  a doc/test resync, with no consumer evidence demanding it.

It closes with the child-stderr rule this PR extends, and names its ADR-0031
exception. `CONTEXT.md` gets its renderer sentence's ADR backing by reference, not by
edit (that file belongs to #805 this wave).

## Surface diff

**None.** Measured, not asserted: `gda schema` and the concatenated `--help` of the
root, all nine groups and `scene preflight` were captured on this branch and on the
base (`git stash` of `src/`), then compared with `cmp` — `IDENTICAL` both times. The
change is output-channel behaviour only; the catalog and SKILL are untouched.

## Gates

- `uv run pytest tests/ -q -m "not e2e"` → **2395 passed**
- `uv run pytest tests/test_e2e_scene_preflight.py -m e2e` → **14 passed** (real engine)
- `uv run ruff check .` / `ruff format --check .` → clean
- `uv run pyright` → 0 errors, 0 warnings

Closes #803

### DIFF
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -212,6 +212,45 @@ the standard build), and they never determine the outcome or a stable code.
 >   meant. This widens no cross-language contract: neither GDScript surface emits or
 >   reads `hint`, and `OperationErrorEnvelope` stays `extra="forbid"`.
 
+> **Outcome (2026-09-02, #685 / #798 / #803) — the HUMAN failure channel, adjudicated.**
+> Everything above specifies the envelope; until #685 that was the whole of what gda
+> emitted, because a caller who passed no `--json` got the serialized envelope anyway.
+> #685 gave the same outcome a second rendering, and the #798 round-1 review flagged
+> that this made the human channel a public output surface with no ADR behind it. This
+> note is the record. It also gives `CONTEXT.md`'s *Error envelope* entry — "Two
+> renderings of ONE outcome, at one exit code, from one renderer" — its ADR backing,
+> by reference rather than by restatement.
+>
+> - **Every gda-classified failure renders for a human on `stdout`, through the one
+>   renderer** (`gda.render.render_failure`, reached only via `gda.headless.emit_failure`),
+>   the `usage` refusals (`unknown_command` / `unknown_option`) included. Routing those
+>   there is what made the one-renderer invariant TRUE rather than approximate, and what
+>   made the `hint:` line reachable at all — it is set nowhere else, so before #798 a
+>   human could never read it.
+> - **The recorded exception is click's own parse errors** (a missing argument, an
+>   invalid value), which stay click-formatted on `stderr`: gda never classified them,
+>   so it has no envelope to render. The same holds where gda recognizes a mistake but
+>   has no correction to add and no `--json` was asked for — it declines to answer and
+>   the parser's message stands. That is gda staying silent, not a second gda layout.
+> - **`--json` is untouched.** The envelope goes to `stdout` as it always did, byte for
+>   byte; the flag chooses the rendering, never the stream.
+> - **Reversal was considered and declined**, twice over. Sending the `usage` refusals
+>   back to `stderr` resurrects the defect #798 closed — two layouts for one category,
+>   and `hint` as a dead branch. Moving ALL human rendering to `stderr` would buy
+>   convention compliance at the cost of a cross-channel asymmetry with `--json`, plus a
+>   documentation and test resync, with no consumer evidence asking for it.
+>
+> One rule follows from the stream choice, and has ONE home
+> (`gda.headless.emit_failure`): a failed CHILD RUN's raw stderr is forwarded to gda's
+> own stderr, EXCEPT when the human channel is about to print those very bytes as
+> `diagnostics` — byte identity decides, so a curated or capped `diagnostics` keeps its
+> tee, and under `--json` the tee is unconditional. Producers attach the stream to the
+> `Failure` and never tee for themselves; #803 brought the last one that did — `scene
+> preflight`, which reaches the launch primitive directly instead of through the shared
+> pipeline — under the rule. Its own recorded exception is `gda script run`, which tees
+> nothing because its success result IS the promoted `Raw run` (ADR-0031): there the
+> child's streams are the operation's published output, not a diagnostic copy of it.
+
 ## `GdaError.code` registry
 
 `GdaError.code` values are a public ABI for agents. Their authoritative source is
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -212,47 +212,51 @@ the standard build), and they never determine the outcome or a stable code.
 >   meant. This widens no cross-language contract: neither GDScript surface emits or
 >   reads `hint`, and `OperationErrorEnvelope` stays `extra="forbid"`.
 
-> **Outcome (2026-09-02, #685 / #798 / #803) — the HUMAN failure channel, adjudicated.**
+> **Outcome (2026-09-02, #685 / #798 / #803) — the human failure channel.**
 > Everything above specifies the envelope; until #685 that was the whole of what gda
 > emitted, because a caller who passed no `--json` got the serialized envelope anyway.
-> #685 gave the same outcome a second rendering, and the #798 round-1 review flagged
-> that this made the human channel a public output surface with no ADR behind it. This
-> note is the record. It also gives `CONTEXT.md`'s *Error envelope* entry — "Two
+> #685 gave the same outcome a second rendering, and the #798 round-1 review noted that
+> this made the human channel a public output surface with no ADR behind it. This note
+> is the record, and it is the one authority for the channel and child-stderr rules
+> below: the implementation sites state only their local behavior and reference this
+> note (#806 review). It also gives `CONTEXT.md`'s *Error envelope* entry — "Two
 > renderings of ONE outcome, at one exit code, from one renderer" — its ADR backing,
 > by reference rather than by restatement.
 >
-> - **Every gda-classified failure renders for a human on `stdout`, through the one
->   renderer** (`gda.render.render_failure`, reached only via `gda.headless.emit_failure`),
->   the `usage` refusals (`unknown_command` / `unknown_option`) included. Routing those
->   there is what made the one-renderer invariant TRUE rather than approximate, and what
->   made the `hint:` line reachable at all — it is set nowhere else, so before #798 a
->   human could never read it.
-> - **The recorded exception is click's own parse errors** (a missing argument, an
->   invalid value), which stay click-formatted on `stderr`: gda never classified them,
->   so it has no envelope to render. The same holds where gda recognizes a mistake but
->   has no correction to add and no `--json` was asked for — it declines to answer and
->   the parser's message stands. That is gda staying silent, not a second gda layout.
-> - **`--json` is untouched.** The envelope goes to `stdout` as it always did, byte for
->   byte; the flag chooses the rendering, never the stream.
-> - **Reversal was considered and declined**, twice over. Sending the `usage` refusals
->   back to `stderr` resurrects the defect #798 closed — two layouts for one category,
->   and `hint` as a dead branch. Moving ALL human rendering to `stderr` would buy
->   convention compliance at the cost of a cross-channel asymmetry with `--json`, plus a
+> - Every gda-classified failure renders for a human on `stdout`, through the one
+>   renderer (`gda.render.render_failure`, reached only via `gda.headless.emit_failure`).
+>   This includes the `usage` refusals (`unknown_command` / `unknown_option`). Routing
+>   them through the renderer makes the one-renderer statement exact and makes the
+>   `hint:` line reachable: it is set nowhere else, so before #798 a human could not
+>   read it.
+> - The recorded exception is click's own parse errors (a missing argument, an invalid
+>   value), which stay click-formatted on `stderr`: gda did not classify them, so it has
+>   no envelope to render. The same holds where gda recognizes a mistake but has no
+>   correction to add and no `--json` was asked for: gda declines to answer and the
+>   parser's message stands.
+> - `--json` is unchanged. The envelope goes to `stdout`, byte for byte; the flag
+>   chooses the rendering, never the stream.
+> - Two reversals were considered and declined. Sending the `usage` refusals back to
+>   `stderr` reintroduces the defect #798 closed (two layouts for one category, `hint`
+>   as a dead branch). Moving all human rendering to `stderr` buys convention
+>   compliance at the cost of a cross-channel asymmetry with `--json` plus a
 >   documentation and test resync, with no consumer evidence asking for it.
 >
-> One rule follows from the stream choice, and has ONE home
-> (`gda.headless.emit_failure`): a failed CHILD RUN's raw stderr is forwarded to gda's
-> own stderr, EXCEPT when the human channel is about to print those very bytes as
-> `diagnostics` — byte identity decides, so a curated or capped `diagnostics` keeps its
-> tee, and under `--json` the tee is unconditional. On a FAILURE producers attach the
-> stream to the `Failure` and never tee it for themselves; #803 brought the last one
-> that did — `scene preflight`, which reaches the launch primitive directly instead of
-> through the shared pipeline — under the rule. A SUCCESS is the other half and stays
-> with the producer: it carries no `diagnostics` for a tee to duplicate, so each
-> forwards the stream itself, at once, that copy being the reader's only one. The
-> rule's own recorded exception is `gda script run`, which tees nothing because its
-> success result IS the promoted `Raw run` (ADR-0031): there the child's streams are
-> the operation's published output, not a diagnostic copy of it.
+> The stream choice implies one child-stderr rule, recorded here:
+>
+> - On a failure, the producer that classified a child run attaches the raw stderr to
+>   the `Failure` (`child_stderr`) and does not print it. `emit_failure` forwards it to
+>   gda's own stderr, except when the human channel prints the same bytes as
+>   `diagnostics`; byte identity decides, and under `--json` the forward is
+>   unconditional.
+> - On a success, the producer forwards the stream immediately: a success carries no
+>   `diagnostics` block to duplicate, so that copy is the reader's only one.
+> - Producers today: `HeadlessCommand.execute` (the sentinel and live pipelines) and
+>   the `scene preflight` recipe, the last private tee, brought under the rule by #803.
+>   A launch-backed channel gda adds joins them by attaching rather than printing.
+> - Exception: `gda script run` neither attaches nor forwards. Its success result is
+>   the promoted `Raw run` (ADR-0031) — the child's streams are the published output —
+>   and its failures publish the captures as curated `diagnostics` sections.
 
 ## `GdaError.code` registry
 
--- a/docs/adr/0002-headless-structured-output-contract.md
+++ b/docs/adr/0002-headless-structured-output-contract.md
@@ -244,12 +244,15 @@ the standard build), and they never determine the outcome or a stable code.
 > (`gda.headless.emit_failure`): a failed CHILD RUN's raw stderr is forwarded to gda's
 > own stderr, EXCEPT when the human channel is about to print those very bytes as
 > `diagnostics` — byte identity decides, so a curated or capped `diagnostics` keeps its
-> tee, and under `--json` the tee is unconditional. Producers attach the stream to the
-> `Failure` and never tee for themselves; #803 brought the last one that did — `scene
-> preflight`, which reaches the launch primitive directly instead of through the shared
-> pipeline — under the rule. Its own recorded exception is `gda script run`, which tees
-> nothing because its success result IS the promoted `Raw run` (ADR-0031): there the
-> child's streams are the operation's published output, not a diagnostic copy of it.
+> tee, and under `--json` the tee is unconditional. On a FAILURE producers attach the
+> stream to the `Failure` and never tee it for themselves; #803 brought the last one
+> that did — `scene preflight`, which reaches the launch primitive directly instead of
+> through the shared pipeline — under the rule. A SUCCESS is the other half and stays
+> with the producer: it carries no `diagnostics` for a tee to duplicate, so each
+> forwards the stream itself, at once, that copy being the reader's only one. The
+> rule's own recorded exception is `gda script run`, which tees nothing because its
+> success result IS the promoted `Raw run` (ADR-0031): there the child's streams are
+> the operation's published output, not a diagnostic copy of it.
 
 ## `GdaError.code` registry
 
--- a/src/gda/commands/scene.py
+++ b/src/gda/commands/scene.py
@@ -1046,6 +1046,11 @@ def run_scene_preflight_operation(
     - **the op reported a verdict** → the public result, with the script errors gda
       read off stderr (#651's parser, the single home of that reading) and the
       CLI-resolved project added to the engine's answer.
+
+    Those four outcomes are decided by :func:`_preflight_verdict` and returned
+    through the ONE tail below, which is what lets this channel answer to the same
+    stderr rule as every other one (#803): a failure carries the launch's stderr to
+    the emission point, a verdict forwards it here.
     """
     run_launch = make_launch or launch
     try:
@@ -1071,13 +1076,35 @@ def run_scene_preflight_operation(
         timeout=params.timeout,
         timeout_label="Godot scene preflight",
     )
-    # Forward the engine's own stream, exactly as the sentinel channel does
-    # (``HeadlessCommand.execute``): a preflight's stderr is where the booted scene's
-    # verbatim complaints are, and this command must not be the one that swallows
-    # them.
-    if raw.stderr:
+    outcome = _preflight_verdict(raw, params, binary, root)
+    # The engine's own stream, forwarded on exactly ONE channel — the same rule
+    # ``HeadlessCommand.execute`` follows since #798, which this recipe did not
+    # inherit because it does not go through it (#803). A preflight's stderr is
+    # where the booted scene's verbatim complaints are, so it must reach the reader;
+    # but a ``Failure`` classified from this run carries that stream as its
+    # ``diagnostics``, and teeing it HERE would say the same bytes twice, across two
+    # streams, for every op-reported refusal. So a failure carries it to
+    # ``emit_failure``, which alone knows the caller's channel and decides there; a
+    # verdict has no diagnostics block to duplicate, so its tee is immediate and
+    # this is the reader's only copy.
+    if isinstance(outcome, Failure):
+        outcome.child_stderr = raw.stderr
+    elif raw.stderr:
         print(raw.stderr, end="", file=sys.stderr)
+    return outcome
 
+
+def _preflight_verdict(
+    raw: RunResult,
+    params: ScenePreflightParams,
+    binary: Path,
+    root: "Path | None",
+) -> "ScenePreflightResult | Failure":
+    """The four-way bifurcation of a finished preflight launch (#664).
+
+    Split out of :func:`run_scene_preflight_operation` so its four exits funnel
+    through one tail there rather than each repeating the stderr decision (#803).
+    """
     diagnostics = parse_script_errors(raw.stderr)
     if raw.launch_failure is LaunchFailure.TIMEOUT:
         # The evidence the launch already measured, which this verdict used to drop
--- a/src/gda/commands/scene.py
+++ b/src/gda/commands/scene.py
@@ -1077,7 +1077,7 @@ def run_scene_preflight_operation(
         timeout_label="Godot scene preflight",
     )
     outcome = _preflight_verdict(raw, params, binary, root)
-    # The engine's own stream, forwarded on exactly ONE channel — the same rule
+    # The engine's own stream, forwarded once for a human — the same rule
     # ``HeadlessCommand.execute`` follows since #798, which this recipe did not
     # inherit because it does not go through it (#803). A preflight's stderr is
     # where the booted scene's verbatim complaints are, so it must reach the reader;
--- a/src/gda/commands/scene.py
+++ b/src/gda/commands/scene.py
@@ -1048,9 +1048,9 @@ def run_scene_preflight_operation(
       CLI-resolved project added to the engine's answer.
 
     Those four outcomes are decided by :func:`_preflight_verdict` and returned
-    through the ONE tail below, which is what lets this channel answer to the same
-    stderr rule as every other one (#803): a failure carries the launch's stderr to
-    the emission point, a verdict forwards it here.
+    through the one tail below, which is what lets this channel answer to the
+    child-stderr rule of ADR-0002's #803 note: a failure carries the launch's
+    stderr to the emission point, a verdict forwards it here.
     """
     run_launch = make_launch or launch
     try:
@@ -1077,16 +1077,10 @@ def run_scene_preflight_operation(
         timeout_label="Godot scene preflight",
     )
     outcome = _preflight_verdict(raw, params, binary, root)
-    # The engine's own stream, forwarded once for a human — the same rule
-    # ``HeadlessCommand.execute`` follows since #798, which this recipe did not
-    # inherit because it does not go through it (#803). A preflight's stderr is
-    # where the booted scene's verbatim complaints are, so it must reach the reader;
-    # but a ``Failure`` classified from this run carries that stream as its
-    # ``diagnostics``, and teeing it HERE would say the same bytes twice, across two
-    # streams, for every op-reported refusal. So a failure carries it to
-    # ``emit_failure``, which alone knows the caller's channel and decides there; a
-    # verdict has no diagnostics block to duplicate, so its tee is immediate and
-    # this is the reader's only copy.
+    # A failure carries the launch's stderr to the emission seam, which alone knows
+    # the caller's channel; a verdict has no diagnostics block to duplicate, so its
+    # tee is immediate and this is the reader's only copy. The rule these two
+    # branches implement is ADR-0002's #803 outcome note (#806 review).
     if isinstance(outcome, Failure):
         outcome.child_stderr = raw.stderr
     elif raw.stderr:
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -77,13 +77,10 @@ class Failure:
     ``child_stderr`` is the raw stderr of the child run this failure classifies,
     attached by its producer instead of being teed there — whether printing it
     would say the same bytes twice depends on the caller's channel, which only the
-    emission point knows (#798 review). Two producers attach it:
-    :meth:`gda.headless.HeadlessCommand.execute` for the sentinel and live
-    pipelines, and ``scene preflight``'s recipe, which reaches the launch primitive
-    directly (#803). The rule they both answer to, and its one recorded exception,
-    are documented at that emission point (:func:`gda.headless.emit_failure`). It
-    stays ``""`` on every failure no child run produced, and it is not part of the
-    serialized envelope.
+    emission point (:func:`gda.headless.emit_failure`) knows (#798 review). The
+    full rule — producers, the success half, and the exception — is ADR-0002's
+    #803 outcome note. It stays ``""`` on every failure no child run produced, and
+    it is not part of the serialized envelope.
     """
 
     error: GdaError
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -75,10 +75,14 @@ class Failure:
     """A classified failure: the stable error shape plus its process exit code.
 
     ``child_stderr`` is the raw stderr of the child run this failure classifies,
-    attached by :meth:`gda.headless.HeadlessCommand.execute` instead of being teed
-    there — whether printing it would say the same bytes twice depends on the
-    caller's channel, which only the emission point knows (#798 review). It stays
-    ``""`` on every failure no child run produced, and it is not part of the
+    attached by its producer instead of being teed there — whether printing it
+    would say the same bytes twice depends on the caller's channel, which only the
+    emission point knows (#798 review). Two producers attach it:
+    :meth:`gda.headless.HeadlessCommand.execute` for the sentinel and live
+    pipelines, and ``scene preflight``'s recipe, which reaches the launch primitive
+    directly (#803). The rule they both answer to, and its one recorded exception,
+    are documented at that emission point (:func:`gda.headless.emit_failure`). It
+    stays ``""`` on every failure no child run produced, and it is not part of the
     serialized envelope.
     """
 
--- a/src/gda/headless.py
+++ b/src/gda/headless.py
@@ -655,15 +655,26 @@ def emit_failure(failure: Failure, *, json_output: bool) -> NoReturn:
     key set, so a record does not read differently depending on which half of the
     contract carried it (:class:`gda.models.FailureEvidence`).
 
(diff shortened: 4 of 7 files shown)


qinhaihong-red:
## Full independent adversarial review — PR #806

**Pinned head:** `75a94eeab` (`fix/803-recipe-stderr-guarantee`) · **Base:** `feat/gda-dogfooding-dev` @ `64136719c`
**Note on scope:** `64136719c` is *not* an ancestor of the head (merge-base `fe72700ec`), so the reviewed change is the single commit `75a94eeab` — 5 files, +249/−19. The `libs/gda-balancing` churn in a naive `git diff base head` is base-side and out of scope.

Reviewed from first principles at the pinned head. The PR body, existing comments, and author responses were excluded from the evidence; every claim below rests on a run I made myself against the real engine (Godot `4.6.3.stable.official.7d41c59c4`) or on the source at the pinned head.

### Verdict

| Axis | Verdict | Findings |
| --- | --- | --- |
| **Artifact** (ADR-0002 note, docstrings, comment, #803 consistency) | Changes required | 1× P2, 1× P3 |
| **Architecture** (locality of the tee rule, `_preflight_verdict`) | Pass | 1× P3 (optional) |
| **Standards** (repo standards, actionable Fowler smells) | Pass | — |
| **Spec** (every #803 AC) | Pass — all 6 AC met | 2× P3 (coverage) |

**Overall: approve after the P2 one-sentence correction.** The defect is real, the fix is correct, and I could not break it on any path I could construct. No P1.

---

### P2

#### P2-1 — the rule's "one home" states the failure half of the policy and denies the success half, which is the half that regresses silently

`src/gda/headless.py:667-673` claims totality, then states a rule that is false as written:

> This is the ONE home of gda's child-stderr policy, and **the whole of it is learnable here**. Two producers attach the field, and **neither tees for itself** … A launch-backed channel gda adds joins them by **attaching rather than printing**.

The same overreach is mirrored into the ADR at `docs/adr/0002-…md:247-248`: "Producers attach the stream to the `Failure` and **never tee for themselves**".

Both producers *do* tee for themselves, on the success path, and it is load-bearing:

- `src/gda/headless.py:836-837` — `if result.stderr: print(result.stderr, end="", file=sys.stderr)`
- `src/gda/commands/scene.py:1092-1093` — `elif raw.stderr: print(raw.stderr, end="", file=sys.stderr)`

**Evidence that this is not pedantry.** I mutated the head to drop *only* the success tee (`elif raw.stderr: print(...)` deleted, attach kept) — precisely the code a new recipe author would write if they took "attaching rather than printing" at face value:

```
2 failed, 33 passed
FAILED tests/test_human_failure_output.py::test_a_preflight_verdict_still_forwards_the_engines_stderr[the-verdict]
FAILED tests/test_human_failure_output.py::test_a_preflight_verdict_still_forwards_the_engines_stderr[the-timeout-verdict]
```

and on the real engine that mutation silently swallows the engine's only copy of its own stderr — the exact failure the pre-#803 comment existed to prevent ("this command must not be the one that swallows them"). `tests/test_e2e_scene_preflight.py:100` (`assert "hero ready" in preflighted.stderr`) is the real-engine pin of the behavior the docstring tells a future author to omit.

**Impact.** The AC asks that "channel policy is learnable from one place". A reader who learns the policy only from that place learns it wrong, and the resulting bug is invisible on the failure paths every test in this PR covers.

**Smallest fix** — one clause at `headless.py:669` and the matching one at ADR line 248, e.g.:

> Two producers attach the field on a **failure** and never tee it there; on a **success** — which has no `diagnostics` block to duplicate — each still forwards the child's stderr itself, that being the reader's only copy.

---

### P3

#### P3-1 — `"forwarded on exactly ONE channel"` uses `channel` in the one sense the codebase does not, and is false under `--json`

`src/gda/commands/scene.py:1080`. Two problems in one clause:

1. **Terminology.** In this project `channel` means the *caller's* channel — human vs `--json`. CONTEXT.md's *Error envelope* entry, `emit_failure`'s "on the channel the caller asked for", and the ADR note's own "cross-channel asymmetry with `--json`" all use it that way; so does line 1087 of this very comment ("which alone knows the caller's **channel**"). Line 1080 uses it to mean *stream*, while lines 1085-1086 correctly say "across two **streams**". One term, two meanings, eleven lines apart.
2. **Accuracy.** Under `--json` the bytes deliberately reach *both* streams. Verified at head:

```
$ gda --json scene preflight res://nope.tscn …
stdout: {"error":{…,"diagnostics":"gda: running operation: scene-preflight\n"}}
stderr: gda: running operation: scene-preflight
```

**Fix:** open with "forwarded once for a human — the same rule …". The rest of the comment is accurate and does correct the stale "exactly as `HeadlessCommand.execute` does" as the AC required.

#### P3-2 — the `child_stderr` attach is unpinned on the `_ended_before_the_verdict` exit; the funnel is the only thing holding it

The funnel is right, but the tests do not hold it. I mutated the head so the attach is skipped on exactly one of the two failure exits:

```python
if isinstance(outcome, Failure):
    if outcome.error.code != "operation_failed":   # skips the ended-before-verdict exit
        outcome.child_stderr = raw.stderr
```

Result: **`2395 passed, 638 deselected`** — the entire non-e2e suite is green with the attach silently lost on that path. Under `--json` that mutation drops the engine's stderr from gda's stderr entirely, which is the one thing the AC says must stay byte-identical.

The gap is asymmetric coverage: `test_the_same_preflight_refusal_under_json_keeps_its_stderr_tee` exists only for the `path_not_found` (`classify_run`) exit; the `_ended_before_the_verdict` exit is covered in human mode only, where `diagnostics == child_stderr` makes a missing attach unobservable. **Fix:** parametrize the existing `--json` test over both failing `RunResult`s already defined in the file (`exit_code=1` + error sentinel, and `exit_code=0` + empty stdout) — roughly four lines.

#### P3-3 — the human-mode no-duplication has no real-engine pin

The red proof monkeypatches `gda.commands.scene.launch`, which is fine for the unit. But `tests/test_e2e_scene_preflight.py:200` already spawns the real engine for exactly this `path_not_found` case and asserts only the `--json` shape, so the defect this PR fixes has no automated real-engine regression cover. RULES.md is explicit that a fake must not be the sole evidence of real integration behavior. Adding `assert preflighted.stderr == ""` to a human-mode variant of that existing test costs one subprocess. (I did verify the behavior on the real engine by hand — see Validation — so this is a durability gap, not a correctness doubt.)

#### P3-4 — one documented rule, two implementations (optional)

`src/gda/headless.py:832-837` and `src/gda/commands/scene.py:1090-1093` are now the same four-line decision written twice, while the ADR and the docstring both insist the rule has ONE home. `scene.py` already imports from `gda.headless` (line 37), so a shared `forward_child_stderr(outcome, stderr)` is layering-feasible and would make the code home match the documented one — and would have killed the P3-2 mutation for free. **Counter-argument, stated honestly:** two call sites for a four-line helper is thin justification, and #803 did not ask for it. Raise only if a third launch-backed recipe appears. I do not consider this blocking.

---

### Axis notes

**Artifact.** The ADR-0002 outcome note is quote-faithful to the adjudication recorded on the issue: human failure channel = stdout through the one renderer *including* the `usage` refusals with the one-renderer/`hint`-reachability rationale; click's own parse errors recorded as the exception on stderr; `--json` untouched; reversal declined with both recorded reasons (usage-back-to-stderr resurrects the two-layout/dead-branch defect; all-to-stderr buys convention at the cost of cross-channel asymmetry plus a doc/test resync, no consumer evidence). I verified all three behavioral claims live and the historical claim in the source (see Validation) — nothing is asserted that the code does not do. The added sentence about gda *declining to answer* a hintless refusal is not in the issue text but is accurate against `hints.py:265-288`. The CONTEXT.md sentence it quotes is verbatim at `CONTEXT.md:244`, and it is referenced rather than restated, as the pre-dispatch note required. The `emit_failure` docstring does name the `script run` / ADR-0031 exception, and that exception is true: `grep` finds no `sys.stderr` write and no `child_stderr` in `src/gda/commands/script.py`. Findings: P2-1, P3-1.

**Architecture.** The fix concentrates the rule correctly and does **not** leak channel knowledge into the recipe — `run_scene_preflight_operation` never reads `json_output`; it only classifies its outcome as `Failure`-or-not, which is channel-agnostic. The resulting tail is a structural mirror of `HeadlessCommand.execute:828-837`, so the two launch-backed producers now have one shape, which is the right outcome. `_preflight_verdict` is a well-drawn seam: a pure decision over `(raw, params, binary, root)` with the emission concern lifted out, converting four scattered returns into one decision plus one tail. Its signature and quoted `"Path | None"` annotation match the neighbouring `_ended_before_the_verdict`. No speculative abstraction, no new extension point. Only P3-4 (optional).

**Standards.** `ruff check` clean, `ruff format --check` clean (196 files), `pyright` 0 errors/0 warnings. Conventional-commit subject and body are accurate to what shipped — I checked each claim in the message against the diff and found no overstatement. `Closes #803` present. No actionable Fowler smell: nothing added is duplicated logic beyond P3-4, the extracted function has a single reason to change, and the comments explain *why* rather than restating the code.

**Spec — every #803 AC.**

| AC | Verdict | Evidence |
| --- | --- | --- |
| Bytes said once on the renderer's channel; `--json` streams byte-identical; red-proofed on `path_not_found` with streams asserted separately | ✅ | Live repro at base + absence at head; `cmp` on both streams; M1 turns the red proof red |
| Preflight-local unconditional tee deleted; success-path forwarding preserved | ✅ | Tee gone at `scene.py:1092`; success forwarding verified live and pinned (M2 → red) |
| Stale "exactly as `HeadlessCommand.execute`" comment corrected | ✅ (see P3-1) | `scene.py:1080-1089` |
| Rule's one home names the `script run` / ADR-0031 exception | ✅ (see P2-1) | `headless.py:670-679`; exception verified in `commands/script.py` |
| Schemas, help, catalog byte-identical | ✅ | `gda schema` identical (1,038,261 B both); 85 `--help` screens identical; `docs/command-catalog.md` untouched by the commit |
| ADR-0002 outcome note recording the adjudication | ✅ (see P2-1) | `docs/adr/0002-…md:215-253` |

The four-exit funnel trap the issue named is handled: one decision in `_preflight_verdict`, one tail, no duplicated stderr decision. The `unresolvable_binary_failure` early return correctly stays outside the funnel — no child run, so `child_stderr=""` is right.

---

### Validation

All runs are mine, at the pinned head/base, in detached scratch worktrees with `UV_PROJECT_ENVIRONMENT` pointed at per-tree venvs and a fresh `--user-data-root` per invocation. `pytest -p no:cacheprovider` throughout.

**Real-engine reproduction (base `64136719c`), human mode, streams captured separately:**
```
$ gda scene preflight res://nope.tscn --project … ; exit 4
stdout: error: path_not_found (operation)
        scene file does not exist: res://nope.tscn
        (blank)
        gda: running operation: scene-preflight     <-- diagnostics
stderr: gda: running operation: scene-preflight     <-- the recipe's tee (the defect)
```
**Same invocation at head `75a94eeab`:** stdout identical, **stderr empty**, exit 4. Said once.

**Byte-identity, base vs head, `cmp` on both streams:**
| Invocation | stdout | stderr |
| --- | --- | --- |
| `--json` failure (`path_not_found`) | identical | identical |
| `--json` success (`ready`) | identical | identical |
| human success (`ready`) | identical | identical (`gda: running operation: scene-preflight`) |
| `gda schema` | identical (1,038,261 B) | — |
| 85 `--help` screens (`COLUMNS=80 NO_COLOR=1`) | identical | — |

**Suites at head:** `tests/test_human_failure_output.py` 35 passed · full non-e2e `2395 passed, 638 deselected` · **real-engine** `tests/test_e2e_scene_preflight.py` **14 passed** (16.2 s). Gates: ruff check / ruff format / pyright all clean.

**Mutation testing** (each applied to the head, suite re-run, then reverted; `git status` clean afterwards):
| # | Mutation | Result |
| --- | --- | --- |
| M1 | Restore the unconditional tee | **RED** — 2 failed (both failure exits) |
| M2 | Drop the success-path tee | **RED** — 2 failed (both verdict shapes) |
| M3 | Drop the `child_stderr` attach | **RED** — 1 failed (the `--json` test) |
| M4 | Drop the attach on the `_ended_before_the_verdict` exit only | **SURVIVES** — 2395 passed → P3-2 |

**ADR claims verified against running code, not read:**
- Hinted usage refusal → stdout through the one renderer with the `hint:` line, exit 2 (`gda scene inspect …`).
- Hintless refusal + no `--json` → gda declines; click's message on stderr, stdout empty (`gda zzzznope`) — matches `hints.py:_refuse`.
- Click parse error → click-formatted on stderr, stdout empty (`gda scene preflight` with no PATH).
- Pre-#685 behavior: `git show ac0b2cf35^:src/gda/headless.py` → `def emit_failure(failure)` with no channel parameter, docstring "Emit a structured error envelope to stdout". The ADR's historical claim holds.
- Scope re-verified independently: `grep -rn "sys.stderr" src/gda/` → 5 hits. Child-stderr routes are `headless.py:682` (the rule's home), `headless.py:837` and `scene.py:1093` (the two success tees). `export.py:574` is a progress notice; `mcp/__init__.py:40` is unrelated. No launch-backed channel still tees a failure's stderr for itself.

**Terminology and prose pass** (required; `artifact-review`). Sources: `CONTEXT.md`, `docs/adr/0002`, `docs/adr/0031`, issue #803. Writing reference: ASD-STE100 per RULES.md. Scope inspected: the complete 39-line ADR note, the `emit_failure` and `Failure` docstrings, the `scene.py` recipe comment and `_preflight_verdict` docstring, and all 151 added test lines. Checks: one-term-one-meaning against the project glossary, causal/scope/ordering accuracy, technical noun and verb use, no second name for an existing concept. Result: one terminology collision (P3-1, `channel`) and one scope-accuracy overreach in two locations (P2-1). Prose is otherwise concrete and free of formulaic wording; "declined, twice over" reads oddly for a count of two options but is disambiguated by the sentences that follow and is not worth changing.

**Not performed:** no Windows or Linux run (stream routing here is platform-independent Python, and `--log-file`/user-data placement is unchanged); no `gda-mcp` surface check (the change touches no schema — verified byte-identical); the engine-quit failure shape was exercised only through the injected launch seam, as #803 itself directs.

qinhaihong-red:
## Review round handled — 4 of 5 adopted, 1 declined

**New head:** `58a76639e` (`docs(gda): the child-stderr rule states its success half too`) · previous head `75a94eeab`.

Every claim was re-verified against the pinned head before deciding — the two mutations you reported were re-run here independently, and the `--json` two-stream behaviour was re-observed on the real engine (Godot `4.6.3`) rather than read off the source.

| Claim | Decision | Evidence (mine) | Change | Validation |
| **P2-1** the rule's "one home" states only the failure half and denies the success half | **Fully Adopt** | Confirmed: `headless.py:837` and `scene.py:1093` both tee on success; the docstring's "the whole of it is learnable here" + unqualified "neither tees for itself" is false as written | `headless.py` docstring + `docs/adr/0002:247-252` — the attach-don't-tee rule scoped to a **FAILURE**, with what a **SUCCESS** does said next to it | M2a (drop the recipe's success tee) → **2 failed**; M2b (drop `execute`'s) → **18 failed** |
| **P3-1** `channel` used for `stream`, and false under `--json` | **Fully Adopt** | Both halves confirmed. Live at head: `--json` puts `diagnostics` on stdout **and** the raw stream on stderr — the bytes reach both streams by design | `scene.py:1080` → "forwarded once for a human" | `--json` bytes unchanged (below) |
| **P3-2** the `child_stderr` attach is unpinned on the `_ended_before_the_verdict` exit | **Fully Adopt** | M4 reproduced independently: **2395 passed, 638 deselected** with the attach dropped on that exit alone | `test_the_same_preflight_refusal_under_json_keeps_its_stderr_tee` → `test_either_preflight_failure_under_json_keeps_its_stderr_tee`, parametrized over both failing runs | M4 re-applied → **RED** on `[the-run-the-project-ended]`; reverted → green. Suite 2395 → **2396** |
| **P3-3** the human-mode no-duplication has no real-engine pin | **Fully Adopt** | Agreed, and it is the load-bearing half: the suppression fires on byte identity between `diagnostics` and the raw stream, and the unit authors **both** sides at the injected launch seam | One extra invocation in the existing `test_a_missing_scene_is_refused_not_reported_as_a_verdict` — `assert human.stderr == ""` as you proposed | M1 (restore the deleted tee) → **RED** on the real engine; `-m e2e` preflight suite **14 passed** |
| **P3-4** one documented rule, two implementations | **Do Not Adopt** | — | none | see below |

### On P3-4

Declined, on your own counter-argument plus one more. Two call sites for a four-line decision is thin, `#803` did not ask for it, and the two sites are not free-standing duplicates: each sits inside a different control flow (`execute` returns the outcome, the recipe falls through to a shared `return`), so a mutating `forward_child_stderr(outcome, stderr)` would add an indirection between a producer and a two-line decision without removing the decision. The "one home" the docs insist on is the home of the **policy** (`emit_failure`), which is already single and now states both halves.

The incidental argument for it — that it would have killed M4 for free — is answered more directly by the P3-2 test, which pins the **contract** rather than the code shape and would survive a later re-inlining. Your trigger is recorded as stated: raise it again if a third launch-backed recipe appears.

### Validation at `58a76639e`

| Gate | Result |
| --- | --- |
| `pytest -m "not e2e" -p no:cacheprovider` | **2396 passed**, 638 deselected (+1, the new parametrized case) |
| `pytest -m e2e tests/test_e2e_scene_preflight.py` | **14 passed** (17.4 s), real engine `4.6.3` |
| `ruff check` / `ruff format --check` | clean / 485 files already formatted |
| `gda schema` | **byte-identical** (1,038,261 B — same figure you measured) |
| `--help` screens (`COLUMNS=80 NO_COLOR=1`, 10 sampled) | **byte-identical** (24,704 B) |
| `--json` preflight failure, both streams | **byte-identical** (stdout 174 B, stderr 40 B) |

Byte-identity was measured working-tree-vs-`75a94eeab` in one pass, not against a remembered figure.

### Mutation re-runs

| # | Mutation | Before this round | After |
| --- | --- | --- | --- |
| M1 | restore the unconditional tee | RED (unit) | RED (unit **and** now real-engine e2e) |
| M2a | drop the recipe's success tee | RED — 2 failed | unchanged; now also **documented** as pinned |
| M2b | drop `execute`'s success tee | *(not previously measured)* | RED — 18 failed |
| M4 | drop the attach on the ended-before-verdict exit only | **SURVIVES** — 2395 passed | **RED** — `[the-run-the-project-ended]` |

M2b was run because the corrected docstring now claims *both* tees are pinned; the claim is measured, not asserted.

### Not verified

No Windows or Linux run — the stream routing is platform-independent Python and the diff adds no platform-sensitive path. The `_ended_before_the_verdict` exit is still exercised only through the injected launch seam, as #803 directs.

qinhaihong-red:
## Exact-head adversarial review — PR #806

**Reviewed base tip:** `64136719c6a615ab94887aeeb3447f895511744e`  
**Reviewed head:** `58a76639ee84eae35de0ab625c14cfaeec08aeb9`  
**Three-dot merge base:** `fe72700ecb6d12b24474560c4b8599266df6ebcd`

The base tip is not an ancestor of the head, so I reviewed `git diff 64136719c...58a76639e` using three-dot merge-base semantics and inspected the exact-head files with `git show`. I also read issue #803 in full. The behavioral fix is correct, but the current head creates conflicting authority and prose-maintenance problems around the policy it fixes.

### Artifact review — Changes required (2 findings)

I reviewed the complete issue #803 and the complete current ADR-0002, plus the changed prose in the `Failure`, `emit_failure`, preflight, and test docstrings/comments where it supports the document's claims.

1. **Required — the ADR records a second full authority while claiming the rule has one home.** `docs/adr/0002-headless-structured-output-contract.md:243-255` reproduces the failure suppression algorithm, producer behavior, success behavior, and `script run` exception, while saying that the rule has one home in `gda.headless.emit_failure`. The exact same policy is independently editable in `src/gda/headless.py:658-681`; producer-count details are repeated in `src/gda/errors.py:77-86` and the recipe decision is restated in `src/gda/commands/scene.py:1080-1089`. Evidence already exists that this drifts: the preceding review had to update both the ADR and docstring after the success half was omitted. This undermines the document's DRY claim and makes the next producer/exception easy to document inconsistently. **Smallest fix:** keep ADR-0002 as the durable decision and scope, keep `emit_failure`'s docstring to its local suppression algorithm, and replace producer-count/global-policy copies with one-way references. If the intended authority is instead the code docstring, make the ADR reference it rather than restating the full operational rule.

2. **Required — the added technical prose obscures the normative stream rules.** The outcome note and related docstrings use emphatic/rhetorical forms such as “the HUMAN failure channel, adjudicated,” “made the ... invariant TRUE,” “declined, twice over,” “resurrects the defect,” “the whole of it is learnable here,” and repeated `ONE` / `ALL` / `FAILURE` / `SUCCESS` (`ADR-0002:215-255`, `headless.py:668-681`, `scene.py:1050-1053,1080-1089`). These formulations make it harder to distinguish requirements from rationale and conflict with `RULES.md`'s plain, practical technical-writing rule. **Smallest fix:** rewrite these passages as short declarative statements naming the actor, stream, condition, and exception; preserve the decision and rationale.

**Terminology and prose assessment.** Sources: `CONTEXT.md` (especially *Headless launch*, *Raw run*, *Error envelope*), accepted ADR-0002 and ADR-0031, issue #803, and `RULES.md`. Writing reference: ASD-STE100 Simplified Technical English, as required by `RULES.md`. Scope: every line of issue #803 and ADR-0002; every changed requirement, decision, docstring, comment, and test explanation; checks covered one-term/one-meaning, technical nouns and verbs, actor/action clarity, causal and scope relations, and procedural ordering. Terms such as `Error envelope`, `Raw run`, `Headless launch`, `diagnostics`, stream, and channel otherwise agree with the project language. The substantive DRY and prose findings are above. I did not run an automated ASD-STE100 checker; the assessment was manual.

### Domain modular architecture — Changes required (1 finding)

Authority sources: `CONTEXT-MAP.md`, `CONTEXT.md`, accepted ADR-0000/0002/0010/0023/0031/0040, issue #803, and the exact-head implementation/tests. The observed control flow is otherwise sound: the scene inbound command stays in its group-owned vertical slice; `_preflight_verdict` owns the four-way recipe decision; the recipe attaches child stderr to the shared `Failure`; `emit_failure` owns channel-aware emission. Source dependencies stay directed (`commands.scene` → shared headless/errors/runner), and no new reciprocal import or speculative extension point appears.

**Required — knowledge ownership is not single or extensible.** The global policy is independently enumerated in ADR-0002, `emit_failure`, `Failure`, and the scene recipe. This fails the architecture's DRY check even though runtime ownership is correctly placed: a new launch-backed recipe changes the producer set and can require edits across unrelated owners. It also makes the claimed extension path (“a launch-backed channel gda adds joins them”) depend on prose synchronization rather than the owned contract. **Smallest fix:** select one durable decision authority, keep only local invariants beside each implementation, and use one-way references for the global rule/exception set.

Completeness: pass; all four post-launch outcomes and the pre-launch binary refusal have owners. Orthogonality: pass for runtime behavior; verdict classification and channel emission remain distinct. DRY: fail as above. Extensibility: fail only at the policy-documentation seam above; the code path itself remains a small extension seam. DDD: no new bounded context, aggregate, repository, or domain service is warranted.

`DM-806-1` — claim-evidence state: `open` (the project defines no stronger claim-evidence vocabulary). Claim: one owner can govern child-stderr policy as producers evolve. Evidence against: four independently editable policy descriptions. Disconfirming condition: one authority plus references or a consistency mechanism. Open validation need: no automated guard currently checks these prose copies.

### Standards — Changes required (2 findings)

1. **Hard — the change creates several independently editable authorities for the child-stderr policy.** `docs/adr/0002-headless-structured-output-contract.md:243-255` says the rule has “ONE home” in `emit_failure`, while the full producer/success/exception policy is repeated in `src/gda/headless.py:658-681`; `src/gda/errors.py:77-86` separately hard-codes that there are “Two producers”; and `src/gda/commands/scene.py:1080-1089` restates the routing decision again. This violates `RULES.md`'s one-authority rule: copies must be generated, one-way references, or consistency-tested. The producer count and exception can drift when another launch-backed recipe is added; no check keeps these prose copies synchronized. Make ADR-0002 the decision authority, keep `emit_failure`'s docstring to the local suppression algorithm, and have `Failure`/`scene` reference the ADR without enumerating the global producer set.

2. **Hard — changed technical prose does not follow the repository's plain-English rule.** The ADR and docstrings rely on rhetorical/legal phrasing and emphatic capitals: “the HUMAN failure channel, adjudicated”, “made the ... invariant TRUE”, “Reversal was ... declined, twice over”, “resurrects the defect”, “the whole of it is learnable here”, and repeated `ONE`/`ALL`/`FAILURE`/`SUCCESS` (`ADR-0002:215-255`, `headless.py:668-680`, `scene.py:1080-1089`). `RULES.md` requires ASD-STE100 as the reference and prohibits formulaic wording, slogans, legal language, and abstract prose. These formulations obscure the normative stream rules. Rewrite as short declarative requirements: who writes which stream, when suppression applies, and the `script run` exception.

No independent Fowler smell beyond the documented DRY breach; diff check is clean. **Count: 2 hard Standards findings.**

### Spec — Pass (0 findings)

At the pinned head, the implementation satisfies issue #803:

- The four preflight exits funnel through `_preflight_verdict`; failures attach `raw.stderr` as `Failure.child_stderr`, while success-shaped verdicts retain immediate stderr forwarding.
- Human `path_not_found` and ended-before-verdict failures suppress duplicate raw stderr; JSON failures retain it. Tests cover both streams independently, both failure exits, and the real-engine `path_not_found` case.
- The stale local comment describes the emission-seam decision, and `emit_failure` names the `gda script run` / ADR-0031 exception.
- ADR-0002 records the adjudicated stdout human-failure channel, Click parse-error exception, unchanged JSON channel, rejected reversals, failure tee rule, success half, and `script run` boundary.
- No unasked behavior or scope creep was found. Exact base/head comparisons found no diff in `gda schema`, root `--help`, or `docs/command-catalog.md`.

### Validation

- Exact-head `tests/test_human_failure_output.py`: **36 passed**.
- Exact-head real-engine `tests/test_e2e_scene_preflight.py`: **14 passed** with Godot available locally.
- Changed files: `ruff check` passed; `ruff format --check` passed; `pyright --pythonpath <review venv>` reported **0 errors, 0 warnings**.
- Independent Spec-axis base/head probes: `gda schema`, root `--help`, and `docs/command-catalog.md` were byte-identical.
- Current CI at the reviewed head: all required reported checks completed successfully; hosted **Godot e2e tests are skipped**.
- Not fully verified locally: the complete non-e2e suite did not finish within the bounded review capture, so the completed GitHub unit/package job is the full-suite evidence. I did not run Windows/Linux stream-routing checks; this diff adds no platform-specific branch. No automated prose-consistency or ASD-STE100 gate exists.

**Axis summary:** Artifact 2 findings (worst: conflicting authority); Domain modular architecture 1 finding (DRY/extension ownership); Standards 2 hard findings (worst: conflicting authority); Spec 0 findings.

qinhaihong-red:
## Round-2 response

New head: `95c555eb1` (one docs/comments commit on the reviewed `58a76639e`).

| The child-stderr policy is enumerated in four independently editable homes while each claims one home; drift already observed within this PR's own rounds (Artifact-1 / DM-806-1 / Standards-1) | **Fully adopt** | The four passages verified at head; RULES.md's one-authority rule | ADR-0002's outcome note is now the one record of the whole rule (streams, suppression, producers, success half, `script run` exception) and says so; `emit_failure`'s docstring states only its local suppression algorithm + a one-way reference; `Failure.child_stderr` states field semantics + reference (the producer roster is gone); the preflight comment states its two local branches + reference | 67 focused + 2396 unit green; docs/comments only — all outputs byte-identical |
| Rhetorical phrasing and argumentative capitals obscure the normative rules (Artifact-2 / Standards-2) | **Partially adopt** | The cited passages verified | Every rewritten passage (the ADR note and the three docstrings/comments) is now short declarative statements — "adjudicated", "TRUE", "declined, twice over", "resurrects", "the whole of it is learnable here" and the argumentative caps are gone. NOT adopted: a repo-wide de-capitalization — functional caps emphasis outside the rewritten passages is the codebase's established comment idiom and out of this PR's scope | The rewritten note reads as actor/stream/condition/exception statements; decision and rationale preserved |

Spec stood at Pass; no behavior changed in this round.

qinhaihong-red:

Reviewed base tip `64136719c6a615ab94887aeeb3447f895511744e`, merge base `fe72700ecb6d12b24474560c4b8599266df6ebcd`, and current head `95c555eb1a2b61b4f0a73ef76e146e10a2de1d32`. I treated the author reply as a claimed-fix list and verified the current artifacts and code independently.

**Required change — `src/gda/render.py:84-91` still independently owns and has already drifted from the child-stderr policy.** The `render_failure` docstring restates attachment, forwarding, byte identity, curated-tail, and `--json` behavior, but names only `HeadlessCommand.execute`. ADR-0002 now explicitly identifies `scene preflight` as the second producer. This also contradicts ADR-0002's statement that implementation sites describe only their local behavior.

Smallest fix: reduce this docstring to the renderer's local responsibility—render diagnostics when present and avoid cross-stream duplication at the emission seam—and refer to ADR-0002's #803 outcome note for the global routing and suppression policy.

**Terminology and prose assessment:** I read issue #803, the complete current ADR-0002, and all changed documentation, docstrings, and comments. I used `CONTEXT.md`, ADR-0031, and `RULES.md` as the terminology and authority sources, with ASD-STE100 as the writing reference. The revised prose is clear and the previous prose problem is no longer present. The only remaining substantive prose problem is the inaccurate policy copy above. No automated ASD-STE100 or policy-authority consistency checker was available.

### Domain modular architecture — Changes required

Runtime ownership and dependency direction now pass: classification and channel emission remain separate, and the four post-launch exits plus pre-launch refusal have owners. However, the presentation owner still contains global emission-policy knowledge in `render.py`, and the copy already omits one producer. The DRY knowledge-ownership and documentation-extensibility checks therefore still fail.

One hard finding remains: `render.py:84-91` is an independently editable and already divergent copy of a rule whose authority is ADR-0002, contrary to `RULES.md`'s one-authority requirement. No additional actionable Fowler smell was found.

### Spec — Pass

Issue #803's four-exit funnel, failure attachment, success forwarding, human duplicate suppression, JSON forwarding, `script run` exception, and public-contract stability remain satisfied.

### Validation

- Focused human-failure and real-Godot preflight tests: `50 passed`.
- Full non-E2E suite: `2394 passed, 2 skipped, 638 deselected`.
- Ruff check/format and Pyright: passed; `git diff --check`: passed.
- Merge-base/head schema, root help, and command catalog: byte-identical.
- Current CI: `failed=[]`, `pending=[]`; hosted Godot E2E is skipped, while the affected preflight path passed locally with the real engine.

Not independently checked: Windows/Linux execution and all 85 rendered subcommand help pages. The current delta is documentation/docstring/comment-only.

qinhaihong-red:
## Round-3 response

New head: `07b86cde4` (one docs commit on the reviewed `95c555eb1`).

| `render.py`'s docstring is a fifth independently editable copy of the child-stderr policy, already drifted (names only `execute` as attacher) — contradicting ADR-0002's local-statements-only rule (all three axes) | **Fully adopt** | Verified at the cited lines; the drift is real — the paragraph predates #803's second producer | The paragraph now states only the renderer's local fact (it renders what the envelope carries; the emission point prevents the cross-stream duplicate) and references ADR-0002's #803 note. Swept for further attacher-enumeration copies: none remain outside the ADR | 67 focused tests green; ruff/format clean; docs-only |

Spec stood at Pass.

PR #807, opened 2026-09-02
refactor(gda): one containment gate on gda.project, three call sites folded (#802)

### LINKED ISSUE #802: gda.project owns the containment gate, not only its primitives
## Parent

Found by the `/improve-codebase-architecture` review of promotion PR #801 (the P4 wave). Companion to #763/#697 (PR #799), which unified the containment PRIMITIVES this issue's orchestration composes.

## What to build

Make `gda.project` own the containment GATE, not only its primitives. Today THREE
call sites write the orchestration by hand — two full copies (`script validate`,
`resource import`) and one ownership-only copy (`script run`, whose address gate
has already decided containment before it asks): ask ownership first (the more specific diagnosis), build the
foreign-owner refusal with its four-coordinate evidence shape (`target_location`, resolved
owner, resolved root, `owner_relative_target`), then ask containment and build the
outside-root refusal. The ordering rule, the argument shapes, and the `.resolve()`
discipline are interface cost every call site pays — and the cross-gate consistency test
exists precisely because these copies drifted once before (#763's own postmortem).

Lift that composition into one gate: the ordered DECISION on `gda.project`
(`containment_violation` — ordering, normalization, coordinates; no envelope), mapped to
the two refusals by one `gda.errors.containment_refusal` the commands call. *(Reshaped
2026-09-02: the original "gate function on `gda.project` returning `Failure`" inverted
ADR-0040 §5's dependency direction — foundation must not import the taxonomy — and needed
a deferred import to hide the cycle; see the round-2 note below.)* All three call sites shrink
to one call each; the consistency test — which already drives all three gates plus the
shared scoping helper — survives as the outer guard over the gate's output rather than
the only thing holding the copies together. Folding the `script run` site into the full
gate is expected to be behavior-identical (its address gate refuses every escaping
spelling first, and a non-escaping `res://` is never outside the root) — VERIFY that
with the consistency table, do not assume it. The gate adopts the cwd-absolutized
project normalization (`resource import`'s #738 form); the two current spellings were
measured to answer identically on relative `--project` forms, so pin that with a
relative-`--project` row in the consistency table. Behavior-preserving: both
refusal envelopes, their codes, evidence, and remediation stay byte-identical.

## Acceptance criteria

- [ ] The ordered ownership-before-containment DECISION has one owner in `gda.project` (`containment_violation`), and one `gda.errors` function (`containment_refusal`) maps it to both refusal envelopes; neither command module builds a containment refusal itself. *(Reworded 2026-09-03 to the two-owner split the round-2 note records — the original one-function-on-`gda.project` wording required the ADR-0040 direction violation.)*
- [ ] The three current call sites call the gate; their local copies are deleted (the `script run` fold verified behavior-identical via the consistency table, including a relative-`--project` row).
- [ ] The cross-gate consistency test still passes unchanged (it now guards the gate's output).
- [ ] Refusal envelopes are byte-identical before/after on all three commands (pin with a regression or assert via the existing e2e), with ONE accepted widening — the complete literal-tilde handling for a relative project directory literally named `~<no-such-user>`, measured per command:
  | command / target | base | head |
  |---|---|---|
  | `script validate`, outside or inside | `RuntimeError`, exit 1 | structured refusal (exit 4) / served (`valid: true`) |
  | `resource import`, outside or inside | `RuntimeError`, exit 1 | structured refusal (exit 4) / served (dry-run result) |
  | `script run`, ordinary inside path | succeeds | succeeds (unchanged) |
  | `script run`, escaping spelling | lexical refusal pre-resolution | unchanged |
  Both changed halves regression-pinned (the pins fail under a swap back to `Path.expanduser()`). *(Widened in full 2026-09-03, round 4; stated as this measured matrix in round 5 — the earlier "EVERY path" wording overclaimed: `script run`'s ordinary paths never crashed, because its ownership-only copy expanded the project only inside the foreign-owner branch.)*
- [ ] Schemas, help, and catalog byte-identical — this is internal orchestration only.

## Blocked by

None — can start immediately (assumes the P4 wave content, on `main` once PR #801 merges).

---

> **Pre-dispatch review (2026-09-02):** the original text undercounted the call sites (two)
> and AC2 contradicted AC1 — following "two" literally leaves `script run`'s hand-built
> ownership refusal in a command module AC1 forbids. Corrected in place to THREE sites, with
> the fold's behavior-identity to be verified, not assumed, and the gate's project
> normalization named.

---

> **Round-2 reconciliation (2026-09-02, PR #807):** three corrections from the exact-head
> re-review, all verified before adoption.
>
> 1. **The prescribed shape violated accepted ADR-0040.** §4 keeps the `*_failure`
>    taxonomy in `gda.errors`; §5 fixes the direction `errors/models → foundation`. A gate
>    on `gda.project` returning `Failure` created `project → errors → models → project`,
>    hidden by a deferred import. Delivered shape: the pure ordered decision stays with the
>    path authority (`gda.project.containment_violation`, returning the fired half and its
>    coordinates), and `gda.errors.containment_refusal` maps it to the envelopes — commands
>    still make one call; ordering and path knowledge stay single-sourced; the direction
>    holds with no deferred import.
> 2. **One measured behavior widening is accepted, not hidden:** the gate's total
>    `_expand_user` normalization turns the base's `RuntimeError` leak on a literal
>    `~<no-such-user>` project directory into the structured refusal. AC4 now names it;
>    a regression pins it.
> 3. AC4's "both commands" corrected to "all three" (the body and AC1/AC2 already said
>    three).

qinhaihong-red:
## What

`gda.project` owned the containment **primitives**; three command sites owned the **composition** over them. This lifts that composition into one gate, split per ADR-0040 §5 (round-2 review): the ordered DECISION on the authority, its envelopes mapped by the taxonomy:

```python
gda.project.containment_violation(target, project) -> ForeignOwnerViolation | OutsideRootViolation | None
gda.errors.containment_refusal(target: str, project: Path | None) -> Failure | None  # what commands call
```

plus `gda.project.project_absolute(project)` beside the decision — the shared cwd-absolutized normalization it adopts (`resource import`'s #738 form), still called directly by `resource import` because it also maps an accepted path back onto `res://` afterwards.

The gate's docstring records what was interface cost at every call site: the **ordering rule** (ownership first — the more specific diagnosis, and the only half a projectless call can make), the four-coordinate evidence shape, the `.resolve()` discipline, and the normalization it adopts.

## Why

The ordering rule, the argument shapes and the `.resolve()` discipline were paid three times, and the copies had already drifted twice — #763's own postmortem (`resource import` refused a net-inside `..`, and read `\\` as a filename character) and #799's (`resource import` reported the project as SPELLED where the two script gates reported it RESOLVED). `tests/test_res_containment_consistency.py` existed precisely to catch that; it now guards **one function's output** instead of being the only thing holding three copies together.

## The three sites

| Site | Before | After |
| --- | --- | --- |
| `_script_validate_recipe` (`src/gda/commands/script.py`) | full copy: ownership → `target_owned_by_another_project_failure`, containment → `target_outside_project_failure`, per batch entry | `containment_refusal(path, project)` per batch entry; first offender still refuses the whole batch |
| `run_script_run_operation` (`src/gda/commands/script.py`) | **ownership-only** copy (its address gate has already decided containment) | `containment_refusal(script, project)` — the whole gate |
| `_asset_res_path` (`src/gda/commands/resource.py`) | full copy, plus its own project normalization and resolved `root` | `containment_refusal(raw, project)`; keeps `project_absolute` only for the `res://` mapping below it |

Neither command module imports `target_outside_project_failure` or `target_owned_by_another_project_failure` any more.

### The `script run` fold, verified not assumed

The claim is that adding the containment half to the ownership-only site changes nothing. It is a theorem, not a sample: `_project_scoped_res_path` returns **only** a canonical `res://` address whose `res_escape_remainder` is `None` — every other spelling is refused there first — and that is exactly the input `path_outside_project` answers `None` for on its lexical `res://` branch. So the folded gate can only ever return what the ownership-only copy returned.

`test_the_containment_half_script_run_folded_in_is_inert` drives that spelling by spelling over the shared `CONTAINMENT` table **plus** the shapes that stay `script run`'s own (`res://`, `res://.`, the trailing-`strip_edges` suffix, the log-line boundary, `user://`), asserting `path_outside_project` is inert for every address that survives. Red-proofed: deleting the address gate's escape clause fails 5 of its rows.

## Byte-identity proof

46 cases driven through the real CLI (`python -m gda … --json`) against a **fixed** fixture tree, at the base commit `fe72700ec` and at this head, then `diff`ed:

```
$ diff base.txt head.txt && echo "FINAL: 46/46 envelopes byte-identical"
FINAL: 46/46 envelopes byte-identical
```

Cases: foreign-owned target (nested project) × 3 commands × {project-relative, `res://`, absolute} spellings; outside-root target × {`res://../`, `res://..\\`, `res://a/../../`, `res://..`, absolute fs path, relative fs path}; **relative `--project`** (`outer`, `./outer`, `alias/../outer`) × 3 commands; projectless call on an owned target; a batch whose second entry offends; and three accepted controls that must stay accepted (`script validate bar.gd`, `resource import res://pic.png`, `resource import res://foo/../pic.png`). 43 refusals + 3 successes, each captured with its exit code, its `--json` body and its stderr.

Sample row (relative `--project`, `script run`), identical on both sides:

```json
{
  "error": {
    "category": "operation",
    "code": "target_outside_project",
    "diagnostics": "",
    "evidence": {
      "owning_project": "/private/tmp/gda802-fixture/outer/inner",
      "project_root": "/private/tmp/gda802-fixture/outer",
      "target_location": "/private/tmp/gda802-fixture/outer/inner/main.gd"
    },
    "message": "/private/tmp/gda802-fixture/outer/inner/main.gd belongs to the Godot project /private/tmp/gda802-fixture/outer/inner, but this call resolved the resolved project /private/tmp/gda802-fixture/outer: its own res:// references would resolve against the wrong root, so nothing was run. Pass --project /private/tmp/gda802-fixture/outer/inner and address the target as 'main.gd' to work on it in the project that owns it."
  }
}
```

## Surface diff: **none**

```
$ gda schema --json | diff base-schema.json -   →  no output
$ gda [--help | <group> --help] ×9 | diff base-help.txt -  →  no output
sha256 base-schema.json = head-schema.json = 27663f01e3e559b21c9202b6db4708789d3b9743b001b5f90ac47bfb49e527f9
sha256 base-help.txt    = head-help.txt    = 7e6be531bedee7136b1245c8b75d236897133b84695175b6f9c33282727587a4
```

Expected, and confirmed: no params model, result model, descriptor, error code or option changed — this is internal orchestration. `docs/command-catalog.md` is untouched for the same reason.

## Tests

`tests/test_res_containment_consistency.py` keeps every existing assertion **unchanged** (it now guards the gate's output) and gains NINE test functions across the four rounds, each red-proofed against a deliberate mutation — among them the both-halves-fire ordering proof (`test_ownership_wins_when_both_halves_of_the_gate_fire`, the discriminating pin for the ordering rule), the two literal-tilde pins (refusal AND inside-target service, both red under a swap back to `Path.expanduser()`), the mapper-purity AST assertion, and:

| New pin | Red-proof |
| --- | --- |
| `test_the_containment_half_script_run_folded_in_is_inert` | delete `_project_scoped_res_path`'s escape clause → 5 rows fail |
| `test_every_gate_answers_a_relative_project_spelling_identically` | anchor `project_absolute` anywhere but the cwd → both rows fail |
| `test_the_three_gates_now_emit_ONE_envelope` | restore an ownership-only copy in `script run` with a drifted re-issue → both rows fail |
| `test_no_command_module_builds_a_containment_refusal_itself` (AST over the whole `gda.commands` package) | call a builder from `resource.py` again → fails |
| `test_the_gate_is_where_both_refusals_are_built` (AST over `gda.errors.containment_refusal`, which must also call the decision and build nothing else) | swap one builder for the other → fails |

**Honest note on the relative-`--project` row:** it does not discriminate between the as-spelled and cwd-absolutized normalizations, and cannot — they genuinely agree, which is the measurement that let the gate adopt one of them. What the row holds is the **agreement**: a normalization that anchors somewhere other than the cwd fails the row. A normalization that RESOLVES before the probes is caught by a different test — `test_the_gate_reads_the_project_as_spelled_rather_than_pre_resolved`, added in round 2 exactly because the row cannot see it (round-5 correction: an earlier version of this note attributed both mutations to the row).

## Verification

- `uv run pytest tests/ -q -m "not e2e"` → **2420 passed, 0 skipped** on the author machine (2390 at merge base); the reviewer's sandbox reports **2418 passed, 2 skipped** — the same set, two environment-gated tests skipping there
- `uv run pytest tests/test_e2e_res_containment.py -m e2e` → **19 passed**, real Godot 4.6.3
- `uv run pytest tests/test_e2e_script.py tests/test_e2e_script_run.py tests/test_e2e_resource.py -m e2e` → **148 passed**, real engine
- `uvx ruff check` / `uvx ruff format --check` → clean (197 files)
- `uv run pyright` (the repo's own src + tests scope, what CI runs) → 0 errors, 0 warnings

## One decision beyond the issue text

**(Superseded in round 2.)** The gate first lived whole on `gda.project` and imported the two builders inside the function to hide a `project → errors → models → project` cycle; the round-2 review found that inverts ADR-0040 §5 and the split above replaced it — no deferred import remains. The normalization knock-on became an ACCEPTED, pinned widening (round 3): a project directory literally named `~<no-such-user>`, spelled RELATIVE, raised `RuntimeError` (exit 1) at the merge base and now gets the structured `target_outside_project` refusal (exit 4); the regression discriminates (it fails under a swap back to `Path.expanduser()`), and issue #802's AC4 records the widening. Round 4 measured that the same normalization changes ACCEPTED paths too — an inside target under the relative tilde project crashed identically at the base and is now served by `script validate` and `resource import` — so the widening is authorized WHOLE on #802's AC4 and pinned on both halves. The change is behavior-preserving on every path except that complete, named, accepted literal-tilde case.

Closes #802

### DIFF
--- a/src/gda/commands/resource.py
+++ b/src/gda/commands/resource.py
@@ -28,8 +28,9 @@ from pydantic import BaseModel, Field, model_validator
 from gda.binary import resolve_godot_binary
 from gda.dispatch import dispatch_domain, dispatch_recipe, params_or_bad_parameter
 from gda.errors import (
-    Failure,
     classify_launch_or_crash,
+    containment_refusal,
+    Failure,
     make_failure,
 )
 from gda.execution import ExecutionKind
@@ -50,7 +51,6 @@ from gda.models import (
 from gda.project import (
     RES_PREFIX,
     canonical_res_path,
-    containment_refusal,
     is_engine_virtual_path,
     project_absolute,
     project_anchored,
--- a/src/gda/commands/resource.py
+++ b/src/gda/commands/resource.py
@@ -31,8 +31,6 @@ from gda.errors import (
     Failure,
     classify_launch_or_crash,
     make_failure,
-    target_outside_project_failure,
-    target_owned_by_another_project_failure,
 )
 from gda.execution import ExecutionKind
 from gda.headless import (
@@ -52,12 +50,10 @@ from gda.models import (
 from gda.project import (
     RES_PREFIX,
     canonical_res_path,
+    containment_refusal,
     is_engine_virtual_path,
-    owner_relative_target,
-    owning_project,
-    path_outside_project,
+    project_absolute,
     project_anchored,
-    target_location,
 )
 from gda.render import render_property_lines, render_set_echo
 from gda.runner import launch
@@ -834,28 +830,17 @@ def _asset_res_path(project: Path, raw: str) -> "str | Failure":
         )
     # One coordinate system for BOTH sides before any comparison: a relative
     # --project must not meet an absolute candidate (#738 re-review 2 — the
-    # mixed comparison raised a bare ValueError on the lexical fallback).
-    project_abs = project.expanduser()
-    if not project_abs.is_absolute():
-        project_abs = Path.cwd() / project_abs
-    # The checks read the project as SPELLED (the double reading is theirs to make);
-    # the refusals report it RESOLVED, which is what `FailureEvidence.project_root`
-    # publishes ("in its resolved form — the same value a successful result
-    # reports") and what the two sibling gates already did. Reporting `project_abs`
-    # itself made this one command answer `/tmp/...` where `script validate`
-    # answered `/private/tmp/...` for the same call (#799 review).
-    root = project_abs.resolve()
-    owner = owning_project(raw, project_abs)
-    if owner is not None:
-        return target_owned_by_another_project_failure(
-            target_location(raw, project_abs),
-            owner.resolve(),
-            root,
-            owner_relative_target(raw, project_abs, owner),
-        )
-    outside = path_outside_project(raw, project_abs)
-    if outside is not None:
-        return target_outside_project_failure(outside, root)
+    # mixed comparison raised a bare ValueError on the lexical fallback). The
+    # gate below normalizes the same way; this local is for the res:// MAPPING
+    # further down, which needs the same absolute root.
+    project_abs = project_absolute(project)
+    # ONE call for ownership-then-containment and both refusal envelopes (#802).
+    # What used to stand here — the two probes, their order, the four coordinates
+    # and the resolved root each refusal reports — now lives on ADR-0006's path
+    # authority, so this gate keeps only what is genuinely about ASSETS.
+    refusal = containment_refusal(raw, project)
+    if refusal is not None:
+        return refusal
     if raw.startswith(RES_PREFIX):
         # Already in the engine's namespace: canonicalize the spelling and stop.
         # Reading it lexically is what the engine does too — a symlink inside the
--- a/src/gda/commands/resource.py
+++ b/src/gda/commands/resource.py
@@ -830,9 +830,11 @@ def _asset_res_path(project: Path, raw: str) -> "str | Failure":
         )
     # One coordinate system for BOTH sides before any comparison: a relative
     # --project must not meet an absolute candidate (#738 re-review 2 — the
-    # mixed comparison raised a bare ValueError on the lexical fallback). The
-    # gate below normalizes the same way; this local is for the res:// MAPPING
-    # further down, which needs the same absolute root.
+    # mixed comparison raised a bare ValueError on the lexical fallback). This
+    # local is for the res:// MAPPING further down, which needs the same absolute
+    # root the gate below reads. Not a two-places-must-agree coupling: there is one
+    # normalization RULE — `project_absolute` — and both sites call it, so the only
+    # cost of asking twice is a second `Path.cwd()` (#807 review).
     project_abs = project_absolute(project)
     # ONE call for ownership-then-containment and both refusal envelopes (#802).
     # What used to stand here — the two probes, their order, the four coordinates
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -64,6 +64,7 @@ from gda.project import (
     RES_PREFIX,
     canonical_res_path,
     containment_refusal,
+    project_absolute,
     res_escape_remainder,
 )
 from gda.runner import LaunchFailure, LaunchFn, RunResult, launch
@@ -2180,7 +2181,11 @@ def _script_validate_recipe(
     built once by the caller, identical on the argv and ``--params-json`` paths
     (ADR-0015).
     """
-    root = None if project is None else project.expanduser().resolve()
+    # The SUCCESS path's root, normalized the way the refusal path's is (#807
+    # review): both sides of one verdict must reach the project by one reading, and
+    # the same call answering two spellings of one root is what #799 was.
+    # `project_absolute` differs only in staying total on an unresolvable `~user`.
+    root = None if project is None else project_absolute(project).resolve()
     for path in params.paths:
         # ONE call for both halves and their ordering (#802): the gate on ADR-0006's
         # path authority owns them, so this recipe states only WHICH targets it is
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -31,9 +31,10 @@ from gda import dispatch
 from gda.binary import resolve_godot_binary
 from gda.dispatch import dispatch_domain, dispatch_recipe, params_or_bad_parameter
 from gda.errors import (
-    Failure,
     classify_launch_or_crash,
     classify_run,
+    containment_refusal,
+    Failure,
     make_failure,
     script_did_not_run_failure,
     script_escapes_project_failure,
@@ -63,7 +64,6 @@ from gda.models import (
 from gda.project import (
     RES_PREFIX,
     canonical_res_path,
-    containment_refusal,
     project_absolute,
     res_escape_remainder,
 )
@@ -2147,7 +2147,7 @@ def _script_validate_recipe(
 
     The refusal has TWO halves since ADR-0006's 2026-08-31 amendment (#697), and
     the second is why a *projectless* call is now checked too. Both are asked by
-    ONE call to :func:`~gda.project.containment_refusal` (#802), which owns their
+    ONE call to :func:`~gda.errors.containment_refusal` (#802), which owns their
     ordering and builds whichever envelope fires; this recipe only chooses the
     targets. Containment (:func:`~gda.project.path_outside_project`) asks whether
     the target is in the resolved project's tree, which only a resolved project
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -2147,9 +2147,9 @@ def _script_validate_recipe(
 
     The refusal has TWO halves since ADR-0006's 2026-08-31 amendment (#697), and
     the second is why a *projectless* call is now checked too. Both are asked by
-    ONE call to :func:`~gda.errors.containment_refusal` (#802), which owns their
-    ordering and builds whichever envelope fires; this recipe only chooses the
-    targets. Containment (:func:`~gda.project.path_outside_project`) asks whether
+    ONE call to :func:`~gda.errors.containment_refusal` (#802), which maps the
+    ordered decision :func:`~gda.project.containment_violation` makes to whichever
+    envelope fires; this recipe only chooses the targets. Containment (:func:`~gda.project.path_outside_project`) asks whether
     the target is in the resolved project's tree, which only a resolved project
     can fail. Ownership (:func:`~gda.project.owning_project`) asks whether that
     project is really the target's OWNER — a ``project.godot`` nearer to the
--- a/src/gda/commands/script.py
+++ b/src/gda/commands/script.py
@@ -42,8 +42,6 @@ from gda.errors import (
     script_run_aborted_failure,
     script_run_project_not_found_failure,
     script_run_timeout_failure,
-    target_outside_project_failure,
-    target_owned_by_another_project_failure,
     termination_phase,
     unresolvable_binary_failure,
 )
@@ -65,11 +63,8 @@ from gda.models import (
 from gda.project import (
     RES_PREFIX,
     canonical_res_path,
-    owner_relative_target,
-    owning_project,
-    path_outside_project,
+    containment_refusal,
     res_escape_remainder,
-    target_location,
 )
 from gda.runner import LaunchFailure, LaunchFn, RunResult, launch
 from gda.script_errors import (
@@ -1564,16 +1559,19 @@ def run_script_run_operation(
     # address gate above is lexical, so it cannot see a `project.godot` nested
     # between the resolved root and the script. Running against the outer root
     # would resolve the script's own `res://` references against a root that is
-    # not its own — GDA-DF-035 in the executing form. One rule, two commands: the
-    # same probe `script validate` applies, reported under the same code.
-    owner = owning_project(script, project)
-    if owner is not None:
-        return target_owned_by_another_project_failure(
-            target_location(script, project),
-            owner.resolve(),
-            project.expanduser().resolve(),
-            owner_relative_target(script, project, owner),
-        )
+    # not its own — GDA-DF-035 in the executing form. One rule, three commands:
+    # the same shared gate `script validate` and `resource import` apply, reported
+    # under the same code.
+    #
+    # This site asked ownership ALONE until #802, because the address gate above
+    # has already decided containment for it — every escaping spelling is refused
+    # there, and a non-escaping res:// address is inside the root by construction.
+    # It now calls the whole gate anyway: the containment half is inert here (the
+    # consistency table proves it spelling by spelling rather than assuming it),
+    # and one call site cannot drift from the other two.
+    refusal = containment_refusal(script, project)
+    if refusal is not None:
+        return refusal
 
     try:
         binary = resolve_godot_binary(godot)
@@ -2147,18 +2145,21 @@ def _script_validate_recipe(
     own entry.
 
     The refusal has TWO halves since ADR-0006's 2026-08-31 amendment (#697), and
-    the second is why a *projectless* call is now checked too. Containment
-    (:func:`~gda.project.path_outside_project`) asks whether the target is in the
-    resolved project's tree, which only a resolved project can fail. Ownership
-    (:func:`~gda.project.owning_project`) asks whether that project is really the
-    target's OWNER — a ``project.godot`` nearer to the target claims it — and that
-    is the half GDA-DF-035 exposed in both its readings: an ancestor that is a
-    project, with the target in a nested one; and a projectless run of a file that
-    does have an owner. Both compiled the target against a root that was not its
-    own and produced the same cascade of false ``res://`` errors. gda refuses and
-    names the owner instead of adopting it: deriving the root from the target is
-    what ADR-0006 rejected and the amendment keeps rejected, so ``--project``
-    naming the owner stays the way to say what you mean. A standalone script with
+    the second is why a *projectless* call is now checked too. Both are asked by
+    ONE call to :func:`~gda.project.containment_refusal` (#802), which owns their
+    ordering and builds whichever envelope fires; this recipe only chooses the
+    targets. Containment (:func:`~gda.project.path_outside_project`) asks whether
+    the target is in the resolved project's tree, which only a resolved project
+    can fail. Ownership (:func:`~gda.project.owning_project`) asks whether that
+    project is really the target's OWNER — a ``project.godot`` nearer to the
+    target claims it — and that is the half GDA-DF-035 exposed in both its
+    readings: an ancestor that is a project, with the target in a nested one; and
+    a projectless run of a file that does have an owner. Both compiled the target
+    against a root that was not its own and produced the same cascade of false
+    ``res://`` errors. gda refuses and names the owner instead of adopting it:
+    deriving the root from the target is what ADR-0006 rejected and the amendment
+    keeps rejected, so ``--project`` naming the owner stays the way to say what
+    you mean. A standalone script with
     no owner is still validated projectless by filesystem path, exactly as before.
 
     ``--all`` has nothing to check: the engine enumerates the resolved project's
@@ -2181,22 +2182,12 @@ def _script_validate_recipe(
     """
     root = None if project is None else project.expanduser().resolve()
     for path in params.paths:
-        # Ownership first: it is the more specific diagnosis, and it is the only
-        # one a projectless call can make. Containment then needs a root to be
-        # outside OF — `root` is set exactly when `project` is, and both are named
-        # so the narrowing stays local to the branch that uses them.
-        owner = owning_project(path, project)
-        if owner is not None:
-            return target_owned_by_another_project_failure(
-                target_location(path, project),
-                owner.resolve(),
-                root,
-                owner_relative_target(path, project, owner),
-            )
-        if project is not None and root is not None:
-            outside = path_outside_project(path, project)
-            if outside is not None:
-                return target_outside_project_failure(outside, root)
+        # ONE call for both halves and their ordering (#802): the gate on ADR-0006's
+        # path authority owns them, so this recipe states only WHICH targets it is
+        # asking about — the batch, in requested order, first offender wins.
+        refusal = containment_refusal(path, project)
+        if refusal is not None:
+            return refusal
     # The runner seam is read off the module at call time — never imported by
     # name — so a test monkeypatch on ``gda.dispatch.make_runner`` still binds.
     # Naming the HEADLESS factory directly is correct only while this command is
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -61,6 +61,7 @@ from gda.models import (
     TerminationPhase,
 )
 from gda.parser import parse_result
+from gda.project import ForeignOwnerViolation, containment_violation
 from gda.runner import DEFAULT_TIMEOUT_LABEL, LaunchFailure, RunResult
 from gda.script_errors import ScriptError, script_error_line
 
@@ -885,6 +886,38 @@ def target_owned_by_another_project_failure(
             owning_project=str(owner),
         ),
     )
+
+
+def containment_refusal(target: str, project: Path | None) -> Failure | None:
+    """The refusal when ``target`` does not belong to ``project`` — or ``None`` (#802).
+
+    THE gate the three path-taking commands call — ``script validate`` per batch
+    entry, ``script run`` for its entry script, ``resource import`` per asset. The
+    DECISION is not made here: :func:`gda.project.containment_violation` owns the
+    ordering (ownership first), the normalization, and the four coordinates; this
+    function maps each half of its answer to the envelope the taxonomy owns. The
+    split follows ADR-0040 §5 — the taxonomy reaches DOWN to the path authority,
+    never the reverse; the composition briefly lived whole on ``gda.project`` and
+    needed a deferred import of this module to hide the inverted edge (#807
+    review).
+
+    One builder of the same code stays outside the gate, deliberately:
+    :func:`script_escapes_project_failure`, ``script run``'s pre-resolution address
+    gate (ADR-0031). It decides on the spelling alone, before there is a project to
+    be outside OF, so it holds none of the four coordinates a refusal from here
+    reports; the argument for keeping it apart is at that builder.
+    """
+    violation = containment_violation(target, project)
+    if violation is None:
+        return None
+    if isinstance(violation, ForeignOwnerViolation):
+        return target_owned_by_another_project_failure(
+            violation.location,
+            violation.owner,
+            violation.root,
+            violation.owner_relative,
+        )
+    return target_outside_project_failure(violation.outside, violation.root)
 
 
 def script_escapes_project_failure(script: str) -> Failure:
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -797,11 +797,13 @@ def target_outside_project_failure(location: Path, project: Path) -> Failure:
     gda therefore refuses *before* the target is parsed and reports the mismatch
     itself, naming both sides so the reader can see which one is wrong.
 
-    The three commands that hold a resolved project when they ask the containment
-    question share this builder through the gate (:func:`containment_refusal`,
-    #802) — ``script validate``, ``script run``, and ``resource import`` — so one
+    Two commands can reach this builder through the gate
+    (:func:`containment_refusal`, #802) — ``script validate`` and ``resource
+    import``, the ones whose target may still be a filesystem spelling — so one
     condition reports one code, one message and one pair of typed coordinates
-    (#763). The message says what is true of BOTH: a
+    (#763). ``script run`` calls the same gate, but only its ownership half is
+    reachable there: an address its own gate accepted is canonical ``res://``,
+    which is never outside the root (#807 round 4). The message says what is true of BOTH: a
     target gda addresses through the project's ``res://`` namespace has no place
     in that namespace. Why that matters differs per command — a compile resolves
     the target's own dependencies, an import writes into the project's cache — and
--- a/src/gda/errors.py
+++ b/src/gda/errors.py
@@ -797,10 +797,11 @@ def target_outside_project_failure(location: Path, project: Path) -> Failure:
     gda therefore refuses *before* the target is parsed and reports the mismatch
     itself, naming both sides so the reader can see which one is wrong.
 
-    The two commands that hold a resolved project when they ask the containment
-    question share this builder — ``script validate``'s recipe and ``resource
-    import``'s asset gate — so one condition reports one code, one message and one
-    pair of typed coordinates (#763). The message says what is true of BOTH: a
+    The three commands that hold a resolved project when they ask the containment
+    question share this builder through the gate (:func:`containment_refusal`,
+    #802) — ``script validate``, ``script run``, and ``resource import`` — so one
+    condition reports one code, one message and one pair of typed coordinates
+    (#763). The message says what is true of BOTH: a
     target gda addresses through the project's ``res://`` namespace has no place
     in that namespace. Why that matters differs per command — a compile resolves
     the target's own dependencies, an import writes into the project's cache — and
--- a/src/gda/project.py
+++ b/src/gda/project.py
@@ -509,9 +509,9 @@ def project_absolute(project: Path) -> Path:
     ``~user`` stays literal rather than raising out of a containment check.
 
     Written for ``resource import``'s asset gate and adopted by
-    :func:`containment_refusal` as the gate's own normalization (#802); the asset
-    gate still calls it directly because it also maps an accepted path back onto
-    ``res://`` afterwards, which needs the same absolute root.
+    :func:`containment_violation` as the decision's own normalization (#802); the
+    asset gate still calls it directly because it also maps an accepted path back
+    onto ``res://`` afterwards, which needs the same absolute root.
     """
     absolute = _expand_user(project)
     if not absolute.is_absolute():
--- a/src/gda/project.py
+++ b/src/gda/project.py
@@ -27,12 +27,26 @@ on, :func:`canonical_res_path` and :func:`res_escape_remainder`, live here too:
 they are pure lexical address rules, so they belong beside
 :data:`ENGINE_VIRTUAL_PREFIXES` and :func:`_lexical_abs` rather than in the
 stderr parser that first needed one (:mod:`gda.script_errors`, now a consumer).
+
+Since #802 the authority owns the **gate** as well as the primitives:
+:func:`containment_refusal` is the whole ordered composition — normalize the
+project, ask ownership, ask containment, build whichever refusal fired — so a
+command module states only WHICH target it is asking about. Before it, three call
+sites wrote that orchestration by hand and the ordering rule, the argument shapes
+and the ``.resolve()`` discipline were interface cost each of them paid.
 """
 
 import os
 import posixpath
 from collections.abc import Mapping
 from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    # `gda.errors` imports `gda.models`, which imports THIS module, so the two
+    # refusal builders `containment_refusal` needs can only be imported inside
+    # it (see the note there). The name is still needed at type-check time.
+    from gda.errors import Failure
 
 GDA_PROJECT_ENV = "GDA_PROJECT"
 
@@ -477,6 +491,107 @@ def owning_project(path: str, project: Path | None) -> Path | None:
             return None
         if (directory / PROJECT_MARKER).exists():
             return directory
+    return None
+
+
+def project_absolute(project: Path) -> Path:
+    """``project`` as an absolute directory, still spelled the caller's way (#738).
+
+    ONE coordinate system for both sides of every containment comparison. A
+    project may arrive relative — ``--project game`` is resolved to a directory,
+    not to an absolute one (:func:`resolve_project_dir`) — and a relative root met
+    an absolute candidate on :func:`path_outside_project`'s lexical fallback,
+    where ``relative_to`` raised a bare ``ValueError`` instead of answering.
+    Anchoring at the cwd is the right anchor because that is where a relative
+    ``--project`` was typed.
+
+    Symlinks are deliberately NOT followed: the two readings
+    :func:`path_outside_project` and :func:`owning_project` make are theirs to
+    make, and pre-resolving here would take the lexical one away from them. ``~``
+    is expanded the module's total way (:func:`_expand_user`), so an unresolvable
+    ``~user`` stays literal rather than raising out of a containment check.
+
+    Written for ``resource import``'s asset gate and adopted by
+    :func:`containment_refusal` as the gate's own normalization (#802); the asset
+    gate still calls it directly because it also maps an accepted path back onto
+    ``res://`` afterwards, which needs the same absolute root.
+    """
+    absolute = _expand_user(project)
+    if not absolute.is_absolute():
+        absolute = Path.cwd() / absolute
+    return absolute
+
+
+def containment_refusal(target: str, project: Path | None) -> "Failure | None":
+    """The refusal when ``target`` does not belong to ``project`` — or ``None`` (#802).
+
+    THE gate behind ``target_outside_project``, and the whole of it: the one
+    question "does this target belong to the resolved project?", asked in one
+    order, answered in one pair of envelopes. Three commands ask it — ``script
+    validate`` per batch entry, ``script run`` for its entry script, ``resource
+    import`` per asset — and until #802 each wrote the composition by hand, so the
+    ordering rule, the four coordinates and the ``.resolve()`` discipline were
+    interface cost every one of them paid. They had drifted once already (#763's
+    postmortem, then #799's), which is why the cross-gate consistency test exists;
+    it now guards this function's output rather than being the only thing holding
+    three copies together.
+
+    **Ordering: ownership first, containment second.** Ownership is the more
+    specific diagnosis — it names the project that CAN serve the call and the
+    spelling to address the target by, so following the sentence verbatim works —
+    while containment can only say "not here, and I found no owner to send you
+    to". The two never both fire on one target (:func:`owning_project` answers
+    ``None`` for anything not inside the resolved tree, which is exactly what
+    :func:`path_outside_project` refuses), so the order is about which diagnosis a
+    caller gets, not about precedence between rival verdicts. Ownership is also
+    the only half a PROJECTLESS call can make: with no root there is nothing to be
+    outside OF, so containment is skipped and a standalone file that no
+    ``project.godot`` claims is served, as ADR-0006's projectless fallback
+    promises.
+
+    **Normalization: the project is cwd-absolutized** (:func:`project_absolute`),
+    the form ``resource import`` adopted in #738 — the checks read the project as
+    SPELLED (the double reading is theirs to make) but they must read it in one
+    coordinate system, or a relative ``--project`` meets an absolute candidate on
+    the lexical fallback. The refusals report it RESOLVED, which is what
+    ``FailureEvidence.project_root`` publishes: "in its resolved form — the same
+    value a successful result reports".
+
+    ``target`` is the caller's own spelling in any form the command accepts —
+    ``res://`` address, project-relative path, or absolute path — because
+    :func:`_anchored_target` anchors each of them the way the engine would.
+    ``project`` is the already-resolved directory (resolution stays CLI-side,
+    ADR-0006); ``None`` means projectless.
+    """
+    # Imported here, not at module scope: `gda.errors` imports `gda.models`, which
+    # imports this module for `is_engine_virtual_path`, so a top-level import would
+    # cycle. The dependency is genuinely one-way in every other direction — the
+    # authority owns the RULE, the taxonomy owns the ENVELOPE — and this is the one
+    # place the two meet.
+    from gda.errors import (
+        target_outside_project_failure,
+        target_owned_by_another_project_failure,
+    )
+
+    if project is None:
+        anchor = root = None
+    else:
+        anchor = project_absolute(project)
+        root = anchor.resolve()
+    owner = owning_project(target, anchor)
+    if owner is not None:
+        return target_owned_by_another_project_failure(
+            target_location(target, anchor),
+            owner.resolve(),
+            root,
+            owner_relative_target(target, anchor, owner),
+        )
+    # `root` is set exactly when `anchor` is; both are named so the narrowing stays
+    # local to the branch that uses them.
+    if anchor is not None and root is not None:
+        outside = path_outside_project(target, anchor)
+        if outside is not None:
+            return target_outside_project_failure(outside, root)
     return None
 
 
--- a/src/gda/project.py
+++ b/src/gda/project.py
@@ -525,26 +525,41 @@ def project_absolute(project: Path) -> Path:
 def containment_refusal(target: str, project: Path | None) -> "Failure | None":
     """The refusal when ``target`` does not belong to ``project`` — or ``None`` (#802).
 
-    THE gate behind ``target_outside_project``, and the whole of it: the one
+    THE gate behind ``target_outside_project`` once a project is RESOLVED: the one
     question "does this target belong to the resolved project?", asked in one
     order, answered in one pair of envelopes. Three commands ask it — ``script
     validate`` per batch entry, ``script run`` for its entry script, ``resource
     import`` per asset — and until #802 each wrote the composition by hand, so the
     ordering rule, the four coordinates and the ``.resolve()`` discipline were
-    interface cost every one of them paid. They had drifted once already (#763's
+    interface cost every one of them paid. They had drifted twice already (#763's
     postmortem, then #799's), which is why the cross-gate consistency test exists;
     it now guards this function's output rather than being the only thing holding
     three copies together.
 
-    **Ordering: ownership first, containment second.** Ownership is the more
-    specific diagnosis — it names the project that CAN serve the call and the
-    spelling to address the target by, so following the sentence verbatim works —
-    while containment can only say "not here, and I found no owner to send you
-    to". The two never both fire on one target (:func:`owning_project` answers
-    ``None`` for anything not inside the resolved tree, which is exactly what
-    :func:`path_outside_project` refuses), so the order is about which diagnosis a
-    caller gets, not about precedence between rival verdicts. Ownership is also
-    the only half a PROJECTLESS call can make: with no root there is nothing to be
+    One builder of that same code stays OUTSIDE this gate, deliberately:
+    :func:`gda.errors.script_escapes_project_failure`, ``script run``'s
+    pre-resolution address gate (ADR-0031). It decides on the spelling alone,
+    before there is a project to be outside OF, so it holds none of the four
+    coordinates a refusal from here reports; it reaches the same rule through
+    :func:`res_escape_remainder`, and the argument for keeping it apart is at the
+    builder. Everything downstream of a resolved project is here.
+
+    **Ordering: ownership first, containment second.** A real precedence rule, not
+    only a choice of wording, because the two halves CAN both fire on one target:
+    their bounds differ. :func:`owning_project` stops its walk by :func:`_within`,
+    which consults the lexical spelling unconditionally, while
+    :func:`path_outside_project` WITHHOLDS the lexical reading when either side
+    carries a ``..``. A file symlinked into the project from a tree that is itself
+    a project, addressed with a ``..``, therefore satisfies the walk's bound and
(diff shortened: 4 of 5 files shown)


qinhaihong-red:
## Full independent adversarial review — PR #807

**Reviewed at** head `e6eb969c6d23665be7b1f9f5378776f4fc490118` (`refactor/802-containment-gate`), against merge-base `fe72700ecb6d12b24474560c4b8599266df6ebcd` (the PR's real "before"; `feat/gda-dogfooding-dev` tip `64136719c` carries unrelated `gda-balancing` work). Diff under review: `src/gda/commands/resource.py`, `src/gda/commands/script.py`, `src/gda/project.py`, `tests/test_res_containment_consistency.py` (+318 / −72).

Everything below is my own evidence. The PR body, its tables and its claims were treated as a checklist to falsify, not as proof.

### Verdict

| Axis | Verdict |
| --- | --- |
| Artifact (issue ↔ diff, docstring-as-contract, terminology, prose) | **Changes required** — 3 × P2 |
| Architecture (boundaries, ownership, dependency direction, deep module) | **Pass**, with notes |
| Standards (RULES.md, repo conventions, Fowler smells) | **Pass** |
| Spec (every AC of #802, literally) | **Pass**, with one P2 carried from the artifact axis (AC1's ordering is *owned* but not *pinned*) |

The refactor does what #802 asked and it is behavior-preserving — I re-derived that independently on the real engine (45 scenarios, byte-identical stdout+stderr+exit at base and head). No P1. The three P2s are all about the same thing: the gate is now the **single authority** for a rule, and the prose that authority carries is wrong in three places, one of which is demonstrably falsifiable and unguarded.

Findings are ranked P1 (blocks) / P2 (fix before merge) / P3 (note).

---

## P2-1 — `containment_refusal`'s central justification is false, and the ordering it now solely owns is not pinned

`src/gda/project.py:539-546`:

> **Ordering: ownership first, containment second.** … The two never both fire on one target (`owning_project` answers `None` for anything not inside the resolved tree, which is exactly what `path_outside_project` refuses), so the order is about which diagnosis a caller gets, not about precedence between rival verdicts.

The parenthetical is not true. The two bounds are **not** complements:

- `owning_project`'s bound is `_within(start, stop)` (`project.py:465`), which consults the lexical reading **unconditionally**.
- `path_outside_project` **withholds** the lexical reading when either side carries a `..` (`project.py:320`, the `_has_dotdot` guard).

So a target that is lexically inside and resolves outside satisfies `owning_project`'s bound *and* `path_outside_project`'s refusal. That is exactly the monorepo shared-addon layout `path_outside_project`'s own docstring exists for, plus a `..` in the spelling.

Self-contained reproduction (run in the PR's own venv):

```python
game/project.godot;  libs/lib/project.godot;  libs/lib/tool.gd
game/addons/lib -> ../../libs/lib
target = "addons/lib/../lib/tool.gd"
```

```
=== HEAD (unmutated) ===
owning_project      -> …/game/addons/lib
path_outside_project-> …/libs/lib/tool.gd          <-- BOTH fire
gate code           -> target_outside_project
gate evidence       -> {'target_location': …, 'project_root': …, 'owning_project': '…/libs/lib'}

=== HEAD + mutation: the two halves swapped ===
gate evidence       -> {'target_location': …, 'project_root': …}      <-- owner GONE
```

Through the real CLI at head vs. the swap (`--project <fx>/game addons/lib/../lib/tool.gd`):

| | head | order swapped |
| --- | --- | --- |
| `script validate` | owner refusal, names `--project …/libs/lib` + `'tool.gd'` | bare "is outside the resolved Godot project", **no owner, no re-issue** |
| `resource import` | owner refusal | bare outside-root refusal |
| `script run` | owner refusal | owner refusal (its canonical `res://` address makes containment inert) |

So swapping the order (a) strictly degrades the diagnosis on a real layout — the actionable half of ADR-0006's amendment disappears — and (b) breaks the PR's own headline invariant, because `script run` no longer agrees with the other two.

**And the suite does not notice.** With the two halves swapped in the gate:

```
$ pytest -p no:cacheprovider -q tests/test_res_containment_consistency.py
50 passed in 0.18s
```

This matters more than a normal test gap. Before #807 the ordering was written three times, so a reviewer of any one copy could see it. After #807 it exists once, its rationale is the docstring above, and that rationale says the order is cosmetic. A future maintainer following the authority verbatim will reorder it and ship green.

**Smallest fix:** (1) correct the parenthetical — the two bounds differ precisely by `path_outside_project`'s `..` guard, so both *can* fire and ownership-first is a real precedence rule, not only a choice of wording; (2) add the both-fire row to the consistency table (the fixture above is ~8 lines) and red-proof it against the swap, alongside the three mutations the PR body already red-proofs.

## P2-2 — "THE gate behind `target_outside_project`, and the whole of it" is not the whole of it

`src/gda/project.py:528`:

> THE gate behind `target_outside_project`, and the whole of it …

There is a third builder of that same code, and a **command module calls it today**:

```
$ grep -rn "script_escapes_project_failure" src/
src/gda/errors.py:890:def script_escapes_project_failure(script: str) -> Failure:
src/gda/commands/script.py:39:    script_escapes_project_failure,
src/gda/commands/script.py:1271:        return script_escapes_project_failure(script)
```

```
$ gda script run --json --project <fx>/game res://../outside.gd
{"error":{"code":"target_outside_project","message":"script 'res://../outside.gd' escapes above the project root: …"}}
```

The code is fine — that builder is the pre-project-resolution lexical edge (ADR-0031), and `errors.py:900-905` argues correctly for keeping it separate. The **prose** overreaches, in two places that a reader will treat as contract:

- the gate docstring, as quoted;
- `tests/test_res_containment_consistency.py:379-386` — *"The two builders that construct a `target_outside_project` refusal. #802's first acceptance criterion is that no command module calls either … so adding a third builder is a change this test makes someone look at."* The third builder already exists, and is already called from a command module. `test_no_command_module_builds_a_containment_refusal_itself` therefore reads as a stronger structural guarantee than it is.

**Smallest fix:** in the gate docstring, bound the claim ("the whole of it once a project is resolved; `script run`'s pre-launch address gate reaches the same code lexically through `script_escapes_project_failure`, before there is a project to be outside OF — ADR-0031"), and in the test, say the set is the two **resolved-project** builders and name the deliberate exclusion.

## P2-3 — the relative-`--project` row claims a discriminating power it does not have

`tests/test_res_containment_consistency.py:351-354`:

> a future normalization that anchored anywhere but the cwd, **or resolved the project before the probes rather than after**, fails here.

First clause: true. Second clause: false.

| mutation of `project_absolute` | consistency suite |
| --- | --- |
| `Path("/") / absolute` (anchor somewhere other than cwd) | **2 failed**, 48 passed ✅ pinned |
| `return _expand_user(project).resolve()` (resolve before the probes) | **50 passed** ❌ not pinned |

The second mutation is not benign. It removes the `..` from the project spelling before `path_outside_project` reads `_has_dotdot(root)`, which re-enables the lexical reading:

```
$ gda script validate --json --project <fx>/alias/../game addons/lib/tool.gd
base : {"code":"target_outside_project", …}       # refusal
head : {"code":"target_outside_project", …}       # refusal (identical)
head + resolve-before-probes : {"valid":true, …}  # ACCEPTED
```

A refusal silently becomes an acceptance and the suite stays 50/50 green.

**Smallest fix:** either delete the second clause (it is a claim the test does not make), or add the row that makes it true — a `--project` carrying a `..` plus a symlinked-in target, which is the input the two normalizations disagree on.

---

## P3 notes

**P3-1 — AST-guard strength.** `_called_names` (`tests/…:389-394`) matches only `ast.Call` with `ast.Name` func. I re-introduced a copy in `resource.py` as `gda.errors.target_outside_project_failure(...)` (module-qualified) and the guard stayed green (50 passed); the plain-name form correctly went red. The glob is also `package.glob("*.py")` — a future `gda/commands/<sub>/` package is not scanned — and a direct `make_failure("target_outside_project", …)` is not caught either. The repo's import idiom is `from gda.errors import …`, so this is low-probability; worth one line saying so, or matching `ast.Attribute` too and using `rglob`.

**P3-2 — two normalizations for one published `project_root`.** `_script_validate_recipe` still derives the SUCCESS path's `project_root` as `project.expanduser().resolve()` (`src/gda/commands/script.py:2183`), while the REFUSAL path now derives it as `project_absolute(project).resolve()` inside the gate. They agree today (I verified across `<abs>`, `game`, `./game`, `alias/../game`), but this is precisely the class of divergence #799 was — the same call answering two spellings of one root. One line: `root = None if project is None else project_absolute(project).resolve()`.

**P3-3 — `resource import` normalizes twice.** `resource.py:836` computes `project_absolute(project)` for the `res://` mapping and the gate recomputes it at `project.py:579`. Harmless (two `Path.cwd()` calls), but the comment "The gate below normalizes the same way" is a two-places-must-agree coupling of exactly the kind #802 set out to delete. If the gate returned the anchor it used (or the mapping took one), the coupling goes away.

**P3-4 — `SCRIPT_RUN_SPELLINGS` covers one of the two accepted forms.** The inertness test's own argument is about "whatever address survives `_project_scoped_res_path`", and that gate accepts project-relative paths too (`script.py:1262`, `lifted = script if …startswith(RES_PREFIX) else RES_PREFIX + script`) — `gda script run --project … inner/main.gd` reaches the shared gate. The table drives only `res://` spellings. Adding `inner/main.gd`, `addons/lib/../lib/tool.gd` and `../outside.gd` makes the table cover the claim it states. (The claim itself is sound — I checked; the fold is genuinely inert.)

**P3-5 — prose.** `project.py:534` — *"They had drifted **once** already (#763's postmortem, **then #799's**)"* says once and then lists two. The commit body and the PR body both say "twice", which is correct.

**P3-6 — pre-existing, not this PR's to change.** `gda script validate --help` still says *"A path OUTSIDE the resolved project refuses the whole batch with `'project_not_found'`"* (`src/gda/commands/script.py:2516-2517`); the code has emitted `target_outside_project` since #697. AC5 requires help to stay byte-identical, so this cannot be fixed here — worth a follow-up issue, since #807 is the PR that makes this refusal single-sourced.

---

## Architecture notes (no change required)

**Deep-module quality: good.** Three sites × ~14 lines of composition → one call each. The interface is `(target: str, project: Path | None) -> Failure | None` — no options, no ordering knowledge, no `.resolve()` discipline leaked to the caller. The command modules now state only *which* target they ask about, which is the shape #802 asked for and the right one.

**The deferred import is a real package cycle, correctly handled.** Verified by AST over `src/gda/*.py`:

```
gda.project → gda.errors      (function-local, col_offset != 0)
gda.errors  → gda.models      (module scope)
gda.models  → gda.project     (module scope)
```

It is safe at runtime (the import happens at call time, when `gda.project` is fully initialized) and it is documented at both ends. Recording the alternative so the next person does not re-derive it: `gda.errors` already reaches `gda.project` transitively through `gda.models`, so hosting the composition in `gda.errors` — or in a small third module importing both at module scope — would need no lazy import at all. #802 specified `gda.project` as the home, and the RULE genuinely belongs with the authority, so I am not asking for a move; but a note in the docstring that the cycle is the price of that placement would help a future reader weigh it.

**`project_absolute` as a second public name.** Justified in its docstring (`resource import` needs the same absolute root for the `res://` mapping after the gate accepts). See P3-3 for the residual.

---

## Terminology and prose assessment

- **Terminology sources:** `CONTEXT.md` (shared language), `docs/adr/0006-project-context-and-path-resolution.md`, `docs/adr/0031`, `docs/command-catalog.md`.
- **Writing reference:** ASD-STE100, per `RULES.md`.
- **Scope inspected:** every changed prose block — the `project.py` module docstring addition, `project_absolute` and `containment_refusal` docstrings, the three call-site comments, the `_script_validate_recipe` docstring re-wrap, and the whole new test section; plus the surrounding unchanged docstrings the new prose refers to.
- **Result:** CONTEXT.md terms are used correctly and consistently — *Error envelope*, *Gda error code*, *refusal*, *gate*, *path authority*, *projectless*. No second name is introduced for an existing concept; "containment gate" composes existing terms and is not a rival to any CONTEXT.md entry. No ADR reconcile is due: ADR-0006 line 208 already says *"`gda.project` now owns the whole answer"*, so #802 realizes that decision rather than changing it (I read the amendment section to confirm).
- **ASD-STE100 checks performed:** word choice and technical nouns/verbs, sentence structure and length, one-topic-per-sentence, accuracy of causal/dependency/scope relationships. Sentence length exceeds STE's descriptive guidance in several new sentences (45–60 words), but that is the established house style of this file and the surrounding unchanged prose, so I am **not** raising it. The relationship-accuracy check is where the real findings came from: **P2-1**, **P2-2** and **P2-3** are all prose stating a causal or scope relationship that the code does not have.
- **Not performed:** no review of `libs/gda-balancing` (outside this PR's diff).

---

## Validation

Scratch worktrees at head (`e6eb969c6`) and merge-base (`fe72700ec`), separate `uv` envs, real engine at `~/Applications/Godot.app/…/Godot` (v4.6.3.stable), per-run `GDA_USER_DATA_ROOT` under `/private/tmp`.

**Behaviour, base vs head — 45 scenarios, byte-identical (stdout + stderr + exit code), `diff -r` clean.** Fixture: `game/` with a nested `game/inner/` project, an outside tree, and a symlinked-in `game/addons/lib -> ../../libs/lib`.

- Foreign-owner refusals on all three commands × `res://`, project-relative and absolute spellings (`v_owner_res/_rel/_abs`, `r_owner_res/_rel`, `i_owner_res/_rel`)
- Outside-root refusals × absolute, `res://` escape and `..` spellings (`v_outside_abs/_res/_dotdot`, `i_outside_abs/_res/_dotdot`)
- Projectless-with-owner (`v_projectless`), batch first-offender in both orders (`v_batch_first/_second`, `i_batch`)
- `script run`'s own edges: `invalid_path` for absolute / `res://` root / `user://` (`r_abs`, `r_root`, `r_user`), lexical escape (`r_escape`)
- Relative `--project` × `game`, `./game`, `alias/../game` × all three commands × owner and outside cases (12 scenarios)
- Symlink cases: monorepo addon accepted (`link_v`, `link_i`, `link_*_abs`), `..`-through-symlink refused (`link_v_dotdot`)
- Real-engine accepted controls: `script validate`, `script run`, `resource import` (plain and through the symlink), with the import cache reset before each side

**Byte-identity of the published surface (AC5).** `gda schema --json`: identical, 1 038 261 bytes. Help: identical across **91** surfaces — root + 15 groups + all 76 commands enumerated from the schema, at `COLUMNS=80`. `docs/command-catalog.md` is untouched by the diff.

**Cross-gate envelope agreement I checked myself,** beyond the PR's own tests: `script validate` and `resource import` emit the identical outside-root envelope for one shared target, at base and at head.

**Suites and gates at head:** `pytest -p no:cacheprovider -q -m "not e2e"` → 2413 passed; `pytest -m e2e tests/test_e2e_res_containment.py` with the real engine → 19 passed; `ruff check .` → all passed; `ruff format --check .` → 485 files formatted; `pyright` → 0 errors. PR CI at this head is green (Godot e2e correctly `skipping`).

**Mutation testing — 6 source-level mutations against `tests/test_res_containment_consistency.py`:**

| # | mutation | expected | result |
| --- | --- | --- | --- |
| M1 | swap ownership/containment in the gate | red | **50 passed** → P2-1 |
| M2 | `project_absolute` anchors at `/` not cwd | red | 2 failed ✅ |
| M3 | `project_absolute` resolves before the probes | red (test says so) | **50 passed** → P2-3 |
| M4 | `script run` stops calling the gate | red | 10 failed ✅ |
| M5 | command module builds the refusal (plain name) | red | 1 failed ✅ |
| M6 | same, module-qualified `gda.errors.…(…)` | red | **50 passed** → P3-1 |

All mutations reverted; both scratch worktrees removed.

**Not performed:** Windows/Linux path behaviour (the `\\` and drive branches of `canonical_res_path` are exercised only through the POSIX table); the full `-m e2e` suite beyond the containment file.

---

### AC ledger (#802, corrected body)

| AC | Verdict |
| --- | --- |
| 1 — one function owns ordering + both envelopes; no command module builds one | **Met** structurally (grep + AST test). The *ordering* half is owned but unpinned and mis-justified → P2-1; the "no command module builds one" claim needs the `script_escapes_project_failure` exception stated → P2-2 |
| 2 — three sites call the gate, copies deleted, `script run` fold **verified** incl. a relative-`--project` row | **Met.** The inertness argument is sound and I confirmed it independently; table coverage note in P3-4 |
| 3 — the cross-gate consistency test still passes unchanged | **Met.** The diff is additive only after line 280 (plus imports and a docstring paragraph); every pre-existing test body is untouched and green |
| 4 — refusal envelopes byte-identical before/after | **Met**, independently re-derived on 45 real-CLI scenarios |
| 5 — schemas, help, catalog byte-identical | **Met**, independently re-derived (schema + 91 help surfaces) |

qinhaihong-red:
## Review handled — 3 × P2, 6 × P3

New head **`83e885ae092eaa943ac2a10b0f0a5259edd7af21`** (one commit, `83e885ae0`, on top of the reviewed `e6eb969c6`).

Every claim was re-derived here before it was acted on. **All three cited surviving mutations were reproduced at the reviewed head** (M1, M3, M6 → `50 passed` each), and all three now go **red**. Thank you — the two unguarded ones were real, and P2-1's premise was falsifiable exactly as described.

### Decisions

| # | Claim | Decision | Evidence I derived | Change | Validation |
| **P2-1** | Ordering rationale false ("the two never both fire"), and the ordering is unpinned | **Fully adopt** | Rebuilt your layout: `owning_project` → `…/game/addons/lib`, `path_outside_project` → `…/libs/lib/tool.gd` on `addons/lib/../lib/tool.gd` — **both fire**. The bounds differ exactly as you say: `_within` reads the lexical spelling unconditionally, `path_outside_project` withholds it when either side carries a `..` | Rationale rewritten as a real precedence rule (`project.py`); new `monorepo` fixture + `test_ownership_wins_when_both_halves_of_the_gate_fire` | M1 (halves swapped): `50 passed` before → **`1 failed, 54 passed`** after |
| **P2-2** | "THE gate … and the whole of it" overreaches; `script_escapes_project_failure` is a third builder, called from a command module | **Fully adopt (prose only)** | `errors.py:890` defines it, `commands/script.py:1271` calls it; `gda script run --json --project <fx> res://../outside.gd` → `target_outside_project` | Docstring bounded to "once a project is RESOLVED", naming the pre-resolution builder and ADR-0031; `CONTAINMENT_REFUSAL_BUILDERS` comment records it as the deliberate **exclusion**, and restates the set as "the builders a command must route through the gate for" | No fold attempted — it is a pre-resolution edge by ADR-0031's decision and holds none of the four coordinates. Suite green |
| **P2-3** | Relative-`--project` row claims a discrimination it lacks; resolve-before-probes flips a refusal to an acceptance | **Fully adopt** | Confirmed: `--project alias/../game addons/plain/tool.gd` → `target_outside_project` at head, **ACCEPTED** under the mutation, on both filesystem-path gates | New `test_the_gate_reads_the_project_as_spelled_rather_than_pre_resolved` (the `..`-carrying project spelling + symlinked-in non-project target); the old row's comment now points at it instead of overclaiming | M3: `50 passed` before → **`1 failed, 54 passed`** after |
| **P3-1** | AST guard misses module-qualified calls and subpackages | **Partially adopt** | Reproduced M6 (`gda.errors.target_outside_project_failure(...)`) → `50 passed` at the reviewed head | `_called_names` now reads `ast.Attribute` as well as `ast.Name`; `glob` → `rglob`. **Not adopted:** scanning for a raw `make_failure("target_outside_project", …)` — that is the generic builder used across the whole taxonomy, so the check would be a string scan with a different failure mode, for a path no module takes | M6: `50 passed` before → **`1 failed, 54 passed`** after |
| **P3-2** | Two normalizations for one published `project_root` | **Fully adopt** | The two agree today (I re-verified on the engine across `<abs>`, `game`, `./game`, `alias/../game`); they differ only in `_expand_user`'s totality on an unresolvable `~user` | `script.py`: `root = None if project is None else project_absolute(project).resolve()` | Success-path results byte-identical vs merge-base on all four spellings (real engine) |
| **P3-3** | `resource import` normalizes twice; gate could return the anchor | **Do not adopt (comment corrected)** | The two sites call the **same function**, `project_absolute` — there is one normalization rule, invoked twice, not two rules to keep in step. Returning the anchor would widen the gate's interface from `Failure \| None` to a pair, which is the interface cost #802 removed and the shape you flagged as good in the architecture notes; the cost paid is one extra `Path.cwd()` | Comment reworded to say so explicitly, since it read as a coupling | Behavior unchanged |
| **P3-4** | `SCRIPT_RUN_SPELLINGS` covers one of two accepted forms | **Fully adopt** | `script.py:1262` lifts a project-relative path onto `res://`, so it reaches the shared gate | Added `inner/main.gd`, `addons/lib/../lib/tool.gd`, `../outside.gd` | Inertness test now 13 rows; green |
| **P3-5** | "drifted **once** … then #799's" | **Fully adopt** | Two postmortems listed | "twice" | — |
| **P3-6** | `script validate --help` still says `project_not_found` | **Do not adopt here** | Confirmed pre-existing (`script.py:2516-2517`); the code has emitted `target_outside_project` since #697 | None — AC5 requires help byte-identical, and my run confirms it still is. Follow-up draft below | Out of scope, as you note |
| *Arch. note* | Record the cycle's alternative placement | **Fully adopt** | `gda.errors` already reaches `gda.project` through `gda.models` | One sentence at the deferred import: hosting it in `gda.errors` needs no lazy import, and #802 declined it because the rule belongs with the authority | — |

### Mutation re-run, this head

| # | Mutation | At `e6eb969c6` | At `83e885ae0` |
| --- | --- | --- | --- |
| M1 | swap ownership/containment in the gate | `50 passed` | **`1 failed`** — `test_ownership_wins_when_both_halves_of_the_gate_fire` |
| M3 | `project_absolute` resolves before the probes | `50 passed` | **`1 failed`** — `test_the_gate_reads_the_project_as_spelled_rather_than_pre_resolved` |
| M6 | module-qualified `gda.errors.…(…)` in a command | `50 passed` | **`1 failed`** — `test_no_command_module_builds_a_containment_refusal_itself` |

M2/M4/M5 continue to fail as they did. Mutations applied non-destructively and reverted; worktree clean.

### Gates at `83e885ae0`

- `pytest tests/ -q -m "not e2e" -p no:cacheprovider` → **2418 passed** (2413 + 5 new rows), 638 deselected
- `pytest -m e2e tests/test_e2e_res_containment.py` (real engine, 4.6.3) → **19 passed**
- `ruff check` → all passed; `ruff format --check` → 485 files formatted; `pyright` → **0 errors**
- **Byte-identity vs merge-base `fe72700ec`**, separate venvs, real engine, per-run `GDA_USER_DATA_ROOT`: **26/26 artifacts identical** (`diff -r` clean) — the owner and outside refusals on all three commands, the two review-driven layouts (both-fire, and the `..`-spelled project), four `--project` spellings on both the refusal and the **success** path (P3-2's line), the projectless row, `gda schema --json` (1 038 261 bytes) and three help surfaces at `COLUMNS=80`

### Not verified

Windows/Linux path behaviour (unchanged by this commit — the new rows are POSIX symlink layouts, and `test_the_gate_reads_the_project_as_spelled_rather_than_pre_resolved` would need a junction on Windows); the full `-m e2e` suite beyond the containment file.

### One thing worth a maintainer's eye

The P2-3 fixture makes a pre-existing asymmetry executable, and it is worth naming rather than burying: on a symlinked-in target, `--project alias/../game` refuses where `--project <abs>` accepts, because `path_outside_project`'s `..` guard reads the PROJECT's spelling too. That is ADR-0006 behavior at base and at head alike (byte-identical above), not something #807 introduces, and the new test pins it as the gate's "read the project as spelled" contract. If it should instead be considered a defect, that is a separate issue against `path_outside_project` — I did not widen this PR to it.

### Follow-up draft (P3-6), not filed

> **`gda script validate --help` names the wrong refusal code**
> `src/gda/commands/script.py:2516-2517` says a path outside the resolved project refuses the batch with `project_not_found`; the code has emitted `target_outside_project` since #697. Not fixable in #807, whose AC5 requires help output byte-identical. Now single-sourced through `gda.project.containment_refusal`, so the help text is the last place stating the old code.

qinhaihong-red:
## Exact-head adversarial review — PR #807

**Changes required.** Reviewed merge-base `fe72700ecb6d12b24474560c4b8599266df6ebcd` against head `83e885ae092eaa943ac2a10b0f0a5259edd7af21` (`git diff origin/feat/gda-dogfooding-dev...refs/codex/review-pr-807-head`). The current base-branch tip is `64136719c6a615ab94887aeeb3447f895511744e`; its extra commits are outside this three-dot diff.

The four review axes below were kept independent. PR/author comments were treated as claims, not closure evidence.

## Artifact review — Changes required (3)

Reviewed the complete current issue #802, including its status (`OPEN`), `ready-for-agent` / `refactor` labels, pre-dispatch amendment, acceptance criteria, and references. This PR changes no Markdown file.

### Required changes

1. **Issue #802 specifies an architecture that conflicts with accepted ADR-0040.** The issue requires a function in `gda.project` to return `Failure` and own both refusal envelopes' construction. ADR-0040 §4 keeps the complete `*_failure` constructor taxonomy in `gda.errors`; §5 fixes the direction as `errors / models -> foundation`. The implementation follows the issue literally and creates `project -> errors -> models -> project`. Impact: the spec cannot be satisfied as written without violating an accepted architecture decision. Smallest fix: revise #802 so `gda.project` owns a pure ordered containment decision and `gda.errors` maps that decision to the two envelopes, or approve and record a different non-cyclic architecture before implementation.

2. **The issue's behavior-preserving claim is false for the current implementation.** It says, “Behavior-preserving: both refusal envelopes, their codes, evidence, and remediation stay byte-identical” and calls the change “internal orchestration only.” At merge-base, a valid literal project directory named `~codex_no_such_user_807` plus an outside target leaks `RuntimeError: Could not determine home directory.` (exit 1). At head, `project_absolute()` uses total `_expand_user()` and the same invocation emits structured `target_outside_project` JSON (exit 4). Impact: reviewers cannot use the issue to decide whether this observable bug fix is in scope. Smallest fix: preserve the prior expansion behavior, or explicitly widen #802 and add this boundary as a regression.

3. **AC4 still says “both commands” after the issue was corrected to three call sites.** The body, AC1, AC2, and the pre-dispatch note all establish `script validate`, `script run`, and `resource import`; AC4 asks for byte identity “on both commands.” Impact: completion scope remains ambiguous at the exact acceptance line. Smallest fix: say “all three commands.”

- Sources: `CONTEXT.md`, accepted ADR-0006, accepted ADR-0031, accepted ADR-0040, and the existing command names/error vocabulary.
- Writing reference: ASD-STE100 Simplified Technical English, as required by `RULES.md`.
- Scope/checks: every line of issue #802; word choice and technical nouns/verbs; causal, dependency, scope, and ordering relationships; internal consistency; all changed-Markdown paths (none).
- Result: established terms such as *project context*, `res://`, refusal, and Error envelope remain consistent. The substantive relationship/scope defects are findings 1–3 above. No other terminology or prose finding.
- Not checked: prose outside issue #802 because the PR changes no Markdown file.

## Domain modular architecture — Changes required (1)

Authority is accepted ADR-0006 (project/path authority), accepted ADR-0040 (shared-core ownership and dependency direction), `CONTEXT.md`, and issue #802 (`ready-for-agent`). Observed profile: command-group slices call a shared project/path foundation; `gda.errors` owns the failure taxonomy and depends on `gda.models`; `gda.models` depends on `gda.project`.

### Required change — remove the hidden reverse dependency cycle

`src/gda/project.py:581-592` imports `gda.errors` inside `containment_refusal`; `gda.errors:55-62` imports `gda.models`; `gda.models:24` imports `gda.project`. A function-local import delays the cycle at runtime but does not remove the source dependency or the inverted `foundation -> errors` edge. The new module also has two reasons to change: path/ownership policy and Error-envelope taxonomy.

Impact: this violates ADR-0040's one-way core graph, makes import safety depend on call timing, and makes a taxonomy/evidence-shape change reach into the path foundation. Smallest practical repair: let `gda.project` return a small inner-owned containment decision (including the chosen violation and coordinates), and let one `gda.errors` function convert it to `Failure`. Commands still make one call; ordering and path knowledge remain single-sourced; dependencies stay `errors -> project/models`.

Structural checks: **Completeness** passes (all three call paths and both outcomes have owners); **Orthogonality** and **DRY** improve at the command sites but fail at the new project/error ownership overlap; **Extensibility** fails because changing the failure taxonomy now changes foundation; directed downward and horizontal communication fail at the reverse edge. No new Bounded Context or tactical DDD construct is justified.

Load-bearing claim `ARCH-807-1` — “the gate placement preserves the accepted core dependency direction”: lifecycle sources are issue #802 (`ready-for-agent`) and ADR-0040 (`accepted`); claim evidence state is `open` (the project defines no stronger evidence-state vocabulary); evidence is the import chain above; the disconfirming condition (a `project -> errors` edge) is present; the unresolved need is to choose and validate a non-cyclic owner split. Retained architecture: command slices depend downward on project/errors; project must not depend back on the Error-envelope layer.

## Standards — Changes required (2)

**Hard violation — `src/gda/project.py`, `containment_refusal` lines 581–592:** the foundation module imports `gda.errors` to construct failure envelopes. ADR-0040 §4 keeps all `*_failure` constructors in `errors.py`; §5 mandates `errors/models -> foundation`. This introduces `project -> errors -> models -> project`; the deferred import merely hides the acknowledged cycle. Keep the ordered path decision pure in `project.py`, then map it to envelopes from `errors.py`.

**Judgement call — possible Divergent Change in the same hunk:** `project.py` now changes for both path policy and failure-envelope taxonomy.

No other documented-standard violation or baseline smell was found in the changed command/test hunks. Tooling-enforced style was excluded.

## Spec — Changes required (1)

**Scope creep:** issue #802 states, “Behavior-preserving: both refusal envelopes, their codes, evidence, and remediation stay byte-identical” and “this is internal orchestration only.” Yet `project_absolute()` now uses total `_expand_user()`, and `script validate` applies it to result-root normalization. For a valid literal project directory named `~unknown-user`, merge-base raises `RuntimeError`/exit 1, while head emits a structured `target_outside_project` envelope/exit 4. This is an unrequested bug fix, not behavior-preserving orchestration. Preserve the former expansion semantics here or explicitly widen #802 and add regression coverage.

No missing acceptance-criterion implementation was otherwise found, and no other scope creep or incorrect implementation was found.

## Validation

- Exact-head targeted unit checks: containment/project `132 passed`; broader affected command set `283 passed`.
- Exact-head fast suite: `2416 passed, 2 skipped, 638 deselected`; the wheel test initially failed only because the default uv cache was sandbox-inaccessible, then passed with `UV_CACHE_DIR=/tmp/pr807-uv-cache`.
- Real Godot 4.6.3 containment E2E: `19 passed`.
- Ruff on all four changed files: check passed; format check passed.
- Public surfaces, base vs head: schema identical (`1,038,261` bytes, SHA-256 `27663f01e3e559b21c9202b6db4708789d3b9743b001b5f90ac47bfb49e527f9`); JSON help identical (`5,584` bytes, SHA-256 `a7a9b440cf5a5ee8ddcb440421b6041eff5f35f12d578d0a8ac180d65248cf3c`).
- Independent literal-`~user` CLI counterexample reproduced at both immutable revisions: base exit 1 traceback; head exit 4 structured refusal.
- Current exact-head CI: all required reported checks pass; hosted Godot E2E is skipped. CI Pyright is green. Green CI does not cover the architecture/spec findings above.
- Not verified: Windows/Linux path behavior and the full local E2E suite beyond `tests/test_e2e_res_containment.py`.

**Two-axis summary:** Standards: 2 findings (worst: hard reverse-dependency/cycle violation). Spec: 1 finding (worst: observable `~user` behavior change outside the behavior-preserving scope).

qinhaihong-red:
## Round-2 response

New head: `e154e320d` (one commit on the reviewed `83e885ae0`).

| The gate on `gda.project` inverts ADR-0040 §5 (`project → errors → models → project`, hidden by a deferred import); issue #802 prescribed the conflicting shape (Artifact-1 / DM-1 / Standards-1) | **Fully adopt** | ADR-0040 §4/§5 verified; both this review and round 1's architecture note converge on the same split | The decision stays with the path authority — `gda.project.containment_violation` returns the fired half with its coordinates (`ForeignOwnerViolation` / `OutsideRootViolation`), ordering and normalization unchanged — and `gda.errors.containment_refusal` maps it to the envelopes. Commands call the taxonomy; the deferred import is gone; direction is `errors → foundation`. The AST guard follows the gate to `gda.errors` and additionally pins that the mapper calls the decision function and builds nothing else. Issue #802 reshaped with a dated round-2 note | 56 consistency + 2419 unit green; envelopes spot-checked byte-identical vs the previous head on real CLI refusals; `gda schema` hash unchanged; 19 containment e2e green; ruff + bare pyright clean |
| The `~<no-such-user>` project spelling changes behavior (base: `RuntimeError` exit 1; head: structured refusal exit 4) outside the issue's behavior-preserving claim (Artifact-2 / Spec-1) | **Partially adopt** | Counterexample accepted as measured | Behavior KEPT — restoring a `RuntimeError` leak to honor the boundary sentence would re-ship a crash; the issue is widened instead (AC4 names the accepted case) and a regression pins the structured refusal (`test_a_project_directory_literally_named_like_a_home_reference_is_served`) | The new regression green; the widening recorded on #802 |
| AC4 still says "both commands" (Artifact-3) | **Fully adopt** | Verified | AC4 → "all three commands" on #802 | Issue diff |

Disclosure: the first containment-e2e run after the restructure failed 17/19 with `ImportError: cannot import name 'containment_refusal' from 'gda.project'` — stale `__pycache__` bytecode from a same-mtime-second file rewrite (the source was correct; the unit suite, which imports in-process after the edit, was green). Clearing `src/**/__pycache__` and rerunning: 19/19. Environment evidence, not a product defect.

qinhaihong-red:

Reviewed base tip `64136719c6a615ab94887aeeb3447f895511744e`, merge base `fe72700ecb6d12b24474560c4b8599266df6ebcd`, and current head `e154e320da6c470fc5c6ae0dbc6f29ec34cc5278`. The implementation-level dependency repair is sound, but the current specification, regression evidence, and explanatory artifacts are not yet reconciled.

1. **Issue #802 AC1 still requires the obsolete ownership model.** AC1 says one function in `gda.project` owns both the ordering decision and both refusal envelopes' construction. The revised issue body, accepted ADR-0040, and current implementation instead place the pure decision in `gda.project.containment_violation` and envelope mapping in `gda.errors.containment_refusal`. Impact: the acceptance criteria cannot be literally satisfied without reintroducing the dependency violation. Smallest fix: update AC1 to state this two-owner split.

2. **The accepted `~<no-such-user>` widening is not regression-pinned by the current test.** `tests/test_res_containment_consistency.py:469-484` constructs the project as the absolute path `tmp_path / "~gda_no_such_user_802"`. `Path.expanduser()` does not interpret a tilde inside an absolute path, so the former direct `expanduser()` implementation also passes this test. Smallest fix: change into `tmp_path`, pass the relative spelling `Path("~gda_no_such_user_802")`, and preferably assert the structured refusal through one affected command.

3. **The PR body still describes the removed implementation.** It refers to `gda.project.containment_refusal(...) -> Failure | None`, a function-local `gda.errors` import, and the former cycle, while also calling the whole change behavior-preserving without acknowledging the now-accepted widening. Update the PR body to describe the current decision/mapper split and the one accepted behavior change.

4. **Code-contract prose still assigns responsibilities inaccurately.** `project.py:511-514` refers to a nonexistent same-module `containment_refusal`; `script.py:2148-2152` says `gda.errors.containment_refusal` owns ordering; and `errors.py:800-809` says only two resolved-project commands use the mapper although `script run` also uses it. Correct these descriptions to match the current owners and three call sites.

**Terminology and prose assessment:** I inspected issue #802, the current PR body, and all changed prose, docstrings, and comments. Terminology sources were `CONTEXT.md` and ADR-0006/0031/0040; ASD-STE100 was the writing reference. The remaining substantive problems are the ownership, scope, and causal inaccuracies above. No additional terminology issue was found, and no automated ASD-STE100 checker was used.

### Domain modular architecture — Changes required

The code now passes completeness, orthogonality, DRY, directed-dependency, and extensibility checks: `gda.project` owns path decisions and coordinates, `gda.errors` owns envelope taxonomy, and the three command sites use the mapper without a reverse dependency. The remaining architecture finding is the authoritative AC1 conflict above: the governing issue still requires the forbidden foundation-to-envelope responsibility.

### Standards — Pass

The ADR-0040 dependency violation and the possible Divergent Change are closed. The current direction is `errors -> project/models`, and the modules now have distinct reasons to change. No new actionable Fowler smell was found.

### Spec — Changes required

- The promised regression does not exercise the relative `~user` spelling that triggered the old failure.
- AC1 contradicts the reconciled design and is not satisfied by the current implementation.

### Validation

- Focused unit tests: `302 passed`; full non-E2E suite: `2417 passed, 2 skipped, 638 deselected`.
- Real-Godot containment E2E: `19 passed`; a wider script/resource E2E run reached `63 passed` with no failure before it was stopped.
- Ruff check/format, Pyright, and `git diff --check`: passed.
- Schema and relevant root/script/resource help surfaces: byte-identical between merge base and head.
- Discriminating CLI replay: merge base with relative `--project '~gda_no_such_user_802'` exits 1 with a traceback; current head exits 4 with `target_outside_project`. Replacing the new helper with direct `expanduser()` still lets the current absolute-path test pass.
- Current CI: `failed=[]`, `pending=[]`; hosted Godot E2E is skipped.

Not independently checked: Windows/Linux path behavior and the remainder of the full local E2E suite.

qinhaihong-red:
## Round-3 response

New head: `bb7767ebe` (one commit on the reviewed `e154e320d`).

| Issue #802 AC1 still requires the obsolete one-function-on-`gda.project` model | **Fully adopt** | Verified | AC1 reworded to the two-owner split (decision on `gda.project`, mapper on `gda.errors`), with a dated marker | Issue diff |
| The `~<no-such-user>` regression does not discriminate — the tilde sits inside an absolute path, which `Path.expanduser()` leaves alone, so the old implementation passes it too | **Fully adopt** | Reproduced: `Path("~u").expanduser()` raises, `Path("/t/~u").expanduser()` does not | The test now chdirs into `tmp_path` and spells the project RELATIVE — the spelling that raised `RuntimeError` at the merge base — and additionally asserts the structured refusal through `_script_validate_recipe`. Red-proofed: swapping `project_absolute`'s total expansion back to `Path.expanduser()` fails exactly this test | 1 red under the mutation, 56 consistency + 2419 unit green restored |
| The PR body still describes the removed implementation (gate on `gda.project`, deferred import) and calls the change behavior-preserving without the accepted widening | **Fully adopt** | Verified | Body updated: the decision/mapper split, the superseded round-1 shape marked as such, and the widening named as the one accepted, pinned behavior change | Body diff |
| Three prose passages assign responsibilities inaccurately (`project.py` cites a nonexistent same-module gate; `script.py` says the mapper owns ordering; `errors.py` counts two commands) | **Fully adopt** | Verified at the cited lines | `project_absolute` cites `containment_violation`; the recipe says the mapper maps the decision's ordering; the builder docstring counts the three commands reaching it through the gate | 2419 unit, ruff/format, bare pyright clean |

qinhaihong-red:

Reviewed base tip `64136719c6a615ab94887aeeb3447f895511744e`, merge base `fe72700ecb6d12b24474560c4b8599266df6ebcd`, and current head `bb7767ebea21c7f01024699a248208dfb2eef1cf`. I treated the author reply as a claimed-fix list and independently replayed the prior findings against the current artifacts, implementation, and runtime behavior. The dependency and gate/mapper ownership repairs now pass; the remaining current-head residue is below.

1. **The actual literal-tilde behavior widening is broader than issue #802 and the PR body authorize.** AC4 and the PR body describe one accepted widening: an outside target under a relative `~<no-such-user>` project changes from the base's `RuntimeError` to a structured `target_outside_project` refusal. The same total project normalization also changes accepted inside-target paths:
   - `script validate --project '~gda_no_such_user_802' main.gd`: base exits 1 with `RuntimeError`; this head exits 0 and reports `valid: true` with the real engine.
   - `resource import --dry-run --project '~gda_no_such_user_802' pic.svg`: base exits 1 with the same `RuntimeError`; this head exits 0 with the normal dry-run result.
   - The comparable `script run` inside-target case is unchanged and succeeds at both revisions.

   The new tests pin only the outside-target refusal, not these two accepted-path changes. Smallest closure: do not restore the crash; update issue AC4 and the PR body to authorize the complete literal-tilde behavior change, then add inside-target regression coverage for `script validate` and `resource import`.

2. **`src/gda/errors.py:800-810` and the test prose still misstate ownership and builder reachability.** The current docstring says `script validate`, `script run`, and `resource import` share `target_outside_project_failure` through the mapper. Only `script validate` and `resource import` can reach that builder. `script run` rejects lexical escapes through `script_escapes_project_failure` before project resolution; an accepted path is already canonical `res://`, so the later mapper's outside-project half is unreachable. This is established by `script.py:1263-1284,1552-1575`, `project.py:305-311`, `tests/test_res_containment_consistency.py:329-348`, and ADR-0031. Also, `tests/test_res_containment_consistency.py:509-512` still says the gate owns both ordering and envelopes; the project gate owns the ordered decision, while the errors mapper owns envelope construction.

   Smallest closure: describe the two commands that can reach the outside builder, then state separately that `script run` calls only the mapper's ownership half after `_project_scoped_res_path`; correct the test comment to preserve the decision/mapper ownership split.

3. **The PR body's review evidence is stale.** It says the consistency suite gains five tests although the exact diff adds eight `test_*` functions; attributes the ordering proof to a relative-project row that its current test comment says is non-discriminating; reports `2413 passed` instead of the current `2417 passed, 2 skipped`; and claims every path except one named case is behavior-preserving despite the two inside-target changes above. Update the body to the current test inventory, discriminating evidence, validation result, and complete widening scope.

**Terminology and prose assessment:** I inspected the complete current issue, PR body, and all changed documentation, docstrings, comments, and test prose. Terminology sources were `CONTEXT.md` and ADR-0006/0031/0040; ASD-STE100 was the writing reference. Apart from the scope, reachability, ownership, and evidence inaccuracies above, no additional substantive terminology or prose issue was found. No automated ASD-STE100 checker was used.

The implementation now has the correct owners and directed dependencies: `gda.project` owns the containment decision and coordinates, `gda.errors` owns failure taxonomy and envelope mapping, and command modules consume them without a reverse import. The `script run` pre-resolution lexical gate remains the distinct ADR-0031 boundary. Completeness, orthogonality, DRY, dependency direction, and current reasons to change pass.

One P2 code-contract finding remains: `errors.py:800-810` confuses “calls the mapper” with “can reach this concrete builder” and contradicts the implementation, its invariant test, and ADR-0031. This is a substantive maintenance-contract error, not a count-word preference. No additional actionable Fowler smell was found.

### Spec — Changes required

The implementation changes valid inside-target behavior for `script validate` and `resource import`, while AC4 authorizes only the outside-target refusal change. Either preserve the former accepted-path behavior or, preferably, authorize the complete non-crashing literal-tilde handling and regression-pin both inside-target paths.

### Reconciliation status

The AC1 ownership split, import graph, and ADR-0006/0031/0040 direction are now synchronized. The remaining unsynchronized edges are: issue/PR → actual inside-target behavior; PR body → current tests and validation; and `errors.py`/test prose → actual builder reachability and decision/envelope ownership. The command catalog and `CONTEXT.md` remain unchanged and do not need a new public surface for this PR.

### Validation

- Focused unit tests: `409 passed`; containment consistency: base `27 passed`, head `56 passed`.
- Full non-E2E suite: `2417 passed, 2 skipped`.
- Real-Godot containment E2E on Godot 4.6.3: `19 passed`.
- Ruff check/format, Pyright, and `git diff --check`: passed.
- Schema and root/script/resource/validate/run/import help surfaces: byte-identical between merge base and head.
- Current CI: `failed=[]`, `pending=[]`; 16 checks succeeded and hosted Godot E2E is skipped. The affected containment paths passed locally with the real engine.

Not independently checked: Windows/Linux path behavior and the full local Godot E2E suite outside containment.

qinhaihong-red:
## Round-4 response

New head: `e82d92243` (one commit on the reviewed `bb7767ebe`).

| The literal-tilde widening is broader than authorized — inside-target ACCEPTED paths change too (`script validate` / `resource import` crashed at base, succeed at head), and only the refusal half was pinned | **Fully adopt** | Your real-engine replays accepted; the crash is the same base `RuntimeError` on every path reaching the containment question | The widening is authorized WHOLE on #802's AC4 (dated round-4 marker) and pinned engine-free at three depths in `test_inside_targets_under_a_literal_tilde_project_are_served`: the decision accepts, the import gate maps the address, and the validate recipe gets PAST containment (reporting an injected `binary_not_found` — an environment verdict, not a containment one, not a crash) | Both tilde pins go RED under the swap-to-`expanduser` mutation (re-run); 57 consistency + 2420 unit green |
| `errors.py:800-810` says three commands share the outside builder; only two can reach it — `script run`'s accepted address is canonical `res://`, so only the mapper's ownership half is reachable; the test comment still says the gate owns ordering AND envelopes | **Fully adopt** | Verified against `_project_scoped_res_path`'s guarantee and the fold-inertness test | The builder docstring now names the two reaching commands and states `script run`'s ownership-only reachability; the builder-set comment carries the decision/mapper ownership split | 2420 unit green; the round-3 overcorrection (two → three) is corrected to the reachability truth (two through the gate, one ownership-only) |
| The PR body's evidence is stale (five vs nine tests; ordering proof attributed to the non-discriminating relative row; `2413 passed`; incomplete widening scope) | **Fully adopt** | Verified: the exact diff adds nine `test_` functions | Body updated: nine tests with the ordering proof attributed to `test_ownership_wins_when_both_halves_of_the_gate_fire`, current suite numbers, and the complete widening scope | Body diff |

qinhaihong-red:
## Round-5 response (wording residuals)

New head: `e05f6bb7b` (one test commit on `e82d92243`).

| AC4's "EVERY path" overclaims — `script run`'s ordinary inside path succeeds at both revisions | **Fully adopt** | Verified against the merge-base source: `script run`'s ownership-only copy expanded the project only inside the foreign-owner branch, so its ordinary paths never reached the crash | AC4 now carries the measured command/outcome MATRIX (validate and import: crash → refusal/served; script run: unchanged on both its rows), with the overclaim's cause named | Issue diff |
| The PR body misattributes the normalization discrimination and reports `2420 passed` where the exact head gives `2418 passed, 2 skipped` | **Fully adopt** | The row holds only the cwd-anchoring agreement; resolve-before-probes is caught by the round-2 spelled-project test. Suite numbers are environment-split: author machine 2420/0 skipped, reviewer sandbox 2418/2 skipped — same set | Both body passages corrected: the two mutations attributed to their actual catching tests; both suite readings stated with their environments | Body diff |
| The mapper-purity assertion only proves two names appear — it cannot support "builds nothing else" | **Fully adopt** (the stronger variant) | Verified: `in`-membership says nothing about other calls | The assertion is now EQUALITY with the complete allowed-call set — the decision, the two builders, `isinstance` — so any new call inside the gate fails by construction, and the comment claims exactly what the assertion checks | 57 consistency + 2420 unit green at `e05f6bb7b` |

PR #809, opened 2026-09-02
fix(gda): refuse a container float the parser destroys, on the raw text (#805)

### LINKED ISSUE #805: a destroyed float inside a JSON container bypasses the #772 refusal; the packer's float32 elision needs its disclosure decision

qinhaihong-red:
## What

A float inside a JSON container `--value` bypassed #772's refusal and was stored silently. It now
takes the same rule, with the same code and the same message, on all four coercing commands.

Also records the second half of #805: the scene packer's float32 elision keeps its documentation
qualification as the permanent contract, with rationale, and no mechanism is added for it.

## Why

#772 refuses a `--value` literal Godot's parser destroys — read as `0.0` when the literal does not
denote zero, or as `NaN` at all. That rule walks the components `_coerce_float` parses. A container
value never reaches it, so the failure #772 calls unacceptable stayed alive on one path.

### Red proof, measured at the base (`dcae947a5`) on Godot 4.6.3

```
$ gda node set main.tscn --node . --property d --value '{"a": 1e-320, "b": 0.000000000000000001}' --json
{"scene_path":"…/main.tscn","path":".","property":"d","type":"Dictionary","value":{"a":0.0,"b":0.0}}
exit=0

$ gda node set main.tscn --node . --property a --value '[1.0, 5e-324]' --json
{"scene_path":"…/main.tscn","path":".","property":"a","type":"Array","value":[1.0,0.0]}
exit=0
```

A following `node get` read the same zeros back: the scene file really held them.

### After

```
$ gda node set main.tscn --node . --property d --value '{"a": 1e-320, "b": 0.000000000000000001}' --json
{"error":{"category":"operation","code":"uncoercible_value","message":"cannot coerce value
{\"a\": 1e-320, \"b\": 0.000000000000000001} to Dictionary for property d on node . — Godot's own
float parser reads 1e-320 as 0.0, so the write would store a number you did not send; gda refuses
it instead of changing your value silently. Try the same value in scientific notation …"}}
exit=4

$ gda node set main.tscn --node . --property a --value '[1.0, 5e-324]' --json
… "code":"uncoercible_value" … "reads 5e-324 as 0.0" …   exit=4
```

The scene file is untouched by a refusal, and the message is the scalar path's, naming the ONE
offending literal rather than the whole argument.

## The scanning rule

There is no per-element traversal to hook: a container's coercion is `JSON.parse_string` as the
gate plus one atomic `str_to_var(raw)`, so by the time a float exists its literal is gone. The
literals are therefore read out of the RAW `--value` text — **and only after the gate has accepted
it**. That ordering is half the rule, not an implementation detail.

Outside strings, valid JSON spells only structure, `true`/`false`/`null`, and numbers, so a maximal
run of number characters is a number token; the contents of every JSON string are skipped, escapes
honoured. Each token is judged by the *existing* `_float_literal_is_destroyed` — one policy
predicate, not a second one.

### The three traps, and their dispositions

| Trap | Disposition | Why |
| --- | --- | --- |
| A STRING value that looks numeric — `{"a": "1e-320"}` | **Not refused**, stored as the six-character string | The scan skips string contents. No float is parsed there, so there is nothing to destroy. Verified on the engine. |
| KEYS — `{"1e-320": 1.0}` | **Not scanned**, member written unchanged | Every JSON key is a string, so the same one rule covers it. The `1.0` beside it is the only number present. Escapes are honoured while skipping, so `{"a\": 1e-320 fake": 1.0}` — valid JSON whose *key* holds that text — is not refused either; without escape handling that write would have failed. |
| `str_to_var`'s richer syntax — `{"a": Vector2(1e-320, 0)}` | **Unreachable, so deliberately not scanned for**; stays the plain `uncoercible_value` it has always been | Measured: `JSON.parse_string` rejects it (`Expected 'true', 'false', or 'null', got 'Vector'`), so the gate refuses the text before `str_to_var` runs. Scanning it would attach a float explanation to a *syntax* refusal — the false cause `test_the_note_explains_only_the_refusal_it_diagnosed` exists to prevent. |

The same attribution rule keeps the fidelity note off a `--value` that is not a container at all
(`'0.000000000000000001'` on a `Dictionary` property — refused by the gate) and off JSON of the
OTHER container type (`'[1e-320]'` on a `Dictionary` property — refused on the type). Both are
pinned as tests.

## Mirror

The scan block lives inside the byte-identical mirrored region, so it was duplicated **verbatim**
into `src/gda/harness/gda_harness.gd` (mechanically, by extracting the marker-delimited block from
`operations.gd` and writing it over the harness copy). `tests/test_harness_coercion_mirror.py` is
green. `HARNESS_VERSION` 17 → 18, the pinned bytes re-pinned, and
`scripts/harness_version_guard.py dcae947a5 32fdf67aa` reports a valid transition.

## The packer-elision decision (the other half of #805)

**Default affirmed: the catalog's measured qualification IS the contract.** gda adds no read-back
verification and no result-level disclosure field. Recorded in `docs/command-catalog.md` beside the
qualification it governs — the existing authority for this fact, so no new home and no ADR. My
rationale does **not** argue for a disclosure field, so none is proposed here; the four reasons, in
the order they bind:

1. It is the **engine's serializer policy**, applied to any write the engine makes. The value gda
   coerced was exact, so a gda-side refusal or annotation would blame the wrong stage.
2. gda cannot **answer** the question at coercion time without either a write-then-read-back engine
   round-trip per write (disproportionate to this band, and out of scope by decision) or a
   re-implementation of `PropertyUtils::is_property_value_different` — a second opinion about an
   engine function, exactly what the refusal rule avoids by *running* the parser instead of
   modelling it.
3. It is **not the loss the rule exists to stop**: the file records nothing, so the property keeps
   the default its own source declares, rather than holding a number nobody sent.
4. The disclosure is **not free** — a result-level field touches the result models, `--schema` and
   the README family, for a rare engine-owned band.

A written trigger is recorded instead: if a case appears where an agent acted on the `set` echo and
the elision made the action wrong, that is a separate issue with the evidence to justify the field.
Until then the remedy is the one the round-trip claim already names — read it back with `node get`.

## Surface diff

- **Behaviour (4 commands).** `node set`, `resource set`, `project set` → `uncoercible_value`
  (exit 4); live `game set` → `live_uncoercible_value` (exit 6). Target untouched in every case.
  No new code, no new message template — the existing `_float_fidelity_note` is reused.
- **`gda schema`: byte-identical to the base.** Measured, not assumed: `sha256` of `gda schema`
  output is `27663f01e3e559b21c9202b6db4708789d3b9743b001b5f90ac47bfb49e527f9` on both
  `dcae947a5` and this head.
- **Docs.** `CONTEXT.md` (the *Value projection* entry's unconditional refusal sentence becomes
  true only with this change — aligned, with the container coverage named);
  `docs/command-catalog.md` (the falsified "one path does not reach the check at all" sentence
  replaced by the container rule and the three dispositions, plus the elision decision);
  `src/gda/skill/SKILL.md` (number-fidelity passage, the `Dictionary`/`Array` `--value` bullets,
  and the live `game set` sentence); `src/gda/live_numbers.py` docstring — the number-domain
  authority carried the same falsified sentence.
- **No Python behaviour change.** The refusal is OPERATION-source, reported from GDScript through
  the existing sentinel envelope. The only non-doc Python edits are `HARNESS_VERSION` and tests.

## Verification

- `uv run pytest tests/ -q -m "not e2e"` → **2392 passed**.
- `tests/test_harness_coercion_mirror.py` → 3 passed.
- Targeted e2e on the real engine, `-m e2e -k "write_value_fidelity or coercion"` → **9 passed**,
  including `test_game_set_refuses_the_same_literals_against_a_real_daemon` (real daemon, real
  session, container cases added).
- Regression bands on the real engine: `test_e2e_node.py`, `test_e2e_resource.py`,
  `test_e2e_project.py`, `test_e2e_headless_number_reads.py` → 169 passed; `test_e2e_game.py`,
  `test_e2e_live_number_transport.py` → 14 passed.
- `ruff check` / `ruff format` clean; `pyright` 0 errors.
- All four commands additionally driven by hand against Godot 4.6.3 before the tests were written
  (dict, array, nested, string value, key, escaped-quote key, exact values, constructor,
  wrong-container-type).

Closes #805


qinhaihong-red:
## Full independent adversarial review — PR #809

**Pinned head:** `32fdf67aa` (`fix/805-container-float-refusal`)
**Stacked base (diffed against, not `dev`):** `fix/804-walk-engine-skip-alignment` @ `dcae947a5`
**Scope:** this PR's own delta only (9 files, +466/−52). The B1 slice beneath is out of scope.
**Independence:** the PR body, its comments and the author's replies were excluded from evidence; every claim below was re-measured against the real engine (Godot 4.6.3, `~/Applications/Godot.app`) from detached scratch worktrees.

### Verdict

| Axis | Verdict | Notes |
| --- | --- | --- |
| **Artifact** (`artifact-review`) | **Pass with minor** | CONTEXT/catalog/SKILL/`live_numbers` are mutually consistent and the CONTEXT "never stored" sentence is now true. One completeness claim is falsified (P2‑1); one ordering nit (P3‑2). |
| **Architecture** (`design-domain-modular-architecture`) | **Pass** | One policy authority (`_float_literal_is_destroyed`) is reused, not re-derived; the scan owns only tokenization. Placement inside the mirrored block is correct. |
| **Standards** | **Pass** | Mirror parity guard green, mirrored region byte-identical, harness version + SHA pins bumped together, schema byte-identical to base. |
| **Spec (#805)** | **Pass with minor** | All four ACs met and independently verified on a real engine. Two of the four coercing commands are behaviourally verified by me but not test-pinned (P3‑1). |

**Findings: 0 P1 · 1 P2 · 4 P3.**

---

### The defect is real and the fix works (red-proof, measured by me)

Base `dcae947a5`, same probe project, same engine:

```
$ gda node set main.tscn --node . --property d \
    --value '{"a": 1e-320, "b": 0.000000000000000001}'
{"...","property":"d","type":"Dictionary","value":{"a":0.0,"b":0.0}}      # exit 0
$ gda node get main.tscn --node .
  {'name': 'd', 'type': 'Dictionary', 'value': {'a': 0.0, 'b': 0.0}}      # written to the .tscn
```

Head `32fdf67aa`: `uncoercible_value`, exit 4, note naming `1e-320`, scene byte-identical afterwards.

---

## P2

### P2‑1 — A JSON **duplicate key** makes the scan refuse a write the engine would have kept, and the block comment claims that class is closed

**Location:** `src/gda/ops/operations.gd:6634` (and its byte-identical mirror `src/gda/harness/gda_harness.gd:2204`); `docs/command-catalog.md` "Three consequences follow from that scanning rule, each a deliberate disposition".

**Evidence.** Godot 4.6.3, measured directly:

```
RAW={"a": 1e-320, "a": 2.0} || JSON=Dictionary:{ "a": 2.0 } || STV=Dictionary:{ "a": 2.0 }
```

The parser **discards** the destroyed literal — no destroyed value can ever be stored from this text. Base stores `{"a": 2.0}` faithfully. Head refuses it, on both channels:

```
head, headless:  uncoercible_value      + "Godot's own float parser reads 1e-320 as 0.0"
head, live daemon: live_uncoercible_value + the same note   (real session, /root/Main)
```

The comment states:

> Reading text needs one rule to be safe, and it is STRING-AWARENESS … That single rule disposes of **both** ways a text scan can refuse a write the engine would have kept faithfully.

There is a third way, and it is live in the code. The catalog's list is likewise presented as exhaustive ("Three consequences … each a **deliberate** disposition") and this disposition was not made.

**Impact.** The over-refusal itself is in the safe direction and the input is pathological, so this is not a correctness emergency. What is actionable is that a comment and a spec paragraph whose entire value is stating the rule exactly both assert a completeness that measurement falsifies — the same standard this PR holds itself to everywhere else.

**Smallest fix — no code change needed.** Either
(a) name it as an accepted over-refusal: one clause in the block comment ("…and a third, accepted: a member the parser then DISCARDS — a duplicate key — is still scanned, so the write is refused although nothing destroyed would be stored"), plus a fourth catalog bullet; or
(b) drop the exhaustiveness: "both ways" → "the two ways ordinary input invites", and "Three consequences" → "The consequences that matter in practice".

---

## P3

### P3‑1 — AC1 says "ALL FOUR commands"; tests pin two of them

`CONTAINER_REFUSED` / `CONTAINER_ACCEPTED` / `CONTAINER_PLAIN_REFUSAL` drive `node set` only; the live band adds two refusals + two traps for `game set`. `test_project_set_shares_the_refusal` still carries only the **scalar** literal, and `resource set` has no case in this module at all (pre-existing).

I verified the two unpinned commands by hand on the real engine, and both are correct:

```
project set gda/probe/dict --value '{"a": 1e-320}'   -> uncoercible_value + note "reads 1e-320 as 0.0"
project set gda/probe/arr  --value '[0e600]'         -> uncoercible_value + note "reads 0e600 as NaN"
project set gda/probe/dict --value '{"a": "1e-320"}' -> OK  {"a":"1e-320"}
project set gda/probe/dict --value '{"1e-320": 1.0}' -> OK  {"1e-320":1.0}
resource set probe.tres --property d --value '{"a": 1e-320}' -> uncoercible_value + note
resource set probe.tres --property a --value '[5e-324]'      -> uncoercible_value + note
resource set probe.tres --property d --value '{"a": "1e-320"}' -> OK
resource set probe.tres --property a --value '[1e-300]'        -> OK  [1e-300]
```

The structural argument (one `_coerce_value`, one shared block, one parity guard) is sound, so this is a coverage nit rather than a risk. Cheapest close: two lines appended to the existing `project set` test.

### P3‑2 — The catalog's new elision decision splits the #772 rule in two

Reading order in `docs/command-catalog.md` is now: refusal rule → packer-elision caveat → **new 14-line elision decision record** → "The rule keys on what the parser PRODUCED…" (which *continues* the refusal rule) → **new container rule**. The decision record is correctly adjacent to the caveat it decides, but the pair now sits inside the refusal rule's body. Moving caveat + decision to *after* the container rule keeps the rule contiguous and costs nothing.

### P3‑3 — The note-attribution walker's container arm is now an exact duplicate of the coercer's two steps

`_coerce_dictionary` / `_coerce_array` run `JSON.parse_string` gate + `_destroyed_json_number`; `_destroyed_float_literal`'s `TYPE_DICTIONARY, TYPE_ARRAY` arm runs the same two calls again. For the *scalar* arms the second walk re-derives a genuinely different shape (split, arity, hex) and the independence is what keeps the note honest — that design (#772) is sound and I am not asking for it to change. For containers the second walk re-derives nothing; it buys a second JSON parse and a second scan. Flagging only so the trigger is written down: if a third container-shaped type ever reaches this rule, hand `_coerce_value` a refusal reason rather than walking a third time.

### P3‑4 — Informational, no action: engine warning noise in `diagnostics`

A container `0e600` refusal now carries **5** `Exponent too high` engine warnings in the structured `diagnostics` field (the scalar path on base already carried **3**). Same pre-existing class, a direct consequence of P3‑3's second walk. Not a regression.

---

## What survived the attack (all verified on the real engine at this head)

The scanner is sound under everything I could throw at it. Verdicts below match the scalar path's rule in every case.

**String-awareness and escape handling — no desync found**

| payload | verdict | correct? |
| --- | --- | --- |
| `{"a\": 1e-320 fake": 1.0}` (escaped quote inside a KEY) | OK, stored verbatim | ✅ |
| `{"k": "a\" 1e-320 fake"}` (escaped quote inside a VALUE) | OK | ✅ |
| `{"1e-320": 1.0}` / `{"a": "1e-320"}` | OK | ✅ |
| `{"a": "😀 1e-320"}`, `{"a": "é1e-320"}` (multi-byte / surrogate) | OK | ✅ |
| `{"héllo": 1e-320}` (multi-byte key, number OUTSIDE it) | REFUSED + note | ✅ |

**Godot's JSON leniencies do not open a hole**

| payload | JSON gate | verdict | correct? |
| --- | --- | --- | --- |
| `{"a": 1e-320,}` (trailing comma — Godot accepts) | accepted | REFUSED + note | ✅ |
| `{"a": 01e-320}` (leading zero — Godot accepts) | accepted | REFUSED + note | ✅ |
| `{"a": 1.0} 1e-320` (trailing garbage) | **rejected** (`Nil`) | plain refusal | ✅ scan can never read text outside the parsed value |

**The constructor unreachability claim is true, and the catalog's quoted error is verbatim**

```
JSON.new().parse('{"a": Vector2(1e-320, 0)}')
  -> err=43  msg="Expected 'true', 'false', or 'null', got 'Vector'"
```
`{"a": Vector2(1e-320, 0)}` and `{"a": Vector2(1.0, 0)}` both get the **plain** `uncoercible_value` (no fidelity note) headless and live. `{"a": nan}` / `{"a": inf}` are likewise rejected by the gate, so no NaN can enter a container by any spelling.

**The `true`/`false` "lone e" token claim is true:** `"e".is_valid_float() == false`, so `{"a": true, "b": false, "c": null}` and `{"a": [true, 1.0, false]}` both store.

**Numeric edges match the scalar path**

`1e-300` ✅ stores exactly · `1e-18` ✅ · `1e308` ✅ stores `1e+308` · `1e400` ✅ stores `inf`, echoes `null` (documented overflow disclosure) · `-0.0` ✅ stores, echoes `0.0` (documented writer residual) · `1e+5` ✅ · `1E-320` ✅ refused · `0e600` ✅ refused as NaN · `{"a": {"b": [0e600]}}` and `[[2.2250738585072014e-308]]` ✅ refused (nesting is not a hiding place).

**No remaining silent-destruction path exists.** `_coerce_value` accepts exactly `bool / int / float / String / StringName / Dictionary / Array / Vector2 / Vector2i / Color`; every other type returns `null` before any parse. So FLOAT + VECTOR2 + COLOR (scalar walk) and DICTIONARY + ARRAY (raw-text scan) is the complete float-bearing surface — the CONTEXT claim "Observing covers every literal a `--value` spells" is now literally true.

**Typed containers cannot smuggle a numeric-looking string into a float.** Measured: `Array[float].assign(["1e-320"])` yields size 0, so the existing size-mismatch check returns `null` (plain refusal). Same for `Dictionary[String, float]`. The string-skipping rule creates no hole here.

---

## Validation

**Mutation testing — 3 mutations applied to `operations.gd` at head, all caught:**

1. Remove the string-skipping branch entirely (scan keys and string values) → `test_node_set_refuses_a_destroyed_number_inside_a_container` **RED** on `{"a": "1e-320"}`.
2. Keep string-skipping, drop the `\\` escape branch → **RED** on `{"a\": 1e-320 fake": 1.0}`.
3. Remove the JSON gate from `_destroyed_float_literal`'s container arm → **RED** in two tests (container corpus + `test_the_note_explains_only_the_refusal_it_diagnosed`), the second showing the fidelity note leaking onto a non-JSON `--value`.

Tree restored to `32fdf67aa` afterwards (`git status --porcelain` empty).

**Suites run** (`pytest -p no:cacheprovider`, `UV_PROJECT_ENVIRONMENT` scoped to the scratch worktree):

- `-m "not e2e"` → **2392 passed**, 643 deselected.
- `-m e2e tests/test_e2e_write_value_fidelity.py -rs` → **8 passed** in 36s, including the live `game set` band against a real daemon and a real engine session.
- Mirror parity guard `tests/test_harness_coercion_mirror.py` → **green**; all five new/changed functions confirmed inside the `BEGIN/END shared coercion` markers in both files (`operations.gd:6312–6862`, `gda_harness.gd:1882–2432`).
- `tests/test_harness_install.py` → green; `HARNESS_VERSION` 17→18 and the pinned SHA256 bumped together.

**Schema:** `gda schema` byte-identical between `dcae947a5` and `32fdf67aa` (1 038 261 bytes each, `diff` empty) — AC4's "no schema diff expected" holds.

**Live band, driven independently of the repo tests** (`daemon start` → 11 payloads → `daemon stop`, real session): headless and live verdicts agreed on every payload, including the escaped-quote trap, the constructor edge, and the P2‑1 duplicate-key case.

**Cross-artifact consistency checked:** no ADR owns the #772 write-side rule (`grep` over `docs/adr/` finds no `#772`/`#805` reference), so catalog + CONTEXT + `live_numbers.__doc__` are the authority and all three were updated together. No README-family statement mentions container coercion, so none went stale. `src/gda/skill/SKILL.md` is the single SKILL copy in the repo.

**Elision half (AC3):** the catalog paragraph affirms the doc qualification as the permanent contract with four ordered reasons, defers a result-level disclosure field to a separate issue behind a written trigger, and introduces **no** read-back mechanism — `grep` confirms no new engine round-trip anywhere in the delta. Matches the issue's bounded default exactly.

**Not performed:** Linux/Windows execution (macOS only); no fuzzing beyond the ~60 hand-crafted payloads above; `gda-mcp` surface untouched by this delta and not exercised.

qinhaihong-red:
## Fix round — "Full independent adversarial review — PR #809"

**Rebase note.** The stacked base moved: B1 appended a review-fix commit, so this branch was
rebased `dcae947a5` → **`96bf1ca11`**. The rebase replayed one commit and **`docs/command-catalog.md`
did not conflict** — B1's edits sit in the walk/skip passages, this slice's in the number-coercion
passages. Parity guard + `harness_version_guard.py 96bf1ca11 HEAD` re-run immediately after: both
green. All measurements below are against the REBASED base.

**New head:** `ec7d39822` (fix round appended as its own commit; `f33c640d2` is the rebased
original).

Every claim was re-measured before deciding. 1 fully adopted, 2 fully adopted, 1 adopted as
prescribed, 1 declined with note.

| # | Finding | Decision |
| --- | --- | --- |
| **P2‑1** | Duplicate key over-refuses; two records claim the class is closed | **Fully adopted** — prose, no code change |
| **P3‑1** | `project set` / `resource set` not test-pinned for containers | **Fully adopted** |
| **P3‑2** | Elision decision splits the #772 rule's body | **Fully adopted** |
| **P3‑3** | Container arm's second walk re-derives nothing; write the trigger down | **Fully adopted** (trigger was NOT written; now is) |
| **P3‑4** | Engine warning noise in `diagnostics` | **Not adopted** — informational, no action requested |

---

### P2‑1 — reproduced, and the completeness claim is false

Reproduced independently before deciding. Godot 4.6.3:

```
RAW={"a": 1e-320, "a": 2.0}  || JSON={"a":2.0} || STV={"a":2.0}
RAW={"a": 2.0, "a": 1e-320}  || JSON={"a":0.0} || STV={"a":0.0}
```

Last-wins, so the destroyed literal is discarded in the first spelling and kept in the second.
At the rebased base `96bf1ca11`, real engine, `node set --property d`:

```
exit 0   {"property":"d","type":"Dictionary","value":{"a":2.0}}
.tscn:   d = {
         "a": 2.0
         }
```

At head: `uncoercible_value`, exit 4, note naming `1e-320`. So the base kept it faithfully and the
head refuses it — the comment's "both ways" and the catalog's "Three consequences … each a
deliberate disposition" are both falsified.

**Agreed that a code fix is disproportionate**, on the reasoning the scan already stands on.
Telling a discarded token from a kept one needs per-object key-and-position bookkeeping — a real
JSON parser — which is exactly the "second opinion about the engine's grammar" `_is_json_number_char`
exists to avoid; last-wins is Godot's own choice (RFC 8259 leaves duplicate keys undefined), so
modelling it would model engine internals. The cost lands in a byte-mirrored region, i.e. twice,
for a pathological input whose over-refusal writes nothing wrong.

Prose fix, option **(a)** — named as a third, accepted over-refusal, in the three places that own
this rule:

- the mirrored block comment (`operations.gd` + `gda_harness.gd`, re-mirrored **verbatim** by
  extracting the marker-delimited block and writing it over the harness copy);
- `docs/command-catalog.md` — "Three consequences" → **"Four consequences"**, with a fourth bullet;
- `src/gda/live_numbers.py` — not in your list, but its docstring carried the same sufficiency
  framing ("which is what keeps it from refusing a write the engine would have kept"), and CONTEXT
  names it the number-domain authority. Leaving it would have re-created the drift.

`CONTEXT.md` and `SKILL.md` were re-read and left alone: neither claims exhaustiveness (CONTEXT's
"covers every literal a `--value` spells" is about what is OBSERVED, which is still true — the
discarded literal is observed, that is the whole finding).

**Deliberately not pinned by a test.** Asserting the over-refusal persists would turn a future
improvement into a test failure. It is recorded as measured behaviour instead.

**Guards.** Comment-only, but the harness bytes change, so `PINNED_HARNESS_SHA256` was re-pinned
(`835dfd9d…` → `687f176b…`). `HARNESS_VERSION` stays **18**: the guard compares base to head, and
17 → 18 still holds against `96bf1ca11`. `harness_version_guard.py 96bf1ca11 ec7d39822` → valid.

### P3‑1 — both commands now pinned

Confirmed your hand-measurements on the real engine, then turned them into tests. A shared
four-row corpus (`CONTAINER_PER_COMMAND`) drives each: the two refusal classes (destroyed to `0.0`
in a Dictionary, to `NaN` in an Array), the string-value trap, and an exact value. `node set` keeps
the full corpus; what these add is that the container half travels with the SHARED coercion.

- `test_project_set_shares_the_refusal` — extended; `project.godot` now seeds
  `probe/dict={"a": 1.0}` / `probe/arr=[1.0]` (the setting's current type IS the declared type).
- `test_resource_set_shares_the_refusal` — new. The probe is a **hand-written script-backed
  `.tres`** naming the script as an `ExtResource`, so it loads without the class registry and needs
  no `--import` pass. Also asserts the file is byte-unchanged after a refusal.

### P3‑2 — reordered

The caveat + decision record moved to **after** the container rule and its four bullets. No anchors
broke (neither paragraph carries a heading), and the two prose back-references still resolve —
"recorded here rather than folded into the rule above" and "what the refusal rule above avoids".
`test_live_numbers.py`'s catalog assertions are substring checks, order-independent; green.

### P3‑3 — the trigger was not written down; it is now

Checked first: `grep` over `src/`, `docs/`, `CONTEXT.md` found no such trigger. The existing comment
explains why the container arm repeats the GATE, but not the escape hatch. Added to
`_destroyed_float_literal`'s comment, in your words: the scalar arms' second walk re-derives a
different shape and that independence is what keeps the note honest; the container arms re-derive
nothing, which is affordable at two arms — a THIRD container-shaped type should have `_coerce_value`
hand back the reason it refused instead. Zero marginal cost (the mirror was being re-pinned anyway).

### P3‑4 — not adopted

Measured and confirmed: 5 `Exponent too high` warnings for a container `0e600`, 3 for the scalar.
Pre-existing engine noise on a path this change did not create, and you ask for no action. Recorded
in the commit body so it is not rediscovered as new.

---

## Validation (head `ec7d39822`, Godot 4.6.3, `~/Applications/Godot.app`)

| Gate | Result |
| --- | --- |
| `pytest tests/ -q -m "not e2e" -p no:cacheprovider` | **2393 passed**, 646 deselected |
| `tests/test_harness_coercion_mirror.py` (parity guard) | **3 passed** |
| `tests/test_harness_install.py` (pinned bytes) | **41 passed** |
| `scripts/harness_version_guard.py 96bf1ca11 HEAD` | valid transition |
| `-m e2e tests/test_e2e_write_value_fidelity.py` | **9 passed** (was 8 — the new `resource set` band), incl. the live `game set` daemon band |
| `-m e2e test_e2e_headless_number_reads.py` + `test_e2e_live_number_transport.py` | **9 passed** |
| `-m e2e test_e2e_resource.py` + `test_e2e_project.py` (regression) | **57 passed** |
| `ruff check` / `ruff format` | clean |
| `pyright` | 0 errors, 0 warnings |
| `gda schema` vs base `96bf1ca11` | **byte-identical** (`27663f01e3e559b21c9202b6db4708789d3b9743b001b5f90ac47bfb49e527f9` both) |

**Not performed:** macOS only (no Linux/Windows); no fuzzing beyond the reproduced payloads.

qinhaihong-red:
# Independent adversarial review — PR #809

**Reviewed base:** `96bf1ca11f9ba0a1ab933565f2eb22aabe28bffb`  
**Reviewed head:** `ec7d3982297ea2d6e93d7f5636b85dd4679a28a1`

## Artifact review — Changes required (1 finding)

### ART-1 — Reconcile #805 with the implementation decision before merge

**Location:** issue #805, **What to build §1** and **Acceptance criteria**.

The issue still says that a Variant constructor such as `Vector2(1e-320, 0)` “inside an accepted dict is reachable.” The exact-head implementation and catalog establish the opposite: `JSON.parse_string` rejects that text before `str_to_var`, and the test pins a plain `uncoercible_value`. The issue's pre-dispatch footer replaces the former traversal machinery, but it does not withdraw this false reachability statement.

The same issue also requires: “Exact values still store exactly.” At `src/gda/ops/operations.gd:6672-6680`, however, the raw scan deliberately refuses `{"a": 1e-320, "a": 2.0}` even though Godot discards the first member and would store the exact surviving value `{"a": 2.0}`. `docs/command-catalog.md:498-504` records this as an accepted over-refusal, but #805 does not authorize that exception.

**Impact:** the closing issue remains internally inconsistent and gives a future implementer two incompatible accounts of both the constructor boundary and exact-value admission.

**Smallest fix:** update #805 before merge: state that the JSON gate makes constructors unreachable, and either explicitly narrow the exact-value criterion to accept the documented duplicate-key over-refusal or require the scanner to consider only surviving members.

**Terminology and prose assessment:** I inspected the complete current issue, all changed Markdown artifacts (`CONTEXT.md`, `docs/command-catalog.md`, and `src/gda/skill/SKILL.md`), and the changed technical prose/comments. Terminology sources were `CONTEXT.md`, `CONTEXT-MAP.md`, the command catalog, and accepted ADRs 0010/0017/0018/0020/0035. ASD-STE100 was the writing reference. I checked term stability, first-use meaning, causal/scope/order statements, procedures, duplicated authority, and changed descriptive prose. Apart from ART-1's substantive contradiction in the issue, I found no terminology or prose problem that prevents the changed Markdown from serving its purpose. Link automation and non-English rendering were not tested.

## Domain modular architecture — Pass (0 findings)

The change keeps the write-admission rule under the existing shared coercion owner. Headless `node set` / `resource set` / `project set` call the operation copy; live `game set` calls the byte-identical harness copy. ADR-0018 explicitly accepts this duplication because the two runtime contexts cannot import one module, and the parity guard plus harness version/pin preserve the one-way authority. `gda.live_numbers` remains a leaf measurement/documentation authority rather than a second executable policy.

- **Completeness:** all four coercing commands, both container types, attribution failures, untouched targets, and headless/live paths have owners and focused evidence.
- **Orthogonality / DRY:** the new scan extends the existing coercion responsibility; it does not add a parallel command policy. The cross-runtime copy is the accepted ADR-0018 seam and is mechanically guarded.
- **Dependencies / communication:** inbound command paths continue through their existing execution adapters; no Domain/Application dependency on a concrete outer adapter or new cycle appears.
- **Extensibility:** the code records the concrete trigger for replacing the repeated container diagnosis (a third container-shaped type), without introducing that abstraction early.

The repository defines no claim-evidence-state vocabulary, so the two load-bearing claims remain `open` under the review method: (A1) every current coercing command reaches this owner, disconfirmed by a setter-specific bypass; and (A2) the live/headless policies are identical, disconfirmed by mirror drift or an invalid harness transition. Exact-head command tests, the byte comparison, and the version guard support both at the reviewed scope. Linux/Windows behavior and unbounded fuzzing remain open validation needs.

## Standards — Pass (0 findings)

No documented-standard violation or actionable baseline smell was found. The exact diff was reviewed against `AGENTS.md`, `RULES.md`, `CLAUDE.md`, `CONTEXT.md`, and accepted ADRs 0000/0002/0009/0010/0017/0018/0020/0023/0035. The mirrored GDScript is not a **Duplicated Code** finding: ADR-0018 requires verbatim copies for irreconcilable runtime contexts, and the drift guard enforces that decision. No hard violation or judgement-call smell remains.

## Spec — Changes required (1 finding)

### SPEC-1 — The duplicate-key over-refusal exceeds the current acceptance decision

`src/gda/ops/operations.gd:6672-6680` scans discarded duplicate-key values, so `{"a": 1e-320, "a": 2.0}` is refused although Godot would faithfully store `{"a": 2.0}`. `docs/command-catalog.md:498-504` explicitly accepts this behavior. That contradicts #805's acceptance criterion: “Exact values still store exactly.”

**Impact:** a valid, lossless write now fails on all four commands.  
**Smallest fix:** make admission consider only surviving duplicate-key members, or explicitly revise #805's acceptance decision before merge. Documentation alone does not satisfy the current spec.

No other missing, partial, incorrect, or unasked behavior was found. The four commands, message family, untouched-target behavior, string/key traps, ordinary exact values, mirror parity, packer-elision decision, and required documentation surfaces are covered. The stale constructor sentence is an issue-artifact defect (ART-1), not an implementation defect: the chosen JSON-gate behavior is pinned.

## Validation

- Exact three-dot diff was non-empty; commits reviewed: `f33c640d2`, `ec7d39822`.
- Exact-head focused unit run: **266 passed** (`test_harness_coercion_mirror`, `test_harness_install`, `test_live_numbers`).
- Exact-head real-engine/daemon run: **9 passed** (`tests/test_e2e_write_value_fidelity.py`).
- Independent Standards run: mirror/install **44 passed**; live-number/write-fidelity **231 passed**; Ruff check/format, Pyright, and `git diff --check` clean.
- `gda schema` was byte-identical at base and head: `27663f01e3e559b21c9202b6db4708789d3b9743b001b5f90ac47bfb49e527f9`.
- Shared coercion blocks were byte-identical; `harness_version_guard.py` reported a valid transition.
- Current CI at this head: all required non-E2E checks succeeded; the GitHub **Godot e2e tests** job was skipped. Local real-engine evidence above covers the changed path.

Not performed: the full repository suite, Linux/Windows execution, fuzzing beyond the focused corpus, or rendered/HITL behavior. These gaps do not erase SPEC-1 or ART-1.

**Summary:** Artifact 1 finding (issue authority is stale); Domain modular architecture 0; Standards 0; Spec 1 (duplicate-key admission conflicts with #805).

qinhaihong-red:
## Round-2 response

Head unchanged: `ec7d39822` — both findings are issue-authority defects, resolved on issue #805.

| #805 still asserts the constructor edge is reachable, contradicting the delivered (and pinned) JSON-gate behavior (ART-1a) | **Fully adopt** | The unreachability is measured (`JSON.parse_string` err 43 on constructor text) and test-pinned | The reachability sentence struck in place with the refutation and a pointer to the pin; a dated round-2 note records it | Issue #805 diff |
| #805's "Exact values still store exactly" conflicts with the duplicate-key over-refusal the implementation and catalog accept (ART-1b / SPEC-1) | **Fully adopt** (on the issue; the code alternative declined) | Reproduced in the fix round: the engine keeps the surviving member; the head refuses the pathological spelling | The AC now names the one accepted over-refusal with its rationale — tracking survivors needs key-and-position bookkeeping, a second opinion about the engine's grammar duplicated into the byte-identical mirrored region, for an input whose refusal writes nothing wrong — and points at the reopen trigger recorded in `_destroyed_float_literal`'s comment. The alternative fix (admit surviving members) is declined for those recorded reasons; over-refusal is the safe direction | Issue #805 diff; no code change; catalog already records the fourth consequence |

DM and Standards stood at Pass.

qinhaihong-red:

Reviewed base/merge base `96bf1ca11f9ba0a1ab933565f2eb22aabe28bffb` and current head `ec7d3982297ea2d6e93d7f5636b85dd4679a28a1`. The code head is unchanged since the prior review; this re-review verifies the amended issue #805 against that exact implementation.

1. **Issue #805 claims that the duplicate-key decision's reopen trigger is recorded in `_destroyed_float_literal`, but it is not.** The trigger there concerns introducing a third container-shaped type and consolidating duplicate diagnostics; it does not state when the accepted duplicate-key over-refusal should be reconsidered. Smallest fix: add a concrete duplicate-key reopen condition to the correct mirrored duplicate-key comment, or remove the unsupported requirement and claim from #805.

2. **Issue #805 states the raw literal must be read “before parsing,” which reverses the accepted processing order.** The implementation, `CONTEXT.md`, command catalog, live-number contract, and tests require a successful `JSON.parse_string` gate before the raw text is scanned for numeric spellings. This ordering matters because constructor syntax and invalid JSON must be rejected by the JSON gate, not diagnosed as coercion loss. Smallest fix: state that the raw spelling is preserved as input, then scanned only after JSON parsing succeeds.

**Terminology and prose assessment:** I inspected the complete amended issue and changed Markdown, using `CONTEXT.md`, `CONTEXT-MAP.md`, the command catalog, and relevant ADRs as terminology sources and ASD-STE100 as the writing reference. Apart from the factual ordering and cross-reference problems above, no substantive terminology or prose issue was found. No automated ASD-STE100 checker was used.

The code structure is unchanged and remains sound: shared coercion has one logical owner, the byte-identical harness copy is required by ADR-0018 and guarded by a mirror test, dependencies remain directed, and all four command paths are covered. The missing reopen trigger is an issue/specification alignment problem, not a new module-boundary defect.

### Standards — Pass

No documented-standard violation or actionable Fowler smell was found. The mirrored GDScript bodies, harness version, and hash pin follow ADR-0018 and the repository's guarded-duplication rule.

### Spec — Changes required

The amended issue makes the duplicate-key over-refusal an accepted exception, so the prior behavioral conflict is superseded. However, its new requirement that a duplicate-key reopen trigger is recorded in code is not satisfied by this head. Add that trigger to the appropriate mirrored comment or remove the claim from the authoritative issue.

### Validation

- Focused unit tests: `266 passed`; focused mirror/live-number/write-fidelity review suite: `275 passed`.
- Real Godot/daemon write-fidelity E2E: `9 passed`; additional four-command probes: `2 passed`.
- All four setters reject destroyed literals and the accepted duplicate-key spelling, while accepting ordinary exact `{"a": 2.0}`. The base stores the duplicate-key example as `{"a": 2.0}`; this head refuses it as the amended issue now specifies.
- Ruff check/format, Pyright, and `git diff --check`: passed.
- Schema is byte-identical across base/head; the coercion mirror is byte-identical and the harness version transition is valid.
- Current CI: `failed=[]`, `pending=[]`; hosted Godot E2E is skipped.

Not independently checked: the full repository suite, Linux/Windows, unbounded fuzzing, rendered output, or HITL behavior.

qinhaihong-red:
## Round-3 response

Head unchanged: `ec7d39822` — both findings are issue-text defects, resolved on #805.

| #805 claims the duplicate-key reopen trigger is recorded in `_destroyed_float_literal`'s comment; the trigger there concerns the container-diagnosis consolidation, not this decision | **Fully adopt** (the remove-the-claim variant) | Verified: the in-code trigger is about a third container-shaped type | The misattribution is corrected on the issue: the catalog's fourth consequence is named as the recorded rationale/remedy, and the reopen condition is stated ON THE ISSUE (new evidence — a legitimate caller payload, not a pathological hand-spelling, refused for a duplicate key). The alternative (adding a mirrored duplicate-key comment) was declined: it would edit the byte-identical region and force a harness SHA re-pin for one sentence the issue can carry | Issue diff; no code change |
| #805 says the raw literal is read "before parsing", reversing the delivered gate-then-scan order | **Fully adopt** | Verified against the implementation and its tests | Reworded: the raw spelling is preserved as input and scanned only after `JSON.parse_string` accepts; the gate, not the scan, refuses invalid JSON and constructor syntax | Issue diff |

Issue #548, opened 2026-07-22
[gda-balancing] Deliver immutable Action plans, cost commitment, and interruption


qinhaihong-red:
## Parent

- #534
- #501

## What to build

Deliver `RPG-COST-01`, `RPG-ACTION-01`, and `RPG-ACTION-02`: admit one immutable Action plan,
reserve its cost, resolve or interrupt that exact identity, commit/refund exactly once, and schedule
or cancel the exact declared child events atomically. All child-Event admission, same-time ordering,
Snapshot visibility, scheduling, cancellation, and `step` behavior must reuse #594's normative
Runtime/Experiment path rather than an Action- or RPG-owned timeline.

## Acceptance criteria

- [ ] The selected plan has an immutable content identity and exact bound execution inputs from admission through completion/interruption.
- [ ] Cost reservation, one-time commit, insufficient-funds outcome, interruption refund, cooldown, and child-event scheduling/cancellation execute in the declared ownership boundaries.
- [ ] Plan mutation, duplicate completion, insufficient resource, and interruption-after-complete vectors produce canonical typed outcomes/refusals.
- [ ] Public trace exposes Action state, plan identity, reservation/commit/refund, canceled and child Event ids, and cooldown state.
- [ ] The current Event commits or rolls back atomically and row closure uses exact receipt-backed public artifacts.
- [ ] Action scheduling/cancellation consumes #594's stable Event identities and queue law; no host callback, private action loop, fixed tick, or new Runtime phase is introduced.

## Blocked by

- #547

Issue #547, opened 2026-07-22
[gda-balancing] Deliver dynamic targeting and named-random check coverage


qinhaihong-red:
## Parent

- #534
- #501

## What to build

Deliver `RPG-TARGET-01` and `RPG-CHECK-01` through one bounded combat decision path: resolve targets
from the current dynamic entity set under stable ordering/cardinality rules, then perform a named-
stream threshold or opposed check with explicit tie policy and success degree.

## Acceptance criteria

- [ ] The admitted `rpg.combat.cast-v1` slice resolves dynamic candidates, filtering, stable order, cardinality, tie, and empty behavior without host-container ordering.
- [ ] Empty-target outcome, missing empty policy, and missing check tie policy vectors produce the specified typed outcomes/refusals.
- [ ] The check records the exact RNG stream identity and draw index under the admitted RNG law.
- [ ] Public artifacts expose target ids/order, `target_count`, and `hit_degree` Metric samples.
- [ ] Receipt-backed claim closure verifies both coverage rows without treating random agreement or fixture labels as evidence.

## Blocked by

- #546

Issue #546, opened 2026-07-22
[gda-balancing] Deliver the RPG-STAT-01 source-to-validated-artifact slice


qinhaihong-red:
## Parents

- #534
- #501

## Problem

`RPG-STAT-01` does not yet have one admitted, executable source-to-artifact slice. The toolkit can
represent typed Symbols and Formula bindings, but it does not yet prove that progression, build,
and effect contributions compose through their owning packages into one derived combat Quantity.
It also lacks a player-facing product that lets an ordinary RPG player vary those contributions,
observe the cap, and report whether the result is understandable.

The slice must use the existing Model build, Experiment, Metrics, audit, validation, Replay, and
local Execution HTTP paths. It must not create a parallel RPG authority, game-specific service
route, generated playtest case, or Godot-side damage calculation.

## Product question

Can an ordinary RPG player understand how Level, Weapon Damage Bonus, and Damage Buff affect Attack
Damage, and recognize when Attack Damage reaches its maximum?

The playable product is **Attack Damage Training**. The player adjusts those three inputs and
attacks a 120-HP training dummy until it is defeated. Settings remain editable between attacks and
take effect on the next attack. Defeat, not an arbitrary trial count, ends the playthrough; Restart
and Give Feedback are separate terminal actions.

## Authority and package design

- Model Source owns the named Formula declarations and exact Formula bindings.
- `game.progression@1.0.0` owns Progression Contribution through
  `game.progression.contribution@1` and its `progression-policy` Formula slot. It depends on
  `core.quantity@2.2.0`.
- `game.build@2.0.0` owns Build Contribution through `game.build.contribution@1` and its
  `build-policy` Formula slot. It selects `core.quantity@2.2.0` and `game.generation@1.1.0`.
  This release does not retain the unrelated `game.build@1.0.0` public contract, so the new
  contribution API uses an honest major boundary. `game.build@1.0.0` remains available
  unchanged.
- `game.effect@2.0.0` owns Effect Contribution through `game.effect.contribute@1` and its
  `effect-policy` Formula slot. It selects `core.quantity@2.2.0`. This release does not retain
  the periodic-effect public contract of `game.effect@1.0.0`, so the new contribution API uses
  an honest major boundary. `game.effect@1.0.0` remains available unchanged.
- `core.quantity@2.2.0` adds only the typed pure Operations required by this slice:
  `quantity.add`, `quantity.multiply`, `quantity.minimum`, and `quantity.floor-divide`.
[...]

PR #856, opened 2026-09-05
docs(balancing): record the dynamic targeting and check design baseline (#547)

qinhaihong-red:
## Summary

- Record the #547 design baseline before any implementation, so that the design can be reviewed on its own.
- Add bADR-0028 (status `proposed`): four bounded collection primitives in the Kernel expression family, exposed through `standard.schema@2.5.0`.
- Amend bADR-0017 with the package ownership that consumes them, the check contract, the trace-level RNG record, the exact package closure of the `rpg.target-check-v1` Golden, and the candidate/open claim scope.
- Add four glossary terms, revise `Target query`, revise the `RPG-TARGET-01` and `RPG-CHECK-01` coverage rows, add the Golden contract and closure rules, and update ARCHITECTURE sections 5.2, 7.1, and 16.
- Docs only. No authority, code, vector, or example change.

## Decisions recorded

1. **Kernel.** The provisional Kernel has no node that can filter, order, truncate, or count an admitted `List`, so a Target query cannot be expressed as a package Operation body. bADR-0028 admits `where-equal`, `order-by`, `take`, and `count` as orthogonal expression nodes, each charging `1 + n` steps, with explicit tie order and equality-only filtering. This is the third provisional-baseline reopening after #640 and #546.
2. **Entity and query boundary.** `game.entity@1.0.0` owns identity and membership types. `game.query@1.0.0` owns a typed candidate view (`TargetCandidate`), the policy, and `select@1`. A candidate row copies faction and life state from its producer and binds identity only through `EntityRef`. The empty policy uses the `List<..., max=1>` shape that `game.generation` already uses.
3. **Check contract.** `game.check@1.2.0` adds `resolve@1`: threshold or opposed mode, d100, success degree equal to the exact integer margin, a tie policy that must be declared, refusal before any draw when it is missing, two named streams (`check-actor`, `check-target`), and a `failed` outcome that commits the negative degree.
4. **Decision path.** `game.combat@2.3.0` adds `CombatProfile` and `target-check@1`, which composes select and resolve and consumes the first resolved target. No damage and no resource spend in this slice.
5. **RNG record.** The per-draw rows in the Event trace (stream, index, candidate, bounds, value) are the `RPG-CHECK-01` observable. The check result does not repeat them.
6. **Golden placement.** `rpg.target-check-v1` lives in a new `examples/schema2/rpg-target-check/` example until #548 admits Action plans and folds the path into `rpg.combat.cast-v1`.
7. **Claim scope.** Verification publishes `candidate`/open. Both rows stay open; Verifier receipts and claim aggregation remain deferred by the milestone.
8. **Playtest.** The slice includes a player-facing Godot application, **Target Practice**. Product question: can an ordinary RPG player predict which enemy the hero targets from the visible targeting rule, and understand from the shown accuracy + roll versus defense margin why an attack hit or missed? The player chooses the ordering rule, cardinality, check mode, and tie rule, marks enemies as defeated to edit the dynamic set, and attacks. Each attack is one complete Experiment revision. The application applies no damage. This paragraph moves into the #547 issue body after approval.
9. **Delivery.** After approval: PR-1 Kernel nodes, `standard.schema@2.5.0`, both evaluators, vectors, and identity re-pinning; PR-2 the four package releases, LDB reseal, the Golden example, and docs; PR-3 the Target Practice playtest and its Godot tests. The three PRs are serial.

## Review focus

- Node set: is it minimal and orthogonal, and is the `1 + n` charge law acceptable?
- Candidate view: the duplication of faction and life state in `TargetCandidate` against bADR-0017's entity/query split.
- Check semantics: refusal before any draw, and `failed` with commit state policy.
- Coverage rows: capability ids without the `-v1` suffix, and the Golden switched from `rpg.combat.cast-v1` to `rpg.target-check-v1`.

## Not in this PR

- No implementation. bADR-0028 stays `proposed` until this review passes; PR-1 flips it to `accepted`.
- The #547 issue body is not edited yet. After approval it receives the product question, the package design, the exact closure, and the non-goals, following the #546 pattern.

## Validation

- Manual read-through of every inserted section for consistency with bADR-0016, bADR-0017, bADR-0022, and the current Kernel node inventory.
- `tools/ci.py classify` reports the changed paths as `required: true`, so the full balancing matrix runs on this PR.

Refs #547


qinhaihong-red:
**Changes required — Artifact Review (pass 1 of 3).**

Reviewed PR #856 at `da6e13bdae354dc7ad93797eaab358b7e8d98b3e`, against base `3f68bf3fb26df2ab54351a8ef4e3e167269bdc16` and merge-base `471e4145117c6746e61658044fa000f087b4e389`. Scope: the five complete current Markdown artifacts and their diff, PR description, issue #547, and relevant #534/#546 and package authority. This is a design review; missing implementation is not a finding. The stated post-approval update to #547 and the milestone's deferred receipt work are recorded as pending work, not unexplained omissions.

### Required changes

1. **[P2] Distinguish the proposed #547 replacement from the current #546 baseline.** [ARCHITECTURE.md, lines 1146–1151](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/ARCHITECTURE.md#L1146-L1151) says the target tracer “showed” the gap, connects its replacement to the “current baseline,” and points to §12.2 for these dogfooding results. But §5.2 calls #547 the *next planned* reopening, §11 and §13 still identify #546 as current, §12.2 contains no #547 product-run evidence, and bADR-0028 is `proposed`. The machine authority is unchanged. Readers cannot reliably determine which identity is current or whether #547 has executable evidence. State that the target requirement exposes a vocabulary gap and that implementation **will** replace #546; scope the §12.2 reference to completed probes. Present-tense decision wording alone is not the issue.

2. **[P2] Complete the dynamic-set capability rename in the coverage contract.** The changed [RPG-TARGET-01 row](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/standard-schema-2.0/genre-coverage.md#L43) and bADR-0017 line 66 reserve `game.entity.dynamic-sets`, while [RPG-ENCOUNTER-01, line 64](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/standard-schema-2.0/genre-coverage.md#L64) still requires `game.entity.dynamic-sets-v1`. These are exact capability identifiers, not interchangeable prose. A release following the new design cannot satisfy the latter reference without an additional, undesigned export. Update the dependent row to the same identifier.

3. **[P2] Use “gameplay outcome” for the check tie result.** [Check resolution, lines 883–887](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/BALANCING-CONTEXT.md#L883-L887) and [coverage line 194](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/standard-schema-2.0/genre-coverage.md#L194) call it a “verdict.” The same glossary, lines 1042–1047, reserves Verdict for a negative completed domain judgment with exit 1. A zero-margin check may instead produce `succeeded`; both check branches are Operation outcomes that commit their degree. Replace these two occurrences with “gameplay outcome” so a check branch does not inherit command/judgment semantics.

### Minor

- **[P3] Correct the theory attribution.** [bADR-0028, lines 98–102](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/badr/0028-kernel-bounded-collection-primitives.md#L98-L102) attributes selection, ordering and truncation together to relational algebra/Codd. Codd's relational model makes row order immaterial (§1.3, p. 379), so it does not establish this ordered-List contract. Attribute the ordered pipeline to SQL query operators and distinguish that analogy from Codd's unordered relations; the proposed four-node design need not change. [Codd's paper](https://ics.uci.edu/~jajones/INF102-S18/readings/25_codd.pdf).

Sources: package-local `BALANCING-CONTEXT.md`, `ARCHITECTURE.md`, bADRs 0016/0017/0022/0028, and the coverage matrix. Writing reference: ASD-STE100, as required by `RULES.md`. Inspected all changed prose, requirement/decision/procedure passages, and each major section of the complete artifacts. Checked one term/one meaning, technical nouns, subject/action clarity, condition and time relationships, causal claims, and abstract noun chains. Findings 1 and 3 are substantive language/meaning issues; no other substantive prose issue was found. This was not a word-by-word certified STE dictionary audit. Codd and Turner were checked from their papers; the exact Knuth §5 passage and full ISO SQL standard were not independently inspected.

Validation: `git diff --check` passed; the repository classifier returned `required: true`. The pinned-head CI checks completed successfully except the explicitly skipped root Godot E2E job. The existing `invoke` outcome mapping, per-draw Event-trace fields, and old-release semantic-identity separation were checked in current authority/source. No new-node implementation, conformance vectors, or Target Practice application exists in this diff, so this review supplies no runtime evidence for them.

qinhaihong-red:
**Changes required — Design Domain Modular Architecture (pass 2 of 3).**

Reviewed `da6e13bdae354dc7ad93797eaab358b7e8d98b3e` against merge-base `471e4145117c6746e61658044fa000f087b4e389`. Authority: package `AGENTS.md`/`docs/agents/domain.md`, `BALANCING-CONTEXT.md`, `ARCHITECTURE.md`, bADRs 0016/0017/0022, proposed bADR-0028, and #547/#534. The bADR-0017 lifecycle remains `accepted`; its new amendment describes the proposed slice. bADR-0028 is `proposed`. Neither lifecycle status establishes executable evidence: the relevant coverage and structural claims remain `open`.

### Observed structure and retained decisions

```text
Kernel: generic List operations, typing, resource law, atomic Runtime
LDB mechanic packages:
  game.combat → game.query → game.entity
              → game.check
              → game.resource
  mechanic packages → core.quantity / standard.schema / standard.runtime
  core.quantity → standard.compiler
```

Entity owns identity/membership contracts; query owns the candidate view and selection policy; check owns named-stream resolution and degree; combat owns composition and defense lookup. Experiment authors supply the initial candidate state; the later encounter producer will derive that view. Existing Runtime owns atomic execution and publishes Snapshot/trace observations; Experiment owns Metrics. Package dependencies are acyclic. The candidate view has an explicit producer and is not, by itself, a second entity authority.

Keep the four generic nodes and the query/check/combat boundaries. Each node serves an observed part of the path; no new context, service, callback, registry, or port abstraction is justified. Existing `invoke` supports exhaustive outcome mappings, refreshes writable ports, and propagates refusals; existing Event traces carry the required per-draw fields.

### Required changes

1. **[P2] Give combat ownership of the multiple-profile-match rule.** [bADR-0017, lines 96–101](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/badr/0017-genre-templates-and-coverage-contract.md#L96-L101) declares a `List<CombatProfile, max=8>`, finds the selected reference with `where-equal`, and uses “that defense,” but specifies only the zero-match refusal. `standard.schema.list` explicitly preserves duplicates, and the nominal Ref law does not make reference keys unique within a List. For E1 profiles with defense 10 and 90, an actor total of 50 yields either hit/+40 or miss/−40 depending on the unassigned choice. Require exactly one matching row, with a typed multiple-match refusal before check RNG, or specify an intentional conflict-selection policy. `where-equal → count → require` can enforce uniqueness without another Kernel node. Add a duplicate-reference/different-defense vector.

2. **[P2] Close the `take.count` value contract.** [bADR-0028, line 42 and lines 47–51](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/badr/0028-kernel-bounded-collection-primitives.md#L42-L51) allows a local reference or integer literal but leaves `count` out of the static operand rules. An integer-valued local with an incompatible kind/unit, or a Boolean local, has no specified admission result. The existing lookup-index checker explicitly matches the selected integer Literal Typing Profile; that lookup-specific law does not automatically apply to a new node. Bind both count forms to the admitted integer profile, explicitly exclude Boolean, and specify static mismatch refusals/vectors for incompatible kind, unit, and Numeric profile. This keeps type policy in Kernel/LDB authority rather than evaluator choice.

### Structural claim checks

| Repository claim | State and available evidence | Disconfirming condition / required validation |
| --- | --- | --- |
| bADR-0028 four-node sufficiency / `RPG-TARGET-01` | `open`; the proposed composition uses four generic nodes plus existing control/invoke | A hidden host query, iteration, or element-construction mechanism is required. Validate the Golden and both independent consumers after closing the two gaps. |
| bADR-0017 package ownership | `open`; roles and dependency edges are distinct and acyclic | Query redefines entity identity, or combat invents target eligibility. Check the eventual Package Release and Operation projections. |
| Core Extension Invariance | `open`; the proposed nodes contain no genre dispatch | A later witness needs Kernel/phase/dispatch changes. Revalidate Gate 5/6 against the replacement identity; do not transfer #546 evidence automatically. |

Completeness requires the two fixes above. No additional ownership overlap, dependency cycle, or unsupported extension mechanism was found. Duplicate candidate references were also examined: the stated List rules deterministically preserve them; without a requirement to count distinct entities, that is not a separate finding.

Validation: a focused probe against the pinned source's current structured-value admission accepted an existing nominal Record List containing the same Ref twice with different scalar values. This establishes the generic multiplicity premise only. The E1 profile calculation is a design counterexample, not an execution of the unimplemented `CombatProfile` or new nodes. New-node charge/limit behavior, new-package vectors, the Golden, and Target Practice still require implementation validation.

qinhaihong-red:
**Review (pass 3 of 3) — Standards and Spec, assessed independently in parallel.**

Pinned head: `da6e13bdae354dc7ad93797eaab358b7e8d98b3e`. Fixed-point diff: `git diff 3f68bf3fb26df2ab54351a8ef4e3e167269bdc16...HEAD`; merge-base `471e4145117c6746e61658044fa000f087b4e389`. Commit: `da6e13bda docs(balancing): record the dynamic targeting and check design baseline (#547)`. The scope is the proposed architecture in this documentation PR.

## Standards

**Documented-standard breach**

- **[P2] Keep forbidden execution mechanisms separate from deferred bounded primitives.** In [bADR-0028, lines 86–92](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/badr/0028-kernel-bounded-collection-primitives.md#L86-L92), “Each is a later minor addition when a consumer demonstrates the need before the freeze” also applies to “unbounded iteration, recursion, or a host callback.” This contradicts accepted [bADR-0022, lines 223–224](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/badr/0022-machine-readable-language-rules-and-formal-semantics.md#L223-L224): “Recursion, user-defined loops, unbounded collection traversal, reflection, dynamic operation lookup, host callbacks, ambient state, and same-name argument capture are not kernel features.” The local documented rule, [docs/agents/domain.md, lines 70–71](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/agents/domain.md#L70-L71), requires conflicts with existing bADRs to be surfaced explicitly rather than silently overridden. The provisional reopening mechanism permits revising a baseline; it does not make these exclusions ordinary deferred minor additions. **Minimal correction:** limit the future-addition sentence to the bounded operations in the first three bullets and explicitly retain the final bullet as an architectural exclusion. If changing that exclusion is intended, explicitly reopen bADR-0022 and its termination/effect guarantees.

**Heuristic smells**

None substantiated. The candidate view's duplicated fields have an explicit ownership/projection rationale, and the optional-policy List pattern follows an existing package convention. The repository standards override the Fowler smell baseline. Tool-enforced checks, absent future implementation/tests, and pending post-approval issue updates were not treated as defects.

## Spec

- **[P2] Define duplicate combat-profile behavior.** [bADR-0017, lines 96–101](https://github.com/aigengame/godot-agent/blob/da6e13bdae354dc7ad93797eaab358b7e8d98b3e/libs/gda-balancing/docs/badr/0017-genre-templates-and-coverage-contract.md#L96-L101). [PRD #534, user story 11](https://github.com/aigengame/godot-agent/issues/534) requires formal rules “so that multiple implementations accept and reject the same programs.” `CombatProfileSet` is a bounded `List`, so two rows may reference the same entity with different defenses. The proposed operation filters profiles with `where-equal`, but defines only the zero-match refusal; it specifies neither which matching defense to consume nor whether multiple matches are invalid. Consequently, independently implemented consumers could select different defenses or disagree on admission. Require exactly one matching profile, with an explicit typed duplicate-profile refusal and boundary vector, or specify a deliberate deterministic selection policy. Existing `count` and `require` primitives suffice.

No additional missing requirements or scope creep was identified within the explicit design phase. The separate Golden, delayed Action integration, and candidate/open verification are intentional phased delivery choices.

Validation: inspected the complete five-file diff, #547, parent PRD #534 and milestone updates, predecessor #546, applicable architecture contracts, and current RNG trace implementation. No implementation or runtime-test completeness claim was assessed.

Standards: **1 finding; worst P2, architectural exclusions weakened without explicit reconciliation.** Spec: **1 finding; worst P2, duplicate-profile behavior is undefined.**

qinhaihong-red:
**Review response — passes 1 to 3 handled at `9634e6a95`.**

Docs-only commit on the same branch. `git diff --check` is clean; CI on the new head is pending.

| Finding | Resolution |
| --- | --- |
| Pass 1 / 1 [P2] proposed #547 replacement vs current #546 baseline | fixed@9634e6a95 — `ARCHITECTURE.md` §7.1 now states that #546 is the current baseline, that bADR-0028 is `proposed`, that its implementation **will** replace that identity, and that §12.2 holds only the completed #640/#546 results; #547 has no product-run evidence yet. |
| Pass 1 / 2 [P2] dynamic-set capability rename | fixed@9634e6a95 — `RPG-ENCOUNTER-01` requires `game.entity.dynamic-sets`. |
| Pass 1 / 3 [P2] "verdict" for the tie result | fixed@9634e6a95 — both occurrences read "gameplay outcome" (glossary `Check resolution`; coverage closure paragraph). |
| Pass 1 / Minor [P3] theory attribution | fixed@9634e6a95 — bADR-0028 attributes the ordered pipeline to SQL query operators (`WHERE`, `ORDER BY`, `FETCH FIRST`/`LIMIT`) and states that Codd's relations are unordered (§1.3), so the analogy is to the SQL query pipeline, not to relational algebra. |
| Pass 2 / 1 and Pass 3 / Spec [P2] duplicate combat profile | fixed@9634e6a95 — bADR-0017 amendment: `target-check@1` filters the profile set with `where-equal`, requires the `count` of the filtered set to equal one with `require` before any named-stream draw, refuses `combat-profile-missing` for zero rows and the new `game.combat.reason.combat-profile-ambiguous` for more than one, and binds a vector with two rows for one entity and different `defense` values whose refusal leaves state unchanged. The coverage closure rules and the `RPG-CHECK-01` row (`rpg.combat.profile-ambiguous-refused-v1`) carry the same rule. |
| Pass 2 / 2 [P2] `take.count` value contract | fixed@9634e6a95 — bADR-0028: a literal `count` must select exactly one reachable integer Literal Typing Profile admitted by the selected Runtime profile; a local `count` must carry only value contracts that match such a profile (the `lookup` local-index law in `declared_lookup_contract`). A Kernel Boolean, a different kind/unit/representation/Numeric policy, and a literal with zero or several matching profiles are `static` `language.structured_value_type_mismatch` refusals; four static vectors are added to Validation. |
| Pass 3 / Standards [P2] bADR-0022 exclusions | fixed@9634e6a95 — Non-goals lists only the three bounded deferred operations; unbounded iteration, recursion, and host callbacks are restated as bADR-0022 architectural exclusions that a provisional reopening cannot admit without reopening bADR-0022 and its termination/effect guarantees. |

A disposable prototype ran in parallel with this review (validation-driven-design, `lightweight` mode). Its findings are in the next comment; three of them need a decision before this PR is approved, and the PR gets one more commit after that.

qinhaihong-red:
**Disposable prototype — dogfooding ledger for the #547 design (decision needed on D6, D7, D8).**

Mode `lightweight`, two charters: Q-A (do the four bADR-0028 nodes reach every layer as generic data: Kernel admission, LDB seal, lowering, both evaluators, trace/Snapshot/Metric, exact Replay, `evidence verify`?) and Q-B (is every runtime-data variation in the bADR-0017 amendment expressible in a straight-line Operation body?). Baseline `da6e13bda`; prototype branch `proto/547-target-check-probe`, code state `6cbce974a`, evidence record `research/547-probe/EVIDENCE.md` at `ebb2ab346`, in the local worktree `.claude/worktrees/agent-a375e54601b031d97` (not pushed, not for merge). The record is not conformance proof and not acceptance.

What was built: the four Kernel nodes with `runtime.node.*` vectors; `standard.schema@2.5.0` with four structured Operations and 12 structured-value vectors; `game.entity@1.0.0`, `game.query@1.0.0`, `game.check@1.2.0`; both evaluators plus the independent replay adapter and the independent bootstrap consumer; `examples/schema2/rpg-target-check/` driven through `model build`, `experiment check/run/replay`, and `evidence verify`. Nine execution vectors agree three ways (expected = production = independent); two mutations (tie direction, inverted predicate) were detected; exact Replay `matched`; `evidence verify` returns `candidate`.

### Per-decision verdict (the nine decisions in the PR body)

| # | Decision | Verdict |
| --- | --- | --- |
| 1 | Four Kernel collection nodes | `refined-adopted` — sufficient for the Golden path, but the Kernel node-contract vocabulary must widen in four places (D2–D5 below). |
| 2 | entity/query boundary | `confirmed-no-change` — cross-package nominal references and nominal `List` types admit. |
| 3 | Check contract | `gap-opened` — refuse-before-draw, two streams, tie policy, and the committing `failed` outcome hold; `CheckPolicy.mode` and a single `resolve@1` do not (D7). |
| 4 | Combat decision path | **open** — `target-check@1` was not built in the time box; `where-equal` over a `Ref` field is unexercised (follow-up probe running). |
| 5 | Trace-level RNG record | `confirmed-no-change` — every draw row carries stream, index, candidate, bounds, value; a refusal before the draw publishes no row. |
| 6 | Separate Golden example | `confirmed-no-change`. |
| 7 | `candidate`/open claims | `confirmed-no-change`. |
| 8 | Playtest | `no-design-effect` — outside the probe. |
| 9 | Versions / equality-only `where-equal` | `gap-opened` (versions, D6) / `confirmed-no-change` (equality-only). |

### Decisions needed before approval

- **D6 — `standard.schema@2.4.0` cannot stay available unchanged beside `@2.5.0`** (`gap-opened`, LDB keying). Three independent refusals reproduced: the typed-envelope Literal Typing Profile must have exactly one owning release (`structured_values.py:125`), the shared artifact wire schemas collide under `kernel.duplicate_identifier` once they differ, and the Formula notation authority must be unique. This contradicts bADR-0028 "Identity consequence" and the amendment's "all earlier releases remain available unchanged". **Recommendation:** treat `standard.schema` as a singleton LDB member: `@2.5.0` supersedes `@2.4.0` inside the sealed LDB, and the maintained examples are rebuilt (the identity churn is already accepted). Drop the "2.4.0 remains available" and "byte-identical payload" claims. Alternative: re-key LDB definitions by (package, version), which reopens bADR-0023 and is a larger change than this slice needs.
- **D7 — a single `resolve@1` with a runtime `mode` cannot make the second draw conditional** (`gap-opened`, package). `draw` and `invoke` are unconditional and `guard-block` completes the Event, so `threshold` mode would publish a spurious `check-target` row, which is exactly the `RPG-CHECK-01` observable. **Recommendation:** `game.check@1.2.0` exports `resolve-threshold@1` and `resolve-opposed@1`, `CheckPolicy{tie_policy}` drops `mode`; `game.combat.target-check@1` invokes `resolve-threshold@1` against `defense`; the Golden covers the opposed case through a direct check entrypoint. Prototype evidence: both operations run with both tie policies; the tie-policy refusal publishes no draw.
- **D8 — with the `1 + n` charge the refusal path ends in `internal_error`** (`gap-opened`, Kernel charge law). The terminal-audit walker recomputes the charge statically from `resource_charge.amount` and has no runtime element counts, so the typed refusal cannot publish (`event_steps` 25 vs 20); with a flat charge the same Experiment publishes `game.query.empty_policy_missing` with rollback. The Kernel node contract also fixes every charge at `{counter: event-steps, amount: 1}` (`runtime_validation.py:1018`), so `1 + n` needs a per-element term (D1). **Recommendation:** charge a flat 1 per collection node and state that the `List` type's static `maximum_length` bounds the work; this removes D1 and D8 and needs no new charge mechanism. Alternative: static `1 + maximum_length` at both sites, which requires the charge walker to resolve operand types.

### Refinements to fold into this PR after those decisions (`refined-adopted`)

- **D2** `count` needs one result-typing kind that resolves through the integer Literal Typing Profiles an integer `constant` selects (the Kernel count law owns no result type).
- **D3** `order-by` takes a Kernel-Boolean `descending` operand instead of a literal `direction` member: bADR-0017 puts the direction in runtime data and the Kernel cannot read a package Enum; the package converts `SortDirection` with `constant` + `equal`. Confirmed from Experiment data in both directions.
- **D4** the two static rules (`key`/`tie_key` is an element field; `operand` contract equals the field) are two named operand-constraint kinds, not implied prose.
- **D5** the `structured-value` conformance-vector kind gains four actions and one optional `operands` member, and every shipped structured-value vector is rewritten.
- **D9** a `guard-block` body requires every visible local to have exactly one value contract, so the `target_count` write must precede the guard in `select@1`; amendment prose only.

Confirmed without change: `take` count from a Record field (B4), effect nodes inside a guard body (B3), refuse before any draw (B5), opposed tie under both policies with `hit_degree` committed (B6), `on_empty` and cardinality refusals with rollback (A5–A7), stable tie order by input permutation (A3).

Non-claims: no conformance, no performance, nothing about the playtest, nothing about the maintained examples that select 2.4.0 (not rebuilt), and the host changes are the smallest that let the checks run, not designed interfaces. Ruff/Pyright were not run in the prototype environment; 11 positional test-data drifts in `test_schema2_bootstrap_language.py` were left red and are explained in the record.

qinhaihong-red:
**Prototype follow-up — decision 4 (combat decision path) is now exercised.**

Same disposable branch; code `112fb6127`, evidence `19f25170a` (`research/547-probe/EVIDENCE.md`, section "B7 (follow-up)", ledger rows D13/D14). Not pushed, not for merge.

`game.combat.target-check@1` was built as the reviewed rule states: invoke `select@1`, read the first candidate's `entity`, `where-equal` over `CombatProfileSet` on that `Ref` field, `count` equal to one checked with `require` before any draw, then invoke `resolve-threshold@1` against that row's `defense`. All five chartered cases run from Experiment data only:

| Case | Observed |
| --- | --- |
| hit / miss | one `check-actor` draw, `target_count` 1, `hit_degree` 50 / -70 committed |
| no-target | no draw row, `target_count` 0, `hit_degree` untouched |
| profile-missing | `game.combat.combat_profile_missing`, no draw, state unchanged |
| profile-ambiguous (defense 10 and 90 for one entity) | `game.combat.combat_profile_ambiguous` before the check invoke, no draw, state unchanged |
| query refusal propagated | `game.query.empty_policy_missing` with `call_path combat.target-check/select-target` |

Production and the independent evaluator agree on the whole bundle (`checked=56 failures=0`, including every pre-existing Formula-bearing package). Dropping the uniqueness `require` is detected: the ambiguous case becomes a silent `hit` against `defense` 10, an order-dependent result. `where-equal` over a `Ref` field needed no Kernel or host change beyond D1–D5.

Design effect for decision 4: `confirmed-no-change` for the path and the uniqueness rule; `gap-opened` for one host prerequisite and for naming the invoked resolve Operation (follows D7).

- **D14 (host, Model lowering; not a design decision):** `model build` refused the Operation with `nested call local operand is incompatible` (`_lowering.py:3769`) because the lowering records only invoke-result locals as `local` invoke arguments, so `defense` from a `lookup` could not be passed. The Kernel and package admission accept it; only lowering refuses. The implementation must mirror the full result-typing dispatch there. Belongs to PR-1 scope.
- **D6 now has five reproduced symptoms** (two more in the Formula notation authority and source schema version dedup). The recommendation is unchanged.
- Not tried: whether `game.combat@2.3.0` can keep the `2.2.0` Operations while depending on `game.check@1.2.0` (two `game.check` releases in one dependency list, subject to the D6 byte-identity rule).
- An Operation `result` sourced from a `lookup`-produced Quantity is a typed envelope that the scalar-result vector contract rejects; the probe sourced the result from the check invoke instead. Implementation note, not a design change.
- The three combat refusals ran with the flat-charge switch, because D8 still blocks the terminal audit under `1 + n`.

Issue #855, opened 2026-09-05
game find: locate runtime nodes by unique name, type, or group


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-051** (2026-08-20, gda 0.9.0). After a CombatScreen rebuild the enemy actor slot moved from `Enemy0` to `Enemy1`; the stale exact path correctly returned `live_node_not_found`, but a fresh full-tree read was the only way to find the actor again. Numeric fallback names are instantiation history, not identity; the entry asks for a selector by unique name, type, group, or script, with a typed ambiguity result.

Verified against gda 0.13.0 (`266c8009f`): every live op addresses a node by an absolute runtime path resolved with `get_node_or_null`; there is no lookup by type, group, script, or `%UniqueName`. The tree's `type` is the ENGINE class (`get_class()`), so a project `class_name` is not visible to a type match.

## What to build

A read-only `game find` that returns runtime paths for the nodes matching one or more selectors, so a caller resolves identity once and keeps exact paths for the follow-up ops:

- Selectors, combinable as AND:
  - `--type <Class>`: engine classes only, subclass-inclusive (`Node.is_class()`), so `--type Button` also returns a `CheckBox` and `--type Control` returns every Control. Help states both facts.
  - `--script <res://path>`: nodes whose attached script, or a script in its base chain, is that resource — the selector that covers a project `class_name`, which `--type` cannot see. This is the actor-slot need GDA-DF-051 recorded.
  - `--group <name>`: `Node.is_in_group()`.
  - `--name <Name>`: exact node name.
  - `--unique-name <Name>`: a node with `unique_name_in_owner` set whose `name` is `<Name>` and whose `owner` is the search root or lies inside the searched subtree (wording fixed 2026-09-05). Unique names are per owner and a running tree holds many owners (autoloads, every instanced sub-scene), so the match is decided against the node's own `owner`, not against `current_scene`: the same `%Name` declared under an owner OUTSIDE the searched subtree is not a match.
- `--root` and `--max-depth` bound the search as `game tree` does (#849).
- The result is the list of matches (path, name, type, script path when any) plus `count`; zero matches is a success with an empty list. Ambiguity is data, not an error: the caller sees every candidate, and the ops that need ONE node keep taking an exact path.
[...]

PR #858, opened 2026-09-05
feat(gda): bound `game tree` with --root and --max-depth (#849)

### LINKED ISSUE #849: game tree: --root and --max-depth with omitted-node counts
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-052** (2026-08-21, gda 0.9.0). An unfiltered `game tree --json` on a gallery owning a production CombatScreen and two dynamic GameView graphs exceeded 36,000 tokens and was truncated by the client; the useful paths sat at the start and the end, the middle consumed the budget, and a second read would truncate again as the UI grows. The workaround was one lucky full read for path discovery and exact `game get` / `game rect` afterwards; a missing node cannot be inferred from a truncated tree.

Verified against gda 0.13.0 (`266c8009f`): `game tree` takes no options beyond the shared ones and serializes the whole subtree of the running **current scene** — `/root` only when no scene is current. (Corrected 2026-09-06: the first wording said "the whole tree from `/root`"; an autoload is the current scene's SIBLING under `/root`, so the default read never showed one. See the dated comment.)

## What to build

- `--root <runtime path>`: serialize the subtree at that node (the same absolute paths `game tree` reports); an unknown path is the existing `live_node_not_found`.
- `--max-depth N` (0 = the root only): nodes deeper than N are not serialized. Two counters, defined exactly: `children_omitted` on a node is the number of its DIRECT children the read did not serialize; `omitted_nodes` on the result is the total number of nodes (all depths) the read did not serialize. A partial read is therefore never mistaken for a complete one.
- The result reports `truncated: bool` (true when `omitted_nodes > 0`).
- Declined here: GDA-DF-052's continuation token. A live tree changes between calls, so a token would page a snapshot that no longer exists; `truncated` plus `omitted_nodes` carry the guarantee the token was for, and the follow-up read is a narrower `--root`.
- Help states the two options and that an unbounded read of a large tree is the caller's choice; the skill's live-inspection guidance shows the two-step pattern (bounded read, then exact addressing). Both say where the default read is rooted — the running current scene — and that an autoload, its sibling under `/root`, takes an explicit `--root /root`; the counters cover the selected subtree only (added 2026-09-06).

## Acceptance criteria

- [ ] On a fixture whose `HUD` node has 3 direct children, one of which has 2 children of its own: `game tree --root /root/Main/HUD --max-depth 1 --json` returns the HUD node and its 3 children, `children_omitted: 2` on that one child, `omitted_nodes: 2`, and `truncated: true`.
- [ ] `--max-depth 0` on the same root returns the root alone with `children_omitted: 3` and `omitted_nodes: 5`.
- [ ] Without the options the result is unchanged except for `truncated: false` and `omitted_nodes: 0` (additive).
- [ ] The schema publishes the new params and fields; a live regression covers the three cases above.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Summary

`gda game tree` serialized the whole runtime subtree of the running current scene
(`/root` only when no scene is current) and took no options. On a
production UI that result exceeded the caller's own budget and was truncated by the
CLIENT (GDA-DF-052) — a truncated dump reads exactly like a complete one, so a missing
node could not be inferred from it.

This slice bounds the read and reports what a bound left out:

- `--root <runtime path>` reads one subtree, addressed by the absolute runtime path
  `game tree` itself reports. A path that resolves to nothing is the existing
  `live_node_not_found` (an empty `--root` included — the same refusal `game get` gives
  for an empty node path).
- `--max-depth N` bounds the walk; `0` is the addressed node alone. A negative value is
  a usage error decided by the params model, before any daemon is involved.
- The result always carries `truncated` and `omitted_nodes` (every unserialized node, at
  every depth — whole omitted subtrees, not just direct children). A node whose children
  were not walked carries `children_omitted`, its DIRECT children only.
- No continuation token (declined in the issue): a live tree changes between calls, so a
  token would page a snapshot that no longer exists. The follow-up read is a narrower
  `--root`.

The harness's tree serializer becomes the one depth-bounded, counting walk,
`_serialize(node: Node, max_depth: int, counts: Dictionary) -> Dictionary`
(`max_depth < 0` = unbounded; `counts["omitted_nodes"]` accumulates by reference), with
`_subtree_size(node) -> int` for the count-only pass over an omitted subtree. It renders
a NESTED subtree, so an op that reports a FLAT match list (#855's `game find`) is a
different walk and does not call it — what such an op inherits from here is the per-node
shape (`name`/`type`/`path`) and this depth-bounding + omitted-count contract, not the
function.

Closes #849

## Surface diff

**`gda schema`** — 104 diff lines, all under `game tree` (base captured before the first
edit, head after the last):

- `argv`: two new option entries — `--root` (`input_property: root`) and `--max-depth`
  (`input_property: max_depth`), both `required: false`.
- `description`: one new paragraph on the command doc (the same six lines the help adds).
- `input.description`: rewritten (the params model is no longer empty).
- `input.properties`: new `root` (`anyOf` string/null, default null) and `max_depth`
  (`anyOf` integer `minimum: 0`/null, default null).
- `output.$defs.GameNode.properties`: new `children_omitted` (integer, `default: 0`) —
  NOT added to `GameNode.required`.
- `output.description`: rewritten.
- `output.properties`: new `truncated` (boolean) and `omitted_nodes` (integer).
- `output.required`: `["root"]` -> `["root", "truncated", "omitted_nodes"]`.

**`gda game tree --help`** — the docstring paragraph above, plus the two option rows:
`--root TEXT` and `--max-depth INTEGER RANGE [x>=0]`, whose help ends "must be >= 0". The
pre-existing rows are unchanged in text; Rich re-columns them because the metavar column
widened. The published schema is byte-identical before and after adding Click's `min=0`
(the input model already carried `minimum: 0`), so the enumerated schema diff above is
the whole schema change.

**`docs/command-catalog.md`** — the `game` bullet's `game tree` sentence now states the
two options, the three counters with their exact definitions, the absent-when-zero rule,
and the declined continuation token with its reason.

**`src/gda/skill/SKILL.md`** — the `game` row points at a new `### Find a node before you
address it` subsection (a peer of the existing `### Structured logging from game code`,
placed so it re-parents no existing prose). It gives the two-step pattern: bounded read
first, then exact addressing, plus how to read the three counters and why there is no
resumed page.

**`README.md`** — none. The `game tree` row describes the command, and per the README
scope rule option detail lives in `--help` / SKILL / the catalog; the shipped
`game get --texture-digest` is likewise absent from its row. No behavioural boundary is
dropped: the unbounded default is unchanged. Consequently no `README.*.md` regeneration
was needed and `scripts/update_readme_i18n.py` was not run.

## Harness version

- `HARNESS_VERSION` 19 -> 20 (`src/gda/harness/install.py`). Merge base re-read at commit
  time: `git merge-base origin/feat/gda-dogfooding-dev HEAD` = `3f68bf3fb` declaring
  `"19"`, so head is base + 1.
- `PINNED_HARNESS_VERSION` / `PINNED_HARNESS_SHA256` re-pinned in
  `tests/harness/test_harness_install.py` to
  `cb5dd790d2d4c4fd040775d77cf40c23403692f544090b343d39c1d77d7dbf8e`. (The review-fix
  commit reworded a harness comment, so the pin was re-taken a second time;
  `HARNESS_VERSION` stays 20 — the guard requires an increase across the MERGE BASE,
  which still declares 19, not one per commit.)
- Guard, run against the committed head (a pre-commit run passes vacuously, since it
  compares committed blobs):

```
$ uv run --frozen python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD
harness version transition is valid
```

## Red-proof evidence

Tests written first and run with the source at the committed base (`3f68bf3fb`, only test
files modified):

```
FAILED tests/live/test_game_commands.py::test_game_tree_emits_runtime_tree_json_through_the_live_channel
FAILED tests/live/test_game_commands.py::test_game_tree_relays_the_bounding_options
FAILED tests/live/test_game_commands.py::test_game_tree_reports_the_omitted_counts_of_a_bounded_read
FAILED tests/live/test_game_commands.py::test_game_tree_renders_the_omission_for_a_human
FAILED tests/live/test_game_commands.py::test_game_tree_refuses_a_negative_max_depth
FAILED tests/live/test_game_commands.py::test_game_tree_help_states_the_bounding_options
FAILED tests/live/test_game_commands.py::test_game_tree_schema_publishes_the_bounding_params_and_counts
7 failed, 128 passed, 10 deselected in 3.70s
```

Sample failures: `assert 'max_depth' in ''` (no such option, so the usage error names
nothing), `assert ('--root' in "Usage: gda game tree [OPTIONS] ...")`, and
`KeyError-shaped` `'max_depth'` on the published input properties.

One caveat, disclosed: `tests/meta/test_schema_command.py`'s sample-validates-schema test
did NOT go red for the added `GAME_TREE_TRUNCATED_RESULT`, because the published output
schema does not set `additionalProperties: false`, so extra keys validate under the old
schema too. That fixture is a forward drift guard, not the AC4 gate; the real gate is
`test_game_tree_schema_publishes_the_bounding_params_and_counts`, which did go red.

The e2e regression could not be run red in a useful way: the harness at base has no
`children_omitted` at all, so it fails on the first assertion. It is reported green below.

Review-fix round (`40d675a0b`): adding Click's `min=0` moves the refusal from the params
model to Click, which names the FLAG rather than the field, so the old pin went red first:

```
>       assert "max_depth" in usage_error_text(result)
E       assert 'max_depth' in "Usage: gda game tree [OPTIONS] Try 'gda game tree --help' for help. Error Invalid value for '--max-depth': -1 is not in the range x>=0."
1 failed in 0.31s
```

The assertion was then flipped to the flag spelling, which is the message an argv caller
actually reads. The `--params-json` path still refuses through the model's `ge=0`
(`invalid_params: max_depth: Input should be greater than or equal to 0`).

## Validation

Host: macOS 15 (Darwin 25.6.0), arm64, Godot `4.6.3.stable.official.7d41c59c4` at the
machine-local `~/Applications/Godot.app/Contents/MacOS/Godot` (passed as `$GDA_GODOT` per
invocation, never hardcoded). No launch timeout occurred in either round, so no re-run was
needed; load average was 9.20 at the first round (sibling wave slices) and 3.17 at the
review-fix round. Every row below is from the review-fix head `40d675a0b`, except the two
"extra" e2e rows, which were run at `539ffff27` and are unaffected by that commit (it
changes one harness COMMENT, the pin, an import, a Click bound and one SKILL sentence).

| Gate | Command | Result |
| --- | --- | --- |
| Unit | `uv run --frozen pytest tests -m "not e2e" -q -p no:cacheprovider` | `2460 passed, 651 deselected in 45.23s` |
| Format | `uv run --frozen ruff format --check .` | `508 files already formatted` |
| Harness guard | `uv run --frozen python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| e2e (game) | `pytest tests/live/test_e2e_game.py -m e2e -rs -p no:cacheprovider -q` | `10 passed in 31.29s` |
| e2e (harness) | `pytest tests/harness -m e2e -rs -p no:cacheprovider -q` | `4 passed, 51 deselected in 11.53s` |
| e2e (other live) | `pytest tests/live/test_e2e_diag.py tests/live/test_e2e_perf.py tests/live/test_e2e_logger.py -m e2e -rs -q` | `10 passed in 21.01s` |
| e2e (input) | `pytest tests/live/test_e2e_input.py -m e2e -rs -q` | `16 passed in 53.50s` |

The last two rows are extra: a GDScript edit that failed to parse would make EVERY live op
fail, and the auto-merge hazard for duplicate GDScript functions is only caught live, so
the other harness-served groups were exercised too. PR CI at `539ffff27` was green on all
seven required jobs (`Godot e2e tests` is `skipping` by repository policy — that tier runs
nightly on `main`, which is why the live evidence above is local). `tests/live/test_e2e_screen.py` was
not run — it needs a windowed session and this run had no attached desktop session; the
slice touches nothing on that path (the screen ops do not call the tree serializer).

Baseline unit count is unchanged at 2460 relative to the pre-edit `main` snapshot plus
this slice's 6 new unit tests and 1 new e2e test.

## Requirement conformance

**AC1** — `game tree --root /root/Main/HUD --max-depth 1 --json` returns the HUD node and
its 3 children, `children_omitted: 2` on the one child that has 2 of its own,
`omitted_nodes: 2`, `truncated: true`. Covered live in
`tests/live/test_e2e_game.py::test_game_tree_bounds_the_read_and_counts_what_it_left_out`
against a real engine session, on a fixture whose HUD has `Panel` (2 children), `Score`,
`Timer`; and at unit speed in
`test_game_tree_reports_the_omitted_counts_of_a_bounded_read`.

**AC2** — `--max-depth 0` on the same root returns the root alone with
`children_omitted: 3` and `omitted_nodes: 5`. Same live test. The 5 is the whole omitted
subtree at every depth (3 children + 2 grandchildren), which is why the harness keeps a
count-only `_subtree_size` pass over what it did not serialize.

**AC3** — without the options the result is unchanged except for `truncated: false` and
`omitted_nodes: 0`. Same live test asserts the full current-scene read still nests
`Main -> HUD -> Panel -> {Title, Value}` and that the literal string `children_omitted`
appears nowhere in the payload; the unit test
`test_game_tree_emits_runtime_tree_json_through_the_live_channel` asserts the same two
totals and the key's absence on both the root and a child.

**AC4** — the schema publishes the new params and fields (enumerated under Surface diff),
and a live regression covers the three cases. Gated by
`test_game_tree_schema_publishes_the_bounding_params_and_counts`, which also asserts
`children_omitted` is NOT in `GameNode.required`.

**`children_omitted` presence decision: present only when non-zero; the key is ABSENT
when a node's children were all serialized.** Reasons:

1. AC3 says the unbounded result is unchanged "except for `truncated: false` and
   `omitted_nodes: 0`". A per-node `children_omitted: 0` would change every node, so an
   always-present field contradicts the stated acceptance.
2. The issue exists because the unbounded read is too big. Adding ~22 bytes per node to
   it would push the exact result GDA-DF-052 could not fit further over the budget.
3. It is what the harness naturally produces: the key is written only where the walk
   stopped.

The cost is one branch for a raw-JSON consumer, so the description tells them the idiom
(`node.get('children_omitted', 0)`). Model-side there is no branch at all: `GameNode`
keeps `children_omitted: int = 0` and a `@model_serializer(mode="wrap")` drops the key
when it is zero, so the presence rule is a serialization rule owned by the writer rather
than a `None` every consumer must handle. The schema, the two `tests/support.py` fixtures
(`GAME_TREE_RESULT` unbounded, `GAME_TREE_TRUNCATED_RESULT` bounded) and the renderer all
follow that one decision.

## Disclosed extras / boundary crossings

- **Dispatch arm touched.** The brief confined harness edits to `_handle_game_tree` and
  `_serialize`. One further line was unavoidable: `_run`'s `OP_GAME_TREE` arm now reads
  `_handle_game_tree(params)` — the handler previously took none. No new `OP_*` const, no
  other harness section, and `_serialize` stayed physically where it was.
- **`_subtree_size` added** next to `_serialize` — the count-only pass AC2's
  `omitted_nodes: 5` needs. It is part of the replaced serializer, not a new section.
- **Human renderer.** `render_game_tree` appends one line, `truncated: <N> nodes
  omitted`, when the read was bounded, so the outline cannot be read as a complete tree on
  the human channel either. `gda/render.py` is deliberately NOT touched: `render_node_tree`
  is shared with the on-disk `scene`/`node` trees and `NodeOutline` has no per-node count,
  so the marker lives in the group that owns the shape (ADR-0040 §5). Consequence: the
  human channel reports the TOTAL, not which node hid what — the JSON channel has that.
- **`tests/support.py`** was edited additively, inside the game-tree fixture region only:
  the two totals were added to `GAME_TREE_RESULT` and `GAME_TREE_TRUNCATED_RESULT` was
  added beside it.
- **`tests/meta/test_schema_command.py`** gained one `jsonschema.validate` call for the
  new fixture (additive; see the caveat under Red-proof).
- **Live contract guards** needed no change: both new result fields are `int`/`bool`, and
  `_carries_float` admits neither, so no float-writer sentence is owed. No new op name, so
  the op-table mirror is untouched.
- **Wire numbers.** `max_depth` rides the relayed request as a JSON number, so it is
  covered by `RelayedLiveParams`: `ge=0` rejects negatives only, and the wire ceiling is
  the inherited `RelayedLiveParams` guard (`9007199254740991` accepted, `9007199254740992`
  refused as `max_depth integer values must be within +/-9007199254740991`). No extra
  depth cap is imposed; the harness reads the value through the existing defensive
  `_int_param` (a null or non-numeric value falls back to the unbounded sentinel).
- **Nothing outside the slice was edited**: `commands/input.py`, `commands/project.py`,
  `commands/script.py`, `commands/scene.py`, `hints.py`, `ops/operations.gd` and
  `gda.project` are untouched, and nothing in them was found to need changing.

## Review round (`40d675a0b`)

The independent review returned READY with five P3 findings; all five are adopted in the
second commit:

1. `SerializerFunctionWrapHandler` now comes from `pydantic`, the public API the two
   existing wrap-serializer sites (`provenance.py`, `commands/scene.py`) use, instead of
   `pydantic_core.core_schema`.
2. Click's `min=0` now mirrors the model's `ge=0` on `--max-depth`, per the convention
   `diag_limit_option` states and `input`/`perf`/`screen`/`resource` follow. The argv
   refusal now names the flag; the help states the bound; the schema is unchanged.
3. The harness serializer's forward reference to a later node-search op was inaccurate —
   `_serialize` returns a NESTED subtree, while a `game find` reports a FLAT match list.
   Reworded to what is actually shared: the per-node shape and the depth-bounding +
   omitted-count contract, not the function.
4. SKILL: "Every live op takes an exact runtime path" was false for the live ops that
   address no node (`diag errors`, `logger tail`, `perf monitors`, `screen capture`, the
   `daemon` group). It now says every live op that ADDRESSES a node.
5. Two unused pytest fixtures dropped from the help test.

## Review round 2 (`092fdec0b`)

The second independent review returned one P2 and two P3:

1. **P2 — `_subtree_size` overflowed the GDScript call stack.** The count-only pass
   recursed; on a 2200-node chain the walk overflowed (1024 frames by default),
   returned a partial total, and the op published it as a SUCCESSFUL read
   (`omitted_nodes: 2044` under `--max-depth 0`, with `Stack overflow in _subtree_size`
   in `diag errors`). Fixed: the walk keeps its own LIFO (`Array[Node]`), so a count
   costs no call depth. Regression
   `tests/live/test_e2e_game.py::test_game_tree_counts_an_omitted_subtree_deeper_than_the_call_stack`
   builds the chain from the main scene's script and asserts `children_omitted: 1`,
   `omitted_nodes: 2200` from the current scene AND from `--root /root --max-depth 1`,
   and an empty `diag errors`. Red on the previous harness (`assert 2044 == 2200`,
   the reviewer's number), green on this one. The nested serializer `_serialize` is
   unchanged, as the review asked: its depth is the caller's `--max-depth`, and an
   UNBOUNDED read of a tree deeper than the call stack is the pre-existing behaviour of
   the base serializer, outside #849.
2. **P3 — the default root.** The issue baseline and the params docstring said the
   default read covers the whole tree from `/root`; the harness selects
   `current_scene` and falls back to `/root` only when none is current, so an autoload
   (the scene's sibling) was never in it. Fixed in the params docstring, the `--root`
   help/schema description, the command docstring, the catalog and the skill's
   two-step pattern — all now say where the default is rooted, that the counters cover
   the SELECTED subtree only, and that discovering an autoload takes `--root /root`.
   #849's body carries a dated correction and comment (2026-09-06). The default itself
   is unchanged (AC3).
3. **P3 — "Wire numbers".** Corrected above: `ge=0` rejects negatives only; the wire
   ceiling is `RelayedLiveParams`'s.

`HARNESS_VERSION` stays 20 (the merge base still declares 19); the pin is re-taken
(`1074e1ecb10913b0e020e4ac99991631840cbd8966dce4c8f928c4e994b0ea18`), and the guard
against `origin/feat/gda-dogfooding-dev` passes. Re-run on this head: unit 369 passed over
game/harness/meta/live-guard modules, ruff/format/pyright clean, e2e game + harness 15
passed (Godot 4.6.3, macOS).

### DIFF
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -1322,16 +1322,18 @@ re-derives every verdict from a running engine.
   [`node`](#node).
 
 - **`game` (the running game's scene graph):** `game tree` reads the runtime scene
-  tree (shipped — the Phase-2 bootstrap tracer, #7), bounded on request by `--root
-  <runtime path>` (the subtree to serialize; an unknown path is `live_node_not_found`)
-  and `--max-depth N` (0 = the addressed node alone) — shipped, #849, from GDA-DF-052,
-  where an unfiltered read of a production UI exceeded the client's budget and was
-  truncated by IT, which a caller cannot tell from a complete read. What a bound leaves
-  out is counted rather than dropped: `omitted_nodes` totals the unserialized nodes at
-  every depth, `truncated` is that total above zero, and a node whose children were not
-  walked carries `children_omitted` — its DIRECT children only, and the key is absent
-  when nothing was omitted, so the unbounded read pays nothing per node. Unbounded stays
-  the default and the caller's choice; the follow-up read is a narrower `--root`, not a
+  tree (shipped — the Phase-2 bootstrap tracer, #7) from the running current scene
+  (`/root` only when none is current — an autoload is that scene's sibling, so seeing
+  one takes `--root /root`), bounded on request by `--root <runtime path>` (the subtree
+  to serialize; an unknown path is `live_node_not_found`) and `--max-depth N` (0 = the
+  addressed node alone) — shipped, #849, from GDA-DF-052, where an unfiltered read of a
+  production UI exceeded the client's budget and was truncated by IT, which a caller
+  cannot tell from a complete read. What a bound leaves out of the selected subtree is
+  counted rather than dropped: `omitted_nodes` totals the unserialized nodes at every
+  depth, `truncated` is that total above zero, and a node whose children were not walked
+  carries `children_omitted` — its DIRECT children only, and the key is absent when
+  nothing was omitted, so the unbounded read pays nothing per node. Unbounded stays the
+  default and the caller's choice; the follow-up read is a narrower `--root`, not a
   continuation token, which would page a snapshot the live tree has already left behind.
   Runtime node property `game get` /
   `game set` (shipped, #220, extended by #422/#473) read and mutate a running node's live
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -1322,7 +1322,18 @@ re-derives every verdict from a running engine.
   [`node`](#node).
 
 - **`game` (the running game's scene graph):** `game tree` reads the runtime scene
-  tree (shipped — the Phase-2 bootstrap tracer, #7); runtime node property `game get` /
+  tree (shipped — the Phase-2 bootstrap tracer, #7), bounded on request by `--root
+  <runtime path>` (the subtree to serialize; an unknown path is `live_node_not_found`)
+  and `--max-depth N` (0 = the addressed node alone) — shipped, #849, from GDA-DF-052,
+  where an unfiltered read of a production UI exceeded the client's budget and was
+  truncated by IT, which a caller cannot tell from a complete read. What a bound leaves
+  out is counted rather than dropped: `omitted_nodes` totals the unserialized nodes at
+  every depth, `truncated` is that total above zero, and a node whose children were not
+  walked carries `children_omitted` — its DIRECT children only, and the key is absent
+  when nothing was omitted, so the unbounded read pays nothing per node. Unbounded stays
+  the default and the caller's choice; the follow-up read is a narrower `--root`, not a
+  continuation token, which would page a snapshot the live tree has already left behind.
+  Runtime node property `game get` /
   `game set` (shipped, #220, extended by #422/#473) read and mutate a running node's live
   properties — the live counterparts of headless `node get` / `node set`, applying the
   **same** value-coercion table and returning the observed read-back value plus
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -20,8 +20,13 @@ import json
 from typing import Any, Optional
 
 import typer
-from pydantic import BaseModel, ConfigDict, Field, model_serializer
-from pydantic_core.core_schema import SerializerFunctionWrapHandler
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    SerializerFunctionWrapHandler,
+    model_serializer,
+)
 
 from gda.dispatch import dispatch_domain, params_or_bad_parameter
 from gda.execution import ExecutionKind
@@ -593,10 +598,11 @@ def game_tree(
     max_depth: Optional[int] = typer.Option(
         None,
         "--max-depth",
+        min=0,
         help=(
-            "Read at most this many levels below the root (0 = the root alone). "
-            "Unset is an unbounded read of the whole subtree, which on a large "
-            "tree is the caller's choice."
+            "Read at most this many levels below the root (0 = the root alone); "
+            "must be >= 0. Unset is an unbounded read of the whole subtree, "
+            "which on a large tree is the caller's choice."
         ),
     ),
     json_output: bool = json_option(),
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -103,10 +103,13 @@ class GameNode(BaseModel):
 class GameTreeParams(RelayedLiveParams):
     """The params of ``gda game tree``: read the running game's runtime scene tree (#849).
 
-    Unbounded by default — the whole runtime tree of the engine session held by
-    ``gda-daemon``, which on a production UI is a very large result. ``root``
-    narrows the read to one subtree and ``max_depth`` bounds how deep it goes;
-    what a bound leaves out is COUNTED rather than silently dropped, so a partial
+    Unbounded by default, and rooted at the running CURRENT SCENE (``/root``
+    only when no scene is current) — the whole subtree below it, which on a
+    production UI is a very large result. An autoload is that scene's SIBLING
+    under ``/root``, so a read that must see one names ``root="/root"``;
+    ``root`` otherwise narrows the read to one subtree and ``max_depth`` bounds
+    how deep it goes. Both counters cover the SELECTED subtree only, and what a
+    bound leaves out of it is COUNTED rather than silently dropped, so a partial
     read is never mistaken for a complete one (see :class:`GameTreeResult`).
     """
 
@@ -115,8 +118,10 @@ class GameTreeParams(RelayedLiveParams):
         description=(
             "Serialize the subtree at this runtime (absolute) node path, as "
             "`game tree` itself reports it (e.g. /root/Main/HUD). Unset reads "
-            "the running current scene. A path that resolves to nothing is "
-            "`live_node_not_found`, the same refusal every other live op gives."
+            "the running current scene — an autoload is its SIBLING under "
+            "/root, so name /root to see one. A path that resolves to nothing "
+            "is `live_node_not_found`, the same refusal every other live op "
+            "gives."
         ),
     )
     max_depth: int | None = Field(
@@ -591,8 +596,9 @@ def game_tree(
         "--root",
         help=(
             "Read the subtree at this runtime node path, as `game tree` reports "
-            "it (absolute, e.g. /root/Main/HUD). Unset reads the current scene; "
-            "a path that resolves to nothing is `live_node_not_found`."
+            "it (absolute, e.g. /root/Main/HUD). Unset reads the current scene "
+            "(autoloads are its siblings: name /root to see them); a path that "
+            "resolves to nothing is `live_node_not_found`."
         ),
     ),
     max_depth: Optional[int] = typer.Option(
@@ -621,12 +627,14 @@ def game_tree(
     `live_unsupported_platform`. The platform/Godot-version precondition is the
     structured `constraints` field of `--schema` (ADR-0021), not restated here.
 
-    `--root` and `--max-depth` bound the read; without them it is unbounded, and
-    a production UI's whole tree is a very large result. What a bound leaves out
-    is counted, never silently dropped: the result carries `truncated` and
-    `omitted_nodes`, and each node whose children were not walked carries
-    `children_omitted`. Read bounded first, then address the nodes you want by
-    their exact path (`game get`, `game rect`, `game set`).
+    `--root` and `--max-depth` bound the read; without them it is unbounded and
+    rooted at the running current scene, and a production UI's whole tree is a
+    very large result. An autoload is that scene's sibling under `/root`, so a
+    read that must see one names `--root /root`. What a bound leaves out of the
+    selected subtree is counted, never silently dropped: the result carries
+    `truncated` and `omitted_nodes`, and each node whose children were not walked
+    carries `children_omitted`. Read bounded first, then address the nodes you
+    want by their exact path (`game get`, `game rect`, `game set`).
     """
     dispatch_domain(
         GAME_TREE_COMMAND,
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -20,7 +20,8 @@ import json
 from typing import Any, Optional
 
 import typer
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, model_serializer
+from pydantic_core.core_schema import SerializerFunctionWrapHandler
 
 from gda.dispatch import dispatch_domain, params_or_bad_parameter
 from gda.execution import ExecutionKind
@@ -66,20 +67,88 @@ class GameNode(BaseModel):
     type: str
     path: str
     children: list["GameNode"] = []
+    children_omitted: int = Field(
+        default=0,
+        description=(
+            "How many of this node's DIRECT children the read did not serialize "
+            "because `max_depth` stopped it here (#849). ABSENT from the JSON "
+            "when it is 0: an unbounded read serializes every node, so it must "
+            "not pay one extra key per node for a bound it never had. Read it as "
+            "`node.get('children_omitted', 0)`; re-read that subtree with "
+            "`--root <this node's path>` to see what it left out."
+        ),
+    )
+
+    # The presence rule above is a SERIALIZATION rule, so it lives on the writer
+    # rather than on every caller: the field stays a plain int with a 0 default
+    # (a consumer reads a number, never None), and the key is dropped from the
+    # emitted JSON when nothing was omitted. `mode="wrap"` runs pydantic's own
+    # serializer first, so nested children are serialized — and pruned — by this
+    # same rule at every depth.
+    @model_serializer(mode="wrap")
+    def _omit_absent_omission_count(
+        self, handler: SerializerFunctionWrapHandler
+    ) -> dict[str, Any]:
+        rendered = handler(self)
+        if not rendered.get("children_omitted"):
+            rendered.pop("children_omitted", None)
+        return rendered
 
 
 class GameTreeParams(RelayedLiveParams):
-    """The params of ``gda game tree``: read the running game's runtime scene tree.
+    """The params of ``gda game tree``: read the running game's runtime scene tree (#849).
 
-    Empty — it reads the whole runtime tree of the engine session held by
-    ``gda-daemon`` (a subtree root may be added by a later slice).
+    Unbounded by default — the whole runtime tree of the engine session held by
+    ``gda-daemon``, which on a production UI is a very large result. ``root``
+    narrows the read to one subtree and ``max_depth`` bounds how deep it goes;
+    what a bound leaves out is COUNTED rather than silently dropped, so a partial
+    read is never mistaken for a complete one (see :class:`GameTreeResult`).
     """
+
+    root: str | None = Field(
+        default=None,
+        description=(
+            "Serialize the subtree at this runtime (absolute) node path, as "
+            "`game tree` itself reports it (e.g. /root/Main/HUD). Unset reads "
+            "the running current scene. A path that resolves to nothing is "
+            "`live_node_not_found`, the same refusal every other live op gives."
+        ),
+    )
+    max_depth: int | None = Field(
+        default=None,
+        ge=0,
+        description=(
+            "Serialize at most this many levels BELOW the read's root: 0 is the "
+            "root node alone, 1 adds its children. Unset reads the whole "
+            "subtree — an unbounded read of a large tree is the caller's choice."
+        ),
+    )
 
 
 class GameTreeResult(BaseModel):
-    """The result of ``gda game tree``: the running game's runtime scene tree."""
+    """The result of ``gda game tree``: the running game's runtime scene tree (#849).
+
+    ``truncated`` and ``omitted_nodes`` are always present, so a bounded read
+    reports what it left out instead of looking like a complete one. They are the
+    ONLY difference an unbounded read shows: no node in it carries
+    ``children_omitted``.
+    """
 
     root: GameNode
+    truncated: bool = Field(
+        description=(
+            "Whether a bound stopped the read short — true exactly when "
+            "`omitted_nodes` is above 0. Always false for an unbounded read."
+        )
+    )
+    omitted_nodes: int = Field(
+        description=(
+            "How many nodes, at every depth, the read did not serialize. It "
+            "counts whole omitted subtrees, not just the direct children a node "
+            "reports in `children_omitted`, so it is the size of what is "
+            "missing from this result."
+        )
+    )
 
 
 class GameGetParams(RelayedLiveParams):
@@ -407,13 +476,22 @@ class GameCallResult(BaseModel):
 
 
 def render_game_tree(game: "GameTreeResult") -> str:
-    """Render the running game's runtime scene tree (ADR-0019).
+    """Render the running game's runtime scene tree (ADR-0019, #849).
 
     The runtime counterpart of ``render_scene_tree``: ``render_node_tree`` reads
     only ``name``/``type``/``children``, which a ``GameNode`` carries, so the
     runtime tree flows through the same indented outline as the on-disk scene.
+
+    A bounded read appends ONE trailing line with the omitted total, so the
+    outline cannot be read as a complete tree on the human channel either. The
+    shared outline itself is left alone: the per-node count is a ``game`` shape,
+    and teaching the tree renderer about it would change what the on-disk
+    ``scene``/``node`` trees print (ADR-0040 §5).
     """
-    return render_node_tree(game.root)
+    outline = render_node_tree(game.root)
+    if not game.truncated:
+        return outline
+    return f"{outline}\ntruncated: {game.omitted_nodes} nodes omitted"
 
 
 def render_game_get(got: "GameGetResult") -> str:
@@ -503,6 +581,24 @@ _app = typer.Typer(
 
 @_app.command(name="tree", cls=GAME_TREE_COMMAND.command_class())
 def game_tree(
+    root: Optional[str] = typer.Option(
+        None,
+        "--root",
+        help=(
+            "Read the subtree at this runtime node path, as `game tree` reports "
+            "it (absolute, e.g. /root/Main/HUD). Unset reads the current scene; "
+            "a path that resolves to nothing is `live_node_not_found`."
+        ),
+    ),
+    max_depth: Optional[int] = typer.Option(
+        None,
+        "--max-depth",
+        help=(
+            "Read at most this many levels below the root (0 = the root alone). "
+            "Unset is an unbounded read of the whole subtree, which on a large "
+            "tree is the caller's choice."
+        ),
+    ),
     json_output: bool = json_option(),
     schema: bool = GAME_TREE_COMMAND.schema_option(),
     params_json: Optional[str] = params_json_option(),
@@ -518,10 +614,17 @@ def game_tree(
     naming the remediation (`gda daemon start`); on an unsupported platform,
     `live_unsupported_platform`. The platform/Godot-version precondition is the
     structured `constraints` field of `--schema` (ADR-0021), not restated here.
+
+    `--root` and `--max-depth` bound the read; without them it is unbounded, and
+    a production UI's whole tree is a very large result. What a bound leaves out
+    is counted, never silently dropped: the result carries `truncated` and
+    `omitted_nodes`, and each node whose children were not walked carries
+    `children_omitted`. Read bounded first, then address the nodes you want by
+    their exact path (`game get`, `game rect`, `game set`).
     """
     dispatch_domain(
         GAME_TREE_COMMAND,
-        GameTreeParams(),
+        params_or_bad_parameter(GameTreeParams, root=root, max_depth=max_depth),
         json_output=json_output,
         godot=godot,
         project=project,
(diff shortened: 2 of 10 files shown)


qinhaihong-red:
**Review record (orchestrator, 2026-09-06).** Independent adversarial review at `539ffff27` (spec vs #849 on the reviewer's own fixture against a real v20 session, harness confinement, standards/surfaces, mutation) returned **READY** — 0 P1 / 0 P2 / 5 P3. Judgments the review was asked for: `children_omitted` absent-when-zero is SOUND (AC3's "unchanged except" is literal; the schema's `default: 0` is the annotation a consumer applies to an absent key; note for maintainers — a wrap serializer collapses `model_json_schema(mode="serialization")`, so the schema builder must stay in validation mode); the `_subtree_size` walk is REQUIRED by the AC's all-depths total and allocates nothing; the AC4 guard is adequate (dropping `truncated` reddens 3 unit tests); README deliberately unchanged, agreed. Base-vs-head harness comparison on one fixture: unbounded subtree IDENTICAL, only `truncated`/`omitted_nodes` added.

| Finding | Disposition |
|---|---|
| P3-1 `SerializerFunctionWrapHandler` imported from `pydantic_core.core_schema` (two existing sites use the `pydantic` re-export) | **Adopted** — `40d675a0b` |
| P3-2 `--max-depth` did not mirror `ge=0` on the Typer option, so the argv refusal named `max_depth`, not the flag (repo convention: `diag --limit`, `perf --frames`, …) | **Adopted** — `min=0` + help; the test pin flipped to the flag spelling (red-proof quoted in the body); schema byte-identical |
| P3-3 harness comment/PR body claimed #855's `game find` "can call `_serialize` as-is" — it renders a NESTED subtree, a flat match list is a different walk | **Adopted** — reworded to the genuinely shared parts (per-node shape; depth-bounding + omitted-count contract; `_subtree_size` reusable); harness bytes changed → sha re-pinned `cb5dd790…`, version stays 20 |
| P3-4 SKILL "Every live op takes an exact runtime path" (false for `diag`, `logger`, `perf monitors`, `screen`, `daemon`) | **Adopted** — "Every live op that ADDRESSES a node …" |
| P3-5 unused `monkeypatch`/`tmp_path` on a new help test | **Adopted** |

Structural observation recorded for the milestone (not a change): the counting semantics have one guard and it is e2e — PR CI does not run it; the harness byte pin is the compensating control. Orchestrator re-verify at `40d675a0b`: 186 unit (game/harness/schema), guard valid, sha matches the pin, `pydantic_core` import gone, `--help` shows `INTEGER RANGE [x>=0]`, schema identical to the reviewed head, `Closes #849`, no attribution, CI 8 success / 4 policy skips, MERGEABLE. Engine e2e will be re-run by the orchestrator serially before the merge slot (order: last of W1).

qinhaihong-red:
**Changes required — a P2 omitted-node counting defect and two minor artifact corrections.**

Independent adversarial review of head **`40d675a0b069f517ccc3ffa57a6b45370443a001`**, against base/merge-base **`3f68bf3fb26df2ab54351a8ef4e3e167269bdc16`**. Order: Artifact → Architecture → independent parallel Standards / Spec. Previous reviews and replies were not used as evidence.

## Artifact

**[P3] State the default root and the scope of the counters accurately.** [#849's baseline description](https://github.com/aigengame/godot-agent/issues/849) says the old default serializes the whole tree from `/root`, and the changed [params-model description](https://github.com/aigengame/godot-agent/blob/40d675a0b069f517ccc3ffa57a6b45370443a001/src/gda/commands/game.py#L103-L118) still calls it the whole runtime tree. Both base and head instead select `current_scene`, falling back to `/root` only when there is no current scene. On a live fixture with `/root/Main` and a sibling `/root/Sidecar` autoload, the default result starts at Main; explicitly passing `--root /root` also shows Sidecar and the harness. The counts apply only inside the selected subtree. Correct the issue, general description and inspection guidance accordingly, including the explicit `/root` spelling when discovering autoloads. Preserve the existing default as AC3 requires. This prevents an agent from treating a complete current-scene read as evidence that a sibling runtime node is absent.

**Minor — [P3] Correct the PR body's upper-bound explanation.** “Wire numbers” says `ge=0` bounds `max_depth` far below `MAX_EXACT_JSON_INT`. It only rejects negatives. Fresh model checks accept `2147483648` and `9007199254740991`; `9007199254740992` is refused by inherited `RelayedLiveParams` validation. Attribute the wire limit to that shared guard. No extra depth cap is requested.

Read the complete command catalog, bundled skill, PR body and current **OPEN / ready-for-agent** issue #849, plus the full diff and relevant accepted ADRs. Checked the definitions of direct children versus all omitted descendants, absent-when-zero serialization, default compatibility, schema/help and the two-step discovery procedure. The small acceptance fixture works, but the all-depth counting claim is disproved by the Spec counterexample below.

**Terminology and prose assessment:** Sources: `CONTEXT.md`, #849, ADR-0017/0018/0019/0020/0023/0040, and the model/harness contracts. Reference: **ASD-STE100 Simplified Technical English**. Checked changed and normative/procedural passages for stable technical nouns, condition/action order, references, selected-root scope, and direct-versus-total counts; sampled each remaining major section's prose. The two factual scope/validation statements above are the artifact findings. No other substantive changed-scope prose issue was found. No formal dictionary-level STE certification was performed.

## Architecture — Pass on module boundaries

Keep the existing topology: the game command group owns params/result models and human rendering; the shared descriptor/live runner carries the request; the harness resolves the runtime node and owns traversal/counting. The omission serializer stays with `GameNode`, and the existing shared outline renderer remains independent of the game-specific count. Imports and communication remain directed; no new registry, reciprocal group dependency or speculative abstraction appears.

Completeness, orthogonality, DRY and extension checks support retaining these owners. Accepted ADR-0018/0020/0023/0040 remain authoritative. Claim evidence state: **open**. The #849 exact-count invariant belongs to the harness; the deep-tree counterexample disconfirms its current implementation. Repair that existing owner rather than introducing another module. Source tracing and macOS engine checks support the retained boundaries; cross-platform and unrestricted-depth correctness are not established.

## Standards — Pass

No documented code-standard violation or actionable Fowler smell was found. Input validation and omission serialization remain model-owned; the human omission marker stays group-local; the harness version advances from 19 to 20 and its byte pin matches. Artifact accuracy and the executable counterexample are reported in their respective axes.

## Spec — Changes required

**[P2] Count omitted subtrees without exhausting the GDScript call stack.** #849 requires: “`omitted_nodes` on the result is the total number of nodes (all depths) the read did not serialize.” The new recursive [_subtree_size call](https://github.com/aigengame/godot-agent/blob/40d675a0b069f517ccc3ffa57a6b45370443a001/src/gda/harness/gda_harness.gd#L1871-L1875) continues through every omitted descendant even with `--max-depth 0`. A finite, valid chain of 2200 ordinary child Nodes overflows this helper, yet the operation publishes a successful undercount.

Reproduce with a main scene named Main whose attached script builds the chain iteratively:

```gdscript
extends Node2D
@export var built_count: int = 0
func _ready() -> void:
    var current: Node = self
    for index in range(2200):
        var child := Node.new()
        child.name = str(index)
        current.add_child(child)
        current = child
        built_count += 1
```

After `gda daemon start`, the observed Godot 4.6.3 results are:

```text
game get /root/Main --property built_count → 2200
diag errors → []
game tree --max-depth 0 → exit 0, children_omitted: 1, omitted_nodes: 2044
game tree --root /root --max-depth 1 → exit 0, omitted_nodes: 2043
diag errors → Stack overflow in _subtree_size
```

Both tree calls omit the same 2200 descendants and must report 2200. The different partial totals cannot be used as the promised exact counts. An independent reviewer restarted the fixture and reproduced both results at the same clean head. Use an explicit stack for the count-only helper and add a deep-chain regression that checks the exact total and clean diagnostics. The evidence does not require rewriting the whole serializer.

## Validation

- **186 passed, 4 deselected**: game-command, schema and harness tests excluding E2E.
- **14 real-Godot E2E tests passed, 51 deselected**: game and harness modules on macOS/Godot 4.6.3.
- Additional live probes covered the current-scene/root distinction and chains of 16, 1200 and 2200 descendants. The 16/1200 counts were correct; 1200 introduced engine stack diagnostics; 2200 produced the successful undercounts above. The decisive case was independently reproduced.
- Applicable PR CI checks passed; hosted Godot E2E was skipped. No source edits. Full repository/nightly, other platforms and a fresh lint/type run were not repeated. Passing existing tests does not cover the demonstrated deep-count failure.

Standards: **0 findings**. Spec: **1 finding, worst P2**.

qinhaihong-red:
**Review round 2 handled (orchestrator, 2026-09-06).** Review at `40d675a0b`; each finding re-checked against that head before disposition. Fix head: `092fdec0b`.

|---|---|---|
| **P2 (Spec)** `_subtree_size` recursed through every omitted descendant; a 2200-node chain overflowed the GDScript call stack and the op published a successful undercount | Reproduced on Godot 4.6.3 with the new regression run against the previous harness: `assert 2044 == 2200` — the review's exact number | **Adopted.** The count-only walk keeps its own LIFO (`Array[Node]`), no call depth per node. New live regression `test_game_tree_counts_an_omitted_subtree_deeper_than_the_call_stack`: chain built by the main scene's script, `children_omitted: 1` / `omitted_nodes: 2200` from the current scene AND from `--root /root --max-depth 1`, `diag errors` empty. Red on the old harness, green on `092fdec0b`. `_serialize` untouched, as asked; an UNBOUNDED read of a tree deeper than the call stack remains the base serializer's pre-existing behaviour, outside #849. |
| P3 (Artifact) Issue baseline and params docstring say the default read is the whole tree from `/root`; the harness selects `current_scene`, `/root` only as fallback | Confirmed in `_handle_game_tree` (both base and head) | **Adopted.** Params docstring, `--root` help/schema description, command docstring, catalog and the skill's two-step pattern now name the default root, say the counters cover the SELECTED subtree only, and give `--root /root` for discovering an autoload. #849 body: dated correction + comment. Default unchanged (AC3). |
| P3 (Artifact) PR body "Wire numbers": `ge=0` bounds `max_depth` far below `MAX_EXACT_JSON_INT` | Confirmed by a model probe: `2147483648` and `9007199254740991` accepted, `9007199254740992` refused by the inherited `RelayedLiveParams` guard | **Adopted.** PR body corrected; the ceiling is attributed to `RelayedLiveParams`; no extra depth cap. |

`HARNESS_VERSION` stays 20 (merge base declares 19); pin re-taken (`1074e1ec…0ea18`); `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` valid. On `092fdec0b`: unit 369 passed (game/harness/meta/live-guard modules), ruff/format/pyright clean, e2e game + harness 15 passed (macOS, Godot 4.6.3). Ready for the user's merge (last in the wave, as the harness slice).

qinhaihong-red:
**Pass — all three findings from the previous review are resolved; no new finding in the repair scope.**

Re-reviewed **`092fdec0bd6f3ee872504632dd6a77e11d886a44`**, previous reviewed head **`40d675a0b069f517ccc3ffa57a6b45370443a001`**, original PR base/merge-base **`3f68bf3fb26df2ab54351a8ef4e3e167269bdc16`**. Scope: the prior findings, their fixes, and adjacent behavior. Order: Artifact → Architecture → independent parallel Standards / Spec. Author replies were checked against source and fresh execution.

## Artifact — Pass

The issue #849 baseline, params/schema descriptions, help, catalog and bundled skill now state that the default root is the current scene, with `/root` as the fallback when no scene is current. They explain explicit `--root /root` discovery for autoloads and scope omission counters to the selected subtree. The default selection code is unchanged. A fresh fixture with a sibling `Sidecar` confirms the distinction.

The PR body's wire-number explanation now correctly separates the model's `ge=0` lower bound from the inherited `RelayedLiveParams` ceiling. Fresh model checks reject `-1`, accept `2147483648` and `9007199254740991`, and reject `9007199254740992` through the inherited wire guard.

**Terminology and prose assessment:** Sources: `CONTEXT.md`, current **OPEN / ready-for-agent** issue #849, accepted ADR-0018/0020/0023/0040, and the command/harness contracts. Reference: **ASD-STE100 Simplified Technical English**. Read the complete current PR body, fix reply and issue, and every changed documentation/comment passage with its relevant surrounding guidance. Checked stable technical nouns, selected-root scope, direct children versus all omitted descendants, condition/action order, and lower-bound versus wire-ceiling attribution. No substantive terminology or prose residue was found. No formal dictionary-level STE certification was performed.

## Architecture — Pass

The retained structure is `commands/game.py` → shared descriptor/live dispatch → `harness/gda_harness.gd`. The game group owns models and human rendering; the harness owns node resolution, serialization and the omitted-subtree count. The repair replaces recursion inside that existing counting owner with a local LIFO; it preserves synchronous execution and `get_children()` semantics. No ownership overlap, dependency cycle, duplicated rule or speculative extension was introduced.

For the load-bearing #849 exact-count claim, claim evidence state remains **open** because no project vocabulary establishes a stronger state. Evidence now includes the exact deep-chain results below and a regression that rejects the old helper. A mismatch between constructed descendants and `omitted_nodes`, or traversal diagnostics, would disconfirm the claim. The repaired count-only boundary has no remaining counterexample in the inspected scope; unrestricted serialization is not established by this evidence.

## Standards — Pass

The independent Standards reviewer found no documented-standard violation or actionable Fowler smell. Harness version **19 at the original PR base → 20 at HEAD** satisfies the guard. The current harness SHA-256 matches its updated pin: `1074e1ecb10913b0e020e4ac99991631840cbd8966dce4c8f928c4e994b0ea18`.

## Spec — Pass

The prior P2 is resolved. Replaying the original independent probe on Godot 4.6.3 produced:

```text
built_count                                      2200
game tree --max-depth 0                           omitted_nodes=2200; children_omitted=1
game tree --root /root --max-depth 1               omitted_nodes=2200; Main.children_omitted=1
diag errors                                      []
```

Both operations returned success. Explicit `/root` also showed `Sidecar`; the default read remained rooted at `/root/Main`.

The new repository regression passed at HEAD. In a separate temporary package copy, replacing only `_subtree_size` with the previous recursive helper made that same new test fail at its exact-count assertion: **`assert 2044 == 2200`**. The reviewed checkout remained unchanged. This proves the regression distinguishes the reported defect.

## Validation

- **193 passed, 4 deselected**: game-command, harness, schema and skill-surface tests excluding E2E.
- **15 real-Godot E2E tests passed, 51 deselected**: game and harness modules on macOS/Godot 4.6.3, including the new deep-chain regression.
- Fresh original 2200-node probe, model boundary probe, harness byte pin and committed-base version guard passed.
- No source changes during review. Full repository/nightly, other platforms and a fresh full lint/type run were not repeated. Deep serialization remains the unchanged pre-existing boundary; this repair proves the count-only traversal.

Current-head CI: all executed checks passed; the GitHub Godot E2E job was **skipped**, so the real-engine evidence here comes from the local runs above.

Standards: **0 findings**. Spec: **0 findings**.

PR #882, opened 2026-09-06
refactor(gda): lift the import-evidence adapter into a core import_evidence module (#741)

qinhaihong-red:
## Summary

`resource import`'s faithful mirror of `EditorFileSystem::_test_for_reimport` —
the read that decides `cached` / `missing` / `stale` / `invalid` — lived as
module-level regexes and private helpers inside the resource command group, so
its only interface was the CLI and every evidence question in the unit suite paid
the full Typer + JSON-envelope toll. This lifts it into the core module
`gda/import_evidence.py`.

ADR-0040 placement: the group keeps its whole command slice (params/result
models, render, recipe, command body) and the core hosts the non-command engine
knowledge, as `binary` and `display` already do. Dependency direction is
commands → core only — the new module imports `hashlib`, `json`, `re`,
`dataclasses`, `pathlib` and `typing`, and nothing from `gda`.

What the module owns: the receipt grammar (`_parse_receipt_assignments`), the
sidecar/receipt regexes, `asset_state`, `project_import_gaps`, `CACHE_ROOT_REL`,
`classify_created_file` + `CreatedFileClass`, and `EvidenceStatus` +
`AssetEvidence`. The `_test_for_reimport` parity contract moved **intact** out of
`_asset_state`'s docstring into the module docstring — remainder and one-way
direction included, re-indented and not reworded.

What stayed in the command: request normalization (`_asset_res_path`), the
created-file **accounting** (`_project_files`, the before/after diff,
`_summarize`), the settlement mapping (`imported` / `not_importable` / `failed`),
and the wire models. `asset_state` returns the narrowed `AssetEvidence`, whose
`EvidenceStatus` is four-valued: the adapter can never return a settlement. A
thin `_asset_record` maps evidence onto the wire `ResourceImportAsset`.

Closes #741

## Surface diff

**byte-identical.** The same harness the ADR-0040 restructure used, with the
baseline generated in a separate clean checkout at the base commit `cfcb8658e`
(never via `git stash` in the working worktree).

| Artifact | Base sha256 | Head sha256 |
| --- | --- | --- |
| `gda schema` | `abb4a866bd22e5fde0fb101217d9aea3a43d9f56eb69a99dcf6eae0b225aa6c3` | `abb4a866bd22e5fde0fb101217d9aea3a43d9f56eb69a99dcf6eae0b225aa6c3` |
[...]

PR #883, opened 2026-09-06
fix(gda): export names both template directories when --user-data-root hides the host's (#840)

### LINKED ISSUE #840: export: --user-data-root hides the host's installed export templates
## Dogfooding source

- **GDA-DF-073** (2026-08-27, gda 0.11.0) — `gda --user-data-root /private/tmp/... export run --preset macOS --json` failed with `export_templates_missing`; the same binary and preset reported `templates_installed: true` and exported when the option was omitted. The installed archive stayed at the standard `export_templates/4.6.3.stable/macos.zip` path throughout.
- **PIPE-DF-164** (2026-09-03, gda 0.12.0) — reconfirmed on a clean sparse worktree; the runner had to drop the isolation for export commands only.

Verified against gda 0.13.0 (`266c8009f`): the template check derives its directory from `OS.get_data_dir()` in the engine-side operation, and the `User-data placement` (`--user-data-root`) relocates exactly that directory (`HOME` on macOS, `XDG_DATA_HOME` on Linux, `APPDATA` on Windows — the only lever Godot exposes; #653, ADR-0022). The native `--export-*` launch runs under the same redirected environment, so it would not find the templates either. The root `gda --help` entry for `--user-data-root` does state that a release/debug `export run` under it finds no installed templates and that `--mode pack` needs none; nothing at the point of failure does: the `export_templates_missing` envelope names only the templates version — no root and no remedy — `export get` reports `templates_installed` and `templates_version` but not the directory it checked, and `export run --help` is silent. This repo's PITFALLS.md records the trap for the e2e suite.

## What to build

Make the interaction visible at the point of failure, without weakening the isolation:

- When the export preflight runs under a redirected root and the templates are absent there but present in the host's standard data directory, the `export_templates_missing` message names both directories, states that `--user-data-root` moved the lookup, and gives the remedy in prose: run the export without the redirect, or use `--mode pack` when no template is needed. No `hint`: the envelope's `hint` is contractually one corrected invocation from the curated near-miss table (CONTEXT.md `Near-miss hint`), and this failure is not a near miss.
- The two directories also ride `evidence` as typed facts (`templates_root_checked`, `templates_root_host`), admitted through ADR-0004's #687 gate: computed on the failure path, unrecoverable without parsing prose, and they change what the caller does next. `FailureEvidence` is `extra="forbid"`, so the fields are declared, not improvised.
- Mechanism: the check runs engine-side inside the redirected environment and cannot see the host directory, and the templates layout rule (`godot` directory name, version directory without a `.0` patch) lives only in the engine-side operation. gda passes the HOST data path (`engine_data_path` over the unredirected environment, already computed for the launch) into the op as a param, so the one layout rule stays where it is and no CLI-side copy appears.
- `export get` reports the template directory it checked (`templates_root`), so the hidden case is readable before an export. (Amended 2026-09-06, PR #883: it also reports `templates_root_host` — present only when a redirect hid templates that the host's own directory holds — because the engine-side check is the only side that can look at the host directory, and the CLI preflight branches on that result; `export get --help` states the effect of the redirect too. Additive to the acceptance criteria; see the dated comment.)
- `export run --help` states that export-template discovery follows the redirected root, as the root option's help already does. Note for the implementer: on macOS the redirect sets `HOME`, so the relocated data directory is `<root>/Library/Application Support`, not `<root>` itself.

Out of scope, decided here: mirroring the host's `export_templates/<version>` into the isolated root (symlink or copy). It duplicates engine-installed files into a task-scoped directory, and the two remedies above cover the need; reopen as its own issue if the disclosure proves insufficient.

## Acceptance criteria

- [ ] With templates installed in the host data directory and an isolated root that has none, `export run --json` fails with `export_templates_missing` whose message names both directories and both remedies, and whose `evidence` carries `templates_root_checked` and `templates_root_host`.
- [ ] The same failure without a redirect (templates genuinely absent) names the one directory checked and no host directory; the two shapes are distinguishable by the presence of `templates_root_host`.
- [ ] `export get --json` reports `templates_root`.
- [ ] `export run --help` states the effect on template discovery.
- [ ] Unit coverage for the hidden-by-redirect branch; the existing missing-templates e2e keeps passing; the templates layout rule is not duplicated CLI-side.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Summary

Godot reads the export templates from the data directory `--user-data-root`
relocates, so a `release`/`debug` `export run` under an isolated root failed with
`export_templates_missing` on a host whose templates were correctly installed —
and named no directory at all, so the remedy read as "install the templates" when
they were already there (GDA-DF-073, PIPE-DF-164).

The check stays engine-side, where the templates layout rule (the `godot`/`Godot`
directory name, the version directory without a `.0` patch) already lives.
`operations.gd` now composes the export-templates root from a data directory given
as an ARGUMENT, so `export-get` can check two of them: the run's own
`OS.get_data_dir()`, and the host's, which gda stamps on the params model from its
own — never redirected — environment. It reports `templates_root` always, and
`templates_root_host` only when the redirect really did hide installed templates.

The preflight hands both to `export_templates_missing_failure`: the message names
both directories and the two remedies, and `evidence` carries
`templates_root_checked` / `templates_root_host`, so the two shapes are told apart
without parsing prose. No `hint` — this is not a near miss.

Live, on this host:

```
$ gda --user-data-root /tmp/…/iso840 export run --preset "Linux/X11" --json   # exit 4
{"error":{"category":"operation","code":"export_templates_missing",
 "message":"export preset \"Linux/X11\" cannot be exported: the export templates for the
  running engine version (4.6.3.stable) are not installed in /tmp/…/iso840/Library/
  Application Support/Godot/export_templates, where --user-data-root (or $GDA_USER_DATA_ROOT) moved the lookup;
  they are installed in the host's /Users/…/Library/Application Support/Godot/
  export_templates. Run the export without the user-data redirect, or use --mode pack, which
  needs no export templates","diagnostics":"",
 "evidence":{"templates_root_checked":"/tmp/…/iso840/Library/Application Support/Godot/
  export_templates","templates_root_host":"/Users/…/Library/Application Support/Godot/
  export_templates"}}}
```

Closes #840

## Surface diff

`gda schema`: 1,058,760 → 1,105,818 bytes (+47,058, +4.4%). Byte-diffed against the
BASE generated in a separate checkout at `cfcb8658e`; every leaf enumerated, nothing
removed:

| Where | Change |
| --- | --- |
| `FailureEvidence` (×76, the one shared `error` schema per command) | `+ templates_root_checked`, `+ templates_root_host` — both optional, defaulted null, ADR-0004's documented per-command payload cost |
| `export get` `input` | `+ host_data_path` (optional string; COMPUTED — a value passed in is ignored, registered in `_COMPUTED_PROPERTIES` beside `script set`/`shader set`'s `mode`, so the "every supplyable property has an argv binding" invariant still holds) |
| `export get` `output` | `+ templates_root` (required string), `+ templates_root_host` (required key, nullable value — the `engine_data_path` shape) |
| `export get` / `export run` `description`, `export get` `input`/`output` `description` | the docstring sentences below |
| `gda export run --help` | + 5 lines: template discovery follows `--user-data-root`; the failure names both directories and the remedies; `--mode pack` needs none |
| `gda export get --help` | + 5 lines: what `templates_root` is, and when `templates_root_host` is set |
| `docs/command-catalog.md` | + one paragraph under `### export`: the two result keys, the two message remedies, the two evidence keys and which one discriminates |
| `src/gda/skill/SKILL.md` | the `--user-data-root` "two limits" bullet now names the evidence keys and how to read them; the `export` row notes `get`'s two new keys |
| `docs/adr/0004-…` | dated addendum to "Who carries it today" (appended; the paragraph's "Seven builders" stays as the record of what #687 decided) |
| `README.md` | **untouched** — it documents neither export templates nor `--user-data-root`'s limits today, so no behavioral boundary it carries changed. No `docs/README.*.md` restamp needed. |

`export get --schema`'s `argv` is byte-identical (the new input property is computed,
so it binds to no flag).

## Red-proof evidence

Tests first, with ONLY test files modified, at `origin/feat/gda-dogfooding-dev`
(`cfcb8658e`) — commit `83278df66`:

```
FAILED tests/export/test_export_commands.py::test_export_get_json_reports_preset_details_and_template_status
FAILED tests/export/test_export_commands.py::test_export_get_reports_the_templates_directory_it_checked
FAILED tests/export/test_export_commands.py::test_export_get_reports_the_host_templates_a_redirect_hides
FAILED tests/export/test_export_commands.py::test_export_get_human_render_names_the_checked_and_hidden_directories
FAILED tests/export/test_export_commands.py::test_export_get_hands_the_operation_the_host_data_path
FAILED tests/export/test_export_commands.py::test_the_host_data_path_reaches_the_operation_through_params_json
FAILED tests/export/test_export_run_commands.py::test_export_run_help_states_that_templates_follow_the_redirected_root
FAILED tests/export/test_export_run_operation.py::test_templates_hidden_by_a_redirect_name_both_directories_and_the_remedies
FAILED tests/export/test_export_run_operation.py::test_templates_genuinely_absent_name_only_the_directory_checked
FAILED tests/cli/test_error_registry.py::test_only_the_recorded_producers_put_evidence_on_the_envelope
10 failed, 79 passed, 18 deselected in 1.82s
```

Two representative failure lines:

```
assert '/home/dev/Library/Application Support/Godot/export_templates' in
  'export preset "Linux/X11" cannot be exported: the export templates for the running
   engine version (4.6.3.stable) are not installed'

AssertionError: assert {'launch_time...failure', ...} == {'export_temp...failure', ...}
  Extra items in the right set: 'export_templates_missing_failure'
```

`test_no_evidence_at_all_when_no_directory_was_reported` passed at the base by
construction (no producer emitted evidence there); it is a guard on the
omitted-never-null rule, not a red-proof test.

## Validation

Host: macOS (Darwin 25.6.0), Godot 4.6.3 at `$GDA_GODOT`, export templates
installed in the real data directory. All from the worktree with the repository
runtime (`uv run --frozen …`).

| Gate | Command | Result |
| --- | --- | --- |
| Unit suite | `pytest tests -m "not e2e" -q -p no:cacheprovider` | `2513 passed, 668 deselected in 46.13s` (base 2501 → +12) |
| Lint | `ruff check .` | `All checks passed!` |
| Format | `ruff format --check .` | `509 files already formatted` |
| Types | `pyright` (src+tests) | `0 errors, 0 warnings, 0 informations` |
| Schema/help byte-diff | `gda schema`, `export run --help`, `export get --help`, `export get --schema` vs the base checkout | only the diffs enumerated above; 0 leaves removed |
| Export e2e (real engine, serialized) | `pytest tests/export -m e2e -rs -q` | `17 passed, 1 skipped in 23.46s` (the skip is the pre-existing Linux-only `XDG_DATA_HOME` test) |
| Adjacent e2e | `pytest tests/harness tests/meta -m e2e -q` | `8 passed` |
| Adjacent e2e (user-data placement) | `pytest tests/runtime -m e2e -q` | `33 passed in 92.26s` |
| Live manual | `gda --user-data-root … export run/get` on a scratch project, both channels | the JSON envelope above; the human channel renders `evidence:` with both paths |

Every Godot-spawning run went through the machine-wide serial lock. The redirect
was passed PER INVOCATION, never exported for a run (PITFALLS.md: exporting
`GDA_USER_DATA_ROOT` hides the host's templates from unrelated tests — notably
`test_templates_installed_is_true_when_host_templates_are_on_disk`, which passed).

## Requirement conformance

| Acceptance criterion | Proof |
| --- | --- |
| Templates on the host + an isolated root with none → `export_templates_missing` naming both directories and both remedies, `evidence` carrying `templates_root_checked` and `templates_root_host` | e2e `test_export_run_under_a_redirect_names_both_template_directories`; unit `test_templates_hidden_by_a_redirect_name_both_directories_and_the_remedies`; the live run quoted above |
| The same failure with no redirect names ONE directory and no host one; the shapes differ by the presence of `templates_root_host` | unit `test_templates_genuinely_absent_name_only_the_directory_checked` — asserts `--user-data-root` is absent from the message and that the emitted evidence object is exactly `{"templates_root_checked": …}` |
| `export get --json` reports `templates_root` | e2e `test_export_get_reports_preset_details_and_template_status` (shape, real engine) and `test_export_get_names_the_host_templates_a_user_data_redirect_hides` (redirected, real engine); unit `test_export_get_reports_the_templates_directory_it_checked` |
| `export run --help` states the effect on template discovery | `test_export_run_help_states_that_templates_follow_the_redirected_root`, plus the help byte-diff above |
| Unit coverage for the hidden-by-redirect branch; the existing missing-templates e2e keeps passing; the layout rule is not duplicated CLI-side | the two operation tests + four `export get` tests; `tests/export -m e2e` is green including the two pre-existing missing-templates paths; the only place `<godot-dir>/export_templates` is composed is `operations.gd::_export_templates_root`, now taking the data dir as an argument — gda passes a data DIRECTORY, never a templates path |
| No `hint` on this failure | asserted in both the unit test (`error.hint is None`) and the e2e (`"hint" not in err`) |

## Disclosed extras and boundary crossings

- **Eighth `Failure evidence` producer.** `export_templates_missing_failure` joins
  the axis, so ADR-0004's "Who carries it today" gets a dated addendum (appended,
  not rewritten — the "Seven builders" count stays as the record of what #687
  decided) and `_EVIDENCE_PRODUCERS` in `tests/cli/test_error_registry.py` gains
  the name. It is the second producer that reports on no run: it is the pre-export
  preflight's verdict. The omitted-never-null rule holds at the producer — a reply
  naming no directory yields no `evidence` key rather than `{}` (pinned by
  `test_no_evidence_at_all_when_no_directory_was_reported`). It is NOT added to
  `test_no_producer_can_emit_an_empty_evidence_object`, which indexes
  `["evidence"]` unconditionally and is scoped to the original five; the dedicated
  test above covers the same rule for this builder.
- **`FailureEvidence` publishes two more fields, so the human failure renderer
  grew two branches** (`gda/render.py`) — the hand-written renderer is kept
  exhaustive by `test_the_sample_table_covers_every_field_the_model_publishes`,
  which reds when the model grows a field.
- **`export get` gains a COMPUTED input property**, not just result fields. The
  issue prescribes passing the host data path "into the op as a param"; making it
  a model field with a `model_validator` (rather than something the argv body
  pastes in) is what keeps ADR-0015's rule that argv and `--params-json` build the
  same model — `--params-json '{"preset": "X"}'` reaches the operation with the
  identical params, asserted by
  `test_the_host_data_path_reaches_the_operation_through_params_json`. A value
  passed in is ignored, which is the existing declared exemption
  (`script set`/`shader set`'s `mode`) rather than a new one.
- **`export get --help` also gained a sentence**, beyond the `export run --help`
  the issue asks for. `export get` is the command that gained `templates_root`, so
  leaving its own help silent about it would have put the disclosure only in the
  catalog.
- **Considered and declined: rewording the `export_templates_missing` registry
  `Meaning`.** Both shapes are the same code with the same cause from the run's
  point of view (the templates are not in the directory it reads), so the
  `error_codes.py` row and the ADR-0002 table are left byte-identical and the
  registry-consistency test needs no change.
- **Out of scope, as the issue decided:** mirroring the host's
  `export_templates/<version>` into the isolated root.

## Guards on PR CI vs the Godot e2e tier

PR CI runs NO Godot e2e, so on a PR run these guards fire:

- every unit test above — the two failure shapes, the evidence contents and the
  emitted-envelope shape, the four `export get` result/param tests, both help
  assertions, the schema-surface invariants
  (`test_every_directly_supplyable_input_property_has_an_argv_binding`,
  `test_the_published_error_schema_declares_the_optional_evidence_key`,
  `test_every_evidence_field_is_optional_in_the_published_schema`), the
  human-render tables, and the AST-read producer set.

Only in the Godot e2e tier (nightly / dispatch):

- `test_export_run_under_a_redirect_names_both_template_directories` and
  `test_export_get_names_the_host_templates_a_user_data_redirect_hides` — both
  SKIP on a runner with no export templates installed, since there is then nothing
  for the redirect to hide. The CI runner does install templates (#301), so both
  RUN there.
- `test_export_get_reports_preset_details_and_template_status`'s new
  `templates_root` shape assertions — no skip, they run wherever the tier runs.
- The pre-existing `test_export_run_without_templates_yields_export_templates_missing`
  stays Linux-only (it forces missing templates through `XDG_DATA_HOME`, which
  only the Linux/BSD engine honours) and still asserts the PLAIN shape there:
  `extra_env` reaches gda's own process too, so gda's host data path resolves to
  the same empty directory and `templates_root_host` is null — no new skip, and
  its assertions are unchanged.

## Review round 1 (`8cf2c7b9c`)

Independent adversarial review at `db6564ced` (four axes + 10 mutants, byte-diff regenerated from
a clean base tree, both remedies exercised on a real engine) returned CHANGES REQUIRED — 0 P1 /
1 P2 / 5 P3, all adopted:

1. **P2** — the message and remedy named `--user-data-root` alone; `$GDA_USER_DATA_ROOT` is the
   same switch and the exported-variable case is the common trigger (PITFALLS), where the caller
   never typed the flag. The message now says `--user-data-root (or $GDA_USER_DATA_ROOT) moved
   the lookup` and the remedy `Run the export without the user-data redirect`; the quoted
   samples above are updated; every existing `"--user-data-root" in message` assertion holds.
2. **P3** — the builder's two speculative defaults dropped (its one caller passes both;
   `ExportGetResult.templates_root` is a required `str`).
3. **P3** — surviving mutant: nothing pinned that a pasted `host_data_path` is ignored. The
   params-json parity test now also passes `"host_data_path": "/pasted/by/the/caller"` and
   asserts the op still receives the resolved host path (reachable via `--params-json` and MCP).
4. **P3** — stale "computed by the production default" comment; the tests now import the one
   resolver instead of re-deriving it; the adjacent pre-existing "grows a sixth field" comment
   corrected.

On `8cf2c7b9c`: `tests/export` + `tests/cli` 480 passed, ruff/format/pyright clean, export e2e
17 passed + the pre-existing Linux-only skip (Godot 4.6.3, macOS, load 3.5). Reviewer-verified
and unchanged by the fix: 774 leaves added / 0 removed / 4 changed; layout rule engine-side only;
host path resolved over the unredirected environment even with `GDA_USER_DATA_ROOT` set; plain
shape omits `templates_root_host` on both renderings; ADR-0002 table + `error_codes.py`
byte-identical; the Linux-only e2e still asserts the plain shape by construction.

## Review round 2 (`9f15cbb19`)

The second independent review at `8cf2c7b9c` re-drove all five ACs and both remedies on a real
engine under both spellings of the redirect and both renderings, regenerated the schema/help
byte-diff (774 / 0 / 4, `argv` byte-identical), and returned 0 P1 / 1 P2 / 6 P3, all adopted by
the orchestrator:

1. **P2** — the round-1 test change aliased the helper to `resolve_host_data_path`, so the three
   dispatch assertions compared gda's request against the function that produced it: a resolver
   returning `None` (the whole #840 mechanism disabled) passed the unit tier. The resolver now
   has its own oracle, `test_the_host_data_path_resolver_reads_the_host_environment`, pinning
   the engine's per-platform rule over a monkeypatched environment; the dispatch tests keep
   proving the plumbing.
2. **P3** — the round-1 message change had no assertion of its own (dropping
   `(or $GDA_USER_DATA_ROOT)` or reverting the remedy survived); both halves are asserted in the
   unit and the live test.
3. **P3** — clause three of the engine-side check (redirect in play, roots differ, host has no
   templates either → plain shape, no host root) had no test at any tier; a live case points
   gda's own environment at an empty home under an isolated root and asserts
   `templates_root_host` is null.
4. **P3** — `export get`'s human line said `hidden by --user-data-root`; it now names the
   redirect with both spellings.
5. **P3** — the registry test's comment names why producers six to eight pin the empty-evidence
   rule in dedicated tests; the ADR-0004 addendum says where the two directories come from (the
   `export get` op's reply) instead of "reports on no run"; CONTEXT.md's `User-data placement`
   entry states that export-template discovery follows the placement (#840).

On this head: export/cli/meta/repo unit 823 passed, ruff/format/pyright clean, export e2e 18 passed
+ the pre-existing Linux-only skip (macOS, Godot 4.6.3, lock-serialized). Commit authorship on the
branch was rewritten to the repository owner's identity (`9e2deb49d` → `9f15cbb19`, trees identical).

## Review round 3 (`ee5de4027`)

The third review at `9f15cbb19` returned CHANGES REQUIRED — 1 P2, adopted:

1. **P2** — `_hidden_host_templates_root` compared the two roots and looked at the host's contents
   but never at the root it reported on, so a redirected root that already held this version read
   as `templates_installed: true` AND `templates_root_host: <host>`, and the human line said
   "installed" and "hidden by" together (reviewer-reproduced on Godot 4.6.3; the renderer treats
   any non-null host root as proof of hiding, which is the field's contract). The op now hands the
   already-computed `templates_installed` to the check, which answers null when nothing is hidden:
   the published predicate is host evidence iff the checked root lacks the version and the host
   has it. New real-engine case
   `test_export_get_under_a_redirect_whose_root_holds_the_templates_names_no_host_root` populates
   both roots and pins installed true / host null / no "hidden by" line; it fails on the previous
   head.

On `ee5de4027`: export unit + e2e 88 passed + the pre-existing Linux-only skip (macOS, Godot 4.6.3,
lock-serialized), ruff/format and `git diff --check` clean.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -141,7 +141,10 @@ attributable to the environment instead of read as a game regression (#850). Onl
 there: a run that ends in an `Error envelope` — `--strict`'s `script_failed`, a
 `launch_timeout` — keeps its pre-#850 shape, since disclosing a fact on a failure
 means entering `Failure evidence`'s producer set, which is a separate ADR-0004
-decision. Headless only: a live `Engine session`'s log is daemon-owned (ADR-0022).
+decision. The engine's export-template lookup follows the same placement, so a
+redirected export can miss templates the host holds — `export run` says so and
+`export get` reports both roots (#840). Headless only: a live `Engine session`'s log
+is daemon-owned (ADR-0022).
 _Avoid_: log redirect, user dir, sandbox
 
 **Raw run**:
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -270,6 +270,21 @@ status: accepted
 > all, rather than a partially invented triple — which is the omitted-never-null
 > rule applied to a producer, not an exception to it.
 >
+> **Addendum (2026-09-06, #840): an eighth producer, and the count above stands as
+> the record of what this amendment decided.** `export_templates_missing_failure`
+> joins the set with `templates_root_checked` and `templates_root_host` — the
+> export-templates directory a run checked, and the host one holding the templates a
+> `--user-data-root` redirect hid from it. Godot reads the templates from the data
+> directory that option relocates, so the ONE code has two shapes needing opposite
+> remedies: drop the redirect, or install the templates. Both paths pass the
+> criterion — each is already computed on the failure path (the engine-side check
+> composes them), neither is recoverable from the envelope without reading the
+> message, and which shape it is decides what the caller does next. Like the two
+> `target_outside_project` refusals it reports on no run: it is the pre-export
+> preflight's verdict. The omitted-never-null rule holds at the producer too — a
+> reply naming no directory yields no `evidence` key rather than an empty object.
+> The asserted set in `tests/cli/test_error_registry.py` is now eight.
+>
 > **Answering ADR-0002's open pointer (#717): the ceiling's PROVENANCE is declined, on
 > the criterion.** That note left the question here — whether a `launch_timeout`'s
 > ceiling was the caller's `--timeout` or one of gda's own fixed bounds "belongs on
--- a/docs/adr/0004-schema-flag-self-description.md
+++ b/docs/adr/0004-schema-flag-self-description.md
@@ -279,9 +279,10 @@ status: accepted
 > remedies: drop the redirect, or install the templates. Both paths pass the
 > criterion — each is already computed on the failure path (the engine-side check
 > composes them), neither is recoverable from the envelope without reading the
-> message, and which shape it is decides what the caller does next. Like the two
-> `target_outside_project` refusals it reports on no run: it is the pre-export
-> preflight's verdict. The omitted-never-null rule holds at the producer too — a
+> message, and which shape it is decides what the caller does next. Its evidence is
+> not a run's residue — the two directories come from the `export get` op's reply —
+> and it is the pre-export preflight's verdict, as the two `target_outside_project`
+> refusals are pre-launch ones. The omitted-never-null rule holds at the producer too — a
 > reply naming no directory yields no `evidence` key rather than an empty object.
 > The asserted set in `tests/cli/test_error_registry.py` is now eight.
 >
--- a/docs/command-catalog.md
+++ b/docs/command-catalog.md
@@ -1196,6 +1196,18 @@ before the native export and reported in `created_dirs`, outermost to innermost;
 an uncreatable parent is reported as `export_output_parent_failed` before Godot
 runs.
 
+Export-template discovery follows the user-data placement (#840). Godot reads the
+templates from its data directory, and `--user-data-root` relocates exactly that,
+so a release/debug run under an isolated root finds none even where the host has
+them installed. `gda export get` therefore reports `templates_root` — the
+export-templates directory checked, which holds the `templates_version` directory
+— and `templates_root_host`, the host's directory when the redirect hid installed
+templates there (null otherwise). The `export_templates_missing` failure names the
+same two directories in its message, gives the two remedies (run without the
+redirect, or `--mode pack`, which needs no templates), and carries them typed on
+`evidence` as `templates_root_checked` / `templates_root_host` — the second key
+present only in the hidden case, which is how the two shapes are told apart.
+
 ### Asset-file groups (create/edit files; headless)
 
 These create or edit resource files (`.gdshader`, `.tres`) and so are headless.
--- a/src/gda/commands/export.py
+++ b/src/gda/commands/export.py
@@ -25,7 +25,7 @@ from pathlib import Path
 from typing import Annotated, Optional
 
 import typer
-from pydantic import AfterValidator, BaseModel, Field
+from pydantic import AfterValidator, BaseModel, Field, model_validator
 
 from gda import dispatch
 from gda.binary import resolve_godot_binary
@@ -50,7 +50,7 @@ from gda.headless import (
     params_json_option,
     project_option,
 )
-from gda.runner import RunResult
+from gda.runner import RunResult, engine_data_path
 
 
 def normalize_export_output_path(path: str) -> str:
@@ -118,17 +118,60 @@ class ExportListResult(BaseModel):
     presets: list[ListedPreset]
 
 
+def host_data_path() -> str | None:
+    """The host's Godot data directory, resolved over gda's OWN environment (#840).
+
+    The default of :attr:`ExportGetParams.host_data_path`, and the whole mechanism
+    behind #840's disclosure. ``--user-data-root`` redirects the CHILD engine's
+    data directory, never gda's own environment, so this stays the directory an
+    unredirected run would use — exactly the one the engine cannot see from inside
+    the redirect, and therefore the one worth passing in.
+
+    ``None`` when the platform's own variable is unset, which is what
+    :func:`gda.runner.engine_data_path` answers rather than fabricating a path; the
+    operation then reports no host directory instead of comparing against a guess.
+    """
+    resolved = engine_data_path()
+    return str(resolved) if resolved is not None else None
+
+
 class ExportGetParams(BaseModel):
-    """The operation params of ``gda export get`` (issue #114).
+    """The operation params of ``gda export get`` (issue #114, #840).
 
     ``preset`` addresses an export preset by its display name (as ``export
     list`` reports it). An unknown name is the ``export_preset_not_found``
     failure. The project is process context (``--project``, ADR-0006).
+
+    ``host_data_path`` is a COMPUTED param, the same shape ``script set``'s
+    ``mode`` has: the model stamps it from the host environment so the engine-side
+    check can name templates a ``--user-data-root`` redirect hid (#840), and a
+    value passed in is ignored. Computed model-side rather than pasted in by the
+    argv body because ADR-0015 makes the model the single source of truth for a
+    request's shape — a ``--params-json`` caller that names only ``preset`` has to
+    reach the operation with the identical params, and it is not a fact a caller
+    is in any position to supply.
     """
 
     preset: str = Field(
         description="The export preset's display name, as 'gda export list' reports it."
     )
+    host_data_path: str | None = Field(
+        default=None,
+        description=(
+            "The host's Godot data directory, so the check can name export "
+            "templates a --user-data-root redirect hides. Resolved model-side "
+            "from gda's own environment; a value passed in is ignored."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def _resolve_host_data_path(self) -> "ExportGetParams":
+        # Stamped on BOTH input channels (ADR-0015): the argv body and
+        # `--params-json` build this same model, so neither can reach the operation
+        # without the host directory and neither can reach it with a caller's guess
+        # at one.
+        self.host_data_path = host_data_path()
+        return self
 
 
 class ExportGetResult(BaseModel):
@@ -142,6 +185,15 @@ class ExportGetResult(BaseModel):
     #121); ``templates_version`` names the version directory that was checked
     (e.g. ``4.6.3.stable``), so the agent knows which templates to install when
     they are missing.
+
+    ``templates_root`` says WHERE it looked (#840) — the export-templates
+    directory that holds the version directory. It is reported because that
+    location is not fixed: Godot reads the templates from its data directory, and
+    ``--user-data-root`` relocates exactly that, so a redirected run reports none
+    installed on a host whose templates are correctly installed.
+    ``templates_root_host`` names the host's directory in that case and ONLY in
+    that case, so the two situations — hidden by a redirect, versus genuinely not
+    installed anywhere — are told apart before an export is ever attempted.
     """
 
     index: int = Field(
@@ -167,6 +219,18 @@ class ExportGetResult(BaseModel):
         description=(
             "The export-templates version directory checked for installation "
             "(e.g. 4.6.3.stable), matching the running engine version."
+        )
+    )
+    templates_root: str = Field(
+        description=(
+            "The export-templates directory that was checked; the "
+            "templates_version directory is looked up inside it."
+        )
+    )
+    templates_root_host: str | None = Field(
+        description=(
+            "The host's export-templates directory, when a --user-data-root "
+            "redirect hid templates that ARE installed there; null otherwise."
         )
     )
 
@@ -268,15 +332,26 @@ def render_export_list(listed: "ExportListResult") -> str:
 
 
 def render_export_get(got: "ExportGetResult") -> str:
-    """Render one preset's details plus its export-template readiness."""
+    """Render one preset's details plus its export-template readiness.
+
+    The template line names the directory that was checked, and — when a
+    ``--user-data-root`` redirect hid installed templates — a second line names
+    where they really are (#840), so the human channel says exactly what the JSON
+    one does.
+    """
     runnable = " [runnable]" if got.runnable else ""
     header = f"{got.name} ({got.platform}){runnable}"
-    templates = (
-        f"templates installed ({got.templates_version})"
-        if got.templates_installed
-        else f"templates missing ({got.templates_version})"
-    )
-    return "\n".join([header, f"  export_path: {got.export_path}", f"  {templates}"])
+    state = "installed" if got.templates_installed else "missing"
+    lines = [
+        header,
+        f"  export_path: {got.export_path}",
+        f"  templates {state} ({got.templates_version}) in {got.templates_root}",
+    ]
+    if got.templates_root_host:
+        lines.append(
+            f"  hidden by --user-data-root; host templates: {got.templates_root_host}"
+        )
+    return "\n".join(lines)
 
 
 def render_export_run(ran: "ExportRunResult") -> str:
@@ -536,7 +611,16 @@ def run_export_operation(
     if not output_path:
         return export_path_unset_failure(got.name)
     if mode is not ExportRunMode.PACK and not got.templates_installed:
-        return export_templates_missing_failure(got.name, got.templates_version)
+        # Both directories ride the failure (#840): the one the engine checked, and
+        # — when a --user-data-root redirect hid installed templates — the host's,
+        # which is what turns "install the templates" into "you already have them,
+        # this run cannot see them".
+        return export_templates_missing_failure(
+            got.name,
+            got.templates_version,
+            got.templates_root,
+            got.templates_root_host,
+        )
     created_dirs = _ensure_output_parent_dirs(output_path)
     if isinstance(created_dirs, Failure):
         return created_dirs
@@ -740,6 +824,12 @@ def run_export(
     preset ``export_path`` values keep Godot's project-relative convention. The
     reported ``output_path`` is the resolved artifact path, and missing output
     parent directories are created and reported in ``created_dirs`` (#402/#403).
+
+    Export-template discovery follows ``--user-data-root``: Godot reads the export
+    templates from the data directory that option relocates, so a release/debug run
+    under it finds none installed unless you put templates there. The failure then
+    names both directories and the remedies; ``--mode pack`` needs no export
+    templates at all.
     """
     # Build the params model from the argv options (the single source of truth,
     # ADR-0015): ExportRunParams.output is an ExportOutputPath, so argv and
--- a/src/gda/commands/export.py
+++ b/src/gda/commands/export.py
@@ -762,7 +762,14 @@ def get_preset(
     godot: Optional[str] = godot_option(),
     project: Optional[str] = project_option(),
 ) -> None:
-    """Report one preset's details plus export-template install status."""
+    """Report one preset's details plus export-template install status.
+
+    ``templates_root`` names the export-templates directory that was checked.
+    Godot reads the templates from its data directory and ``--user-data-root``
+    relocates that, so a redirected run can report none installed on a host
+    that has them; ``templates_root_host`` names the host's directory in
+    exactly that case.
+    """
     dispatch_domain(
         EXPORT_GET_COMMAND,
         ExportGetParams(preset=preset),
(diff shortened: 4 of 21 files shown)


qinhaihong-red:
**Review record (orchestrator, 2026-09-07).** Independent adversarial review at `db6564ced` (Artifact → Architecture → Standards / Spec-vs-issue, 10 mutants, schema/help byte-diff regenerated from a clean base tree, both remedies exercised on a real engine) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 5 P3**. Each finding re-checked against the head before disposition. Fix head: `8cf2c7b9c`.

|---|---|---|
| **P2** the remedy names only the flag half of the switch; `$GDA_USER_DATA_ROOT` is the same redirect and the exported-variable case (PITFALLS) is the common trigger | Confirmed at `errors.py:760-763`; reproduced with the variable exported for one run | **Adopted.** Message: `--user-data-root (or $GDA_USER_DATA_ROOT) moved the lookup`; remedy: `Run the export without the user-data redirect`. Existing assertions hold; docs quote neither old sentence. |
| P3 speculative defaults on `export_templates_missing_failure` | Confirmed; the one caller passes both, `templates_root` is a required `str` on the result | **Adopted.** Defaults dropped; the omitted-never-null guard stays. |
| P3 mutant M6 survived: a pasted `host_data_path` honoured by the validator keeps 2513 tests green | Confirmed | **Adopted.** The params-json parity test now pastes a value and asserts the resolved host path still reaches the op. |
| P3 stale "computed by the production default" comment | Confirmed | **Adopted.** |
| P3 tests re-derive the host path beside a sibling that imports the resolver | Confirmed | **Adopted.** One resolver, imported. |
| P3 (pre-existing, adjacent) "grows a sixth field" comment | Confirmed | **Adopted.** |

On `8cf2c7b9c`: `tests/export` + `tests/cli` 480 passed, ruff/format/pyright clean, export e2e 17 passed + 1 pre-existing Linux-only skip (macOS, Godot 4.6.3). Reviewer-verified and unchanged by the fix: 774 leaves added / 0 removed / 4 changed and the enumeration complete; layout rule engine-side only; host path over the unredirected environment even with `GDA_USER_DATA_ROOT` set; plain shape omits `templates_root_host` on both renderings; renderer exhaustiveness guard red-on-growth; ADR-0002 + `error_codes.py` byte-identical; ADR-0004 addendum appended with the #687 count preserved; both remedies work on a real engine. Noted for CI planning (no action here): the engine-side half of the mechanism is guarded by the Godot e2e tier only, which the wave's e2e dispatch runs. Ready for merge in the wave order (after #882).

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-07).** Second independent review at `8cf2c7b9c` (all five ACs + both remedies re-driven live under both redirect spellings and both renderings; schema/help byte-diff regenerated; 12 mutants) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 6 P3**. Fix head: `9f15cbb19` (orchestrator-direct; the P2 was the orchestrator's own round-1 change).

|---|---|---|
| **P2 (F2)** aliasing the test helper to `resolve_host_data_path` made three assertions tautological; mutant "resolver returns `None`" survived the unit tier (720 passed) while the round-1 form killed it | Confirmed | **Adopted.** Independent oracle test for the resolver (engine per-platform rule over a monkeypatched environment); dispatch tests keep the alias for the plumbing claim. |
| P3 (F1) the round-1 message fix unpinned (mutants M5/M6 survived) | Confirmed | **Adopted.** `$GDA_USER_DATA_ROOT` and `without the user-data redirect` asserted in the unit and the live test. |
| P3 (F4) clause three of `_hidden_host_templates_root` untested (mutant E2 survived e2e) | Confirmed | **Adopted.** Live case: empty home for gda's own process + isolated root → `templates_installed false`, `templates_root` under the isolated root, `templates_root_host null`. Passes on this host. |
| P3 (F5) `render_export_get` names one spelling | Confirmed | **Adopted.** Both spellings; human-output test updated. |
| P3 (F3) registry test comment "five producers" | Confirmed | **Adopted.** Names the split and the dedicated-test pattern. |
| P3 (F6) ADR addendum "reports on no run" | Confirmed | **Adopted.** The directories come from the `export get` op's reply; pre-export preflight verdict. |
| P3 (F7) CONTEXT.md `User-data placement` silent on template discovery | Confirmed | **Adopted.** One clause (#840). |
| Recommendation: a behavioural "pasted value ignored" pin per `_COMPUTED_PROPERTIES` member (`script set`/`shader set` `mode` have none) | Confirmed pre-existing | **Recorded** for a follow-up issue; not this PR. |

On `9f15cbb19`: export/cli/meta/repo unit 823, ruff/format/pyright clean, export e2e 18 + 1 pre-existing Linux-only skip; commit authorship rewritten to the repository owner's identity (`9e2deb49d` → `9f15cbb19`, trees identical). **READY** — merges third in the wave (after #882).

qinhaihong-red:
**Changes required — 1 P2** at exact head `9f15cbb1934f469d101771bb2f3bf501fce39ad1` (base `cfcb8658e67df418a69694840a37a22a9cd3cbe0`).

## Artifact Review

**[P2] Keep `templates_root_host` null when the redirected root already has this template version.** The issue, command catalog, bundled skill, model, and schema description all define this field as non-null only when a redirect actually hides templates. `src/gda/ops/operations.gd:3443-3460`, however, checks only that the roots differ and that the host contains `version_dir`; it does not check the current root, even though `templates_installed` is computed separately at lines 3341-3348.

With real Godot 4.6.3 and `4.6.3.stable` present in both roots, this head returned `templates_installed: true` together with a non-null `templates_root_host`. Human `export get` consequently printed both “templates installed” and “hidden by the user-data redirect” because `src/gda/commands/export.py:334-355` treats every non-null host root as proof of hiding. This can make agents classify a healthy redirected root as hiding templates.

I reviewed issue #840, the complete PR body, and all changed Markdown against `CONTEXT.md`, ADR-0004, project terminology, and ASD-STE100. Apart from this factual contradiction, I found no substantive terminology or prose problem.

## Architecture

The topology is otherwise sound: the Python export slice owns params/results/rendering/recipes, while the Godot operation owns engine template layout and probing. The defect is a completeness failure in that engine-side owner: it does not preserve the published predicate `host evidence iff current version absent and host version present`. Reuse or pass the already-computed `templates_installed` and return null when the checked root already contains the version.

## Standards

The same P2 violates the model-driven result contract in ADR-0004: the semantic field name says “hidden-only”, while the renderer treats a weaker predicate as proof. No separate actionable Fowler smell was found.

## Spec

The other #840 acceptance criteria pass. Add a real-engine matrix case with both roots populated that asserts `templates_installed == true`, `templates_root_host == null`, and no human `hidden by` line. Current coverage includes unredirected, current-missing/host-present, and current-missing/host-missing states, but not both-present.

Validation: focused unit and Godot E2E tests passed, as did Ruff, format, targeted Pyright, and `git diff --check`. The counterexample was reproduced locally on macOS with real Godot 4.6.3. Hosted Godot E2E was skipped; Linux/Windows runtime was not re-run. The PR head remained unchanged at final verification.

qinhaihong-red:
**Review round 3 handled** — the P2 adopted at `ee5de4027`.

|---|---|---|
| **P2** `templates_root_host` non-null when the checked root already holds the version | Confirmed: `_hidden_host_templates_root` never looked at the root it reported on | **Adopted.** The op passes the already-computed `templates_installed`; the check answers null when nothing is hidden. New real-engine case populates both roots and pins installed true / host null / no "hidden by" line — it fails on the previous head. |

On `ee5de4027`: export unit + e2e 88 passed + the pre-existing Linux-only skip (macOS, Godot 4.6.3, lock-serialized), ruff/format and `git diff --check` clean. PR body carries the round-3 record.

Issue #845, opened 2026-09-05
script: a case-mismatched script path is a typed refusal, not valid:true


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-062** (2026-08-26, gda 0.10.0, macOS). `gda script validate res://Content/combat/combat_session.gd` (tracked spelling `res://content/...`) returned `valid: true`, `error_string: null`, and empty `diagnostics`, while Godot warned that the case does not match the stored file and that the file will not open on a case-sensitive platform. A portability gate on macOS therefore accepts a path that fails on Linux and on a case-sensitive export host.

Verified against gda 0.13.0 (`266c8009f`) and Godot 4.6.3: the engine's message is a `WARN_PRINT` from `drivers/unix/file_access_unix.cpp` (`Case mismatch opening requested file '%s', stored as '%s' in the filesystem. This file will not open when exported to other case-sensitive platforms.`). gda's script-error classifier (`gda.script_errors`) skips every warning by contract, so the record is never classified (the runtime-log reader `gda diag errors` uses does surface warnings, but that is a different channel), and gda's own path normalization compares nothing against the on-disk spelling.

## What to build

Decide the mismatch on gda's side, deterministically and with the same code on every platform, at ADR-0006's path authority in `gda.project`, reached through the one containment gate (#763, folded to one call per site by #802). The gate has three call sites today — `script validate`, `script run`, and `resource import`; every other command group hands its `res://` target straight to the engine and never consults `gda.project` for it (scope note of 2026-09-05 below):

- Resolve the requested path against the project and compare each component with the directory entry's actual spelling. A mismatch is a typed refusal `path_case_mismatch` — category `operation`, exit bucket `EXIT_OPERATION` (the same as `path_not_found`), source `classifier`, since gda decides it CLI-side like `target_outside_project` — whose message names the requested and the stored `res://` spellings. The two spellings also ride `evidence` as declared fields (`requested_path`, `stored_path`), admitted through ADR-0004's #687 gate (`FailureEvidence` is `extra="forbid"`). No `hint`: the envelope's `hint` is contractually one corrected invocation from the curated near-miss table (ADR-0004's #670 note; CONTEXT.md `Near-miss hint`), and a case mismatch is not a near miss — the corrected spelling is the typed `stored_path`, the same shape #840 chose for its two directories.
[...]

Issue #842, opened 2026-09-05
project add-input-action: joypad button and axis events


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-068** (2026-08-26, gda 0.11.0). `gda project add-input-action pause_menu --key Escape` cannot author the required Escape-plus-controller-Start action, nor the A/B accept/cancel and D-pad additions a controller-accessible game needs: the command requires `--key` and exposes no joypad button, axis, or device option. The production had to hand-write `InputEventJoypadButton` records into `project.godot` — the reverse-engineered serialization #380 was created to remove.

#380 recorded mouse, joypad, and axis events as out of scope for its first slice, with the requirement that the op shape extend to them later without breaking. This is that extension; the trigger has arrived. The whole-file reserialization the same entry observed is #843.

## What to build

Extend `add-input-action` so one call declares keyboard and controller bindings together, persisted through `ProjectSettings` exactly as today:

- `--joy-button <name|index>` (repeatable): an `InputEventJoypadButton` bound to a `JoyButton` name (e.g. `A`, `B`, `Start`, `DPadLeft`) or its integer value.
- `--joy-axis <axis>[:<sign>]` (repeatable): an `InputEventJoypadMotion` bound to a `JoyAxis` name or index with the axis sign (`+` / `-`), so one left-stick direction is one binding.
- `--device <n>`, applied to the joypad events of this call; default `-1` (`InputMap.ALL_DEVICES`, matched by every joypad). gda must set it explicitly: a script-constructed joypad event starts at device `0` (`InputEvent::device` defaults to 0 in the engine), which matches only the first joypad — the same reason the existing key path sets `device = -1` itself.
- `--key` becomes optional; at least one binding of any kind is required.
- The written value keeps Godot's own serialization for the new event objects, as #380 required for keys; an unknown button or axis name is a structured error naming the accepted set.

Coordinate with #843 (bounded project writes): both change `add-input-action` and neither blocks the other; whichever lands second extends the other's regression (the byte-equivalence check must cover the joypad event types, and the restored-line check must cover a call with joypad events).

## Acceptance criteria

- [ ] `add-input-action jump --key Space --joy-button A` writes an action with one `InputEventKey` and one `InputEventJoypadButton` that the engine and the editor load as a normal action.
[...]

Issue #829, opened 2026-09-04
daemon start: refuse a project with no main scene before the engine pops a native alert


qinhaihong-red:
## What to build

Add a conservative file precheck to `gda daemon start` and the daemon's lazy Engine session launch. With no `--scene` selector, refuse a project when the files establish either of these conditions:

- `application/run/main_scene` is empty or absent: `live_main_scene_undefined` (LIVE, exit 6). Name both remedies: set the main scene, or pass `--scene <res://path|uid://...>`.
- The main scene is a `uid://` and the configured project data directory has no `uid_cache.bin`: `live_main_scene_unresolved` (LIVE, exit 6). Name the import remedy (`gda resource import <any existing res:// asset>` once, or open the project in the editor) and the explicit `--scene res://<scene>.tscn` alternative.

The CLI refusal must precede daemon launch and harness installation; the daemon-side refusal must precede Engine session launch. The existing engine version probe is allowed. A determinate main-scene refusal precedes the `--windowed` display check at both boundaries, so a simultaneous display failure does not mask the scene code or its remedies. Use one predicate owned by `gda.project`, checked by the CLI and again at the daemon's authoritative launch boundary because the project can change after `daemon start`. Register the codes in `gda.error_codes`, with their documented rows in ADR-0002 and the category of `live_scene_not_found`.

Godot 4.6.3 calls `OS::alert()` when its GAME startup path has no main scene or cannot resolve a main-scene UID without a cache. On macOS this is a native modal `NSAlert` even with `--headless`. The readiness deadline eventually tears down the session, but the dialog can remain until then. This was observed on 2026-09-04 during #827's boot-latency probes; #830 records the operational pitfall.

## Accepted scope and limitations

The precheck must not reject a project the engine can run. It does not reproduce Godot's feature selection or settings-overlay parser. The following cases defer to the existing engine launch and readiness deadline:

- A feature-tagged `run/main_scene.<feature>` declaration, an existing `override.cfg`, a nonempty `application/config/project_settings_override` path, or a feature override of that path. These can change the effective main scene.
- An escaped application key, whose decoded name the minimal reader does not determine.
- For the UID-cache verdict only: a feature override or an unrecognized boolean value for `application/config/use_hidden_project_data_directory`. The active cache directory is then uncertain; an independently known empty main scene can still be refused.
- An unreadable or undecodable project file is no scene verdict; the existing file-access failure path remains responsible for it.

[...]

PR #813, opened 2026-09-03
docs(gda): trim the README to what its reader acts on — keep it simple

qinhaihong-red:
## Why

The README is the human-facing front door. Over the dogfooding waves (2026-08-17 → 2026-09-03) each feature PR "synced its contract" into whichever README sentence mentioned its flag or command. The command reference grew from about 180 to about 1,700 words of prose that restated what `--help`, the Skill, `docs/command-catalog.md` and the ADRs already say — a fifth hand-kept copy of the contract, replicated into three translations.

This PR trims the README back to its reader. One rule, applied throughout:

- **User-facing.** A first-time reader needs what a command does and the one thing that will surprise them. Not error-code inventories, result-field lists, or engine mechanics.
- **Keep it simple.** A table row is one line per command plus at most one clause of gotcha. After a table, at most a short note for a contrast the rows cannot make.
- **Cut what already has an authoritative home; do not compress it.** Per-command contract detail lives in `--help` (which also feeds `--schema` and the MCP tool descriptions), the Skill, the command catalog and the ADRs. The README links out instead of restating.

The goal is lower cognitive load for the person reading the README, not fewer facts in the project. Every removed sentence still exists where it belongs. What a user must *know* — a compatibility change, a setup that stops working — stays, in one sentence.

## What changed

Six commits, same rule (the last two answer the review below):

1. **Command reference** (354a1d4c2). The prose blocks after the `scene` (618 words), `script` (571) and `game` (172) tables are deleted. The one fact a reader needs moves into the command's own row. A note of at most 30 words stays after `scene` and `game` for the contrast the rows cannot make. The "Fails loudly" bullet drops its evidence-key walkthrough.
2. **Remaining notes and the daemon table** (07a3dd7d3). The `Control` offset note folds into the `node set` row, the stale-mouse note becomes one line after `input`, and the four long `daemon` cells say what the command does and the one reason to run it.
3. **Outside the tables** (b88de343a). Configuration keeps one resolution table instead of a list plus a table saying the same thing. The project-code-execution disclosure keeps its six points at one sentence each. Meta and global-flag rows lose payload shapes and field enumerations. The "Reading a 124" walkthrough goes: the `launch_timeout` message already says it, and the Skill keeps its copy. The Skill flag paragraph that repeated the code comments goes.
[...]

PR #898, opened 2026-09-07
fix(gda): bound a project write to the request and report the file it reserialized (#843)

### LINKED ISSUE #843: project writes: ProjectSettings.save() drops explicit-default settings and reorders the file — bound and report the mutation
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-068** (2026-08-26, gda 0.11.0). A keyboard-only `add-input-action` reserialized the entire `project.godot`: it added boilerplate and `config/features`, reordered sections, and dropped the explicit `file_logging/enable_file_logging=false` and `textures/default_filters/use_nearest_mipmap_filter=false` lines. None of that was reported; the agent believed it had made one bounded InputMap edit.

Verified against the engine and gda 0.13.0 (`266c8009f`): every writer in the `project` command group (`project set`, `add-autoload`, `remove-autoload`, `add-input-action`, `remove-input-action`) persists through `ProjectSettings.save()`. (`daemon install` / `uninstall` edit the `[autoload]` section with a line-level text editor, and `export run` snapshots and restores the file, ADR-0028 — those paths are not in scope here.) The engine's writer (`ProjectSettings::save_custom`, `core/config/project_settings.cpp`, Godot 4.6.3) skips any setting whose value equals its initial value (`if (v->variant == v->initial) continue;`) and sets or rewrites `application/config/features` (adds the list when absent, and on an existing list appends the rendering method, adds or removes `C#`, and trims to supported features). An explicit default-equal line is therefore deleted on every gda write, and the file's order becomes the engine's.

## What to build

Bound the write to what the caller asked for, and disclose the rest:

- After the engine save, compare the pre- and post-write file with a section/key reader of Godot's `ConfigFile` text. gda has two partial readers today — the main-scene reader in `gda.project` and the `[autoload]` line editor the harness installer uses — so generalize one into a single key/value reader rather than adding a third. Report settings whose explicit line vanished (default-equal values the engine dropped), settings the engine added or rewrote (`config/features`), and whether the section order changed.
- Restore the dropped explicit lines verbatim into their original section. They are the caller's declarations (tooling and humans read the file), and restoring bytes the caller wrote is not gda-authored serialization — ADR-0033's objection to hand-built Godot literals does not apply.
- The addressed key itself is covered from the REQUEST, not from the pre-write file: when the caller writes a value equal to the engine default (`project set debug/file_logging/enable_file_logging false` on a project that lacks the line), the engine drops it and the pre-write file has nothing to restore, so gda writes the line from the request's key and coerced value — the same value the result already echoes. (Mechanism amended 2026-09-07 with PR #898: gda writes no Godot literal — the operation moves the engine's initial value aside with `set_initial_value` so the ENGINE writes the line from the coerced value, ADR-0033; the observable outcome is unchanged and the line is still reported under `restored_settings`. See the dated comment.)
- Report the residual mutation on every project-write result: `added_settings`, `rewritten_settings`, `restored_settings`, and `sections_reordered`. Human rendering shows one line per NON-EMPTY category (amended 2026-09-07 with PR #898: a clean write renders only its own line, so the common case stays one line).
- Decision recorded here: gda does not preserve the file's section order — the engine owns the layout — but it says when the order changed.

## Acceptance criteria

- [ ] A `project.godot` carrying an explicit default-equal setting (e.g. `debug/file_logging/enable_file_logging=false`) keeps that line after `add-input-action`, `add-autoload`, and `project set` on an unrelated key.
- [ ] `project set` of a key to its engine default on a file that lacks the line leaves the line present afterwards, and the result reports it under `restored_settings`.
- [ ] Each project-write result reports the settings the engine added or rewrote and those gda restored; `config/features` appears under `added_settings` on a file that lacked it and under `rewritten_settings` on a file whose list the engine changed.
- [ ] The addressed setting itself is written exactly as before (#380's byte-equivalence for InputMap entries holds).
- [ ] A regression covers all five writers on one fixture; the engine still loads the restored file (a headless `project get` of the restored key).
- [ ] The help of the five commands states that the engine reserializes the file and what gda restores and reports.

## Blocked by

None - can start immediately. Coordinate with #842 (joypad events on `add-input-action`): whichever lands second extends the other's regression.

qinhaihong-red:
## Summary

`ProjectSettings.save()` does not edit `project.godot` — it **reserializes** the whole file from
the engine's merged settings (`ProjectSettings::save_custom`, Godot 4.6.3). So every writer in the
`project` group changed lines the caller never named, and reported none of it: an explicit
declaration whose value equals the engine's initial value was **deleted**
(`if (v->variant == v->initial) continue;`), `application/config/features` was **added or
rewritten**, and the sections came back in the **engine's own order**. That is GDA-DF-068: a
keyboard-only `add-input-action` silently dropped `debug/file_logging/enable_file_logging=false`
and reordered the file while the agent believed it had made one bounded InputMap edit.

End to end, after this PR:

- the five writers (`set`, `add-autoload`, `remove-autoload`, `add-input-action`,
  `remove-input-action`) run through the **recipe channel** (ADR-0023) — the one descriptor-driven
  hook argv and `--params-json` share — so the CLI holds the file as it stood **before** the engine
  ran, which no engine started afterwards can recover;
- after a successful write gda **restores each dropped declaration verbatim** into the section it
  was written in, and **reports the residual mutation** on the result: `added_settings`,
  `rewritten_settings`, `restored_settings`, `sections_reordered` (human rendering: one line per
  non-empty category, so a clean write stays a one-liner);
- the **layout is not restored** — the engine owns it, so gda says the order changed instead of
  fighting it (the decision recorded on the issue);
- the restore is an **optimistic, atomic replace** of the engine's output (ADR-0018 Decision 4,
  the guarantee `operations.gd` already gives its own read-modify-write ops): staged in a sibling
  file, re-checked against the target's `st_mtime_ns` and `st_size`, committed with one
  `os.replace`. A file that moved in that window is left to its writer and reported as
  `file_changed_externally`; a write that fails outright stays `save_failed`;
- the **addressed** setting is in no category: its new value, appearance or removal IS the request.
  The one line that cannot come from the pre-write file — a `set` of a setting TO the engine
  default on a file that never declared it — is kept by `_op_project_set` moving that default aside
  (`ProjectSettings.set_initial_value`), so the **engine** writes the line in its own
  serialization and names it in `restored_settings`. gda hand-builds no Godot literal (ADR-0033).

**Prefactoring commit** (`065431bc0`, behavior-preserving): `gda.project_file` becomes the ONE
reader of Godot's `ConfigFile` text. It replaces the two partial readers the issue names — the
#829 main-scene reader in `gda.project` (`_strip_comment`/`_unquote`/`_section_name`) and
`harness/install.py`'s line primitives (`_split_config`/`_line_ending`/`_is_section_header`/
`_section_of`) — and adds what the third caller needs: multi-line values (an `input/<action>`
Dictionary is ONE entry), raw per-entry lines (so a restore puts back the exact bytes), and
decoding of the engine-encoded key spelling (`property_name_encode`), with a key it cannot decode
excluded from every comparison rather than guessed at — a key read as two keys is how a "restore"
would write a duplicate line. The installer keeps its own `[autoload]` **edit**, which is what
preserves its byte-identity guarantee; only the reading primitives moved.

**Reading it the way the engine reads it** (round 3). ONE lexical scan carries the quote, escape
and bracket state across lines and records each section header's line index, and the restore
consumes those recordings rather than rescanning the raw text — so a `[debug]` line inside a
multi-line String value is not a section. `config_key` mirrors the engine on both key spellings:
a bare key drops every character of code 32 or less (`parse_tag_assign_eof` accumulates only
`c > 32`), and a quoted key is decoded as `VariantParser::get_token` decodes a string (four hex
digits after `\u`, SIX after `\U`, every other escape standing for the character it precedes).
The harness installer reduces every raw line through the same `config_line`, so a commented
`[autoload] ; note` header is the one recognizer's autoload section too.

**No `HARNESS_VERSION` bump.** `src/gda/harness/gda_harness.gd` is untouched — `install.py` is the
INSTALLER, not the harness, and `scripts/harness_version_guard.py` keys on the harness body.

Closes #843

## Surface diff

Byte-diffed against the main checkout at `cfcb8658e` (`gda schema`, and each affected `--help` at
`COLUMNS=88`). Every diff below is intended; nothing else in the surface moved.

| Surface | Diff |
| --- | --- |
| `gda schema` | 5 commands changed, no command added/removed: `project set`, `add-autoload`, `remove-autoload`, `add-input-action`, `remove-input-action`, each in `description` (the shared help paragraph) and `output` (the four new properties `added_settings`, `rewritten_settings`, `restored_settings`, `sections_reordered`). They carry defaults, so they are schema-optional and `required` is unchanged — the same shape as `daemon install`'s `created_paths`/`created_sections` receipt. |
| `--help` (×5) | +5 lines each (the shared 4-line paragraph and its blank), 0 removed. No option, argument or summary line changed. |
| `docs/command-catalog.md` | New "**What a write does to `project.godot`**" paragraph in the `project` section: the engine's three reserialization behaviours, what gda restores/reports, the addressed-setting exception, the FIVE out-of-scope residuals (comments, key order inside a section, line endings, a byte-order mark, an undecodable key spelling — the last now stated against the engine's own tokenizer rules), and the two failure codes a restore can end in (`save_failed`, `file_changed_externally`) with the atomic-replace guarantee behind them. |
| `src/gda/skill/SKILL.md` | The `project` row gains the write gotcha (the group's command list is unchanged, so `tests/meta/test_skill_surface_sync.py` still passes). |
| `README.md` | Two lines after the `project` table (one gotcha, per the README scope rule). Translations re-translated and re-stamped with `scripts/update_readme_i18n.py`; `tests/repo/test_readme_i18n_sync.py` green. |
| `CONTEXT.md` | None. No term is falsified: the report is a per-group result contract, not shared language, and `Project-code execution surface` is unchanged (the new engine-side call is `set_initial_value`, which runs no project code; the CLI restore is a file write). |

## Red-proof evidence

Base `cfcb8658e` exported with `git archive` into a scratch tree, the two new test files copied in,
nothing else changed. Unit: **13 failed, 1 passed**; e2e: **6 failed**.

```
FAILED tests/project/test_project_bounded_write.py::test_a_dropped_explicit_default_line_is_restored_verbatim
FAILED tests/project/test_project_bounded_write.py::test_the_engine_added_setting_is_reported
FAILED tests/project/test_project_bounded_write.py::test_a_rewritten_setting_is_reported
FAILED tests/project/test_project_bounded_write.py::test_a_reordered_file_is_reported_but_not_reordered_back
FAILED tests/project/test_project_bounded_write.py::test_an_unchanged_section_order_reports_no_reorder
FAILED tests/project/test_project_bounded_write.py::test_the_addressed_setting_is_never_reported_as_residual
FAILED tests/project/test_project_bounded_write.py::test_the_engine_reported_restore_of_the_addressed_key_rides_through
FAILED tests/project/test_project_bounded_write.py::test_every_writer_reports_the_mutation
FAILED tests/project/test_project_bounded_write.py::test_human_output_lines_each_non_empty_category
FAILED tests/project/test_project_bounded_write.py::test_a_projectless_writer_reports_no_mutation
FAILED tests/project/test_project_bounded_write.py::test_params_json_takes_the_same_path
FAILED tests/project/test_project_bounded_write.py::test_every_writer_publishes_the_four_report_fields
FAILED tests/project/test_project_bounded_write.py::test_every_writer_help_states_the_reserialization
13 failed, 1 passed in 1.08s
```

The distinct failures — the file the engine wrote is the bug itself, in the second line:

```
E   KeyError: 'restored_settings' / 'added_settings' / 'rewritten_settings' / 'sections_reordered'
E   assert 'file_logging/enable_file_logging=false' in '; Engine configuration file.\n\nconfig_version=5\n\n[application]\n\nconfig/name="fixture"\nconfig/features=PackedStringArray("4.6")\n\n[debug]\n\nfile_logging/enable_file_logging.pc=false\n\n[zsection]\n\nmy/custom=42\n'
E   assert {'added_setti...ns_reordered'} <= {'setting', 'type', 'value'}     # project-set
E   assert 'engine added: application/config/features' in ['set application/config/name (String) = "fixture"']
E   assert 'reserializes' in ' \n Usage: gda project set [OPTIONS] S...e and exit. │\n...'
```

e2e, same base, real engine:

```
FAILED tests/project/test_e2e_project.py::test_every_project_writer_keeps_an_explicit_default_line
FAILED tests/project/test_e2e_project.py::test_project_set_to_the_engine_default_leaves_the_line_present
FAILED tests/project/test_e2e_project.py::test_the_features_list_the_engine_writes_is_reported
FAILED tests/project/test_e2e_project.py::test_a_stale_features_list_is_reported_as_rewritten
FAILED tests/project/test_e2e_project.py::test_a_hand_ordered_file_reports_the_engine_reordering_it
FAILED tests/project/test_e2e_project.py::test_a_bounded_write_renders_its_report_for_a_human
6 failed, 35 deselected in 4.00s
E   KeyError: 'restored_settings'
```

Disclosed: one of the 14 unit tests, `test_a_clean_write_renders_only_its_own_line`, PASSES at the
base. It is a noise guard (a write that changed nothing must render one line only), so it is
vacuously true where no report exists; it is red-proof-free by nature, not by omission.

## Validation

Host: macOS 15 (Darwin 25.6.0), local dev machine, Godot 4.6.3 at the RULES.md path passed per
invocation. Load at the e2e run: `load averages: 3.77 3.54 3.37` (a busy but not saturated host;
no timing-sensitive assertion in this slice). All from the worktree with `uv run --frozen`.

| Gate | Command | Result |
| --- | --- | --- |
| Unit | `uv run --frozen pytest tests -m "not e2e" -q -p no:cacheprovider` | `2528 passed, 672 deselected in 45.75s` (base 2501 + 27 new) |
| Format | `uv run --frozen ruff format --check .` | `512 files already formatted` |
| Godot e2e | `python <lock> uv run --frozen pytest tests/project tests/harness -m e2e -rs -q -p no:cacheprovider` | `119 passed, 300 deselected in 143.89s` (`[e2e-lock] acquired after 121s`) |
| Schema | `gda schema` vs the main checkout at `cfcb8658e` | 5 changed commands, all enumerated above |
| Help | 5 × `gda project <cmd> --help` vs the same base | +5 lines each, 0 removed |

Every Godot-spawning run went through the machine-wide e2e lock, one process at a time across
worktrees. `GDA_USER_DATA_ROOT` was never exported for a run.

## Requirement conformance

| Acceptance criterion | Proof |
| --- | --- |
| A `project.godot` carrying an explicit default-equal setting keeps that line after `add-input-action`, `add-autoload` and `project set` on an unrelated key | e2e `test_every_project_writer_keeps_an_explicit_default_line` — all FIVE writers in sequence on one fixture, asserting both the file text and `restored_settings` after each |
| `project set` of a key to its engine default on a file that lacks the line leaves the line present, reported under `restored_settings` | e2e `test_project_set_to_the_engine_default_leaves_the_line_present` (`config/description` → `""`), unit `test_the_engine_reported_restore_of_the_addressed_key_rides_through` |
| Each write result reports what the engine added or rewrote and what gda restored; `config/features` under `added_settings` on a file that lacked it, under `rewritten_settings` when the engine changed the list | e2e `test_the_features_list_the_engine_writes_is_reported` (added on the first write, neither on the second) and `test_a_stale_features_list_is_reported_as_rewritten`; unit `test_the_engine_added_setting_is_reported`, `test_a_rewritten_setting_is_reported` |
| The addressed setting itself is written exactly as before (#380 byte-equivalence for InputMap entries) | the existing e2e `test_project_add_input_action_persists_var_to_str_form_and_reports_keycodes` and `..._persists_key_and_joypad_events_in_var_to_str_form` still pass — unchanged assertions, now running with the restore in the path; unit `test_the_addressed_setting_is_never_reported_as_residual` pins that the addressed key is never restored or reported |
| A regression covers all five writers on one fixture; the engine still loads the restored file | e2e `test_every_project_writer_keeps_an_explicit_default_line` ends with `project get debug/file_logging/enable_file_logging` → `false` in a fresh engine process, plus `project get application/config/name` to show the requested write landed |
| The help of the five commands states the reserialization and what gda restores and reports | unit `test_every_writer_help_states_the_reserialization`; the help diff above |
| Section order: gda does not preserve it but says it changed | e2e `test_a_hand_ordered_file_reports_the_engine_reordering_it` (asserts the file keeps the ENGINE's order), unit `test_a_reordered_file_is_reported_but_not_reordered_back` + `test_an_unchanged_section_order_reports_no_reorder` |

The joypad binding #842 added rides the five-writer regression's `add-input-action` step
(`--key J --joy-button A --joy-axis LeftX:-`), so the two slices' regressions meet on the final
shape rather than each running its own sequence.

## Disclosed extras / boundary crossings

- **Module placement.** The wave plan said the one reader would live in `gda.project`. It landed in
  a NEW `src/gda/project_file.py` instead: `gda.project` is the project-DIRECTORY and `res://` path
  authority (ADR-0006/#763/#802), and ~470 lines of ConfigFile text handling is a different reason
  to change; a separate module also keeps the diff off the file W3's #845 edits. `gda.project`
  imports it (stdlib-only, no cycle) and its own file shrank by 73 lines.
- **`operations.gd` touched, in one op.** Only `_op_project_set` — the one writer that can address a
  setting whose value equals the engine default. The other four write custom settings (`autoload/…`,
  `input/…`) whose initial is unset, so the engine never drops them.
- **The five commands now pass `help=` in the decorator** rather than carrying a docstring Typer
  reads. That is what lets the shared paragraph have ONE authority (`BOUNDED_WRITE_HELP` +
  `_write_help`) instead of five copies that drift.
- **`_RECIPE_OPERATIONS`** in `tests/cli/test_command_descriptor_registry.py` gains the five
  operations — the descriptor invariant it exists to pin.
- **Two out-of-scope residuals, documented not coded.** The engine also replaces the file's
  COMMENTS with its own header, and reorders keys WITHIN a section. Neither is in the issue's four
  categories; both are stated in the catalog so a reader is not surprised.
- **Test collateral.** Six result round-trip tests in `tests/cli/test_models.py` and two exact-dict
  assertions in `tests/project/test_project_commands.py` / `test_e2e_project.py` asserted the old
  key set; they now assert their own echo plus the (empty) report, with one new test that a
  populated report round-trips.
- **es/ja README notes are agent-written**, like the rest of those files; the zh-CN note is
  native. A native reviewer pass on es/ja remains the standing gap recorded for #813.

## Guards on PR CI vs the Godot e2e tier

PR CI runs **no** Godot e2e (that tier is schedule/dispatch-only). So on PR CI:

- **Runs:** every unit test above — the whole `tests/project/test_project_bounded_write.py`
  (19 tests, including the fake-engine restore, the report, both input channels, the two restore
  failure codes, and the model-field and help invariants), `tests/project/test_project_file.py`
  (26: the scanner, the key normalizer, the escape decoder, the restore's placement and its
  optimistic write), `tests/harness/test_harness_install.py` (48), the descriptor-registry recipe
  invariant, the model round-trips, the SKILL surface sync and the README i18n stamp; plus ruff,
  ruff format and pyright.
- **Only in the e2e tier (dispatched once per wave):** the nine e2e regressions in the `#843`
  block and the harness/install e2e — i.e. everything that proves the REAL engine's drop rule, its
  features rewrite, its section order, its multi-line String serialization, its key spellings, and
  that the restored file still loads. The unit tier fakes the engine's reserialized output, so a
  change in Godot's writer would be caught only there.

---

## Review round 1 (`454c4c15b`)

_(The branch was rebased after rounds 1 and 2 were written; the heads named in this section and
the next are the commits that now carry those trees.)_

All 10 findings addressed to me are fixed in one commit; P3-6 (the issue text) is the
coordinator's. Nothing was declined.

| Finding | What I did |
| --- | --- |
| **P2-1** — `sections_reordered` false while the caller's final order changed | `bound_project_write` now measures the flag on the **final** text (`read_config_text(restored).sections`) rather than on the engine's intermediate, so a section the save emptied away and the restore re-opened is counted where it actually landed. Took the smaller of the two options the brief named: re-inserting the section at its ORIGINAL position would need a predecessor lookup and its own edge cases (no predecessor, a predecessor the save also dropped) for a placement that is still only cosmetic — the file stays in the engine's layout either way, which is the recorded decision, so the flag is what had to become true. `_reordered`'s docstring now says which file it compares. Both pre-existing reorder tests stay green. |
| **P3-2** — the missing-section restore branch untested (mutant M11) | `test_a_section_the_save_emptied_away_is_re_opened_and_the_reorder_reported` in `tests/project/test_project_file.py` is exactly that shape (a section whose only key is default-equal): it asserts the re-opened section, the restored line, the reorder flag, and that the result reads back as a section rather than a stray key. |
| **P3-3** — the shared-set rule unpinned (mutant M8) | `test_a_section_only_one_file_holds_is_not_a_reorder` (a section only the save's output has is not a reorder) and `test_a_reordering_the_save_made_is_reported` (a genuine swap is). Disclosed: both PASS on the reviewed head — they are characterization tests that pin the rationale and kill the mutant, not regression proofs. |
| **P3-1** — `[ autoload ]` half-seen; uninstall left an empty section | Every header RECOGNITION on the installer's edit path is now by NAME (`section_name(...) == _AUTOLOAD_SECTION`), in all three places that compared to the `[autoload]` literal — `_ensure_autoload`, `_remove_autoload`, `_emptied_autoload_span`. Two tests: the uninstall drops the whitespaced header it empties, and an install into a whitespaced section with a sibling entry JOINS it (no duplicate section) and round-trips byte-identically. Both are red on `ccb8dbb8f`. Disclosed as a behaviour change beyond restoring the base: the base skipped such a section entirely, so it appended a second `[autoload]`; head now joins the existing one, which is what Godot's parser says the file means. |
| **P3-4** — the save→restore window and the unguarded write | The restore runs before a `Failure` is returned: a run that never saved compares equal and writes nothing, a run that saved and then failed gets its declarations back, and the failure envelope's code and diagnostics are untouched. The write is wrapped: an `OSError` becomes `ProjectFileRestoreError` (raised from `project_file`, which stays below `gda.errors`), which the CLI turns into **`save_failed`** — an existing registered code, reused by semantic match per the registry's own rule, **no new code minted** — naming the declarations to put back by hand; when the operation had ALSO failed, its envelope stands instead. Four tests (two CLI, two unit) plus a catalog paragraph on what remains: the repair is unreported on the failure path, and a half-written file is not detectable from outside. |
| **P3-8** — a BOM'd `project.godot` got a duplicate line | `read_config_text` drops a leading U+FEFF, so gda names the first key the way the engine's writer does. Test asserts the marked file's `config_version` is not "restored", that the file ends with exactly one `config_version=5`, and that the key the ENGINE mangled is reported under `added_settings` rather than hidden. |
| **P3-5** — the catalog's residual list incomplete | Now four residuals with a sentence each: comments, key order inside a section, **line endings** (the engine's writer emits LF, so a CRLF file comes back LF — stated against the harness install, which is gda's own line edit and preserves CRLF), and **undecodable keys** (excluded from every comparison; near-unreachable, since `c_escape_multiline` emits only `\\` and `\"` — verified in `core/string/ustring.cpp`). |
| **P3-7** — falsified comment | `_customized_settings`' reason is reworded: the initial value IS reachable (`property_get_revert`, which `_op_project_set` now uses), and the ConfigFile read stays right because "written in project.godot" and "differs from the default" are different facts — a key declared AT the default is customized, and `property_get_revert` would hide it from a bare `project list`. |
| **P3-9** — five command functions with no docstring | One line each; `help=`/`_write_help` stays the single authority for the help text, and the byte-diff below confirms the docstrings did not leak into it. |

### Red proof (round 1)

Base = the reviewed head `ccb8dbb8f`, exported with `git archive`, the two changed test modules
copied in and `ProjectFileRestoreError` stubbed so the behavioural assertions run rather than the
whole module failing to import (both write-guard tests are red on the missing symbol alone).

```
FAILED tests/project/test_project_file.py::test_a_section_the_save_emptied_away_is_re_opened_and_the_reorder_reported
FAILED tests/project/test_project_file.py::test_a_leading_byte_order_mark_is_not_part_of_the_first_key
FAILED tests/project/test_project_file.py::test_a_restore_that_cannot_be_written_is_raised_as_a_typed_error
FAILED tests/project/test_project_bounded_write.py::test_a_failed_operation_still_gets_the_file_repaired
FAILED tests/project/test_project_bounded_write.py::test_a_restore_that_cannot_be_written_fails_the_command
5 failed, 31 passed in 1.67s

E   AssertionError: assert False is True                       # sections_reordered
E   At index 0 diff: '﻿config_version' != 'debug/file_logging/enable_file_logging'
E   PermissionError: [Errno 13] Permission denied: '.../project.godot'   # the raw traceback
E   assert 'file_logging/enable_file_logging=false' in '; Engine configuration file.\n\n...'
E   assert 1 == 4                                              # no typed failure
```

And the two installer tests, same base:

```
FAILED tests/harness/test_harness_install.py::test_install_joins_a_whitespaced_autoload_section_it_finds
FAILED tests/harness/test_harness_install.py::test_uninstall_drops_a_whitespaced_autoload_header_it_empties
2 failed, 41 passed in 0.19s

E   assert 'autoload' not in 'config_vers...toload ]\n\n'   # the emptied header left behind
```

### Gates re-run at `454c4c15b`

| Gate | Command | Result |
| --- | --- | --- |
| Unit | `pytest tests -m "not e2e" -q -p no:cacheprovider` | `2540 passed, 672 deselected in 68.43s` (round 0: 2528) |
| Lint | `ruff check .` | `All checks passed!` |
| Format | `ruff format --check .` | `512 files already formatted` |
| Types | bare `pyright` | `0 errors, 0 warnings, 0 informations` |
| Harness install | `pytest tests/harness/test_harness_install.py` | `43 passed` (41 + the two new) |
| Godot e2e | `python <lock> pytest tests/project tests/harness -m e2e -rs -q -p no:cacheprovider` | `119 passed, 312 deselected in 235.44s` — the reviewer's 119 baseline held (the deselected count rose by the 12 new unit tests in those directories); host load `4.98 5.39 4.62` |
| Schema | `gda schema` vs the main checkout at `cfcb8658e` | unchanged from round 0: the same five writers, `description` + the same four output properties; no other command touched |
| Help | 5 × `gda project <cmd> --help` vs the same base | `+5 -0` lines each — identical to round 0, so the new docstrings did not change the help |

`src/gda/harness/gda_harness.gd` is byte-identical to the base (`git diff` empty); no
`HARNESS_VERSION` bump.

---

## Review round 2 (`b06b3d249`)

All 5 findings fixed in one commit; nothing declined. The surface is unchanged from round 1
(same schema diff, same `+5 -0` help diff).

| Finding | What I did |
| --- | --- |
| **P2-1** — the round-1 fix destroyed an EMPTY `[ autoload ]` | Split the two questions the round-1 change had merged. RECOGNITION stays by name in `_ensure_autoload` / `_remove_autoload` — that is how Godot's parser reads the file, and it is what lets gda take its own entry out of any spelling of the section. The header DROP goes back to the literal in `_emptied_autoload_span` (`lines[index].strip() == _AUTOLOAD_HEADER`): the only header gda writes is the only one it removes. Base behaviour for the whitespaced header is restored, and the asymmetry is stated at the comparison itself. |
| **P2-1** — tests | Round 1's `test_uninstall_drops_a_whitespaced_autoload_header_it_empties` is re-scoped and renamed to `..._keeps_...`: the entry goes (it is gda's), the header stays (it is not), `removed_sections == ()`. New `test_round_trip_leaves_an_empty_whitespaced_autoload_section_intact`, parametrized at EOF and mid-file: byte-identical round trip with `created_sections == ()` and `removed_sections == ()` — symmetric receipts. All three are red on `454c4c15b`. |
| **P2-1** — the pre-existing plain-`[autoload]`-empty loss | **Documented, not fixed**, with a pinning test (`test_round_trip_drops_a_pre_existing_empty_autoload_section`) — it is NOT the same one-line rule. Telling "gda wrote this header" from "gda joined it" needs pre-install state this module refuses to persist, and the install's `created_sections` receipt cannot supply it: an uninstall is a SEPARATE run (`daemon stop`, the `export run` strip) that never sees the install's receipt. Comparing the section's content to the gda line is what the code already does — that is exactly the check that cannot tell the two apart. So: a fourth bullet on the module docstring's shapes list (which read as exhaustive and was not, in both directions), the reason stated once in `uninstall_harness` where the guarantee lives, and the test that pins it as a decision rather than drift. |
| **P3-1** — 125-char docstring line | The `ConfigText` paragraph is rewrapped; no line I added anywhere in `src/` exceeds 88 characters (checked by CHARACTER count, not bytes). |
| **P3-2** — the `save_failed` reuse was not authorised by the registry's own clause | `src/gda/error_codes.py`'s module docstring gains one paragraph: reuse also covers a failure the CLI itself SUFFERS while finishing what an engine run started, not only one it reads out of the engine's output — with this restore write as the case, and the same "semantic match, no mirror change" test restated as the boundary. The pinned row description and its ADR-0002 mirror are untouched (`tests/cli/test_error_registry.py` green). |
| **P3-3** — an affirmative "bounded" claim on the one path gda knows nothing | The catalog now says the four keys on that path are **unknown**, not "nothing changed": gda has no reading to compare, so it makes no claim about what the save did. The shape is unchanged (still four empty keys), so the schema is byte-identical to round 1. |
| **P3-4** — the BOM residual was not listed | Now the fifth residual, with what the engine actually leaves: a permanent `"ï»¿config_version"=5` beside its own `config_version=5`, which gda reports under `added_settings` and does not remove, because it is a setting the file now declares. |

### Red proof (round 2)

Base = the reviewed head `454c4c15b`, `git archive` export, only
`tests/harness/test_harness_install.py` copied in.

```
FAILED tests/harness/test_harness_install.py::test_uninstall_keeps_a_whitespaced_autoload_header_it_empties
FAILED tests/harness/test_harness_install.py::test_round_trip_leaves_an_empty_whitespaced_autoload_section_intact[at-eof]
FAILED tests/harness/test_harness_install.py::test_round_trip_leaves_an_empty_whitespaced_autoload_section_intact[mid-file]
3 failed, 43 passed in 0.12s

E   assert '[ autoload ]' in 'config_version=5\n\n[application]\n\nconfig/name="t"\n'
E   assert b'config_vers...ig/name="t"\n' == b'config_vers... autoload ]\n'      # at EOF
E   At index 51 diff: b'o' != b' '                                              # mid-file
```

The fourth new test (`..._drops_a_pre_existing_empty_autoload_section`) passes on both heads —
it pins the documented exception rather than proving a fix.

### Gates re-run at `b06b3d249`

| Gate | Command | Result |
| --- | --- | --- |
| Unit | `pytest tests -m "not e2e" -q -p no:cacheprovider` | `2543 passed, 672 deselected in 46.28s` (round 1: 2540) |
| Lint | `ruff check .` | `All checks passed!` |
| Format | `ruff format --check .` | `512 files already formatted` |
| Types | bare `pyright` | `0 errors, 0 warnings, 0 informations` |
| Harness install | `pytest tests/harness -m "not e2e"` | `56 passed, 4 deselected` (43 + the four new, parametrized) |
| Error registry | `pytest tests/cli/test_error_registry.py` | `21 passed` (the pinned row descriptions still match ADR-0002) |
| Godot e2e | `python <lock> pytest tests/project tests/harness -m e2e -rs -q -p no:cacheprovider` | `119 passed, 315 deselected in 145.06s`; host load `2.99 2.88 2.47` |
| Schema | `gda schema` | byte-identical to round 1 — five writers, `description` + the four output properties, vs the main checkout at `cfcb8658e` |
| Help | 5 × `gda project <cmd> --help` | byte-identical to round 1 (`+5 -0` vs the same base) |

`src/gda/harness/gda_harness.gd` still byte-identical to the base (`git diff` empty); no
`HARNESS_VERSION` bump.

---

## Review round 3 (`383edb0d8`)

All 5 findings fixed, in two commits (`bcc81b10e` the reader/restore, `383edb0d8` the installer).
Nothing declined. The public surface is unchanged from round 2: `gda schema` is byte-identical to
the round-2 head, and each of the five `--help` outputs is byte-identical to it (still `+5 -0`
against the main checkout at `cfcb8658e`).

Every one of the three parser counterexamples was reproduced against the real engine on the
REVIEWED head before anything was changed, and each now has a regression that reads the result
back through a NEW engine launch.

| Finding | What I did |
| --- | --- |
| **P1** — a multi-line quoted string breaks section identification, so a successful `project set` can restore the OLD addressed value | ONE lexical scan (`_scan_line`) now carries the quote, escape and bracket state from line to line and records every section header's line index (`ConfigHeader`) plus each entry's lines. A value continues while a double-quoted string is open OR bracket depth > 0 — `VariantParser::get_token` reads a literal newline inside a string and `String::c_escape_multiline` escapes only `\` and `"`, so the engine WRITES such values back that way. `_section_span` is gone: `_restored`/`_section_end` consume the recorded boundaries, and every insertion point is computed before a single line moves, so the insertions cannot shift each other's targets. `ConfigText.sections` is now derived from `headers` rather than recorded beside it, so one list says where the sections are. |
| **P1 (second form)** — bare-key normalization | `config_key` drops every character of code 32 or less from an unquoted token, exactly as `parse_tag_assign_eof` does (`if (c > 32) … what += c`). `foo bar` and `foobar` are one key again, so a `project set debug/foobar` no longer sees a second setting to "restore" over the value it just wrote. |
| **P2** — `\U` escape width | `_unescape` is rewritten against `VariantParser::get_token` rather than `String::c_unescape`: four hex digits after `\u`, SIX after `\U`, `\b\t\n\f\r` as the switch has them, and `default: res = next` for everything else — that switch knows no `\a` or `\v`, which gda was mis-decoding. UTF-16 pairs are combined the way the tokenizer combines them. What stays `None` is what the engine refuses the whole FILE for: a truncated or non-hex sequence, an unpaired surrogate, a code point outside Unicode. |
| **P2** — make the restore an atomic, optimistic write | `read_config` takes the change token (`st_mtime_ns`, `st_size`) from the SAME open handle the text came from, so it describes exactly the bytes that were scanned. `_replace_file` stages the restored text in a sibling temporary file (mode copied from the target), re-stats the target against that token, and commits with one `os.replace`; a mismatch discards the staged file and raises the new `ProjectFileChangedError`, which `_bounded_write` maps to the registered `file_changed_externally` (exit 4) beside the existing `ProjectFileRestoreError` → `save_failed`. No registry change was needed: `ErrorCodeSource` is explicitly "NOT an exclusive emitter list", the same clause the `save_failed` reuse rests on, and `tests/cli/test_error_registry.py` is green with no producer table to join (the envelope carries no `evidence`). |
| **P2** — the harness installer bypasses the reader's comment handling | `config_line` (comment stripped, line trimmed) is the ONE reduction; `section_name`, `section_of` and `is_section_header` apply it, so they take a raw line, and `_ensure_autoload` / `_remove_autoload` / `_emptied_autoload_span` all read their lines through it. The round-2 rule is untouched — recognition by section NAME, deletion of the literal `[autoload]` header only — so a user's `[autoload] ; note` keeps its comment when gda empties it. |

### Red proof (round 3)

**Real engine, on the reviewed head `b06b3d249`.** Each counterexample was run through the shipped
CLI before any change, and the file the engine left was read back with a fresh `gda project get` /
`project list`:

| Counterexample | What the reviewed head did |
| --- | --- |
| `application/config/description` holding a literal multi-line value with a `[debug]` line, then `config/name="old"`; `gda project set application/config/name --value renamed` | reported `value: "renamed"`, then wrote `config/name="old"` back BELOW the engine's `config/name="renamed"`. A fresh `gda project get application/config/name` returned **`"old"`**. |
| `[zsection]` holding `foo bar=false`; `gda project set zsection/foobar --value true` | reported `value: true` and `restored_settings: ["zsection/foo bar", …]`, and put `foo bar=false` back below `foobar=true`. A fresh `gda project get zsection/foobar` returned **`false`**. |
| `[debug]` holding `"\U000066ile_logging/enable_file_logging"=false`; an unrelated `gda project set` | `restored_settings: []` — the declaration was gone from the file and gda said nothing had been restored. |

The three e2e regressions are red on that head and green on `383edb0d8`:

```
FAILED tests/project/test_e2e_project.py::test_a_header_spelling_inside_a_value_does_not_move_a_restored_line
FAILED tests/project/test_e2e_project.py::test_a_bare_key_holding_a_space_is_the_key_the_engine_reads
FAILED tests/project/test_e2e_project.py::test_a_six_digit_escaped_key_is_restored_like_any_other
3 failed, 41 deselected in 3.63s

E   AssertionError: assert ['debug/config/features'] == ['application/config/features']
E   At index 0 diff: 'zsection/foo bar' != 'debug/file_logging/enable_file_logging'
E   AssertionError: assert [] == ['debug/file_logging/enable_file_logging']
```

And the two installer regressions, same head:

```
FAILED tests/harness/test_harness_install.py::test_install_joins_a_commented_autoload_section_it_finds
FAILED tests/harness/test_harness_install.py::test_uninstall_keeps_a_commented_autoload_header_it_empties
2 failed, 46 deselected in 0.04s

E   assert 'GdaHarness="*res://addons/gda_harness/gda_harness.gd"' not in '…; kept by the user\nGdaHarness=…'
```

The unit-tier additions cannot be run against the reviewed head as a module (they import
`ProjectFileChangedError`, which does not exist there). The boundary defect behind the new
`test_a_restore_lands_in_the_section_the_scan_recorded` was instead shown directly: replaying the
old raw-line rescan over the same engine output returns the span `(6, 9)` — opened by the `[debug]`
INSIDE the description — where the recorded headers are `[(2, 'application'), (9, 'debug')]`, so
the dropped declaration was written into `[application]`.

### Gates at `383edb0d8`

Host: macOS 15 (Darwin 25.6.0), Godot 4.6.3 at the RULES.md path, passed per invocation. Host load
at the e2e run: `load averages: 8.07 20.32 26.51` — busy, and no assertion in this slice is
timing-sensitive. Every Godot-spawning run went through the machine-wide e2e lock;
`GDA_USER_DATA_ROOT` was never exported for a run.

| Gate | Command | Result |
| --- | --- | --- |
| Unit | `pytest -m "not e2e" -q -p no:cacheprovider` | `2554 passed, 675 deselected in 107.12s` (round 2: 2543) |
| Lint | `ruff check .` | `All checks passed!` |
| Format | `ruff format --check .` | `512 files already formatted` |
| Types | bare `pyright` | `0 errors, 0 warnings, 0 informations` |
| Project + harness unit | `pytest tests/project tests/harness -m "not e2e" -q` | `326 passed, 122 deselected` |
| Error registry | `pytest tests/cli/test_error_registry.py -q` | `21 passed` |
| Harness version guard | `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| Godot e2e | `python <lock> pytest tests/project tests/harness -m e2e -rs -q -p no:cacheprovider` | `120 passed, 2 failed, 326 deselected` — see the environment note below; a run at the same code state before the contamination appeared gave `122 passed, 326 deselected in 401.47s` |
| Schema | `gda schema` vs `b06b3d249` and vs `cfcb8658e` | byte-identical to round 2; still exactly the same five changed commands against main |
| Help | 5 × `COLUMNS=88 gda project <cmd> --help` | byte-identical to round 2; `+5 -0` against `cfcb8658e` |

**Environment note on the two e2e failures.** `test_project_list_without_project_is_a_clean_error`
and `test_project_info_without_project_is_a_clean_error` run `gda` from `/tmp` and expect
`project_not_found`. A concurrent process on this machine left a probe project at
`/private/tmp/project.godot` (`config/name="gda888-probe"`, 20:57) with a `/private/tmp/.godot/`
cache beside it, so gda correctly RESOLVED a project and both refusals stopped firing. Proof it is
the environment and not this branch: the same two tests pass on this head when that file is moved
aside (`2 passed, 42 deselected in 3.14s`, the file restored immediately after), and the earlier
run at the same code state, before the file existed, was `122 passed`. Nothing in this branch
touches project resolution. The stray file is left where it is — it is another process's.

`src/gda/harness/gda_harness.gd` is byte-identical to the base (`git diff` empty); no
`HARNESS_VERSION` bump.


qinhaihong-red:
**Review record, round 1 (orchestrator, 2026-09-07).** Independent adversarial review at `c0026fa33` (four axes; 7 real-engine probes; an 18-shape install/uninstall differential base vs head; 12 unit + 4 e2e mutants) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 9 P3**. Every cited line was re-checked at that head before disposition; the fix round is the implementer's. Fix head: `a2fa571e4`.

|---|---|---|
| **P2-1** `sections_reordered: false` while the caller's final section order changed — the save dropped a whole section (its only key default-equal) and the restore re-opened it at the END; `_reordered` compared the engine's intermediate file on shared sections only | Confirmed by reading `_reordered` / `_insert`; engine-reproduced by the reviewer on the GDA-DF-068 shape one key smaller | **Adopted.** Flag computed on the FINAL text; re-inserting at the original position was weighed and declined (edge cases; the file's final layout is the engine's, as the issue records). |
| P3-1 `[ autoload ]` (inner whitespace): uninstall no longer byte-identical — literal header compare vs the name-based `section_of` | Confirmed at `install.py:543` | **Adopted.** All three literal compares are name-based now (matches `VariantParser::_parse_tag`'s `strip_edges`); disclosed: base appended a second `[autoload]` on that shape, head joins the existing section. |
| P3-2 / P3-3 untested missing-section restore branch (M11) and shared-set rule (M8) | Confirmed by the surviving mutants | **Adopted.** `bound_project_write`-level tests; the two rationale tests are characterisation tests (green on the reviewed head), disclosed as such. |
| P3-4 failure-before-restore window; unguarded `write_text` | Confirmed at `project.py:499-502`, `project_file.py:413` | **Adopted.** Restore runs on the failure path too (unsaved → compares equal, no write); the restore write raises `ProjectFileRestoreError`, mapped to the registered `save_failed` (no new code); the op's own failure envelope is preserved; catalog states the residual window. |
| P3-5 catalog residuals incomplete | Confirmed | **Adopted.** Four residuals: comments, in-section key order, line endings (LF; the installer keeps CRLF), undecodable keys. |
| P3-6 #843's mechanism text stale ("gda writes the line"; "one line per category") | Confirmed | **Adopted by the orchestrator** — dated amendment + comment on #843. |
| P3-7 `operations.gd` comment: "no initial-value binding" while `property_get_revert` is used | Confirmed | **Adopted.** Reason reworded. |
| P3-8 BOM'd `project.godot` gets a duplicate line | Confirmed by reading `read_config_text` | **Adopted.** Leading U+FEFF stripped; test. |
| P3-9 five command functions without a docstring | Confirmed | **Adopted.** One line each; help byte-diff unchanged. |

On `a2fa571e4`: unit 2540 (2528 + 12), ruff/format/pyright clean, harness install 43, project+harness e2e 119 (lock-serialized, load ≈5), schema/help diff identical to round 0 (five writers only), `gda_harness.gd` byte-identical (no version bump). Reviewer-verified and unchanged: five-writer sequence diffs to nothing but the request; the four categories match the diff on every write; argv/`--params-json` parity through one recipe; no gda-spelled Godot literal; one ConfigFile-text reader; #829 refusals unchanged; red-proof 13/1 reproduced. A second independent round runs on this head per the wave rule.

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-07).** Second independent review at `a2fa571e4` reproduced every round-1 fix (real engine + fake-engine probes incl. a genuine `OSError` → `save_failed` on both channels with no traceback; 14/16 mutants killed, the two survivors expected) and returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 4 P3**. Fix head: `c5005a246`. The P2 fix was re-verified by the orchestrator directly (diff small; no third Opus round).

|---|---|---|
| **P2-1** the round-1 fix regressed the installer: an EMPTY `[ autoload ]` (inner whitespace) section was deleted by install → uninstall — recognition by name made install JOIN it, and the header drop then removed a line gda never wrote (receipt asymmetric); plain empty `[autoload]` mid-file was already lost at base | Confirmed at `install.py:549`; base/reviewed/fix-head table reproduced | **Adopted, split in two:** RECOGNITION stays name-based (`_ensure_autoload` / `_remove_autoload`, the engine parser's rule); the header DROP is literal again — gda writes only `[autoload]`, so it deletes only `[autoload]`. Orchestrator probe on `c5005a246`: empty `[ autoload ]` at EOF and mid-file and `[ autoload ]` + sibling all round-trip byte-identical with `created=() removed=()`; a section-less file creates and removes `[autoload]` symmetrically. The pre-existing plain-`[autoload]`-empty loss is **recorded and pinned, not fixed**: telling "gda wrote the header" from "gda joined it" needs pre-install state, and uninstall is a separate run that never sees the install receipt — added as the fourth bullet of the module's exclusion list (which read as exhaustive and was not), with a pinning test. |
| P3-1 125-char docstring line | Confirmed | **Adopted.** Reflowed; measured in characters. |
| P3-2 `save_failed` reuse outside the registry's stated rule | Confirmed (`error_codes.py:20-27`) | **Adopted.** Module docstring admits a CLI-owned IO failure while finishing engine-started work; the pinned row description and its ADR-0002 mirror untouched (registry test 21 passed). |
| P3-3 unreadable post-save file reported the four keys as EMPTY | Confirmed | **Adopted.** Catalog says the keys are UNKNOWN on that path; shape unchanged, schema byte-identical to round 1. |
| P3-4 BOM residual absent from the catalog list | Confirmed | **Adopted.** Fifth residual: the engine's permanent `"ï»¿config_version"=5` duplicate, reported under `added_settings`, never deleted. |

On `c5005a246`: unit 2543, ruff/format/pyright clean, harness install 56, error registry 21, project+harness e2e 119 (lock-serialized), schema and the five helps byte-identical to round 1 (vs base: five writers only, +5/−0 each), `gda_harness.gd` byte-identical to base (no version bump), PR CI green. **READY** for merge in the wave order (after #883). Note: commit AUTHOR identity on this branch will be rewritten to the repository owner's before merge (environment artefact, tree hashes unchanged).

qinhaihong-red:
Authorship rewrite (orchestrator, 2026-09-07): every commit on this branch is re-authored to the repository owner's identity — the agent environment had stamped `trial <trial@local>`, which GitHub's squash turns into a `Co-authored-by` trailer. `c5005a246` → `b06b3d249`; tree hashes identical commit by commit, so the reviewed content is unchanged and the pinned-SHA reviews stand by content.

qinhaihong-red:
**Changes required** at exact head `b06b3d2497ec6b08f4db21a9533f6fcb9f7fe097` (base `cfcb8658e67df418a69694840a37a22a9cd3cbe0`).

## Artifact Review

The public artifacts promise that the addressed setting remains the request's value and that every unrelated explicit default declaration is restored verbatim. The real-Godot counterexamples below falsify those guarantees. The PR body is also stale: its Surface diff still cites `41c6d8ae7` and two residuals, while the current refactor commit is `065431bc0` and the catalog lists five. I reviewed issue #843, the complete PR body, and all changed Markdown against the project glossary/ADRs and ASD-STE100. Apart from these factual defects, I found no separate substantive terminology or prose problem.

## Architecture

Centralizing the ConfigFile read/restore model in `project_file.py` and wrapping the five project writers is a sound ownership direction. However, the owner does not carry one lexical state consistently through parsing and insertion: `read_config_text` and `_section_span` independently recognize boundaries, and neither preserves multiline string state. That incompleteness is the root of the P1 below.

**[P2] The new module is not yet the single semantic reader requested by #843.** `read_config_text` strips comments before recognizing sections (`src/gda/project_file.py:308-316`), but harness install/uninstall pass `raw.strip()` directly to lower-level `section_of` (`src/gda/harness/install.py:434-442,592-599`). A valid `[autoload] ; comment` header is therefore missed: install can append a duplicate section and uninstall can leave its entry behind. Have the harness consume the full reader, or expose one raw-line API that owns comment handling. The dependency direction is otherwise sound.

## Standards

**[P2] Make the restore an atomic, optimistic write.** `src/gda/project_file.py:454-457` calls `path.write_text(...)` directly after reading the engine's output. ADR-0018 Decision 4 requires headless writes to be atomic and protected by an optimistic mtime check, returning `file_changed_externally` on interference. This path can truncate a valid engine-written file before a write failure and can overwrite an editor change made after the read. Write a sibling temporary file, recheck the read version/mtime, and atomically replace the target. No other actionable Fowler smell was found.

## Spec

**[P1] A multiline string can make a successful `project set` restore the old addressed value.** `src/gda/project_file.py:323-328` continues values only by bracket depth, not quoted-string state; `_section_span` at lines 518-525 then treats a line such as `[debug]` inside that string as a real section header. In a real Godot 4.6.3 fixture where `application/config/description` contains a literal multiline `[debug]` line before `application/config/name="old"`, setting the name to `renamed` returned success and reported `renamed`, but restoration inserted `config/name="old"` into the misidentified span. A fresh engine read returned `old`. Preserve multiline quote state and make insertion reuse the same authoritative lexical boundaries; add a fresh-engine reload regression.

The same setting-identity failure has a second real-engine form: Godot normalizes the bare key `foo bar` to `foobar`, while `config_key` at lines 156-161 retains the internal space. Setting `debug/foobar=true` returned success, restored the old `debug/foo bar` declaration, and a fresh engine returned `false`. Normalize unquoted keys exactly as Godot does and cover this form too.

**[P2] Preserve Godot's valid six-digit `\\U` key escape.** `_unescape` at lines 182-190 requires eight digits for uppercase `U`, while Godot's Variant parser consumes six. A valid explicit-default declaration keyed as `"\\U000066ile_logging/enable_file_logging"=false` was accepted by Godot but silently disappeared after an unrelated `project set`, with `restored_settings=[]`. Match Godot's six-digit rule and add this exact regression.

Validation: 204 targeted unit tests and six issue-specific real-Godot E2E tests passed; Ruff, format, and isolated exact-head Pyright passed. All three parser/restorer counterexamples used the exact-head CLI and Godot 4.6.3; the two addressed-value failures were verified with fresh engine reads. Required CI was green, while hosted Godot E2E was skipped. The PR head remained unchanged at final verification.

qinhaihong-red:
**Review round 3 handled** — all five findings adopted at `383edb0d8` (implementer under the orchestrator's brief; each disposition re-verified on the head, the escape rules against `4.6.3-stable` `core/variant/variant_parser.cpp`).

|---|---|---|
| **P1** multiline quoted string breaks section identification (two independent recognizers, neither with quote state) | Confirmed, real-engine repro on the old head (fresh `get` returned `old`) | **Adopted.** One lexical scan carries quote/escape/bracket state across lines and records each header's line index and each entry's span; `_section_span` is gone, the restore consumes the recorded boundaries, every insertion point computed before a line moves; `ConfigText.sections` derived from the headers. Fresh-engine reload regression. |
| **P1** bare-key normalization (`foo bar` = `foobar`) | Confirmed (`parse_tag_assign_eof` accumulates only `c > 32`) | **Adopted.** `config_key` drops every char ≤ 0x20 from an unquoted token. Fresh-engine regression. |
| **P2** `\U` six digits | Confirmed (`hex_len = (next == 'U') ? 6 : 4`) | **Adopted and widened to the whole decoder**: `_unescape` now mirrors `get_token` (`\u` 4, `\U` 6, `\b\t\n\f\r`, `default: res = next`, UTF-16 pairing) — the old `\a`/`\v` mappings were mis-decodes. Exact `\U000066ile_logging/…` regression. |
| **P2** atomic, optimistic restore (ADR-0018 D4) | Confirmed | **Adopted.** Change token (`st_mtime_ns`, `st_size`) taken by `fstat` on the handle the text came from; sibling temp + re-stat + one `os.replace`; a mismatch raises `ProjectFileChangedError` → registered `file_changed_externally` beside `save_failed`. Tests: external edit in the window refused and left intact; a failed restore never truncates. |
| **P2** installer bypasses the reader's comment handling | Confirmed | **Adopted.** `config_line` is the one reduction; `section_name`/`section_of`/`is_section_header` apply it and the installer consumes it per line. Round-2 rule intact (recognize by name, delete the literal only). Install + uninstall round-trips with `[autoload] ; comment`, both red pre-fix. |
| **Body** stale commit refs and counts | Confirmed | **Adopted.** `065431bc0`, five residuals, three further stale SHAs repointed, new guarantees stated; catalog paragraph updated. |

On `383edb0d8`: fast tier 2554 passed, project+harness+registry 347, ruff/format/pyright clean, harness guard valid (no harness bytes touched), schema and the five help pages byte-identical to the round-2 head; under the lock: project+harness e2e 122 passed. Disclosed: an earlier probe of this session left a stray `project.godot` in `/private/tmp` which made two cwd-based e2e cases resolve a project — removed; both pass again.

Issue #838, opened 2026-09-05
input action / input tap --action: disclose the state-injection route in help, schema, skill, and the result


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`):
- **GDA-DF-048** (2026-08-20, gda 0.8.0) — `gda input action ui_cancel` press/release returned success, but a modal listening in `_unhandled_input` stayed visible; a native Escape press/release through the same handler closed it. Runtime diagnostics stayed empty.
- **GDA-DF-075** (2026-09-04, gda 0.13.0 at `a938cd4e`) — `input tap --action ui_down` left a scrollable CardView unchanged while `input tap --key Down` scrolled it. The successful action-tap results were read as GUI-path validation and produced a false product diagnosis.

Verified against gda 0.13.0 (`266c8009f`): the harness injects `--action` through `Input.action_press` / `Input.action_release` (a change to the polled action state), while `--key` and mouse events go through `Viewport.push_input` (an `InputEvent` delivered to `_input`, `_gui_input`, `_unhandled_input`). A state change is never an event: nothing that `Input.is_action_pressed` polls is delivered to a Control. `input action --help` currently says the game "observes the action exactly as a real binding would fire", which overstates what the route does.

## What to build

Make the two injection routes explicit everywhere an agent chooses between them, without changing either route's behavior:

- The help of `input action`, `input tap`, and the `action` event kind of `input sequence` states which route each takes and what it reaches: polled state versus delivered event; event-driven UI needs an event.
- The `--schema` descriptions and the bundled skill carry the same distinction and the rule of thumb: drive event-driven UI with key or mouse events; use actions for `Input.is_action_*` polling.
- Every `input` result names the route it used, with the value `action_state` or `viewport_event`: `input action`, `input key`, and `input mouse-move` report one top-level `injection_route`; `input tap`, `input mouse-click`, and `input sequence` report it per phase (a new field on `InputEventPhase`), because one sequence can mix routes; a tap targets exactly one of `key` / `action`, and that target selects the route both its phases take, reported in the same per-phase shape. Additive fields only. (Corrected 2026-09-06: the first wording said a tap could mix routes. See the dated comment.)

No injection behavior changes in this issue. The opt-in event mode for actions is #854.

## Acceptance criteria

[...]

PR #899, opened 2026-09-07
feat(gda): input action --as-event delivers an InputEventAction through the viewport (#854)

### LINKED ISSUE #854: input: opt-in event mode for actions — deliver an InputEventAction through the viewport, with a handler-path conformance fixture
## Parent and evidence

#838 defines route disclosure. This issue adds an explicit event route for actions.
The user need remains valid: GDA-DF-048 (2026-08-20) and GDA-DF-075 (2026-09-04)
show that successful state injection was mistaken for delivery to event-driven UI.
An agent needs to choose the intended route without changing existing calls.

## Requirements

- Keep `input action` and action taps on `Input.action_press/action_release` by
  default. They change polled state, not event-handler input.
- Add `--as-event` to `input action` and to `input tap --action`. Reject it on key
  taps. Accept `as_event: true` only on the action variant of sequence events,
  including the same union used by `screen capture --await-events`.
- The opt-in constructs `InputEventAction` and calls the root viewport's
  `push_input`, not `Input.parse_input_event`. It must leave polled state unchanged.
- Report `injection_route`: top-level for action, per phase for tap and sequence.
  Mixed-route sequences are allowed. No public `as_event` echo or capability count
  is needed in results.
- Help, schema descriptions, and the agent skill state the same conformance matrix:

| injection | `Input.is_action_pressed` | `_input` / `_unhandled_input` | `_gui_input` |
| --- | --- | --- | --- |
| `input action` | yes | no | no |
| `input action --as-event` | no | yes | yes, with a focused Control |
| `input key` of the mapped key | no | yes | yes, with a focused Control |

Handlers above are eligible according to Godot's normal propagation and event
consumption rules. A successful injection reports its route, not a guarantee that
every handler ran or that a Control implemented a particular action binding.

## Design and support assumptions (revised 2026-09-08)

Use the current harness bundled with gda. Mixed-version sessions are not a target;
after updating gda, stop/start an existing session so it loads the current harness.
File synchronization alone does not reload code in a running game. No automatic
restart, capability negotiation, old-version fallback, or event-mode counter is
required by this issue.

The input command group owns input types and route projection. Its existing
descriptor classifier decodes action/tap wire replies and returns public results.
The raw boolean mode is consumed there; it is not kept as a private result field.
Recipes compare requested and applied public routes. Sequence phases remain
request-derived after the existing event-count check, not delivery telemetry.
The screen group reuses the input event union and the harness event application
path; it does not own input compatibility policy.

Public models validate public values and must preserve value equality across
Python and JSON serialization/revalidation. Do not distinguish raw replies from
public values by field shape. Keep malformed-reply validation, but do not diagnose
an old harness version from a missing field. Use the existing classifier and recipe
seams, not a new generic framework. See ADR-0018, ADR-0023, and ADR-0040.

## Acceptance criteria

- [x] Default state behavior is unchanged; opt-in press/release reaches eligible
  event handlers while leaving polled state unchanged.
- [x] CLI, params JSON, schema, help, and skill agree on the option and matrix.
- [x] A real-Godot fixture asserts every matrix cell for an action and its mapped
  key, including press/release, action tap, and a mixed-route sequence.
- [x] The shared capture event channel accepts action event mode.
- [x] Wire decoding and public results have separate responsibilities; malformed
  replies fail, public results round-trip by value, and no hidden provenance or
  old-version capability machinery remains.

Implementation: #899. #838 is the completed prerequisite, not an open blocker.

qinhaihong-red:
## Purpose

Closes #854. Builds on the route disclosure delivered by #838.

State injection is useful for games that poll `Input.is_action_*`, but it cannot
exercise event-driven handlers. This PR adds an explicit action-event route without
changing the default. It does not claim that successful injection proves a UI
action succeeded.

## Public behavior

- `input action --as-event` and `input tap --action NAME --as-event` deliver
  `InputEventAction` through the root viewport's `push_input`.
- Default action injection still uses `Input.action_press/action_release`.
  Event-mode injection leaves polled action state unchanged. It does not use
  `Input.parse_input_event`, which would also change that state.
- Sequence action events accept `as_event: true`; other event variants reject it.
  Key taps reject `--as-event`. A sequence can mix routes, reported per phase.
- `screen capture --await-events` reuses the same event union and harness event
  application path, including scheduled release events after an early match.
- Help, schema, skill, catalog, and README translations describe the opt-in.
  The conformance matrix in help/schema/skill includes normal propagation and
  consumption limits. `injection_route` remains the public mode disclosure.

## Revised design (2026-09-08)

This section supersedes the earlier review-round implementation notes in the PR
history. The requirement is retained; old-session compatibility is not a target.
Use the current harness bundled with gda. After updating gda, stop/start an existing
daemon session: syncing its installed file does not reload the running game.
No input command automatically restarts a user's game.

| Responsibility | Owner and existing seam |
| --- | --- |
| Input event types and route projection | `commands/input.py` |
| Decode action/tap wire replies | Group-local descriptor classifiers, using private reply adapters with `classify_live` |
| Public result values and gesture invariants | `InputActionResult` / `InputTapResult`; no wire provenance |
| Compare requested and applied routes | Action/tap recipes, using public results only |
| Sequence phase projection | Existing recipe, after the total event-count check; request-derived, not handler telemetry |
| Capture predicate, image, and receipt | `commands/screen.py`, reusing the public input event union |
| Engine event application | Shared harness action-event helper and sequence application path |

Removed: private echo-presence attributes, source-shape guessing, wrap validators
that retain wire metadata, old-harness diagnosis/fallback logic, and the per-window
event-mode counter across input and capture. Ordinary malformed-reply checks and
route correlation remain. There is no new generic adapter framework, descriptor
field, or routing registry. The public JSON/schema structure is unchanged; only
help/schema explanatory prose gains the propagation limit.

The earlier claim that raw reply normalization had to stay in public models to
preserve their schema was incorrect: the existing classifier seam can do that
conversion first. `_RECIPE_OPERATIONS` remains a legitimate test expectation, not
a production registration table.

Authorities updated: #854, CONTEXT.md, ADR-0018 (current harness/session lifetime),
ADR-0023 (reply/public result boundary), and ADR-0040 (command-group ownership).
The bundled harness version is 22 with an updated content hash pin.

## Scope and integration

Only #854 is implemented here. No target-branch merge is included in the delivered
history. Integration with the latest target branch and the order of the six PRs
are deferred to the maintainer's coordinated merge plan.

## Validation

At `1256cec90537ae31d8e977015908d495312d080f`:

- Fast tier: 2548 passed, 2 skipped, 668 E2E cases deselected.
- Real Godot input and screen suites: 36 passed, including the complete route
  matrix and capture event-mode press/release drain test.
- Ruff check/format and Pyright: clean. Harness version transition: valid.
- Aggregate schema comparison with `64db8ad93`: description text changes only;
  fields, required keys, types, defaults, and bindings are preserved.
- Independent Standards and Spec reviews: one P3 documentation finding each,
  both corrected and rechecked; no remaining finding.

[Exact-head CI passed](https://github.com/aigengame/godot-agent/actions/runs/34188622567).
The hosted Godot job was skipped by workflow policy; it does not replace the
local real-engine evidence above. See the completion comment for the review record.


qinhaihong-red:
**Review record, round 1 (orchestrator, 2026-09-07).** Independent adversarial review at `08c5e3b4b` (four axes; live probes on Godot 4.6.3 incl. `_gui_input` on a focused `Button`, strength on the event route, both refusal channels and the published schema; 11 mutants) returned **CHANGES REQUIRED — 0 P1 / 2 P2 / 6 P3**. Cited lines re-checked before disposition; the fix round is the implementer's. Fix head: `b4a216fd0`.

|---|---|---|
| **P2-1** the SKILL matrix cells were unguarded — flipping the opt-in row left the fast tier green while help/schema copies are pinned by `_MATRIX_ROWS` | Confirmed (SKILL `:272-276`; mutant M9) | **Adopted.** The consistency test now reads the skill table, normalises rows, parses `_MATRIX_ROWS` into claims and consumes rows longest-spelling-first with no leftover (pins the table size too). Flipped row → 1 failed. |
| **P2-2** no test at any tier for the single-frame `--as-event --release` (harness release→pressed mutant survived; the tap's is caught) | Confirmed (mutant M5b) | **Adopted.** `MATRIX_UI_GD` gained `is_action_released` counters; the fixture adds an `--as-event --release` step asserting the release counters advance while press counters and polled state do not; the reviewer's mutant now fails (`assert 0 == 1`). |
| P3-1 help absolute "however the action is bound" contradicted two paragraphs later | Confirmed | **Adopted.** Scoped to the state route; paragraph re-flowed (clears a pre-existing orphan). |
| P3-2 catalog orphan line after the insertion | Confirmed | **Adopted.** Re-flowed, wording unchanged. |
| P3-3 "all three injections" is two | Confirmed | **Adopted.** "the two DISTINGUISHING injections", with the substring reason. |
| P3-4 comment promised a refusal `_check_tap_evidence` does not make | Confirmed | **Reworded, not enforced** — enforcing would require publishing `as_event` as a result field only so the validator could see it (declined in this PR); such a reply routes `viewport_event` either way, a key tap's real route, so nothing wrong is published. |
| P3-5 human render identical for both routes of `input action` | Confirmed (`render_input_action`) | **Adopted.** ` as event` appended when the route is `viewport_event`; pinned by the first test on an input renderer. |
| P3-6 fixture comment's arrow-key reason refuted on the engine | Confirmed | **Adopted.** X kept; reason rewritten (independence from focusable siblings and `ui_*` overlap). |

On `b4a216fd0`: fast tier 2521 (+1 render pin), ruff/format/pyright clean, harness bytes unchanged (21, pin `78cbfec5…`, guard valid), schema diff unchanged from round 0 (4 commands, strictly additive), help: `input action` gains the P3-1 scoping and re-flow, `tap`/`sequence` unchanged from round 0, the other three byte-identical; input+harness live e2e 22 passed, no skips. Reviewer-verified and unchanged: one derivation with two inputs and no fallback, no new event kind, key-set pin intact; one harness primitive pushing to the same viewport as key/mouse; unmapped action refused (`live_unknown_action`); refusals identical on argv, `--params-json` and structurally in the published schema; the echo fallback judged sound. A second independent round runs on this head per the wave rule.

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-07).** Second independent review at `b4a216fd0` reproduced all eight round-1 fixes (matrix guard kills the flip and eight structural mutants; the release step kills the harness mutant and three more; help/catalog diffs exactly as claimed; own-fixture AC1/AC2/AC4 on a real engine incl. the `_input` half) and returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 5 P3**. Fix head: `9aef79b86`; the diff is small (one module, SKILL, one test module), re-verified by the orchestrator directly (no third Opus round).

|---|---|---|
| **P2-1** the new render rationale ("`input action` is the ONE command whose route varies per call") is false — a tap's route has varied by target since #838 — and this PR made `tap action X` ambiguous on the human channel while leaving it alone | Confirmed: `render_input_tap` byte-identical for `as_event` true/false | **Adopted.** `render_input_tap` appends ` as event` on an action tap whose phases took the event route; both rationales rewritten to the true trigger (ambiguity, not variation; a sequence's line was never route-bearing); pinned by `test_the_human_render_names_the_event_mode_on_an_action_tap` (mutant → 1 failed). |
| P3-5 the PR body's reason for not refusing a key-tap echo of the mode was false (`_name_the_route` is `mode="before"` and reads the raw payload) | Confirmed | **Adopted — the two-line refusal is implemented** ("a key tap result cannot echo the event mode."), the malformed reply added to `_MALFORMED_TAP_REPLIES`, the round-1 comment and PR-body claim corrected. Orchestrator probe on `9aef79b86`: `{**INPUT_TAP_KEY_RESULT, "as_event": True}` → `ValidationError`. |
| P3-1 guard comment overstated what the SKILL matrix test pins | Confirmed | **Adopted.** Comment states the scope (three yes/no verdicts; header and non-yes/no rows skipped; `(focused Control)` not asserted — deliberately, since no surface owns that qualifier as a claim). |
| P3-2 module docstring's pre-#854 absolute | Confirmed | **Adopted.** Scoped to the route; names `--as-event` / `as_event`. |
| P3-3 SKILL "drive `Input.action_press`" stated of the commands | Confirmed | **Adopted.** "By default …". |
| P3-4 narrower re-wrap | Confirmed | **Adopted.** 76–80 columns; one short line kept so a backticked span is not split. |

On `9aef79b86`: fast tier 2522, ruff/format/pyright clean, harness unchanged (21, pin `78cbfec5…`, guard valid), schema diff unchanged from rounds 0–1 (4 commands, additive; only `input action.description` moved with the re-wrap), input+harness live e2e 22 with no skips; input unit module 111 passed in the orchestrator's re-run; PR CI green. **READY** for merge in the wave order (after #898; first of the harness pair, lands as 21). Note: earlier commits' AUTHOR identity on this branch will be rewritten to the repository owner's before merge (environment artefact; the round-2 commit already carries the right identity).

qinhaihong-red:
Authorship rewrite (orchestrator, 2026-09-07): every commit on this branch is re-authored to the repository owner's identity — the agent environment had stamped `trial <trial@local>`, which GitHub's squash turns into a `Co-authored-by` trailer. `9aef79b86` → `0d93ef063`; tree hashes identical commit by commit, so the reviewed content is unchanged and the pinned-SHA reviews stand by content.

qinhaihong-red:
**Changes required** at exact head `0d93ef0631aff117178f27b1ea9a7d57d07bb07c` (base `cfcb8658e67df418a69694840a37a22a9cd3cbe0`).

## Artifact Review

The PR body overstates two guarantees. First, it says the hidden `as_event` echo is validated, but `src/gda/commands/input.py:536-550` and lines 760-784 apply Python truthiness rather than requiring a boolean. A reply containing `"as_event": "false"` publishes `viewport_event`; a key-tap reply containing `as_event: false` is accepted despite the stated contract. Second, the requirement table says the live fixture asserts every matrix cell, but `tests/live/test_e2e_input.py:226-249` defines no `_input` observer.

I reviewed issue #854, parent #838, the complete PR body, and all changed Markdown against project terminology and ASD-STE100. No other substantive terminology or prose issue was found. Spanish and Japanese were not native-reviewed.

## Architecture

The ownership is otherwise sound: `input` owns the event union and route vocabulary, `screen -> input` is the one-way reuse allowed by ADR-0040, and `_push_action_event` centralizes the three harness injection sites.

**[P2] The protocol addition has no safe transition for an already-running old Engine session.** `src/gda/commands/daemon.py:816-847` deliberately syncs a stale harness on disk without retiring the session already serving the daemon; only the next session loads it. Base harness v20 ignores `as_event` and does not echo it. With the v21 CLI, action/tap event-mode requests can therefore exit successfully as state injections, while sequence phases are synthesized from the request and can falsely report `viewport_event`. `screen capture --await-events` forwards the same union to the same session.

Please either retire/relaunch the active session after a harness sync, or capability/version-gate it and correlate authoritative applied-mode evidence with every request. The output boundary must reject missing, non-boolean, or request-mismatched mode evidence rather than applying truthiness.

## Standards

**[P2] Exercise event-mode sequence release against Godot.** The mixed sequence at `tests/live/test_e2e_input.py:1093-1129` has an opted-in press, but its release is a state-route event. Changing `src/gda/harness/gda_harness.gd:1499` from `not releasing` to constant `true` left all 111 focused fast tests and the real-Godot matrix green. Add an `as_event: true, release: true` event, assert its `viewport_event` phase and handler release counters, then separately release the state-route press. No other actionable Fowler smell was found.

## Spec

**[P2] Add `_input` to the live conformance fixture.** Issue #854 requires the opted-in event to reach `_input`, focused `_gui_input`, and `_unhandled_input`, while the default route reaches none. Add `_input` press/release counters and assert state, opted-in action, and mapped-key behavior.

Validation: 111 focused input tests, 58 focused screen tests, and the exact-head real-Godot matrix passed. The release mutant survived both the focused fast suite and live matrix. Godot source inspection confirmed the intended `Viewport.push_input` route. Required CI is green, but its Godot E2E job is skipped; the extra windowed screen-capture probe could not run because this environment rejected WindowServer. The PR head remained unchanged at final verification.

qinhaihong-red:
**Review round 3 handled** — all four findings adopted at `7646cad29` (implementer under the orchestrator's brief; each disposition re-verified on the head).

|---|---|---|
| **P2 (Architecture)** no safe transition for a running pre-v21 session | Confirmed: the idempotent start syncs the file only; a v20 session ignores `as_event`, and sequence phases are request-derived | **Adopted — capability gate at the output boundary, no relaunch op.** Single-event ops (`input action`, `input tap`) correlate the requested door with the reply's strictly-boolean echo in their descriptor recipes; list-carrying ops (`input sequence`, `screen capture --await-events`) correlate the request's opt-in count with a new harness echo `event_mode_actions` (carried, never published — schema byte-identical). Missing echo/count for an opted-in request, a non-boolean, a non-integer, or a mismatch → `contract_violation` naming the stale harness and the `daemon stop` / `daemon start` remedy. A missing count with zero opt-ins is tolerated (a v20 session still answers such a request honestly). 14 fake-runner cases cover the six shapes named plus the opposite-direction mismatch. |
| **Artifact** two overstated body guarantees | Confirmed (truthiness at both echo sites; no `_input` observer) | **Adopted.** `echoed_event_mode` reads a JSON boolean only; a key-tap reply carrying ANY `as_event` key is refused. Body sentences rewritten; totals and head reconciled. |
| **P2 (Spec)** `_input` missing from the live fixture | Confirmed | **Adopted.** `input_hits` / `input_release_hits` counters; all three handlers asserted together, press and release, at every matrix step. |
| **P2 (Standards)** event-mode release mutant survived | Confirmed | **Adopted.** The mixed sequence carries an opted-in release (frame 2) and a state release (frame 3); phases assert `viewport_event` then `action_state`, handler release counters move by exactly one. Mutant re-run under the lock: `1 failed` (`assert 5 == 4`), reverted. |

On `7646cad29`: fast tier 2536 passed (+14), ruff/format/pyright clean, harness guard valid (version 21, pin `3f5934a6…`), `gda schema` and the three `input` help pages byte-identical to the round-2 head; under the lock: `test_e2e_input.py` 18 passed (no skips), harness e2e 4, screen `--await-events` e2e 6. Residual: the gate proves the session APPLIED the mode, not that the engine delivered the event — that stays the live matrix's job; only the macOS leg was run locally. The private-attribute carry on `InputSequenceResult` is deliberate (keeps the shared `execute` seam and the schema) and commented as such.

qinhaihong-red:
**Re-review: changes required** at exact head `7646cad29fc10fe571a71e075c0fef833b395c90` (base `cfcb8658e67df418a69694840a37a22a9cd3cbe0`).

The round-3 changes close the stale-session correlation, strict-echo, `_input`, and event-mode release findings. One P2 residue remains.

**[P2] Keep public input results stable when they are validated again after the private harness echo has been consumed.** `InputActionResult._name_the_route` (`src/gda/commands/input.py:625-641`) and `InputTapResult._name_the_route` (`:850-874`) recompute the public `injection_route` on every validation from the private raw `as_event` echo. That echo is intentionally omitted from the public result, so validating the model's own dump silently changes truthful event-route evidence back to the state route:

```text
InputActionResult: viewport_event -> model_dump -> model_validate -> action_state
InputTapResult:    [viewport_event, viewport_event] -> model_dump -> model_validate -> [action_state, action_state]
```

I reproduced both transitions directly on this exact head. This contradicts the model's stated contract that `injection_route` is the mode's whole public disclosure and means a schema-valid result cannot survive a normal serialize/revalidate boundary without changing meaning.

Please separate raw-harness normalization from public-result validation, or preserve and validate an already-published route when the private echo is absent. Add `model_validate(model_dump())` regressions for both an event-mode action and an event-mode action tap; each should retain `viewport_event`.

Focused verification otherwise passed: 28 relevant unit tests and the real-Godot event-mode conformance matrix. The release-to-press mutant is killed by that matrix (`assert 5 == 4`), confirming the new release coverage is effective. Required hosted checks pass; hosted Godot E2E remains skipped.

qinhaihong-red:
**Review round 4 handled** — the one P2 finding is adopted at `4f4f14c95` (implemented and verified by the orchestrator; harness bytes untouched, `HARNESS_VERSION` stays `21`, pin unchanged).

**[P2] Public input results changed meaning when validated again after the private echo was consumed** — **fixed**. Both transitions reproduced verbatim on `7646cad29`. `InputActionResult._name_the_route` and `InputTapResult._name_the_route` now keep an already-published route when the payload echoes no `as_event`: a payload with the echo is a harness reply and is normalized strictly as before; one without the echo that already names its route (the head field / every phase, `_phases_published`) is a published result coming back through validation and is kept, with the field's vocabulary check still running; one with neither reads as the state route, which the recipe's correlation then judges. A key tap is re-derived on every pass as before (its route is fixed by its target) and one that echoes the mode at all is still refused. Chosen over moving the normalization out of the validator because `InputEventPhase.injection_route` is required and a harness reply's phases carry none — the fold has to happen before the phase models are built (the #838 design on `main`), so the second remedy you offered is the one that changes no schema. Regressions: `model_validate(model_dump())` and `model_validate_json(model_dump_json())` retain `viewport_event` for an event-mode action and for an event-mode action tap (both fail on the round-3 head), and the state-route pair stays `action_state` across the same boundary.

Gates at `4f4f14c95`: fast tier `2539 passed` (+3); ruff / format / pyright clean; `gda schema` and `gda input action|tap --help` byte-identical to `7646cad29`; the real-Godot event-mode matrix `tests/live/test_e2e_input.py` `18 passed` under the machine-wide lock. The PR body carries a "Review round 4" section with the same record.

qinhaihong-red:
**Re-review: changes required** at exact head `4f4f14c95363dd9b52e511e86a71c3c4d11b3ca8` (base and merge base `cfcb8658e67df418a69694840a37a22a9cd3cbe0`).

The original serialize/revalidate failure is fixed: event-mode action and action-tap results now retain `viewport_event` across both Python and JSON round trips. The implementation is nevertheless **Partially Adopted** because the chosen source-shape test introduces one P2 boundary regression.

**[P2] Do not infer that a payload is a published result merely because it already carries public route fields.** `InputActionResult._name_the_route` (`src/gda/commands/input.py:652-662`) and `InputTapResult._name_the_route` / `_phases_published` (`:462-476`, `:901-908`) use the presence of `injection_route` to distinguish a revalidated public result from a raw harness reply. The same output models validate both sources, however, so a raw or drifted harness reply can supply those public fields itself and bypass the round-3 authoritative-echo gate.

Reproduced through the CLI's fake live-runner seam on this exact head:

- `input action --as-event`: a reply with no `as_event` echo but top-level `injection_route: viewport_event` exits 0 instead of `contract_violation`.
- An action tap with no echo but `viewport_event` on both phases likewise exits 0.
- An action tap with `press=viewport_event` and `release=action_state` exits 0 and publishes the contradictory routes. `_check_tap_evidence` validates frames and phase names but not their one-route invariant, while `_input_tap_recipe` correlates only the first phase.

The last payload violates #838's contract that one tap's target selects the route both phases take. The identical command-level counterexample is rejected on `7646cad29` and accepted on `4f4f14c95`, so this is a regression introduced by the round-4 fix. The assertion that a harness reply “never” carries these fields is not enforceable at the validator that receives both shapes.

Please distinguish raw-harness normalization from public-result revalidation explicitly rather than by field presence. If public `injection_route` is instead intended to become a second authoritative protocol echo, document and validate that contract and still reject mixed tap-phase routes. Add regressions for the two no-echo/public-route replies and for an action tap whose two phases disagree.

Verification: all 123 focused input unit tests pass; the real-Godot event-mode matrix passes; the three independent boundary counterexamples above fail their expected-refusal assertions. The exact-head `workflow_dispatch` CI run succeeded, with hosted Godot E2E skipped.

qinhaihong-red:
## Residual review finding addressed

**Status: Fully adopted in [`64db8ad93`](https://github.com/aigengame/godot-agent/commit/64db8ad93ccc569df9918c479f4af68abfc282cb).**

The event-mode evidence boundary now distinguishes an `as_event` echo present in the raw harness reply from the public `injection_route` field carried by a published result. Consequently:

- `input action --as-event` and `input action tap --as-event` reject replies that omit the authoritative `as_event` echo, even when a public route field says `viewport_event`.
- Action taps reject mixed press/release injection routes instead of publishing an internally inconsistent result.
- Public `InputActionResult` and `InputTapResult` payloads still preserve their route across `model_dump()` / `model_validate()` and JSON round trips; the schema and command help surfaces remain byte-identical to the previous head.

Regression tests reproduced all three counterexamples before the fix (`3 failed`) and pass after it.

Validation on exact head `64db8ad93ccc569df9918c479f4af68abfc282cb`:

- Targeted regression/round-trip tests: `7 passed`
- Focused input/screen unit tests: `189 passed`
- Full fast tier: `2540 passed, 2 skipped, 667 deselected`
- Real Godot input E2E: `18 passed`
- Ruff lint and format: passed
- Pyright: `0 errors, 0 warnings`
- [CI run 34184444686](https://github.com/aigengame/godot-agent/actions/runs/34184444686): passed (hosted Godot E2E was conditionally skipped; the real-engine suite above ran locally under the repository lock)

GitHub currently still reports the PR itself as `CONFLICTING` / `dirty` against base `cfcb8658e67df418a69694840a37a22a9cd3cbe0`; that merge-state issue is separate from this reviewed behavior fix.

qinhaihong-red:
## Redesign completed

The #854 requirement is retained. The previous compatibility-driven design is
replaced with the current-harness policy and the existing classifier/recipe
boundaries. Issue and PR bodies, CONTEXT, ADR-0018/0023/0040, catalog, skill, and
help/schema explanations now agree.

| Review claim | Decision | Result |
| --- | --- | --- |
| Public results should not carry hidden raw-source evidence | Fully adopted | Reply adapters consume wire mode before returning ordinary public values; Python/JSON round trips preserve full model equality |
| Old harness compatibility justified per-feature gates and counters | Not retained under the revised requirement | Removed old-session fallback/diagnosis and input/capture mode counters; use the current bundled harness |
| Raw normalization had to remain in public models to preserve schema | Rejected | Existing descriptor classifiers perform the conversion; public schema structure is preserved |
| A manually enumerated test expectation is a forbidden production registry | Not adopted | `_RECIPE_OPERATIONS` is retained; its obsolete compatibility comments are corrected |

## Standards

Independent review found one P3 stale test comment describing removed capability
counters. It was corrected. No remaining documented-standard or material
architecture finding.

## Spec

Independent review found one P3 missing propagation/consumption qualification in
action help/schema. It was added and is covered by the matrix surface test. No
remaining functional requirement gap or scope-expansion finding.

## Artifact and prose assessment

Checked the issue/design/code relationships and the bundled skill's complete text
against CONTEXT and the relevant ADRs, using ASD-STE100 as the technical-prose
reference. The changed text distinguishes installation from an engine session,
wire replies from public values, and route disclosure from handler/UI success.
No remaining substantive terminology or prose finding in the changed scope.

## Delivery and verification

Exact head: `1256cec90537ae31d8e977015908d495312d080f`.
New commits: `a4c5f20a4` (implementation/design) and `1256cec90` (review follow-up).

- Fast tier: **2548 passed, 2 skipped, 668 deselected**.
- Real Godot input + screen E2E: **36 passed**, run serially under the shared lock.
  This includes every conformance-matrix cell, action/tap/sequence press/release,
  and a capture that matches early but still applies the scheduled release.
- Ruff check/format and Pyright: clean; harness version/hash and transition guard
  pass (version 22).
- Compared the aggregate manifest with `64db8ad93`: only explanatory descriptions
  changed; schema fields, required keys, types, defaults, and bindings are intact.
- Net change from the previous PR head: **410 fewer lines** across implementation,
  tests, and documentation; no added compatibility framework.

Hosted CI: [exact-head run](https://github.com/aigengame/godot-agent/actions/runs/34188622567)
**passed**, including all required gda-balancing checks. The hosted Godot job was
skipped by workflow policy; the 36 real-engine tests above were run locally.

No target-branch merge or PR merge is included. The six-PR integration order is
left to the maintainer's coordinated plan. Earlier PR comments remain historical;
their old-session gates and private-provenance recommendations are superseded by
the revised issue and this design.

qinhaihong-red:
ADR form change at `80490fb33` (maintainer's decision): the three `##` sections that `a4c5f20a4` added mid-ADR (ADR-0018 current-version policy, ADR-0023 reply decoding, ADR-0040 input boundary) are now the dated `> **Outcome (2026-09-08, #854)**` notes these ADRs already use, placed where each file keeps them (ADR-0018 before `## Decision`, the other two after `## Consequences`). The #225 outcome note in ADR-0018 keeps its original wording; the new note carries the session-lifetime clarification and records the policy's rationale as decided by the maintainer: old-version support is not a priority and warrants no disproportionate complexity. Docs-only: no code, schema or harness change (`tests/repo` 102 passed). Head is now `80490fb33`.

qinhaihong-red:
`cd0200025`: `HARNESS_VERSION` reset from 22 to 21, the merge base's version plus one (the wave rule; 21 was never shipped). The sha256 pin is unchanged. Harness unit tests 51 passed; `harness_version_guard` valid. Head is now `cd0200025`.

Issue #876, opened 2026-09-06
[gda-balancing] Remove duplicate value primitives and compose extrema with explicit charges


qinhaihong-red:
## Implementation refinement — 2026-09-07

Implementation starts from reviewed dev `6a32d8ced05e4dcd6c02e59d7053144500528b83` after #612 (PR #910). The accepted deletion endpoint is unchanged.

An actually admitted in-memory candidate that replaces the two quantity maximum instructions with existing less-than/select and increments their exact bounds causes four of six maintained Models to refuse. In `rpg-stat-composition`, `floor_zero([-1000,4040])` loses its `[0,4040]` interval and becomes `[-1000,4040]`, producing `language.formula_type_mismatch`; nominal type, unit, representation and numeric policy are unchanged. Pure result equivalence was not sufficient evidence of compositional compiler admission.

The delegated engineering refinement is compiler-owned predicate-aware selection inference in the existing Formula owner. Restrict each branch to possible values, then join; preserve actual Kernel Boolean contracts and eager evaluation/refusal. Do not retain maximum-specific host matching, widen authored contracts to hide the regression, or introduce a solver/Runtime fallback. Generic precision improvements can also sharpen existing minimum domains; refresh affected authored expression/body pairs and current identities consistently. The new branch records the exact evidence, alternatives and validation in `docs/refactor/current-language/PRIMITIVE-COMPOSITION.md`.

This is an implementation prerequisite inside #876, not a new HITL blocker. No AC is complete until production deletion, current public cases, independent consumers, full checks and review pass.

---

## Parent

#865 · stage S6a · AFK

## What to build

Delete the exact value/copy synonym and move maximum into admitted compare/select composition through the real Formula Runtime seam owned by #612. Replace the internal current resource baseline explicitly; do not add a compatibility charge exemption.

## Acceptance criteria

- [x] The unused value alias and its production contracts/dispatch/vector obligations are deleted; one binding/read meaning remains.
- [x] Maximum executes as LDB composition with declared current resource charges; its redundant irreducible Kernel node and host branch are removed after all consumers migrate.
- [x] Int64 extremes and at/below/above resource limits distinguish result equivalence from changed charging; old evidence is rebuilt or withdrawn, not reused under false identity.
- [x] Unsafe algebraic reductions are rejected by permanent counterexamples (MIN−MIN versus intermediate negation; comparison versus overflowing difference); no unproven state/cancel/guard deletion is bundled.
[...]

Issue #865, opened 2026-09-06
[gda-balancing] Refactor to one current language and delete obsolete version and execution bindings


qinhaihong-red:
## Documentation delivery

PR #881 records this direction at `fc7bac0025fde98c183ccddc6f82eddc58a9622e`. Start with the [document index](https://github.com/aigengame/godot-agent/blob/fc7bac0025fde98c183ccddc6f82eddc58a9622e/libs/gda-balancing/docs/refactor/current-language/README.md), [implementation plan](https://github.com/aigengame/godot-agent/blob/fc7bac0025fde98c183ccddc6f82eddc58a9622e/libs/gda-balancing/docs/refactor/current-language/PLAN.md), [issue graph](https://github.com/aigengame/godot-agent/blob/fc7bac0025fde98c183ccddc6f82eddc58a9622e/libs/gda-balancing/docs/refactor/current-language/ISSUES.md), [evidence and reproduction](https://github.com/aigengame/godot-agent/blob/fc7bac0025fde98c183ccddc6f82eddc58a9622e/libs/gda-balancing/docs/refactor/current-language/EVIDENCE.md), and [validation limits](https://github.com/aigengame/godot-agent/blob/fc7bac0025fde98c183ccddc6f82eddc58a9622e/libs/gda-balancing/docs/refactor/current-language/VALIDATION.md). Existing internal tags/release metadata are distinct from the owner-defined formal product release; they create no historical language-retention promise.

The owner has now authorized sequential implementation on `codex/gda-balancing-refactor-delete-dev`, created from the documentation head above. PR #884 completes #866 delivery to that development branch after independent review; PR #881 remains the original documentation proposal. #865 and implementation #867–#879 require their own acceptance evidence. Each issue uses a branch and reviewed PR into dev; the accumulated dev branch waits for final review before main integration. The primary agent owns implementation decisions under orthogonality, DRY, architectural consistency and control of unnecessary complexity.

## Parent

- #501
- #534

## Problem and evidence

Internal revisions have been treated as permanently immutable releases even though gda-balancing has not formally released. This has split required capabilities across package lines and amplified changes through historical selection and exact-build execution bindings. The owner accepted a first-principles refactor: delete unnecessary mechanisms, keep one current compositional language, and defer any formal compatibility commitment until no earlier than toolkit v1.0.

Bounded experiments at `3f68bf3fb26df2ab54351a8ef4e3e167269bdc16` provide evidence:

[...]

Issue #612, opened 2026-08-03
[gda-balancing] Route value-program conformance vectors through the real Formula Runtime seam


qinhaihong-red:
## Delivery completed — 2026-09-07

PR #910 is merged into `codex/gda-balancing-refactor-delete-dev` at `6a32d8ced05e4dcd6c02e59d7053144500528b83`. All nine ACs are verified at reviewed head `075a5aae74d257f8968c408faa4b4b304055f85b`; [independent reviews, full CI and rollback evidence](https://github.com/aigengame/godot-agent/pull/910#issuecomment-5568406818) record the completion scope. Complete CI accounts for 1,741 tests: 1,718 passed, 23 existing allowed skips, zero failures/errors. Standards, Spec and domain-architecture reviews each report zero findings.

The shared production seam and vector-helper deletion are complete. The two public resource-refusal defects are corrected under the existing audit laws. Closure means reviewed delivery to dev, not main integration, a formal release or #876 primitive deletion.

---

## Implementation scope — 2026-09-07

Implementation starts from the reviewed development baseline `aa94ec79d63b0595e5d6af8bce6180aba9d5b397` after #875 (PR #906). The delegated engineering decision is to extract one charged per-program evaluation function inside the existing Runtime owner, use it from normal initialization/Event/observation and the test-side vector adapter, and delete the shipped vector-only entry point. Inline Event instruction ledgers and independent reference/replay implementations remain separate responsibilities.

The byte-preservation AC below is clarified to follow #875: unchanged execution meaning keeps its semantic artifacts, while changed Runtime source must change truthful evaluator provenance and its publication records. Freezing the producer fingerprint would violate the accepted execution identity contract. This does not authorize a Kernel/LDB, Formula, numerical, resource or public result-contract change. The issue remains open until its implementation, independent review and complete checks pass.

---

## Confirmed public resource-refusal defect — 2026-09-07

Actual public build/check/run with correctly admitted Runtime profile node limits of 5 (Event Formula) and 20 (observation Formula) reproduces an internal publication failure on baseline `aa94ec79d` and the initial seam extraction. The prepared terminal audit omits attempted Formula charges: an exception prevents the caller assignment from receiving updated consumed steps. The independent artifact validator correctly rejects that counter mismatch.

[...]

PR #910, opened 2026-09-07
fix(gda-balancing): share Formula evaluation and preserve refusal charges

qinhaihong-red:
## Problem and change

The shipped value-program vector helper had its own meter, cache and refusal loop, so passing conformance vectors did not prove that normal Formula execution used the same behavior. Initialization, derived Event and observation Formula programs now use one production-owned evaluation function; the vector adapter translates requests under tests, and the shipped vector-only helper is deleted. Independent reference evaluation and artifact replay remain independent. Inline Operation Formula keeps its existing per-instruction Event ledger.

Actual public low-budget runs also exposed a pre-existing fault: Event and observation callers lost charged steps when Formula evaluation raised, causing their terminal audits to fail semantic validation and return internal errors. Both callers now retain the fault charge. Limits 5 and 20 publish the existing typed refusal with 6 and 21 charged steps instead of invalid audits with 3 and 18 steps. No validator, Kernel/LDB, public schema or resource law is weakened or changed.

## Validation

- Integrated seam/public/vector checks: **24 passed**. Each of the ten shipped vectors is independently checked against its expectation by production and reference consumers; production runs in all three lifecycle contexts.
- Nonempty public build/check/run checks cover actual compiled programs, exact sites/frames, declared charge larger than body length, instrumented cache hits, numerical refusal and atomic resource-refusal prefixes.
- Real before/after capture: **26 CLI subprocess calls**, all **28 authority files** and previously valid semantic artifacts byte-identical. Only nine truthful evaluator manifests and their dependent publication/receipt identities change. Confirmed invalid baseline resource refusals are separately tested as repairs.
- Inventory: **1,741 tests**, nine disjoint shards, all **849** required baseline IDs and **313** current package vectors retained; no missing or overlapping tests.
- Ruff, Pyright and sealed-LDB checks pass. Full CI passed at `075a5aae74d257f8968c408faa4b4b304055f85b`: 17 successful checks and one intentionally skipped root Godot job. All nine package JUnit sets account for 1,741 tests: 1,718 passed, 23 existing allowed skips, zero failures or errors; actual executed IDs equal the complete collection and each shard digest. Independent review completion is recorded in the final acceptance comment.

The [Formula Runtime record](libs/gda-balancing/docs/refactor/current-language/FORMULA-RUNTIME.md) defines scope, source fingerprints, evidence limits, reproduction tests and rollback. Independent reversal of the complete issue diff restored all 436 package files and modes exactly to `aa94ec79d63b0595e5d6af8bce6180aba9d5b397`. Actual restored public build/check/run with explicit `--rir` passed; expected event order and final metrics were preserved. This rollback also restores the baseline resource-refusal defect.

Refs #612. This PR targets `codex/gda-balancing-refactor-delete-dev` under #865. It does not merge accumulated work into main, publish a formal release, or perform the primitive deletions owned by #876.

### DIFF
--- a/libs/gda-balancing/docs/refactor/current-language/FORMULA-RUNTIME.md
+++ b/libs/gda-balancing/docs/refactor/current-language/FORMULA-RUNTIME.md
@@ -36,8 +36,25 @@ moving the shared Event helpers would add relocation without changing ownership.
 
 Inline Event Formula instructions retain their existing per-instruction total,
 Event and Operation counters and exact failure index. Extracting the program
-function does not replace this ledger with a bulk charge or change the published
-refusal counters, committed prefix or rollback boundary.
+function does not replace this ledger with a bulk charge. The committed prefix
+and rollback boundary remain unchanged.
+
+## Confirmed fault-charge correction
+
+Public runs with admitted Runtime profile node limits of 5 and 20 expose a
+pre-existing failure in the Event and observation Formula fault paths. Both the
+development baseline and the initial extraction return an internal error when
+publication rejects the prepared terminal audit. The call assigning
+`total_steps = _evaluate_initialization_programs(...)` never receives a result
+when the function raises, so the callers report their pre-call counter. The
+independent artifact validator correctly requires the attempted Formula charge.
+
+The shared fault now carries consumed steps. Both callers must retain that value
+before constructing the refusal audit. This repairs conformance to the existing
+charge and audit laws; it does not weaken the validator or change the language,
+resource policy or public refusal schema. Previously valid artifacts retain their
+behavior. The affected runs now publish the required typed refusal instead of an
+internal publication failure. Permanent public tests cover both cases.
 
 ## Consumers and refusal scope
 
@@ -78,7 +95,9 @@ general-purpose program admission implementation.
    domain architecture independently before integration into dev.
 
 The byte comparison preserves Kernel/LDB identities, RIR meaning, admitted Formula
-semantics and semantic execution artifacts. Under the
+semantics and previously valid semantic execution artifacts. The two confirmed
+fault-charge failures above require a separate before/after result, since the
+baseline could not publish a valid refusal artifact. Under the
 [execution identity contract](EXECUTION-IDENTITY.md), source changes must update
 the evaluator implementation fingerprint. Producer manifests and publication
 records that identify their bytes therefore change truthfully. Enumerate those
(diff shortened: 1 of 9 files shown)


qinhaihong-red:
Verified acceptance at **075a5aae74d257f8968c408faa4b4b304055f85b**, base/merge-base **aa94ec79d63b0595e5d6af8bce6180aba9d5b397**.

| Independent review | Findings | Independent verification |
| --- | --- | --- |
| Standards (`review`) | 0 | 25 targeted tests; all 151 existing test definitions/parameters/decorators retained; moved reference AST unchanged; removing the two charge assignments restores both public internal failures. |
| Spec (`review`) | 0 | 25 targeted tests; all ten vector expectations separately checked in the reference and all three production contexts; fresh baseline/candidate subprocess build/check/run and exact member comparison; eight forged-RIR public inputs refused, including recomputed outer identities. |
| Domain architecture | 0 | 35 tests; shared Runtime ownership and independent reference/replay preserved; both public charge regressions verified counterfactually; raw byte capture and exact Domain source trees independently checked. |

The investigated warm-cache counterexample required replacing an admitted program or selected contract while retaining its identity. Public admission canonically detaches and rederives these inputs, and the cache is invocation-local. Actual public mutation probes refuse. This does not justify a second general-purpose Runtime admission/hash pass; no new fallback or validation mechanism was added.

[Complete CI](https://github.com/aigengame/godot-agent/actions/runs/34104169094) passed: **17 successful checks**, plus the intentionally skipped root Godot E2E job. All nine package JUnit sets were downloaded and parsed: **1,741 total, 1,718 passed, 23 existing allowed skips, zero failures/errors**. Every executed test ID equals the full collection and each shard digest; no duplicates, missing tests, new skips or xfails. All 849 required baseline IDs and 313 current package vectors remain. Wheel/subprocess smoke and both-product co-install passed.

The independent full reverse patch restores every **436** baseline package file, all **8,703,880 bytes** and file modes. Actual restored public build/check/run with explicit `--rir` passed and produced applied/ticked/ticked/expired, health 70 and inactive final Effect. The public run used the earlier test-head reversal; its restored tree and the final-head reversal were independently proved identical to the complete same baseline Git tree. Rollback also restores the original low-budget defect.

The 26-call byte capture compared 184 raw files: 157 byte-identical and 27 changed only at 54 enumerated producer/publication identity leaves. No broad normalization or stale publication reuse was used. Event limit 5 and observation limit 20 are separately evidenced bug repairs: exit 4 with invalid audits at 3/18 steps becomes exit 2 with valid audits at 6/21 steps.

All #612 ACs are satisfied for reviewed integration into `codex/gda-balancing-refactor-delete-dev`. This is not integration into main, a formal release, #876 primitive deletion, or completion of #865.

PR #937, opened 2026-09-08
feat(gda): resource import says why an asset is invalid or failed (#853)

### LINKED ISSUE #853: resource import: per-asset reason on failed and invalid
## Dogfooding source

- **PIPE-DF-191** (2026-09-04, gda 0.13.0) — a second forced-cache `resource import` reported the addressed asset `status: failed` with no diagnostics; the caller could not tell an importer error from a stale-cache decision and had to prove the outcome from unchanged resource bytes and a later passing test.
- **GDA-DF-064** (2026-08-26, gda 0.11.0) — the same command's project-wide pass and its residue. The pass being project-wide is the recorded #668 decision (the engine's one import primitive) and stays; what remains from that entry is the per-asset explanation.

Verified against gda 0.13.0 (`266c8009f`): `ResourceImportAsset` carries `path`, `status`, `sidecar`, and `dest_files`; the `invalid` and `failed` verdicts have no reason field, although the `Import evidence` adapter (`gda/import_evidence.py` since #741) knows which check failed: an unparsable sidecar, `valid=false` in the sidecar, or a receipt outside the documented subset produce `invalid`; a destination still absent after the pass produces `failed`. (A missing sidecar is `missing`, not `invalid`, and the pass then settles it.) `invalid` is a pre-pass evidence state reported under `--dry-run`; a real run settles every `invalid` request to `failed` without a pass — which is why the dogfooded `failed` carried no explanation.

## What to build

- Each asset with `status` `invalid` or `failed` carries `reason` (a small closed enum the adapter already decides, e.g. `sidecar_unparsable`, `sidecar_marked_invalid`, `receipt_unsupported`, `dest_missing_after_pass`) and `detail` (the offending path or line, when known). The reason SURVIVES the invalid-to-failed settlement: a real run's `failed` names the pre-pass check that decided it.
- For a `failed` that went through the pass, the pass's stderr lines that name the asset are attached as `engine_output`: at most 20 lines, and `engine_output_truncated: true` when more matched (the same bounded-stream rule as #665, without a spill file).
- Human rendering shows the reason on the asset's line.

## Acceptance criteria

- [ ] `--dry-run` on a fixture with an unparsable `.import` sidecar reports `invalid` with `reason: sidecar_unparsable` and the sidecar path; the same fixture on a real run reports `failed` carrying the same reason.
- [ ] A fixture whose import genuinely fails in the pass reports `failed` with `reason: dest_missing_after_pass` and the engine's lines for that asset in `engine_output`.
- [ ] The schema publishes the enum and the cap; existing fields are unchanged.
- [ ] Coverage in the adapter's unit tests (in #741's module if it has landed) and one e2e for the `failed` case.

## Blocked by

None - can start immediately. Coordinate with #741 if it lands first: the reason is decided in the evidence adapter it moves.

qinhaihong-red:
## Summary

`gda resource import` reported `failed` with nothing behind it, so a caller
could not tell an importer error from a stale-cache decision and had to prove
the outcome from unchanged resource bytes (PIPE-DF-191). Each `invalid` or
`failed` asset now says which check decided it, and a `failed` the engine pass
ran over carries that pass's own lines for the asset.

Placement follows the issue's 2026-09-08 note. `reason` / `detail` are decided
in the `Import evidence` adapter (`gda/import_evidence.py`) beside the state it
already reads; `engine_output` is the pass's stderr and stays command-side in
`commands/resource.py`'s settlement, because the lifted module is the pre-pass
read and never sees the pass. The commands -> core direction is unchanged:
`import_evidence.py` still imports nothing from `gda`.

## Public behavior

Three new keys on each entry of `assets`, present on every asset and empty
where there is nothing to explain:

- **`reason`** — which check decided an `invalid` or `failed` verdict; `null`
  on every other status.
  - `sidecar_unparsable` — the `.import` bytes do not decode, or its
    `dest_files=` / `files=` list does not parse.
  - `sidecar_marked_invalid` — the ENGINE marked the last import `valid=false`.
  - `receipt_unsupported` — the `.md5` receipt falls outside gda's documented
    engine-written assignment subset.
  - `dest_missing_after_pass` — no artifact check refused the asset, the pass
    ran, and it is still not cached.
- **`detail`** — the offending line or path behind `reason`, when the check
  knows one the record does not already carry: the malformed `dest_files=` /
  `files=` line, or the derived `.md5` receipt's `res://` path (which appears
  nowhere else on the result). `null` otherwise.
- **`engine_output`** / **`engine_output_truncated`** — the import pass's
  stderr lines that name the asset's `res://` path, verbatim and in order, for
  a `failed` asset whenever THIS request ran a pass; at most 20, with the flag
  set when more matched (#665's bounded-stream rule, without its spill file).
  Empty on every
  other status and when no pass ran. It is **independent of `reason`**, and the
  two must not be made to gate each other: `reason` / `detail` are gda's
  PRE-PASS evidence, `engine_output` is what the ENGINE said about the asset
  during this request's pass. The engine names an asset it then skips, so an
  asset carrying an artifact reason can carry lines too (round 2, P2-1 —
  see Backtrace).

The three ARTIFACT reasons **survive the settlement**: an `invalid` request
settles to `failed` without a pass, and the settled record still names the
pre-pass check that refused it. `dest_missing_after_pass` is assigned only
where no such check refused the asset, so a batch that spends a pass for one
asset does not overwrite another asset's pre-pass reason.

Human rendering shows the reason (and its detail) on the asset's line:

```
resource dry run: 1 asset(s), no pass needed (cache root res://.godot)
         invalid  res://icon.png  (receipt_unsupported: res://.godot/imported/icon.png-….md5)
```

Two shape decisions the issue leaves open, and why:

- **Nullable, not omitted.** The issue is silent; this result's existing
  pattern is nullable-and-always-present (`sidecar` is emitted as `null`), so
  the four new keys follow it rather than the `Failure evidence`
  omitted-rather-than-null rule, which governs the error half.
- **`engine_output` names the asset as the engine delimits a path.** gda knows the
  path it is looking for, so it never cuts an unknown path out of the line: a line
  counts when the asset's `res://` path (or its sidecar, `.import` / `.import.md5`,
  each optionally `:<line>`) is spelled in it whole and delimited as the engine
  delimits a path — a quoted path closes with the SAME quote character whatever it
  holds in between; a bare one ends at whitespace, the end of the line, a
  sentence-final period or a `:<line>` (all measured on the 4.6 sources). Never a
  substring, a prefix-plus-lookahead, or a token cut by a character class: a legal
  neighbouring path (`res://icon.png copy`, `res://icon.png]backup`,
  `res://icon.png.import-backup`) can carry any punctuation inside it, and a legal
  asset path can carry a quote (`res://icon"hero.png`). A broader needle (the
  filename, the filesystem path) would attribute a neighbour's error to this asset;
  the engine's `at:` continuation lines name a source file, not the asset, so they
  stay out. Evidence that over-claims is worse than evidence that is short. Declared
  residual: a BARE path containing whitespace, and a quoted neighbour whose path
  continues after a quote character identical to its delimiter — the engine's own
  quoting cannot distinguish either from this asset's line.

## Surface diff

Captured against a detached checkout of the base (`b7c485693`), with
`gda schema` and `gda resource import --help` run in both. Every difference is
enumerated below; the rest of the 1.1 MB manifest is byte-identical.

**`gda schema` — 23 added leaves, 0 removed, 2 changed.** All additions are
under `commands[…"resource-import"].output.$defs.ResourceImportAsset.properties`:

| Added property | Shape |
| --- | --- |
| `reason` | `anyOf: [{type: string, enum: [sidecar_unparsable, sidecar_marked_invalid, receipt_unsupported, dest_missing_after_pass]}, {type: null}]`, `default: null`, title + description |
| `detail` | `anyOf: [{type: string}, {type: null}]`, `default: null`, title + description |
| `engine_output` | `type: array`, `items: {type: string}`, **`maxItems: 20`**, title + description |
| `engine_output_truncated` | `type: boolean`, `default: false`, title + description |

The 2 changed leaves are prose only: `ResourceImportAsset.description` (the
model docstring gains one paragraph naming the three keys) and the
`resource-import` command `description` (the same sentence as the `--help`
below). No existing property, enum, required list, or `argv` binding changes;
no command, recipe, or `kind` changes.

**`gda resource import --help` — one added sentence**, inside the existing
paragraph (3 wrapped lines replaced by 7):

```
sidecars). An `invalid` or `failed` asset says why: `reason` names the
check that decided it (an invalid one survives into the settled `failed`),
`detail` the offending line or path when there is one, and `engine_output`
the pass's own stderr lines for that asset (at most 20, with
`engine_output_truncated` when more matched). `--dry-run` reports the states
and the decidable predictions (including `pass_will_also_import`) and writes
nothing. Plain `gda script run` never triggers an import pass. The pass
```

**Other contract surfaces carried by this slice:**

- `src/gda/skill/SKILL.md` — the `resource` row gains the reason vocabulary and
  what each reason tells the agent to do next.
- `docs/command-catalog.md` — the "Scoped import surface" section gains one
  paragraph on the reasons, their survival, and the `engine_output` bound.
- `CONTEXT.md` `Import evidence` — amended: the adapter's `invalid` verdict now
  also names the check that decided it, and the settlement clause names the one
  reason the command decides alone.
- `README.md` — **unchanged**. The `resource import` row's one-line contract
  ("Ensure assets are imported into the project cache (clean-worktree
  loading)") does not change, so no row edit and no i18n run.

## Validation

Commands, all re-run on the committed head `a751d80a9`:

| Gate | Command | Result |
| --- | --- | --- |
| Fast tier | `uv run --frozen pytest -m "not e2e" -q -p no:cacheprovider` | **2645 passed**, 686 deselected |
| Format | `uv run --frozen ruff format --check .` | 515 files clean |
| Targeted e2e | `… e2e-lock.py uv run --frozen pytest tests/resource tests/project/test_project_walk.py -m e2e -q -p no:cacheprovider` | **42 passed**, 117 deselected, real Godot 4.6.3 |
| Surface diff | `gda schema` + `gda resource import --help`, here and in the base checkout `/private/tmp/pr937-base-b7c485693` (`b7c485693`, read-only, `uv run --frozen --no-sync`) | 23 added leaves, 0 removed, 2 changed (enumerated above); 17965 -> 17988 manifest leaves. Unchanged across both rounds — only the two description strings moved |

`tests/resource` + `tests/project/test_project_walk.py` collect **159** tests,
up from 138: 8 adapter-seam, 11 command, 2 e2e.

**Red then green.** The 7 adapter tests were written first and failed with
`AttributeError: 'AssetEvidence' object has no attribute 'reason'`; the 10
command-side assertions (9 new tests plus the wire-shape pin on the existing
dry-run ABI test) failed with `KeyError: 'reason'` /
`KeyError: 'engine_output'` / `ImportError: cannot import name 'AssetReason'`
before the command change.

**Coverage.** One adapter test per reason the adapter can decide
(`sidecar_unparsable` in both its spellings — undecodable bytes and a malformed
list line — `sidecar_marked_invalid`, `receipt_unsupported` in both its
spellings), plus one that pins `cached` / `missing` / `stale` as reasonless.
Command-side: the survival pair (dry-run `invalid` -> settled `failed`, no pass
spent), the settlement reason with the engine's lines, the 20-line bound, the
exactly-20 boundary, `imported` and `not_importable` carrying no reason, the
dry-run mode invariant, the human render, and a guard that the wire enum covers
every adapter reason plus exactly one of its own.

**Mutants** — each was applied, the suite run, and the mutation reverted:

| Mutation | Caught by |
| --- | --- |
| drop `reason="sidecar_marked_invalid"` | `test_a_valid_false_sidecar_names_the_engines_own_verdict` |
| drop `reason="receipt_unsupported"` | 3 tests, adapter + render |
| settlement ignores the surviving reason (`asset.reason or …` -> constant) | `test_an_invalid_reason_survives_the_settlement_unchanged` |
| truncation boundary `>` -> `>=` | `test_exactly_twenty_matching_lines_are_not_reported_truncated` |
| drop `reason="sidecar_unparsable"` (decode branch) | 2 tests, adapter + the survival pair |
| `engine_output` unbounded (`matched[:CAP]` -> `matched`) | `test_engine_output_is_bounded_to_twenty_lines` |
| carried `detail` -> `None` | `test_an_invalid_reason_survives_the_settlement_unchanged` |
| **round 2**: gate `engine_output` on the reason again (round 1's shape) | `test_each_asset_takes_only_the_passs_lines_that_name_it` |
| **round 2**: drop the `AssetEvidence` invariant | `test_an_invalid_verdict_without_a_reason_is_refused_at_construction` |
| **round 2**: remedy footnote back on the `invalid` STATE | `test_the_remedy_footnote_follows_the_artifact_reason_into_a_real_run` |

PR CI runs **no** Godot e2e (the `godot-e2e` workflow is schedule/dispatch
only), so what gates this change on the PR is:

- the 2645-test fast tier, which covers everything here except the engine's own
  words: the adapter's five reason branches (plus its `invalid`-needs-a-reason
  invariant) against real temp trees, and the whole command surface through the
  launch seam — a fake pass whose stderr and file effects are scripted, which is
  how the 20-line bound, the truncation flag, the per-asset filter, the reason's
  survival, and the remedy footnote are exercised. The fake stderr is shaped
  from output reproduced against the real engine, which is the round-2 lesson:
  a scripted stream is only as true as the run it was copied from;
- `ruff check`, `ruff format --check`, `pyright`;
- the schema-shape guards already in `tests/cli/test_command_descriptor_registry.py`
  (unchanged — this slice adds no command and no recipe).

What needs the real engine, and ran locally under the machine-wide lock:

- `test_a_failed_asset_names_its_reason_and_the_engines_own_lines` — the AC's
  second half. Only a real pass proves that Godot exits **0** while leaving an
  asset unimported, that it prints `ERROR: Error importing 'res://broken.png'.`
  on **stderr** (not stdout, where its progress lines go), and that the second
  run's artifacts (`valid=false`, written by that pass) make the pre-pass check
  decide instead — the survival direction, on artifacts no fixture wrote.
- `test_the_engine_names_a_sidecar_it_skips_and_those_lines_ride_along`
  (round 2) — the assumption round 1 got wrong. No fake can establish that the
  engine emits `ResourceFormatImporter::load - 'res://bad.png.import:8'` twice
  while deciding to SKIP an asset; that is exactly the fact a scripted stream
  would have kept inventing.

## Review round 1

All six findings adopted, in `00007491f`.

| # | Finding | Resolution |
| --- | --- | --- |
| P2-1 | `engine_output` attached to every settled `failed`, contradicting its own description; the mutant survived | **SUPERSEDED by round 2 — see Backtrace below.** The narrowing shipped in `00007491f` and was reverted in `1d90f7193`: its premise (the engine says nothing about an asset it skips) is false, so the description was the thing that had to move, not the behaviour. |
| P2-2 | the carried `detail` had no assertion; mutating it to `None` survived | `test_an_invalid_reason_survives_the_settlement_unchanged` now uses a decodable sidecar with `dest_files=[oops]` and asserts `detail == "dest_files=[oops]"` on both halves. The mutant now fails. |
| P3-1 | for some importers the filtered lines only restate the verdict | One clause added to the SKILL `resource import` row: the lines are the ones NAMING the asset, so keep the whole engine stream with `gda --user-data-root DIR resource import …` and read `DIR/logs/godot.log`. |
| P3-2 | the cap docstring claimed the whole stream "stays where it always was — the engine's log" | Reworded: "the engine's log when `--user-data-root` keeps it, and the `operation_failed` diagnostics of a pass that exited non-zero". Without a root the log is a private temporary file removed on exit (`runner.py`). |
| P3-3 | the `reason`/`detail` `None` resets on the `imported` and `not_importable` branches are dead | Deleted. The adapter reports a reason only on `invalid`, and the two existing tests that assert `reason is None` there now pin that guarantee rather than a redundant overwrite. |
| P3-4 | mixed optional spelling; the cap duplicated in prose; the dry-run invariant ignored the flag | `reason: AssetReason | None` and `detail: str | None` (as `sidecar` is spelled); the description derives "at most 20" from `ENGINE_OUTPUT_LINE_CAP`; the dry-run invariant now refuses `engine_output_truncated=True` as well as a non-empty `engine_output`. The schema is byte-identical through all three. |

Nothing was declined.

### Backtrace

Round 1's P2-1 disposition is **superseded**, not refined. It narrowed
`engine_output` to a settlement-decided reason on an assumption nobody had
checked against the engine — that the engine says nothing about an asset it
skips. Round 2 disproved it on Godot 4.6.3, and it reproduces here: an
unparsable sidecar draws two
`ERROR: ResourceFormatImporter::load - 'res://bad.png.import:8' error …` lines
naming the asset before the skip (from `_test_for_reimport` and
`_get_import_dest_paths` in `editor/file_system/editor_file_system.cpp`), and a
receipt outside gda's narrower subset is re-imported by the engine, which then
fails in the open. The batch test round 1 added encoded the false assumption
with a stderr line the engine never emits, so it could not catch it. The
recovery is not a third variant but the literal rule #853 states, with no
reason-gating at all: two orthogonal facts on one record, neither gating the
other.

| # | Finding | Resolution |
| --- | --- | --- |
| P2-1 | round 1's narrowing rests on a false engine-behaviour assumption | `engine_output` computed for every settled `failed` whenever this request ran a pass; `reason` / `detail` come from the carried evidence or `dest_missing_after_pass`, independently. The round-1 comment claiming the engine never names a skipped asset is deleted. Field description, `docs/command-catalog.md` and the SKILL row state the literal rule; no "the pass ran over [the asset]" wording is left, including in `--help`. |
| P2-1 (tests) | the round-1 batch test encoded the false assumption | Replaced by `test_each_asset_takes_only_the_passs_lines_that_name_it`: a `sidecar_unparsable` asset beside a `missing` sibling, fake stderr in the engine's real shape (both `ResourceFormatImporter::load` lines with their `at:` continuations, plus the sibling's `Error importing 'res://fresh.png'.`) — the skipped asset carries exactly its two lines, the sibling exactly its own, nothing crosses; then a no-pass run leaves both empty. |
| P2-1 (e2e) | the engine behaviour was never pinned | New real-engine e2e `test_the_engine_names_a_sidecar_it_skips_and_those_lines_ride_along`: an unparsable sidecar beside a missing sibling; the `sidecar_unparsable` asset's `engine_output` names it (asserted on the `.import:` substring, not the whole line). The pass-failed e2e is kept. |
| P3-1 | an `invalid` verdict could be built with no reason | `AssetEvidence.__post_init__` raises `ValueError` for `status == "invalid"` with `reason is None` — an internal invariant of gda's own dataclass, with one test. |
| P3-2 | the "delete the .import sidecar" remedy was keyed on the `invalid` STATE, which a real run settles away | The footnote now fires when any asset carries an artifact reason (`sidecar_unparsable` / `sidecar_marked_invalid` / `receipt_unsupported`), so a settled `failed` gets the hint; `dest_missing_after_pass` still gets none, because no sidecar edit helps an import the engine attempted. One test covers both directions. |
| P3-3 | `detail` is unbounded | Recorded, not changed — it is one sidecar line or one receipt path. |

## Review round 3 (independent, base to head `00007491f`)

CHANGES REQUIRED. The matcher was a bare substring test, so a prefix neighbour (`res://icon.png2`, `res://icon.png.backup`) was attributed to `res://icon.png`. Handled at `7137e0ef1` with a lookahead over the punctuation that may follow the path — superseded in round 4.

## Review round 4 (recheck at `7137e0ef1`) — handled by backtrace at `e90979849`

The lookahead still admitted `'res://icon.png.import-backup'`, `'res://icon.png copy'`, `'res://icon.png]backup'`. Fourth change to one matcher: the round was traced with `/backtrace-review` before any code moved. Root cause: every round refined what may FOLLOW the asset's name while nothing defined where a path token begins and ends in the engine's own output; a path's alphabet holds every character a lookahead can stop at. Two concerns were fused in one regex — cutting the line into path tokens (the engine's grammar, measured on the 4.6 sources) and comparing a token to the asset (equality) — and are now separate (`_ENGINE_PATH_TOKEN` + a whole-token equality against the path or its sidecar family).

| whole-token claim not closed (three quoted neighbours) | Fully Adopt | reproduced | tokenizer by the engine's quoting and grammar; equality of the whole token | the three cases are unit neighbours; the round-3 regex as a mutant fails the test at index 0 |
| recognize the complete path within the engine grammar; `.import:<line>` as a complete suffix | Fully Adopt | grammar measured in `editor_file_system.cpp`, `resource_importer.cpp`, `resource_loader.cpp`, `resource_format_binary.cpp`, `image_loader.cpp` | the unit fixture's own lines are every measured form with the path where its source puts it | unit 112 passed |
| add the cases and a neighbour flood | Fully Adopt | — | 25 quoted neighbours beside the own lines → own lines only, not truncated; real-engine e2e `test_a_neighbours_lines_are_never_the_assets_own` (four broken assets whose paths extend each other, one pass, each keeps its own lines) | e2e 43 passed under the lock; red on the round-3 code |

## Review round 5 (recheck at `e90979849`) — handled at `a38aec4b6`

The tokenizer's `[^'"]*` cut a quoted path at EITHER quote, so `res://icon"hero.png` — which the real engine names inside single quotes — reported no evidence. Re-derived rather than widened: four devices (substring, lookahead, tokenizer, character class) had each tried to find where an UNKNOWN path ends, and a path's alphabet defeats all of them; gda KNOWS the path, so `_names_asset` finds that spelling and checks the two boundaries the engine's delimiting fixes. Unit cases for the other quote, the same quote and both (the round-4 tokenizer as a mutant fails all three); `icon"hero.png` joins the real-engine neighbour e2e. Residual declared in the code comment.

## Open questions

1. The human render shows `reason` and `detail` but not `engine_output`; the
   issue's rendering bullet asks only for the reason. A human on a `failed`
   asset must re-run with `--json` to read the engine's lines. Worth a
   follow-up if that reads as a hole in review. (The adjacent question — the
   lines being thin for some importers — is answered in the SKILL and the
   catalog by pointing at `--user-data-root`, not by widening the filter, which
   would attribute a neighbour's error to the asset.)
2. `dest_missing_after_pass` reads "the destination is missing" but is assigned
   for any post-pass verdict that is not `cached`, including a pass that wrote
   `valid=false` over a previously stale sidecar. The name is the issue's; the
   description says exactly what it covers.
3. `detail` is unbounded (round 2, P3-3). It is one sidecar line or one receipt
   path today, so no cap was added; a sidecar with a very long `dest_files=`
   list would make it large.

Closes #853


qinhaihong-red:
**Review record, round 1 (orchestrator, 2026-09-09).** Independent adversarial review at `4284e2cd0` (four axes; real-engine probes on Godot 4.6.3 incl. a genuinely failing Ogg and PNG pass; 10 mutants beyond the PR's six) returned **CHANGES REQUIRED — 0 P1 / 2 P2 / 4 P3**. Both P2s re-verified by the orchestrator on the head before disposition; the fix round is the implementer's. Fix head: `00007491f`.

|---|---|---|
| **P2-1** `engine_output` was attached to EVERY settled `failed`, so a pre-pass `invalid` asset (which the engine skips) collected stderr lines when a sibling spent the pass — against the field's own description; the narrowing mutant survived the suite | Confirmed (`resource.py` settlement called `_engine_output` unconditionally) | **Adopted — behaviour narrowed to the published description:** lines ride only where the settlement decided the reason (`dest_missing_after_pass`); a carried-through pre-pass reason settles with `[]` / `false`. New batch test (`invalid` beside a pass-spending `missing` sibling, stderr naming both) kills the mutant. Descriptions unchanged. |
| **P2-2** `detail` carried through the settlement had no assertion (`detail → None` mutant survived) | Confirmed | **Adopted.** The survival test uses a decodable sidecar with `dest_files=[oops]` and asserts `detail` on both the dry-run and the settled half. |
| P3-1 for a class of importers the filtered lines only restate the verdict (Ogg's root-cause line carries no asset path) | Confirmed on the engine | **Adopted as disclosure, not a wider filter** (a broader needle would attribute a neighbour's error — the failure mode P2-1 closed): the SKILL row says the whole engine stream is kept with `--user-data-root DIR` (`DIR/logs/godot.log`). |
| P3-2 the cap docstring claimed the full stream "stays in the engine's log", but without `--user-data-root` that log is a private temporary file removed on exit | Confirmed (`runner.py:361`) | **Adopted** — reworded. |
| P3-3 two dead `reason/detail = None` updates on the `imported` / `not_importable` branches | Confirmed (mutants deleting them survived) | **Adopted — deleted**; the two tests asserting `reason is None` there now pin the adapter's guarantee. |
| P3-4 one optional spelling in the model; "at most 20" hard-coded beside `ENGINE_OUTPUT_LINE_CAP`; dry-run invariant did not refuse `engine_output_truncated=True` | Confirmed | **Adopted** (all three; the schema renders byte-identically). |

Reviewer-verified and unchanged: schema 23 added / 0 removed / 2 prose leaves, all under `ResourceImportAsset.properties`, `required` unchanged; the help sentence; fast tier 2642, e2e 41, 155 collected; 17 red-then-green failures reproduced by swapping the base sources back in; stdlib-only adapter; the reason enum covers all five `invalid` exits; the `elif` simplification proven equivalent; decision (b) — the settled reason comes from the PRE-pass record, so a `stale` sidecar the pass rewrites to `valid=false` reports `dest_missing_after_pass` with the engine's lines and the next dry-run reports `sidecar_marked_invalid` — judged the right truth; the reason-enum duplication plus its guard test judged proportionate (a `Union[…, Literal]` would publish a three-branch `anyOf`). On `00007491f`: fast tier 2643, e2e 41 under the lock, 9 mutants killed, PR CI green. A second independent round runs on this head per the wave rule.

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-09).** Second independent review at `00007491f` (real-engine probes on Godot 4.6.3 with the full engine log kept under `--user-data-root`; engine source `editor/file_system/editor_file_system.cpp` read for the skip path; both round-1 mutants re-run) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 3 P3**. Because the P2 reversed round 1's disposition, the orchestrator ran a **backtrace review** before any further change (the wave's convergence rule after PR #899). Fix head: `a751d80a9`, verified directly by the orchestrator (no third review round — the budget is two).

**Backtrace.** Round 1's narrowing of `engine_output` to the settlement-decided reason was the orchestrator's choice of the reviewer's second option, made on the implementer's code comment ("the pass skips a refused asset, so a line naming it never belongs to it") and pinned by a batch test whose fake stderr line the engine never emits — an unverified external-behaviour assumption, with a mock as its only evidence. Round 2 measured the engine: for a `sidecar_unparsable` asset it prints two `ResourceFormatImporter::load - 'res://x.import:N' error …` lines naming the asset BEFORE skipping it (ERR_PRINT, then `return false`; `_test_for_reimport` / `_get_import_dest_paths`), and it re-imports a `receipt_unsupported` asset outright (gda's receipt subset is narrower than the engine's parser). The narrowing therefore dropped the most informative evidence, and the field's description became false in the other direction. Root cause: a prose contract with two readings ("the pass ran over [the asset | the batch]") resolved by assumption instead of measurement. Recovery: the one literal rule the issue itself states — `engine_output` is the pass's stderr lines naming the asset whenever THIS request ran a pass, empty when none ran, independent of `reason` (gda's pre-pass evidence and the engine's own words are two facts with two owners; neither gates the other). Round 1's P2-1 disposition is superseded; its finding (behaviour ≠ description) stands and is closed by the description becoming literal.

|---|---|---|
| **P2-1** the round-1 narrowing dropped lines the engine prints about an asset it skips; the description ("for a `failed` the pass ran over … empty on a `failed` no pass was spent on") was false the other way; the round-1 test pinned a fabricated engine behaviour | Confirmed on the engine and in its source | **Adopted (the backtrace's recovery).** Lines follow the pass, not the reason; the false comment deleted; field / class / `_engine_output` docstrings, catalog, SKILL and `--help` state the literal rule; the synthetic batch test replaced by one shaped after the engine's real output (both `.import:8` lines with their `at:` continuations for the skipped asset, the sibling's `Error importing` line for the other, a no-pass run empty for both); one new real-engine e2e pins the skipped-sidecar lines. |
| P3-1 an adapter `invalid` exit without a reason would settle as `dest_missing_after_pass` | Confirmed (all five exits carry one today) | **Adopted** — `AssetEvidence.__post_init__` raises on `invalid` without a reason, with a test. |
| P3-2 the human remedy footnote appeared only on dry runs (real runs settle `invalid` into `failed`) | Confirmed | **Adopted** — the footnote keys on the artifact reasons, so a settled `failed` gets the "delete the sidecar" hint and `dest_missing_after_pass` does not. |
| P3-3 `detail` is unbounded (one sidecar line or one receipt path today) | Noted | **Recorded** as open question 3; no cap. |

Orchestrator verification on `a751d80a9` (detached checkout): no "ran over" wording anywhere; the literal rule in the settlement comment and the field description; `import_evidence.py` imports stdlib only; resource fast tier 117 passed; the re-gating mutant (lines only when `asset.reason is None`) fails the new batch test; resource + project-walk e2e **42 passed** under the machine lock on Godot 4.6.3. Implementer's gates: fast tier 2645, ruff / format / pyright clean, 7 mutants killed, schema diff still 23 added / 0 removed / 2 prose, PR CI green. Reviewer-verified in round 2 and unchanged: both round-1 mutants killed; nested paths, spaces and unicode in asset paths filter correctly; `at:` continuations carry no path; the engine's ERROR text is locale-independent; argv ↔ `--params-json` byte-identical; the MCP output schema inherits the 23 leaves; the 20-line cap and the dry-run invariant behave at 20 / 21. **READY** for the maintainer's review; merges second in the W3 order (after #939).

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `a751d80a9f282570d5cfda19349bb984e7708680`. Existing reviews and replies were not used as evidence.

The changed public prose says that `engine_output` contains the pass's stderr lines that name this asset and explicitly distinguishes it from a neighbour's output (`src/gda/commands/resource.py:622-631,922-934`). The implementation uses an unbounded substring test at line 938, so that prose is false for prefix-neighbour paths. Update the implementation and keep the current wording, rather than weakening the per-asset contract.

**Terminology and prose assessment:** I inspected issue #853, the PR body, the changed `CONTEXT.md`, command catalog, skill, help/schema descriptions, and their code-backed claims against `CONTEXT.md`, ADR-0004/0015/0040, and ASD-STE100. `reason`, `detail`, and `engine_output` are distinct and consistently owned. No other substantive terminology or prose issue was found. I did not validate every engine error spelling across Godot versions.

## Architecture review — Pass

The responsibility split is sound: `gda.import_evidence` owns pre-pass artifact evidence, while `gda.commands.resource` owns post-pass settlement and pass stderr. Dependencies remain one-way and no new abstraction is needed. The matcher below is a local correctness defect, not a reason to redraw modules.

## Standards

- Hard: `_engine_output()` at `src/gda/commands/resource.py:922-939` promises asset-specific evidence but line 938 uses `if res_path in line`. For requested `res://icon.png`, a line naming only `res://icon.png.backup` is published for the wrong asset. This violates `RULES.md:86`. Match a delimited engine path while retaining the same asset's `.import:<line>` form, and add a prefix-collision regression.

No additional baseline smell survived review.

Standards summary: 1 hard finding; worst is wrong diagnostic evidence.

## Spec

- Required: issue #853 says “the pass's stderr lines that name the asset are attached as `engine_output`.” The exact-head probe reproduced the violation:

  ```text
  _engine_output("ERROR: Error importing 'res://icon.png2'.\n", "res://icon.png")
  => (["ERROR: Error importing 'res://icon.png2'."], False)
  ```

  Both are legal asset names, and the false match can also consume the 20-line cap. Existing tests use only non-prefix siblings. Parse or anchor the asset token and add this counterexample.

No other missing requirement or scope creep was found.

Spec summary: 1 finding; worst is incorrect per-asset attribution.

qinhaihong-red:
## Third review handled at `7137e0ef1`

The Artifact, Standards and Spec findings are one claim — the matcher's token boundary — and it is
the third change to `_engine_output`, so it was traced before it was patched. History: round 1
gated the evidence on a settlement-decided reason (assumption falsified on the engine); round 2
made the rule literal — the lines that NAME the asset — with a bare substring test; the third
review shows the substring over-claims for a prefix neighbour. Each change fixed a distinct,
measured defect; the thing never defined was the lexical boundary of "names the asset". Root
cause: an undefined token, not the rule. The design (pre-pass evidence in `gda.import_evidence`,
post-pass settlement and pass stderr in the command) is kept, as the Architecture axis found.

| `if res_path in line` attributes `res://icon.png2` / `res://icon.png.backup` lines to `res://icon.png` and can spend its 20-line cap on them | Fully Adopt | reproduced: `_engine_output("ERROR: Error importing 'res://icon.png2'.\n", "res://icon.png")` matched | The asset is matched as a whole token: its path followed by what the engine puts after a path — a closing quote, its own `.import:<line>` sidecar, a colon, a comma, a bracket, whitespace or the end of the line — never another path character. Wording kept | new regression `test_engine_output_names_the_asset_as_a_whole_token` (prefix neighbours excluded, three real engine spellings included, a 25-line neighbour flood not counted against the cap); reverting to the substring reds exactly that test; resource fast tier 112; resource e2e **42 passed** on a real engine under the lock (the `Error importing 'res://broken.png'` and `.import:8` lines still match) |

Head `7137e0ef1`, one commit on `a751d80a9`; ruff / format / pyright clean; CI green.

qinhaihong-red:
## Follow-up recheck — still open at exact head `7137e0ef1`

The earlier `icon.png2` / `icon.png.backup` examples are now covered, but the underlying claim — match the asset as a whole token — is not closed. The new expression

```python
re.escape(res_path) + r"(?=$|['\"\s:,)\]]|\.import\b)"
```

still treats characters that can occur *inside a legal neighbouring resource path* as the end of the requested path. Against this exact head, each direct call below incorrectly returns the neighbour's line for `res://icon.png`:

```text
ERROR: Error importing 'res://icon.png.import-backup'.
ERROR: Error importing 'res://icon.png copy'.
ERROR: Error importing 'res://icon.png]backup'.
```

The first case enters through the special `\.import\b` alternative; the other two enter through the whitespace/bracket alternatives before the quoted path itself has ended. The new regression covers `icon.png2` and `icon.png.backup`, but none of these remaining boundaries, so 41 focused tests pass while the direct counterexamples still reproduce.

Required closure: recognize the complete path within the engine line's grammar rather than treating punctuation after the requested prefix as sufficient. In particular, match the real sidecar spelling `.import:<line>` as a complete suffix, and apply separators only after the complete quoted/unquoted token. Add the cases above, including a neighbour flood, to prove they cannot consume the 20-line cap.

Current CI is green with no pending jobs; that does not cover these inputs.

qinhaihong-red:
## Fourth review handled by backtrace at exact head `e90979849`

Four changes to the same matcher without closure is the cycle `/backtrace-review` exists for, so this round traced the decisions before touching the code. Root cause: every round refined WHAT MAY FOLLOW the asset's name — a substring, then a lookahead over punctuation — while nothing ever defined where a path token BEGINS and ENDS in the engine's own output. A path's alphabet holds every character such a lookahead can stop at, so no finite alternative list closes it: `res://icon.png` is a prefix of the legal neighbours `res://icon.png copy`, `res://icon.png]backup`, `res://icon.png.import-backup`. Two concerns were fused in one regex — cutting the line into path tokens (the engine's grammar) and comparing a token to the asset (equality). They are now separate.

Measured on the 4.6 sources, the import pass and the loaders print a path in two shapes. Quoted: `'%s'` in `Error importing`, `Cannot open file`, `Error opening file`, `Cannot open import file '%s.import'`, the editor's `'%s.import:%d'` / `'%s.import.md5:%d'` sidecar records; `"%s"` in the UID warnings. Bare: ended by whitespace, the end of the line, a sentence-final period (`Failed loading resource: %s.`), a following `(` (`Resource file not found: %s (expected type: …)`), or `:<line>` (the core `ResourceFormatImporter::load - %s.import:%d error:`).

| The whole-token claim is not closed: `'res://icon.png.import-backup'`, `'res://icon.png copy'`, `'res://icon.png]backup'` each attribute a neighbour's line to `res://icon.png` | **Fully Adopt** | reproduced; the `\.import\b` alternative and the whitespace/bracket alternatives fire inside a quoted path | `_ENGINE_PATH_TOKEN` cuts each line into path tokens by the engine's quoting and grammar above; a line names the asset when a token EQUALS the path, or is its sidecar (`.import` / `.import.md5`, each optionally `:<line>`), matched as a complete token | the three counterexamples are in the unit fixture as neighbours; the third-review regex, put back as a mutant, fails that test at index 0 with the `.import-backup` line |
| Recognize the complete path within the engine line's grammar; match `.import:<line>` as a complete suffix; apply separators only after the complete token | **Fully Adopt** | the grammar is measured above, not assumed | as above; the sidecar family is a full-token match, so `'res://icon.png.import:8'` (editor form) and the bare `res://icon.png.import:8 error:` (core form) both count and `'res://icon.png.import-backup'` does not | the unit fixture's `ours` set is every measured engine form with the path where its source puts it, including both sidecar spellings |
| Add the cases, including a neighbour flood, to prove they cannot consume the 20-line cap | **Fully Adopt** | — | unit: 25 `'res://icon.png copy <i>'` lines beside nine neighbours and ten own lines → exactly the ten own lines, `engine_output_truncated: false`. Real engine: `test_a_neighbours_lines_are_never_the_assets_own` imports `icon.png`, `icon.png copy.png`, `icon.png]backup.png`, `icon.png.png` (all broken) in one pass and checks each asset keeps only lines carrying its own quoted path | unit green; e2e green against Godot 4.6.3 under the machine-wide lock, and red on the third-review head (it ran against that code once by accident and failed) |

Declared residual, in the code comment: a BARE path that itself contains whitespace cannot be told from a path followed by prose; the engine quotes a path in every import-pass message measured, so the residual is stated rather than chased with more alternatives.

CI is green on `e90979849`. Locally: resource unit tier 112 passed, resource e2e 43 passed under the lock, `ruff` / `pyright` clean.

qinhaihong-red:
## Follow-up recheck — still open at exact head `e90979849`

The earlier prefix-neighbour over-attribution is closed, but the new tokenizer now under-matches a quoted asset path that itself contains a quote character. `_ENGINE_PATH_TOKEN` uses:

```python
r'''(?P<quote>['"])(?P<quoted>res://[^'"]*)(?P=quote)'''
```

The character class excludes **both** quote characters regardless of which one delimits this occurrence. A real Godot 4.6.3 import pass reproduces the gap with an invalid asset named `icon"hero.png`: Godot emits

```text
ERROR: Error importing 'res://icon"hero.png'.
```

but `gda resource import 'res://icon"hero.png'` settles the asset as `failed` / `dest_missing_after_pass` with `engine_output: []`. A direct call to `_engine_output()` with that exact engine line also returns `([], False)`.

This is not an invented output spelling: gda accepts the path, Godot scans it, and the engine emits the quoted line shown above. It also falls within the implementation comment's own premise that a path's alphabet can contain punctuation. The public per-asset evidence contract therefore remains incomplete even though the neighbour exclusions now work.

Required closure: recognize the complete known asset path inside the engine's quoted form without excluding the *other* quote character, while preserving exact-token equality, the current prefix-neighbour exclusions, and the `.import` / `.import.md5` sidecar forms. Add the quote-bearing filename as a regression; preferably keep the real-engine case because it proves the output grammar rather than assuming it.

Independent validation on this head: the existing resource selection passed (**42 tests**, including the real-engine neighbour test), while the real-engine quote-bearing counterexample above still reproduces. Current required CI jobs are green; the GitHub Godot E2E job is skipped.

qinhaihong-red:
## Fifth review handled at exact head `a38aec4b6`

Reproduced first: on `e90979849`, `_engine_output("ERROR: Error importing 'res://icon\"hero.png'.", 'res://icon"hero.png')` returned `([], False)`, and so did a path carrying the delimiting quote itself (`'res://it's.png'`). The tokenizer's `[^'"]*` cut a quoted path at EITHER quote.

Rather than widen the character class, this round re-derived the mechanism, since the class was the fourth device applied to one question: a substring, a lookahead, a tokenizer, and now a character class each tried to find where an UNKNOWN path ends in the line, and a path's alphabet defeats every one of them (it holds every punctuation mark, every quote). gda never has to cut an unknown path out of the line — it KNOWS the path it is looking for. `_names_asset` now finds that spelling (or its sidecar, `.import` / `.import.md5`, each optionally `:<line>`) in the line and checks the two boundaries the engine's delimiting fixes: a quoted path closes with the SAME quote character, whatever it holds in between; a bare one ends at whitespace, the end of the line, a sentence-final period or a `:<line>`. Exact-token equality, the prefix-neighbour exclusions and the sidecar forms are preserved by construction.

| A quoted asset path that contains a quote character is under-matched; the real engine prints `Error importing 'res://icon"hero.png'.` and gda reports `engine_output: []` | **Fully Adopt** | reproduced on `e90979849` (direct call and the engine line) | the known-path boundary check above; the comment in `resource.py` records the measured delimiting and why the known needle makes the unknown-alphabet problem disappear | `test_engine_output_names_an_asset_whose_path_carries_a_quote` × 3 (the other quote, the same quote, both), each with its own prefix neighbours excluded; the round-4 tokenizer as a mutant fails all three |
| Recognize the complete known path inside the quoted form without excluding the other quote, preserving exact equality, the neighbour exclusions and the sidecar forms | **Fully Adopt** | — | as above; the neighbour fixture of round 4 is unchanged and still passes (`'res://icon.png copy'`, `'res://icon.png]backup'`, `'res://icon.png.import-backup'`, the 25-line flood) | resource unit tier 115 passed |
| Keep the real-engine case | **Fully Adopt** | it proves the output grammar | `icon"hero.png` joins the four assets of `test_a_neighbours_lines_are_never_the_assets_own` (skipped only on Windows, where the name is illegal): one pass, each asset keeps only lines carrying its own quoted path | resource e2e 43 passed against Godot 4.6.3 under the machine-wide lock |

Declared residual, in the code comment: a bare path that itself contains whitespace, and a quoted NEIGHBOUR whose path continues after a quote character identical to its delimiter (`'res://icon.png'x.png'`) — the engine's own quoting cannot distinguish that line from one naming `res://icon.png`, and no reader can. Not chased.

CI is green on `a38aec4b6`.

PR #938, opened 2026-09-09
refactor(gda): project_file owns every section boundary, and one key decoder (#930)

### LINKED ISSUE #930: project_file owns every section boundary: the harness installer reads through ConfigText, and one key decoder
## What to build

The `project.godot` reader keeps its own invariant — one lexical scan owns every boundary — for every consumer, including the harness installer, and its interface shrinks to the names its consumers use.

Today the reader threads one scan across lines when it reads a whole file, but the installer that adds and removes the `GdaHarness` autoload entry re-recognizes sections line by line with the per-line reduction, which restarts the scan on every line. That is the second recognizer the module's own docstring says was removed after the #898 review. Reproduced on the dev head: a quoted value spanning several lines with an `[autoload]`-shaped line inside it is read by the whole-file reader as one `[application]` value, while the installer's loop enters the autoload section on that line — so an install can edit inside a string. Re-express the installer's two loops (ensure and remove) over the reader's parsed headers and entries, which already record line spans; add one section-scoped accessor to the reader if needed. The installer's guarantees stay as they are: a no-op when the entry is already exact, `[autoload]`-scoped edits, the literal comparison that decides an emptied section, byte-identical round trips.

Trim the interface: the names that exist only for tests (the line primitives, the key decoder, the parsed-text types) become private, so the reader is used through `read_config` / `read_config_text`, `bound_project_write`, and their result types and errors (19 public names to about 11). Tests that exercised the primitives move to the reader's interface.

One key decoder: `ConfigEntry` already carries the decoded key (`name`), yet the main-scene lookup in `gda.project` decodes the raw key token itself with the weaker quote stripper and re-implements the refusal of an escaped key. It reads `name` instead; the quote stripper stays only as the value-literal reader that lookup needs (a Godot value literal is not the reader's concern) and is documented as such.

No schema, help or harness change.

## Evidence

Found by the architecture review of the W2 promotion (#927) on 2026-09-08 and reproduced the same day in-process on the dev head `6d0e52306`: for a `[application]` value spanning three lines with `[autoload]` on the middle one, the whole-file reader reports headers at lines 0 and 5 while the installer's per-line loop reports the autoload section from line 2 onward. Seven of the reader's nineteen public names have no production caller.

## Acceptance criteria

- [ ] `daemon start` / `daemon uninstall` on a `project.godot` whose multi-line quoted value contains a header-shaped line install and remove the autoload entry in the real `[autoload]` section and leave the string untouched (installer regression, no engine needed).
- [ ] A commented header `[autoload] ; note` is the autoload section to the installer (installer regression; today pinned only at the primitive).
- [ ] The installer's existing guarantees hold unchanged: a repeat start is a no-op, uninstall removes only the entry and an emptied section, round trips are byte-identical (existing harness tests green).
- [ ] The reader's public names are the readers, the bounded write, their types and errors; the former test-only primitives are private, and no production caller outside the module uses them.
- [ ] `gda.project`'s main-scene lookup uses the entry's decoded name; an escaped key is refused by the one decoder; the existing main-scene tests stay green.
- [ ] `gda schema` and every `--help` are byte-identical to the base; harness bytes unchanged.

## Blocked by

None - can start immediately (forks from `feat/gda-dogfooding-dev` once the W2 promotion #927 has merged).

qinhaihong-red:
## Summary

`gda.project_file` now owns every `project.godot` section boundary, for the harness
installer too, and one decoder names every key.

The installer recognized `[autoload]` line by line with the reader's per-line reduction,
which restarts the scan on every line. That second recognizer carried none of the scan's
quote state, so a header-shaped line inside a multi-line quoted value opened a section the
file does not have. Reproduced on the dev head with an `[application]` description that
quotes the autoload declaration: the install re-pointed the line INSIDE the string (and
added no real entry), and the uninstall then deleted that line out of the value.

Both installer loops now run on what the reader's one scan recorded:

- `ConfigText.sections_named(name)` returns every part of a section with its bounds
  (`header`, `start`, `end`) and the entries inside them - the one section-scoped
  accessor the edit needed, which `_section_end` (the bounded write's restore) now reads
  as well;
- `ConfigEntry` carries the line it starts on, so an entry is re-pointed or removed by the
  SPAN it was written on, multi-line values included - and where a key is declared more
  than once, the install decides on the LAST declaration, which is the one `ConfigFile`
  loads;
- `ConfigText.spelled(lines)` is the one place a line list becomes this file's text - its
  terminator, its final break and its leading byte-order mark. Without it a rebuild from
  the scanned lines would have dropped the mark, so the installer's byte-identity
  guarantee is kept by construction and now pinned.

One key decoder: `gda.project`'s main-scene lookup decoded the raw key token itself and
re-implemented the refusal of a key it could not read. It now addresses each entry by
`ConfigEntry.name`, and only a key the reader could not name defers to the engine.
`ConfigEntry.key_token` existed for that second decoding and has no reader left, so it is
gone. The quote stripper moves to that lookup as `_unquoted_literal`, the one place in gda
that reads a Godot VALUE literal - `project_file` reads the format and leaves a value as
the text the file spells.

Interface trim: the line primitives become private, so the reader is used through the two
readers, the bounded write, and what they answer with.

## Public behavior

No command-surface change: no schema change, no help change, no result-model change, no
harness change. Inside the installer and the main-scene precondition, seven differences,
all of them consequences of the two rules above. Six are improvements; one shape (6) is
disclosed rather than fixed. Every engine claim below was checked with `ConfigFile.load`
on Godot 4.6.3.

1. **Fixed.** A header-shaped line inside a multi-line quoted value is no longer a section
   to the installer: the entry goes into the file's real `[autoload]`, and the string keeps
   its bytes. (The reproduced defect.)
2. **Narrowed to the engine's own reading.** A GdaHarness entry is matched by the reader's
   DECODED key, so ANY spelling the reader decodes to that key is the same entry to gda
   that it is to the engine - a quoted `"GdaHarness"`, a bare key with internal whitespace
   (`Gda Harness=`, `Gda\tHarness=`, which `parse_tag_assign_eof` reads as `GdaHarness`),
   or an escaped `\uXXXX` form. Before, the install added a duplicate beside it and the
   uninstall left it.
3. **Fewer writes.** The no-op is decided on the DECLARATION (key and value text), not on
   the raw line:
   - `GdaHarness= "*res://..."` (a space AFTER the `=`) already declares this value, so
     head leaves it alone; base rewrote it to the canonical spelling.
   - `GdaHarness = "*res://..."` (a space BEFORE the `=`) is the same entry to head; base
     did not recognize it at all (its check was `startswith("GdaHarness=")`) and appended
     a duplicate beside it.
   - A trailing comment was already a no-op on base - its own reduction stripped it - so
     nothing changes there.

   The receipt follows the write: `daemon start` on such a project now reports
   `installed_harness: false`, where base reported `true` once (it rewrote the entry, or
   added a duplicate beside it, and converged on the next start).
4. **One decoder decides, and a false refusal goes.** The main-scene lookup addresses each
   entry by the name the reader decoded, which changes two of the three spellings:
   - a plain quoted `"run/main_scene"` base already read (its quote stripper handled it) -
     unchanged;
   - an ESCAPED `"run/main_scene"` base DEFERRED on (any backslash in the token made it
     give up on the whole file); head reads it, so an escaped declaration of an EMPTY main
     scene is refused with `live_main_scene_undefined` instead of being left to the macOS
     alert;
   - a bare `run/main _scene=` base read as a DIFFERENT key, so it refused a project the
     engine runs perfectly well - `live_main_scene_undefined` on a declared main scene.
     Verified: the engine reports that key as `run/main_scene` and loads `res://main.tscn`.
     Removing that false refusal is the strongest single improvement in this PR, and it is
     now pinned.
5. **Degenerate input, one blank line.** On a hand-mangled file with TWO `[autoload]`
   sections that both hold a harness entry and are dropped together, the blank separator in
   front of the first section survives; the old walk re-measured after each drop and took
   it. Godot's own writer never emits two parts of one section, and no test pinned it.
6. **Disclosed, not fixed - the one shape where head is worse than base.** A file with a
   byte-order mark whose FIRST line is the `[autoload]` header. The reader drops the mark
   before it splits the text, so gda joins that header; the engine does not. Verified,
   `ConfigFile.load` of that file:

   ```
   load_err=0
   sections=[""]
   section= keys=["<mark>[autoload]GdaHarness"]
   ```

   The engine reads one section-less KEY and registers no autoload, while the install
   reports `installed_harness: true` with `created_sections: []` (it believes it joined a
   section). Base appended a fresh `[autoload]` at EOF, which the engine did read. The same
   file declares no readable `config_version`, so the engine loads no project from it at
   all: it belongs to the "Godot cannot write this" class the installer docstring already
   scopes out, and it is added to that list rather than coded around. The `_BOM` comment in
   the reader now states that the drop decides a SECTION on line 0, not only a key.
7. **A repeated key is decided the engine's way.** When the file declares GdaHarness more
   than once - `"GdaHarness"=<canonical>` followed by `GdaHarness="*res://old.gd"`, say -
   the install re-points the LAST declaration and leaves the earlier ones, because that is
   the one `ConfigFile` loads (verified: `get_section_keys` reports ONE `GdaHarness` and
   `get_value` returns `*res://old.gd`). Base matched only a literal `GdaHarness=` prefix,
   so it rewrote the first such line and appended a duplicate when the only entry was
   spelled otherwise. Review round 2 caught an interim state here: widening the match by
   decoded key while still stopping at the FIRST entry made the install report nothing to
   do while the engine went on loading `res://old.gd`, so the harness never registered and
   `daemon start` waited out its readiness deadline. Fixed in `f6fee77b8`, pinned by a test
   the first-entry mutant kills. The uninstall already removed every duplicate, so the two
   halves stay symmetric.

## Surface diff (byte-identity proof)

Captured from a detached checkout of the base `b7c485693` and from this branch's HEAD, with
`COLUMNS=100`: `gda schema` plus 98 `--help` outputs - the root, every group (`project` and
`daemon` included), every command in the schema manifest, and the meta commands. Re-run
after each review round against the reviewer's own base checkout:

```
$ diff -r /tmp/w3-930-surface-base2 /tmp/w3-930-surface-head4   # /private/tmp/pr938-base-b7c485693 vs HEAD f6fee77b8
(no output)
IDENTICAL: gda schema + 98 --help outputs
```

Harness bytes untouched: `git diff base..HEAD -- src/gda/harness/gda_harness.gd` is empty
and `HARNESS_VERSION` stays `"22"` (the pin in `tests/harness/test_harness_install.py` is
not touched; the new tests sit in their own block far above it).

## The reader's interface, before and after

19 public names to **11**; one is new (`ConfigSection`).

| | before | after |
| --- | --- | --- |
| readers | `read_config`, `read_config_text` | unchanged |
| bounded write | `bound_project_write` | unchanged |
| result types | `ConfigHeader`, `ConfigEntry`, `ConfigText`, `ProjectWriteMutation`, `SECTIONLESS` | unchanged, plus `ConfigSection` |
| errors | `ProjectFileRestoreError`, `ProjectFileChangedError` | unchanged |
| line primitives | `line_ending`, `split_config`, `strip_comment`, `config_line`, `is_section_header`, `section_name`, `config_key` | private (`_line_ending`, ...) |
| removed | `section_of` (no caller after the rework), `unquote` (moved to its one consumer) | - |

Consumers after the change:

- `src/gda/harness/install.py`: `ConfigEntry`, `ConfigSection`, `ConfigText`,
  `read_config_text` (was six line primitives);
- `src/gda/project.py`: `read_config` (was `read_config` + `unquote`);
- `src/gda/commands/project.py`: `read_config`, `bound_project_write` and the two errors
  (unchanged);
- tests: `tests/project/test_project_file.py` exercises the interface, with one
  pre-existing exception it keeps - it monkeypatches the private `project_file._restored`
  to land an external edit inside the restore's write window, the only seam a test can
  reach that race through; `tests/project/test_project_bounded_write.py` uses the two
  errors.

`ConfigHeader` has no importer of its own: it stays public as the element type of
`ConfigText.headers`, which callers do read.

No production caller outside the module uses a privatized name:

```
$ grep -rn --include='*.py' -E '\b(line_ending|split_config|strip_comment|config_line|is_section_header|section_name|section_of|config_key|unquote)\b' src | grep -v '^src/gda/project_file.py'
src/gda/mcp/server.py:47:from urllib.parse import unquote, urlparse
src/gda/mcp/server.py:238:    dirs = [unquote(urlparse(str(root.uri)).path) for root in result.roots]
```

Both hits are `urllib.parse.unquote` (a URL decoder), not this module's.

## Validation

Red then green. The two boundary regressions were written first and run against the
unchanged installer (commit `ccc2e150a`):

```
$ uv run --frozen pytest tests/harness/test_harness_install.py -q -k "quoted_one or documented_project or commented_autoload or byte_order_mark or emptied_commented"
FAILED ...::test_install_writes_into_the_real_autoload_section_not_a_quoted_one
FAILED ...::test_uninstall_takes_only_the_real_entry_out_of_a_documented_project
2 failed, 3 passed, 41 deselected
```

The three that passed are guards, not repairs: the commented-header pins move the
`[autoload] ; note` rule from the reader's primitive to the installer whose behavior it is,
and the byte-order-mark round trip guards a byte-identity the rework could have dropped in
silence. All five pass on HEAD.

Two rules the reviews found unpinned are now pinned by mutant. Each mutant was applied to
`install.py` and reverted, the file restored byte-identically (md5 checked):

```
# round 1 - the span edit
both span uses cut to one line:   1 failed, 46 passed  (test_an_entry_written_across_lines_...)
the remove use alone:             1 failed, 46 passed  (same test)

# round 2 - the last declaration
declared[-1] -> declared[0]:      1 failed, 47 passed  (test_install_re_points_the_entry_the_engine_reads_last)

restored:                        48 passed
```

Gates on HEAD (`f6fee77b8`):

| gate | result |
| --- | --- |
| `uv run --frozen pytest -m "not e2e" -q -p no:cacheprovider` | **2636 passed**, 684 deselected (base `b7c485693`: 2626 passed - +10 tests, none removed in net) |
| `uv run --frozen ruff check .` | All checks passed |
| `uv run --frozen ruff format --check .` | 515 files already formatted |
| `uv run --frozen pyright` | 0 errors, 0 warnings, 0 informations |
| `pytest tests/harness tests/project -m e2e` (real Godot 4.6.3, machine lock) | **122 passed** in 2:27 |

The existing harness suite (the install/uninstall guarantees, the #654 receipts, the
snapshot restores) is unchanged and green; the migrated reader tests keep every case the
privatized primitives pinned, asked of the text the rule applies to.

Everything this change is about runs in the **PR gate**: `_ensure_autoload` and
`_remove_autoload` are pure text edits, so the boundary regressions, the commented header,
the multi-line span, the repeated key, the byte-order mark and the whole reader suite are
unit tests. Nothing here needs an engine.

The **Godot e2e tier** is schedule/dispatch-only, so it does not run on this PR. It was run
locally instead - `tests/harness tests/project -m e2e`, 122 tests, all passing - because the
daemon's harness install writes a real `project.godot` an engine then boots, and the
main-scene precondition gates a real session launch. Every engine CLAIM in this PR was
checked the same way, by loading the file in question through `ConfigFile.load` on 4.6.3:
the repeated key, the marked header, the spaced bare key, and the orphan continuation line
(which the engine refuses outright - `ERR_PARSE_ERROR`, "Unterminated string").

## Review round 1

| finding | resolution |
| --- | --- |
| P2-1 marked `[autoload]` on line 0, disclosure only | Adopted as asked, no code change: a fifth bullet on the installer's scoped list, item 6 in Public behavior above, and the `_BOM` comment now says the drop decides a SECTION on line 0. Re-verified on Godot 4.6.3 before writing it down. |
| P3-2 span edit unpinned | Adopted: `test_an_entry_written_across_lines_is_re_pointed_and_removed_whole` covers both halves - the install re-points the whole span, and an uninstall that meets the multi-line entry takes all of its lines, which is what empties the section. Both mutants killed; evidence above. |
| P3-3 `ConfigSection.name` has no reader | Adopted: field dropped. A section is asked for by name, so the answer only has to say where it is. |
| P3-4 "Extend ``removed``" | Adopted: "Return ``removed`` widened by the sections it emptied". |
| P3-5 "Three shapes" over five bullets | Adopted: five, counting the new one. |
| P3-6 behaviors 2 and 4 under-state the rule | Adopted: both now say "any spelling the reader decodes to this key", with the whitespace and escaped forms named. |
| P3-7 PR body: test-file claim, receipt line | Adopted: the `_restored` monkeypatch is named as the one seam the reader tests still reach, and behavior 3 carries the receipt change. |

## Review round 2

| finding | resolution |
| --- | --- |
| P2-1 the install decided on the FIRST harness entry, the engine takes the LAST | Adopted, two lines: `_ensure_autoload` collects the harness entries across the `[autoload]` sections and decides on `declared[-1]` - a no-op when its value is canonical, otherwise that entry's span is re-pointed. `test_install_re_points_the_entry_the_engine_reads_last` pins install, repeat install and uninstall for the shape, and the first-entry mutant kills it. Engine behavior re-verified before the fix (`get_value` returned `*res://old.gd`). |
| P3-2 Public behavior 3 mis-stated base | Adopted: the space-AFTER-`=` spelling is the one base rewrote, the space-BEFORE-`=` one is the duplicate it appended, the trailing comment was already a no-op on base, and the receipt line now says base reported `true` once and then converged. |
| P3-3 Public behavior 4 mis-stated base, and its strongest case was unpinned | Adopted: the item separates the three spellings (quoted: already read; escaped: read instead of deferred; bare with inner whitespace: a FALSE refusal removed), and `test_an_escaped_application_key_is_read_as_the_setting_it_spells` gains the `run/main _scene=` case. The engine's reading of it was verified. |
| P3-4 three stale lines | Adopted: the `_drop_emptied_autoload_sections` summary no longer names a return shape it lost; the span test says the engine refuses the whole file (`ERR_PARSE_ERROR`, verified) instead of reading a stray declaration; `sections_named` no longer claims the section-less head is always exactly one section - a literal `[]` reopens it, and the engine merges the parts. |
| P3-5 `ConfigHeader` has no importer | Recorded, no change: it is the element type of `ConfigText.headers`, which callers read. Noted in the interface section above. |

The three open questions of round 0 were judged YES by review and are settled: the wider
no-op matches what the engine reads, the two-section blank line is unreachable by a clean
round trip, and indices (not lines) is the right shape for `ConfigSection` - only the unread
`name` field was wrong, and it is gone.

Closes #930

### DIFF
--- a/src/gda/harness/install.py
+++ b/src/gda/harness/install.py
@@ -61,7 +61,7 @@ string, and an uninstall deleted a line out of one (#930). Before that, a
 ``[autoload] ; note``: the install appended a SECOND autoload section and the
 uninstall left its entry behind (PR #898 review, round 3).
 
-Three shapes of input still come back changed, so the byte-identity guarantee of
+Five shapes of input still come back changed, so the byte-identity guarantee of
 :func:`uninstall_harness` is scoped to exclude them:
 
 - a file with MIXED terminators is normalized to its first one;
@@ -72,14 +72,21 @@ Three shapes of input still come back changed, so the byte-identity guarantee of
 - a file whose ``[autoload]`` section was ALREADY EMPTY loses that header, for the
   reason :func:`uninstall_harness` states — it is that function's guarantee, so the
   reasoning lives there and this list only names the shape (PR #898 review,
-  round 2: the list read as exhaustive and was not).
+  round 2: the list read as exhaustive and was not);
+- a file with a BYTE-ORDER MARK whose FIRST line is the ``[autoload]`` header: gda
+  joins that header, the engine reads no section there. The shared reader drops the
+  mark before it splits the text, so gda sees ``[autoload]`` where ``ConfigFile``
+  sees one marked, section-less KEY — and the install then reports a registration
+  the engine will not make. The same file declares no readable ``config_version``,
+  so the engine loads no project from it at all (#930 review, round 1).
 
 None is reachable for a ``project.godot`` the engine itself wrote: Godot's
 ``ConfigFile`` writer emits uniformly ``\\n``-terminated lines, always terminates
-the last one, and never emits an empty section. They need a hand-edited or
-tool-mangled file, so they are documented rather than coded around — the code stays
-a plain line-oriented edit instead of growing a per-line terminator model, or a
-record of pre-install state, for inputs Godot cannot produce.
+the last one, never emits an empty section, and never writes a byte-order mark.
+They need a hand-edited or tool-mangled file, so they are documented rather than
+coded around — the code stays a plain line-oriented edit instead of growing a
+per-line terminator model, a record of pre-install state, or a second reading of
+the format for inputs Godot cannot produce.
 """
 
 from dataclasses import dataclass
@@ -540,7 +547,7 @@ def install_harness(project: Path) -> HarnessInstall:
 def _drop_emptied_autoload_sections(
     config: ConfigText, sections: list[ConfigSection], removed: set[int]
 ) -> tuple[set[int], tuple[str, ...]]:
-    """Extend ``removed`` with the sections it emptied; (removed lines, dropped).
+    """Return ``removed`` widened by the sections it emptied; (lines, dropped).
 
     ``sections`` holds the ``[autoload]`` sections a harness entry was actually
     removed from — the ONLY sections this may drop. A section gda emptied would
--- a/src/gda/harness/install.py
+++ b/src/gda/harness/install.py
@@ -466,16 +466,24 @@ def _ensure_autoload(text: str) -> _ConfigEdit:
     # Re-point an existing GdaHarness entry, or insert a fresh one — both scoped to
     # the [autoload] section, so a same-named key in another section is never
     # touched (PR #247 review; symmetric with _remove_autoload).
-    for section in sections:
-        for entry in _harness_entries(section):
-            # A no-op when the entry already declares this value, whatever spacing
-            # or trailing comment the line carries: gda writes only where the
-            # DECLARATION differs, so it never rewrites a line it did not need to.
-            if entry.value == _AUTOLOAD_VALUE:
-                return _ConfigEdit(text, False)
-            lines = list(config.lines)
-            lines[entry.index : entry.index + len(entry.lines)] = [line]
-            return _ConfigEdit(config.spelled(lines), True)
+    declared = [entry for section in sections for entry in _harness_entries(section)]
+    if declared:
+        # The LAST declaration, because that is the one the engine loads
+        # (`ConfigText.settings`, verified on 4.6.3: a bare `GdaHarness=` after a
+        # quoted `"GdaHarness"=` wins). Deciding on the first let a canonical entry
+        # hide a later one pointing elsewhere, so the install reported nothing to
+        # do and the harness never registered (#938 review, round 2). The earlier
+        # duplicates are left as the file wrote them — the engine ignores them, and
+        # `_remove_autoload` takes them all out again.
+        entry = declared[-1]
+        # A no-op when that entry already declares this value, whatever spacing or
+        # trailing comment the line carries: gda writes only where the DECLARATION
+        # differs, so it never rewrites a line it did not need to.
+        if entry.value == _AUTOLOAD_VALUE:
+            return _ConfigEdit(text, False)
+        lines = list(config.lines)
+        lines[entry.index : entry.index + len(entry.lines)] = [line]
+        return _ConfigEdit(config.spelled(lines), True)
 
     # An existing [autoload] section with no GdaHarness entry — insert right after
     # its header, preserving any sibling autoloads.
@@ -547,7 +555,7 @@ def install_harness(project: Path) -> HarnessInstall:
 def _drop_emptied_autoload_sections(
     config: ConfigText, sections: list[ConfigSection], removed: set[int]
 ) -> tuple[set[int], tuple[str, ...]]:
-    """Return ``removed`` widened by the sections it emptied; (lines, dropped).
+    """Return ``removed`` widened by the sections it emptied, and their headers.
 
     ``sections`` holds the ``[autoload]`` sections a harness entry was actually
     removed from — the ONLY sections this may drop. A section gda emptied would
--- a/src/gda/harness/install.py
+++ b/src/gda/harness/install.py
@@ -67,8 +67,8 @@ Three shapes of input still come back changed, so the byte-identity guarantee of
 - a file with MIXED terminators is normalized to its first one;
 - a file with NO final terminator gains one (install terminates the line it
   appends after; uninstall has no way to know the file never ended in a break);
-- a CR-only (classic-Mac) file comes back CRLF — ``line_ending`` only tells
-  ``\\r\\n`` from ``\\n``, while ``str.splitlines`` also splits a bare ``\\r``;
+- a CR-only (classic-Mac) file comes back CRLF — the reader's terminator rule only
+  tells ``\\r\\n`` from ``\\n``, while ``str.splitlines`` also splits a bare ``\\r``;
 - a file whose ``[autoload]`` section was ALREADY EMPTY loses that header, for the
   reason :func:`uninstall_harness` states — it is that function's guarantee, so the
   reasoning lives there and this list only names the shape (PR #898 review,
@@ -113,7 +113,7 @@ HARNESS_VERSION = "22"
 _VERSION_HEADER_PREFIX = "# gda-harness-version:"
 _AUTOLOAD_HEADER = "[autoload]"
 # The same section by NAME. Every RECOGNITION on the edit path goes through this
-# (via the shared reader's `config_line` + `section_name`/`section_of`), because
+# (via the shared reader's `sections_named`, which reads its scan), because
 # Godot's own parser strips a header's inner whitespace and reads past a trailing
 # comment — `[ autoload ]` and `[autoload] ; note` are both the autoload section
 # (`VariantParser::_parse_tag`, `parse_tag_assign_eof`). Comparing a line to the
--- a/src/gda/harness/install.py
+++ b/src/gda/harness/install.py
@@ -43,17 +43,23 @@ uninstall each RETURN the exact set they touched (``created_paths`` /
 agent (or a reviewer) needs to audit what gda wrote into a tracked project.
 
 **Line endings (#654).** ``project.godot`` is read and written with newline
-translation OFF and rejoined with the terminator its FIRST line uses, so a CRLF
-project file stays CRLF — Python's default text mode would otherwise silently
-rewrite the whole file to LF on any autoload edit. The line primitives this
-edit runs on (``split_config``, ``config_line``, ``section_of``,
-``is_section_header``) belong to :mod:`gda.project_file`, the one reader of
-Godot's ``ConfigFile`` text (#843); this module contributes the
-``[autoload]``-specific EDIT, not a second reading of the format. Every raw line
-this edit looks at is reduced by that reader's :func:`~gda.project_file.config_line`
-first — a ``.strip()`` of its own left the comment on, and a header written
-``[autoload] ; note`` was then missed: the install appended a SECOND autoload
-section and the uninstall left its entry behind (PR #898 review, round 3).
+translation OFF and spelled back with the terminator its FIRST line uses, so a
+CRLF project file stays CRLF — Python's default text mode would otherwise
+silently rewrite the whole file to LF on any autoload edit.
+
+**Every boundary is the reader's (#843, #930).** :mod:`gda.project_file` is the
+one reader of Godot's ``ConfigFile`` text; this module contributes the
+``[autoload]``-specific EDIT, not a second reading of the format. The edit asks
+that reader for the ``[autoload]`` sections and the entries inside them
+(:meth:`~gda.project_file.ConfigText.sections_named`) and works on the line SPANS
+its single scan recorded. Recognizing the sections here, line by line, carried
+none of that scan's state — the per-line reduction restarts it on every line — so
+a header-shaped line INSIDE a multi-line quoted value opened an ``[autoload]``
+section the file does not have: an install wrote its entry into a description
+string, and an uninstall deleted a line out of one (#930). Before that, a
+``.strip()`` of this module's own left the comment on and missed a header written
+``[autoload] ; note``: the install appended a SECOND autoload section and the
+uninstall left its entry behind (PR #898 review, round 3).
 
 Three shapes of input still come back changed, so the byte-identity guarantee of
 :func:`uninstall_harness` is scoped to exclude them:
@@ -80,14 +86,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
 
-from gda.project_file import (
-    SECTIONLESS,
-    config_line,
-    is_section_header,
-    section_name,
-    section_of,
-    split_config,
-)
+from gda.project_file import ConfigEntry, ConfigSection, ConfigText, read_config_text
 
 # The autoload name and the res:// location the bundled harness is installed to.
 HARNESS_AUTOLOAD_NAME = "GdaHarness"
@@ -122,6 +121,11 @@ _AUTOLOAD_HEADER = "[autoload]"
 # emptied header left behind (PR #898 review). The literal above stays what the
 # install WRITES and what the #654 receipt reports.
 _AUTOLOAD_SECTION = "autoload"
+# The setting the reader names the harness entry by — `<section>/<key>`, with the
+# key DECODED. Matching the entry by that name is what makes the two spellings of
+# one key (`GdaHarness` and `"GdaHarness"`) the same entry to gda, as they are to
+# the engine, so an install can never add a duplicate beside the quoted one.
+_HARNESS_SETTING = f"{_AUTOLOAD_SECTION}/{HARNESS_AUTOLOAD_NAME}"
 _PROJECT_FILE = "project.godot"
 _BUNDLED_HARNESS = Path(__file__).parent / HARNESS_FILE
 
@@ -193,9 +197,12 @@ class _ConfigEdit:
     sections: tuple[str, ...] = ()
 
 
+# Enabled-singleton form: the res:// path prefixed with "*" (issue #119).
+_AUTOLOAD_VALUE = f'"*{HARNESS_RES_PATH}"'
+
+
 def _autoload_line() -> str:
-    # Enabled-singleton form: the res:// path prefixed with "*" (issue #119).
-    return f'{HARNESS_AUTOLOAD_NAME}="*{HARNESS_RES_PATH}"'
+    return f"{HARNESS_AUTOLOAD_NAME}={_AUTOLOAD_VALUE}"
 
 
 def _read_config(path: Path) -> str:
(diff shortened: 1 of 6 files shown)


qinhaihong-red:
**Review record, round 1 (orchestrator, 2026-09-09).** Independent adversarial review at `b2c5c2f30` (four axes; install→uninstall on four real `project.godot` files and 24 hand-built shapes; a 600- and a 900-case differential fuzz of generated files through both checkouts with every divergence classified; the engine's own `ConfigFile.load` on Godot 4.6.3 as the oracle; three mutants) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 6 P3**. The fix round is the implementer's. Fix head: `f6f83fc09`.

|---|---|---|
| **P2-1** a sixth behaviour difference, undisclosed and the only one where head is worse than base: a byte-order-marked file whose FIRST line is the `[autoload]` header — the reader strips the mark before splitting, so head joins that header, while the engine reads no section there (`sections=[""]`, key `<mark>[autoload]GdaHarness`) and the harness is not registered though `installed_harness: true` is reported; base appended a fresh section the engine did read | Confirmed on the engine (`ConfigFile.load`) | **Adopted as DISCLOSURE, not code** (such a file has no readable `config_version` either — the "Godot cannot write this" class the installer already scopes out; coding around it would be the disproportionate move): a fifth bullet on the installer's scoped list, a sixth item in the PR's Public behavior, the `_BOM` comment now says the drop decides a section on line 0. |
| P3-2 the span edit the Summary advertises ("re-pointed or removed by the SPAN it was written on, multi-line values included") was unpinned — both span uses mutated to single-line left 327 tests green | Confirmed | **Adopted** — `test_an_entry_written_across_lines_is_re_pointed_and_removed_whole`; both mutants now fail exactly that test. |
| P3-3 `ConfigSection.name` had no reader | Confirmed | **Adopted** — field dropped. |
| P3-4 `_drop_emptied_autoload_sections` docstring described a mutation that does not happen | Confirmed | **Adopted** — "Return `removed` widened by the sections it emptied". |
| P3-5 "Three shapes of input" over a longer list (pre-existing, #898) | Confirmed | **Adopted** — "Five shapes"; the paragraph also says Godot's writer never emits a byte-order mark. |
| P3-6 disclosures 2 and 4 under-stated the rule (any spelling the reader decodes to the key: quoted, bare with internal whitespace, escaped) | Confirmed on the engine | **Adopted** — PR body reworded. |
| P3-7 PR-body inaccuracies: the reader tests still monkeypatch `project_file._restored`; the widened no-op also changes the `installed_harness` receipt on a spacing-only match (`false` where every start said `true` — #654's intent) | Confirmed | **Adopted** — both stated. |

Reviewer-verified and unchanged: 19 → 11 public names with no production use of a privatized one (only `urllib.parse.unquote` matches); `gda schema` and every `--help` byte-identical over the reviewer's own 94-output enumeration; harness bytes and `HARNESS_VERSION` untouched; fast tier 2634 (base 2626), e2e 122 under the lock; the red-then-green trail reproduced by overlaying base's installer under head's tests (2 failed / 3 passed); all five disclosed behaviours reproduced (the fixed defect: base wrote the entry INSIDE a quoted description and uninstall deleted a line out of it; the quoted-key case: base left a dangling autoload pointing at the deleted script — ADR-0018's point-3 hazard — head no-ops and removes it); four real projects install / re-install / uninstall byte-identically on both checkouts; `_section_end`'s refactor behaviour-preserving over 500 cases; the deliberately literal comparison (`install.py` `config.lines[header].strip() != _AUTOLOAD_HEADER`) intact and now pinned at the installer; `bom` + `spelled`, `sections_named` and `ConfigEntry.index` each survive a deletion test and the installer shed its hand-kept section state; `_unquoted_literal`'s home judged right; the three open questions all judged YES (the wider no-op matches what the engine reads and serves #654; the two-section blank line is unreachable by a clean round trip; indices, not lines). On `f6f83fc09`: fast tier 2635, e2e 122, schema + helps identical, PR CI green. A second independent round runs on this head per the wave rule.

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-09).** Second independent review at `f6f83fc09` (round-1 mutants re-run through an overlay copy; an independent 2 000-text fuzz of generated `project.godot` shapes; the engine's own `ConfigFile.load` on Godot 4.6.3 as the oracle for every claim about what it reads) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 4 P3**. Fix head: `f6fee77b8`, verified directly by the orchestrator (no third review round — the budget is two).

|---|---|---|
| **P2-1** a seventh behaviour difference, undisclosed and the only one where head is worse: `_ensure_autoload` decided on the FIRST harness entry, while `ConfigFile` keeps the LAST declaration — which `project_file`'s own `settings()` documents. With a canonical quoted entry followed by a bare `GdaHarness="*res://old.gd"`, head reported `installed_harness: false` while the engine loads `old.gd`, so the harness never registers and `daemon start` fails at its readiness deadline; base rewrote the bare line | Confirmed on the engine (`ConfigFile.load` returns one key, value `*res://old.gd`) | **Adopted — two lines**: the install decides on the last declaration across the autoload sections and re-points that entry's span; `_remove_autoload` already removed every duplicate, so the halves stay symmetric. `test_install_re_points_the_entry_the_engine_reads_last` covers install / repeat / uninstall; the `declared[0]` mutant fails exactly it. |
| P3-2 "Public behavior" item 3 mis-stated base: a space BEFORE `=` made base append a duplicate, a trailing-comment line was already a no-op there, and base converged after ONE install | Confirmed on both checkouts | **Adopted** — the item now uses the space-after-`=` example and says base rewrote or duplicated once. |
| P3-3 item 4 mis-stated base on two of three spellings: a plain quoted `"run/main_scene"` was already read, and a bare key with inner whitespace was REFUSED (`live_main_scene_undefined`) on a project the engine runs — head removes a false refusal, the strongest thing in the PR, and nothing pinned it | Confirmed on the engine | **Adopted** — the item separates the three spellings, and the whitespace case is now a third assertion in the escaped-key test. |
| P3-4 three stale lines: the `_drop_emptied_autoload_sections` summary still named its old return shape; a test comment said the engine reads a stray continuation line as a declaration (it refuses the whole file — `ERR_PARSE_ERROR`, "Unterminated string"); `sections_named` promised the section-less head is exactly one part (a literal `[]` reopens it) | Confirmed | **Adopted** — all three reworded. |
| P3-5 `ConfigHeader` has no importer outside the module (it survives as the element type of `ConfigText.headers`) | Confirmed | **Recorded**, no change; the PR's interface section says so. |

Reviewer-verified and unchanged: all five round-1 fixes, each with its mutant; `gda schema` plus 98 `--help` outputs byte-identical to the base over the reviewer's own enumeration; harness bytes and `HARNESS_VERSION` untouched, the test-file diff one appended block; the red trail reproduced by overlaying base's two modules under head's tests; all six disclosed behaviours reproduced against the engine; `read_config_text` total over 2 000 generated texts with every rebuild byte-exact, and 1 403 non-excluded shapes byte-identical through install → uninstall; `bound_project_write`'s restore identical to base over LF / CRLF / no-final-newline / a multi-line dictionary (the one divergence, a BOM'd `after`, is unreachable — Godot's writer emits none); a repeat install never touches `project.godot`'s mtime (ADR-0018 point 4); the crash-safe uninstall ordering intact; entry spans cannot cross a section boundary by construction. Orchestrator verification on `f6fee77b8`: harness diff empty at version 22; `tests/harness` + `tests/project` fast tier 329 passed; the `declared[0]` mutant fails exactly the new test; `gda schema` byte-identical to the base and five spot-checked helps identical; harness + project e2e **122 passed** under the machine lock on Godot 4.6.3. **READY** for the maintainer's review.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `f6fee77b8e7765eae7cb163582e5f509e6b8bea0`. Existing reviews and replies were not used as evidence.

## Artifact review — Pass

Issue #930 and the PR body give a coherent scope: one stateful `ConfigText` scan owns section and entry spans, the harness installer consumes those spans, the key decoder has one owner, and command/schema/harness behavior is otherwise unchanged.

**Terminology and prose assessment:** I inspected the complete issue, PR body, and relevant project-file/installer prose against `CONTEXT.md`, ADR-0006/0018/0040, and ASD-STE100. The terms `ConfigText`, section boundary, entry span, and key decoder are consistent. No substantive Markdown/issue terminology or prose problem was found. The code-docstring contradiction below is reported on the Standards axis. I did not compare the scanner with every malformed `ConfigFile` grammar production.

## Architecture review — Pass

The observed structure is the intended one-context topology: `gda.project_file` owns lexical ConfigFile knowledge; project reads and the harness installer depend on that owner; the installer owns only the autoload edit. The public reader surface is smaller, dependencies are directed, and the change removes rather than creates overlapping boundary recognizers.

## Standards

- Hard: `src/gda/harness/install.py:702-704` says the uninstall byte-identity guarantee excludes only three malformed shapes. The same module's authoritative list at lines 60-81 now has five, including a BOM before the first `[autoload]` header. A caller of `uninstall_harness` is therefore told that the BOM case is covered when it is expressly excluded. This violates the synchronization rule in `RULES.md:119-123`. Refer to the complete module-level list or enumerate the BOM case in the function contract.

No justified baseline smell was found.

Standards summary: 1 hard finding; worst is a contradictory byte-identity contract.

## Spec

No Spec finding. The installer uses scanned spans, multiline/header cases are covered, the public reader surface is reduced, main-scene lookup uses the decoded key, and the diff does not alter harness bytes or command descriptors. Independent validation: 168 targeted tests passed; Ruff passed; Pyright reported 0 errors.

Spec summary: 0 findings; worst: none.

qinhaihong-red:
## Third review handled at `04cd9a618`

| `uninstall_harness`'s docstring says the byte-identity guarantee excludes three malformed shapes while the module's authoritative list has five (the BOM case included) | Fully Adopt | `install.py` lines 60–81 list five shapes; the function docstring enumerated three | The function contract now names all five and says the module list is the authority it points at; ADR-0018 gains a dated `> **Outcome (2026-09-13, #930)**` note recording that the list grew (its "excludes three" summary is the point-in-time record) | harness + project fast tier 329 passed; ruff / format clean; CI green |

Head `04cd9a618`, one commit on `f6fee77b8`. Artifact, Architecture and Spec axes had no finding; nothing else changed.

PR #939, opened 2026-09-09
fix(gda): game tree serializes every tree its result model accepts (#929)

### LINKED ISSUE #929: game tree serializes every tree its result model accepts, and game tree / game find counters are checked
## Re-scope (2026-09-09, decided by the maintainer)

The first implementation attempt measured the depth ladder of `game tree` on `main` `b7c485693` (Godot 4.6.3, pydantic 2.13) and showed that the diagnosis below was incomplete: the recursive serializer is the HIGHEST of four ceilings, and fixing it changes no CLI outcome.

| nesting levels | outcome today | ceiling |
| --- | --- | --- |
| about 127 to 253 | exit 1, bare traceback, no error envelope | the per-node `children_omitted` wrap serializer (#849) trips pydantic's serialization depth guard, uncaught — a regression: `scene get`'s plain nested model serializes to 254 and then refuses typed |
| 254 to 512 | `tree_too_deep` (typed) | pydantic's recursion limit |
| 513 and deeper | `contract_violation`, the reply JSON is cut off | the engine's `JSON.stringify` bails at `Variant::MAX_RECURSION_DEPTH` (1024 Variant levels, `core/io/json.cpp`) |
| about 1024 and deeper | the same, plus "Stack overflow" in the engine log | the GDScript call stack — the only ceiling the original text targeted |

Revised scope — a proportionate contract, no promise for depths no real project reaches:

- Fix the first ceiling with the wire shape unchanged: replace the per-node wrap serializer with ONE top-level serializer on the tree result that dumps the tree and prunes the zero `children_omitted` keys iteratively, so every tree the validator accepts (up to pydantic's limit) also serializes, and the JSON is byte-identical to today's for every existing fixture. Confirm the new ceiling by measurement (the ladder method), not by reasoning.
- The second ceiling stays as it is: `tree_too_deep` is the documented refusal for a tree deeper than about 250 nesting levels; `game tree --help` states that ceiling and the remedy (`--root` / `--max-depth`) in one sentence — the only help/schema-description change, enumerated in the PR.
- The third and fourth ceilings are NOT addressed: no harness change, no `HARNESS_VERSION` bump, no explicit-stack rewrite — a garbled reply from a 500-level tree is a `contract_violation` today and stays one; the ceiling sentence in help discloses it. "Exact at any depth" would need a flat wire shape (an ADR-level decision) and is out of this milestone.
- The result invariants stay in scope (the acceptance criteria below marked "kept").

Acceptance criteria (superseding the list below):

- [ ] A real-engine e2e reads a chain about 200 levels deep unbounded: exit 0, every node serialized, `truncated` false, `omitted_nodes` 0, `diag errors` empty — this case crashes on the base.
- [ ] A real-engine e2e reads a chain about 300 levels deep unbounded and gets `tree_too_deep`, and the same chain bounded with `--max-depth` reads exactly (the existing 2200-node bounded tests stay green).
- [ ] The zero-`children_omitted` omission is byte-identical to today's output for every existing fixture; `gda schema` differs from the base only in `game tree`'s description text, and that diff is enumerated in the PR.
- [ ] (kept) `GameTreeResult` and `GameFindResult` refuse a reply where `truncated` disagrees with `omitted_nodes > 0`, and `GameFindResult` refuses `count != len(matches)`; the CLI reports `contract_violation` for such a reply.
- [ ] No harness change: harness bytes and `HARNESS_VERSION` are untouched.

---

_The original text follows for the record; its "What to build" first paragraph and its acceptance criteria 1, 2 and 5 are superseded by the re-scope above._

## What to build

A `game tree` read of a scene deeper than the GDScript call stack returns the whole tree, and the two tree results refuse a reply whose counters contradict each other.

Today the harness serializes the tree recursively, so an unbounded `game tree` over a 2200-node chain overflows the engine's stack: the reply is cut off mid-JSON, gda publishes `contract_violation` ("Expecting value: line 1 column 6683"), and `diag errors` shows "Stack overflow" in the serializer. The count-only walk (#849) and the `game find` walk (#855) already use an explicit stack; the serializer is the one walk that still recurses. Convert it to the same explicit-stack, document-order walk so the three walks share one rule. Unifying the three walks into one is out of scope unless it falls out for free.

`GameTreeResult` and `GameFindResult` document invariants they do not enforce: `truncated` is true exactly when `omitted_nodes` is above zero, and `count` equals the number of `matches`. A reply with `count` 5 beside two matches, or `truncated` false beside `omitted_nodes` 7, is published today as a successful, complete read. Add `after` validators on both result models so such a reply fails output validation as `contract_violation`, the way the `input` results already refuse a reply outside their contract. The published schema stays byte-identical (validators add no properties) and the field descriptions are unchanged.

Harness version: the merge base's plus one at rebase time, content pin re-taken. This is the first harness slice of W3, so it forks and merges before the other two harness slices of the wave.

## Evidence

Found by the architecture review of the W2 promotion (#927) on 2026-09-08 and reproduced the same day on the dev head `6d0e52306`: the unbounded read on Godot 4.6.3 (macOS) exits 5 with the envelope above and the engine error in `diag errors`; the inconsistent `GameFindResult` payload validates in-process. Pre-existing on `main` since #849; not a W2 regression.

## Acceptance criteria

- [ ] Unbounded `gda game tree` over the existing deep-chain fixture (a chain deeper than the call stack) exits 0, serializes every node with `truncated` false and `omitted_nodes` 0, and `diag errors` reports no engine error afterwards; the case is added to the real-engine e2e beside the bounded reads.
- [ ] The bounded reads and `game find` over the same chain keep their exact counts (existing e2e green).
- [ ] `GameTreeResult` and `GameFindResult` refuse a reply where `truncated` disagrees with `omitted_nodes > 0`, and `GameFindResult` refuses `count != len(matches)`; the CLI reports `contract_violation` for such a reply (fast-tier tests through the fake live runner).
- [ ] `gda schema` is byte-identical to the base for both commands; the human rendering is unchanged.
- [ ] `HARNESS_VERSION` is the merge base's plus one with the content pin updated; the version guard passes.

## Blocked by

None - can start immediately (forks from `feat/gda-dogfooding-dev` once the W2 promotion #927 has merged).

qinhaihong-red:
## Summary

`gda game tree` crashed on a runtime tree its own result model accepts. `GameNode`
carried a `model_serializer` that dropped a zero `children_omitted` from the emitted
JSON; a model serializer is a Python callback, and pydantic stops calling those at
about 128 nested levels — far below the 254 levels the model validates. A tree
between those two depths validated and then raised an uncaught
`PydanticSerializationError` on the way out: a `--json` read exited 1 with a bare
traceback and no `Error envelope` at all, while `scene get`'s plain nested model
refused the same depth as the typed `tree_too_deep`. The human channel was
unaffected — it renders from the model and never calls `model_dump_json` — so the
failure hit exactly the channel an agent reads. Regression from #849.

The prune moves to ONE `model_serializer` on `GameTreeResult`, which runs pydantic's
own serializer over the whole tree once and then walks the dumped dictionaries with
an explicit stack. `GameTreeResult` and `GameFindResult` also gain the `after`
validators that enforce the invariants their field descriptions already promised.

The wire shape is unchanged, and the three ceilings above the result model's are
disclosed rather than fixed, per the 2026-09-09 re-scope.

## Public behavior

- A `game tree --json` read of 128 to 254 nesting levels now returns the tree instead
  of crashing. The JSON is what it always was: `children_omitted` present exactly
  where the count is above zero. The human channel already worked at those depths;
  only `--json` failed.
- Past 254 levels the refusal is unchanged: the typed `tree_too_deep`.
- Past about 512 levels the reply is still `contract_violation` — Godot's own
  `JSON.stringify` bails at `Variant::MAX_RECURSION_DEPTH` and cuts the reply short.
  Not addressed here.
- `game tree --help` now discloses BOTH ceilings and the remedy in one sentence:
  "A tree nesting deeper than about 250 levels is refused (`tree_too_deep`; past
  about 500 the engine's own JSON writer cuts the reply short and the refusal is
  `contract_violation`): bound such a read with `--root` and `--max-depth`."
- A caller's `exclude=` / `include=` / `exclude_defaults=True` on the result model
  keeps working; the prune finds nothing to do rather than raising.
- A `game tree` or `game find` reply whose counters contradict each other
  (`truncated` disagreeing with `omitted_nodes > 0`, or `count != len(matches)`) is
  now `contract_violation` instead of a successful, complete-looking read.
- No harness change: `src/gda/harness` and `HARNESS_VERSION` are untouched.

## Surface diff

Compared against a detached checkout of the base `b7c485693`:

| surface | result |
| --- | --- |
| `gda schema` | ONE structural difference in the whole document: `commands[15].description` (`game tree`) gains the ceiling sentence quoted under Public behavior. No property, `$defs`, type or required-list change anywhere. |
| `gda game tree --schema` | byte-identical (the per-command document carries no command description) |
| `gda game tree --help` | the same one sentence added, nothing else |
| `gda game find --help` | byte-identical |
| emitted JSON for `GAME_TREE_RESULT` and `GAME_TREE_TRUNCATED_RESULT` | byte-identical on `model_dump()`, `model_dump(mode="json")` and `model_dump_json()`, key order included |

The `TypedDict` return annotation on the new serializer is not published: `--schema`
is generated in validation mode, where a serializer's return type does not appear.
The zero-difference `game tree --schema` diff is the evidence.

Re-measured after review round 1 against the reviewer's base checkout
`/private/tmp/pr939-base-b7c485693`: the same four rows hold, with the reworded
sentence in place of the first one.

## Validation

Measured ladder, `GameTreeResult` with a single chain of N nesting levels
(pydantic 2.13.4 / pydantic-core 2.46.4, CPython 3.13.13):

| levels | base `b7c485693` | this PR |
| --- | --- | --- |
| 127 | validate OK, dump OK, dump_json OK | same |
| 128 | validate OK, dump OK, **dump_json raises `PydanticSerializationError` (uncaught: a `--json` read exits 1 with no envelope; the human channel still prints the tree)** | validate OK, dump OK, dump_json OK |
| 200 | same crash | OK |
| 253 | same crash | OK |
| 254 | same crash | OK |
| 255 | `ValidationError` (`recursion_loop`) -> `tree_too_deep` | unchanged |
| 300 | `tree_too_deep` | unchanged |

Bisected ceilings: base `model_dump_json` 127 levels, this PR 254 levels — exactly the
model's validation limit, which is the condition the re-scope set. `model_dump` was
254 on both.

Red then green, real engine (Godot 4.6.3, macOS), `tests/live/test_e2e_game.py`:

- `test_game_tree_serializes_a_chain_the_result_model_accepts` (201 levels unbounded)
  with the per-node serializer restored: `1 failed`, the child process exiting 1 with
  `PydanticSerializationError: Error calling function _omit_absent_omission_count`.
  With the fix: passes — exit 0, 201 levels walked, `truncated` false, `omitted_nodes`
  0, no `children_omitted` anywhere, `diag errors` empty.
- `test_game_tree_refuses_a_chain_past_the_result_model_limit` (301 levels): unbounded
  gives `tree_too_deep` / category `parse`; the same chain with `--max-depth 1` reads
  exactly (`children_omitted` 1 on the one node, `omitted_nodes` 299).

Fast tier, red first: `test_game_tree_emits_a_deep_tree_the_model_accepts` failed on
the base with the same uncaught error, while the two byte-identity cases passed on the
base — so they pin the rule, not the new mechanism.

Mutant checks (each reverted, run, restored):

| mutant | result |
| --- | --- |
| per-node `children_omitted` serializer restored | the 201-level e2e fails (exit 1, uncaught `PydanticSerializationError`) |
| serializer return annotation back to `dict[str, Any]` | `test_game_tree_emits_a_deep_tree_the_model_accepts` fails |
| `children_omitted` removed from `EmittedGameNode` | `test_the_emitted_shape_names_every_field_the_models_hold` fails |
| `omitted_nodes` removed from `EmittedGameTree` | the same drift test fails |
| the prune's defensive reads removed | `test_the_prune_leaves_an_excluded_tree_alone` fails (`PydanticSerializationError`) |
| the refusal case's reply made shallow (stands for the model limit rising) | `test_game_tree_refuses_a_reply_past_the_result_model_limit` fails |
| `GameTreeResult._check_omission_counters` body removed | `test_a_tree_reply_whose_counters_disagree_is_a_contract_violation` fails |
| `GameFindResult._check_result_counters` body removed | `test_a_find_reply_whose_counters_disagree_is_a_contract_violation` fails |

Gates:

| gate | result |
| --- | --- |
| `uv run --frozen pytest -m "not e2e" -q` | 2634 passed |
| `uv run --frozen ruff check .` | All checks passed |
| `uv run --frozen ruff format --check .` | 515 files already formatted |
| `uv run --frozen pyright` | 0 errors, 0 warnings, 0 informations |
| `pytest tests/live/test_e2e_game.py tests/harness -m e2e` (real engine, serialized by the machine-wide lock) | 20 passed |
| `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | valid (harness unchanged) |
| `git diff <base> -- src/gda/harness tests/harness/test_harness_install.py` | empty |

PR CI runs NO Godot e2e (that tier is schedule/dispatch-only), so the two tiers guard
different halves of this change:

- **On PR CI**: the fast tier, which carries the whole serializer contract — the
  byte-identity cases against both shared fixtures, the 200-level deep-emit case, the
  300-level refusal case that pins the documented ceiling from the other side, the
  emitted-shape drift test, and the five contradictory-reply cases; plus `ruff`,
  `pyright`, and
  `harness_version_guard.py`, which proves the harness bytes and `HARNESS_VERSION` did
  not move. The 200-level case reaches the same code path as the e2e one through the
  fake live runner, so the regression itself cannot come back unnoticed on PR CI.
- **Needs the real engine** (run locally for this PR, and on the nightly `godot-e2e`
  job): that a real `Engine session` actually builds a 201-node chain and that the
  daemon relays a reply of that size; that `diag errors` is empty afterwards, which
  only a live session can report; and that the 301-level chain refuses as
  `tree_too_deep` rather than failing earlier inside the engine.

## Review round 1

| finding | resolution |
| --- | --- |
| P2-1 — the emitted `TypedDict`s are a hand-written parallel of the models' fields; a forgotten field is dropped from the JSON silently while `--schema` keeps publishing it | Adopted. `test_the_emitted_shape_names_every_field_the_models_hold` requires `set(EmittedGameNode.__annotations__) == set(GameNode.model_fields)` and the same for the result pair, so the models stay the one authority for WHICH fields exist and the declarations only restate `children_omitted`'s optionality. Two drift mutants kill it. Fixed at `312c3593d`. |
| P2-2 — the help sentence names only gda's ceiling; the re-scope requires it to disclose the engine's too | Adopted verbatim. The sentence now reads "...is refused (`tree_too_deep`; past about 500 the engine's own JSON writer cuts the reply short and the refusal is `contract_violation`)...". Schema and help re-diffed against `/private/tmp/pr939-base-b7c485693`: still exactly one structural difference in `gda schema`, `game tree --schema` and `game find --help` byte-identical. Fixed at `312c3593d`. |
| P3-1 — the prune indexes `root` and `children`, so `exclude=` / `include=` / `exclude_defaults=True` raise a `KeyError` where the base returned a value | Adopted. Both reads are defensive; `test_the_prune_leaves_an_excluded_tree_alone` covers `exclude={"root"}` on both dump paths. Measured after the change: `exclude`, `include` and `exclude_defaults` all succeed, no serialization warnings under `warnings="error"`, and the ceiling is still 254. Fixed at `312c3593d`. |
| P3-2 — fold the pre-existing 2200-node test's three writes into `_chain_project` | Not adopted, by the orchestrator's decision: no rule is at stake and the existing test is left untouched. |

## Review round 2

| finding | resolution |
| --- | --- |
| P3-1 — the documented refusal ceiling is guarded only by the nightly engine tier; the fast tier catches the limit falling but not rising | Adopted. `test_game_tree_refuses_a_reply_past_the_result_model_limit` sends a 300-level reply through the fake live runner and asserts `EXIT_PARSE` + `tree_too_deep`, so a dependency that lifted the limit past 300 goes red on PR CI instead of only on the nightly tier. Its own mutant (a shallow reply) kills it. Fixed at `6c2811b8d`. |
| P3-2 — the body said "a read of 128 to 254 nesting levels … exit 1"; only the `--json` channel crashed, the human channel printed the tree | Adopted. Summary, Public behavior and the ladder row now say `--json` and name the human channel explicitly. No code change. |
| P3-3 — a bare `GameNode` dumped alone now keeps `children_omitted: 0` | Adopted. The comment on the field says so in half a sentence, with the note that no gda path dumps a node alone. Comment only. Fixed at `6c2811b8d`. |
| Splitting the help sentence into two | Not adopted, by the orchestrator's decision: the re-scope asked for one sentence and round 1 adopted this wording verbatim. |
| `ge=0` on the counters | Not adopted, by the orchestrator's decision: it would add `minimum: 0` to the published schema, and the harness cannot emit a negative count. Recorded as a non-issue. |

## Open questions

- The ceiling sentence says "about 250 levels" rather than the exact 254. The exact
  number is a pydantic-core internal and could move with a dependency bump, so the
  help states the order of magnitude and the remedy. Say if you would rather pin the
  measured number.
- The two ceilings above this one stay unaddressed by decision: the engine's
  `JSON.stringify` guard (about 512 tree levels, `contract_violation`) and the GDScript
  call stack (about 1024). Making `game tree` exact at any depth needs a flat wire
  shape and is an ADR-level decision outside this milestone.

## History note

The first attempt on this branch also carried an explicit-stack rewrite of the
harness `_serialize` (`73ebcd292`). It was dropped at the 2026-09-09 re-scope: the
measurement showed that ceiling is the highest of four and that removing it changes no
CLI outcome. This branch was reset to the base and only the validator commit was
cherry-picked forward.

Closes #929

### DIFF
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -17,7 +17,7 @@ against the engine session it holds, reading the runtime ``SceneTree`` after
 """
 
 import json
-from typing import Any, Optional
+from typing import Any, NotRequired, Optional, TypedDict
 
 import typer
 from pydantic import (
@@ -88,17 +88,8 @@ class GameNode(BaseModel):
     # The presence rule above is a SERIALIZATION rule, so it lives on the writer
     # rather than on every caller: the field stays a plain int with a 0 default
     # (a consumer reads a number, never None), and the key is dropped from the
-    # emitted JSON when nothing was omitted. `mode="wrap"` runs pydantic's own
-    # serializer first, so nested children are serialized — and pruned — by this
-    # same rule at every depth.
-    @model_serializer(mode="wrap")
-    def _omit_absent_omission_count(
-        self, handler: SerializerFunctionWrapHandler
-    ) -> dict[str, Any]:
-        rendered = handler(self)
-        if not rendered.get("children_omitted"):
-            rendered.pop("children_omitted", None)
-        return rendered
+    # emitted JSON when nothing was omitted. The writer is ONE serializer on
+    # :class:`GameTreeResult`, not one per node — see it for why.
 
 
 class GameTreeParams(RelayedLiveParams):
@@ -139,6 +130,31 @@ class GameTreeParams(RelayedLiveParams):
     )
 
 
+# The shape ``game tree`` EMITS, as opposed to the shape it holds: identical to
+# :class:`GameNode` / :class:`GameTreeResult` except that `children_omitted` is
+# optional, which is the one difference the prune below makes. Declared because a
+# `model_serializer`'s return annotation is what pydantic builds the writer from
+# — see the method for why a bare mapping is not enough here. Not published: the
+# `--schema` document is generated from the models in validation mode, where a
+# serializer's return type does not appear.
+class EmittedGameNode(TypedDict):
+    """One node of the emitted runtime tree (#929)."""
+
+    name: str
+    type: str
+    path: str
+    children: list["EmittedGameNode"]
+    children_omitted: NotRequired[int]
+
+
+class EmittedGameTree(TypedDict):
+    """The emitted ``game tree`` result (#929)."""
+
+    root: EmittedGameNode
+    truncated: bool
+    omitted_nodes: int
+
+
 class GameTreeResult(BaseModel):
     """The result of ``gda game tree``: the running game's runtime scene tree (#849).
 
@@ -163,6 +179,38 @@ class GameTreeResult(BaseModel):
             "missing from this result."
         )
     )
+
+    # The ONE writer of :class:`GameNode`'s `children_omitted` presence rule
+    # (#929). It sits here, on the whole tree, because the obvious home — a
+    # `model_serializer` on the node itself — is a Python callback at EVERY
+    # level, and pydantic stops calling those at about 128 nested levels: a
+    # chain the model happily VALIDATED then raised an uncaught
+    # `PydanticSerializationError`, so the CLI exited 1 with a bare traceback
+    # and no `Error envelope` at all, while `scene get`'s plain nested model
+    # refused the same depth typed. One callback at the top runs pydantic's own
+    # serializer over the whole tree first and then walks the dumped
+    # dictionaries with an explicit stack — never recursion, which would only
+    # move the ceiling into Python. The emitted JSON is unchanged and every tree
+    # the model accepts now also serializes; past that the refusal stays the
+    # typed `tree_too_deep` (issue #37) that `game tree --help` states.
+    #
+    # The return type is spelled out rather than left as a bare mapping BECAUSE
+    # of the depth: pydantic serializes a callback's result through the schema
+    # the annotation gives it, and an unannotated one falls back to the inferred
+    # writer, whose own JSON guard stops at the same ~128 levels this method
+    # exists to clear. The annotation is what makes the tree schema-driven again.
+    @model_serializer(mode="wrap")
+    def _omit_absent_omission_counts(
+        self, handler: SerializerFunctionWrapHandler
+    ) -> "EmittedGameTree":
+        rendered = handler(self)
+        pending: list[Any] = [rendered["root"]]
+        while pending:
+            node = pending.pop()
+            if not node.get("children_omitted"):
+                node.pop("children_omitted", None)
+            pending.extend(node["children"])
+        return rendered
 
     @model_validator(mode="after")
     def _check_omission_counters(self) -> "GameTreeResult":
@@ -914,7 +962,9 @@ def game_tree(
     selected subtree is counted, never silently dropped: the result carries
     `truncated` and `omitted_nodes`, and each node whose children were not walked
     carries `children_omitted`. Read bounded first, then address the nodes you
-    want by their exact path (`game get`, `game rect`, `game set`).
+    want by their exact path (`game get`, `game rect`, `game set`). A tree
+    nesting deeper than about 250 levels is refused as `tree_too_deep`: bound
+    such a read with `--root` and `--max-depth`.
     """
     dispatch_domain(
         GAME_TREE_COMMAND,
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -89,7 +89,8 @@ class GameNode(BaseModel):
     # rather than on every caller: the field stays a plain int with a 0 default
     # (a consumer reads a number, never None), and the key is dropped from the
     # emitted JSON when nothing was omitted. The writer is ONE serializer on
-    # :class:`GameTreeResult`, not one per node — see it for why.
+    # :class:`GameTreeResult`, not one per node — see it for why; a node dumped
+    # on its own therefore keeps the key, which no gda path does.
 
 
 class GameTreeParams(RelayedLiveParams):
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -164,6 +164,22 @@ class GameTreeResult(BaseModel):
         )
     )
 
+    @model_validator(mode="after")
+    def _check_omission_counters(self) -> "GameTreeResult":
+        # The two counters ARE the partial-read contract (#929): a caller that
+        # reads `truncated` false stops looking, so a reply carrying it beside a
+        # non-zero `omitted_nodes` presents an incomplete tree as a complete one
+        # — and the reverse claims an omission the read never made. A stale or
+        # drifted harness must fail output validation and classify as
+        # contract_violation, never pass as a success. Same rule, same reason as
+        # the `input` gesture evidence (``InputTapResult``, #652).
+        if self.truncated != (self.omitted_nodes > 0):
+            raise ValueError(
+                "a tree result reports truncated true exactly when "
+                "omitted_nodes is above 0."
+            )
+        return self
+
 
 # The selectors `gda game find` ANDs together, in the order the params model
 # declares them. One authority: the model reads it to refuse a selector-less
@@ -376,6 +392,24 @@ class GameFindResult(BaseModel):
             "is the size of what was never tested against the selectors."
         )
     )
+
+    @model_validator(mode="after")
+    def _check_result_counters(self) -> "GameFindResult":
+        # `count` is published so a caller can branch on the number WITHOUT
+        # walking the list, which makes a count that disagrees with the list
+        # worse than no count at all. The bounding pair carries
+        # :class:`GameTreeResult`'s rule for the same reason it does there: a
+        # search that under-reports its omission reads as proven absence. Both
+        # are harness drift, so both fail output validation and classify as
+        # contract_violation (#929).
+        if self.count != len(self.matches):
+            raise ValueError("a find result's count is the length of matches.")
+        if self.truncated != (self.omitted_nodes > 0):
+            raise ValueError(
+                "a find result reports truncated true exactly when "
+                "omitted_nodes is above 0."
+            )
+        return self
 
 
 class GameGetParams(RelayedLiveParams):
--- a/src/gda/commands/game.py
+++ b/src/gda/commands/game.py
@@ -204,12 +204,16 @@ class GameTreeResult(BaseModel):
         self, handler: SerializerFunctionWrapHandler
     ) -> "EmittedGameTree":
         rendered = handler(self)
-        pending: list[Any] = [rendered["root"]]
+        # Both reads are defensive because `exclude` / `include` /
+        # `exclude_defaults` let a caller drop either key: the prune must then
+        # find nothing to do, not raise a KeyError the caller reads as a
+        # serialization failure.
+        pending: list[Any] = [rendered["root"]] if "root" in rendered else []
         while pending:
             node = pending.pop()
             if not node.get("children_omitted"):
                 node.pop("children_omitted", None)
-            pending.extend(node["children"])
+            pending.extend(node.get("children", ()))
         return rendered
 
     @model_validator(mode="after")
@@ -963,8 +967,9 @@ def game_tree(
     `truncated` and `omitted_nodes`, and each node whose children were not walked
     carries `children_omitted`. Read bounded first, then address the nodes you
     want by their exact path (`game get`, `game rect`, `game set`). A tree
-    nesting deeper than about 250 levels is refused as `tree_too_deep`: bound
-    such a read with `--root` and `--max-depth`.
+    nesting deeper than about 250 levels is refused (`tree_too_deep`; past about
+    500 the engine's own JSON writer cuts the reply short and the refusal is
+    `contract_violation`): bound such a read with `--root` and `--max-depth`.
     """
     dispatch_domain(
         GAME_TREE_COMMAND,
--- a/tests/live/test_e2e_game.py
+++ b/tests/live/test_e2e_game.py
@@ -14,7 +14,7 @@ import time
 
 import pytest
 
-from gda.exit_codes import EXIT_LIVE
+from gda.exit_codes import EXIT_LIVE, EXIT_PARSE
 from tests.support import PNG_1X1_B64, Gda
 
 from tests.conftest import LIVE_PROJECT_GODOT, project_godot
@@ -804,16 +804,27 @@ def test_game_tree_bounds_the_read_and_counts_what_it_left_out(
 # total, and the op published that as a successful read — the exact-count promise
 # of #849 broken on a finite, valid tree.
 DEEP_CHAIN_DEPTH = 2200
-DEEP_CHAIN_MAIN_GD = (
-    "extends Node2D\n"
-    "func _ready() -> void:\n"
-    "\tvar current: Node = self\n"
-    f"\tfor index in range({DEEP_CHAIN_DEPTH}):\n"
-    "\t\tvar child := Node.new()\n"
-    "\t\tchild.name = str(index)\n"
-    "\t\tcurrent.add_child(child)\n"
-    "\t\tcurrent = child\n"
-)
+
+
+def chain_main_gd(depth: int) -> str:
+    """The main-scene script building ONE chain `depth` nodes deep below Main.
+
+    The chain length is the only thing the depth cases vary, and each of them
+    reads a different ceiling, so the script is built rather than copied.
+    """
+    return (
+        "extends Node2D\n"
+        "func _ready() -> void:\n"
+        "\tvar current: Node = self\n"
+        f"\tfor index in range({depth}):\n"
+        "\t\tvar child := Node.new()\n"
+        "\t\tchild.name = str(index)\n"
+        "\t\tcurrent.add_child(child)\n"
+        "\t\tcurrent = child\n"
+    )
+
+
+DEEP_CHAIN_MAIN_GD = chain_main_gd(DEEP_CHAIN_DEPTH)
 DEEP_CHAIN_MAIN_TSCN = (
     "[gd_scene load_steps=2 format=3]\n\n"
     '[ext_resource type="Script" path="res://main.gd" id="1"]\n\n'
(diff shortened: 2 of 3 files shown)


qinhaihong-red:
**Review record, round 1 (orchestrator, 2026-09-09).** Independent adversarial review at `af7be5e32` (four axes; the depth ladder re-measured in-process on both checkouts and on Godot 4.6.3 at 201 / 300 / 600 levels; five mutants through a pytest plugin patching the decorators; the serializer-variant matrix re-run on replicas) returned **CHANGES REQUIRED — 0 P1 / 2 P2 / 2 P3**. The fix round is the implementer's. Fix head: `312c3593d`.

|---|---|---|
| **P2-1** `EmittedGameNode` / `EmittedGameTree` are a hand-written parallel declaration of the two models' field sets; because the annotation selects pydantic's writer, a field added to a model and forgotten there vanishes from the JSON silently (`warnings="error"` included) while `--schema` keeps publishing it | Confirmed on a replica | **Adopted.** One consistency test pins `set(Emitted*.__annotations__) == set(Model.model_fields)` for both pairs (fails on drift; the model stays the one authority for field names). |
| **P2-2** the help/schema sentence said a deeper tree "is refused as `tree_too_deep`", but past ~500 levels the engine's JSON writer cuts the reply short and the refusal is `contract_violation` (measured at 600 levels: rc 5, `diag errors` "JSON structure is too deep. Bailing.") — the 2026-09-09 re-scope requires that disclosure | Confirmed on the engine | **Adopted.** The sentence now names both: refused (`tree_too_deep`; past about 500 the engine's own JSON writer cuts the reply short and the refusal is `contract_violation`); bound the read with `--root` / `--max-depth`. Still the only description change in the schema. |
| P3-1 the prune indexed `rendered["root"]` / `node["children"]`, so `model_dump(exclude=…/include=…/exclude_defaults=True)` raised where the base returned a value | Confirmed | **Adopted** — `.get`-style reads, one test for `exclude={"root"}`; the 254 ceiling is unchanged and `warnings="error"` is silent. |
| P3-2 fold the pre-existing 2200-node test's three writes into the new `_chain_project` helper | Optional | **Declined** — no rule at stake; the older test stays as written. |

Reviewer-verified and unchanged: the mechanism is one serializer annotated correctly (the variant matrix shows per-node + TypedDict still fails at 254 and an `Any`/`dict` annotation re-imposes the 128 guard; `int | None` + `exclude_none` would change the schema), the deletion test passes, no `Emitted*` name reaches `gda schema` or `--schema`, the prune is iterative and total, a 20 000-child dump is faster (35 → 15 ms), round-trip equality holds (ADR-0023's 2026-09-08 rule), the validators match what the harness emits (`gda_harness.gd:443,509-511`), the ladder reproduces row for row (base `model_dump_json` ceiling 127, head 254 = the validation limit), the schema differs in exactly one description, `game tree --schema` / `game find --schema` / `game find --help` byte-identical, both fixtures byte-identical across three dump paths, harness diff empty at version 22, `73ebcd292` not an ancestor, AC1/AC2 reproduced independently on the engine (201 whole, 301 `tree_too_deep`, bounded exact); "about 250" judged the right disclosure because the e2e depths 200 / 300 bracket the constant. On `312c3593d`: fast tier 2633, game + harness e2e 20 under the lock, seven mutants killed, PR CI green. A second independent round runs on this head per the wave rule.

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-09).** Second independent review at `312c3593d` (the four `Emitted*` key-deletion mutants through a runtime plugin; a seeded 300-tree differential fuzz across three dump paths; seven dump-option combinations on both checkouts; the engine at 240 / 320 / 320 bounded / 600 levels under the machine lock; the human renderer, the MCP relay, `--params-json` parity and the find validators against what the harness emits) returned **READY — 0 P1 / 0 P2 / 3 P3**. Touch-up head: `6c2811b8d`.

|---|---|---|
| P3-1 the documented refusal ceiling was guarded only by the nightly e2e (`REFUSED_CHAIN_DEPTH = 300`); the fast tier's 200-level case catches the ceiling FALLING, not RISING | Confirmed | **Adopted** — a fast-tier mirror case: a 300-level reply through the fake live runner asserts `EXIT_PARSE` / `tree_too_deep`. |
| P3-2 the PR body said "a read of 128 to 254 nesting levels … exit 1"; on the base only the `--json` channel crashed (the human channel exits 0 at 200 levels) | Confirmed | **Adopted** — body wording: "a `--json` read". |
| P3-3 a bare `GameNode` dumped on its own now keeps `children_omitted: 0` (the prune lives on the tree result; no gda path dumps a node alone) | Confirmed | **Adopted** — half a sentence in the comment. |
| Note: the help sentence's length; `ge=0` on the counters | — | **Declined** — the sentence was adopted verbatim in round 1 and the re-scope asked for one; `ge=0` would add `minimum: 0` to the schema and the harness cannot emit a negative. |

Reviewer-verified and unchanged: all four `Emitted*` deletions killed by the consistency test (and the risk shown real on a replica: a missing key vanishes silently with `warnings="error"`); the whole-document schema differs only in `commands[15].description`; `game tree --schema`, `game find --schema`, `game find --help`, `gda --help`, `gda game --help` byte-identical; the ladder base 127 → head 254 = the validation limit; the PR's 200-level case fails on the base with the exact serializer error; 240 whole / 320 `tree_too_deep` / 320 bounded exact / 600 `contract_violation` with the engine's "JSON structure is too deep" line; 300 random trees byte-identical across three dump paths with round-trip equality 300/300; fast tier 2633, e2e 20 (18 + 2), ruff / format / pyright clean, harness guard valid, harness diff empty at 22; the human renderer already walks on an explicit stack; the MCP relay is a `json.loads` passthrough; `--params-json` parity untouched; the find validators refuse no reply the harness can emit; `NotRequired` is documentary — the consistency test is the sole guard. On `6c2811b8d`: PR CI green, MERGEABLE. **READY** for the maintainer's review; merges first in the W3 order.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `6c2811b8d06372f3d9f25ecc5f6450a580e92c31`. Existing reviews and replies were not used as evidence.

1. **The issue title contradicts its superseding scope.** Issue #929 is still titled “game tree reads a scene exactly at any depth,” but its 2026-09-09 re-scope explicitly leaves the ~250-level Pydantic refusal and the ~500-level engine JSON ceiling in place, and says that “exact at any depth” is out of scope. Rename the issue so boards, search, and later closure checks state the contract that this PR implements.
2. **The command catalog omits the new public depth contract.** `src/gda/commands/game.py:970-973` discloses both refusal ceilings and the `--root` / `--max-depth` remedy. `docs/command-catalog.md:1418-1431` still says only that an unbounded read is the caller's choice. Synchronize the catalog with the help.

**Terminology and prose assessment:** I inspected the complete current issue, the PR body, changed help/schema prose, and the relevant current command-catalog section against `CONTEXT.md`, ADR-0015/0017/0019, and ASD-STE100. Apart from the two contradictions above, the `Game tree`, `truncated`, `omitted_nodes`, and `children_omitted` terms are consistent and the changed prose is concrete. I did not test external links.

The command slice remains in the existing `gda` context and correctly keeps runtime-tree validation and serialization in `gda.commands.game`. However, `EmittedGameNode` / `EmittedGameTree` at `src/gda/commands/game.py:141-156` duplicate the models' names, types, nesting, and requiredness because the annotations drive Pydantic serialization. The guard at `tests/live/test_game_commands.py:186-194` checks field names only. A later type or requiredness change can therefore update validation/schema while the writer keeps a stale contract. Derive the emitted shape from the model if practical, or extend the explicit synchronization guard to the semantic dimensions that control serialization. Claim-evidence state: `open` until a drift regression proves the complete synchronization rule.

## Standards

- Hard: synchronize `docs/command-catalog.md:1418-1431` with the refusal contract, as required by `RULES.md:121-123`.
- Hard: the incompletely guarded emitted TypedDicts are a second authority for the wire shape, contrary to `RULES.md:117-120`.
- Judgment call — Duplicated Code: `truncated == (omitted_nodes > 0)` is repeated at `src/gda/commands/game.py:229-233` and `:460-464`. A small group-local predicate could prevent drift while retaining result-specific messages.

Standards summary: 3 findings; worst is a hard public-contract synchronization defect.

## Spec

No Spec finding. The superseding #929 scope is implemented: accepted deep trees serialize, counter invariants reject contradictory replies, help names the retained ceilings, and the harness is unchanged. Independent validation passed the 200/300-level real-engine cases (2 tests) and all 60 fast game-command tests.

Spec summary: 0 findings; worst: none.

qinhaihong-red:
## Third review handled at `f469e75c1`

| Issue #929's title ("reads a scene exactly at any depth") contradicts its 2026-09-09 re-scope | Fully Adopt | the re-scope leaves the ~250 and ~500 ceilings in place | Issue renamed to "game tree serializes every tree its result model accepts, and game tree / game find counters are checked" — PR #939's own title | — |
| The catalog omits the depth contract the help states | Fully Adopt | `game.py:970-973` names both ceilings and the remedy; the catalog's `game tree` entry did not | One sentence added to the catalog entry, mirroring the help | catalog is prose; help / schema unchanged |
| The emitted TypedDicts are a second authority for the wire shape, guarded on names only | Partially Adopt | true: the annotations drive pydantic's writer, and the guard compared field names alone | The guard now compares every field's TYPE and nesting against the model (`EmittedGameNode` standing in for `GameNode`), and pins the optional-key set to exactly `children_omitted`. NOT adopted: deriving the TypedDict from the model — the recursive `children: list[EmittedGameNode]` makes a dynamic TypedDict the heavier mechanism, and the TypedDict itself is required (pydantic selects the writer from the wrap-serializer's return annotation, measured in #929). Also learned and stated in the guard: writer-side "requiredness" is omission-when-absent, not the model's input requiredness — `children` has a default on input and is always emitted, so the first draft of the guard reddened on it | retyping `children_omitted` to `str` in the TypedDict reds exactly the guard; 60 game-command tests pass |
| `truncated == (omitted_nodes > 0)` repeated in two validators | Adopt | two copies of one rule | one group-local `_check_omission`, each result keeping its own subject in the message | dropping the rule from the find validator reds exactly `test_a_find_reply_whose_counters_disagree_is_a_contract_violation` |

Head `f469e75c1`, one commit on `6c2811b8d`; ruff / format / pyright clean; CI green. Spec axis had no finding.

Issue #844, opened 2026-09-05
script run: recognize engine shutdown-leak errors as diagnostics; --strict fails on them


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-063** (2026-08-26, gda 0.10.0). A fixture left callback reference cycles; Godot printed `WARNING: ObjectDB instances leaked at exit` and `ERROR: 7 resources still in use at exit`, yet `gda script run --strict` returned shell zero, nested `exit_status: 0`, and `diagnostics: []`, with both lines only inside the raw `stderr` string. The production had to add "stderr must be empty" as its own gate. The leaking fixture was uncommitted and repaired at once, so a synthetic reproduction is needed.

Verified against gda 0.13.0 (`266c8009f`) and Godot 4.6.3: `--strict` keys on the exit status only; the script-error parser (`gda.script_errors`) recognizes `SCRIPT ERROR:` records, the entry-load failures, the bind-time refusals, and a project `push_error` (#722), and by contract skips every warning. The two shutdown records differ in kind: `%d resources still in use at exit` is an `ERR_PRINT` from `core/io/resource.cpp` and carries a count; `ObjectDB instances leaked at exit (run with --verbose for details).` is a `WARN_PRINT` from `core/object/object.cpp` and carries no count. Neither is recognized, so they are neither a diagnostic nor a strict failure. The shared parser's consumers are `script run` (including its timeout path) and `scene preflight`; `script validate` has its own diagnostic channel and is not in scope.

## What to build

- Recognize the two shutdown records as one new diagnostic kind, `shutdown_leak`, on `script run` and `scene preflight`. The count stays in `message` (the resources record has one, the ObjectDB record has none); `ScriptError` gains no field — it is published on both halves of the contract (success `diagnostics` and `evidence.script_errors`), so a new field would be a two-sided contract change this issue does not make.
[...]

PR #940, opened 2026-09-09
feat(gda): daemon readiness reports the session's startup script errors (#848)

### LINKED ISSUE #848: daemon wait-ready: report recognized startup script errors from the session log
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-047** (2026-08-20, gda 0.8.0, marked high). `daemon start --windowed --scene <fixture>` returned a successful PID and the exact-scene result although the root's attached script failed to parse (`Cannot infer the type of "model"`). The first live tree held only the scriptless root; the failure was visible only in `diag errors` after a live read had established the session. Automation captured a blank frame and mistook daemon availability for scene readiness.

Verified against gda 0.13.0 (`266c8009f`): `daemon start` launches no engine session (lazy, ADR-0017). #657 added `daemon wait-ready`, whose success means the harness connected and presented its token; it says nothing about the scene's scripts. #829 refuses only an undefined or unresolved main scene. `scene validate` and `scene preflight` (#664) are the one-shot gates, but the live launch itself reports no degraded start.

## What to build

- `daemon wait-ready` (and the lazy launch a first live op triggers) reads the Session log up to the handshake through the shared script-error parser (`gda.script_errors`, imported by the daemon for the first time — no copy under `gda.daemon`) and reports what it recognized: the result gains `startup_diagnostics` (the same `ScriptError` shape `script run` and `scene preflight` emit) and `clean_start: bool`.
- Boundary with the existing log read: `startup_diagnostics` is bounded to the pre-handshake prefix of the Session log and is a convenience at the readiness boundary; `gda diag errors` remains the full-log read (ADR-0022), and the two are projections of the same daemon-owned file, not competing authorities.
- The session still serves — a broken scene is exactly when `diag errors`, `game tree`, and a screenshot are wanted — so this is a disclosure on success, not a refusal. Whether a scene whose ROOT script failed to bind (the node boots scriptless) should become a typed refusal is a follow-up decision; record it on this issue once the disclosure exists.
- `daemon status` reports the same two fields for the serving session, so a later caller reads them without a relaunch.
- The help of `wait-ready` and `daemon start --scene`, and the skill, state the readiness contract: harness connected is not scene started cleanly; read `clean_start` before capturing evidence.

## Acceptance criteria

- [ ] With a daemon started `--scene <fixture>` whose root script has a parse error, `daemon wait-ready --json` returns success with `clean_start: false` and the recognized diagnostic naming the script and line; on a clean scene, `clean_start: true` with an empty list.
- [ ] `daemon status --json` reports the same fields for the serving session.
- [ ] The parser is the shared one (no second recognizer); diagnostics carry the same shape as `script run`.
- [ ] A daemon e2e covers both scenes; help and skill are updated in the same change.
- [ ] A prefix gda cannot read (no session log, a read failure, or a launch that could not measure the bound) reports `startup_diagnostics: null` and `clean_start: null` on both commands — never `clean_start: true`. _(added 2026-09-14, see the note below)_

## Blocked by

None - can start immediately.

## Note (2026-09-14, PR #940 third review)

- The boundary is implemented as measured, not described: the launch records the Session log's size at the instant the handshake completes (token and scene verification accepted), and the verdict is read from that prefix. A record the game emits during the handshake's own frames lands on whichever side of that instant it was written; everything after it is `gda diag errors` (ADR-0022). PR #940's first two rounds had read the whole log after the launch returned and re-worded the contract to fit; the third review restored the boundary this body asks for.
- Third state, added to the contract (acceptance line above): when gda could not read that prefix, both keys are null on `wait-ready` and `status` — no verdict, never a clean start for a log nobody read. `gda diag errors` reports `live_log_unavailable` for the same condition, and the human output says so in one line. The session still serves.
- Typed-refusal follow-up (the decision this body asks to record once the disclosure exists): the recommendation is to keep it a disclosure. A refusal would take away the reads that diagnose the failure (`diag errors`, `game tree`, a capture), and `clean_start` lets a caller refuse for itself. Open for the maintainer's decision; no issue filed.

qinhaihong-red:
## Summary

`gda daemon wait-ready` returned plain success while the launched scene's root script had
failed to compile. The node booted script-less, the harness connected, and the caller read
daemon availability as scene readiness — then captured a blank frame as evidence
(GDA-DF-047). The engine had already written the parse error to the daemon-owned `Session
log`; nothing read it at the readiness boundary.

The daemon's launch boundary now reads that log UP TO THE HANDSHAKE: `launch_session`
measures the log's size at the instant the handshake completes (token and scene
verification both accepted) and the session carries that bound; the daemon reads that
prefix through the shared `gda.script_errors` recognizer. This is the daemon's first use of
that module — imported and called, never copied — so what counts as a recognized script
error stays that module's rule (which #844 amends in parallel).

The verdict is recorded where the session is established, so BOTH paths that establish one
record it: `daemon wait-ready`, and the lazy launch a first live op triggers.

## Public behavior

- `gda daemon wait-ready` success gains `startup_diagnostics` (the same `ScriptError[]`
  `script run` and `scene preflight` publish) and `clean_start`. An idempotent repeat
  (`launched: false`) reports the establishing launch's verdict, not a fresh read.
- `gda daemon status` reports the same pair for the session the daemon established, so a
  later caller reads the verdict without relaunching the game.
- **Three states, on both commands.** A list and its boolean (a session started; the list
  says what was recognized in the prefix, the boolean says whether that list is empty);
  an empty list with `clean_start: true`; and **null for both** — when no session was
  established this daemon lifetime (`status` only), when no daemon is running or the
  status round trip missed transiently (`status` only), and when gda could not read the
  log up to the handshake (no session log, a read failure, or a launch that could not
  measure the bound). Null is never a clean start for a log nobody read (ADR-0022 keeps
  unavailable apart from empty); `gda diag errors` reports `live_log_unavailable` for the
  same condition.
- **Boundary with `diag errors`:** the verdict covers the bytes the Session log held at the
  instant the harness handshake completed — engine startup, the project's autoloads and the
  scene's own scripts; a record the game emits during the handshake's own frames lands on
  whichever side of that instant it was written. Everything after it is `gda diag errors`,
  which stays the full-log read (ADR-0022): the two are projections of one daemon-owned
  file, not competing authorities. A unit test pins the bound: a recognized record written
  after the measured size is not in the verdict.
- A **disclosure on success, never a refusal.** A scene whose script did not compile still
  serves, and that is exactly when `diag errors`, `game tree` and a capture are wanted.
- Human output adds nothing on a clean start; a degraded one adds `startup not clean:` and
  one `script_error_line` per recognized error; an established session with a null verdict
  adds one line saying the log up to the handshake was not read — silence would read as a
  clean start. On both `wait-ready` and `status`.
- **The pair is ONE fact**, owned by the result models: both null, or a list and
  `clean_start` exactly "that list is empty" (`check_startup_verdict_pair`, enforced as a
  `model_validator` on `DaemonWaitReadyResult` and `DaemonStatusResult`). On `wait-ready` a
  reply that says otherwise fails output validation as `contract_violation` — the channel a
  missing key already takes; `status` degrades the same violation to the null pair through
  the same function, because a status read must not crash on a drifted daemon.
- **No new `Gda error code`** (the issue asks for none; the registry's W3 owner is #845).
  No harness change, no `HARNESS_VERSION` bump. `script_errors.py` changes only in its two
  docstring consumer lists — the daemon boundary is that module's fourth consumer and the
  renderers its fifth writer — taken over from #844's PR in round 3 because the sentences
  are true on this head only.

## Surface diff (enumerated)

Captured from a detached checkout of the base (`b7c485693`) and this branch, then diffed.

`gda schema` — 77 commands before and after, **0 removed paths, 86 added, 2 changed**, all
inside `commands[2]` (`daemon status`) and `commands[5]` (`daemon wait-ready`). Everything
outside those two command objects is byte-identical (verified by swapping the two objects
back and comparing the whole document).

1. `daemon status` output: `+ startup_diagnostics` (`anyOf: [array of ScriptError, null]`)
   and `+ clean_start` (`anyOf: [boolean, null]`), both added to `required` — the
   required-but-nullable shape `session_id` already uses.
2. `daemon wait-ready` output: the same two keys with the same required-but-nullable shape
   (null is the unread-prefix state; success always carries both keys).
3. Both outputs gain the `$defs.ScriptError` / `$defs.ScriptErrorKind` definitions — the
   models already published by `script run` and `scene preflight`, unchanged.
4. `commands[2].description` and `commands[5].description` changed: the docstring
   paragraphs below.

`gda daemon wait-ready --help` — two added paragraphs, nothing else changed:
> A connected harness is NOT a cleanly started scene: a script that fails to compile
> leaves its node script-less, and the session serves anyway. Success therefore reports
> `clean_start` and the `startup_diagnostics` gda recognized in the session log UP TO THE
> HANDSHAKE — the bytes it held at the instant the handshake completed, which the launch
> measures — so engine startup, the autoloads and the scene's own scripts; a record
> emitted during the handshake's own frames lands on whichever side of that instant it was
> written. Read `clean_start` before you treat a screenshot or a runtime read as evidence
> about the scene; both keys are null when gda could not read that prefix (`gda diag
> errors` says `live_log_unavailable`), never a clean start it did not see; and read `gda
> diag errors` for the whole log (#848).
>
> A daemon started by an OLDER gda answers without these two keys, which the CLI reports as
> `contract_violation`; run `gda daemon stop`, then `gda daemon start`, so the daemon
> serves the current contract. A mixed-version session is not a compatibility target and
> gets no negotiation — the CLI/daemon leg of ADR-0018's current-harness policy.

`gda daemon status --help` — one added paragraph, nothing else changed:
> Also reports the last engine session the daemon established: its `session_id` (#660) and
> its startup verdict — `clean_start` and the `startup_diagnostics` read at that session's
> readiness boundary — so a later caller reads what `daemon wait-ready` saw without
> relaunching the game (#848). Both are null together — when no session was established
> this daemon lifetime, when gda could not read the session log up to the handshake (`gda
> diag errors` says `live_log_unavailable`), when no daemon is running, and when the
> status round trip missed transiently.

`gda daemon start --help` — the `--scene` option help gains one sentence (the issue asks
this command's help to state the readiness contract). Not in `gda schema`: the option help
is Typer-side, while the schema publishes the params model, which is unchanged.
> Booting a scene does NOT check that it started cleanly: a script that fails to compile
> leaves its node script-less and the session still serves, so read `clean_start` on
> `gda daemon wait-ready` before you read the scene as evidence (#848).

Prose surfaces: `src/gda/skill/SKILL.md` (readiness paragraph + the `daemon` group row),
`docs/command-catalog.md` (the `wait-ready` and `status` daemon rows), and the README
`daemon wait-ready` / `daemon status` rows — one sentence each, since the boundary
"ready is not clean" is user-facing. `scripts/update_readme_i18n.py` re-stamped the three
translations, whose changed rows were translated here.

## Validation

Repository runtime throughout (`uv run --frozen`, this checkout), at head `4d1c2f015`.

| Gate | Command | Result |
| --- | --- | --- |
| Fast tier | `uv run --frozen pytest -m "not e2e" -q -p no:cacheprovider` | **2646 passed** (base `b7c485693`: 2626 — +20 unit tests) |
| Lint | `uv run --frozen ruff check .` | All checks passed |
| Format | `uv run --frozen ruff format --check .` | already formatted |
| Daemon e2e (real engine, Godot 4.6.3) | `pytest tests/daemon -m e2e -q` under the machine-wide Godot lock | **38 passed** (base: 36 — +2 e2e) |

**Red then green.** Every new test was written first and failed for the right reason:
`KeyError: 'clean_start'` on the daemon replies, `AttributeError` on the two result
models, `'startup_diagnostics' not in required` on the published schema, and the e2e's
first real-engine run failing on the diagnostic list (which is how the engine's actual
three-record output for the fixture was learned, rather than guessed). Round 3's tests
failed first against the round-2 head: `['parse_error', 'compile_failed',
'runtime_error']` for the unbounded read, `[] is None` for the unread log.

**Mutant check (round 3).** Three reverts of the new mechanism, each killed by exactly the
test written for it, then restored:
- the bound is ignored (`data` instead of `data[:bound]`) →
  `test_the_verdict_is_bounded_to_the_log_up_to_the_handshake` fails with a third kind in
  the verdict. The first version of this test used a record the recognizer does not
  classify and SURVIVED that mutant — a lesson kept in the test's comment;
- an unmeasured bound or an unreadable log returns `[]` (the round-1 behaviour) →
  `test_a_log_gda_could_not_read_is_no_verdict_not_a_clean_start` and
  `test_a_launch_that_could_not_measure_the_handshake_bound_reports_no_verdict` fail;
- the CLI accepts a contradictory pair →
  `test_status_startup_verdict_is_null_before_a_launch_or_on_drift` fails on `([], false)`.

**Real-engine evidence.** The two e2e tests boot the same scene shape with a root script
that does not compile and one that does. The broken one reproduces the dogfooding sentence
verbatim in the session log (`Parse Error: Cannot infer the type of "model" variable …` at
`res://main.gd:5`) and reports `clean_start: false` with the recognized records; the clean
one reports `clean_start: true` with an empty list — which also proves the real launch
measures a bound and reads a readable prefix, since an unmeasured one would report null.
`daemon status` returns the identical list for the serving session.

PR CI runs **no Godot e2e** (the `godot-e2e` workflow is schedule/dispatch-only), so what
this PR proves on CI and what needs the real engine differ:

**Runs on PR CI** — the fast tier, which covers every seam below the engine:
- the daemon serve loop and its launch boundary, against the injected `SessionLaunch` seam
  with a hand-written Session log and a hand-measured bound
  (`tests/daemon/test_daemon_server.py`, `tests/daemon/test_daemon_socket_lifecycle.py`):
  the verdict is recorded on a lazy launch as well as on `wait-ready`, is null before any
  session, is bounded to the measured prefix (a recognized record written after it is
  excluded), is null for an unreadable log or an unmeasured bound, is not re-read later,
  is replaced by a relaunch and retained across a failed replacement;
- the CLI mapping of the control reply, including the drift/degrade-to-null rule and the
  contradictory pair (`tests/daemon/test_daemon_lifecycle_payload.py`), and the null
  verdict passing the LIVE seam as success
  (`tests/daemon/test_daemon_wait_ready_commands.py`);
- the published schema shape (`--schema` required/nullable assertions) and all three
  human renderings (`tests/cli/test_render.py`,
  `tests/daemon/test_daemon_wait_ready_commands.py`);
- lint, format and pyright.

**Needs the real engine** (run locally here, under the machine-wide Godot lock, green):
- that Godot actually writes the parse error to the `--log-file` Session log, and flushes
  it, BEFORE the harness completes its handshake — which is what makes the measured
  prefix cover the scene's own scripts at all. No fake can establish this;
- that a scene whose root script fails to compile still loads, still sets `current_scene`,
  and still lets the harness connect, so the disclosure has a serving session to ride on;
- that the shared recognizer classifies this engine build's actual sentences
  (`parse_error` + `compile_failed`, naming `res://main.gd:5`).

The e2e assertions deliberately do not pin the NUMBER of records the engine prints for one
broken declaration (4.6.3 prints three), only that every record names the scene's script,
that both kinds appear, and that a `parse_error` carries the reported line.

## Review round 1

CHANGES REQUIRED, 0 P1 / 2 P2 / 6 P3. Every finding adopted; head `d99d2357c` is one
further commit, no force-push. The reviewer confirmed each acceptance criterion against the
real engine, so nothing below changes a mechanism.

| Finding | Resolution |
| --- | --- |
| **P2-1** the published description claims the read is "bounded to the pre-handshake prefix … and stops there" — falsified 3/3 against a fixture erroring on its first three frames | **Adopted, wording only** — and SUPERSEDED in round 3, which implements the boundary instead: the falsified claim was the description of an unbounded read, not a wrong requirement. |
| **P2-2** the remembered verdict's lifecycle is unpinned; 2 realistic mutants survive the suite | **Adopted, two tests** in `tests/daemon/test_daemon_socket_lifecycle.py`, beside the two that pin `session_id`'s lifecycle. `test_a_relaunch_replaces_the_previous_sessions_startup_verdict` (broken log on launch 1, clean on launch 2, session killed between) kills mutant 5 — `if self._startup_diagnostics is None:` — with `assert False is True`. `test_a_failed_replacement_launch_retains_the_previous_startup_verdict` (the fake truncates the log on the failing attempt, as the real launcher does) kills mutant 6 — the write moved above `if self._session is None: return None` — with `assert True is False`. Each mutant was applied, the whole daemon suite run, and only its own test failed; restored after. |
| **P3-1** catalog says the null pair is "exactly when" no session was established | Adopted: the catalog now enumerates the cases the field description already listed. The same overstatement in `daemon status --help` was fixed too. |
| **P3-2** SKILL.md "showing an empty tree and a blank frame" over-generalizes | Adopted verbatim: "leaving the tree without whatever the script would have built, and often a blank frame". |
| **P3-3** the e2e boots the fixture as `main_scene`, the AC names `--scene <fixture>` | Adopted the stronger option: the startup-verdict e2e now starts with `--scene res://main.tscn`, with a comment saying the two boot the same scene here but the selector also runs the launch-boundary scene verification, so the disclosure is proven on the path GDA-DF-047 used. |
| **P3-4** an unreadable log reads as `clean_start: true` | Adopted as a wording change in round 1 — SUPERSEDED in round 3, which makes the unread prefix its own (null) state. |
| **P3-5** record the typed-refusal follow-up on #848 | Recorded on #848 in round 3 (dated note), with the recommendation to keep it a disclosure: a refusal would take away the reads that diagnose the failure, and `clean_start` already lets a caller refuse for itself. |
| **P3-6** `script_errors.py`'s docstring still says "the four places that write one"; the daemon renderers make five | Taken over here in round 3 (see Public behavior); #844's PR no longer carries the daemon sentences. |

## Review round 2

CHANGES REQUIRED, 0 P1 / 1 P2 / 3 P3, with all eight round-1 fixes confirmed against their
mutants. One finding needed a change; head `b0c32e26c` is one further commit, no
force-push, no mechanism change.

| Finding | Resolution |
| --- | --- |
| **P2-1** the Open question claimed the version-skew asymmetry is "inherited, not introduced here" — disproved on the base | **Adopted; my claim was wrong.** On `b7c485693` `DaemonWaitReadyResult` is `{pid, launched}`, and a daemon old enough to miss those keys does not know the `daemon-wait-ready` op at all, so no "serves the op but omits a required key" state was reachable. Adding two required keys makes it reachable: a daemon is long-lived per project, a repeat `daemon start` against a running one only answers `already_running`, and there is no CLI/daemon version handshake — so upgrading gda mid-session leaves the older daemon serving and `wait-ready` exits 5 with `contract_violation` over a raw validation dump, which reads as a gda defect for a one-command operator fix. `daemon wait-ready --help` (and so its `gda schema` description) and the `docs/command-catalog.md` daemon rows now name the symptom and the remedy: `gda daemon stop`, then `gda daemon start`. The Open question below is rewritten. **Fields stay required, no handshake added** — a mixed-version session is not a compatibility target, the CLI/daemon leg of ADR-0018's 2026-09-08 current-harness policy, cited in the same sentence. No follow-up issue filed. |
| **P3-1** the first commit's now-corrected "pre-handshake prefix" wording could reach `main` in a concatenated squash body | **Declined with evidence** (orchestrator): the repo sets `squash_merge_commit_message: BLANK` and the merge procedure passes an explicit empty body and verifies it after every squash — all six W2 squashes on `main` carry a one-byte body. Branch history is discarded. |
| **P3-2** `script_errors.py`'s module docstring lists three consumers and misses the daemon readiness boundary | Taken over here in round 3. |
| **P3-3** the human render keys on the list being empty while `--json` keys on `clean_start`, so a drifted pair would render inconsistently | Declined in round 2 — REVERSED in round 3: the CLI now degrades a contradictory pair to null, so the two renderings cannot disagree. |

## Review round 3 (independent, base to head `b0c32e26c`)

CHANGES REQUIRED on all four axes. Handled by backtrace rather than a third wording patch:
round 1 had reduced the acceptance to fit the implementation (the issue asks for the log
up to the handshake; the code read the whole log after the launch returned; when
measurement showed post-handshake records inside, the WORDS were changed, not the read),
and an unreadable log collapsed into the empty state. Heads `2604602d7` + `4d1c2f015`, two
further commits, no force-push. Every finding adopted; they are two underlying claims.

| Artifact / Architecture 1 / Spec 1: the implementation is not bounded to the issue's pre-handshake prefix; the prose published the contrary behaviour as guidance; no owner for the boundary | **Fully Adopt** | `server.py` read the whole file after `launch_session` returned; the round-1 reply admitted first-frame records | `launch_session` measures the Session log's size at the instant the handshake completes and the session carries it (`handshake_log_size`, on the `SessionHandle` protocol); the daemon reads `data[:bound]`. Help, both field descriptions, catalog, SKILL, docstrings say "up to the handshake" again, with the honest edge (a record emitted during the handshake's own frames lands on whichever side it was written) | `test_the_verdict_is_bounded_to_the_log_up_to_the_handshake` (a recognized runtime error written after the bound is excluded); the unbounded-read mutant killed; clean-scene e2e proves the real launch measures a bound |
| Architecture 2 / Standards hard 1 / Spec 2: an unreadable log returns `[]` and so `clean_start: true`, contrary to ADR-0022 | **Fully Adopt** | reproduced by the reviewer's probe: `{'startup_diagnostics': [], 'clean_start': True}` for a nonexistent log | third state: `_UNAVAILABLE` in the daemon (no bound measured, no log, or a read failure) → both keys null on `wait-ready` and `status`; `DaemonWaitReadyResult` fields nullable; human renderers add one line for an established session with no verdict; a serving session is not refused | `test_a_log_gda_could_not_read_is_no_verdict_not_a_clean_start`, `test_a_launch_that_could_not_measure_the_handshake_bound_reports_no_verdict`, the CLI pass-through and render tests; the `[]` mutant killed |
| Standards hard 2: `_startup_verdict()` validates types only; a contradictory pair is accepted | **Fully Adopt** | no test pinned it; the round-2 decline rested on "same deployment" | the CLI degrades `clean != (not list)` to null — half a verdict is no verdict | `([], false)` and `([recognized], true)` in `test_status_startup_verdict_is_null_before_a_launch_or_on_drift`; the accept-anything mutant killed |
| Claim-evidence state `open` until missing/unreadable-log and post-handshake-record regressions pass | **Fully Adopt** | — | both regressions exist and fail against the round-2 head | closed |

Re-verified after the round: fast tier **2646 passed**, `ruff check` / `ruff format
--check` / `pyright` clean, daemon e2e **38 passed** under the lock. Against `b7c485693`
the schema diff is **86 added / 0 removed / 2 changed** (the two new leaf paths are the
`null` branches of the `wait-ready` fields), confined to `commands[2]` and `commands[5]`,
everything outside byte-identical. All three `--help` diffs stay additive. README
untouched this round, so the i18n stamp is still in sync (its gate runs in the fast tier).

## Review round 4 (recheck at `4d1c2f015`) — handled at `b08ab2d7b`

The bound and the unavailable state were confirmed; one invariant remained: `DaemonWaitReadyResult` had no cross-field validator, and the round-3 reply's "the CLI validates the result model" established nothing. Short backtrace: "the pair is one fact" had been a property of the daemon's computation and of the tolerant status adapter, never of the published VALUE, on a "same deployment" premise this PR's own round 2 had falsified (the CLI/daemon skew is reachable). Fully adopted: `check_startup_verdict_pair` states the rule once; both result models enforce it; `wait-ready` refuses (`contract_violation`), `status` degrades to null through the same function — one rule, two documented failure behaviours. Tests: four contradictory fake replies on `wait-ready` → `contract_violation`; the same four on each model raise, the consistent pairs construct. Mutants: the wait-ready validator emptied → 12 tests fail; the adapter bypassing the check → the status drift test fails. Schema shape unchanged.

## Open question

**This slice makes CLI/daemon version skew reachable for the first time.** Both
`wait-ready` fields are required, matching `launched` (#657). Before this PR the result
was `{pid, launched}` and a daemon that lacked them did not serve the op at all, so the
skew had no reachable state; now an older running daemon answers the op without the two
keys and the CLI reports `contract_violation`. That is deliberate — the help and the
catalog name the remedy (`gda daemon stop`, then `gda daemon start`), and a mixed-version
session is not a compatibility target per ADR-0018's 2026-09-08 policy. What is left open
is only whether the error's PRESENTATION should improve: today it is exit 5 with a raw
validation dump, and a caller has to read the prose to learn that restarting the daemon
fixes it. Not this slice's to change — the shared classifier owns that shape.

Closes #848


qinhaihong-red:
**Review record, round 1 (orchestrator, 2026-09-09).** Independent adversarial review at `1603d8e66` (four axes; the real engine for every acceptance criterion — a broken root script, a clean one, an error two seconds after startup, a relaunch after an edit, no daemon, the `--scene` selector, the lazy first-live-op launch; six mutants) returned **CHANGES REQUIRED — 0 P1 / 2 P2 / 6 P3**. The implementation was judged behaviourally correct; both P2s were a wording pass and one test. Fix head: `d99d2357c`.

|---|---|---|
| **P2-1** the published description ("bounded to the log's pre-handshake prefix … and stops there") is falsified: the read happens after `_launch` returns, so a `push_error` on the game's frame 2 lands in `startup_diagnostics` (3/3 runs on a fixture erroring on its first three frames). The slice is otherwise right — a `_ready` error is included, an error two seconds later is not | Confirmed on the engine | **Adopted as wording, no mechanism.** Seven sites now say what the code does: read ONCE right after the harness handshake, everything the log held at that instant, so it may include the game's first frames; the whole log is `diag errors`. `grep -rn "pre-handshake"` is empty; the test comment now pins the ONCE, not a prefix. An offset captured at the handshake would still not be exact (the engine's flush boundary) and is not worth the mechanism. |
| **P2-2** the remembered verdict's lifecycle was unpinned while `session_id`'s is pinned twice; two realistic-regression mutants survived the whole suite — never refreshing on relaunch (the retired session's verdict is reported after an edit) and writing above the `if self._session is None` guard (a failed replacement clobbers it) | Confirmed (both mutants passed 2639 tests) | **Adopted.** Two tests beside the `session_id` pins: a relaunch replaces the previous verdict; a failed replacement retains it. Each mutant now fails exactly its own test. |
| P3-1 the catalog claimed the pair is null "exactly when" no session was established — a biconditional the CLI does not honour (also null with no daemon, and on a missed status round trip) | Confirmed | **Adopted** in the catalog and in `daemon status --help`, which carried the same overstatement. |
| P3-2 SKILL's "showing an empty tree and a blank frame" over-generalizes — a failed root script drops only the script, the authored nodes still load | Confirmed on the engine | **Adopted** — reworded. |
| P3-3 the e2e booted the fixture as `main_scene` where the acceptance criterion names `--scene` | Confirmed equivalent | **Adopted the stronger option** — the e2e now starts with `--scene res://main.tscn`, which also exercises the launch boundary's scene verification. |
| P3-4 an unreadable session log reads as `clean_start: true` while `diag errors` answers `live_log_unavailable` for the same condition | Confirmed | **Adopted as disclosure** — both `clean_start` descriptions carry the clause; a third state would be a contract wart. |
| P3-5 the issue asks that the follow-up decision (typed refusal for a root script that fails to bind) be recorded on #848 | — | **Orchestrator's at merge**; the PR body keeps the recommendation (stay a disclosure — a refusal removes the reads that diagnose the failure, and `clean_start` lets a caller refuse for itself). |
| P3-6 `script_errors.py`'s "the four places that write one" is now five | Confirmed | **Recorded** as a follow-up; that file is #844's this wave and stays untouched here. |

Reviewer-verified and unchanged: schema 0 removed / 84 added / 2 changed, all inside `daemon status` and `daemon wait-ready`, with the required-but-nullable shape `session_id` uses on status and the non-nullable pair on wait-ready; the three helps each gain one paragraph; README's two rows and the three translations re-stamped to the README's own digest, and the renderings judged accurate; fast tier 2639 (base 2626) and daemon e2e 38 (base 36); both PR mutants reproduced plus two more; the verdict is written in exactly ONE place on the shared `_ensure_session` boundary, so a lazy first-live-op launch and `wait-ready` record the same thing, and `diag errors` stays the full-log read; `script_errors.py`, `src/gda/harness/`, `error_codes.py` and `docs/adr/` have empty diffs at `HARNESS_VERSION = "22"`; ~75 new lines, no new error code, no third state. On `d99d2357c`: fast tier 2641, daemon e2e 38 under the lock, both mutants killed, the schema diff unchanged apart from description strings, PR CI green. A second independent round runs on this head per the wave rule.

qinhaihong-red:
**Review record, round 2 (orchestrator, 2026-09-09).** Second independent review at `d99d2357c` (all eight round-1 fixes re-checked with their own mutants applied in overlay copies; the real engine probed for an error only at frame 30, a failing autoload, errors on frames 1..5, an error on EVERY frame, a session log deleted mid-session, and the stop / start / no-daemon states; the MCP tool surface built from the head schema) returned **CHANGES REQUIRED — 0 P1 / 1 P2 / 3 P3**. Fix head: `b0c32e26c`, verified directly by the orchestrator (no third round — the budget is two).

|---|---|---|
| **P2-1** the PR body called the CLI/daemon version-skew asymmetry "inherited, not introduced here". On the base, `DaemonWaitReadyResult` is `{pid, launched}`, and a daemon old enough to miss those keys does not serve the `daemon-wait-ready` op at all — so no "serves the op but omits a required key" state was reachable before this slice. It is reachable now: a daemon is a long-lived per-project process, a repeat `daemon start` only answers `already_running`, and the CLI/daemon leg has no version handshake, so upgrading gda while a daemon runs makes `wait-ready` exit 5 with `contract_violation` over a raw validation dump | Confirmed by driving the CLI with the base's payload | **Adopted as prose, no mechanism.** `daemon wait-ready`'s docstring (hence its published schema description and `--help`) and the catalog's daemon rows now name the symptom, the `gda daemon stop` / `gda daemon start` remedy, and the standing decision: a mixed-version session is not a compatibility target and gets no negotiation — the CLI/daemon leg of ADR-0018's 2026-09-08 current-harness policy. No optional fields, no handshake, no follow-up issue. |
| P3-1 the first commit's message still asserts the falsified "bounded to the pre-handshake prefix", which a concatenated squash body would carry to `main` | Confirmed in the branch history | **Declined with evidence** — this repository sets `squash_merge_commit_message: BLANK` and the merge procedure passes an explicit empty body and verifies it after every squash (all six W2 squashes on `main` carry a one-byte body). Branch history is discarded. |
| P3-2 `script_errors.py`'s module docstring lists three consumers and misses the daemon readiness boundary, beside the "four places" line round 1 already noted | Confirmed | **Recorded** — the PR's follow-up row now names BOTH stale lists; that file belongs to #844 this wave and stays untouched. |
| P3-3 the human render keys on the list being empty while `--json` keys on `clean_start`, so a drifted reply carrying `false` with an empty list would render inconsistently | Confirmed as unreachable in practice | **Declined** — `clean_start` is computed in one place from the same recognition, and unlike a harness reply this one comes from gda's own Python in the same deployment; a disagreement is a bug the tests catch, not a runtime state to validate. |

Reviewer-verified and unchanged: `grep -rn "pre-handshake" src/ docs/ tests/` empty, with the two schema field descriptions checked by dumping `gda schema` rather than reading the source; both lifecycle tests sit beside the `session_id` pins and each dies to its own mutant (228 passing otherwise); the catalog and `daemon status --help` enumerate all three null cases; the SKILL wording; the e2e's `--scene res://main.tscn`; the unreadable-log clause on both `clean_start` descriptions, with the real-engine split confirmed (a log deleted mid-session still reports the snapshot while `diag errors` answers `live_log_unavailable`); schema 0 removed / 84 added / 2 changed, all inside the two daemon commands, byte-identical elsewhere; MCP builds all 77 tools from the head schema and both output schemas validate real samples; `--params-json` parity; the translations genuinely rendered. New engine evidence worth keeping: an error only at frame 30 is absent from the snapshot and present in `diag errors`; a failing autoload yields three records all naming it; errors on frames 1..5 leave the first two in the snapshot; an error on EVERY frame still leaves two while `diag errors` holds 529 — the list has no unbounded-payload risk. Orchestrator verification on `b0c32e26c`: the skew sentence present in `--help`, the catalog and the published schema description; `src/gda/harness/`, `script_errors.py`, `error_codes.py` and `docs/adr/` diffs empty at `HARNESS_VERSION = "22"`; fast tier 2641; ruff / format / pyright clean; the never-refresh mutant fails exactly `test_a_relaunch_replaces_the_previous_sessions_startup_verdict`; daemon e2e **38 passed** under the machine lock on Godot 4.6.3. **READY** for the maintainer's review.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `b0c32e26c35458102903378a31c3aa683eb38aa1`. Existing reviews and replies were not used as evidence.

Issue #848 requires `startup_diagnostics` to be bounded to the **pre-handshake prefix** and requires callers to use `clean_start` before accepting runtime evidence. The changed help/schema prose instead says that the read happens after the handshake, can include first-frame records, and that an unreadable log still means `clean_start: true` (`src/gda/commands/daemon.py:320-347`). These are material contract changes, not clarifications. Implement the issue boundary and preserve an unknown/unavailable state; do not publish the contrary behavior as guidance.

**Terminology and prose assessment:** I inspected the full issue, PR body, changed README/translations, command catalog, skill, help/schema descriptions, and daemon docstrings against `CONTEXT.md`, ADR-0017/0018/0020/0022/0023, and ASD-STE100. `readiness`, `startup diagnostics`, and `clean start` are otherwise used consistently. The two inaccurate causal statements above are substantive prose findings. I did not measure the engine's log-flush timing on every platform.

The daemon is the correct owner of the Session log and the remembered session verdict. Two load-bearing facts nevertheless have no faithful state/boundary:

1. `src/gda/daemon/server.py:629-653` snapshots the whole file after `launch_session` returns, so there is no owner for the required pre-handshake offset. First-frame errors can be mislabeled as startup errors.
2. Lines 649-652 collapse log unavailability into the empty evidence state; lines 622-626 then derive `clean_start: true`. This violates ADR-0022's explicit unavailable-versus-empty distinction.

Capture the handshake log boundary and preserve availability (for example, a nullable verdict or a typed availability field) without turning a serving session into a refusal. Claim-evidence state for both guarantees: `open` until missing/unreadable-log and post-handshake-record regressions pass.

## Standards

- Hard: the unreadable-log collapse above can authorize screenshots/runtime reads without readable evidence, contrary to ADR-0022 and `RULES.md:84-88,117-123`.
- Hard: `_startup_verdict()` at `src/gda/commands/daemon.py:1026-1045` validates only types, not `clean_start == not startup_diagnostics`. A contradictory status reply is accepted; `DaemonWaitReadyResult` also has no equivalent invariant. Reject the pair or derive the boolean from the validated list.

No additional baseline smell warranted reporting.

Standards summary: 2 hard findings; worst is false clean-start evidence.

## Spec

- Required: the implementation is not bounded to the issue's pre-handshake prefix. `server.py:636-643` explicitly admits first-frame records, making `clean_start` timing-dependent.
- Required: a missing/unreadable log returns `[]` at `server.py:649-652`, then `clean_start: true`. A direct exact-head probe reproduced `{'startup_diagnostics': [], 'clean_start': True}` for a nonexistent log, while `diag errors` correctly reports `live_log_unavailable` for the same condition.

Independent validation passed 19 focused startup/wait-ready tests; none covers either failure path.

Spec summary: 2 findings; worst is unsupported readiness evidence.

qinhaihong-red:
Third review handled by backtrace, not a third wording patch. The two rounds before it had reduced the acceptance to fit the implementation: issue #848 asks for the log UP TO THE HANDSHAKE, the code read the whole file after `launch_session` returned, and when measurement showed post-handshake records inside, the words were changed instead of the read. The second defect — an unreadable log collapsing into `[]` and so `clean_start: true` — fell out of the same "fit the words to the code" move. Both are fixed at the mechanism. Exact head `4d1c2f015` (`2604602d7` + `4d1c2f015`, two commits over `b0c32e26c`, no force-push). Every finding adopted; they are two underlying claims plus one invariant.

| Artifact / Architecture 1 / Spec 1 — the implementation is not bounded to the pre-handshake prefix, the prose publishes the contrary behaviour as guidance, and no owner holds the boundary | **Fully Adopt** | `server.py` snapshotted the whole file after the launch returned; round 1's reply admitted first-frame records | `launch_session` measures the Session log's size at the instant the handshake completes (token and scene verification both accepted) and the session carries it — `handshake_log_size`, on the `SessionHandle` protocol — so the instant is captured where it is known, once; the daemon reads `data[:bound]`. Help, both field descriptions, the catalog, the skill and the docstrings say "up to the handshake" again, with the one honest edge: a record emitted during the handshake's own frames lands on whichever side of that instant it was written | `test_the_verdict_is_bounded_to_the_log_up_to_the_handshake`: a recognized runtime error written after the measured size is not in the verdict. Mutant `data` instead of `data[:bound]` → that test fails with a third kind. (Its first version used a record the recognizer does not classify and survived the mutant; corrected in `4d1c2f015`.) The clean-scene e2e reports `[]` / `true`, which under the new rule proves the real launch measures a bound and reads a readable prefix |
| Architecture 2 / Standards hard 1 / Spec 2 — an unreadable log returns `[]`, then `clean_start: true`, against ADR-0022's unavailable-versus-empty distinction | **Fully Adopt** | your probe: `{'startup_diagnostics': [], 'clean_start': True}` for a nonexistent log | third state. In the daemon, no measured bound, no log, or a read failure is `_UNAVAILABLE`, and the verdict is null for both keys on `wait-ready` and `status`; `DaemonWaitReadyResult`'s two fields become required-but-nullable (the shape `status` already used); the human renderers add one line for an established session with no verdict, since silence would read as a clean start; `diag errors` names the same condition `live_log_unavailable`. A serving session is not refused | `test_a_log_gda_could_not_read_is_no_verdict_not_a_clean_start` (both commands) and `test_a_launch_that_could_not_measure_the_handshake_bound_reports_no_verdict`; the LIVE-seam pass-through and human-render tests in `test_daemon_wait_ready_commands.py` / `test_render.py`. Mutant `[]` for the unread cases → both server tests fail |
| Standards hard 2 — `_startup_verdict()` validates types only; a contradictory pair is accepted, and `DaemonWaitReadyResult` has no equivalent invariant | **Fully Adopt** | no test pinned it; round 2's decline rested on "same deployment" | the CLI degrades `clean != (not list)` to null: the pair is one fact and half a verdict is no verdict. For `wait-ready` the daemon derives the boolean from the list in one place and the CLI validates the result model; the status path is where a drifted reply is tolerated, so that is where the invariant is checked | `([], false)` and `([recognized], true)` added to `test_status_startup_verdict_is_null_before_a_launch_or_on_drift`. Mutant "accept any pair" → that test fails |
| Claim-evidence state `open` until missing/unreadable-log and post-handshake-record regressions pass | **Fully Adopt** | — | both regressions exist and were red against `b0c32e26c` (`['parse_error', 'compile_failed', 'runtime_error']`; `[] is None`) | closed |

Also in this round: the `script_errors.py` docstring consumer lists (module docstring: four consumers; `script_error_line`: five writers) are taken over from #844's PR #964, since the sentences are true only on this head; the round-2 decline of P3-3 is reversed by the invariant above; the round-1 P3-5 follow-up is recorded on #848 as a dated note (with the acceptance line for the third state); the PR body is rewritten to the delivered contract.

CI is green on `4d1c2f015`. Locally: fast tier 2646 passed, `ruff check` / `ruff format --check` / `pyright` clean, daemon e2e 38 passed against Godot 4.6.3 under the machine-wide lock. Schema against `b7c485693`: 86 added / 0 removed / 2 changed, confined to `daemon status` and `daemon wait-ready`, everything outside byte-identical. Not verified: the engine's log-flush timing relative to the handshake on platforms other than macOS — the e2e pair establishes it on this host only, and the bound makes the verdict independent of how long the launch takes to RETURN, not of when the engine flushes.

qinhaihong-red:
## Follow-up recheck — one invariant remains open at exact head `4d1c2f015`

The pre-handshake byte bound and the unavailable-versus-empty state are now implemented and independently rechecked. The remaining part of the earlier finding is the `daemon wait-ready` result invariant.

`DaemonWaitReadyResult` still has no cross-field validator. Direct model validation at this exact head accepts all three contradictory states:

```text
startup_diagnostics=None, clean_start=True
startup_diagnostics=[],   clean_start=None
startup_diagnostics=[],   clean_start=False
```

This matters at the command boundary: `DAEMON_WAIT_READY_COMMAND.output_model` is `DaemonWaitReadyResult`, so the reply's model validation checks the two field types but neither rejects nor normalizes an inconsistent pair. The response's statement that “the CLI validates the result model” therefore does not establish the claimed invariant. The status adapter now checks `clean != (not recognized)` and degrades drift to the null pair; the wait-ready path has no equivalent check.

Required closure: enforce one contract at the wait-ready output boundary — either both values are null, or diagnostics is a list and `clean_start == (not diagnostics)` — and add fake-reply cases for both boolean contradictions and the mixed-null states. A deliberate difference from the status behavior would instead need an explicit contract decision.

Independent validation: 47 focused daemon tests and the clean/degraded real-Godot E2E pair pass. Current CI is green with no pending jobs; none of those tests supplies a contradictory wait-ready reply.

qinhaihong-red:
## Fourth review handled at exact head `b08ab2d7b`

The remaining finding is the third round on one invariant, so it got a short backtrace before the change. Root cause: "the pair is one fact" was implemented as a property of the daemon's computation (it derives the boolean from the list in one place) and, in round 3, of the tolerant status adapter — never of the published VALUE. Round 2's decline rested on "same deployment", a premise this PR's own round-2 finding had already falsified: the CLI/daemon skew is reachable, and a second deployment is exactly where the daemon's computation stops being evidence. The invariant now has one owner, the result models.

| `DaemonWaitReadyResult` has no cross-field validator; model validation accepts `(None, true)`, `([], None)`, `([], false)`; "the CLI validates the result model" established nothing | **Fully Adopt** | reproduced at `4d1c2f015`; the previous reply's claim was wrong | `check_startup_verdict_pair` states the rule once — both null, or a list and `clean_start` exactly "that list is empty" — and both `DaemonWaitReadyResult` and `DaemonStatusResult` enforce it as a `model_validator`. On `wait-ready` a contradictory reply now fails output validation as `contract_violation`, the channel a missing key already takes (help and catalog name it beside the older-daemon case); the status adapter degrades the same violation to the null pair THROUGH the same function, because a status read must not crash on a drifted daemon | `test_wait_ready_rejects_a_contradictory_verdict_pair` (four fake replies: both boolean contradictions and both mixed-null states → `contract_violation`); `test_both_result_models_own_the_one_verdict_rule` (the same four on each model raise, the two consistent pairs construct); the existing status drift test keeps its `([], false)` / `([recognized], true)` cases. Mutants: the wait-ready validator emptied → 12 tests fail; the adapter bypassing the shared check → the status drift test fails |
| A deliberate difference from the status behaviour would need an explicit contract decision | **Fully Adopt** | — | the difference is deliberate and now stated in the function's docstring and the field description: `wait-ready` refuses (output validation), `status` degrades to null — one rule, two documented failure behaviours, both computed by the same function | schema unchanged in shape (both keys still required-but-nullable; a model validator publishes nothing) |

CI is green on `b08ab2d7b`. Locally: daemon + cli + meta tiers 869 passed, `ruff` / `pyright` clean.

PR #964, opened 2026-09-09
feat(gda): script run --strict fails a run the engine reports leaking at exit (#844)

### LINKED ISSUE #844: script run: recognize engine shutdown-leak errors as diagnostics; --strict fails on them

qinhaihong-red:
## Summary

Godot reports leaked OBJECTS and RESOURCES at engine exit with two records — `WARNING: ObjectDB
instances leaked at exit (run with --verbose for details).` (a `WARN_PRINT` from
`ObjectDB::cleanup`) and `ERROR: <n> resources still in use at exit (run with --verbose for
details).` (an `ERR_PRINT` from `ResourceCache::clear`) — and `gda.script_errors` recognized
neither. A suite could print a pass, `quit(0)` and leave its objects alive, and `gda script run
--strict` still returned shell zero with `exit_status: 0` and `diagnostics: []`; both records were
visible only inside the raw `stderr` string, so the production gated on "stderr must be empty"
instead (GDA-DF-063, gda 0.10.0).

This PR classifies both records as one new `shutdown_leak` diagnostic kind and gives `--strict` a
second trigger. No new error code, no new `FailureEvidence` field, no new evidence producer.

## Public behavior

- **`script run` (and `scene preflight`) report a `shutdown_leak` diagnostic** for each engine leak
  record. The count stays in `message` (the resources record has one, the ObjectDB record has none);
  `ScriptError` gains no field, because that model is published on both halves of the contract.
  Neither sentence names a `res://` resource, so `path` / `line` are `null`.
- **`script run --strict` now fails on either of two triggers**: the non-zero exit status, or a
  `shutdown_leak` diagnostic — so a run that exits `0` and leaks is `script_failed` (exit 4). Without
  `--strict` the diagnostic stays data on the success result, as #651 decided for everything the
  script survived.
- **One verdict, one producer.** The widened rule reuses `script_exit_status_failure`: the same
  registered `script_failed` code, the same `evidence.exit_status` (the child's `0`) and
  `evidence.script_errors`, the same place in ADR-0004's producer set
  (`tests/cli/test_error_registry.py` unchanged). Only the MESSAGE branches, because a zero status
  cannot explain why a strict gate fired: `script run --strict: res://leaky.gd exited with status 0,
  but the engine reported a leak at exit — ObjectDB instances leaked at exit (run with --verbose for
  details).` The message says the ENGINE reported a leak, not that this script leaked: the engine
  reports what the whole process still held at exit, autoloads included. A run that trips both
  triggers keeps the status message; the leak is on the evidence and in the diagnostics either way.
- **`scene preflight` reports the record as DATA and does NOT gate `started` on it.** The same
  parser over the same captured stderr, so a leaking scene carries the `shutdown_leak` entries in
  `diagnostics` — but `started` answers how the BOOT went, while the engine prints this record after
  the run and about the whole process. Gating on it would report a scene whose nodes carry no script
  at all as `started: false` because one of the project's autoloads leaked. Both derivations of
  `started` (the ordinary verdict and the splash-quit route) ask one shared reading, so the field
  cannot mean two things by the route that produced it. BOTH renderings carry the records: `--json`
  lists them in `diagnostics`, and the human channel prints them under the plain `ready` headline
  (round 2). A record that does not gate the verdict still has to be visible in a terminal — that is
  what GDA-DF-063 was filed for — so the one short line is now keyed on `started` AND an empty
  `diagnostics`, and `ready with errors` keeps meaning `started: false`. No new headline word.
- **The closed set admits its first WARNING record**, argued in the module docstring against #722's
  three admission criteria: (1) both sentences are C++ format-string literals, so nothing keys on
  project prose — a `push_warning` that spells the same words is skipped by its frame, the mirror of
  the guard the error level already had via `_push_error`; (2) the record says what became of the
  objects and resources of the PROCESS this run was, which an agent branches on; (3) the kind states
  that the PROCESS REACHED SHUTDOWN — nothing about whether the entry script ran — so it stays out of
  `_ENTRY_FAILURE_PRECEDENCE` and, naming no resource, could not decide an entry verdict even if it
  were in it. The warning LEVEL is still skipped.
- **Two boundaries the set states rather than leaves implicit.** The frame guard covers the
  `push_warning` / `push_error` BUILTIN paths; a script that writes the sentence straight to stderr
  with `printerr` is not keyed out — the same pre-existing opening every recognized sentence has,
  new only in that forging THIS kind flips `--strict`. And the two records are not everything Godot
  says about leaks at exit: it also reports leaked RIDs
  (`servers/rendering/renderer_canvas_cull.cpp` `WARN_PRINT`, and
  `core/templates/rid_owner.h` `print_error`, the latter not behind `DEBUG_ENABLED`), which this set
  deliberately does not recognize — a run that leaks only RIDs is a clean `--strict` with an empty
  `diagnostics`. What may enter the set is #722's admission question; #844 answered it for these two
  records only, and no pattern is widened here.

## Surface diff (enumerated, `gda schema` + `--help` vs `b7c485693` in a detached checkout)

`gda schema`: 1,136,730 → 1,146,644 bytes (+9,914, +0.9%). Same 77 commands; `kind`, `argv` and
`constraints` byte-identical for every one. Exactly four kinds of leaf change, and nothing else:

1. `error.$defs.ScriptErrorKind.enum` gains `"shutdown_leak"` — **77 commands** (the shared
   `GdaErrorEnvelope` embeds it), inserted before `incompatible_script`, matching source order.
2. `error.$defs.ScriptError.properties.kind.description` — **77 commands**: the "say the script RAN"
   clause now names `shutdown_leak`.
3. `script run` and `scene preflight` only: the same two changes on their `output` schema (each
   publishes `ScriptError` on its success result).
4. Six single-command descriptions: `script run.input.properties.strict.description`,
   `scene preflight.output.properties.started.description`,
   `scene preflight.output.properties.diagnostics.description`,
   `scene preflight.output.description` (the `ScenePreflightResult` docstring, changed by round 1),
   and the `description` of both commands (their docstrings, which are also the `--help` bodies).
   Re-derived from a real `gda schema --json` diff against `origin/feat/gda-dogfooding-dev`, not
   from the earlier count.

`gda script run --help`: two hunks — the `--strict` option help (both triggers, the process
boundary, the RID exclusion, and the message now naming the leak sentence on a zero status) and one
paragraph of the command body. `gda scene preflight --help`: the recognition sentence plus the
statement that the record does not gate `started`. Everything else byte-identical.

Other surfaces: `src/gda/skill/SKILL.md` (the `script` and `scene` rows), `docs/command-catalog.md`
(the `script run` `--strict` paragraph and the preflight recognition set), ADR-0031 (a dated
`> **Outcome (2026-09-09, #844):**` note appended after the existing outcome notes). **README
unchanged** (and its three translations): its `script run` row states the non-zero-`quit()`
passthrough and that `--strict` inverts it, which stays true; the row never enumerated the strict
rule, and the contract detail belongs in `--help` / SKILL / catalog. Harness bytes untouched (no
version bump).

## Validation

Recorded engine fixture — **no invented lines**. Godot 4.6.3 (`v4.6.3.stable.official.7d41c59c4`,
macOS), a script that parks a loaded `.tres` in a `RefCounted` cycle and `quit(0)`s, run through
`gda script run`:

```
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
   at: cleanup (core/object/object.cpp:2663)
ERROR: 1 resources still in use at exit (run with --verbose for details).
   at: clear (core/io/resource.cpp:810)
```

and the same leak under `godot --headless --verbose` (the engine drops the tail):
`ERROR: 1 resources still in use at exit.` / `   at: clear (core/io/resource.cpp:805)`. Both
spellings are in the parser tests; both `ERR_PRINT`/`WARN_PRINT` sites are unchanged from 4.4-stable
to 4.6.3-stable (checked in the engine source), which is gda's supported range (ADR-0003).

The `printerr` forgeability and the `push_warning` guard were **measured**, not reasoned about — a
probe script on 4.6.3 printing all four lines returns three diagnostics: the forged
`Failed loading resource:` reads as `resource_load_failed` (the pre-existing opening), the two
forged leak sentences read as `shutdown_leak`, and the `push_warning` spelling the leak sentence is
correctly skipped by its frame.

Red then green: the parser tests were written first and failed on the missing `leaked_at_exit`
import and the unrecognized records; the strict e2e/unit tests failed with a passthrough success
before the rule was widened.

| Gate | Result |
| --- | --- |
| `uv run --frozen pytest -m "not e2e" -q` | **2645 passed** (base 2626 collected; +19 unit tests) |
| `uv run --frozen pytest tests/script tests/scene -m e2e -q` (under the machine-wide lock) | **213 passed** in 393 s, 0 skipped (base 210 collected; +3 e2e) |
| `uv run --frozen ruff check .` | clean |
| `uv run --frozen ruff format --check .` | clean (515 files) |
| `uv run --frozen pyright` | 0 errors, 0 warnings |

Mutants (all restored afterwards, `git diff` confirmed back to the intended state):

- **A — drop the two recognizer branches in `_classify`** (nothing is a leak any more) →
  **11 fast-tier tests + 3 e2e tests** fail (4 parser, 4 preflight, 3 `script run` operation, and
  all three real-engine leak tests). Re-measured on this head; round 2's three new guards raise the
  fast-tier half from 9, and the e2e half is unchanged because this round adds no e2e test.
- **B — revert the strict rule to `strict and raw.exit_code != 0`** (the second trigger is gone) →
  exactly **1 fast-tier test + 1 e2e test** fail:
  `tests/script/test_script_run_operation.py::test_strict_fails_a_zero_exit_that_leaked_at_exit` and
  `tests/script/test_e2e_script_run.py::test_script_run_strict_fails_a_leaking_run_that_exited_zero`.
- **C — de-anchor the two leak patterns** (`^…` dropped, `.match` → `.search`, so the wording
  anywhere in a record is a leak) → exactly **1** test fails,
  `tests/script/test_script_error_parser.py::test_the_leak_wording_inside_another_sentence_is_not_a_leak`
  — the guard added in round 2. Before it, this mutant survived the whole fast tier.
- **D — let `SHUTDOWN_LEAK` gate `started` again**, mutated in each derivation separately →
  the ordinary verdict kills
  `tests/scene/test_scene_preflight_commands.py::test_an_exit_time_leak_is_reported_but_does_not_gate_started`,
  the splash-quit route kills
  `tests/scene/test_scene_preflight_commands.py::test_the_splash_quit_route_reads_a_leak_the_same_way`.
- **E — keep the old `--strict` message wording** (`but leaked at exit`) →
  `tests/script/test_script_run_operation.py::test_strict_fails_a_zero_exit_that_leaked_at_exit`
  fails, and the e2e
  `tests/script/test_e2e_script_run.py::test_script_run_strict_fails_a_leaking_run_that_exited_zero`
  asserts the same wording against the real engine.
- **F — widen `--strict` to `raw.exit_code != 0 or bool(diagnostics)`** (any recognized diagnostic
  fails a run) → exactly **1** test fails,
  `tests/script/test_script_run_operation.py::test_strict_leaves_a_surviving_runtime_error_a_success`
  — the guard added in round 2. Before it this mutant passed the whole fast tier **and** the whole
  e2e tier.
- **G — leave the preflight renderer's early return at `if preflight.started:`** (a record that did
  not gate the verdict is invisible in a terminal) → exactly **1** test fails,
  `tests/scene/test_scene_preflight_commands.py::test_the_human_channel_shows_a_record_that_did_not_gate_the_verdict`.
  Before it the whole fast tier stayed green.
- **H — restrict `_RESOURCES_LEAKED` to one digit** (`^\d` instead of `^\d+`) → exactly **1** test
  fails, `tests/script/test_script_error_parser.py::test_a_multi_digit_resource_count_is_recognized`.
  Every captured fixture in the slice says `1`, so this mutant survived the whole fast tier.

PR CI runs **no** Godot e2e (the `godot-e2e` workflow is schedule/dispatch-only), so what CI proves
here is the pure-function half plus the contract guards: the parser tests over the recorded
captures, the `script run` operation tests through the injected launch seam (including the strict
widening, the message branch and the evidence key set), the anchoring guard, the two preflight
derivation tests, the enum-drift guard that forces every new kind to declare whether it can fail an
entry, the error-registry producer-set assertion, and the schema/help snapshot tests — all in the
2641-test fast tier, plus ruff and pyright.

What needs the real engine, and was run locally under the machine-wide lock: that Godot really
prints these two records for a leaking run, that it really exits `0` while doing so, that the
records reach gda's stderr capture after the sentinel on the preflight channel, and that `--strict`
turns that into exit 4. Those are the three `@pytest.mark.e2e` tests; a fixture-only proof would pin
what we imagined the engine prints.

## Review round 1 (handled at `f934f919a`)

Seven findings, all adopted. One was a behavior change, two were falsified or overstated claims,
one was a missing test, three were wording.

1. **`scene preflight.started` must not be gated by an exit-time leak** — fixed. `SHUTDOWN_LEAK` is
   excluded from both derivations through one shared reading, the records stay in `diagnostics`, and
   `started`'s published description says so. This was the PR's only non-opt-in behavior change and
   AC3 never asked for it.
2. **"Godot reports a leak at exit twice" was falsified** — the RID family is a third one. Fixed as
   WORDING: a closed set of two, the RID reports named as deliberately outside it, and the
   `--strict` help plus the SKILL row narrowed to leaked objects or resources. No pattern widened.
3. **The regex anchoring had no test guard** — added; see mutant C.
4. **"nothing keys on project prose" needed the `printerr` caveat** — stated, and measured.
5. **The `--strict` message attributed a process leak to the entry script** — reworded.
6. **Admission clause (2) overstated ownership** — reworded, in the module docstring and in the
   ADR-0031 note.
7. **This body understated its own mutants** — corrected above, and re-measured on this head.

Four surfaces the first round left stale were repaired with them, not the two an earlier version of
this body claimed: the `scene preflight` paragraph of `docs/command-catalog.md` and the `scene` row
of `src/gda/skill/SKILL.md` (both still said `started: false`), the ADR-0031 outcome note (which
carried both the `started: false` claim and the old ownership clause), and the
`ScriptErrorKind.SHUTDOWN_LEAK` enum comment (which still said "the fate of the run's own objects";
that one is deliberately out of the schema, so it moves no description). The ADR change stays inside
the existing dated note — no new section, no rewrite of the point-in-time prose.

## Review round 2 (handled at `bf6d39f36`)

Converged in substance: no round-1 finding reopened. Five findings, all adopted — 0 P1, 2 P2, 3 P3.

1. **`--strict`'s two-trigger boundary had no guard** — added
   (`test_strict_leaves_a_surviving_runtime_error_a_success`). The rule is two triggers, never "any
   recognized diagnostic fails", because a runtime error the script survived is data on a successful
   result (#651). Mutant F above; before the guard the widened condition passed both tiers.
2. **The round-1 fix made the HUMAN channel drop the leak records** — fixed, and this was the only
   behavior change of the round. Measured on the running engine before and after; see the bullet in
   *Public behavior*. Two renderer tests were added: the started-with-records render, and the clean
   run that must stay one short line.
3. **The resources count had no multi-digit guard** — added
   (`test_a_multi_digit_resource_count_is_recognized`). The shipped regex was already correct; only
   the guard was missing. GDA-DF-063 reported 7.
4. **`gda.script_errors`'s two consumer lists were stale** — the module's *Consumers* list and
   `script_error_line`'s "the four places that write one" now name the daemon readiness boundary
   (#848), which PR #940's review record handed to this slice. #848 merges before #844, so the
   counts (four consumers, five writers) are the ones that hold in the merged tree.
5. **This body's arithmetic** — corrected above: six single-command descriptions, not five, and four
   stale surfaces, not two. Both re-derived from a real schema diff and the round-1 commit.

`gda schema --json` at this head is **byte-identical** to `f934f919a`: the renderer and the module
docstrings are not published, so this round moves no leaf at all.

## Review round 3 (independent, base to head `bf6d39f36`) — handled by backtrace at `8acdc7534`

CHANGES REQUIRED. `shutdown_leak`'s meaning was restated in six places and the restatements said "proves the script RAN" — false on the very head where a scriptless scene with a leaking autoload emits it. One meaning, stated at the enum (the owner) and every restatement: the record proves the PROCESS reached shutdown; the precedence exclusion stays right, its REASON changes; `WARNING:` added to the published prefix list; the daemon-consumer sentences (true only on #848's head) moved to PR #940; issue #844's body corrected in place with the dated comment kept; the per-script-shape debt recorded in ADR-0031's note with its trigger.

## Review round 4 (recheck at `8acdc7534`) — handled at `1341853ac`

Three restatements the round-3 sweep missed, because it searched by PHRASE and these used other words: issue #844's criterion (3) ("proves the script RAN"), the module's admission criterion (2) (a "SCRIPT's fate" rule that did not cover the process record it admits), and the entry-failure test comment. This round enumerated every mention of the identifier across `src`, `docs`, `tests`, the issue and this body, read each, and corrected exactly those — plus the same sentence in this body above. Criterion (2) now admits a record about a script's fate OR about the fate of the process the run was, naming the exit-time record as the one such member. Prose only; no wire change.

## Review round 5 (recheck at `1341853ac`) — handled at `3e33f2d48`

One more stale rationale: `test_a_shutdown_leak_never_fails_an_entry_verdict`'s comment ("prints it after the run, so it proves the script ran") — the round-4 enumeration listed the test by identifier but read only the matching line. This round swept by RATIONALE ("proves the script", "script ran", "after the run", "printed the leak" and variants) across `src`, `tests`, `docs` and the README; this was the only remaining old-semantics claim about the leak record. Prose only.

## Open questions

**The ObjectDB record carries no count in 4.4–4.6.3**, but Godot's `master` reformats that sentence
to `%d ObjectDB %s leaked at exit (run with \`--verbose\` for details).`. The pattern anchors on
`^ObjectDB instances leaked at exit`, so it would stop matching on a future engine that ships that
wording. Deliberate (recognition is closed and keyed on what the supported range prints); worth a
note when the version floor moves.

Closes #844


qinhaihong-red:
## Round-1 review handled — all seven findings adopted, fixed at `f934f919a`

No finding was declined. One was a behavior change, one was a falsified claim, one a missing test,
three wording, one the PR body itself. Three stale surfaces the fixes exposed were repaired with
them.

### P2-1 — `scene preflight.started` gated by an exit-time leak — **FIXED** at `f934f919a`

Adopted as stated. `SHUTDOWN_LEAK` is now excluded from BOTH derivations of `started`, through one
shared reading (`_startup_was_clean` in `src/gda/commands/scene.py`) so the ordinary verdict and
`_ended_before_the_verdict`'s splash-quit route cannot answer differently. The records stay in
`diagnostics`, which is all AC3 asks of this command, and `started`'s published description now
carries the clause. The finding's false negative is real and is what the new tests pin: a scene
whose nodes carry no script at all read `started: false` because a project autoload leaked at exit.

Three new tests, all in `tests/scene/test_scene_preflight_commands.py`:
`test_an_exit_time_leak_is_reported_but_does_not_gate_started`,
`test_a_leak_beside_a_startup_error_still_gates_started` (the exclusion is by KIND, not by "some
diagnostic was recognized"), and `test_the_splash_quit_route_reads_a_leak_the_same_way`. The e2e
`test_a_scene_that_leaks_at_exit_is_reported_without_losing_its_verdict` was inverted to assert
`started: true` against the real engine.

Mutated back, one derivation at a time: gating in the ordinary verdict kills the first test, gating
in the splash-quit route kills the third. Neither mutant is silent.

### P2-2 — "Godot reports a leak at exit twice" falsified by the RID family — **ADOPTED AS WORDING**

Confirmed against the engine source at `070dc9897`: `servers/rendering/renderer_canvas_cull.cpp`
prints `WARN_PRINT(vformat("1 RID of type \"%s\" was leaked.", …))` / `"%d RIDs of type \"%s\" were
leaked."`, and `core/templates/rid_owner.h:430` prints
`ERROR: %d RID allocations of type '%s' were leaked at exit.` through `print_error`, NOT behind
`DEBUG_ENABLED`. Neither matches any pattern in the set, so a run that leaks only RIDs is a clean
`--strict` with an empty `diagnostics`.

Remedy applied exactly as specified — **wording only, no pattern widened**. The module docstring now
states the recognized set as a closed set of TWO records about leaked objects and resources, with a
dedicated paragraph naming the RID family as deliberately outside it and pointing at #722's
admission question as the place that decides what may enter. `--strict`'s option help, the
`ScriptRunParams.strict` description, the `run_script` docstring, the SKILL `script` row and
`docs/command-catalog.md` all now say "leaked objects or resources at exit" and name the RID
exclusion.

### P2-3 — the regex anchoring had no test guard — **FIXED**

Added `tests/script/test_script_error_parser.py::test_the_leak_wording_inside_another_sentence_is_not_a_leak`:
an `ERROR:` record quoting `… said 1 resources still in use at exit.` and a `WARNING:` record
quoting `… said ObjectDB instances leaked at exit …`, both from a project frame, must parse to `[]`.

Mutant, measured: drop the `^` from both patterns and switch `.match` to `.search`. Before the test
the mutant survived the entire fast tier. With it, exactly one test fails — this one — and it covers
both patterns (the mutant produces two `shutdown_leak` records, one per sentence).

### P3-1 — the "nothing keys on project prose" claim needed the `printerr` caveat — **ADOPTED, AND MEASURED**

Stated in the module docstring, and not from code reading. Probe on Godot 4.6.3, a script printing
four lines through `printerr` / `push_warning`, run through `gda script run --json`:

| printed | gda read |
| --- | --- |
| `printerr("ERROR: Failed loading resource: res://forged.tres.")` | `resource_load_failed` |
| `printerr("ERROR: 1 resources still in use at exit (…).")` | `shutdown_leak` |
| `printerr("WARNING: ObjectDB instances leaked at exit (…).")` | `shutdown_leak` |
| `push_warning("ObjectDB instances leaked at exit (…).")` | *skipped* (frame guard) |

So the frame guard does what it claims for the builtin paths, the `printerr` opening is pre-existing
(the first row), and what is new is only that forging THIS kind flips `--strict` without having to
look like an entry-load failure. The docstring says exactly that.

### P3-2 — the `--strict` message attributed a PROCESS leak to the entry script — **FIXED**

`script run --strict: <script> exited with status 0 but leaked at exit — …` became
`… exited with status 0, but the engine reported a leak at exit — …`. The script stays named as the
subject of the RUN, never of the leak; `src/gda/errors.py`'s docstring says why. Both the unit test
and the real-engine e2e assert the new wording, so keeping the old one is not silent.

### P3-3 — admission clause (2) overstated ownership — **FIXED**

"the objects and resources of the run the script owned" → "of the PROCESS this run was" in
`src/gda/script_errors.py`. The same clause was stale in the ADR-0031 outcome note and was corrected
there too, inside the existing dated note block — no new section and no rewrite of the
point-in-time prose.

### P3-4 — the PR body understated its own mutants — **FIXED, and re-measured on this head**

The body now lists five mutants with exact counts re-measured on this head, not the three-test claim
it carried. Measured, not copied: dropping both recognizer branches kills **9 fast-tier + 3 e2e**
tests (it was 6 + 3 before round 2 added the preflight tests); reverting the strict rule kills
exactly **1 unit + 1 e2e**, the two the finding named.

### Three stale surfaces repaired with the fixes

- `docs/command-catalog.md`'s **`scene preflight`** paragraph still said a leaking scene reads
  `started: false` — the exact sentence P2-1 removed everywhere else. Corrected, with the reason.
- ADR-0031's outcome note carried both the old `started: false` claim and the P3-3 ownership clause.
  Corrected in place.
- The `ScriptErrorKind.SHUTDOWN_LEAK` enum comment still read "the fate of the run's own objects".
  Corrected (comment only — this one is deliberately kept out of the schema, so it moves no
  description).

`README.md` and its three translations are untouched, as the row contract does not change.

For the record: issue #844's own rationale carries the same ownership clause ("the fate of the run
the script owned") in its *What to build* section. The implementation and ADR now say PROCESS; the
issue text is superseded by this PR rather than edited.

### Gates on `f934f919a`

| Gate | Result |
| --- | --- |
| `pytest -m "not e2e"` | 2641 passed |
| `pytest tests/script tests/scene -m e2e` (machine-wide lock, one process) | 213 passed, 0 skipped |
| `ruff check` / `ruff format --check` | clean / 515 files |
| `pyright` | 0 errors, 0 warnings |
| `gda schema --json` vs the pushed head `eab9a87dd` | 5 leaf **descriptions** changed, 0 keys added or removed |

#844 still mints no builder, error code, evidence field or producer: `src/gda/harness/`,
`docs/adr/0002*`, `docs/adr/0004*`, `src/gda/error_codes.py` and `src/gda/models.py` are byte-identical
to `origin/feat/gda-dogfooding-dev`.

qinhaihong-red:
## Review round 2 — CONVERGED in substance. CHANGES REQUIRED, 0 P1 / 2 P2 / 3 P3 (all adopted)

Independent adversarial review at `f934f919a` against the base `b7c485693`, on a running 4.6.3
engine. **All seven round-1 findings are genuinely fixed and each was confirmed on the running
system** — the P2-1 false negative reproduced end to end (`started: false` at `eab9a87dd`,
`started: true` at `f934f919a`, on a project with a leaking autoload and a `main.tscn` whose nodes
carry no script at all). Nothing from round 1 reopened, no fix caused a regression in the
recognizer or the verdict, and the design stopped moving. Both P2s are round-2 discoveries.

### The convergence rule, applied rather than waived

This wave's rule says a fix that breaks an adjacent case stops the patch cycle and sends the slice
to `/backtrace-review` instead of a third patch. **P2-2 is literally that shape**, so the rule was
applied, not skipped. The question it raises is: after round 1, does the slice still have a clean
responsibility boundary?

It does, and the answer is not a judgement call. `ScenePreflightResult` publishes three facts:
`status` (the engine's readiness), `diagnostics` (everything recognized), and `started` (the
one-boolean gate). Round 1 deliberately severed the 1:1 relation between `started` and
`diagnostics` — a record can now exist without changing the verdict, which is the whole point of
excluding an exit-time record from a question about the boot. The human renderer shortcut on the
OLD invariant (`started ⟹ no diagnostics`) and nobody re-examined it. So this is one consumer
holding a stale invariant, not a boundary that has come apart: fixing the renderer RESTORES the
round-1 design rather than compensating for it. There is no growing mechanism here — no
capability gate, no hidden state, no flag added to defend another flag — which is what the rule
exists to catch. A backtrace would cost more than the one line it would recommend.

**Two REVIEW rounds are the budget and they are spent.** The fix below is the closing fix; I will
verify it directly rather than dispatch a third review.

### Findings and dispositions

**P2-1 — `--strict`'s two-trigger boundary has NO guard. ADOPTED (test).** The issue, the ADR note,
the `--help`, the catalog and the PR body all assert the same boundary: no general "any `ERROR:`
line fails strict" rule, because runtime errors the script survived are data on success (#651).
Nothing pins it. Widening the condition to `raw.exit_code != 0 or bool(diagnostics)` passes the
fast tier (2641) **and** the e2e tier (213) with zero failures — `--strict` could silently start
failing every surviving `push_error` run, inverting ADR-0031's decided default, and no test would
notice. The two tests that look like they cover it are blind: one uses `stderr=""`, the other
asserts `diagnostics == []`. This is round-1's P2-3 one level up. The reviewer wrote the guard,
ran it on head (passes) and under the mutant (reds) — take it as given.

**P2-2 — the round-1 fix makes `scene preflight`'s HUMAN channel drop the leak records entirely.
ADOPTED, and I am choosing the remedy.** `render_scene_preflight` early-returns the one-line form
whenever `started` is true. Now that a leak no longer gates `started`, a recognized `shutdown_leak`
reaches `--json` and reaches **nothing** in a terminal. Measured on one project across three
revisions: base `ready`, `eab9a87dd` `ready with errors` + both records, head `ready` — the records
gone. The PR body's "a leaking scene carries the `shutdown_leak` entries in `diagnostics`" is true
on `--json` and silently false on the human channel, and neither the body nor the round-1 handling
comment mentions that channel, so it was an oversight, not a decision. It is also untested:
narrowing the early return leaves the entire fast tier green, so nothing covers the
started-with-diagnostics render path.

**Remedy: show the records.** Narrow the early return to `started and not diagnostics`, and derive
the headline from `started` so the existing vocabulary is enough — `ready` when the boot was clean,
the records printed beneath it; `ready with errors` stays for `started: false`. **No new headline
word.** The alternative — deciding that a `started: true` verdict never shows diagnostics in a
terminal — is rejected: the two renderings are of ONE outcome, and a leak sitting invisible in a
terminal is the exact complaint GDA-DF-063 was filed for. Add the renderer test the gap exposed.

**P3-1 — the resources count has no multi-digit guard. ADOPTED (test).** Every fixture uses
`1 resources still in use at exit`, so restricting the pattern to a single digit survives the whole
fast tier. GDA-DF-063 — the dogfooding report this issue comes from — reported **7**, and a real
leak routinely exceeds nine, so a 10-resource leak would silently stop being recognized and stop
failing `--strict`. The shipped regex is correct; only the guard is missing.

**P3-2 — `script_errors.py`'s two stale consumer lists, explicitly handed to #844, were not picked
up. ADOPTED (wording).** PR #940 (#848) added a third consumer and its review record says verbatim
that the file "belongs to #844 this wave and stays untouched". Still stale: the module docstring's
Consumers list names three and misses the daemon readiness boundary, and `script_error_line`'s
docstring says "the four places that write one" where the daemon's `wait-ready` / `status`
renderers make five. #848 merges BEFORE #844, so this is the last cheap chance, and the only other
record of it is a PR-body row that a `BLANK` squash message discards.

**P3-3 — PR body arithmetic, again. ADOPTED.** The surface-diff item says five single-command
descriptions and lists five; the measured base→head set is **six** — `scene preflight`'s
`output.description`, changed by this very round, is missing. And "two surfaces the first round
left stale were repaired" is really four. Round 1's P3-4 was exactly "this body understates its own
numbers".

- **All seven round-1 fixes on the running system.** The exclusion is by KIND, not by "some
  diagnostic exists": the same leaking project with a `push_error` in `_ready` reads
  `started: false` with `[push_error, shutdown_leak, shutdown_leak]`. The splash-quit route answers
  identically, and it really takes that route — mutant D2 flips only `splash.tscn`, D1 flips only
  `main.tscn`.
- **The RID claim, at two engine tags.** `object.cpp` and `resource.cpp` at **4.4-stable and
  4.6.3-stable** print exactly the two recognized sentences, and the fixtures' line numbers are the
  real 4.6.3 ones. RID family confirmed at `renderer_canvas_cull.cpp:2690/2692` (`WARN_PRINT`) and
  **`rid_owner.h:440` `print_error` with no `DEBUG_ENABLED` guard** (every `#if` in the file
  checked). Neither RID sentence matches any pattern in the set. The recorded caveat is real:
  master rewrote the line to a counted form the `^ObjectDB instances` anchor will miss.
- **`printerr` forging measured**: the forged `Failed loading resource:` reads as
  `resource_load_failed` (pre-existing), both forged leak sentences read as `shutdown_leak`, and
  `push_warning` with the same words is skipped by the frame guard. Under `--strict` the run is
  exit 4 with the corrected, de-attributed message.
- **All six claimed mutants re-run**, each killed by exactly the named test and only it — including
  B, C, D1, D2 and E at exactly one test each, as claimed. Five further mutants written by the
  reviewer are killed cleanly (`_startup_was_clean` keyed on `path`, `leaked_at_exit` returning the
  last record, the message branch not restricted to a zero status, dropping the `push_warning`
  frame guard, dropping the READY half).
- **Gates and surfaces:** fast 2641 / 687; `tests/script tests/scene -m e2e` 213 passed, 0 skipped,
  under the lock; ruff clean, 515 formatted, pyright 0/0. Base collects 2626 / 210, so the +15 / +3
  claims are exact. Schema: 77 commands, same names and order, **`kind` / `argv` / `constraints`
  byte-identical for every one**; `shutdown_leak` inserted at index 8 matching source order; r1→head
  exactly 5 leaf descriptions changed, 0 keys added or removed.
- **F3/F4 hold:** harness, ADR-0002, ADR-0004, `error_codes.py`, `models.py`, README and the three
  translations are absent from the changed-file list; `script_exit_status_failure` gains a docstring
  paragraph and four body lines and restructures nothing.
- **A full stale-copy sweep found none remaining** other than P3-2's consumer lists, across `src/`,
  `docs/`, `tests/`, `README*` and the live PR body.
- **Shared-file hotspots for the serial merge:** `errors.py` — #966's insert and #964's region are
  disjoint, four clean lines apart in the import block; `commands/script.py` — six lines apart, no
  restructure on either side; `command-catalog.md` and `SKILL.md` — three PRs edit different
  sections, so watch for keep-both appends at merge time. #964 merges before #845, so #966 rebases,
  which is the expected direction.
- **Proportionality:** the `printerr` forgery is disclosed and measured with no defence built —
  correct under "no product promise for extremely small technical situations" — and there is no
  compatibility machinery anywhere in the slice.

### Not verified, and why

A RID-only leak producing an empty `diagnostics` on a real run: the canvas / rendering-device sites
need a live rendering server, so it could not be synthesized headless. Verified instead by engine
source at 4.6.3-stable and master — sound, but source reading rather than a measured run, and the
docstring states it as a fact about behaviour. Also the PR body's "268 s" e2e timing (measured 245 s
at load 3.4; immaterial).

qinhaihong-red:
## Round 2 handled at `bf6d39f36` — verified directly by the orchestrator. **READY.**

The implementer's session ended after the push and the body update, before its reply comment, so
this comment is both the finding → resolution map and the closing verification. Two review rounds
were the budget and both are spent; no third review was dispatched.

### Finding → resolution

| # | Finding | Resolution at `bf6d39f36` |
|---|---|---|
| P2-1 | `--strict`'s two-trigger boundary had no guard — widening it to "any diagnostic" survived both tiers | `test_strict_leaves_a_surviving_runtime_error_a_success` added; under the mutant it is the ONLY test that reds (1 failed / 130 passed) |
| P2-2 | the round-1 fix made the human channel drop the leak records | `render_scene_preflight` short-circuits only on `started and not diagnostics`; the headline is derived from `started`, so `ready` keeps its word with the records beneath it and `ready with errors` still means `started: false`. **No new headline word.** `test_the_human_channel_shows_a_record_that_did_not_gate_the_verdict` added; under the mutant it is the only test that reds (1 / 35) |
| P3-1 | no multi-digit guard on the resources count | `test_a_multi_digit_resource_count_is_recognized` added; under the single-digit mutant it is the only test that reds (1 / 53) |
| P3-2 | `script_errors.py`'s consumer lists stale since #848 | module docstring names the daemon readiness boundary as the fourth consumer, and `script_error_line` names the daemon renderers as the fifth writer |
| P3-3 | PR-body arithmetic | six single-command descriptions, four repaired stale surfaces — corrected in the body's `## Review round 2` section |

### Verified by the orchestrator

- **The human-channel defect, reproduced and gone, on a real engine (4.6.3) under the machine-wide
  lock.** A project with a leaking autoload (an orphan `Node` holding a loaded `.tres`) and a
  `main.tscn` whose nodes carry no script:
  - at `f934f919a`: `ready res://main.tscn` and nothing else — the two records the JSON carries are
    dropped;
  - at `bf6d39f36`: `ready res://main.tscn`, then `project:` and the two `shutdown_leak:` lines.
  - The startup-error scene (`push_error` in `_ready`) reads `ready with errors` with its
    `push_error:` line and the two leak lines, identically on both revisions.
  - `--json` on both revisions: `main` → `started: true, status: ready, [shutdown_leak,
    shutdown_leak]`; `boom` → `started: false, [push_error, shutdown_leak, shutdown_leak]`. The
    verdict did not move; only the rendering did.
- **The closing commit changes no published leaf.** `gda schema --json` at `f934f919a` vs
  `bf6d39f36`: 0 added / 0 removed / 0 changed. Against dev: 79 added / 0 removed / 164 changed,
  matching the round-2 review's count exactly; 77 commands, `kind` / `argv` / `constraints`
  identical on every one.
- **Untouched, byte-identical to dev:** `src/gda/harness/`, ADR-0002, ADR-0004,
  `src/gda/error_codes.py`, `src/gda/models.py`, `README.md` and the three translations (F3/F4).
- **Gates:** fast tier **2645 passed** (+4 over `f934f919a`'s 2641 — the four new guards);
  `tests/script tests/scene -m e2e` **213 passed, 0 skipped** in 234 s; ruff clean, 515 files
  formatted, pyright 0 / 0 / 0. PR CI green, `mergeStateStatus: CLEAN`.
- **Commits:** three on top of dev, each authored and committed by
  `haihong.qin <haihongqin@gmail.com>`, no attribution of any kind; the closing commit was a
  fast-forward push, no force.
- **The source diff is exactly the adjudication:** read line by line — the early return, the
  headline derivation, the two consumer-list sentences, and nothing else in `src/`.

W3 merge order for this slice is unchanged: `#939 → #937 → #938 → #940 → #964 → #966 → #967 → #973`.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `bf6d39f36e1b32388c89affa6fbea511f671e4e0`. Existing reviews and replies were not used as evidence.

1. Issue #844's primary body still says that the leak describes “the fate of the run the script owned.” Its later correction says that the record is process-wide and an autoload can produce it. Move that correction into the body and state the non-gating `scene preflight.started` consequence there; a trailing supersession note leaves the implementation authority internally contradictory.
2. The public `ScriptError.message` description at `src/gda/script_errors.py:450-453` says only `ERROR:` / `SCRIPT ERROR:` prefixes are stripped, but this PR admits and strips an ObjectDB `WARNING:` record. Add the third prefix.
3. `src/gda/script_errors.py:25` and `:479` claim that `daemon wait-ready` / `status` consume this parser/renderer, but this exact head has no daemon import or consumer. Remove the forward claim from this independent PR, or land it only with the implementing dependency.

**Terminology and prose assessment:** I inspected the complete issue and correction, PR body, ADR-0031 amendment, command catalog, skill, help/schema prose, and parser docstrings against `CONTEXT.md`, ADR-0002/0004/0023/0031, and ASD-STE100. `shutdown_leak` is stable, but `script ran` and the prefix/consumer claims above are factually inaccurate. I did not validate leak strings outside Godot 4.4-4.6.3.

`gda.script_errors` is the correct shared owner for the admitted engine record, and strict policy remains in the script command while scene preflight owns its non-gating startup verdict. However, a process-lifecycle record now travels in `ScriptError` / `evidence.script_errors`. The current issue explicitly avoids a public shape migration, so a broader diagnostic type is non-blocking design debt; within this slice, every contract must say only that the **process reached shutdown**, not that the entry script ran. The claims at `src/gda/script_errors.py:354-358,390-395,439-447` and ADR-0031:625-626 violate that boundary. Claim-evidence state for a future type rename: `open`; no rename is required in #844.

## Standards

- Hard: the public contract says `shutdown_leak` proves the script ran, although the same head proves an autoload can emit it for a scriptless scene and tests combine a missing entry with a shutdown leak. Replace that claim with process-shutdown semantics.
- Judgment — Mysterious Name / Primitive Obsession: `ScriptError` now contains a process-wide lifecycle record. Keep this as explicit design debt because the issue forbids a new wire shape.

Standards summary: 2 findings; worst is a hard false semantic claim.

## Spec

- Required: add `WARNING:` to the public message-prefix description (`src/gda/script_errors.py:450-453`).
- Required: remove the nonexistent daemon consumers claimed at `src/gda/script_errors.py:25,479` from this head.

The requested parser admission, strict precedence, non-strict data behavior, and non-gating preflight behavior otherwise match the corrected issue. Independent validation passed 221 focused unit/CLI tests and 3 real-Godot leak E2E tests.

Spec summary: 2 findings; worst is an incomplete public schema contract.

qinhaihong-red:
## Third review handled at `8acdc7534`

The Standards "false semantic claim" and the Artifact/Spec findings share one root, and it is the
third pass over the same wording family, so it was traced before it was patched. History: round 1
corrected "the run the script owned" to "the process" in the strict message and the admission
clause; round 2 found four more stale copies; the third review finds the claim one level up — that
the record "proves the script RAN" — still standing in the enum, the module's criterion (3), the
precedence comment, the published `kind` description and the ADR note. Root cause: the kind's
meaning was RESTATED in six places instead of stated once at its owner, and every restatement
inherited the original mistake. The record proves that the PROCESS reached shutdown still holding
objects or resources; it says nothing about the entry script — a scene with no script emits it
under a leaking autoload, and the tests combine it with a missing entry. The precedence exclusion
stays, for the reason that survives: it names no resource.

| The contract says `shutdown_leak` proves the script ran | Fully Adopt | this head's own e2e: a scriptless scene under a leaking autoload; a missing entry beside a leak | process-shutdown semantics at the enum (the owner) and at every restatement — criterion (3), the precedence comment, the published `kind` description, ADR-0031's note | script/scene/cli fast tier 797 passed; ruff / format / pyright clean; CI green; `grep` for the old claims empty |
| `ScriptError.message` says only `ERROR:` / `SCRIPT ERROR:` are stripped, but the set now strips a `WARNING:` record | Fully Adopt | the ObjectDB leak is a `WARN_PRINT` | the published description names the third prefix | schema: description-only change |
| `script_errors.py` claims daemon consumers that do not exist on this head | Fully Adopt | added in round 2 as "#844's last cheap chance" because #848 merges first — but false at THIS head as an independent PR | the two sentences are removed here and land in PR #940, the PR that adds the consumer | — |
| Issue #844's body still says "the fate of the run the script owned" while a trailing note contradicts it | Fully Adopt | the body is the implementation authority and was internally contradictory | the body's criterion (2) is corrected in place — process semantics and the non-gating `scene preflight.started` consequence — and the dated comment stays as the record of the change | — |
| A process-lifecycle record now travels in the per-script `ScriptError` shape (Architecture: non-blocking debt) | Acknowledged, not changed | the issue forbids a wire-shape migration | recorded in ADR-0031's note as debt, with its trigger: a SECOND process-lifecycle record | — |

Head `8acdc7534`, one commit on `bf6d39f36`; `src/gda/harness/`, ADR-0002/0004, `error_codes.py`, `models.py`, README and translations still byte-identical to dev.

qinhaihong-red:
## Follow-up recheck — runtime fixed, authority still inconsistent at exact head `8acdc7534`

The enum, published field descriptions, ADR, `WARNING:` message contract, and consumer list now carry the intended process-wide behavior. The claim that every restatement was corrected and that a grep for the old semantics is empty is not borne out by the exact head:

1. Issue #844's current **What to build** bullet first says the record “says nothing about whether the entry script ran,” then its criterion (3) still says “the kind states in the enum that it proves the script RAN.” This is an internal contradiction in the primary issue body.
2. `src/gda/script_errors.py:40-42` still defines the general admission rule as a record saying something specific about a **SCRIPT's fate**, while the admitted `shutdown_leak` paragraph now correctly says it describes the process and says nothing about the entry script. The rule either needs to include process/run-lifecycle facts or identify this as an explicit exception.
3. `tests/script/test_script_error_parser.py:169-175` still says `runtime_error`, `push_error`, and `shutdown_leak` all prove the script ran, including the old “printed the leak at exit, after the run” rationale.

These are the same semantic authority surface as the original finding, not a new runtime redesign request. The minimum closure is prose-only: correct issue criterion (3), reconcile the module's admission rule with its process-wide exception, and update the stale test comment. No wire-shape migration is required.

Independent validation: 77 focused parser/script tests pass and current CI is green with no pending jobs. Those results confirm the implementation remains healthy, but they do not test these contradictory comments and issue text.

qinhaihong-red:
## Fourth review handled at exact head `1341853ac`

The third round's sweep found restatements by PHRASE ("proves the script RAN"); the three you name use other words, which is why "a grep for the old semantics is empty" was true and useless. This round enumerated every mention of the identifier (`shutdown_leak` / `SHUTDOWN_LEAK` / "shutdown leak") across `src`, `docs`, `tests` and the issue — 60 sites — and read each; the three below were the only ones still carrying the old meaning. Prose only, as you scoped it; no wire change.

| Issue #844's What-to-build criterion (3) still says the kind "proves the script RAN", contradicting (2) two sentences earlier | **Fully Adopt** | read on the live issue | criterion (3) now says the kind proves nothing about whether the entry script ran — a record about the process reaching shutdown that names no resource — with a dated correction in place | re-read after the edit |
| `script_errors.py` admission criterion (2) defines admissible records as saying something about a SCRIPT's fate, while the admitted leak record describes the process | **Fully Adopt** | the rule did not cover the one record it admits | criterion (2) now admits a record about a script's fate OR about the fate of the process the run was, naming the exit-time record as the one such member — the rule includes it rather than an unstated exception | `tests/script` + `tests/scene` + `tests/meta` 623 passed |
| `test_every_never_ran_kind_is_an_entry_failure_candidate`'s comment still says `shutdown_leak` proves the script ran ("printed the leak at exit, after the run") | **Fully Adopt** | read on the head | the comment separates the two kinds that prove the script ran from the leak record, which says nothing about the entry script and names no resource | same run |

CI is green on `1341853ac`.

qinhaihong-red:
## Follow-up recheck — one stale semantic claim remains at exact head `1341853ac`

The live issue, the module admission rule, the enum/public descriptions, and the comment in `test_every_never_ran_kind_is_an_entry_failure_candidate` now carry the corrected contract: `shutdown_leak` proves that the **process** reached shutdown and says nothing about whether the entry script ran.

However, a second test comment was missed. `tests/script/test_script_error_parser.py:846-850`, in `test_a_shutdown_leak_never_fails_an_entry_verdict`, still says:

```text
the engine prints it after the run, so it proves
the script ran
```

That is the same old semantic claim and directly contradicts both the current enum contract and this head's author reply. The assertion itself is correct (`entry_load_failure(...) is None`); only its rationale is stale.

Minimum closure is prose-only: state here that the record names no resource and says nothing about the entry script, so it cannot establish an entry-load verdict. No runtime or wire-shape change is required.

Independent validation on this head: **54 parser tests passed** and current required CI jobs are green; those checks do not detect contradictory comments.

qinhaihong-red:
## Fifth review handled at exact head `3e33f2d48`

| `test_a_shutdown_leak_never_fails_an_entry_verdict`'s comment still says the record "proves the script ran" | **Fully Adopt** | read on `1341853ac`; the round-4 enumeration listed that test's `def` line by identifier but read only the matching line, not the comment under it | the rationale now says the record is printed while the PROCESS exits, about everything the process held, names no resource and says nothing about whether the entry script ran, so it can never establish an entry-load verdict | script tier 311 passed |

This time the sweep was by RATIONALE, not by identifier or phrase: every line in `src`, `tests`, `docs` and the README matching "proves the script", "script ran", "after the run", "printed the leak" and their variants was read. The only remaining old-semantics claim about the leak record was this one; every other hit is about `runtime_error`, `push_error` or `script_failed`, where "the script ran" is the correct claim.

CI is green on `3e33f2d48`.

PR #966, opened 2026-09-09
fix(gda): a case-mismatched res:// path is a typed refusal (#845)

### LINKED ISSUE #845: script: a case-mismatched script path is a typed refusal, not valid:true

qinhaihong-red:
## Summary

`gda script validate res://Content/combat_session.gd` returned `valid: true` on macOS
while the tracked file is `res://content/…`. Godot opened it (APFS is case-insensitive by
default) and warned that the same path will not open on a case-sensitive platform, but
that warning is a `WARN_PRINT` and `gda.script_errors` skips every warning by contract, so
a portability gate accepted a path that fails on Linux and on a case-sensitive export host
(dogfooding GDA-DF-062).

gda now decides the spelling itself, at ADR-0006's path authority
(`gda.project.case_mismatch`) and through the ONE containment gate
(`gda.errors.containment_refusal`), so all three commands the gate protects report the new
typed `path_case_mismatch`.

Closes #845

## Public behavior

- **New `Gda error code` `path_case_mismatch`** — category `operation`, exit `4`
  (`EXIT_OPERATION`, the same bucket as `path_not_found`), source `classifier`, because
  ADR-0006 keeps the path authority CLI-side exactly as it does for
  `target_outside_project`. Registered in `error_codes.py` with its ADR-0002 row; the
  description pin holds.
- **Where it fires** — `script validate` (per batch entry), `script run` (its entry
  script), `resource import` (per asset). That is the gate's call set and the scope the
  issue's 2026-09-05 note decided; `scene get` and the rest of the surface still hand the
  `res://` string to the engine and are unchanged.
- **How it is decided** — the requested path is anchored against the project the way the
  engine anchors it, then walked component by component from the project root, comparing
  each component with the directory entry's ACTUAL spelling. It LISTS the directory rather
  than asking whether a path exists, so the verdict is identical on both kinds of
  filesystem: a case-variant entry is the mismatch, and a component no entry matches under
  any case yields NO refusal — that stays the operation's own `path_not_found`.
- **What the refusal says** — the message names both `res://` addresses and the one to
  re-issue with; the same two ride typed as `evidence.requested_path` /
  `evidence.stored_path`. **No `hint`** (issue finding F2): that key is contractually one
  corrected invocation from the curated near-miss table (ADR-0004's #670 note), and a
  computed correction is not a near miss — the same line #840 drew for its two directories.
- **Ordering** — ownership, then containment, then the spelling. The spelling is last
  because it is the least specific of the three (asking how the project spells an entry
  only means something once the target is established as this project's to serve) and
  because it is the only half that touches the filesystem.
- **One code moved, and only where the target really exists** (round-1 disclosure). A
  mis-cased NON-`.gd` target that the project DOES store under another case now reports
  `path_case_mismatch` where `invalid_path` came back before, because the gate runs BEFORE
  the operation. Measured at this head: `gda script validate res://Content` and
  `res://Content/note.txt` (stored `res://content/note.txt`) both answer the new code.
  Both codes are refusals at exit `4`, so no contract breaks, but an agent branching on
  `invalid_path` sees a different one. The ordering is NOT changed for it: the gate is the
  earlier and more specific diagnosis, and it names the spelling to re-issue with where
  `invalid_path` only says the extension is wrong. Where a component names nothing — a
  mis-cased directory with an absent leaf, `res://Content/absent.txt` — the gate still says
  nothing and `invalid_path` comes back unchanged, measured too.
- **Bounds, each deliberate** — the project ROOT's own spelling is not compared (a
  case-differing `--project` stays the separate gap ADR-0006's amendment records); a target
  that is not LEXICALLY under the root is not compared (the symlinked-in file
  `path_outside_project` admits through its resolved reading); `user://` / `uid://` are
  left alone; and a projectless call is not compared, because there is no `res://`
  namespace to name a stored spelling in.

## Surface diff (enumerated)

Captured in a separate detached checkout at the base (`git worktree add --detach
/private/tmp/845-base b7c485693`), running `uv run --frozen gda schema` and the three gated
commands' `--help` there and on this branch, then diffing. The worktree is removed.

`gda schema` — **157 leaf differences, all four of them expected, nothing else**:

| Count | Difference |
| --- | --- |
| 77 × 2 | `error.$defs.FailureEvidence.properties.requested_path` and `.stored_path` added — the one shared envelope schema, repeated once per command in the aggregate manifest (ADR-0004's measured multiplication). |
| 3 | `commands[*].description` — the three help sentences below. |

Size: 1,136,730 → 1,170,595 bytes (+33,865, +3.0%). No other key changed, added, or
removed; no command was added or removed (`_RECIPE_OPERATIONS` untouched — no new command,
no recipe).

The new error CODE does not appear in `gda schema` and is not expected to: the code
registry is published through ADR-0002's table and `gda.error_codes`, not through the
per-command contract.

`--help` — exactly one added sentence per command, nothing else:

- `script validate`: "A path whose CASE does not match the stored file refuses the batch
  with 'path_case_mismatch' naming the stored res:// spelling, because such a path opens on
  a case-insensitive filesystem and fails on a case-sensitive one."
- `script run`: the same rule, in that command's own ``…`` markup.
- `resource import`: the same rule, phrased for an asset.

Not touched: `README.md` and its three translations. No README row's one-line contract
changes — the rows say what each command does, and the README points at ADR-0002's table
for the code registry rather than listing codes. So no `scripts/update_readme_i18n.py` run
and no translation edits. Harness bytes untouched (`HARNESS_VERSION` unchanged).

Other artifacts changed in this PR because the two evidence fields force them: ADR-0004's
`#687` producer paragraph gains a dated addendum (a ninth producer; never a rewrite),
`_EVIDENCE_PRODUCERS` in `tests/cli/test_error_registry.py`, the exhaustive evidence
renderer in `render.py` (two branches), and `_EVIDENCE_SAMPLES` in
`tests/cli/test_human_failure_output.py`. Three further published-shape pins needed the
two field names as well: `tests/cli/test_models.py`, `tests/meta/test_schema_command.py`,
and the gate's complete-call-set assertion in
`tests/project/test_res_containment_consistency.py` (which is exactly the guard that makes
a third builder inside the gate a change someone has to look at). SKILL and
`docs/command-catalog.md` gained the rule.

## Validation

Run in this worktree at `4f5f58242`, host macOS 15.6 / APFS (case-insensitive), Godot 4.6.3.

| Gate | Command | Result |
| --- | --- | --- |
| Fast tier | `uv run --frozen pytest -m "not e2e" -q -p no:cacheprovider` | 2643 passed, 1 skipped, 685 deselected in 53 s |
| Lint | `uv run --frozen ruff check .` | All checks passed |
| Format | `uv run --frozen ruff format --check .` | 516 files already formatted |
| Types | `uv run --frozen pyright` | 0 errors, 0 warnings, 0 informations |
| Targeted e2e | `uv run --frozen pytest tests/script tests/resource tests/project -m e2e -q` under the machine-wide lock, `GDA_GODOT` set | 278 passed, 673 deselected in 456 s |

**Which filesystem branch ran on this host.** The new module probes the behaviour with a
real temp file rather than reading `sys.platform`. On this host the probe reports
case-INSENSITIVE, so
`test_a_case_insensitive_host_would_have_opened_the_mis_cased_path` ran (it asserts its own
premise — the mis-cased path really does `exists()` — and then the refusal) and
`test_a_case_sensitive_host_reports_the_mismatch_rather_than_a_missing_file` is the one
skip in the fast-tier count. On Linux CI the two swap. Every other test in the module is
filesystem-independent, which is the point of the fix: the authority reads directory
entries, so it reaches the same verdict on both.

**Red then green.** The module was written first and failed on the missing
`gda.project.case_mismatch` import; after the authority, the registry row, the two evidence
fields, the builder, the gate branch and the two render branches it passed. Three published
pins then reddened as intended and were updated with the two names
(`test_every_evidence_field_is_optional_in_the_published_schema`,
`test_the_published_error_schema_declares_the_optional_evidence_key`,
`test_the_gate_is_where_both_refusals_are_built`).

**Mutants.**

- Drop the per-component caseless comparison in `gda.project._stored_entry` (keep only the
  exact match) → 9 of 16 tests in `tests/project/test_path_case_mismatch.py` fail, including
  all three gated commands; `script validate` goes back to reporting `valid: true` for the
  mis-cased path, which is GDA-DF-062 reproduced.
- Drop the `stored_path` branch in `render._evidence_lines` →
  `test_every_evidence_field_alone_reaches_the_human_block[stored_path]` fails.

Both mutations were reverted and the gates re-run.

## Review round 1 — APPROVE, 0 P1 / 0 P2 / 5 P3 (+1), re-run at `4a7978096`

All six adopted. Three surviving mutants became three guards; four stale comments and one
"third half" were reworded, with the orphaned line in `containment_refusal`'s docstring
reflowed; two findings are disclosures, above. No executable code changed, so the e2e tier
was not re-run — its inputs are byte-identical.

| Gate | Result |
| --- | --- |
| Fast tier | 2644 passed, 2 skipped, 685 deselected in 71 s — was 2643 / 1: one new test, and one new skip because this host's filesystem is case-insensitive |
| Lint / Format / Types | All checks passed / 516 files already formatted / 0 errors, 0 warnings, 0 informations |
| `tests/project/test_path_case_mismatch.py` on a mounted case-sensitive APFS volume (`hdiutil create -fs "Case-sensitive APFS"`, `--basetemp` on the volume) | 16 passed, 1 skipped — the skip is the case-INSENSITIVE premise test, mirroring the host run |
| `gda schema --json` vs `4f5f58242` | byte-identical, 1,170,595 bytes |
| `--help` of the three gated commands at `COLUMNS=100` vs `4f5f58242` | byte-identical: 6,495 / 16,498 / 6,790 bytes |

**Each new test kills its mutant, and nothing else does.**

| Mutant | Killed by |
| --- | --- |
| `case_mismatch`: `entry = part` instead of `return None` when a component names no entry | `test_a_name_no_entry_matches_is_left_to_the_operation` — 1 failed / 2643 passed; the mutant answers a correction naming `res://content/absent.gd` |
| `_stored_entry`: drop the `if name in entries: return name` shortcut | `test_a_case_sensitive_host_answers_with_the_spelling_the_caller_named`, on the case-sensitive volume — 1 failed / 15 passed / 1 skipped, refusing `res://Content/combat_session.gd` although that file exists exactly. The mutant SURVIVES the whole fast tier on this case-insensitive host (2644 passed), which is why the volume is part of the evidence rather than a skipped test |
| `containment_violation`: ask `case_mismatch` before `path_outside_project` | `test_containment_wins_when_the_spelling_half_fires_too` — 1 failed / 2643 passed, `assert 'path_case_mismatch' == 'target_outside_project'` |

Every mutant was reverted and the working tree checked back to the intended diff before the
commit.

PR CI runs **no** Godot e2e (schedule/dispatch-only, this repo's standing arrangement), so
what it proves for this change is:

- the whole fast tier, which is where every guard in this PR except one lives: the
  authority's own decisions, the three gated commands driven through the real CLI, the
  argv ↔ `--params-json` parity (ADR-0015), the `scene get` non-refusal, the published
  `FailureEvidence` shape, the registry ↔ ADR-0002 description pin, the evidence-producer
  set, and the exhaustive human renderer;
- `ruff check` / `ruff format --check` / `pyright`.

CI runs on **Linux**, whose filesystem is case-sensitive, so the branch this macOS host
skips is the branch CI runs — and vice versa. Between the two hosts both premises are
asserted at least once per change.

What needs the real engine, and therefore the e2e tier (run locally here, green above):
`test_script_validate_case_mismatch_is_refused_and_the_stored_spelling_is_clean`. It is the
control this fix rests on — the ACCEPTED spelling compiles with no diagnostic and no "Case
mismatch" line on either stream — beside the refusal of the mis-cased spelling of the same
file, out of process through the shipped CLI.

## Review round 3 (independent, base to head `4f5f58242`) — handled at `87e052950`

CHANGES REQUIRED. `str.casefold` is FULL folding (`straße` == `strasse`, two different files) → simple folding via `str.lower`; an ambiguous folded match (two entries) was enumeration-order dependent → no correction; the "listing only on the not-found path" promise was unimplementable on a case-insensitive host → reconciled on #845 with a dated note and a cost-bound regression; the skill's OS parentheticals qualified as common defaults. All adopted; proven on a mounted case-sensitive APFS volume.

## Review round 4 (recheck at `87e052950`) — handled at `4f90d4515`

One wording claim: "simple folding, one character to one" overstates `str.lower` (`"İ".lower()` is two code points). Fully adopted as a precise description — equality under `str.lower()` and nothing narrower, expansions included — in the docstring, the test comment and the issue note (dated correction); `test_the_relation_is_equality_under_lower_expansions_included` pins the `İ` case. Defining or rejecting multi-code-point expansions was not adopted: no product promise covers it, and the two rules already in place (ambiguity is no correction; the result is re-issued, never acted on) bound the residual.

## Open questions

- **Widening beyond the gate is not done here, by decision.** `scene get`, `node add
  --instance` and the rest still pass their `res://` target to the engine unchecked, so the
  same mis-spelling is a warning there and nothing more. Routing every `res://` target
  through the authority is an ADR-0006 design change with its own surface; the issue's
  reopen trigger is a second command group dogfooded into the same trap.
- **Case is checked; Unicode NORMALIZATION is not** (round-1 disclosure). An NFD spelling
  of a name stored in NFC has exactly the property this refusal exists for — it opens on
  one filesystem and fails on another — and gda passes it through. Measured at this head:
  `gda script validate res://content/café.gd` spelled NFD (`e` + U+0301) returns
  `valid: true` against a file stored NFC (U+00E9); the gate says nothing, because the two
  spellings are not a CASE difference. It is deliberately not widened here. The published
  description claims a case check rather than general path portability, so no promise is
  broken today, and a normalization check is a different and larger mechanism than this
  fix's scope. **Limit of the evidence: only the macOS half was measured.** Both an
  ordinary APFS volume and a mounted case-SENSITIVE APFS volume are still
  normalization-INSENSITIVE, so the NFD spelling opens on both of them here; the half where
  it FAILS needs a byte-comparing filesystem (Linux ext4), which was not run. Recorded as a
  dated note on issue #845.
- **Caseless matching is equality under `str.lower()`** — never `str.casefold`, whose FULL
  folding made `straße` and `strasse` equal and "corrected" one to the other (round 3).
  `str.lower` is one character to one except where Unicode expands it (`İ`, U+0130 → `i` +
  U+0307); an expanded spelling folds equal to the stored one like any other case
  difference (round 4). A name that folds equal to MORE than one entry is no correction.
  A host's own table can differ in those corners, and the result is a spelling to
  re-issue with, never an action taken on the caller's behalf, so the failure mode is a
  wrong sentence and not a wrong write.
- **Cost**: one `os.listdir` per component of the target, on the accepted path as well as
  the refused one, bounded by the project root. It is the first filesystem read in this
  module's otherwise lexical decision set, and it is last in the ordering so a refused call
  never pays it.


qinhaihong-red:
## Review round 1 — APPROVE, 0 P1 / 0 P2 / 5 P3 (all adopted)

Independent adversarial review at `4f5f58242`, against the base `b7c485693`, with a real engine
(Godot 4.6.3) and **a real case-sensitive APFS volume** mounted for the second filesystem branch.
The behaviour is correct and every published claim that could be checked held. The five P3s are
three surviving mutants, one batch of stale prose in files this PR edits, and one adjacent gap
that needs a disclosure rather than code.

### Findings and dispositions

**P3-1 — ADOPTED (test).** `case_mismatch` returns `None` the moment a component names no entry,
deliberately ("guessing a spelling would replace a true `path_not_found` with a false
correction"). No test covers the case where that matters: a **mis-cased directory plus an absent
leaf**. Mutant `if entry is None: entry = part` survives the whole fast tier (2643 passed), and it
is not equivalent — `res://Content/combat/absent.gd` turns from `path_not_found` into
`path_case_mismatch` naming a "stored" spelling for a file that exists under neither. Add
`assert case_mismatch("res://Content/absent.gd", project) is None`.

**P3-2 — ADOPTED (test).** `_stored_entry` documents that the exact spelling is preferred over a
caseless match, so a case-sensitive filesystem holding both `Content` and `content` answers with
the one the caller named. Nothing pins it. Deleting the `if name in entries: return name`
shortcut lets `listdir` order decide and **falsely refuses a path that exists exactly** — measured
on a mounted case-sensitive APFS volume holding both directories. The mutant survives here and
would survive Linux CI, the host where the condition is reachable. Add the pin in the module's
case-sensitive branch, and verify it on a mounted case-sensitive volume rather than leaving it
skipped.

**P3-3 — ADOPTED (test).** The third arm's precedence is argued at length in
`containment_violation` ("the spelling comes LAST") but not pinned, while the previous pair IS
pinned by `test_ownership_wins_when_both_halves_of_the_gate_fire` (the #807 ordering pin).
Asking `case_mismatch` before `path_outside_project` survives the fast tier and is not
equivalent: on a `..`-through-a-symlink target where both arms fire, the mutant answers
`path_case_mismatch` naming a different, in-project file instead of `target_outside_project`.
Add the sibling pin.

**P3-4 — ADOPTED (wording).** The composition now has three arms; four comments still say two and
one new sentence says "third half": `project.py:34`, `:36`, `:587`, `errors.py:986`,
`script.py:2266`, `resource.py:791`. Two of those sit directly beside lines this PR already
updated. Also fix the orphan line left at `errors.py:984-986`.

**P3-5 — ADOPTED as a DISCLOSURE only; no code, no published-surface change.** A Unicode
NFD-vs-NFC spelling has exactly the property this refusal exists for — it opens on one filesystem
and fails on another — and gda passes it through (`valid: true`, measured). Widening the check to
normalization is scope creep against the standing proportionality rule, and the published
description claims a CASE check rather than general path portability, so no false promise is made
today. Record the bound in the PR body and as a dated note on issue #845. Do not touch the help,
the catalog or the schema. Honest limit of the evidence: case-sensitive APFS is still
normalization-insensitive, so only the macOS half was measured.

**P3-6 (added by the orchestrator) — DISCLOSURE only.** A mis-cased NON-`.gd` path now reports
`path_case_mismatch` where `invalid_path` came back before (`gda script validate res://Content`).
Both refuse, so no contract breaks, but an agent branching on `invalid_path` sees a different
code. This follows from the design — the gate runs before the operation — so the ordering is NOT
to be changed. State it in the PR body.

- **The defect reproduces on base and is fixed on head.** Base: `valid: true` with the engine's
  own `WARNING: Case mismatch …` on stderr — GDA-DF-062 exactly. Head: `path_case_mismatch`, exit
  4, both spellings in the message and in `evidence.requested_path` / `evidence.stored_path`, no
  `hint` key (F2 held).
- **All three gated commands refuse**, on argv and on `--params-json` (ADR-0015 parity),
  `resource import --dry-run` included. `scene get` is unchanged — scope decision F1 held.
- **The refusal precedes any engine launch — proved, not inferred.** With `GDA_GODOT` pointed at a
  nonexistent path all three still answer `path_case_mismatch`, while the correctly-spelled
  control answers `binary_not_found`. The gate runs before binary resolution, let alone a spawn.
- **Identical verdict on both filesystem kinds, measured on this host**, on a mounted
  case-sensitive APFS volume: mis-cased → `path_case_mismatch`, absent → `path_not_found`,
  correct → `valid: true`; the new module 15 passed / 1 skipped there (the case-INSENSITIVE
  premise test skips), mirroring the host run exactly. AC1 and AC2 are satisfied by direct
  measurement, not only by CI.
- **The casefold caveat the PR discloses is accurate**, measured: `STRASSE.gd` does not open
  `straße.gd` on the case-sensitive volume, and gda reports `path_case_mismatch` naming
  `res://content/straße.gd` — a wrong reason carrying a correction that works.
- **Boundaries probed, all sane:** mid-path component, trailing slash, `res://` vs
  project-relative vs absolute, `..` inside the target, `\`, `//`, `res://` root and `res://.`,
  `user://` / `uid://`, a target outside the root, a 32-component path, a symlinked-in directory,
  a broken symlink, an unreadable directory (`OSError` swallowed — no refusal, no traceback), a
  `--project` spelled relatively / with `..` / in a different case, and batch ordering.
- **Published surface, head vs base:** `gda schema` 0 leaves removed, 3 changed — exactly
  `resource import`, `script run`, `script validate`, each by the one added paragraph; additions
  are 77 commands × 2 properties, byte sizes matching the PR body to the byte. `--help` of all 77
  commands plus every group and the root rendered at `COLUMNS=100`: exactly three differ, each by
  that paragraph. The new code appears in `gda schema` only inside a field description, as
  claimed. README and the three translations correctly untouched — no README row enumerates error
  codes.
- **Guards are live:** dropping the caseless comparison reds 9 of 16 module tests; removing the
  `_EVIDENCE_PRODUCERS` entry reds the producer-set test; rewording the ADR-0002 row reds the
  registry pin; the `render.py` exhaustiveness guard reds on either dropped branch.
- **Gates:** head 2643 passed / 1 skipped vs base 2626 / 0 (+18 collected); Linux CI 2641 / 3
  skips — internally consistent at 2644 collected on both hosts, exactly one of the filesystem
  pair skipping per host. ruff clean, 516 files formatted, pyright 0/0/0. `tests/script
  tests/resource tests/project -m e2e` → 278 passed under the machine-wide lock. PR CI green,
  `mergeStateStatus: CLEAN`.
- **W3 hotspots:** `HARNESS_VERSION` still 22; `src/gda/harness/` and `src/gda/ops/`
  byte-identical to base. The new builder sits above `containment_refusal` without disturbing it
  or `target_outside_project_failure`. `FailureEvidence` keeps `extra="forbid"`. ADR-0004's
  addendum is a dated blockquote in the existing series — no mid-ADR `##` section — and its
  arithmetic checks out as the ninth producer.
- **Proportionality:** ~35 lines of executable logic, no general path-portability mechanism, scope
  held at the gate's three commands. Cost measured at 3.4 ms per target in a project with a
  3000-entry directory (host load 9.4, so an upper bound), paid per target with no memoization —
  negligible for a 4–6 path batch, ~1.7 s for a 500-path batch. Not worth changing.

### Not verified, and why

Linux CI's own filesystem branch directly (a real case-sensitive APFS volume was substituted, and
CI's +1 skip independently confirms its temp is case-sensitive), and the NFD/NFC failure on a
byte-comparing filesystem — case-sensitive APFS is still normalization-insensitive, so only the
macOS half of P3-5 was measured.

qinhaihong-red:
## Review round 1 handled — all six adopted, one commit `4a7978096`

Nothing was declined and nothing was deviated from. Three guards, one wording pass, two
disclosures. PR CI is green at the new head.

| # | Finding | Resolution |
| --- | --- | --- |
| P3-1 | No test for a mis-cased directory with an absent leaf | Fixed at `4a7978096` — `assert case_mismatch("res://Content/absent.gd", project) is None` added to `test_a_name_no_entry_matches_is_left_to_the_operation`, with the reason on the line |
| P3-2 | The exact spelling's preference over a caseless match is unpinned | Fixed at `4a7978096` — `test_a_case_sensitive_host_answers_with_the_spelling_the_caller_named` in the module's case-sensitive branch; run on a mounted case-sensitive APFS volume, not left skipped |
| P3-3 | The third arm's precedence is argued but unpinned | Fixed at `4a7978096` — `test_containment_wins_when_the_spelling_half_fires_too`, beside the #807 ordering pin in `tests/project/test_res_containment_consistency.py`, on the `monorepo` fixture shape |
| P3-4 | Four comments say two arms, one says "third half", one orphan line | Fixed at `4a7978096` — `project.py` module docstring and `case_mismatch`, `errors.py` `containment_refusal` (orphan reflowed), `commands/script.py`, `commands/resource.py` |
| P3-5 | NFD/NFC is the same class of mistake and passes through | Disclosed, no code — PR body ("Open questions") and a dated note on #845 ([comment](https://github.com/aigengame/godot-agent/issues/845#issuecomment-5645533288)) |
| P3-6 | A mis-cased non-`.gd` path reports a different code than before | Disclosed in the PR body ("Public behavior"); the ordering is unchanged |

### The three mutants, each killed by its own test and by nothing else

| Mutant | Killed by |
| --- | --- |
| `case_mismatch`: `entry = part` instead of `return None` | `test_a_name_no_entry_matches_is_left_to_the_operation` — 1 failed / 2643 passed, the mutant offering `res://content/absent.gd` as a stored spelling |
| `_stored_entry`: drop `if name in entries: return name` | `test_a_case_sensitive_host_answers_with_the_spelling_the_caller_named` on the case-sensitive volume — 1 failed / 15 passed / 1 skipped, refusing `res://Content/combat_session.gd` although it exists exactly. It SURVIVES the whole fast tier on this case-insensitive host (2644 passed), which is why the volume run is the evidence |
| `containment_violation`: ask `case_mismatch` before `path_outside_project` | `test_containment_wins_when_the_spelling_half_fires_too` — 1 failed / 2643 passed, `assert 'path_case_mismatch' == 'target_outside_project'` |

Every mutant was reverted and the tree checked back to the intended diff before the commit.

### Gates at `4a7978096`

- Fast tier 2644 passed / 2 skipped / 685 deselected (was 2643 / 1 — one new test, one new
  skip on this case-insensitive host). Linux CI: 2643 passed / 3 skipped, so both new tests
  RUN there and the case-sensitive pin is executed on the host where the condition is
  reachable.
- `ruff check` clean, 516 files already formatted, pyright 0 errors / 0 warnings /
  0 informations.
- `tests/project/test_path_case_mismatch.py` on a mounted case-sensitive APFS volume:
  16 passed, 1 skipped (the case-INSENSITIVE premise test).
- **No published surface changed**: `gda schema --json` byte-identical to `4f5f58242`
  (1,170,595 bytes), and `--help` of the three gated commands at `COLUMNS=100` byte-identical
  (6,495 / 16,498 / 6,790). `src/gda/harness/` untouched, `HARNESS_VERSION` still 22.
- No e2e re-run: the round changes comments, docstrings and tests only, so the e2e tier's
  inputs are byte-identical.

### The two disclosures, measured rather than asserted

- **P3-5.** With `res://content/café.gd` stored NFC (U+00E9), the NFD spelling (`e` +
  U+0301) returns `valid: true`. Measured on an ordinary APFS volume and on the mounted
  case-SENSITIVE one — both are normalization-INSENSITIVE, so the file opens under either
  spelling on this host and only the macOS half of the claim is measured. The failing half
  needs a byte-comparing filesystem (Linux ext4) and was not run. Not widened: the
  description claims a case check, not path portability.
- **P3-6.** The code shift is narrower than "every non-`.gd` path", measured:
  `gda script validate res://Content` and `res://Content/note.txt` (stored
  `res://content/note.txt`) now answer `path_case_mismatch` where `invalid_path` came back,
  while `res://Content/absent.txt` — a mis-cased directory with an absent leaf — still
  answers `invalid_path`, because the gate says nothing when a component names no entry.
  Both codes are refusals at exit `4`.

### One line beyond the six, called out rather than folded in

`containment_refusal`'s summary line said "The refusal when `target` does not belong to
`project`", which the spelling arm contradicts — a mis-cased target DOES belong to the
project. It is the head of the very docstring P3-4 asks to fix, so it is reworded to "is
not `project`'s to serve as spelled". Docstring only; no published surface.

qinhaihong-red:
## Round 2 — verified directly by the orchestrator. **READY at `4a7978096`.**

Round 1 was APPROVE with five P3s and no P2, and the fix round changed no executable line — three
tests, six comment/docstring rewordings and two disclosures — so a second full review would have
cost more than it could find. I verified it myself instead.

**The round-2 diff is exactly what it claims.** Six files: `tests/project/test_path_case_mismatch.py`
(+25), `tests/project/test_res_containment_consistency.py` (+36), and comment/docstring-only
changes in `src/gda/project.py`, `src/gda/errors.py`, `src/gda/commands/script.py`,
`src/gda/commands/resource.py`. I read all four source hunks line by line: no executable statement
changed.

**No published surface moved.** `gda schema --json` is **byte-identical** to `4f5f58242`
(1 170 595 bytes, `cmp` clean). `script validate`, `script run` and `resource import` `--help`
byte-identical too (6 495 / 16 498 / 6 790 bytes — distinct sizes, so the argv was really split and
the comparison is not vacuous). `src/gda/harness/` untouched, `HARNESS_VERSION` still 22. README
and the three translations untouched.

**Gates:** fast tier **2644 passed / 2 skipped / 685 deselected** (was 2643 / 1 — one new test, one
new skip on this case-insensitive host, two collected). `ruff check` clean, 516 files formatted,
`pyright` 0 errors / 0 warnings / 0 informations.

**The precedence pin kills its mutant, and only it.** I re-applied the P3-3 mutant myself — ask
`case_mismatch` before `path_outside_project` in `containment_violation` — and the fast tier
answered **1 failed, 2643 passed**: exactly
`test_containment_wins_when_the_spelling_half_fires_too`, on
`AssertionError: assert 'path_case_mismatch' == 'target_outside_project'`. Reverted and confirmed
clean.

**Reported by the implementer, accepted:** the other two mutants each killed by their own test
alone, the `_stored_entry` one shown on a **mounted case-sensitive APFS volume** (it survives the
whole fast tier on this case-insensitive host, which is precisely why the volume run is the
evidence and why the pin belongs in the case-sensitive branch, where Linux CI runs it). Linux CI
moved 2641 → 2643 passed, so both new tests execute there.

**Both disclosures were re-measured rather than relayed**, and one came back NARROWER than the
finding: P3-6 is not "every non-`.gd` path" — `res://Content` and `res://Content/note.txt` answer
`path_case_mismatch`, while `res://Content/absent.txt` still answers `invalid_path`, because the
gate says nothing when a component names no entry. Both exit 4. P3-5's NFD/NFC case was measured on
both the ordinary and the case-sensitive APFS volume (both normalization-insensitive), so only the
macOS half is evidence; the failing half needs Linux ext4 and was not run.

**One line beyond the six, disclosed rather than folded in silently:** `containment_refusal`'s
summary said "the refusal when `target` does not belong to `project`", which is untrue of a
mis-cased target — it does belong. Reworded to "is not `project`'s to serve as spelled". It is the
head of the very docstring P3-4 asks to fix, it is docstring-only, and schema and helps stayed
byte-identical. Accepted.

Commits: two on top of dev, author and committer `haihong.qin <haihongqin@gmail.com>`, no
attribution of any kind. PR CI green, `mergeStateStatus: CLEAN`.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `4a7978096589b8220ce195aef14ff65cfabb608f`. Existing reviews and replies were not used as evidence.

`src/gda/skill/SKILL.md:145-147` equates macOS/Windows with case-insensitive filesystems and Linux with case-sensitive filesystems. Those operating systems can host either behavior; the PR's own test/probe prose acknowledges that. Remove the OS parentheticals or qualify them as common defaults.

**Terminology and prose assessment:** I inspected issue #845 and all scope comments, the PR body, ADR-0002/0004/0006 changes, the command catalog, skill, help/schema prose, and error text against `CONTEXT.md` and ASD-STE100. `path_case_mismatch`, `requested_path`, and `stored_path` are consistently used. The OS-wide claim above is the only independent prose finding. I did not test every filesystem's Unicode folding rules; that uncertainty is itself relevant to the code findings below.

`gda.project` is the correct owner of project-path spelling, `gda.errors` correctly owns the envelope, and all three gated commands depend downward on that decision. The authority is not deterministic or semantically total in two legal directory states, however: `src/gda/project.py:576-579` returns the first `casefold()` match, so ambiguous matches depend on enumeration order, and Unicode casefold equivalence can identify a different lowercase filename as a case-only correction. Preserve the current owners, but make ambiguity an explicit no-correction/refusal outcome and define a portable case-only relation. Claim-evidence state: `open` until ambiguity and Unicode counterexamples pass on the authority directly.

## Standards

- Hard: Python `casefold()` makes `straße.gd` and `strasse.gd` equal although both are lowercase and distinct. Lines 576-579 can replace truthful `path_not_found` with `path_case_mismatch` and recommend another file, contrary to `src/gda/error_codes.py:286-289`.
- Hard: when both `FOO.gd` and `foo.gd` exist, request `Foo.gd`; the first `os.listdir()` result wins, so `stored_path` is nondeterministic.
- Hard: the OS-wide skill statement above violates the affected-artifact truthfulness rule in `RULES.md:107-113`.

No additional baseline smell was actionable.

Standards summary: 3 hard findings; worst is an unsafe/nondeterministic correction.

## Spec

- Required: issue #845 requires a deterministic result. The ambiguous case above alternates between two `stored_path` values as enumeration order changes. Detect ambiguity or define a stable, safe rule.
- Required: the issue says the extra directory listing occurs only on the not-found path. `case_mismatch()` is called unconditionally at `src/gda/project.py:728` and lists once per component at lines 635-636; an exact two-component path performs two listings. Implement the promised fast path or reconcile the requirement and add a cost-bound regression.

The three shared gates, evidence/no-hint shape, exact spelling, unchanged `scene get`, registry, and accepted-spelling E2E otherwise match the issue. Independent validation passed 73 targeted tests (2 skipped) and the new real-engine E2E.

Spec summary: 2 findings; worst is nondeterministic typed evidence.

qinhaihong-red:
## Third review handled at `87e052950`

| `casefold()` makes `straße.gd` and `strasse.gd` equal, so a truthful `path_not_found` became a `path_case_mismatch` recommending a different file | Fully Adopt | reproduced on the authority | the relation is SIMPLE case folding (`str.lower`, one character to one — what a case-insensitive filesystem does), never full folding | `test_full_case_folding_never_corrects_to_a_different_file`: `strasse` and `STRASSE` → no correction, `STRAßE` (a true case-only pair) → corrected; reverting to `casefold` reds exactly it |
| Both `FOO.gd` and `foo.gd` present, request `Foo.gd`: the first `os.listdir` result wins, so `stored_path` is nondeterministic | Fully Adopt | reproduced on a mounted case-sensitive APFS volume | a name that folds equal to MORE than one entry is no match; the operation's `path_not_found` stands | `test_an_ambiguous_caseless_match_is_no_correction` (runs on a case-sensitive host — 19 passed / 1 skipped on the volume; the "first match wins" mutant reds exactly it there) |
| The issue promises the extra listing only on the not-found path; the authority lists once per component on every call | Partially Adopt | true — and unimplementable on a case-insensitive host, where `exists()` says "yes" for a mis-cased path, so the listing is the only way to learn "not found" | the REQUIREMENT is reconciled (dated note on #845) and the cost is pinned rather than a fast path built: `test_the_authority_lists_each_directory_component_once` counts one `os.listdir` per component, exact or mis-cased. Round 1 measured the cost at 3.4 ms per target in a 3 000-entry directory | test passes on head |
| The skill equates macOS / Windows with case-insensitive and Linux with case-sensitive | Fully Adopt | either OS can host either kind | "the common default on …" | prose |

Head `87e052950`, one commit on `4a7978096`; project fast tier 286 passed / 3 skipped (the ambiguity case skips on this case-insensitive host and runs on the volume and on Linux CI); the two case-mismatch e2e cases pass on a real engine under the lock; ruff / format / pyright clean; CI green. Schema and helps unchanged.

qinhaihong-red:
## Follow-up recheck — functional findings closed; one P3 wording claim remains at exact head `87e052950`

The original functional findings are closed: `strasse` no longer aliases `straße`, ambiguous folded matches return no correction, the per-component listing cost is recorded, and the Skill now describes OS behavior as common defaults rather than universal rules.

The replacement rationale introduces one inaccurate contract statement in the PR response, the issue #845 correction, and `src/gda/project.py`: it calls `str.lower()` “simple case folding” and says it is “one character to one.” Python lowercase mapping can expand a character. A direct probe on this exact environment gives:

```text
"İ".lower() == "i̇"
input code points: 1; output code points: 2
```

This does **not** reopen the fixed `ß`/`ss` or enumeration-order defects. It leaves the intended behavior for expanding Unicode lowercase mappings inaccurately specified, which matters because the prose uses non-expansion as the reason this relation is safe.

Minimum closure: either describe the implemented relation precisely as equality under Python `str.lower()` and acknowledge expanding mappings, or actually define/reject multi-code-point lowercase expansions and add an `İ` case. Do not call the current primitive one-to-one.

Independent validation: 17 focused tests pass with 3 filesystem-dependent skips; current CI is green with no pending jobs. The direct string probe is outside the existing matrix.

qinhaihong-red:
## Fourth review handled at exact head `4f90d4515`

| "Simple case folding, one character to one" misdescribes `str.lower()`: Python's lowercase mapping can expand (`"İ".lower()` is two code points), and the prose used non-expansion as the reason the relation is safe | **Fully Adopt** (describe the relation precisely) | reproduced: `len("İ".lower()) == 2` on this environment | the docstring in `project.py`, the test comment, and the issue #845 note (dated correction appended to the note) now state the relation as EQUALITY UNDER `str.lower()` and nothing narrower: the mapping is one-to-one except where Unicode expands it, an expanded spelling folds equal to the stored one like any other case difference, and the ambiguity rule still applies. What makes the relation safe is not non-expansion but the two rules already in place — an ambiguous match is no correction, and the result is a spelling to re-issue, never acted on | `test_the_relation_is_equality_under_lower_expansions_included`: stored `İcon.gd`, a request spelled `i` + U+0307 is corrected to it, a request spelled plain `icon.gd` is not |
| Alternatively define or reject multi-code-point lowercase expansions | **Do Not Adopt** | no product promise covers it; the `ß`/`ss` defect was a false correction to a DIFFERENT file, which an expansion cannot produce (both spellings name one stored entry, or the ambiguity rule declines) | no new rule; the `İ` case is pinned as data, not as a supported scenario | residual: a case-insensitive host whose own table disagrees with Python's in that corner answers its own way, which the "re-issue, never act" design already bounds |

CI is green on `4f90d4515`. Locally: project tier 287 passed, `ruff` clean.

Issue #846, opened 2026-09-05
perf monitors --frames: aggregate-only mode and bounded observer memory


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-066** (2026-08-26, gda 0.11.0, Godot 4.6.3). `gda perf monitors --frames 600 --budget ... --json` in a two-enemy production combat returned every raw sample (a 68,215-token result, truncated by the client) and, during an otherwise stable window, `static_memory` rose by 1,229,440 bytes while object, node, resource, and orphan counts stayed flat. The harness retaining 600 sample dictionaries is the most plausible source, and it turns the observer into a monotonic slope that reads like a game leak. The production moved its no-growth gate to a separate in-process test.

Verified against gda 0.13.0 (`266c8009f`): in window mode the `perf-sample` sampler appends one Dictionary per frame through the shared #223 multi-frame window base (also used by `perf monitor`, `perf sample`, `screen capture`, `screen frames`, and `input sequence`) and the harness reply carries them all; the CLI computes the aggregates from the samples; the window schema requires `samples`. The sibling `screen frames` already has `--summary` (#665): every frame is still written, and the result carries the aggregate instead of the per-frame list.

## What to build

- `--summary` on `perf monitors --frames`, the same spelling and meaning as `screen frames --summary`: the result carries `stats`, `budget`, and `passed` but `samples: null` with `samples_omitted: true`; the schema's window branch admits the omitted form. The default stays as today (additive change).
- Bound the observer (the harness-side `perf-sample` sampler): the sampler and its finalizer store the window per monitor in packed arrays (a `PackedFloat64Array` per monitor) instead of per-frame Dictionaries; the shared window base is unchanged, so the five other window ops are untouched. With `--summary` the harness reply still carries the packed values (the CLI computes the aggregates), but the RESULT the caller reads omits them.
- Disclose the observer's own footprint: the result reports the retained sample count and the approximate bytes the collector held (`collector_bytes`, computed from the packed sizes), so a `static_memory` delta can be attributed.
- Help states that a long window allocates on the engine side and how to read `collector_bytes` against `static_memory`.

## Acceptance criteria

- [ ] `perf monitors --frames 600 --summary --json` returns `stats`, `budget`, `passed`, `samples: null`, `samples_omitted: true`, and `collector_bytes`.
[...]

PR #967, opened 2026-09-09
feat(gda): game rect reports the minimum sizes and the local rect (#852)

### LINKED ISSUE #852: live geometry: game rect adds minimum sizes; game get on Control layout properties names game rect
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-071** (2026-08-27, gda 0.11.0, marked high for UI layout diagnosis). `game get NODE --property custom_minimum_size` succeeded, but `position` and `size` on the same Control returned `live_unknown_property`, so the caller could read authored minima but not the post-layout rectangles behind a visible overlap. The entry asks for `position`, `size`, `global_position`, `global_rect`, `get_minimum_size()`, and `get_combined_minimum_size()`.

Verified against gda 0.13.0 (`266c8009f`) and Godot 4.6.3: `game get` reads storage properties (plus explicit script variables). A Control's `position` and `size` are editor-usage properties and `global_position` carries no usage flags at all, so none of the three is in the storage set; `global_rect` is not a property — `Control.get_global_rect()` is a method. `game rect` already returns the rendered global rect (`position` and `size` from `get_global_rect()`), which is what the diagnosis needed; the entry did not use it. Two gaps remain: the minimum-size getters, and a `live_unknown_property` message that leaves an agent stranded on exactly the spellings `game rect` serves.

## What to build

- `game rect` additionally reports `minimum_size` (`get_minimum_size()`), `combined_minimum_size` (`get_combined_minimum_size()`), and the Control's local rect as `local_position` / `local_size`, beside the existing global `position` / `size`, whose meaning is unchanged.
- `game get` on a Control keeps returning `live_unknown_property` for the four spellings a caller reaches for — `position`, `size`, `global_position`, `global_rect` (the last a method, not a property) — but the message names `game rect` as the read for layout output and, for a free Control, the offsets and anchors as the storage properties — a direct child of a `Container` does not carry them (the engine strips their storage flag), so it is told `custom_minimum_size`, the size flags and the parent's layout instead; the `position` setter refusals (live and headless) name the same inputs from the same shared statement (corrected 2026-09-13 after the third review of PR #967; the dated comment below records the change).

## Acceptance criteria

- [ ] `game rect --json` on a Control inside a Container reports the global rect, `local_position`, `local_size`, `minimum_size`, and `combined_minimum_size`; a live regression asserts them against the fixture's layout.
- [ ] `game get --property size` and `game get --property global_rect` on a Control return `live_unknown_property` whose message names `game rect`; a non-Control keeps the current message.
- [ ] Schema and help are updated for the new `game rect` fields.

## Blocked by

None - can start immediately.

qinhaihong-red:
## Summary

`gda game rect` reported only `Control.get_global_rect()`, so a UI diagnosis could read the
authored `custom_minimum_size` through `game get` but never the laid-out result behind a
visible overlap (GDA-DF-071, kung-fu dogfooding). It now reports the whole layout read, and
`game get`'s refusal for the four spellings a caller reaches for on a `Control` names the
command that serves them.

## Public behavior

**`game rect` — four new required result fields.** Beside the unchanged global
`position` / `size`:

| Field | Engine read | Why it is its own field |
| ----- | ----------- | ----------------------- |
| `local_position` | `Control.get_rect().position` | the same rectangle in the PARENT's space. It is the origin of the node's own transform — the `position` property while `pivot_offset` is zero, or while `scale` and `rotation` are default, because the internal origin is `pivot - R*S*pivot` — and it differs from the global point by the ancestors' TRANSFORM |
| `local_size` | `Control.get_rect().size` | the node's `size` property multiplied by the node's own `scale`; it differs from the global size only under an ANCESTOR scale |
| `minimum_size` | `Control.get_minimum_size()` | the class's intrinsic minimum, EXCLUDING the authored `custom_minimum_size` (`[0, 0]` for a plain `Control`) |
| `combined_minimum_size` | `Control.get_combined_minimum_size()` | the per-axis maximum of the two — the size a parent `Container` honors |

All four are required, so a stale installed harness is a `contract_violation` rather than a
partial rect a caller could read as complete. The human rendering keeps its first line
byte-identical and adds two labelled lines.

**`game rect` runs project code, narrowly.** `Control::get_minimum_size()` is
`GDVIRTUAL_CALL(_get_minimum_size)` with no cache, so where a class leaves that getter to
`Control` the addressed node's script override of it runs once per request, and TWICE where
the combined read finds the minimum-size cache stale. `get_combined_minimum_size()` beside it
reads that cache first and recomputes through the SAME virtual when it is stale, so it adds no
point of its own — but it does add a call. Measured on a counting override in a running game,
two nodes of the same script:

| | visible `Control` | `visible = false` |
| - | - | - |
| each `game rect` | +1 | +1 |
| a `custom_minimum_size` write | +1 | **+0** |
| the next `game rect` | +1 | **+2** |
| the one after it | +1 | +1 |

`update_minimum_size()` invalidates the cache and then returns at `!is_visible_in_tree()`
BEFORE queueing its deferred revalidation, so on a hidden Control nothing refills the cache and
the next `game rect` pays for both reads. Idle `game get` reads never move the counter. A
`Button` subclass with the same script stays at 0, because `Button::get_minimum_size()`
overrides the C++ getter and never reaches the virtual — hence the "where a class leaves that
getter to `Control`" qualifier rather than a claim about every `Control`. This is one more
point on `CONTEXT.md`'s `Project-code execution surface`, inside ADR-0009's `Trusted project`
assumption and adding no new trust axis; the term is updated.

**`game get` — a redirect in the existing refusal, branched on the parent.** On a `Control`,
`--property position`, `size`, `global_position` and `global_rect` stay
`live_unknown_property` (none of them is a storage property: the first two carry editor usage
only, `global_position` no usage flags, and `global_rect` is a method). The message now names
`gda game rect <path>` and the fields that read reports, then the layout INPUTS — and WHICH
inputs a node carries depends on its parent, so the last sentence branches on the same
`_has_container_parent` predicate `game set --property position` already uses:

| Parent | The refusal names | Why |
| ------ | ----------------- | --- |
| not a `Container` | `offset_left` … `offset_bottom`, `anchor_left` … `anchor_bottom` | they are on the node's storage set |
| a `Container` | `custom_minimum_size`, `size_flags_horizontal`, `size_flags_vertical`, and the parent `Container`'s own layout | the engine drops `PROPERTY_USAGE_DEFAULT` from `offset_*` / `anchor_*` / `grow_*` / `anchors_preset` on a direct child of a `Container` (`Control::_validate_property`, not `is_editor_hint()`-gated), so naming them there would send the caller into a SECOND `live_unknown_property` — the dead end this slice exists to remove, one step further in |

Any other node keeps the current generic message. **No new `Gda error code`.**

Harness `HARNESS_VERSION` 22 -> 23 (first of W3's two harness slices; #846 takes 24).

## Surface diff (enumerated)

Captured from a separate detached checkout of the base (`b7c485693`) and this branch:
`gda schema` plus every `--help` surface the live command tree has (93 — the root, the 15
groups, and the 77 commands). Exactly three of those helps differ, plus `gda schema`;
everything else is byte-identical, including every other command in `gda schema`.

`gda schema` — exactly two commands differ (`game get`, `game rect`); no command added or
removed, no top-level key changed:

1. `game rect` `output.properties` gains `local_position`, `local_size`, `minimum_size`,
   `combined_minimum_size` (each `array` of `number`, with its description).
2. `game rect` `output.required` gains the same four names.
3. `game rect` `output.description` (the result model's docstring) rewritten for the layout read.
4. `game rect` `description` (the command help) rewritten — see the help diff below.
5. `game get` `description` gains one paragraph about the four Control spellings and the two
   branches of the redirect.

`gda game rect --help`: the summary line becomes "Read a running Control's layout output
(live)."; the body paragraph now names all six geometry fields, points at the four spellings
`game get` refuses, and states the one `_get_minimum_size` call. Options, arguments and the
precision paragraph are unchanged.

`gda game get --help`: one new paragraph after the `--texture-digest` paragraph. Nothing
else changed.

`gda game --help`: the group listing carries `game rect`'s one-line summary, which is derived
from the docstring above, so its `rect` row changes with it. Nothing else in the listing
changes.

Other agent-facing surfaces changed by hand: `CONTEXT.md` (the `Project-code execution
surface` term), `src/gda/skill/SKILL.md` (the `game` group row and the Control-layout
paragraph), `docs/command-catalog.md` (the `game rect` / `game get` entries and the
Control-position policy line), `README.md` (the `game rect` row plus one sentence for the
`game get` boundary) and the three translations, re-stamped with
`scripts/update_readme_i18n.py`.

## Validation

Red then green: the model/render/fixture/guard tests were written first and failed for the
right reasons (7 failures naming the missing fields, the missing `CONTROL_LAYOUT_READS`
declaration and the old render line), then passed after the harness and command changes.

| Gate | Command | Result |
| ---- | ------- | ------ |
| Fast tier | `uv run --frozen pytest -m "not e2e" -q -p no:cacheprovider` | 2633 passed, 686 deselected |
| Format | `uv run --frozen ruff format --check .` | 515 files already formatted |
| Real engine | `pytest tests/daemon tests/harness tests/live -m e2e -q` under the machine-wide Godot lock | 102 passed (0 skipped), 3m21s |
| Harness guard | `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | "harness version transition is valid" |

The behaviour both branches of the redirect promise was then read off a RUNNING game: on a
container-managed `Control`, `game get --property size` returns the container branch and every
property it names (`custom_minimum_size`, `size_flags_horizontal`, `size_flags_vertical`)
resolves, while `--property offset_left` on that same node is `live_unknown_property`; on a
free `Control`, `game get --property position` returns the `offset_*` / `anchor_*` branch and
those resolve. The e2e test asserts both halves, including that FOLLOWING the redirect
succeeds.

Mutants (each reverted after the check; the harness hash pin cannot discriminate a harness
mutant, so it is deselected when judging one):

| Mutant in `gda_harness.gd` | Killed by |
| -------------------------- | --------- |
| `"local_position": _jsonify(control.position)` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| `"local_size": _jsonify(control.size)` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| `"combined_minimum_size": _jsonify(control.custom_minimum_size)` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| `"minimum_size": _jsonify(control.get_combined_minimum_size())` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| the container branch never fires (`if control == null:`) | e2e `test_daemon_game_get_names_game_rect_for_a_controls_layout_reads`, and fast-tier `test_game_get_names_game_rect_for_the_control_reads_it_cannot_serve` |
| the container branch APPENDS its sentence (`inputs +=`) instead of replacing the free-Control one | the same two tests, since round 2. Both asserted only PRESENCE, so this mutant survived them and reopened the dead end: the container child was told the container inputs AND the `offset_*` / `anchor_*` list. The e2e test now asserts those names are ABSENT for the container child, and the fast-tier guard reads the assignment itself |

Engine facts the assertions rely on, recorded in the tests that use them
(`tests/daemon/test_e2e_daemon.py`, beside the `MIN_SIZE_MAIN_TSCN` fixture):

- `Control::get_minimum_size()` returns the `_get_minimum_size` virtual, which a scriptless
  plain `Control` does not implement — so the fixture's `Box` must report `[0, 0]`, which is
  what makes it a different read from the combined one.
- `Control::_update_minimum_size_cache()` takes the per-axis maximum of that and
  `custom_minimum_size` — so `Box` must report `[80, 20]` for the authored
  `custom_minimum_size = Vector2(80, 20)`, and the fixture's `Tag` (a `Button` authored
  `Vector2(300, 1)`) must report a combined minimum equal to NEITHER input, because the
  authored width wins on x and the theme-driven intrinsic height wins on y. The intrinsic
  height is read from the result rather than pinned, so no theme number enters the assertion.
- `Control.get_rect()` is built from the LOCAL TRANSFORM: its origin is
  `position + pivot_offset - R*S * pivot_offset` (so it is `position` itself at a zero
  `pivot_offset`) and its size is the transform's own scale times `size`, which a rotation
  does not change — measured at 45 degrees. The
  fixture's `Scaled` node (`scale = Vector2(2, 3)`, `pivot_offset = Vector2(10, 10)`, offsets
  30/40/130/90) therefore reports `local_position [20, 20]` against a `position` property of
  `(30, 40)`, and `local_size [200, 150]` against a `size` property of `(100, 50)`.
- `Control.position` is in the parent's rect space, so the container-managed child's local
  origin is its global origin minus the container's. The test asserts that RELATION against a
  second `game rect` read rather than a pixel value, and asserts the local rect differs from
  the global one.
- `BoxContainer` fits each child into a rect no smaller than its combined minimum, asserted
  as an inequality (no theme-dependent number).

PR CI runs **no Godot e2e** (that tier is schedule/dispatch-only), so the cross-language half
of this change is guarded at unit speed by two source-level mirrors in
`tests/live/test_live_contract_guards.py`:

- `test_game_rect_reply_keys_mirror_its_published_result_model` — the keys
  `_handle_game_rect` writes must equal `GameRectResult`'s fields. A field the model requires
  and the harness never writes would otherwise be a `contract_violation` visible only on a
  real engine.
- `test_game_get_names_game_rect_for_the_control_reads_it_cannot_serve` — the four declared
  spellings, the redirect's content (both branches, and the container-parent predicate it
  branches on), and its Control-only guard, read off the harness source.

Also on PR CI: the schema/help contract (`tests/meta`), the result-model and refusal relay
through the fake live runner (`tests/live/test_game_commands.py`), the human rendering
(`tests/cli/test_render.py`), the float-writer contract guard (every new float declares the
engine writer), and the harness hash pin + version guard.

Needing the real engine (the e2e tier, run locally for this PR): that Godot actually lays the
fixture out and produces those numbers, that the four spellings really are absent from a live
`Control`'s storage set, and — the container branch's whole point — that each property the
redirect names IS on the addressed node's storage set.

## Open questions

- The dispatch brief said this issue carried dated 2026-09-07 comments fixing four round-1
  findings. Issue #852 has **no comments** and was last updated 2026-09-05, so the issue body
  was used as the authority. Nothing in the delivered scope depends on the difference.
- The issue body asked the refusal to name "the offsets and anchors as the storage
  properties". That premise does not hold for a container-managed `Control`, so the delivered
  message branches instead. #852 carries a dated note recording it; the acceptance criteria
  are untouched.
- `game set` on the same four spellings keeps the generic message (the issue scopes the
  redirect to `game get`); the free-Control `position` write is unchanged.
- Out of scope, **to be filed (the user's call)**: `game set --property position`'s container
  refusal (`_control_position_unavailable_message`) tells the caller to address `offset_left` …
  on the same container child where those names are not storage properties either. The same
  wrong advice also stands in two documents this PR edits — `SKILL.md`'s container paragraph
  ("use those offset properties explicitly instead") and `docs/command-catalog.md`'s `node set`
  row — so the follow-up covers the harness message AND those two sentences. It is left out of
  this slice on purpose: it is pre-existing, it is a different command and a different message,
  and fixing the harness would move the version ladder W3 depends on. No issue exists yet.

Closes #852


qinhaihong-red:
## Review round 1 — CHANGES REQUIRED, 0 P1 / 2 P2 / 3 P3 (all five adopted)

Independent adversarial review at `96e9e093f` against the base `b7c485693`, on a real engine
(Godot 4.6.3). The mechanism is sound and **measured** correct: all four new numbers were compared
field by field against what the running game itself reports, on **eleven Controls** — offset
ancestor, scaled ancestor, self-scaled and pivoted node, theme-driven minimum, a
`custom_minimum_size` below the theme minimum, a per-axis maximum, a container child whose
combined minimum exceeds its own, nested containers, and a script-provided `_get_minimum_size` —
fetched through `game call` + `GDA_CALLABLE`. Every field matched on every node.

Both P2s are **published text that is false against the running engine**, not broken behaviour.

### Findings and dispositions

**P2-1 — the redirect names layout inputs a container-managed Control does not have. ADOPTED, and
adopted as the CODE remedy.** The new refusal ends by naming `offset_*` / `anchor_*` as "the
storage properties". For a Control whose parent is a `Container` — the dogfooding case, GDA-DF-071
is literally "after container layout" — the engine drops `PROPERTY_USAGE_STORAGE` from exactly
those names (`Control::_validate_property`, not `is_editor_hint()`-gated), so they are absent from
the storage set and following the redirect lands in a SECOND `live_unknown_property` carrying the
generic message. That is the dead end this slice exists to remove, recreated one step further in.
Measured: `game get …/Box --property size` → the redirect; `--property offset_left` → generic
`live_unknown_property`; `game set --property offset_left` → the same; while the same calls on a
free Control succeed. The PR pins the false claim in an e2e assertion on a container-managed child.

Wording-only was the cheaper option and is **rejected**: qualifying five surfaces with "unless the
Control is in a Container" leaves the agent with a redirect that does not say where to go in the
common case. The harness already has `_has_container_parent(control)` and already branches on it
for `game set --property position`, so the fix is one branch on an existing predicate — a reuse,
not a new mechanism. Free Control keeps the `offset_*` / `anchor_*` list; a container-managed
child gets `custom_minimum_size`, `size_flags_horizontal` / `size_flags_vertical` and the parent
Container's own layout. The e2e assertion must then cover both branches, and the five surfaces
that carry the sentence must say the same thing.

Note for the record: the issue body itself asked for "the offsets and anchors as the storage
properties", so the implementer followed the brief and the brief's premise is what is wrong. The
issue is amended with a dated note rather than silently deviated from.

**P2-2 — `game rect` now runs project code and `CONTEXT.md`'s `Project-code execution surface` is
not updated. ADOPTED (wording).** `Control::get_minimum_size()` is `GDVIRTUAL_CALL(_get_minimum_size)`
with no cache, so every `game rect` invokes the addressed Control's override once. Measured on a
counter: 4 → `game rect` → 5, and idle reads do not move it, so the call is attributable to the
command. That term is meticulous about this — it gives `game call` its own "ONE narrow point"
clause and states that reading `GDA_CALLABLE` adds no point at all — and the repo's DRY/state-sync
rule makes the update required. One clause in `CONTEXT.md`, plus an honest half-sentence in
`docs/command-catalog.md`, which currently says `game rect` "remains a read-only query" while the
command can now run project code with side effects. Stays inside ADR-0009's Trusted-project
assumption; no new trust axis.

**P3-1 — `local_position`'s description is false under `pivot_offset` + scale, and the choice is
unpinned. ADOPTED, both halves.** The text says `get_rect().position`, "which is the node's own
`position`"; the engine builds `get_rect()` from the local transform, whose origin is
`position + pivot − S·pivot`. Measured on a node with `scale (2,3)`, `pivot_offset (10,10)`,
offsets `(30,40)`: `local_position [20,20]` against `position [30,40]`. Two mutants —
`control.position` for `local_position`, `control.size` for `local_size` — survive the **entire**
fast tier AND both new e2e tests. So the clause is corrected (the implementation is the better
choice) **and** the fixture gains a scaled node, so the guard discriminates. Same family: the PR
body and the harness comment say local differs from global "by the ancestors' offsets" — under an
ancestor scale it is the ancestor **transform** (measured `(10,5)` → `(50,55)`).

**P3-2 — `combined_minimum_size` is not distinguished from `custom_minimum_size`. ADOPTED (test).**
That mutant passes the full fast tier and both new e2e tests, because the fixture's `Box` has the
two equal. The field's defining semantics — the per-axis maximum of the intrinsic and the authored
minimum — is never observed. Non-equivalent, measured: a `Button` with `custom_minimum_size
(300,1)` and intrinsic `(23,31)` gives `combined (300,31)`. One fixture node plus two asserts;
it can share the P3-1 fixture edit.

**P3-3 — the PR body's surface enumeration is incomplete. ADOPTED (wording).** `gda game --help`
(the group listing) also changed, because it carries `game rect`'s summary line. Derived and
harmless, but "everything not listed below is byte-identical" is then untrue. Verified by
capturing all 97 help surfaces on both checkouts: exactly three differ.

- **The four numbers are what Godot reports**, on 11 Controls, each field compared against the
  running game's own `get_global_rect()` / `get_rect()` / `position` / `size` /
  `get_minimum_size()` / `get_combined_minimum_size()`.
- **The four `game get` spellings really are absent** from a live Control's storage set, measured
  on three Controls, with the engine bindings confirming `position`/`size` as
  `PROPERTY_USAGE_EDITOR`, `global_position` as `PROPERTY_USAGE_NONE`, and `global_rect` as a
  method.
- A non-Control keeps the generic message; a script variable still wins over the redirect, so it
  fires only when nothing was found.
- **The human render's first line is byte-identical**: both renderers fed the same five shared
  values emit the same line; head adds two indented lines.
- **Schema diff is exactly the two claimed commands** — 77 commands both sides, none added or
  removed, `game get` changing only its description and `game rect` gaining the four properties,
  their four `required` entries and two descriptions.
- **ADR-0023 holds:** `classify=None`, `recipe=None`, no validators, no private attrs, empty
  `model_config`; on a real engine reply `model_dump() == reply`, `validate(dump(m)) == m`, JSON
  round trip equal, `model_fields_set` exactly the nine public fields. No hidden wire-presence flag.
- **ADR-0018 holds:** no version negotiation, fallback, source flag or applied-mode counter; all
  four fields required, so a stale harness yields a typed `contract_violation`.
- **Harness:** `sha256(gda_harness.gd)` recomputed and matching the pin; 22 → 23; the guard valid;
  edits confined to `_handle_game_get`, `_handle_game_rect`, one const and one message helper —
  `_handle_perf_sample` (#846's region) and `_serialize` untouched, `operations.gd` untouched.
- **Merge hazards checked, not assumed:** `git merge-tree` against #939 (#929) is **clean**, and
  the hunk regions are disjoint in `game.py` and `test_game_commands.py`; #939 does not touch
  `tests/support.py` at all, so the `GAME_RECT_RESULT` edit has no keep-both hazard. Against #973
  (#846) the only conflicts are the expected ones — `HARNESS_VERSION`, the hash pin, the three
  README stamps — and the two `gda_harness.gd` regions do not touch.
- **Gates:** unit 2633; ruff clean; 515 files formatted; pyright 0/0/0; `tests/daemon
  tests/harness -m e2e` 42 passed under the machine-wide lock; PR CI green, CLEAN.
- **README + translations:** re-running the stamp script is a no-op, all three stamps equal
  `sha256(README.md)`, 93 row keys in each of the four files with identical sets — no row lost.
- The PR's "Open questions" are accurate: issue #852 carries **0 comments**, so its body is the
  authority.
- The PR's three mutant claims all reproduced. Useful caveat recorded: any harness edit also fails
  the hash-pin test, so that test cannot discriminate harness mutants and must be deselected when
  judging them.

### Out of scope, recorded for its own issue

`_control_position_unavailable_message` — the `game set --property position` container refusal —
tells the caller to "address offset_left, offset_top, offset_right, and offset_bottom instead" on
the same container child where those writes fail, measured above. Pre-existing, a different
command and a different message; it is not folded into this slice.

### Not verified, and why

The red-then-green history (not reconstructible from the head commit); a real shipped v22 harness
against a v23 CLI (ADR-0018's current-harness policy puts it out of scope by design); non-macOS
behaviour (nothing in the diff is platform-conditional); MCP tool regeneration (it derives from
`gda schema`, which was diffed).

qinhaihong-red:
## Review round 1 handled — all five findings landed in one commit (`1f52d7e06`)

Every finding was adopted as written. The two P2s are corrections to behaviour and to a
published fact; the three P3s close a test gap and two wording defects.

| # | Finding | Resolution |
| - | ------- | ---------- |
| P2-1 | the redirect names layout inputs a container-managed `Control` does not have | FIXED (code) — `_control_layout_read_message` now branches on the existing `_has_container_parent(control)` predicate |
| P2-2 | `game rect` runs project code; `CONTEXT.md` does not say so | FIXED (wording) — one clause in the `Project-code execution surface` term, plus `docs/command-catalog.md` |
| P3-1 | `local_position`'s description is false under `pivot_offset` + scale, and the choice is unpinned | FIXED (wording + test) — the clause is corrected and the fixture gained a scaled + pivoted node |
| P3-2 | `combined_minimum_size` is not distinguished from `custom_minimum_size` | FIXED (test) — one `Button` fixture node whose combined minimum equals NEITHER input |
| P3-3 | the PR body's surface enumeration is incomplete | FIXED (wording) — `gda game --help` added; the body is rewritten |

### P2-1 — the redirect branches on the parent

`Control::_validate_property` drops `PROPERTY_USAGE_DEFAULT` from `offset_*`, `anchor_*`,
`grow_*` and `anchors_preset` on a direct child of a `Container`, and that branch is not
`is_editor_hint()`-gated — so the previous message sent a container-managed caller into a
second `live_unknown_property`. The last sentence now branches:

* free `Control` — the `offset_*` / `anchor_*` list, byte-identical to before;
* direct child of a `Container` — "This Control is a direct child of a Container, which owns
  its position and size: the offset_* and anchor_* properties are not in its storage set. The
  layout inputs it does carry are the storage properties custom_minimum_size,
  size_flags_horizontal and size_flags_vertical; the rest is the parent Container's own
  layout".

Both branches read off a RUNNING game (Godot 4.6.3, `VBoxContainer > Control` and
`Control > Control` with `scale` / `pivot_offset`):

```
$ gda game get /root/Main/HUD/Box --property size --json          # container branch
... "message":"... read them with `gda game rect /root/Main/HUD/Box` ... This Control is a
direct child of a Container ... custom_minimum_size, size_flags_horizontal and
size_flags_vertical; the rest is the parent Container's own layout"
$ gda game get /root/Main/HUD/Box --property custom_minimum_size --json
{"path":"/root/Main/HUD/Box",...,"properties":[{"name":"custom_minimum_size","type":"Vector2","value":[80.0,20.0]}]}
$ gda game get /root/Main/HUD/Box --property size_flags_horizontal --json
{... "properties":[{"name":"size_flags_horizontal","type":"int","value":1}]}
$ gda game get /root/Main/HUD/Box --property offset_left --json   # the old dead end
{"error":{"code":"live_unknown_property","message":"node /root/Main/HUD/Box has no runtime,
storage, or script property: offset_left"}}

$ gda game get /root/Main/Frame/Scaled --property position --json # free branch, unchanged
... "The layout inputs are the storage properties offset_left, offset_top, offset_right,
offset_bottom and anchor_left, anchor_top, anchor_right, anchor_bottom"
$ gda game get /root/Main/Frame/Scaled --property offset_left --json
{... "properties":[{"name":"offset_left","type":"float","value":30.0}]}
$ gda game get /root/Main/Frame/Scaled --property anchor_right --json
{... "properties":[{"name":"anchor_right","type":"float","value":0.0}]}
```

`test_daemon_game_get_names_game_rect_for_a_controls_layout_reads` now covers both branches
and, on each, asserts that FOLLOWING the redirect succeeds — the assertion the old test could
not make, because it pinned the false claim. The unit guard
`test_game_get_names_game_rect_for_the_control_reads_it_cannot_serve` additionally requires the
predicate and both name sets in the harness source.

Five surfaces were reconciled to the same statement: `gda_harness.gd`, `game.py`'s `game_get`
and `game_rect` docstrings (so `--help` and `gda schema` agree), `SKILL.md` and
`docs/command-catalog.md`. `README.md` and the three translations carry no claim about the
inputs, so they are untouched.

### P2-2 — the project-code point

Reproduced the measurement: a plain `Control` with a `_get_minimum_size` that increments a
counter reads 4, stays 4 across two idle `game get` reads, and reads 5 after one `game rect`.
A `Button` subclass with the same script stays at 0, because `Button::get_minimum_size()`
overrides the C++ getter and never reaches the virtual — so the clause is qualified with
"where a class leaves that getter to `Control`" rather than claiming every `Control`.
`CONTEXT.md`'s term gains one clause in its own "ONE narrow point" style, states that
`get_combined_minimum_size()` reads a cache instead, and joins the closing list of points that
widen the surface without a new trust axis. `docs/command-catalog.md`'s "remains a read-only
query" is replaced.

### P3-1 and P3-2 — the test gap

`MIN_SIZE_MAIN_TSCN` gains `Frame > Scaled` (a free `Control`, `scale = Vector2(2, 3)`,
`pivot_offset = Vector2(10, 10)`, offsets 30/40/130/90) and `HUD > Tag` (a `Button` with
`custom_minimum_size = Vector2(300, 1)`). Measured on the engine: `Scaled` reports
`local_position [20, 20]` against a `position` property of `(30, 40)` and `local_size
[200, 150]` against a `size` property of `(100, 50)`; `Tag` reports `minimum_size [8, 8]`,
`custom_minimum_size [300, 1]` and `combined_minimum_size [300, 8]` — a third value. The
`Button` theme number is read from the result, not pinned.

`local_position`'s published description no longer says it "is the node's own `position`"; it
says `get_rect().position` is the origin of the node's own transform, equal to `position` only
while `scale` and `rotation` are default. `local_size` gains the matching clause. The harness
comment, `docs/command-catalog.md` and the PR body now say the local and global rects differ by
the ancestors' TRANSFORM, not their offsets.

### Mutants re-run

The harness hash-pin test is deselected when judging a harness mutant — any harness edit fails
it, so it cannot discriminate.

| Mutant in `gda_harness.gd` | Killed by |
| -------------------------- | --------- |
| `"local_position": _jsonify(control.position)` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| `"local_size": _jsonify(control.size)` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| `"combined_minimum_size": _jsonify(control.custom_minimum_size)` | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| `"minimum_size": _jsonify(control.get_combined_minimum_size())` (new) | e2e `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect` |
| the container branch never fires (`if control == null:`) (new) | e2e `test_daemon_game_get_names_game_rect_for_a_controls_layout_reads`, and the fast tier's `test_game_get_names_game_rect_for_the_control_reads_it_cannot_serve` |

The first four still survive the whole fast tier (2632 passed with the pin deselected), which
is why the e2e assertions are where they had to land.

### Gates

| Gate | Result |
| ---- | ------ |
| `pytest -m "not e2e"` | 2633 passed, 686 deselected |
| `ruff check .` | clean |
| `ruff format --check .` | 515 files already formatted |
| `pytest tests/daemon tests/harness tests/live -m e2e` under the machine-wide Godot lock | 102 passed, 0 skipped, 6m12s |
| `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | "harness version transition is valid" |

Harness bytes changed, so `PINNED_HARNESS_SHA256` moves to
`6c71a0ad8b3f3d29ba26106a58fb0d2023dce0386adae17e05d9afd5d31b82ab`; `HARNESS_VERSION` stays
23.

### Surface diff

Against this PR's previous head `96e9e093f`: `gda schema` changes in exactly three places —
`game get`'s `description`, `game rect`'s `description`, and the `local_position` /
`local_size` descriptions in `game rect`'s `output.properties`. No command added or removed
(77 both sides), no `output.required` change, no result model changed. Two of 93 `--help`
surfaces differ (`gda game get`, `gda game rect`).

Against the base `b7c485693` (dev): three `--help` surfaces differ — `gda game get`, `gda game
rect` and `gda game` (the group listing carries `rect`'s summary line, which is P3-3) — plus
`gda schema`, where `game rect` gains the four properties and their four `required` entries.

### Out of scope

`_control_position_unavailable_message` — `game set --property position`'s container refusal,
which points the caller at `offset_left` … on the same container child — is untouched, as
adjudicated. It is a different command and a different message, and gets its own issue.

Issue #852 carries a dated note recording that its body's premise ("the offsets and anchors as
the storage properties") does not hold for a container-managed `Control`; its acceptance
criteria are unchanged.

qinhaihong-red:
## Review round 2 — CONVERGED. APPROVE, 0 P1 / 0 P2 / 4 P3 (all adopted; closing fix, then orchestrator verification)

Independent adversarial review at `1f52d7e06` against the base `b7c485693`, on a running Godot
4.6.3. **All five round-1 findings are genuinely fixed and none could be reopened.** The P2-1
remedy is correct *by construction*: the engine's stripping branch is
`Object::cast_to<Container>(parent_control)`, not editor-gated, and `data.parent_control` is set
from `get_parent()` on `NOTIFICATION_PARENTED` — so `_has_container_parent` computes exactly the
engine's own condition and cannot drift from it. Measured on eight shapes — a `Control` in a
`VBoxContainer`, an `HBoxContainer` in a `VBoxContainer`, a `Control` in a nested `HBoxContainer`, a
`top_level` `Control` in a container, a `Control` in a `Control`, a `Control` under the root
`Window`, a `Control` in a `SubViewport`, and a `Control` REPARENTED free → container at runtime —
every property the taken branch names is readable, present in the node's unfiltered storage list,
and **writable** (all three container inputs take a `game set` and change the layout). No fix broke
an adjacent case; the mechanism did not grow (one branch on an existing predicate, one string).
No `/backtrace-review` is warranted. Two review rounds are the budget and they are spent: the four
items below land as one closing commit and the orchestrator verifies the head directly.

### Findings and dispositions

**P3-1 — "runs once per request" is measurably false; it is twice where the minimum-size cache is
stale. ADOPTED (wording, five surfaces).** The precision the implementer added beyond the brief is
half right: the cache claim and the `Button` qualifier hold (`get_combined_minimum_size()` consults
`data.minimum_size_valid` first; `Button::get_minimum_size()` overrides the C++ getter, and 29
`scene/gui` classes do the same — a `Button` subclass's counting script stayed at 0 across two
`game rect` calls). The count does not: `update_minimum_size()` returns before its deferred
revalidation when the Control is not visible in the tree, so on a hidden Control both of `game
rect`'s reads reach the script virtual. Measured with plain gda commands: counter 4 → a
`custom_minimum_size` write leaves it 4 → ONE `game rect` → 6 → a second → 7. `CONTEXT.md`
contradicts itself inside one paragraph ("runs once per request" beside "recomputes through the
SAME virtual"). Fix the count in `CONTEXT.md`, the catalog, the `game rect` docstring (published in
`--help` and `gda schema`), `SKILL.md` and the `_handle_game_rect` comment; keep "adds no point of
its own", which is right — same virtual, same point on the surface.

**P3-2 — the redirect tests assert presence and never absence; a mutant that REOPENS round-1's
P2-1 survives both tiers. ADOPTED (test).** Making the container branch APPEND its sentence
instead of replacing the free-Control one passes the fast tier (2632, pin deselected) and both
e2e tests, because every assertion is `X in error["message"]`. Under it a container child's
refusal names the container inputs AND the `offset_*` / `anchor_*` list that strands the caller —
the round-1 dead end, undetected. One negative assertion in
`test_daemon_game_get_names_game_rect_for_a_controls_layout_reads` (`"offset_left" not in
error["message"]` for `/root/Main/HUD/Box`), mirrored in the unit guard.

**P3-3 — `local_position`'s "only while `scale` and `rotation` are default" over-restricts.
ADOPTED (wording, the published Field description).** Measured: a free Control with
`scale (2,3)` and the default `pivot_offset` reports `local_position [30,40]` == its own
`position`. The engine's `get_rect()` origin is `_get_internal_transform().origin + position`,
and that internal origin is `pivot − R·S·pivot`, zero at the default pivot. The harness comment
and the catalog already say it correctly ("moved by `pivot_offset` where a `scale` or a `rotation`
is set"); only the Field description carries the stricter form. Reword to: equals the node's
`position` while `pivot_offset` is zero, or while `scale` and `rotation` are default; a
`pivot_offset` with a `scale` moves the origin away from it.

**P3-4 — the PR body says the out-of-scope item is "filed separately"; no such issue exists.
ADOPTED as a PR-body correction; the filing itself is the user's call.** The round-1 record said
`_control_position_unavailable_message` "gets its own issue" — future tense — and the body reads
as done. `gh issue list --state all` finds nothing. Correct the body to "to be filed". The
reviewer also scoped the follow-up honestly: the same wrong advice ("use those offset properties
explicitly instead") lives in **two documents this PR edits** — `SKILL.md`'s container paragraph
and the catalog's `node set` row — and was confirmed live (`game set …/Box --property position` →
"address offset_left … instead", which fails on that node). The follow-up issue therefore covers
the harness message AND those two sentences. It is pre-existing, it touches the harness (a version
bump that would disturb W3's ladder), and issue creation goes through the user — so it is NOT
folded into this slice and NOT filed unilaterally; it is put to the user at the W3 authorization
point.

- **Round-1 P2-2's `CONTEXT.md` clause, word by word against `4.6.3-stable` source and a live run:**
  `Control::get_minimum_size()` is `GDVIRTUAL_CALL(_get_minimum_size)` with no cache; the combined
  read checks the cache flag first; a stale cache recomputes through the same virtual; the `Button`
  qualifier is necessary and correct. Only the word "once" is wrong (P3-1).
- **The five claimed mutants, exact killers confirmed** with the pin deselected: the four value
  mutants survive the fast tier (2632) and die at e2e on the new `Scaled` node in
  `test_daemon_game_rect_reports_the_minimum_sizes_and_the_local_rect`; the container-branch
  mutant dies at both tiers.
- **No adjacent breakage:** the free-Control sentence byte-identical to round 1 (string
  reconstruction of both revisions plus the live message); the human render's head line
  byte-identical to the base; `MIN_SIZE_MAIN_TSCN` has exactly two users, both updated, and the
  added nodes leave the pre-existing assertions as the same relations.
- **Gates re-run on the FINAL source** (covering the two post-run comment edits): fast 2633; ruff
  clean; 515 formatted; pyright 0/0/0; `tests/daemon tests/harness tests/live -m e2e` **102
  passed, 0 skipped** in 3 m 44 s under the machine-wide lock; the harness guard valid 22 → 23.
- **Harness and merge hazards:** `sha256(gda_harness.gd)` recomputed = the pin; `git merge-tree`
  against #939 **clean**; against #973 exactly the expected set and no more (`install.py`, the
  pins, the three translation stamps) — `gda_harness.gd`, `commands/game.py`,
  `test_game_commands.py`, `tests/support.py`, `SKILL.md`, the catalog and README all auto-merge,
  and **the merged harness has 100 functions with no duplicate names** (the silent keep-both
  GDScript hazard, checked). The W3 order #967-before-#973 is load-bearing and holds.
- **The PR body's surface numbers are exact:** 77 commands on all three revisions; `gda schema`
  differs from the base only in `game get.description`, `game rect.description` and `game
  rect.output` (its `required` gaining exactly the four names); 3 helps differ from the base
  (`game`, `game get`, `game rect`), 2 from round 1; within `output.properties` only
  `local_position` and `local_size` moved, as claimed. The stamp script is a no-op.
- The `custom_minimum_size`-absent case asked for in the brief was ruled out at the source: every
  `_validate_property` override in `scene/gui/*.cpp` at 4.6.3 was read, and none strips
  `custom_minimum_size` or `size_flags_*` at runtime; measured present on all seven shapes.
- Issue #852's dated 2026-09-12 note matches the independent measurement, transcript included.

### Not verified, and why

The red-then-green history (one squashed commit); non-macOS behaviour (nothing platform-conditional
in the diff); a stale v22 harness against a v23 CLI (out of scope by ADR-0018's current-harness
policy); MCP regeneration (derives from the diffed schema); a PROJECT script stripping
`custom_minimum_size` via its own `_validate_property` — ruled out for every built-in class, not
constructed for a project-authored one, and inside the Trusted-project assumption.

qinhaihong-red:
## Round 2 handled — 4 of 4 P3 adopted, one closing commit `477d530f4`

| Finding | Resolution |
| ------- | ---------- |
| P3-1 the count | Fixed on all five surfaces, same sentence everywhere |
| P3-2 the test gap | Mutant reproduced, then killed at both tiers |
| P3-3 `local_position` | Field description reworded; `local_size` checked and left alone, with evidence |
| P3-4 the PR body | "filed separately" -> "to be filed (the user's call)", scoped to three places |

### P3-1 — the count (wording, five surfaces)

`CONTEXT.md`, `docs/command-catalog.md`, the `game rect` docstring (so `--help` and `gda
schema` follow), `src/gda/skill/SKILL.md` and `_handle_game_rect`'s comment now all read
"runs once per request, and twice where the combined read finds the minimum-size cache
stale". "Adds no point of its own" is kept — same virtual, same point.
`grep -rn "once per request" src docs CONTEXT.md` shows only the corrected form (the harness
comment wraps the phrase across two lines and is checked separately).

Reproduced independently before writing it, with a counting `_get_minimum_size` on two nodes
of one script in a running game:

| | visible `Control` | `visible = false` |
| - | - | - |
| each `game rect` | +1 | +1 |
| a `custom_minimum_size` write | +1 | **+0** |
| the next `game rect` | +1 | **+2** |
| the one after it | +1 | +1 |

Same shape as the review's 4 -> 6 -> 7 (the absolute start differs because the scene does),
and it isolates the cause: `update_minimum_size()` invalidates the cache and then returns at
`!is_visible_in_tree()` BEFORE queueing the deferred revalidation, so on a hidden Control
nothing refills the cache and the next request pays for both reads. On a visible Control the
deferred pass refills it between requests, so there it is always once. Idle `game get` reads
never move the counter.

### P3-2 — the test gap (test)

Reproduced the append mutant first (`inputs +=` in `_control_layout_read_message`) and
confirmed it SURVIVES: the fast-tier guard passed in 0.18 s and both e2e tests passed in
6.96 s. Under it the live refusal for `/root/Main/HUD/Box` read
"... anchor_right, anchor_bottom This Control is a direct child of a Container ...", naming
the container inputs AND the `offset_*` / `anchor_*` list — round 1's dead end.

Two assertions were then added with the mutant still applied, and each failed on it:

- `tests/daemon/test_e2e_daemon.py::test_daemon_game_get_names_game_rect_for_a_controls_layout_reads`
  now carries an `absent` set per path (the whole `offset_*` / `anchor_*` list for
  `/root/Main/HUD/Box`, empty for the free Control):
  `assert 'offset_left' not in "node /root/... s own layout"` at line 539.
- `tests/live/test_live_contract_guards.py::test_game_get_names_game_rect_for_the_control_reads_it_cannot_serve`
  reads the assignment itself, because that guard inspects harness SOURCE and both branches'
  strings live in one function, so absence is not expressible there:
  `assert (<re.Match ... match='\t\tinputs += '> is not None and '+' == '')` at line 1037.

The mutant was reverted and the harness is byte-identical to `1f52d7e06` apart from the P3-1
comment.

### P3-3 — `local_position` (wording, the published Field description)

Reworded to "equals the node's `position` property while `pivot_offset` is zero, or while
`scale` and `rotation` are default: a `pivot_offset` with a `scale` (or a rotation) moves the
origin away from it", which is what the harness comment and the catalog already said.

**`local_size` needs no change, and that is measured, not inferred.** At `4.6.3-stable`,
`Control::get_rect()` is `Rect2(xform.get_origin(), xform.get_scale() * get_size())`, and
`Transform2D::get_scale()` returns the basis column magnitudes, which a rotation does not
change. Four shapes on a running 4.6.3, all with `position (30, 40)` and `size (100, 50)`:

| node | authored | `local_position` | `local_size` |
| ---- | -------- | ---------------- | ------------ |
| `ScaleOnly` | `scale (2, 3)` | `[30, 40]` = `position` | `[200, 150]` |
| `RotOnly` | `rotation` 45 deg | `[30, 40]` = `position` | `[100, 50]` |
| `RotScale` | `rotation` 45 deg + `scale (2, 3)` | `[30, 40]` = `position` | `[200, 150]` |
| `PivotRot` | `pivot (10, 10)` + `rotation` 45 deg | `[40, 35.86]` != `position` | `[100, 50]` |

`ScaleOnly` is the finding: a scale alone leaves the origin at `position`. `RotOnly` shows a
rotation alone does too, and `PivotRot` shows a pivot with a rotation and NO scale moves it —
which is why the reword says "(or a rotation)". `RotScale` is the `local_size` check: `size *
scale` holds under a rotation.

### P3-4 — the PR body

The Open-questions bullet now says **to be filed (the user's call)** and no longer implies an
issue exists (`gh issue list --state all` finds none). It also scopes the follow-up honestly:
it covers `_control_position_unavailable_message` AND the same wrong advice in two documents
this PR edits — `SKILL.md`'s container paragraph ("use those offset properties explicitly
instead") and `docs/command-catalog.md`'s `node set` row. None of the three is touched here:
pre-existing, a different command, and a harness edit would move the version ladder W3 depends
on. Issue creation is the user's decision.

Also corrected in the body while it was open: the P3-1 measurement (with the table above), the
`local_position` row, the engine-fact bullet's origin formula (now `position + pivot - R*S *
pivot`, with the rotation-invariance of the size stated), a new mutant-table row for the append
mutant, and the e2e wall clock of the final run.

### Gates on `477d530f4`

| Gate | Result |
| ---- | ------ |
| `pytest tests -m "not e2e"` | 2633 passed, 686 deselected, 48 s |
| `ruff check .` | clean |
| `ruff format --check .` | 515 files already formatted |
| `pytest tests/daemon tests/harness tests/live -m e2e` under the machine-wide Godot lock | 102 passed, 0 skipped, 3 m 21 s |
| `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | "harness version transition is valid" |
| PR CI | green |

Harness bytes moved (the comment only), so `PINNED_HARNESS_SHA256` is now
`67a610c6842360fe10743adb3724a79cb8a319a877ec91e193f41d8d0102ebc5`; `HARNESS_VERSION` stays
**23**.

`gda schema` against `1f52d7e06`: exactly two leaf values differ — `game rect`'s `description`
and `game rect.output.properties.local_position.description`. 77 commands on both sides,
`output.required` unchanged, `local_size` unchanged. Nothing structural.

qinhaihong-red:
## Round 2 handled at `477d530f4` — verified directly by the orchestrator. **READY.**

Two review rounds were the budget and both are spent; this is the closing verification of the
implementer's round-2 commit.

### Verified

- **The reopening mutant dies at both tiers.** Making the container branch APPEND its sentence
  instead of replacing the free-Control one — the mutant round 2 showed survived everything —
  now reds exactly `test_game_get_names_game_rect_for_the_control_reads_it_cannot_serve` on the
  fast tier (pin test deselected) and exactly
  `test_daemon_game_get_names_game_rect_for_a_controls_layout_reads` on a real engine under the
  machine-wide lock. Reverted clean.
- **Gates on the head:** fast tier **2633 passed**; ruff clean; 515 files formatted; pyright 0 / 0
  / 0; `tests/daemon tests/harness tests/live -m e2e` **102 passed, 0 skipped** in 197 s under
  the lock — run after the last edit, in one process.
- **Schema, leaf by leaf:** `1f52d7e06 → 477d530f4` changes exactly two leaves, both under `game
  rect` — its description and `local_position`'s description; `local_size` untouched, nothing
  structural. `dev → 477d530f4` is 20 added / 0 removed / 3 changed and touches only `game get`
  and `game rect`.
- **All four surfaces carry the corrected count** — `game.py`'s docstring, `SKILL.md`, the catalog
  and `CONTEXT.md` each continue "runs once per request, and twice where the combined read finds
  the minimum-size cache stale"; the fifth (the harness comment) wraps the same clause. The old
  terminating form exists nowhere.
- **Harness:** `HARNESS_VERSION` 23; `sha256(gda_harness.gd)` recomputed =
  `67a610c6842360fe10743adb3724a79cb8a319a877ec91e193f41d8d0102ebc5` = the pin; the version guard
  valid against dev. The closing diff is eight files, 45 insertions / 17 deletions — prose, two
  test files, the pin, and a harness COMMENT.
- **Commits:** three on top of dev, each authored and committed by `haihong.qin
  <haihongqin@gmail.com>`, no attribution of any kind, no force-push. PR CI green.
- **Accepted as judged differently, with evidence:** `local_size` left unchanged — at
  `4.6.3-stable` `get_rect()` is `Rect2(xform.get_origin(), xform.get_scale() * get_size())` and
  `get_scale()` is rotation-invariant; four live shapes confirm it (a 45° rotation with scale still
  reads `local_size [200, 150]`). Its description was already literal.

### Follow-up recorded, not folded in

`_control_position_unavailable_message` (the `game set --property position` container refusal)
and the same advice in `SKILL.md`'s container paragraph and the catalog's `node set` row: to be
filed at the user's decision, after W3's harness ladder. Not this slice's scope.

W3 merge order unchanged: `#939 → #937 → #938 → #940 → #964 → #966 → #967 → #973`; #967 (harness 23)
lands before #973 (24), which the version guard makes load-bearing.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `477d530f4042ada7d1e2f059aa96b47b116a271f`. Existing reviews and replies were not used as evidence.

1. Issue #852's body/acceptance text still requires offsets and anchors for every Control, while its 2026-09-12 correction proves that those properties are absent on a direct `Container` child. The correction also says the acceptance criteria are unchanged. Reconcile the primary body and acceptance text so one current requirement remains.
2. Current shipped guidance still gives the disproved remedy: `docs/command-catalog.md:373-380`, `src/gda/skill/SKILL.md:442-446`, README.md, and all three translations tell a Container child to use offsets explicitly. Following that advice reaches the second `live_unknown_property` that the correction exists to prevent.

**Terminology and prose assessment:** I inspected the complete issue/correction, PR body, changed README/translations, command catalog, skill, help/schema prose, and runtime messages against `CONTEXT.md`, ADR-0017/0020/0023/0035, Godot Control layout terms, and ASD-STE100. The new `local_position`, `local_size`, `minimum_size`, and `combined_minimum_size` descriptions are coherent. The two contradictions above are substantive usability findings. I did not validate every translated paragraph with a native-language reviewer.

The harness correctly owns runtime layout observation, and the new `game get` branch derives guidance from the actual parent type. But the Container remediation rule has overlapping owners: `_control_layout_read_message()` now gives valid inputs, while the live and headless position-set refusal helpers retain the opposite rule (`src/gda/harness/gda_harness.gd:1015-1016`; `src/gda/ops/operations.gd:1923-1924`). Align the existing helpers and add a behavioral synchronization test; no new layer is needed. Claim-evidence state: `open` until following each suggested property succeeds or the parent-owned layout boundary is stated.

## Standards

- Hard: the setter messages and docs direct Container children to properties that the new real-engine test proves are absent. The retained E2E checks only that the false offset names appear. Replace the Container-case remedy with `custom_minimum_size`, `size_flags_horizontal`, `size_flags_vertical`, and the parent Container's layout; update the docs/translations and make the regression follow the advice. This violates `RULES.md` consistency and affected-artifact synchronization requirements.

No independent baseline smell was substantiated.

Standards summary: 1 hard finding; worst is unusable public remediation.

## Spec

- Required: the 2026-09-12 correction says not to restore the false offset/anchor claim. The new `game get` branch is correct, but `src/gda/skill/SKILL.md:442-447`, `docs/command-catalog.md:373-380`, both setter helpers, and their old test still recreate the same dead end. Update the entire user path, not only the getter.

The six new rect fields/read redirects otherwise match the corrected requirement. Independent validation passed 169 focused unit tests and 6 relevant live E2E tests; the old false-remedy E2E also passed, confirming that the defect is codified.

Spec summary: 1 finding; worst is a directly unusable prescribed next step.

qinhaihong-red:
The out-of-scope item this PR's reviews surfaced — the `game set --property position` container
refusal naming `offset_*` on a node that does not carry them, and the same advice in the skill's
container paragraph and the catalog's `node set` row — is now filed as #974 (milestone
"gda dogfooding II", after #847 in the harness serial chain). The headless `node set` twin is in
its scope as a measure-first criterion. This PR's body reads "to be filed"; #974 is that filing.

qinhaihong-red:
Third review handled at exact head `c04f178d1` — one commit over `477d530f4`, no force-push. The findings are one underlying claim: the Container remediation rule had three code owners (the `game get` redirect and the two position-set refusals) and five prose copies, and round 2 corrected only the getter. Every finding adopted; the fix is one statement of the rule, used everywhere it is spoken.

| Issue #852's body and acceptance text still require offsets and anchors for every Control while its 2026-09-12 correction says a direct `Container` child carries neither | Fully Adopt | the two texts contradicted each other; the correction is the measured one (`Control::_validate_property` strips storage from `offset_*` / `anchor_*` under a Container parent, Godot 4.6.3) | the body's requirement and acceptance text now state the container branch — `custom_minimum_size`, `size_flags_horizontal`, `size_flags_vertical`, and the parent Container's own layout; the dated correction stays as the record | issue body re-read after the edit |
| Shipped guidance still gives the disproved remedy: catalog `node set` row, SKILL container paragraph, README row and the three translations | Fully Adopt | following it reaches the second `live_unknown_property` (measured live in round 2) | all five rewritten to the container inputs; the three translated rows were translated here and re-stamped by `scripts/update_readme_i18n.py` | fast tier, including the README i18n sync gate; CI green |
| Overlapping owners: `_control_layout_read_message()` gives valid inputs while both position-set refusal helpers (`gda_harness.gd`, `operations.gd`) keep the opposite rule | Fully Adopt | three code sites decided the same fact independently | one `_control_layout_inputs(control)` per payload — the existing container predicate picks the inputs — now used by the getter redirect AND by `_control_position_unavailable_message`; the two payloads are mirrored and the mirror list names the helper. In `operations.gd` the refusal is built BEFORE `root.free()` (the first draft read a freed node — caught by the headless e2e, not by inspection) | the unit guard reads the shared helper's container branch and asserts it names no offset; harness re-pinned at `HARNESS_VERSION` 23, sha `7c78691056bc…`, from the merged file |
| The retained e2e only checks that the false offset names appear, so the defect was codified | Fully Adopt | a test that asserts the wrong advice is present protects the defect | replaced by two tests that FOLLOW the advice: live `test_daemon_game_set_container_managed_control_position_names_the_inputs_it_carries` and headless `test_node_set_container_managed_control_position_names_the_inputs_it_carries` set `custom_minimum_size` on the container child and succeed, and assert the refusal names no `offset_*` | both green against Godot 4.6.3 under the machine-wide lock |
| Claim-evidence state `open` until following each suggested property succeeds or the parent-owned layout boundary is stated | Fully Adopt | — | the two tests above follow each named input on the real engine; the parent-owned boundary is stated in the shared helper's message and in the docs | closed by the e2e pair |

Consequence for #974: it was filed from this PR's earlier rounds for exactly this setter advice; this head ships that fix, so #974 is commented as subsumed and the wave's promotion PR closes it.

CI is green on `c04f178d1`. Locally: fast tier, `ruff check` / `ruff format --check` / `pyright` clean, live + headless e2e pair under the lock. Not verified: the three translated rows had no native-language review.

PR #973, opened 2026-09-12
feat(gda): perf monitors --summary omits the rows, and a window reports what it retained (#846)

### LINKED ISSUE #846: perf monitors --frames: aggregate-only mode and bounded observer memory

qinhaihong-red:
## Summary

`gda perf monitors --frames 600 --budget ... --json` returned every raw sample — a 68,215-token
result the dogfooding client truncated — and, over an otherwise stable window, `static_memory`
rose 1,229,440 bytes while object, node, resource and orphan counts stayed flat (GDA-DF-066,
kung-fu dogfooding). The harness retaining 600 sample Dictionaries was the most plausible
source, so the observer read as a monotonic leak in the very window it measured, and the
production moved its no-growth gate to a separate in-process test.

This PR bounds the observer, discloses what it still costs, and lets the caller ask for the
window without its rows.

## Public behavior

**`--summary` on `perf monitors --frames`** — the spelling and meaning `screen frames
--summary` has. The window is still sampled in full; only the per-frame rows are left out of
the RESULT:

| Field | Default | `--summary` |
| ----- | ------- | ----------- |
| `stats`, `budget`, `passed`, `frames`, `max_frames` | the full window's | identical |
| `samples` | the per-frame rows | `null` |
| `samples_omitted` | `false` | `true` |
| `collector_bytes` | what the sampler retained | identical |

`samples_omitted` is the ONLY way `samples` may be absent from a window: the result validator
and the published schema both refuse a window that drops the rows without saying so, and one
that claims the omission while carrying them. `--summary` requires `--frames`, like `--monitor`
and `--budget` (a snapshot has no rows to omit) — refused by name, and published as a schema
conditional so a client validating against `gda schema` reaches the same verdict.

**`collector_bytes`** — the LOGICAL size of what the harness's sampler kept for this window:
8 bytes per stored value, over one packed column per sampled monitor plus one of timestamps.
That makes it exact and testable, and a **lower bound** on what the window costs inside the
game: a packed column over-allocates as it grows, and the shared window base's own per-frame
accumulator is outside the number. So a window's own `static_memory` rise is read against it by
ORDER OF MAGNITUDE — a few times `collector_bytes` is still the observer, an order of magnitude
past it is the game. `null` in snapshot mode, which retains nothing. It is the HARNESS's
number, relayed — not one the CLI recomputes from the request — so a harness that retained
something else reports what it really held.

**The gap, measured** (round-1 review, re-measured independently before publication): a
600-frame window over all 16 monitors reports `collector_bytes` 81,600 against a `static_memory`
rise of **113,872 bytes** — about **1.4x** — on macOS with Godot 4.6.3, reproduced byte for byte
on three consecutive windows. Roughly 17,128 bytes of that is the shared base's per-frame
accumulator; the rest is over-allocation on the columns already counted (a 600-element packed
column measures 5,736 bytes against a logical 4,800). That ratio is published as a MEASUREMENT
with its host, engine, frame count and monitor count — not as a constant gda promises. Only the
direction generalizes: the number understates wherever a growing packed array over-allocates.

**Human rendering.** The window header gains the same two facts:
`perf window: 600 frames, 16 monitors (ceiling 600, collector >=81600 bytes, samples omitted)`.
The `>=` is round 2's correction: `~` read as *approximately*, the word every other surface
dropped, while the number is exact (600 x 17 x 8) and the measured in-game cost was 1.40x it.
A human reading only the render, then a 113,872-byte `static_memory` rise, reached exactly the
false "game leak" this PR exists to remove. The render is human-only, so no published
description moves with it.

**Unchanged:** snapshot mode, the default window (rows kept), the statistics and percentile
semantics, the budget contract, and the five other multi-frame ops.

## Wire shape

The `perf-sample` sampler now keeps the window COLUMN-wise in packed arrays — one
`PackedFloat64Array` per monitor plus one `PackedInt64Array` of timestamps — instead of one
Dictionary per frame. A stored value costs 8 bytes instead of a hash node and two boxed
Variants. The reply is therefore column-wise too, with the frame index positional (entry *i*
is frame *i*) rather than transmitted:

```
{"kind": "sample", "frames": N, "monitors": [...],
 "timestamps": [...N...], "values": {"<monitor>": [...N...]}, "collector_bytes": B}
```

Per ADR-0023, that is the WIRE shape and it is not the public result: the private
`_SampleReply` adapter decodes and self-checks it (the declared length matches the timestamp
column, the value columns are exactly the declared monitors, every column is one value per
frame, and `collector_bytes` is present), and the recipe assembles the public
`PerfMonitorsResult` from it. The columns are the STORAGE; the per-frame rows are a projection
rebuilt from them only when the caller wants them, so one aggregation serves both forms and no
packed column reaches the published result. Both new public fields are plain values, so the
result keeps value equality across serialize/revalidate.

The reply shape this replaces is **not decoded anywhere**. Live commands target the harness
bundled with the running gda (ADR-0018's 2026-09-08 current-harness note), so a dictionary-shaped
reply is contract drift and classifies as `contract_violation` — no version negotiation, no
fallback route.

The change is local to `_handle_perf_sample`: the shared multi-frame base (`_begin_window` /
`_advance_window`) is untouched, so `perf monitor`, `perf sample`, `screen capture`,
`screen frames` and `input sequence` keep their behaviour and their tests.

## Harness version and pin

`HARNESS_VERSION` 22 -> **24**. 23 belongs to #852 (PR #967), the first of W3's two harness
slices, which lands before this one; the merge base is 22, so the version guard passes today
and stays valid after #967 lands. `PINNED_HARNESS_SHA256` re-pinned to
`3ba2ba9fd142d5b43fe62ceff2790aaf9b4efd0d652648518d47bb3e9c4eab04` (round 2 named the
measurement conditions in the harness comment, which moved the bytes again; the version is
unchanged, and the guard requires only `head > base`).

## Surface diff (enumerated)

`gda schema --json` captured from a separate detached checkout of the base (`b7c485693`) and
from this branch, compared leaf by leaf: **42 added, 1 removed, 8 changed — all 51 under
`commands[35]` (`perf monitors`)**. 77 commands on both sides; no command added or removed, no
other command's leaf touched.

1. `input.properties.summary` — new boolean, `default: false`, with its description.
2. `input.allOf[2]` — new conditional: `summary: true` requires `frames`.
3. `output.properties.samples_omitted` / `collector_bytes` — new nullable fields
   (`collector_bytes` with `minimum: 0`), with descriptions.
4. `output.oneOf[0]` (snapshot) — both new fields pinned to `null`.
5. `output.oneOf[1]` (window) — both added to `required`; `samples` widened from `array` to
   `["array", "null"]` (the 1 removed leaf); two new conditionals tying `samples` to
   `samples_omitted` in both directions.
6. `argv` gains `summary` between `monitors` and `budget` (9 new leaves; `budget`'s own index
   moves 2 -> 3, its content unchanged).
7. Four descriptions rewritten: the command help, the params docstring, the result docstring,
   and `samples`'s field description.

**Round 1's follow-up round adds no leaf**: re-captured the same way against `21c7bfc87`, it is
**0 added / 0 removed / 3 changed**, all descriptions and all under `commands[35]` — the command
help, the result docstring, and `output.properties.collector_bytes.description`.
**Round 2's closing round moves no leaf at all**: against `d31b38e13`, **0 / 0 / 0**. It touches
the human render, three test files and one harness comment; no published description changes.

Other agent-facing surfaces changed by hand: `src/gda/skill/SKILL.md` (the `perf` row),
`docs/command-catalog.md` (the `perf monitors` entry), `README.md` (the `perf monitors` row)
and the three translations, re-stamped with `scripts/update_readme_i18n.py` (row keys diffed:
96 rows in each file before and after, identical keys).

## Validation

| Gate | Command | Result |
| ---- | ------- | ------ |
| Fast tier | `uv run --frozen pytest tests -m "not e2e" -q -p no:cacheprovider` | **2638 passed**, 685 deselected (base dev 2626; +12) |
| Format | `uv run --frozen ruff format --check .` | 515 files already formatted |
| Types | `uv run --frozen pyright` | 0 errors, 0 warnings, 0 informations |
| Real engine | `pytest tests/live tests/harness tests/daemon -m e2e -rs -q` under the machine-wide Godot lock | **101 passed**, 629 deselected, 3:22 (host load ~1.5) |
| Harness guard | `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | "harness version transition is valid" |

**Measured sizes** (real daemon -> engine session, all 16 monitors, 600 frames, Godot 4.6.3 on
macOS): **about 1.7 KB with `--summary` against about 260 KB without** — about 150x. The last
digits are session-dependent, so the figures are stated to that precision and not as exact
byte counts: three runs gave 1,700 / 1,690 / 1,689 bytes and 260,179 / 260,178 / 259,928.
`collector_bytes` = 81,600 = 600 x 17 x 8, exactly what the packed columns hold, and that one
IS exact. The e2e regression pins a ceiling of 8,192 bytes — several times the measured size:
headroom for a widened monitor table, far below anything that grows with `--frames`. Round 2
removed the last copy of the session-dependent digits, which sat in that test's own comment and
disagreed with two re-measurements; it now states the same figures the catalog does.

## Mutants

Each applied to the committed tree, run against
`tests/live/test_perf_commands.py tests/live/test_live_contract_guards.py tests/meta/test_schema_command.py tests/cli/test_render.py`,
then reverted. No survivors.

| Mutant | Killed by |
| ------ | --------- |
| Aggregate only the first monitor (`reply.monitors[:1]`) | `test_the_packed_aggregation_equals_the_recorded_dictionary_aggregates` (+5 more) |
| Invert the derived flag (`samples_omitted=not params.summary`) | `test_summary_omits_the_rows_and_keeps_every_other_window_field` (+12 more; 13 failures over the four files) |
| Harness stores 4 bytes per value but the constant still says 8 | `test_perf_packed_value_bytes_mirrors_the_harness` |
| Rebuilt rows all take `timestamps[0]` | `test_the_rebuilt_rows_carry_the_recorded_window_frame_by_frame` |
| Drop the `samples_omitted -> samples: null` schema conditional | `test_perf_monitors_schema_and_models_reach_the_same_verdict` |
| CLI recomputes `collector_bytes` from the request instead of relaying the harness's | `test_a_window_reports_what_the_observer_retained` |
| **(round 1)** Harness stores `PackedFloat32Array`, constant left at 8 | `test_perf_packed_value_bytes_mirrors_the_harness` — it survived the whole suite before |
| **(round 1)** Harness narrows the timestamp column to `PackedInt32Array` | `test_perf_packed_value_bytes_mirrors_the_harness` |
| **(round 1)** Harness stores a constant timestamp (`timestamps.append(0)`) | `test_a_600_frame_summary_window_stays_compact_and_reports_its_own_cost` (e2e; `assert 0 > 0`) |
| **(round 1)** Delete `"samples_omitted": {"type": "null"}` from the snapshot schema branch | `test_perf_monitors_schema_and_models_reach_the_same_verdict` |
| **(round 1)** Drop `ge=0` from `PerfMonitorsResult.collector_bytes` | `test_perf_monitors_schema_and_models_reach_the_same_verdict`, at its new one-sided assertion |
| **(round 2)** Revert `collector_bytes`' PUBLISHED description to the pre-fix "approximate bytes" | `test_perf_monitors_schema_and_models_reach_the_same_verdict`, at its new description assertion — the same revert passed the whole fast tier (2638 passed) before it |
| **(round 2)** Revert the human render to `collector ~` | `test_render_perf_monitors_window_states_the_observer_cost_and_rows` |
| **(round 2)** Harness reads the clock in microseconds (`Time.get_ticks_usec()`) | `test_a_600_frame_summary_window_stays_compact_and_reports_its_own_cost` (e2e; `AssertionError: 4133279`, `assert 4133279 < 30000`) — it passed the fast tier (2637, pin deselected) and the whole perf e2e module before |

## PR CI vs the e2e tier

PR CI runs **no Godot e2e** (schedule/dispatch-only), so the cross-language half is guarded at
unit speed by `test_perf_packed_value_bytes_mirrors_the_harness`: the harness's
`PERF_PACKED_VALUE_BYTES` must equal the fixtures' twin, be what the reply multiplies by, AND —
since round 1 — be backed by the storage TYPES that make it true (`PackedFloat64Array` columns,
a `PackedInt64Array` of timestamps). Without those two, a `PackedFloat32Array` harness passed
the fast tier and the real-engine perf e2e alike: `collector_bytes` would report twice the
truth, and the full-binary64 fidelity this command's help and CONTEXT.md's `Value projection`
publish would be destroyed silently (`22001251` reading back as `22001252`).
What still needs a real engine — and was run locally for this PR — is that a packed array is
appended through a lambda capture at all (Godot 4 passes packed arrays by reference, per the
engine's own class reference; if it did not, the columns would arrive empty and the reply would
fail its own self-check), that Godot's JSON writer emits the columns as numbers, and the
measured sizes above.

AC2's equality is proven on a RECORDING, since two live windows are never equal: 20 frames of
two monitors in the old dictionary shape, with the aggregates produced by running the pre-#846
aggregation (`_stats_over` over the dictionary rows at `b7c485693`) and pinned as literals. The
values are chosen to be discriminating, not captured from a session — the means are not sampled
values, p50 is not the mean, and p95 is the second-largest rather than the max.

## Not done, and why

- **No decoding of the previous reply shape**, and no version negotiation or fallback route —
  ADR-0018's current-harness policy. A dictionary-shaped reply is `contract_violation`, pinned
  by `test_the_previous_dictionary_reply_shape_is_contract_drift`.
- **The base accumulator is not counted in `collector_bytes`.** The shared window base
  accumulates one Array entry per frame — a nil, since this sampler returns none — and measures
  **17,128 bytes at 600 frames**. The earlier reason given here was wrong: `finalize` already
  receives `samples`, so counting it would be one line inside `_handle_perf_sample` and would
  need no change to the shared base (#846's prohibition is on MODIFYING that base, not on
  reading what it hands you). The real reason is that it would not close the gap —
  81,600 + 17,128 = 98,728 against a measured 113,872 — because the remainder is
  over-allocation on the columns already counted, and closing that would mean buying an
  engine-internal `sizeof(Variant)` estimate. `collector_bytes` therefore stays the exact,
  testable logical size, published as a lower bound; a true total, if it is ever wanted, belongs
  in a separately measured field. Stated in the harness comment and in the catalog.
- **`perf monitor`** (the per-node property/signal timeline) keeps its per-frame rows and gets
  no `--summary`. Out of #846's scope; it is a different op on the same base.
- **No ADR amendment.** The change adds no decision — ADR-0023 already governs the wire/public
  split and ADR-0018 the current-harness policy.

## Disclosures

- `collector_bytes` is the SAMPLER's columns measured logically, not a measurement of the
  engine's allocator. Every surface now says lower bound rather than "approximate", which read
  as an error bar in either direction and was wrong by 1.4x in one of them.
- The 1.4x ratio and the 17,128-byte accumulator are figures from ONE host and ONE engine build
  (macOS, Godot 4.6.3). They are published with those conditions and are not a promise; the
  direction is what holds generally.
- The measured sizes were taken with host load average ~10. Byte sizes are load-independent; no
  timing figure is claimed except the wall clock of the e2e run itself.
- A harness-side mutant on the sampler's VALUES (for example a constant timestamp, or the
  clock read in the wrong UNIT) is still killed only by the e2e tier, because every unit fake
  is CLI-side — round 1 added the timestamp assertion and round 2 bounded the span
  (`0 < span_ms < 30_000`, unfalsifiable in milliseconds under the daemon's 30 s window guard,
  and ~4,133,000 in microseconds), and both fire in the scheduled run, not on PR CI. What moved
  to unit speed is the storage TYPE and the byte constant.

## Open questions — closed in round 1

- **Harness version ordering.** Settled: the guard requires only `head > base`, never exactly
  +1, so **24 stays**. It passes over dev's 22 today and over 23 once #967 lands, and this PR
  needs no re-pin in either order. The live hazard runs the other way — #967's 23 would become
  `head < base` if this PR landed first — so W3's merge order (#967 before #846) is
  load-bearing and is recorded as such.
- **`argv` index shift.** Settled: no action. `mcp/server.py` builds `argv_by_tool` from the
  entry NAME, a test asserts `argv` never reaches the MCP tool schemas, and both SKILL and
  catalog tell agents to link bindings by `input_property`. No stability promise exists or is
  needed.
- **Should `collector_bytes` count the base's per-frame accumulator?** Settled: no — see the
  bullet above for the reason, which is not the one this PR first gave.

Closes #846


qinhaihong-red:
## Review round 1 — CHANGES REQUIRED, 0 P1 / 2 P2 / 6 P3 (all adopted, plus one added by the orchestrator)

Independent adversarial review at `21c7bfc87` against the base `b7c485693`, on a real engine
(Godot 4.6.3). **The slice's core value is reproduced, not taken on trust:** the same 600-frame,
16-monitor window costs **1 229 440 B of `static_memory` on the base** — byte for byte the figure
GDA-DF-066 reported — and **113 872 B on this branch**, a 10.8× reduction, with the allocation
fully released between windows (identical `first` on three consecutive runs; zero idle drift).

Neither P2 needs a redesign.

### Findings and dispositions

**P2-1 — the published attribution rule understates the observer by 1.40×. ADOPTED (wording).**
Three surfaces (`commands/perf.py:1239-1248`, `docs/command-catalog.md:1743-1748`,
`src/gda/skill/SKILL.md:248`) tell the caller that "a rise of about `collector_bytes` is the
observer, and a rise well past it is the game". Measured on a game that allocates nothing, the
observer's own rise is **1.40×** `collector_bytes` — so the rule points the caller at a false
"game leak", which is the judgement #846 exists to remove. The mechanism was measured in-engine,
not guessed: a 600-element packed column really costs 5 736 B against a published 4 800 (CowData's
geometric growth), and the shared window base's per-frame accumulator adds 17 128 B that
`collector_bytes` does not count — together 114 640 B against a measured 114 032.

Make `collector_bytes` a **lower bound** in all three places, name both parts of the gap, and keep
the guidance qualitative: a few times `collector_bytes` is still the observer, an order of
magnitude past it is the game. **Publish the 1.4× as a MEASUREMENT with its host, engine, frame
and monitor count — not as a constant the product promises.** The reviewer is explicit that the
magnitude is platform-specific while the direction (`collector_bytes` understates) holds wherever
`append` over-allocates; a platform-specific ratio must not become a product promise.

**P2-2 — the guard does not do what its own docstring says; a `PackedFloat32Array` mutant survives
the whole suite. ADOPTED (code).** `test_perf_packed_value_bytes_mirrors_the_harness` states its
purpose as catching a harness that "ever stored something wider than 8 bytes per value", but it
only greps the CONSTANT and the multiplication — never the storage TYPE, and nothing else in the
repo pins it either. Changing the three `PackedFloat64Array` occurrences to `PackedFloat32Array`
while leaving the constant at 8 passes the fast tier (207) **and** the real-engine
`tests/live/test_e2e_perf.py -m e2e` (7). The harm is double: `collector_bytes` becomes 2× the
truth, and the published full-binary64 fidelity is silently destroyed — `22001251` reads back as
`22001252` — contradicting this command's own help paragraph and `CONTEXT.md`'s `Value projection`,
which makes full precision on both engine payloads a published property. Two assertions in the
grep idiom the test already uses (`"= PackedFloat64Array()"`, `"timestamps := PackedInt64Array()"`)
restore the guard at unit speed, which matters because PR CI runs no Godot. The unmutated packed
path was confirmed byte-identical to the old dictionary path, so precision is not a regression of
this PR.

**Related gap, same class, also adopted:** no e2e assertion touches the window's TIMESTAMPS. The
CLI-side twin of that mutant is guarded; the harness-side one is not. One line in the 600-frame
e2e closes it.

**P3-1 ADOPTED.** The accumulator estimate "~10–14 KB at 600 frames" is wrong; measured **17 128 B**.
Fix the figure wherever it is repeated.

**P3-2 ADOPTED.** Deleting `"samples_omitted": {"type": "null"}` from the snapshot output branch
survives `test_perf_monitors_schema_and_models_reach_the_same_verdict` — the published schema would
be wider than the runtime ABI, the exact divergence that test exists to prevent. Its twin
(`collector_bytes`) IS covered. Add `{**snapshot, "samples_omitted": False}` to the output corpus.

**P3-3 ADOPTED.** Dropping `ge=0` from `PerfMonitorsResult.collector_bytes` survives too, although
the PR's own enumerated surface diff lists `minimum: 0` as a deliberate published fact. One corpus
entry fixes it.

**P3-4 ADOPTED.** The catalog states the measured byte figures as exact ("1 667 … 259 569"), but
the stats block's digits are session-dependent — independent re-measurement on the same host and
engine gave **1 679** and **259 396**. Use "about 1.7 KB against about 260 KB (measured …)". This
is not a licence to return to the earlier hand-waving, which was wrong by a different order of
magnitude; it is dropping false precision while keeping the measurement and its conditions.

**P3-5 ADOPTED.** The PR body undercounts M2's killers: the named test "+9 more" is really +12
(13 failures against the PR's own four files).

**P3-6 ADOPTED.** `tests/live/test_e2e_perf.py:333-336` says the same sentence twice.

**P3-7 — added by the orchestrator. ADOPTED (wording).** The PR justifies leaving the base
accumulator uncounted by saying the issue forbids touching that base. The reviewer showed that
reason is **not correct**: `finalize` already receives `samples`, so counting it is one line inside
`_handle_perf_sample` and needs no base change — the prohibition is about MODIFYING the base, not
about reading what it hands you. The conclusion stands and the reason must be replaced by the real
one (below). A right answer resting on a wrong reason is what reopens later.

### The three open questions — decided

**(a) Harness 24: keep it, and the ordering risk belongs to #967, not here.** The guard requires
only `head > base`, never exactly +1, so 24 passes over dev's 22 today and over 23 after #967
lands; this PR needs no re-pin in either order. The real hazard is the reverse: **if #973 landed
first, #967's 23 would become `head < base` and would have to move to 25.** W3's merge order already
puts #967 before #846 — that order is now load-bearing and is recorded as such. The two harness
hunks do not overlap at all (#967 at 589–711, #973 at 85–86 and 996–1058), so only `install.py`'s
one line and the two pins conflict, trivially.

**(b) `budget`'s `argv` index 2→3: no action.** `mcp/server.py` builds `argv_by_tool` from the
entry NAME, never from the `argv` list, and `test_argv_metadata_cannot_reach_the_two_schema_halves_gda_mcp_maps`
asserts `argv` never reaches the MCP tool schemas at all. A repo-wide grep for positional access
finds only `sys.argv`. Both `SKILL.md` and the catalog tell agents to link bindings by
`input_property`. No stability promise exists or is needed.

**(c) `collector_bytes` must NOT count the base accumulator.** Not for the reason the PR gives
(see P3-7) but because counting it would not fix anything: `81 600 + 17 128 = 98 728` against a
measured 113 872, since ~15 KB of the gap is CowData growth on the columns already counted. Closing
the rest would mean buying an engine-internal `sizeof(Variant)` guess — exactly the disproportionate
complexity the standing rule forbids. Keep `collector_bytes` as the exact, testable logical size of
the columns and land P2-1's reading rule instead. If a true total is ever wanted, the honest form is
a separately measured field, not a constant.

Fast tier 2638 (+12 over dev's 2626); ruff clean, 515 files formatted, pyright 0/0/0; `tests/live
tests/harness tests/daemon -m e2e` **101 passed** in 338.7 s under the machine-wide lock; harness
sha256 recomputed and matching the pin; version guard valid 22 → 24. Schema 42 added / 1 removed /
8 changed reproduced leaf by leaf, all under `commands[35]` = `perf monitors`, 77 commands both
sides. **AC2 proven by replaying the BASE's own `_SampleReply` + `_stats_over` at `b7c485693` over
the recorded window — every pinned literal matched.** AC1's `--summary` shape confirmed on a real
session; AC4's window measured at 1 679 B against 259 396 B (154×), well under the pinned 8 192.
`collector_bytes` tracks frames × (monitors+1) × 8 exactly in five different shapes. The published
aggregates recomputed from the 600 published rows equal the result's `stats` for all 16 monitors.
Frames `0..599` in order, timestamps non-decreasing, every row carrying all 16 monitors. **ADR-0023
holds:** decoding is `classify_live(result, None, _SampleReply)` in the recipe over a PRIVATE
adapter — the same pattern as `screen capture`'s `_CaptureReply`, which the ADR names as
precedent — and the public result round-trips with value equality and byte-stable JSON, with no
hidden wire-presence flag. **ADR-0018 holds:** the old dictionary shape is a plain refusal (three
`Field required` errors), with no version, fallback or compat branch anywhere. All six named
mutants killed, four of them by exactly the named test and nothing else. Harness edits confined to
the const block and `_handle_perf_sample`, with `_serialize`, `_json`, `_ok`, `_begin_window`,
`_advance_window` and `operations.gd` untouched and zero overlap with #967's region.
`tests/support.py` is one clean region. README and the three translations keep 96 identical row
keys, the stamp re-runs to the same value, and each translation follows its own file's register.
Further probes all sane: 1-frame window, `--monitor` × `--summary`, `--budget` with and without
`--summary` (identical verdicts), the snapshot reporting both new fields null, `--summary` without
`--frames` refused by name at exit 2, duplicate `--monitor` names deduped before dispatch, and
inf/NaN through the packed path identical to the dictionary path.

### Not verified, and why

The PR's exact byte figures are not reproducible (session-dependent digits — see P3-4). The Float32
survivor was run against the PR's four mutation files and the full `test_e2e_perf.py`, not every
other e2e module. CI's "no Godot e2e" is taken from the documented policy rather than re-run. The
allocator figures are one host and one engine build: the 1.40× magnitude is platform-specific, the
mechanism and the direction are not.

qinhaihong-red:
## Round 1 handled — all nine items, one commit `d31b38e13`

Every finding adopted as adjudicated. No behaviour changed: the schema diff against
`21c7bfc87` is **0 added / 0 removed / 3 changed**, all three descriptions under
`commands[35]` (`perf monitors`) — the command help, the result docstring, and
`output.properties.collector_bytes.description`.

| # | Finding | Resolution |
| - | ------- | ---------- |
| P2-1 | The attribution rule understates the observer by 1.40x | Rewritten on **four** surfaces (see note below): `collector_bytes` is now the columns' LOGICAL size and a LOWER bound, both parts of the gap named, guidance kept qualitative (a few times = observer, an order of magnitude = game). The 1.4x is published in the catalog WITH its conditions, explicitly as a measurement of one host and one engine build. |
| P2-2 | `PackedFloat32Array` mutant survives the whole suite | Two assertions added to `test_perf_packed_value_bytes_mirrors_the_harness`, in its own grep idiom: `"= PackedFloat64Array()"` and `"timestamps := PackedInt64Array()"`. |
| P2-2b | No e2e assertion touches the window's timestamps | One line in the 600-frame e2e: `rows["samples"][-1]["timestamp"] > rows["samples"][0]["timestamp"]`. |
| P3-1 | Accumulator estimate "~10-14 KB" wrong | The figure appears in no committed artifact and in no PR/issue comment (checked the body, the issue, the PR comments, the review API and the branch's commit messages), so there was nothing to correct. Every new mention states the measured **17,128 bytes at 600 frames** — harness comment, catalog, PR body. |
| P3-2 | Deleting the snapshot branch's `"samples_omitted": {"type": "null"}` survives | `{**snapshot, "samples_omitted": False}` added to the output corpus. |
| P3-3 | Dropping `ge=0` survives | Corpus entry added **and** a one-sided assertion — see "one deviation" below. |
| P3-4 | Catalog states session-dependent digits as exact | Now "about 1.7 KB against about 260 KB (Godot 4.6.3; the exact digits vary between sessions)". The PR body carries the three raw runs behind that. |
| P3-5 | M2's killer count | Re-run: **13 failures** over the PR's four files, so `+9 more` -> `+12 more`. |
| P3-6 | Duplicated sentence in the e2e | Deleted. |
| P3-7 | The wrong reason for leaving the base accumulator uncounted | Replaced in all three places (harness comment, catalog, PR body) with the real one: `finalize` already receives `samples`, so counting it needs no base change; it is left out because it would not close the gap (81,600 + 17,128 = 98,728 against a measured 113,872, the rest being over-allocation on the columns already counted), and closing the rest would need an engine-internal `sizeof(Variant)` estimate. A true total belongs in a separately measured field. |

### P2-1's ratio, re-measured here

Not relayed — re-measured on this branch before publishing. A 600-frame window over all 16
monitors, read from the window's own `static_memory` column (first row against last row), on
macOS with **Godot 4.6.3-stable (official)**:

```
collector_bytes = 81,600   static_memory rise = 113,872 B   ratio = 1.395
```

Identical byte for byte on three consecutive windows. The catalog publishes exactly that, with
the host, the engine build, the frame count and the monitor count beside it, and says the ratio
is a measurement rather than a constant gda promises — only the direction generalizes, because
the number understates wherever a growing packed array over-allocates.

### Every new guard, shown killing its mutant

| Mutant | Killed by | Evidence |
| ------ | --------- | -------- |
| Three sampler `PackedFloat64Array` -> `PackedFloat32Array` | `test_perf_packed_value_bytes_mirrors_the_harness` | `AssertionError: the perf-sample value columns must stay PackedFloat64Array`. Before the fix the same mutant gave **2637 passed** on the fast tier (sha-pin test deselected, since any harness edit fails it and it cannot discriminate). |
| `timestamps := PackedInt32Array()` | same test | `AssertionError: the perf-sample timestamp column must stay PackedInt64Array` |
| `timestamps.append(0)` in the harness | `test_a_600_frame_summary_window_stays_compact_and_reports_its_own_cost` (real engine) | `assert 0 > 0` |
| Delete `"samples_omitted": {"type": "null"}` | `test_perf_monitors_schema_and_models_reach_the_same_verdict` | fails in the output-corpus loop on the snapshot instance |
| Drop `ge=0` from `PerfMonitorsResult.collector_bytes` | same test, at the new one-sided assertion | `assert not True` |

Each applied to the committed tree, confirmed red, reverted, and the tree confirmed back to its
intended state.

### One deviation, with evidence

**P3-3's remedy as written does not kill its mutant.** The corpus loop asserts only PARITY, and
the published `minimum: 0` is DERIVED from the model's `ge=0` — dropping the bound removes it
from both halves at once, so schema and model keep agreeing and the corpus entry stays green.
Verified: with `ge=0` removed, the run failed at the one-sided assertion (line 1474), not at the
corpus loop (line 1447). I therefore added both — the corpus entry the review asked for, and

```python
assert not schema_ok(doc["output"], {**window, "collector_bytes": -1})
assert not model_ok(PerfMonitorsResult, {**window, "collector_bytes": -1})
```

which is the same shape recheck 3's range keywords already use in that test, and is what makes
the published fact hold.

### Two judgement calls

**A fourth surface for P2-1.** The review named three (help, catalog, SKILL). The published
`collector_bytes` field description — the text an MCP client reads out of `--schema` — carried
the same rule in the same wrong form ("the approximate bytes ... retained ... Read it against
the window's own `static_memory` rise to tell the observer's footprint from the game's"), and
the result docstring carried its one-line echo. Leaving them would have left the false rule
standing in the surface agents consume most directly, so both were corrected. That is why the
schema diff has three changed leaves rather than one. Two assertions in
`test_perf_monitors_help_states_the_observer_cost_and_summary` now pin the corrected help so the
wording cannot quietly regress.

**Left alone:** the human render's `collector ~81600 bytes`. The tilde reads as an
approximation from below, which is now exactly what the number is; changing it would churn the
render tests for no gain.

### Gates

| Gate | Result |
| ---- | ------ |
| Fast tier | **2638 passed**, 685 deselected |
| ruff check / format | clean, 515 files formatted |
| pyright | 0 errors, 0 warnings, 0 informations |
| Real engine, `tests/live tests/harness tests/daemon -m e2e` under the machine-wide lock | **101 passed**, 629 deselected, 6:07 (host load ~10) |
| `harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | valid |
| PR CI at `d31b38e13` | all checks pass |

`HARNESS_VERSION` stays **24**; the harness comments changed, so `PINNED_HARNESS_SHA256` moves
to `c368e2c82f8ecd818897427e09634d907bca0ff051e49b05cb42c8c6166f87c6`.

qinhaihong-red:
## Review round 2 — CONVERGED. APPROVE, 0 P1 / 0 P2 / 6 P3 (all adopted; closing fix, then orchestrator verification)

Independent adversarial review at `d31b38e13` against the base `b7c485693`, on a running Godot
4.6.3. **Every round-1 finding is closed on the running system, every mutant the PR claims to
guard dies at exactly its named killer, and both re-measured figures reproduce byte for byte** on
the reviewer's host. Nothing reopened, nothing adjacent broke; the mechanism grew by two test
assertions, two corpus entries and one e2e line. No `/backtrace-review` is warranted. Two review
rounds are the budget and they are spent: the six items land as one closing commit and the
orchestrator verifies the head directly.

### Findings and dispositions

**P3-A — the `gda schema` field description, the fix round's own "fourth surface", has no
regression guard. ADOPTED (test).** Reverting `collector_bytes`' description to the pre-fix false
rule while leaving the help intact passes the whole fast tier (2638), whereas the same revert on
the help is killed at once by the two assertions the fix round added there. The asymmetry is the
wrong way round: the schema is the surface agents and gda-mcp read. One line in the parity test,
which already holds `doc`.

**P3-B — P3-4's false precision survives in a test comment. ADOPTED (wording).**
`tests/live/test_e2e_perf.py:289-296` still states "1,667 bytes … against 259,569 … ~4.9x" as the
measurement — the exact digits the catalog dropped, and the PR body says ~4.8x. Two independent
re-measurements disagree with both (1 700 / 1 690 / 1 689 and 1 677 / 1 678 / 1 685; 259 928 …
260 179). The round-1 finding fixed in three places and stale in a fourth — the class this round
was told to hunt. Use the catalog's own phrasing; keep the 8 192 bound and its rationale.

**P3-C — the PR body names the superseded harness pin. ADOPTED (body).** "re-pinned to
`738c71670e…`" — the tree pins `c368e2c8…`, recomputed and matching. Only the reply comment
carries the new value.

**P3-D — the human render still says `collector ~81600 bytes`. ADOPTED, and the reviewer is
right to flag it as the one worth elevating.** The implementer kept the tilde as "an
approximation from below". It is not: `~` reads as *approximately*, the word this round removed
from every other surface, and it is wrong in BOTH directions — the number is exact (600 × 17 × 8),
and the real in-game cost is 1.40× it. A human reading only the render and then a 113 872-byte
`static_memory` rise reaches precisely the false "game leak" P2-1 exists to remove, with `--help`
one command away. One token — `collector >=81600 bytes` — plus the two pinned strings in
`tests/cli/test_render.py`.

**P3-E — the mutant the suite still does not kill: the timestamp UNIT. ADOPTED (test).**
`PerfSampleFrame.timestamp` publishes "(ms)"; changing the one line this PR wrote to
`Time.get_ticks_usec()` passes the fast tier (2637, pin deselected) and the real-engine
`test_e2e_perf.py` (7). Round 1's `rows[-1] > rows[0]` catches a constant clock, not a wrong one,
and no other e2e module reads a perf window's timestamps. Bound the span on the same line: the
daemon's 30 s guard caps a 600-frame window, so `0 < span_ms < 30_000` is unfalsifiable in ms and
fails at ~4 133 000 in µs.

**P3-F — the harness comment publishes measured bytes without host or engine. ADOPTED (wording).**
`gda_harness.gd:1050-1055` gives frames and monitors but not the host or engine build, in a file
that is installed into the user's project. Round 1's rule was "conditions wherever the measurement
is published"; a comment shipped into a project is published. One clause.

### The two deviations, judged

**The fourth surface — correct, and the right call.** The pre-fix field description carried the
identical false rule in the surface `gda schema` and gda-mcp publish; leaving it would have left
round 1's defect standing where agents read it. It explains the schema diff honestly: three
changed description leaves, no shape change, verified leaf by leaf. The only reservation is P3-A.

**P3-3's corrected remedy — the reasoning is right and the proof reproduced.** With `ge=0`
dropped the run fails at the new one-sided assertion (`test_perf_commands.py:1474`), NOT at the
corpus loop, because `minimum: 0` is derived from `ge=0` and the mutant moves both halves
together, so parity stays silent. The corpus entry is not dead weight either: it would catch a
hand-written schema override diverging from a retained `ge=0`. Adding both was correct.

- **Gates, re-run in a byte-verified copy of `d31b38e13`:** fast **2638** (base re-run 2626 — the
  +12 holds); ruff clean; 515 formatted; pyright 0/0/0; `tests/live tests/harness tests/daemon
  -m e2e` **101 passed** in 200.6 s under the machine-wide lock; the harness guard valid. PR CI
  green, and its job list shows "Godot e2e tests: skipping" — the "PR CI runs no Godot" premise
  observed, not assumed.
- **The ratio, re-measured independently:** 600 frames × 16 monitors → `collector_bytes` 81 600,
  `static_memory` rise **113 872**, ratio **1.40**, identical byte for byte on four consecutive
  windows and matching the implementer's 1.395. **And checked at its boundary, which nobody had:**
  across 300 / 100 / 60 / 20 / 10 / 5 / 2 frames the ratio stays within 1.24–2.18, so the
  published qualitative rule ("a few times = observer, an order of magnitude = game") holds over
  the whole `--frames` domain on this host, not only at 600.
- **Both engine parentheticals in the catalog reproduce exactly** (headless probe, same engine):
  a 600-element `PackedFloat64Array` built by `append` costs 5 736 B against a logical 4 800; an
  Array of 600 nils costs 17 128 B. The catalog's arithmetic and its conclusion — counting the
  accumulator would not close the gap — are sound.
- **Every mutant killed by exactly its named killer**, eleven of them re-run, including the
  Float32/Int32 storage-type guards, the `ge=0` one at line 1474, `samples_omitted=not
  params.summary` at 13 failures (the corrected +12), and the help reverted to the old rule.
- **All five statements of the rule agree** (help, catalog, SKILL, the published description, the
  result docstring); a repo-wide grep for the old phrasings returns nothing; the "~10–14 KB"
  figure exists nowhere; the 1.4× ratio appears only in the catalog and one test comment — **no
  machine-read surface carries a number**, as claimed. README + three translations re-stamp to
  the value `shasum README.md` produces.
- **Merge hazards:** `sha256(gda_harness.gd)` = `c368e2c8…` = the pin; 24 > dev's 22; the guard
  requires only `head > base`, so **if #973 landed first, #967's 23 would fail — W3's order
  #967 → #973 is load-bearing, exactly as recorded.** `merge-tree` against #967: the same five
  files as at `21c7bfc87` (three stamps, `install.py`, the pin test) — the fix round added no
  overlap; against #939: clean.
- **Schema:** `21c7bfc87 → d31b38e13` 0 / 0 / 3 (all three descriptions under `commands[35]`);
  `b7c485693 → d31b38e13` 42 / 1 / 8, zero leaves outside `commands[35]`, 77 commands both sides.

### Not verified, and why

One host, one engine build (the direction generalizes, the magnitudes do not — as the PR says);
the Float32 mutant's PRE-fix survival taken from round 1, only the post-fix kill re-run; the µs
mutant run against the fast tier and `test_e2e_perf.py` only, justified by a grep showing no other
module reads a window's timestamps; the PR body's edit history not checked, only the live body.

### Operational note

The review checkout handed over was DIRTY: P3-2's mutant (`"samples_omitted": {"type": "null"}`
deleted from the snapshot branch) had been left behind in it. The reviewer could not restore it
under its permissions, worked instead in a copy verified hash-for-hash against `git ls-tree -r
d31b38e13` before and after every mutant, and read the schema branch from `git show`, so the
leftover entered no result above. The orchestrator restored the checkout on receipt.

qinhaihong-red:
## Round 2 closing fix — all six P3 items landed in one commit (`7b30ba7e7`)

Nothing deviated: every remedy was taken as written. Head is `7b30ba7e7`, one commit on top of
`d31b38e13`.

| Item | Resolution |
| ---- | ---------- |
| **P3-A** published description unguarded | One assertion added to `test_perf_monitors_schema_and_models_reach_the_same_verdict` (it already holds `doc`): `"LOWER bound on the in-game cost" in doc["output"]["properties"]["collector_bytes"]["description"]`. |
| **P3-B** stale digits in the e2e comment | `tests/live/test_e2e_perf.py` now says "about 1.7 KB at 600 frames, against about 260 KB for the same window with its rows (the exact digits vary between sessions)" — the catalog's own phrasing — and "several times the measured size" instead of the ratio. The 8,192 bound and its rationale are unchanged. |
| **P3-C** superseded pin in the PR body | Body rewritten from the FINAL tree: `PINNED_HARNESS_SHA256` = `3ba2ba9fd142d5b43fe62ceff2790aaf9b4efd0d652648518d47bb3e9c4eab04`, recomputed from `src/gda/harness/gda_harness.gd` after P3-F. `HARNESS_VERSION` stays 24. |
| **P3-D** the render's tilde | `collector >={collector_bytes} bytes`. `>=` reads as a lower bound at a glance, which is what the number is; `~` read as *approximately*, and was wrong in both directions. Both pinned strings in `tests/cli/test_render.py` moved with it. Human-only surface, so no schema leaf moves. |
| **P3-E** the timestamp UNIT | The same e2e line now reads `span_ms = last - first` and asserts `0 < span_ms < 30_000`. It keeps round 1's "not a constant clock" and adds the unit bound, which the daemon's 30 s window guard makes unfalsifiable in milliseconds. |
| **P3-F** harness comment without conditions | `gda_harness.gd` now says "measured at 600 frames over 16 monitors, on macOS with Godot 4.6.3: 81,600 + 17,128 against a 113,872-byte static_memory rise" — the catalog's form, in a file installed into the user's project. |

### The three new guards, each shown killing its mutant

| Mutant | Before | After |
| ------ | ------ | ----- |
| `collector_bytes`' published description reverted to the pre-fix "approximate bytes" (from `21c7bfc87`) | fast tier **2638 passed**, 685 deselected — survives | `tests/live/test_perf_commands.py:1479`: `assert 'LOWER bound on the in-game cost' in "The approximate bytes the gda harness's sampler retained for this window (#846)..."` |
| Human render reverted to `collector ~{collector_bytes}` | (the pinned strings were the old ones) | `tests/cli/test_render.py:291`: `- collector >=32 bytes, samples kept)` / `+ collector ~32 bytes, samples kept)` |
| `timestamps.append(Time.get_ticks_usec())` in the harness sampler | fast tier **2637 passed** (pin deselected) and `tests/live/test_e2e_perf.py` **7 passed** on the real engine — survives both | `tests/live/test_e2e_perf.py:358`: `AssertionError: 4133279` / `assert 4133279 < 30000` — the reviewer's ~4,133,000 reproduced exactly |

Each mutant was applied to the committed tree, run, then reverted; the harness file was restored
byte for byte from a pre-mutant copy before P3-F was applied, and the pin recomputed from the
final bytes.

### Gates at `7b30ba7e7`

| Gate | Result |
| ---- | ------ |
| Fast tier | **2638 passed**, 685 deselected, 47 s (unchanged — the two new assertions live inside existing tests) |
| ruff check / format | clean; 515 files already formatted |
| pyright | 0 errors, 0 warnings, 0 informations |
| `tests/live tests/harness tests/daemon -m e2e` on a real Godot 4.6.3, under the machine-wide lock, after the last edit | **101 passed**, 629 deselected, 3:22 (host load ~1.5) |
| `harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | "harness version transition is valid" |
| `sha256(src/gda/harness/gda_harness.gd)` | `3ba2ba9f…` = the pin |
| PR CI at `7b30ba7e7` | all checks pass ("Godot e2e tests: skipping", as expected) |

### Schema and greps

`gda schema` at `7b30ba7e7` against `d31b38e13`, leaf by leaf: **0 added / 0 removed / 0
changed**. The comparison mechanism was validated on the same run — the same script over a
`git archive` of `21c7bfc87` reproduces round 1's **0 / 0 / 3**, so the base schema genuinely
comes from the archived revision and the 0/0/0 is a real result, not a silent self-comparison.

A tree-wide grep for `~81600`, `collector ~`, `1,667`, `1 667`, `259,569`, `259 569`, `4.9x` and
`4.8x` over `*.py`, `*.md`, `*.gd`, `*.json` and `*.toml` returns nothing. No fifth copy of the
session-dependent digits survives; the only remaining `collector ~` anywhere is the mutant's
name in the PR body and in this comment.

qinhaihong-red:
## Round 2 handled at `7b30ba7e7` — verified directly by the orchestrator. **READY.**

Two review rounds were the budget and both are spent; this is the closing verification of the
implementer's round-2 commit.

### Verified

- **The three round-2 guards each kill their mutant, by exactly the intended test:**
  - the `collector_bytes` field description reverted to the pre-fix "approximate" rule (two
    sites) → `test_perf_monitors_schema_and_models_reach_the_same_verdict` reds alone;
  - the human render's `collector ~` restored →
    `test_render_perf_monitors_window_states_the_observer_cost_and_rows` reds alone;
  - `Time.get_ticks_usec()` in the sampler (pin test deselected) → the 600-frame e2e reds on a
    real engine at `assert 4133108 < 30000` — the reviewer's ~4 133 000 µs reproduced.
  All reverted clean.
- **Gates on the head:** fast tier **2638 passed**; ruff clean; 515 files formatted; pyright 0 / 0
  / 0; `tests/live tests/harness tests/daemon -m e2e` **101 passed** in 208 s under the
  machine-wide lock, after the last edit.
- **No published leaf moved:** `gda schema --json` **byte-identical** to `d31b38e13`
  (1 140 621 bytes, `cmp` clean). The closing diff is six files, 30 insertions / 18 deletions —
  one render token, two render-test strings, two test files, the pin, a harness COMMENT.
- **Harness:** `HARNESS_VERSION` 24; `sha256(gda_harness.gd)` recomputed =
  `3ba2ba9fd142d5b43fe62ceff2790aaf9b4efd0d652648518d47bb3e9c4eab04` = the pin; the version guard
  valid against dev at 22.
- **Greps:** `~81600`, `collector ~`, `1,667`, `1 667`, `259,569`, `259 569`, `4.9x`, `4.8x` —
  none in `src`, `docs`, `tests`, `CONTEXT.md`, `README.md`. The render token reads
  `collector >={collector_bytes} bytes`.
- **Commits:** four on top of dev, each authored and committed by `haihong.qin
  <haihongqin@gmail.com>`, no attribution of any kind, no force-push. PR CI green.

### Integration, on the whole wave

In the W3 trial gate — the eight slice heads merged onto dev in the wave's order — this slice
lands eighth, conflicts only on the `HARNESS_VERSION` line, the two pin lines and the three
translation stamps (all resolved in place, the auto-merged remainder kept), and is re-pinned at
**24** with the sha of the MERGED harness. On that merged tree: unit 2734 passed / 2 skipped,
ruff / format / pyright clean, the harness guard valid, and 500 real-engine e2e passed / 1
Linux-only skip across project / harness / export / resource / live / daemon / script / scene.

W3 merge order unchanged: `#939 → #937 → #938 → #940 → #964 → #966 → #967 → #973`; #967 (23)
lands before this slice (24), which the version guard makes load-bearing.

qinhaihong-red:
Independent adversarial review of base `b7c48569377e9796088137a61bb89f1ade39f9c9` to exact head `7b30ba7e7f1ad89e6e47fd80b919c1d91684c6a3`. Existing reviews and replies were not used as evidence.

The public docs turn one measurement into a causal rule: a rise of “a few times” `collector_bytes` is still observer cost, while an order-of-magnitude rise is game memory (`docs/command-catalog.md:1751-1757`, `src/gda/commands/perf.py:699-700,1249-1252`, `src/gda/skill/SKILL.md:248`). The implementation says the true total is unknown and excludes packed-array capacity plus the shared accumulator (`src/gda/harness/gda_harness.gd:1047-1058`). One 600-frame/16-monitor/macOS/Godot-4.6.3 ratio of 1.4x cannot support those thresholds. Remove the categorical attribution rule and state that this lower bound cannot attribute the residual by itself; recommend a matched control/baseline. Also reconcile issue #846's “approximate bytes” wording with the delivered exact logical-element lower-bound contract.

**Terminology and prose assessment:** I inspected issue #846 and its comment, the PR body, changed README/translations, command catalog, skill, help/schema descriptions, and sampler comments against `CONTEXT.md`, ADR-0015/0017/0018/0020/0023, and ASD-STE100. `summary`, `samples_omitted`, and `collector_bytes` are otherwise stable. The causal threshold above is the substantive prose/evidence defect. I did not benchmark allocator behavior across platforms, engine builds, frame counts, or monitor selections.

## Architecture review — Pass

Ownership and dependency direction are sound: the harness owns packed in-game collection, the CLI owns aggregation/budgets and the compact result projection, and the shared window base remains unchanged. No new module or abstraction is justified. The unsupported threshold is an evidence/documentation defect, not a boundary defect.

## Standards

- Hard: the categorical threshold conflicts with `RULES.md:36,67-71`. `collector_bytes` is only logical packed elements; omitted fixed/capacity costs are not proportional for small windows/selections, and concurrent game allocation is uncontrolled. State the lower bound and uncertainty, or publish a separately measured total with a validated bound.

No other documented violation or baseline smell was found; `git diff --check` passed.

Standards summary: 1 hard finding; worst is unsupported causal attribution.

## Spec

- Required: issue #846 asks help to explain how to read `collector_bytes` against `static_memory`; it does not authorize an unvalidated threshold. A game-side 5x rise can be mislabeled observer cost, while platform overhead above 10x can be mislabeled a game leak. Remove the thresholds and use a matched baseline, or supply evidence that establishes a portable bound.

The compact summary shape, packed sampler-local columns, recorded aggregate equivalence, unchanged snapshot/default behavior, schema admission, and 600-frame live size bound otherwise match the issue. Independent validation passed 136 focused tests and the new 600-frame real-daemon E2E.

Spec summary: 1 finding; worst is unsupported memory attribution.

qinhaihong-red:
## Third review handled at `ad6e3be8b`

The Artifact, Standards and Spec findings are one claim, and it is the third rewrite of the same
rule, so it was traced before it was patched. History: round 0 said a rise "of about
`collector_bytes`" is the observer; round 1 measured 1.4× and rewrote it as "a lower bound — a few
times is the observer, an order of magnitude is the game"; the third review shows those thresholds
are a causal rule one measurement cannot support. Root cause: the docs kept trying to hand the
caller a DECISION RULE from a number that cannot carry one — the parts outside `collector_bytes`
are not proportional to it (fixed costs dominate a small window or a narrow monitor set) and the
game allocates concurrently, so a game-side rise of five times the number reads as observer under
such a rule and platform overhead past ten times reads as a game leak. The design (harness owns the
packed collection, the CLI the projection) is kept, as the Architecture axis found.

| The published thresholds turn one measurement into a causal attribution rule | Fully Adopt | one host, one engine build, one frame/monitor pair; the two excluded parts are not proportional to the number | all five surfaces (published field description, help docstring, result docstring, catalog, skill) state the exact logical size as a LOWER bound that cannot attribute a `static_memory` rise on its own, and name the method that can: a matched baseline window — the same `--frames` and `--monitor` set on a scene that allocates nothing, on the same host — and compare the two rises. The measured figures stay as scale, with their conditions | the help pin now asserts `MATCHED BASELINE` is present and `order of magnitude` is ABSENT; live/cli/meta fast tier 1050 passed; ruff / format / pyright clean; CI green |
| Reconcile issue #846's "approximate bytes" with the delivered exact-logical-size contract | Fully Adopt | the issue says "approximate bytes … computed from the packed sizes"; the delivered number is exact and a lower bound | dated note on #846 | — |

Head `ad6e3be8b`, one commit on `7b30ba7e7`; schema moves only in the `collector_bytes` description and the two `perf monitors` descriptions; the harness is untouched (pin unchanged, `HARNESS_VERSION` 24).

Issue #979, opened 2026-09-15
export smoke: run a caller-selected Export artifact headlessly and return its completed process evidence


qinhaihong-red:
## Parent

#841 (the design decision). ADR-0042, `docs/adr/0042-artifact-smoke-runs-the-exported-build.md`, merged into `feat/gda-dogfooding-dev` as `86dae1f84`, is the authority for this slice. Where this issue and the ADR differ, the ADR wins.

## What to build

One new command, end to end (CLI, command descriptor, `--schema`, MCP discovery, `--help`, catalog, README row, SKILL guidance, unit and real-engine e2e tests):

```text
gda export smoke <artifact> [--arg VALUE ...] [--quit-after FRAMES] [--timeout SECONDS] [--strict]
```

Behaviour (ADR-0042, "Product contract"):

- **Projectless.** The descriptor sets `inherits_project=False`; the signature has no `--project`; the cwd and `GDA_PROJECT` are never used as project context. A relative `<artifact>` resolves against the invocation cwd, so the absolute `output_path` from `export run` passes through directly.
- **Artifact resolution.** Accept a directly host-runnable file; on macOS also resolve the main executable of a `.app` bundle. An absent path is `export_artifact_not_found`; a present input that cannot be resolved to a host-runnable Godot executable is `export_artifact_not_runnable`. No platform-format classifier. The resolution rule (2026-09-17): a regular file the host may execute is accepted as given; a `.app` bundle names its main executable in `Contents/Info.plist` (`CFBundleExecutable`), and the resolved path is `Contents/MacOS/<that name>`, which must be a regular file the host may execute; a bundle without that plist, key, or file is `export_artifact_not_runnable`, and so is every other directory and a file without execute permission. Nothing else is inspected: whether the resolved file is a Godot build is what the run shows.
- **Launch.** The shared one-shot Godot `Headless launch`, with the resolved executable as its second explicit executable source: `[executable, --headless, --log-file <gda-owned>, (--quit-after FRAMES), --, *args]`. Headless is fixed; there is no windowed mode.
- **`--timeout SECONDS`.** Defaults to `script run`'s 120 s (the same completed-run contract; no second number to explain); a longer game raises it. The `launch_timeout` message and the `Raw run`'s `timeout_bound` name the launch `Godot artifact smoke`, beside the existing `Godot script` / `Godot export` / `Godot import` / `Godot scene preflight` labels.
[...]

PR #978, opened 2026-09-14
docs(gda): define bounded headless Export artifact smoke (#841)

qinhaihong-red:
## Summary

Docs-only design decision for #841. It defines one capability:

```text
gda export smoke <artifact> [--arg VALUE ...] [--quit-after FRAMES] [--timeout SECONDS] [--strict]
```

The command runs a caller-selected exported Godot game headlessly on the current
host, within a timeout, and returns its completed process evidence. It exists
because `export run` can report `warnings: []` without observing defects that occur
only when the exported game exits (GDA-DF-072).

## Contract after the first-principles reset

- Reuse the existing **one-shot Godot Headless launch** in fixed headless mode.
  The configured editor executable and a resolved exported-game executable are its
  two explicit target sources. This does not create a generic process or mode
  platform.
- Keep `export smoke` projectless: set `inherits_project=False`, declare no
  `--project`, and resolve relative filesystem artifact paths against the invocation
  cwd. The absolute `export run.output_path` remains a direct handoff.
- Accept a direct host-runnable file; on macOS also resolve a `.app` bundle's main
  executable. Use `export_artifact_not_found` and
  `export_artifact_not_runnable`; do not add a platform-format taxonomy. These two
  refusals and `smoke_failed` are classifier-source `operation` codes with process
  exit 4.
- Return artifact path, resolved executable path, exit status, bounded stdout and
  spill metadata, stderr, and diagnostics. Extract only the completed-result base
  now shared with `script run`; each command keeps its own fields and policy.
- Normal completion returns exit status as data. `--strict` returns
  `smoke_failed` only for non-zero exit or `shutdown_leak`, reusing the existing
  typed exit-status and script-error evidence. The implementation must reconcile
  ADR-0004's producer set, field descriptions, and guard tests.
- Optional non-negative `--quit-after FRAMES` maps directly to Godot's release
  engine flag before `--`; omission or zero disables it. It lets the engine finish
  normally and emit shutdown diagnostics, while `--timeout` remains the external
  hard bound. It adds no completion policy, result field, state, or error code.
- Honor the existing global user-data override. Otherwise create a private root
  after artifact resolution and attempt to remove it in `finally` for every outcome;
  cleanup failure is internal, does not replace the verdict, and adds no public
  field, code, or evidence. Do not remove caller-owned overrides or publish transient
  placement paths.
- Add `ExecutionKind.ARTIFACT_SMOKE` as schema self-description only, with no new
  dispatch strategy.

[...]

Issue #976, opened 2026-09-14
script_errors: own the per-kind policy of ScriptErrorKind in one place (boot-verdict predicate, entry attribution, code map)


qinhaihong-red:
## Source

Candidate 1 of the architecture review of promotion PR #975 (W3 of milestone "gda dogfooding II", 2026-09-14; deletion test applied). Verified against the dev head `2858630c5`.

The shared script-error recognizer (`gda.script_errors`) owns the closed `ScriptErrorKind` set, the entry-failure precedence and `leaked_at_exit`. What it does NOT own is the rest of the per-kind policy — "does kind K gate verdict V" — which four consumers answer for themselves:

- `scene preflight`'s `started` excludes the process record by hand (`all(d.kind is ScriptErrorKind.SHUTDOWN_LEAK for d in diagnostics)`).
- the daemon's readiness verdict computes `clean_start = not recognized` with no exclusion at all. Unreachable today — the verdict reads the Session log up to the handshake, and an exit record cannot precede it — but the rule is unstated and untested, and the two boot verdicts (preflight `started`, daemon `clean_start`) can drift without anything noticing.
- `script run`'s entry-failure code map restates the precedence tuple's key set; the pair is held in lockstep only by a test (`test_every_entry_failure_kind_has_a_verdict_code`), whose comment admits a missing row would be a `KeyError` on a real failure path.
- `script run`'s completion-marker attribution (`_entry_attributable`) re-derives the recognizer's private path match because it cannot import it.

`--strict` and the `script_failed` message already ask the shared `leaked_at_exit`. #862 (the failure-evidence half of #850) lands FIRST and touches the same two files; it asks no per-kind question of its own, so it is not a consumer — this issue, landing second, adapts to whatever #862 leaves in `commands/script.py` and `errors.py` and keeps `leaked_at_exit`'s existing callers on the table. The spread is the same one that took PR #964 five review rounds to reconcile in prose, and ADR-0031's dated outcome note records a WIRE-SHAPE debt: a process-lifecycle record travels in `ScriptError`, a per-script shape, and a SECOND such record is the trigger for a broader diagnostic type. This issue does not pay that debt (no wire-shape change); what the table gives a second record is its POLICY row, by construction. The trigger stays recorded and open.

## What to build

Own the per-kind policy of `ScriptErrorKind` once, beside the enum, and make every consumer ask it instead of re-deriving it:

[...]

Issue #850, opened 2026-09-05
script run: cross-reference --user-data-root in help and report the effective user-data root


qinhaihong-red:
## Dogfooding source

Distilled from the kung-fu card-game dogfooding notes (`goal-dev` repo, `examples/card-games/kung-fu/Dogfooding.md`): **GDA-DF-049** (2026-08-20, gda 0.8.0) with follow-ups on 2026-08-25 and 2026-08-26 (gda 0.10.0), and **PIPE-DF-077** (2026-08-21). A persistence-bearing `script run --strict` failed with `write_failed` / `ERR_FILE_CANT_OPEN` under a restricted application-data directory, and later runs that omitted the option failed twelve suites and burned three 120-second timeouts. The global `--user-data-root` (#653) fixes it, but it appears only in `gda --help`, must precede the subcommand, and the result never states which `user://` root the script had — so the failure keeps being diagnosed as a game regression at the point where authors read `script run --help`.

Verified against gda 0.13.0 (`266c8009f`): `script run --help` mentions `user://` only as an address scheme, never the option; `scene preflight --help` does not mention it at all; the success result has no user-data field; only the `user_data_unwritable` refusal names the placement. On the default path the log lives in a private temporary directory that the launch deletes before the result is emitted, so a default-path `log_file` would be a dangling name. #653 scoped success-path disclosure out and pointed to #659, which covers install provenance, not per-launch placement.

## What to build

- `script run --help` cross-references the global option with the exact argument order and the `GDA_USER_DATA_ROOT` alternative, in the paragraph that describes `user://` writes; `scene preflight --help` gains that paragraph (it has none today).
- The `script run` result reports the launch's `User-data placement` with fields named for what gda holds: `user_data_root` — the `--user-data-root` / `GDA_USER_DATA_ROOT` value the launch ran under, omitted when none was given (omitted, not null); `engine_data_path` — the platform application-data directory the engine resolved `user://` beneath (`gda.runner.engine_data_path`), always present; `log_file` — only when a root was given, since that is the only case in which the file outlives the launch. Additive; the same facts the `user_data_unwritable` envelope already names.
- The skill's script-testing guidance shows the redirect as the default recipe for any script that writes `user://`.

## Acceptance criteria

- [ ] `script run --help` and `scene preflight --help` name `--user-data-root` / `GDA_USER_DATA_ROOT` with the ordering example.
[...]

PR #980, opened 2026-09-15
feat(gda): script run failure envelopes carry the user-data placement (#862)

### LINKED ISSUE #862: script run: carry `engine_data_path` on its `script_failed`, `launch_timeout` and `script_aborted` envelopes
## Parent

#850 published the launch's `User-data placement` on the `script run` SUCCESS result. This issue is the failure-envelope half, raised by PR #860's independent review (P2-2, 2026-09-05) and deferred there as out of #850's scope.

## Dogfooding source

The same misdiagnosis #850 addresses — **GDA-DF-049** (2026-08-20, gda 0.8.0, follow-ups 2026-08-25/26) and **PIPE-DF-077**: a persistence-bearing `script run --strict` failed on an unwritable `user://` and was read as a game regression, and later runs burned three 120-second timeouts on the same cause. `--strict` and the timeout are that record's own spellings, and on exactly those two paths the result is a failure envelope, which #850 left without the placement.

Verified against PR #860's head (`541eedd2f`): the `Raw run` carries `user_data` (root, engine data path, log file) out of the one launch primitive, and `script run` promotes it on the success result only. Its three run-reporting failure builders — `script_exit_status_failure` (`script_failed`), `script_run_timeout_failure` (`launch_timeout`) and `script_run_aborted_failure` (`script_aborted`) — are called with the raw run in scope but are not handed the placement, so their `evidence` carries the clocks, the child's exit status and the parsed `script_errors` and nothing about where `user://` was. `script run`'s timeout does NOT go through the shared `launch_timeout_failure`, so the other launch-backed channels are not on this path.

## What to build

`script run`'s three run-reporting failure envelopes carry the launch's `User-data placement` on `Failure evidence`, with the same presence rules the success result uses:

- `engine_data_path`: the directory the engine resolved `user://` beneath; present whenever the platform's data variable resolved (omitted, never null, when it did not — ADR-0004's rule for this object's own fields, where the success result says `null`).
- `user_data_root`: only when `--user-data-root` / `GDA_USER_DATA_ROOT` was given.
- `log_file`: only under a root, since that is the one case in which the log outlives the launch — and on a timeout it is where the caller looks next.

The facts pass ADR-0004's three-clause admission test: they are already computed on the failure path (the raw run holds them at every call site), they are not recoverable from `diagnostics` without parsing engine prose, and they change what the caller does next — re-run under `--user-data-root` instead of debugging the game. The producer set is unchanged (the three builders already carry evidence); what grows is the FACT set, so ADR-0004's "Who carries it today" paragraph gets a dated amendment saying which producers carry the placement, and the `FailureEvidence` schema publishes the three new optional fields.

Out of scope, stated so it is not read as an omission: `script_did_not_run_failure` (`script_not_found`, `script_compile_failed`, `incompatible_script_type`) and the pre-launch `script_escapes_project` refusal — a script that never ran fails clause three (the caller's next step is the script, not the environment); every other channel's `launch_timeout` and every sentinel envelope — #850's one-channel boundary stands, and the registry test that holds it extends to error envelopes in this change.

The routing-reversal rule applies: the published contract this slice changes lands with it. `script run --help`, the catalog and the bundled skill still say a failure envelope does not carry the placement; each stops saying so here, and the corrected sentences name all three envelopes (`script_failed`, `launch_timeout`, `script_aborted`), not the two the current sentences name. CONTEXT.md no longer says it: PR #978 removed the `User-data placement` sentence "Only there: a run that ends in an `Error envelope` … keeps its pre-#850 shape" and `Raw run`'s "the ONE channel that publishes the placement" clause while the behaviour still holds, so this slice ADDS the positive statement to `User-data placement`: `script run`'s three run-reporting failure envelopes carry the placement on `Failure evidence`, and every other channel's envelope does not. ADR-0031's `Outcome (2026-09-05, #850)` note says "byte identity is what the OTHER launch-backed channels and this channel's failure envelopes keep"; its failure-envelope half becomes false here and gets a dated line saying so (the other channels' byte identity stands). While in `tests/cli/test_error_registry.py`, correct its count prose ("a sixth builder", "Producers six to eight") to the current nine.

## Acceptance criteria

- [ ] `gda script run --strict <script that calls quit(1)> --json` reports `script_failed` whose `evidence` carries `engine_data_path` beside `exit_status` and `script_errors`, and neither `user_data_root` nor `log_file`; the same run under `gda --user-data-root DIR …` carries all three, with `engine_data_path` under `DIR` and a `log_file` under `DIR` that exists after the command returns.
- [ ] `gda script run --timeout <small> <script that never quits> --json` reports `launch_timeout` whose `evidence` carries the same keys beside `elapsed_seconds`, `timeout_seconds` and `termination_phase`; a `script_aborted` envelope (completion marker declared, recognized error, silence) carries them too.
- [ ] Every other failure envelope of `script run` and every other channel's `launch_timeout` envelope is byte-identical to before; the boundary is pinned for error envelopes by a builder-level guard (which builders pass the placement — the AST shape `_EVIDENCE_PRODUCERS` already uses), because the shared `FailureEvidence` cannot express a single-channel boundary by model shape and the descriptor-registry test walks result models only; the evidence producer-set test still passes with an unchanged set.
- [ ] `gda schema` publishes the three optional fields on `FailureEvidence`; ADR-0004's producer paragraph carries the dated amendment; `script run --help`, the catalog and the skill no longer say the failure envelopes lack the placement and their corrected sentences name all three envelopes; CONTEXT.md's `User-data placement` entry states that `script run`'s three run-reporting failure envelopes carry the placement on `Failure evidence` and that every other channel's envelope does not; ADR-0031's #850 outcome note carries the dated line superseding its failure-envelope byte-identity clause; a real-engine test covers the two shapes of the `--strict` case.

## Amendments

- **2026-09-15, W4 artifact review:** the CONTEXT.md clause of AC4 was satisfiable by making no edit (PR #978 had already removed the negative sentence), so it is now a positive requirement; ADR-0031's #850 outcome note joined the reconcile list (its byte-identity clause would have become false); AC3's boundary guard is builder-level, since the descriptor-registry test walks result models only; the corrected sentences must name three envelopes, not two; `Failure evidence` was dropped from the list of entries that "currently say" it (that entry never did).

## Blocked by

- #850 (PR #860, on `feat/gda-dogfooding-dev`; this slice can start once that PR is on the dev branch, and it appends to `FailureEvidence`, the hotspot #845 owns in its wave — serialize behind #845's merge).
- **2026-09-15:** both gates are discharged — #850 merged (`36b96bc23`, W1) and #845 merged on dev (`d0886b9a2`, W3). The slice can start now; in W4 it merges after #839 and before #976 (shared `commands/script.py` + `errors.py`).

qinhaihong-red:
## Summary

`gda script run`'s three failure verdicts that report on a RUN — `--strict`'s
`script_failed`, this channel's own `launch_timeout`, and `script_aborted` — now carry the
launch's `User-data placement` on `Failure evidence`, with the presence rules #850 gave the
success result.

This is the failure-envelope half of #850. The dogfooding record (GDA-DF-049,
PIPE-DF-077) is about a persistence-bearing `--strict` run whose `user://` write failed and
was read as a game regression, and about three 120-second ceilings burned on the same cause.
Both of those paths end in an `Error envelope`, which is exactly what #850 left without the
placement.

The producer set of ADR-0004's evidence axis is UNCHANGED — all three builders already carry
evidence. What grows is the fact set, so the schema publishes three new optional fields and
the ADR's producer paragraph records which producers report the placement.

Closes #862

### Behaviour — `src/gda/models.py`, `src/gda/errors.py`, `src/gda/commands/script.py`, `src/gda/render.py`

- `FailureEvidence` gains `engine_data_path`, `user_data_root` and `log_file`, each optional
  and omitted rather than null. The presence rules are the success result's with ONE
  divergence, which is this object's own rule: the success result reports a null
  `engine_data_path` when the platform's data variable is unset, while here the key is
  absent.
- `models.py` also gains the ONE projection of the placement into those public key names:
  `PLACEMENT_FIELD_NAMES` and `placement_fields()`, beside the fields that declare them,
  returning only the facts the launch had. It sits in the contract core because the core
  owns what the wire CALLS those paths, while `runner.py`'s `UserDataReport` stays the RAW
  fact owner — three optional paths and the lifetime rule that decides which of them are
  facts. (Round 1 created that single projection on `UserDataReport` itself; round 3 moved
  it to the core. Same semantics, same callers.) `errors.py` gains
  one private helper, `_placement_evidence`, that reshapes that mapping into the three
  positional values the builders spell as explicit keyword arguments.
  `script_exit_status_failure`, `script_run_timeout_failure` and
  `script_run_aborted_failure` take a REQUIRED keyword-only `user_data` and pass the
  triple into their existing `FailureEvidence`. No new builder, no new error code, no
  change to the envelope's key set.
- `commands/script.py` passes `raw.user_data` at the three call sites. The facts stay the
  launch's — read off the `Raw run`, never resolved a second time — so an envelope cannot
  name a placement the run did not have. The edits are three argument lines plus the
  reversed help paragraph, kept small for #976, which lands on the same two files.
- `render.py` grows the three matching branches, so the human rendering of a failure carries
  what `--json` does.
- `src/gda/runner.py`: the comment on `RunResult.user_data` stated the old rule ("Today
  exactly one does: `script run`'s SUCCESS result … Follow-up: extend `Failure evidence`
  with `engine_data_path` on `script_failed` / `launch_timeout`"). It now names this change
  instead. That comment is the module's only lasting edit — round 3 moved the key names and
  the projection out to `models.py`, so the primitive holds the raw placement facts and
  nothing about the contract's spelling.

Out of scope, kept out deliberately: `script_did_not_run_failure`
(`script_not_found` / `script_compile_failed` / `incompatible_script_type`) and the
pre-launch `script_escapes_project` refusal — a run that never started fails the admission
criterion's third clause, since the caller's next step is the script, not the environment;
and every other channel's `launch_timeout` plus every sentinel envelope — #850's
one-channel boundary stands.

### Guards — `tests/cli/test_error_registry.py`, `tests/cli/test_human_failure_output.py`, `tests/cli/test_models.py`, `tests/meta/test_schema_command.py`, `tests/script/**`

- `test_only_the_three_named_builders_put_the_placement_on_evidence` reads out of
  the source which builders construct a `FailureEvidence` with a placement key — the same
  AST shape `_EVIDENCE_PRODUCERS` uses. The boundary is held at the BUILDER because the two
  guards that exist cannot hold it: `FailureEvidence` is one schema shared by every command,
  so the model's shape cannot say "this channel only", and the descriptor-registry placement
  guard walks RESULT models.
- `test_the_shared_launch_timeout_discloses_no_placement_it_was_handed` pins the same
  boundary on behaviour: the shared builder is handed the whole `Raw run`, which since #850
  carries the placement.
- `test_only_the_recorded_producers_put_evidence_on_the_envelope` passes with an UNCHANGED
  nine-name set. While in that file, two stale counts were corrected to the current nine:
  "a sixth builder" → "a TENTH builder", and "Producers six to eight (the two `target_*`
  refusals and `export_templates_missing_failure`)" → "Producers six to nine (… and
  `path_case_mismatch_failure`)".
- `_EVIDENCE_SAMPLES` gains the three samples, which the sample-table equality against
  `FailureEvidence.model_fields` and the per-field render test both require.
- The two published-schema field guards (`tests/cli/test_models.py`,
  `tests/meta/test_schema_command.py`) gain the three names.
- `tests/script/test_script_run_operation.py`: seven operation-level tests for the presence
  rules — default run, redirected run, timeout, abort, an unresolved platform path omitted
  rather than nulled, a run with no placement report, and a never-ran verdict carrying none.
- `tests/script/test_e2e_script_run.py`: four real-engine tests — `--strict` under a root
  (all three keys, the data path derived under the root, the log still on disk after the
  command returns), the timeout under a root, the abort under a root, and the never-ran
  verdict under a root carrying nothing. The two existing evidence key-set EQUALITIES now
  read `{"exit_status", "script_errors", "engine_data_path"}`, which is AC1's first shape:
  a default run gains exactly one key and the other two stay out.

### Published contract — the routing reversal

`script run --help`, `docs/command-catalog.md` and `src/gda/skill/SKILL.md` each stopped
saying a failure envelope does not carry the placement, and each corrected sentence names
all THREE envelopes rather than the two the old sentences named.

`CONTEXT.md`, `User-data placement` — PR #978 had already REMOVED the negative sentence, so
this adds the positive one:

> Three of that channel's failure envelopes carry it too, on `Failure evidence` —
> `script_failed`, `launch_timeout` and `script_aborted` — since that same misdiagnosis is
> exactly what a `--strict` failure and a timeout report. Those three codes and no others:
> every other `script run` verdict carries no placement, whether or not the script ran
> (`engine_crashed` and `stdout_spill_failed` did run and carry none), and so does every
> other channel's envelope (#862). The presence rules are the success result's, with the one
> difference that object imposes: a field that is not a fact is omitted, never null.

`Raw run` and `Failure evidence` were checked and need no edit: `Raw run` describes the
SUCCESS promotion, which is unchanged, and `Failure evidence` describes the object
generically and names its fields illustratively. Nothing else in `CONTEXT.md` was touched —
the `export smoke` present tense ahead of #979 is left as it is.

ADR-0004 gains a dated amendment after the #687 block:

> **Outcome (2026-09-15, #862): the producer SET is unchanged; what grows is the FACT set
> three of them report.** … The facts pass the criterion above on all three clauses: the
> `Raw run` holds the placement at every one of those call sites (#850), the directory the
> engine resolved `user://` beneath is not recoverable from `diagnostics` without parsing
> engine prose, and it decides the caller's next move — re-run under `--user-data-root`
> instead of debugging the game. … the boundary is ONE channel's, which this schema cannot
> express — `FailureEvidence` is shared by every command — so it is held at the builder
> instead.

ADR-0031 gains a dated outcome note after the #844 block:

> **Outcome (2026-09-15, #862) — the failure-envelope half of the #850 note's byte-identity
> clause is superseded, and so is its "SUCCESS result alone" boundary; the other channels'
> half of both stands.** … So both sentences now read the same way: the success result is no
> longer alone, and byte identity is what the other launch-backed channels keep, and what
> this channel's OTHER envelopes keep — the never-ran verdicts …, the pre-launch
> `target_outside_project` refusal, and the two verdicts about a run that DID run and are on
> no evidence axis, `engine_crashed` and `stdout_spill_failed`. Three named codes, not a
> category.

Both are appended blocks; no existing ADR prose was rewritten.

README: checked and NOT changed. It mentions `--user-data-root` only as a capability in one
table row and never states the failure-envelope rule, so no translation or i18n marker
refresh was needed (`tests/repo/test_readme_i18n_sync.py` passes unchanged).

## Schema / help diff — the declared delta, enumerated

Baseline captured from a detached worktree at `origin/feat/gda-dogfooding-dev`
(`86dae1f84`), not a stash: `uv run --frozen gda schema` plus `--help` for the root, every
group, all 77 commands and the five meta commands (98 invocations), at the base and at this
head, then diffed.

**`gda schema`** — 77 commands at the base and at the head, the same names, and:

1. `error.$defs.FailureEvidence.properties` gains exactly `engine_data_path`,
   `user_data_root` and `log_file`. No property removed, no existing property changed.
   Because ADR-0004 publishes ONE shared `error` schema per command, this appears in all
   **77** commands' error schemas — asserted: the `error` object is byte-identical across
   every head command, and after deleting the three new properties it is byte-identical to
   the base's.
2. `script run`'s `description` changes — the reversed help paragraph below.

Nothing else: `input`, `output`, `kind`, `constraints` and `argv` are identical for all 77
commands, and `description` is identical for the other 76.

Manifest size 1,216,293 → 1,290,672 bytes (+74,379, +6.1%), which is the three properties
times 77.

**`--help`** — one hunk, in `gda script run --help` (13 changed lines, whole capture
363,844 → 364,413 bytes). Removed:

```
since by default it is a private temporary file gda removes. A failure envelope
(``--strict``'s ``script_failed``, a ``launch_timeout``) does not carry them.
```

Added:

```
since by default it is a private temporary file gda removes. Three failure
envelopes say the same, under ``evidence``: ``script_failed``, ``launch_timeout``
and ``script_aborted``. There the keys follow the omitted-never-null rule of that
object, so an unresolved ``engine_data_path`` is absent rather than null. Those
three and no others — every other failure of this command, ``engine_crashed`` and
``stdout_spill_failed`` included, carries no placement.
```

Every other command's `--help` is byte-identical. Both figures are re-measured at head
`686a760f5`, against the same base capture. Round 1's prose fix grew this one help hunk from
12 to 13 lines; round 2's schema fix changed the `log_file` property's `description` text
inside the same shared error object and moved nothing else — the help capture is
byte-identical to round 1's, and `script run`'s description did not change again.

**Runs on PR CI** (ruff, pyright, the `not e2e` suite): the builder-level AST guard, the
shared-`launch_timeout` behaviour guard, the unchanged nine-name producer-set guard, the
seven operation-level presence tests, the `_EVIDENCE_SAMPLES` equality plus the per-field
render tests, and the two published-schema field guards.

**Runs only in the wave's e2e dispatch** (a real Godot binary): the four new
`tests/script/test_e2e_script_run.py` tests, the two amended evidence key-set equalities in
that file, and `tests/runtime/test_e2e_launch_timeout.py`'s
`set(evidence) == {elapsed_seconds, timeout_seconds, termination_phase}` — the pre-existing
real-engine pin that the shared `launch_timeout` is still byte-identical on another channel.
All of those were run locally for this PR; see Validation.

## Validation

Host: this Mac, macOS (Darwin 25.6.0), Godot 4.6.3-stable. Every Godot suite ran in the
foreground through the machine-wide e2e lock, serially.

| command | result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `516 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2770 passed, 3 skipped, 1 warning in 20.85s` (superseded by the round-1 and round-2 tables below) |
| `pytest -m e2e tests/script -q -rs` | `126 passed, 318 deselected in 149.30s (0:02:29)` |
| `pytest -m e2e tests/runtime/test_e2e_launch_timeout.py -q -rs` | `3 passed in 64.71s (0:01:04)` |

The second e2e suite is not a module this PR edits. It was run because it holds the
real-engine pin for the other half of the boundary: the shared `launch_timeout` builder
still reports exactly three evidence keys on `resource import`.

The full set was re-run at the later rounds; the current head's numbers are in
**Review round 3 -> Round-3 validation**.

## Mutant checks

Each mutant was applied to a copy-restored file (`cp` from a saved pristine copy, never
`git checkout --`), confirmed RED, restored, and confirmed GREEN. `git diff HEAD` is empty
after all seven.

| # | mutant | RED |
| --- | --- | --- |
| M1 | `script_run_aborted_failure` stops passing the three keys | AST builder guard + `test_an_aborted_run_reports_the_placement_too` |
| M2 | the SHARED `launch_timeout_failure` starts passing them | `test_the_shared_launch_timeout_discloses_no_placement_it_was_handed` + AST builder guard |
| M3 | `_placement_evidence` always reports nothing | the four presence tests (strict default, strict under a root, timeout, abort) |
| M4 | an unresolved `data_path` becomes `""` instead of an omitted key | `test_an_unknown_platform_data_path_is_omitted_from_a_failure_envelope` |
| M5 | `script_did_not_run_failure` starts carrying the placement | `test_a_never_ran_verdict_carries_no_placement` + AST builder guard |
| M6 | the human renderer drops the `log_file` branch | `test_every_evidence_field_alone_reaches_the_human_block[log_file]` |
| M7 | the model drops `user_data_root` | both published-schema field guards + the sample-table equality |

Round 3 re-checked the projection's two rules at its new home and reports one mutant that
SURVIVES; see **Review round 3 -> Round-3 mutant checks**.

## Deviations from the issue

None on behaviour or on the reconcile list. Three things the issue does not name, all small:

- `src/gda/runner.py`'s comment on `RunResult.user_data` stated the rule this change
  reverses and named the follow-up as open. Corrected; comment only.
- `render.py`'s `_evidence_lines` docstring twice said "a sixth field", which the three new
  fields falsify. Made count-free ("a further field") rather than restating a number the
  next field would falsify again.
- `tests/cli/test_models.py` and `tests/meta/test_schema_command.py` hold the published
  `FailureEvidence` property set as an equality, so both had to gain the three names. They
  are not in the issue's artifact list but are the same kind of guard as
  `_EVIDENCE_SAMPLES`.

- **Linux and Windows**: not run. Nothing added is platform-specific — the placement comes
  off the existing launch primitive and the fields are plain strings — but the real-engine
  arms were exercised on macOS only.
- **The `engine_data_path`-omitted arm on a real engine**: covered at the operation level
  only. Reaching it live needs a host whose platform data variable is unset, which this
  machine does not offer.
- **The AST builder guard's two known limits**, both stated in the test and both inherited
  from `_EVIDENCE_PRODUCERS` since #687 rather than introduced here. It reads keyword
  arguments at the `FailureEvidence(...)` call, so a `FailureEvidence(**something)` would
  pass unseen; and it parses ONE module, `gda.errors`, so a `FailureEvidence(...)` built
  anywhere else would be invisible to it. Every producer today lives in that module and
  spells its fields — verified by grep at review: `FailureEvidence(` appears only in
  `errors.py` and `models.py`, and no `evidence=` argument exists outside `errors.py`.
- **Payload growth**: the shared error schema repeats once per command, so the three
  properties cost +74 KB (+6.1%) on the aggregate `gda schema`. ADR-0004's #687 amendment
  already prices this axis and names the structural fix (emitting the envelope once) as
  ADR-0012's decision, not this one's.

---

## Review round 1

Round 1 returned "Changes required" against `977c81a37`: P1 0, P2 2, P3 6. All eight are
adopted. One commit, **`59ab8300b`**.

| finding | decision | fix |
| --- | --- | --- |
| **P2-1** The prose introduced the three as a CATEGORY ("the failure envelopes that report on a RUN"), which `script run`'s `engine_crashed` and `stdout_spill_failed` also are while carrying nothing (reproduced live: `OS.crash()` under a root → `engine_crashed`, no `evidence` key, `DIR/logs/godot.log` on disk) | Adopted, documentation only. The two verdicts stay WITHOUT the placement | Every surface now names the three codes instead of defining a category — `script run --help` / the schema description, the catalog, `SKILL.md`, `CONTEXT.md`, both ADR notes, plus `runner.py`'s comment, `models.py`'s comment and the test comment blocks. ADR-0004's #862 note gains the recorded reason: neither verdict is on the evidence axis at all, `engine_crashed` is the SHARED `classify_launch_or_crash` verdict every launch-backed channel reaches (so disclosing there would reach `export run`, `resource import` and `scene preflight` in one edit) and its `diagnostics` already carries the crash account, and `stdout_spill_failed` names its own path in its own message |
| **P2-2** The human rendering was guarded for presence only; a `log file:` label printing `user_data_root` passed the whole fast tier | Adopted | `test_the_placement_renders_as_three_labelled_lines` renders the three together and asserts the three exact lines, label and value. The reviewer's own mutant now goes RED (R1-M1 below) |
| **P3-1** ADR-0031's note superseded the #850 note's byte-identity clause but not its "SUCCESS result alone" boundary, which it also falsifies | Adopted | The note's heading and first paragraph now quote and supersede BOTH sentences |
| **P3-2** The placement's string projection existed twice (`_placement_evidence` and the success half's inline triple) | Adopted | `UserDataReport.as_strings()` in `runner.py` — the module that owns the lifetime rule — is now the ONE projection, returning only the PRESENT facts. Both halves read it; each keeps its own null contract without re-deciding it (the success model's `engine_data_path` default is `None`, so a missing key still publishes `null`; `Failure evidence` omits it). `_placement_evidence` survives only as the shape the builders need — three positional values they spell as explicit keywords, which is what keeps the AST boundary guard able to read them |
| **P3-3** `user_data` defaulted to `None`, so a dropped call-site argument was a silent revert | Adopted. Grep confirmed the only callers are the three sites in `commands/script.py` plus four test call sites | The parameter is now REQUIRED and keyword-only on all three builders. Dropping it at a call site is a pyright error (R1-M4) |
| **P3-4** The three field names were held in two test modules | Adopted | Both read `PLACEMENT_FIELD_NAMES` from `gda.runner`, which `as_strings()` itself builds from |
| **P3-5** `CONTEXT.md`'s "the record it answers" had no antecedent in the entry | Adopted | Reworded to "that same misdiagnosis", whose referent is the preceding sentence |
| **P3-6** `_classify_ended_run`'s "All of it is PROSE … Do not add envelope fields here" was stale | Adopted | Corrected to the current rule: prose plus the typed `evidence` keys #687 and #862 admitted; a new envelope FIELD is still ADR-0004's decision |

### Round-1 mutant checks

Each applied to a `cp`-saved pristine copy, `__pycache__` deleted after every edit and
every restore, confirmed RED, restored, confirmed GREEN. `git status` empty afterwards.

| # | mutant | RED |
| --- | --- | --- |
| R1-M1 | the `log file:` label prints `user_data_root` — the reviewer's exact mutant D, which was GREEN at `977c81a37` | `test_the_placement_renders_as_three_labelled_lines` |
| R1-M2 | the `engine data path:` label prints `log_file` | the same test |
| R1-M3 | `UserDataReport.as_strings()` stops filtering absent facts | four tests across BOTH halves — `test_default_run_reports_the_engine_data_path_only`, `test_an_unknown_platform_data_path_is_reported_as_null_not_omitted`, `test_a_strict_failure_reports_the_engine_data_path_of_a_default_run`, `test_an_unknown_platform_data_path_is_omitted_from_a_failure_envelope`. This is the P3-2 pin: the success result's byte identity and the evidence object's omitted-never-null rule both hold the one projection |
| R1-M4 | drop `user_data=raw.user_data` at the timeout call site | pyright: `Argument missing for parameter "user_data" (reportCallIssue)`, plus 12 tests — the silent revert P3-3 names is now a type error |
| R1-M5 | rename one key in `PLACEMENT_FIELD_NAMES` | `test_only_script_run_publishes_the_launch_user_data_placement` — the constant really is the result-model guard's authority (P3-4) |

### Round-1 validation

Same host: this Mac, macOS (Darwin 25.6.0), Godot 4.6.3-stable. Godot suites in the
foreground through the machine-wide lock.

| command | result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `516 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2771 passed, 3 skipped, 1 warning in 27.19s` |
| `pytest -m e2e tests/script -q -rs` | `126 passed, 318 deselected in 152.18s (0:02:32)` |
| `pytest -m e2e tests/runtime/test_e2e_launch_timeout.py tests/runtime/test_e2e_user_data.py -q -rs` | `12 passed in 72.52s (0:01:12)` |

The third suite widened from round 0: `test_e2e_user_data.py` joins it because
`UserDataReport` gained the shared projection, and `test_e2e_launch_timeout.py` still holds
the real-engine pin that the shared `launch_timeout` reports exactly three evidence keys.

The schema and help re-diff is folded into the "Schema / help diff" section above, measured
at `59ab8300b` against the same base capture. The delta is the same in kind — the three
properties in all 77 error schemas plus `script run`'s one description — with the one help
hunk grown from 12 to 13 changed lines by the P2-1 wording. No other command's description,
`input`, `output`, `kind`, `constraints`, `argv` or `--help` moved.

### Not verified in this round

- `stdout_spill_failed` on a real engine. Its exclusion is argued from the code path and
  ADR-0031's #665 note, as the reviewer's own finding was; only `engine_crashed` was
  reproduced (by the reviewer).
- Linux and Windows, unchanged from round 0.

---

## Review round 2

Round 2 is a **PASS** against `59ab8300b`, with four P3 findings. All four are adopted.
One commit, **`686a760f5`**.

| finding | decision | fix |
| --- | --- | --- |
| **P3-1** The category wording survived in two TEST surfaces: `tests/script/test_script_run_operation.py`'s section header named "the FAILURE envelopes that report on a run", which its own paragraph three lines below refutes; `tests/cli/test_error_registry.py`'s guard comment and test name said "run-reporting" without ADR-0004's "that ALREADY carried evidence" qualifier | Adopted | The header now reads "THREE named failure envelopes of this channel"; the guard comment names the three ADR-0004's note names and spells out why the bare category is wrong; the test is renamed `test_only_the_three_named_builders_put_the_placement_on_evidence`, and the sections above use the new name — the round-0 and round-1 mutant tables name the guard by role, not by symbol, so neither went stale |
| **P3-2** This body still quoted the ROUND-0 text of two artifacts it says round 1 fixed (`CONTEXT.md` carried both round-1 defects, the ADR-0031 quote omitted the P3-1 fix), and described `user_data` as optional | Adopted | Both block-quotes above are re-taken from the current files; "Behaviour" now says `UserDataReport.as_strings()` is the projection and `_placement_evidence` only reshapes it, and that `user_data` is a REQUIRED keyword-only parameter. The round-0 fast-tier row is marked superseded rather than deleted, since it is what ran at `977c81a37` |
| **P3-3** `FailureEvidence.log_file`'s published description was a garden-path sentence ("and on a run gda ended the file to read next" parses first as *gda ended the file*) | Adopted | Two sentences now: "… the one case in which it outlives the launch. On a run gda ended, it is the file to read next." Schema re-diffed; numbers above refreshed |
| **P3-4** The AST guard disclosed its `**splat` limit but not that it parses ONE module (`gda.errors`) | Adopted | Both limits are stated in the test comment and in the residual-risk bullet above, with the grep that shows the convention holds today |

### Round-2 schema / help re-diff

Re-captured at `686a760f5` against the same base capture, and also diffed against round 1's
capture to isolate what this commit alone moved:

- **base → head**: unchanged in kind. 77 commands both sides, same names; `FailureEvidence`
  gains exactly the three properties, none removed, no PRE-EXISTING property changed; the
  error object is byte-identical to the base's after deleting those three, and identical
  across all 77 head commands; `input` / `output` / `kind` / `constraints` / `argv`
  unchanged everywhere; `script run`'s `description` is the only other change. Manifest
  1,216,293 → **1,290,672** bytes.
- **round 1 → round 2**: the ONLY difference is `FailureEvidence.log_file`'s `description`
  string, inside the one shared error object (so it shows on all 77 commands). `script run`'s
  description did not change again, and the whole 98-invocation `--help` capture is
  byte-identical to round 1's.

### Round-2 validation

Same host: this Mac, macOS (Darwin 25.6.0), Godot 4.6.3-stable.

| command | result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `516 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2771 passed, 3 skipped, 1 warning in 25.04s` |

No Godot e2e re-run this round: the only non-comment change is one published `description`
string, which no e2e test reads. Round 1's `tests/script` (126 passed) and
`tests/runtime` (12 passed) results stand for the code.

### Round-2 mutant check

One, because a rename is exactly the kind of edit that can quietly detach a guard. Round 0's
M1 — `script_run_aborted_failure` stops passing the three keys — re-applied to a `cp`-saved
pristine `errors.py` with `__pycache__` cleared: RED on
`test_only_the_three_named_builders_put_the_placement_on_evidence` (the renamed guard) and on
`test_an_aborted_run_reports_the_placement_too`; restored, GREEN, `git status` empty. No other
mutant was re-run: the remaining three findings are comments, a docstring `description` string
and PR prose, and no rule changed.

---

## Review round 3 (external)

An independent adversarial review of `686a760f5` returned **Standards: Pass** and **Spec:
Pass**, with two artifact findings and one architecture finding. All three are adopted in
full. Two commits, **`070bdcc9e`** (docs) and **`cea4a1fd0`** (refactor).

| finding | decision | fix |
| --- | --- | --- |
| **P2-1** ADR-0004's #862 note argued `engine_crashed`'s exclusion with a FALSE clause: "its `diagnostics` already carries the engine's crash account, which is the second clause". That clause asks whether the PROPOSED fact — the placement — is recoverable without parsing prose, and the crash classifier is passed only `raw.stderr`, which need not hold a placement or even a crash account (a crash with empty stderr gives empty diagnostics) | Fully adopted, documentation only. The exclusion stands; its stated reason changes | `070bdcc9e` first replaced the false second-clause argument with a third-clause claim. The follow-up `5a250e107` corrects that claim: the pre-spawn probe belongs to `runner.user_data_placement`, does not check the default `user://` directory, and cannot establish that placement is unrelated to every later signal death. The current note retains the three named producers and the shared-verdict/channel-scope reason without asserting an unproven admission-clause failure. Both corrections edit this PR's own unmerged note in place |
| **P3-1** The same note attributed BOTH the failed-write misdiagnosis and the three later 120-second ceilings to ADR-0031's #850 note, which records the former only | Fully adopted | `070bdcc9e`. The note now splits the record between its two sources: ADR-0031's #850 note holds the misdiagnosis (GDA-DF-049, PIPE-DF-077), and issue #862's "Dogfooding source" section holds the timeout history, which is cited for that half |
| **P3-2 (architecture)** `runner.py` had grown the JSON key names of `ScriptRunResult` and `FailureEvidence` plus a description of their two different null behaviours. ADR-0040 §5 puts the public contract models in the shared core and the launch primitive below it, so the primitive should not know the contract's spelling | Fully adopted | `cea4a1fd0`. `PLACEMENT_FIELD_NAMES` and ONE function, `placement_fields(report)`, now live in `gda.models` beside the `FailureEvidence` fields that declare the three names. `UserDataReport` stays in `gda.runner` as the RAW fact owner (three optional paths, the lifetime rule) and names no public key string, `ScriptRunResult` or `FailureEvidence` anywhere. `models.py` imports the record under `TYPE_CHECKING` only, so the core takes no runtime edge. Both halves of `script run` read the one function — the success result in `commands/script.py`, `_placement_evidence` in `errors.py` — and both boundary guards read the names from `gda.models`, with their guard logic unchanged. The function also folds in the `None`-report case that the two call sites each handled identically |

Round 1's **P3-2** and **P3-4** rows record where the single projection and the single name
tuple lived when round 1 created them (`gda.runner`). Those fixes stand — one projection,
one authority both guards read; round 3 moves only the home.

### Round-3 schema / help byte-diff

A refactor must change no behaviour, so the measurement here is EQUALITY, not a delta. A
detached worktree at `686a760f5` and this head each dumped `gda schema` plus `--help` for
the root, every group and every command (93 invocations), then `diff -rq`:

```
FINAL: schema + 93 help pages byte-identical to 686a760f5
```

The base → head delta reported in "Schema / help diff" and refreshed in round 2 therefore
stands unchanged.

One edit was REVERTED to hold that equality: `ScriptRunResult`'s class docstring is
PUBLISHED — it is that command's `output.description` in `gda schema` — so the pointer at
the projection's new home lives in the inline comment at the call site instead.

### Round-3 validation

Same host: this Mac, macOS (Darwin 25.6.0), Godot 4.6.3-stable, load average 2.4–3.2 at the
start of the runs. Godot suites in the foreground through the machine-wide e2e lock.

| command | result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `516 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2771 passed, 3 skipped, 1 warning in 22.48s` |
| `pytest -m e2e tests/script tests/runtime/test_e2e_launch_timeout.py tests/runtime/test_e2e_user_data.py -q -rs` | `138 passed, 318 deselected in 221.98s (0:03:41)` |
| `gda schema` + 93 `--help` pages vs `686a760f5` | byte-identical |

### Round-3 mutant checks

Applied to a `cp`-saved pristine `models.py`, `__pycache__` cleared after every edit and
every restore, restored from the COPY; `git status` empty afterwards.

| # | mutant | result |
| --- | --- | --- |
| R3-M1 | drop `log_file` from `PLACEMENT_FIELD_NAMES` at its new home | RED — 32 tests, including the placement tests of both halves. NOT red on the two boundary guards themselves: each reads the tuple as a membership / subset basis, so a SHORTER tuple still passes them. Measured rather than assumed, and it is not a regression of the move: the same mutant applied at `686a760f5`, with the tuple still in `runner.py`, is RED on exactly the same single test of those two guard files (`test_no_producer_can_emit_an_empty_evidence_object`, through the three-value unpack) |
| R3-M2 | `placement_fields` stops omitting, so a `None` path becomes a present key | RED — the failure half (`test_a_strict_failure_reports_the_engine_data_path_of_a_default_run`, `test_an_unknown_platform_data_path_is_omitted_from_a_failure_envelope`), the success half (`test_default_run_reports_the_engine_data_path_only`, `test_an_unknown_platform_data_path_is_reported_as_null_not_omitted`) and three e2e tests. This is round 1's R1-M3 re-run at the new home |
| R3-M3 | `placement_fields` emits each key with a literal `None` VALUE instead of omitting it | SURVIVES, reported rather than patched: both call sites read the mapping with `.get(name)`, which cannot tell an absent key from a `None` value, so the mutant changes no observable byte on either half. The omission rule is observable only in the form R3-M2 states, and that form is pinned |

### Not verified in this round

- Linux and Windows, unchanged from rounds 0 and 1.
- `runner.py` still spells `engine_data_path`, `resolve_user_data_root` and a `log_file`
  field as its OWN vocabulary — a function, a resolver and a dataclass attribute that
  pre-date #862 and are not the contract's key names. Renaming them is outside this PR.


qinhaihong-red:
## Reply to review round 1 (head `977c81a37` → fixes at `59ab8300b`)

| P2-1 — six prose surfaces DEFINE "the failure envelopes that report on a run" while `engine_crashed` and `stdout_spill_failed` also report on a run and carry no placement | Fully Adopt | Reproduced by the reviewer: `OS.crash()` under `--user-data-root` → `engine_crashed` with no `evidence` key while `DIR/logs/godot.log` exists | Every surface (help / schema description, catalog, SKILL, CONTEXT.md `User-data placement`, both ADR notes, `runner.py` / `models.py` comments) names the three codes instead of a category; ADR-0004's #862 note records why the two stay out: neither is on the evidence axis, `engine_crashed` is the shared `classify_launch_or_crash` verdict every launch-backed channel reaches (disclosing there would reach `export run`, `resource import`, `scene preflight` in one edit) and its `diagnostics` already carries the crash account, `stdout_spill_failed` names its own path in its message. Neither gained the placement. | Repo-wide grep for the category wording; `script run --help` re-diffed (one hunk, 13 lines) |
| P2-2 — the human rendering is guarded for presence only (mutant D green) | Fully Adopt | Reviewer's mutant D passed the fast tier at `977c81a37` | `test_the_placement_renders_as_three_labelled_lines` asserts the three exact lines | R1-M1 (mutant D) and R1-M2 (`engine data path:` printing `log_file`) RED, GREEN after restore |
| P3-1 — ADR-0031's #862 note supersedes one clause of the #850 note but not the "SUCCESS result alone" sentence | Fully Adopt | `docs/adr/0031` #850 note | The note quotes and supersedes both sentences | Doc read |
| P3-2 — the placement's string projection exists twice | Fully Adopt | `_placement_evidence` vs the success half's inline triple | `UserDataReport.as_strings()` in `runner.py` is the one projection (present facts only); both halves read it and keep their own null contract | R1-M3 (`as_strings()` stops filtering absent facts) RED on four tests across both halves; `tests/runtime/test_e2e_user_data.py` 12 passed |
| P3-3 — `user_data` defaulted to `None` on the three builders (a dropped argument is a silent revert) | Fully Adopt | Three production call sites + four test sites were the only callers (grep) | Required keyword-only on all three builders | R1-M4 (dropping `user_data=` at the timeout site) → pyright `reportCallIssue` + 12 tests RED |
| P3-4 — the three field names live in two test modules | Fully Adopt | `_PLACEMENT_FIELDS` / `_PLACEMENT_EVIDENCE_FIELDS` | Both guards read `PLACEMENT_FIELD_NAMES` from `gda.runner`, which `as_strings()` builds from | R1-M5 (renaming a key in the constant) RED on the result-model guard |
| P3-5 — CONTEXT.md "the record it answers" has no antecedent | Fully Adopt | `CONTEXT.md` `User-data placement` | "that same misdiagnosis", referent in the preceding sentence | Doc read |
| P3-6 — `_classify_ended_run` docstring one step more stale | Fully Adopt | `commands/script.py` | Corrected to the current rule (prose plus the typed `evidence` keys #687 and #862 admitted; a new envelope FIELD stays ADR-0004's decision) | Doc read |

Head `59ab8300b` (one commit on top of `977c81a37`). Gates on this Mac (macOS, Godot 4.6.3-stable): `ruff check` clean; `ruff format --check` 516 files; `pyright` 0/0/0; fast tier `2771 passed, 3 skipped`; `tests/script` e2e `126 passed`; `tests/runtime/test_e2e_launch_timeout.py` + `test_e2e_user_data.py` `12 passed`. Schema re-diff: still exactly the three optional `FailureEvidence` properties in all 77 commands' error schema plus `script run`'s description (1,216,293 → 1,290,441 bytes); help: one hunk in `script run --help`. PR CI green. Not verified: `stdout_spill_failed` on a real engine (argued from the code path, as in the finding); Linux/Windows; the `engine_data_path`-omitted arm live. Round 2 follows at this head.

qinhaihong-red:
## Reply to review round 2 (head `59ab8300b` → tidy at `686a760f5`)

Round 2 verdict: Pass, four P3 findings; all eight round-1 fixes re-verified by the reviewer against artifacts and five reverse mutants.

| P3-1 — the category wording survives in two test surfaces (`test_script_run_operation.py` section header; `test_error_registry.py` comment and test name) | Fully Adopt | Reviewer's grep at `59ab8300b` | Header names the three envelopes; the guard comment states the licensed reading and why the bare category is wrong; guard renamed `test_only_the_three_named_builders_put_the_placement_on_evidence` | Fast tier 2771 passed / 3 skipped; round-0 mutant M1 re-applied → RED on the renamed guard, GREEN after restore |
| P3-2 — the PR body's "Published contract" / "What changed" quoted round-0 text; `user_data` called optional | Fully Adopt | PR body vs current CONTEXT.md / ADR-0031 / `errors.py` | Block-quotes re-taken; `user_data` stated as required keyword-only; superseded validation row marked, not deleted | Live body read back |
| P3-3 — garden-path sentence in `FailureEvidence.log_file`'s description (all 77 error schemas) | Fully Adopt | `models.py:275-283` | Two sentences: "… the one case in which it outlives the launch. On a run gda ended, it is the file to read next." | Schema re-diff: the only round-1→round-2 change is that description string in the shared error object; help capture byte-identical to round 1 |
| P3-4 — the AST guard's disclosed limits omit its single-module scope | Fully Adopt | `test_error_registry.py` | Both limits stated (keyword-reading, `gda.errors` only) with the grep that shows the convention holds | Comment + PR body |

Head `686a760f5` (one docs commit on top of `59ab8300b`). Gates on this Mac: `ruff check` clean; `ruff format --check` 516 files; `pyright` 0/0/0; fast tier `2771 passed, 3 skipped`. No e2e re-run for a description-only change; round 1's `tests/script` 126 and `tests/runtime` 12 stand for the code. Base → head schema delta unchanged in kind (1,216,293 → 1,290,672 bytes; three optional properties in every command's error schema + `script run`'s description); help one hunk. The maintainer's independent re-verification follows before the merge.

qinhaihong-red:
## Maintainer re-verification — READY at `686a760f5`

Independent run on a detached checkout of the pinned head (this Mac, macOS, Godot 4.6.3-stable, host load ~4.9 with two sibling agents active):

| Check | Result |
| --- | --- |
| Commits base..head | five, all `haihong.qin <haihongqin@gmail.com>`, Conventional with `(#862)`, no trailers |
| PR body | `Closes #862`; no attribution; "Review round 1" and "Review round 2" sections present |
| File scope | 17 files; no file owned by a sibling W4 slice; `script_errors.py` untouched |
| `ruff check .` | All checks passed |
| `ruff format --check .` | 516 files already formatted |
| `pytest -m "not e2e" -n 4 --dist loadgroup` | 2771 passed, 3 skipped |
| `pytest -m e2e tests/script tests/runtime/test_e2e_launch_timeout.py tests/runtime/test_e2e_user_data.py` (machine-wide lock) | 138 passed in 3:47 |

Two independent review rounds closed (round 1: 0/2/6 all adopted; round 2: Pass, 0/0/4 all adopted). Merges in W4 order (after #981) on the wave's authorization; the wave's full e2e dispatch runs on the dev ref after the last merge.

qinhaihong-red:
Independent adversarial review of PR #980 at 686a760f548ee69e9ca1ae19f7d6a939ff54dfd4 against merge base 86dae1f84b522ea234339926a2c5b6dbbff908c5. I reviewed the issue and exact diff from first principles; earlier PR review conclusions were not evidence.

## Artifact and issue — Changes required

1. **[P2] Correct the accepted ADR's evidence rationale.** ADR-0004:406-410 says engine_crashed fails the second evidence-admission clause because diagnostics carries the engine's crash account. The actual rule at ADR-0004:228-244 asks whether the proposed fact, here user-data placement, is recoverable without parsing prose. errors.py:551-559 passes only raw.stderr into that verdict; it need not contain a placement or even a crash account. A constructed crash result with user_data present and stderr empty produced engine_crashed with diagnostics empty. The one-channel/unchanged-producer boundary can remain; remove this false clause argument or replace it with a reason that addresses the placement facts.

2. **[P3] Correct the evidence citation.** ADR-0004:381-384 attributes both the failed-write misdiagnosis and three later 120-second timeouts to ADR-0031's #850 note. That note records the former; issue #862's dogfooding source records the timeout history. Cite #862 for that half.

**Terminology and prose assessment.** Checked every changed Markdown passage and the relevant glossary, accepted ADRs, catalog, bundled skill, and issue #862 against project terms and ASD-STE100 (technical word choice, sentence scope, causal and dependency claims). No further substantive terminology or prose issue was found. No independent run of every command's published help/schema diff was performed.

## Architecture — Minor

**[P3] Keep public field names at the contract boundary.** runner.py:96-101,135-159 now defines the JSON names of ScriptRunResult and FailureEvidence and describes their different null behavior. That shared launch primitive previously held only placement Path/lifetime facts; this public contract change also altered the lower runner and the unchanged success path (commands/script.py:1746-1769). ADR-0040 assigns public error models to the shared contract core and the script result to its command group. Keep UserDataReport as the raw fact owner, and put the shared string/key projection at a contract-level boundary. The three named failure builders can retain the same disclosure rule.

## Standards — Pass

Zero documented-standard violations and zero actionable Fowler baseline smells in the 17-file diff. The three failure builders remain in errors.py as ADR-0040 requires, and their explicit keyword fields are a bounded part of the existing guard.

## Spec — Pass

Zero missing, extra, or incorrect requirements found against #862. The three named builders read the same Raw run; other failure and channel paths retain their prior envelopes. Schema, help, catalog, bundled skill, glossary, and ADR updates cover the issue's published-contract requirements.

Exact-head validation: 205 focused unit tests passed locally; another focused run including models, schema, descriptor guards, and five real-Godot cases (default/root strict, root timeout, root abort, never-ran) passed 178 tests. git diff --check passed.

Axis totals: Standards 0 (none); Spec 0 (none).

qinhaihong-red:
## Reply to the external review (head `686a760f5` → fixes at `cea4a1fd0`)

| Artifact P2 — ADR-0004's #862 note argues `engine_crashed`'s exclusion with the second admission clause ("its `diagnostics` already carries the engine's crash account"), while that clause is about the proposed fact — the placement — and the crash classifier sees only `raw.stderr` | Fully Adopt | `errors.py`'s `engine_crashed` branch passes `raw.stderr` alone; a crash with empty stderr yields empty diagnostics; the placement is not in either | `070bdcc9e` — the note (this PR's own, edited in place) now gives the third-clause reason: an unusable placement is refused BEFORE the spawn by the same shared classifier (`user_data_unwritable`), so a signal death is not the placement's doing and naming it would not change the caller's next move; the shared-verdict point and the one-channel / unchanged-producer boundary are kept | Read |
| Artifact P3 — the note cites ADR-0031's #850 note for both the failed-write misdiagnosis and the three 120-second timeouts | Fully Adopt | The #850 note records the misdiagnosis; the timeout history is #862's dogfooding source | `070bdcc9e` — the two halves cite their own records | Read |
| Architecture P3 — `runner.py` defines the public key names (`PLACEMENT_FIELD_NAMES`, `UserDataReport.as_strings`) and describes the two public models' null behaviour, inverting ADR-0040's layering | Fully Adopt | The launch primitive named `ScriptRunResult` / `FailureEvidence` and their key strings | `cea4a1fd0` — the names tuple and ONE projection `placement_fields(report)` live in `gda.models` beside `FailureEvidence`'s three fields (the record imported under `TYPE_CHECKING` only); `UserDataReport` stays in `runner.py` as the raw fact owner; both halves of `script run` and both boundary guards read the one authority; `runner.py` names no contract key (grep) | Schema + all help pages byte-identical to `686a760f5` (measured both sides, `diff -rq` clean); mutants: a name dropped from the tuple → 32 tests RED (the two boundary guards are membership-based and pass a shorter tuple at either head, measured); a `None` path published as a key → RED on both halves plus 3 e2e; a key with a literal `None` value → survives, behaviourally equivalent because both call sites read `.get(name)` (disclosed, not patched) |

Head `cea4a1fd0` (two commits on top of `686a760f5`, append-only so #976's stack base is intact; all `haihong.qin`). Gates on this Mac (macOS, Godot 4.6.3-stable): `ruff check` clean; `ruff format --check` 516 files; `pyright` 0/0/0; fast tier `2771 passed, 3 skipped`; under the machine-wide lock `tests/script` + `tests/runtime/test_e2e_launch_timeout.py` + `tests/runtime/test_e2e_user_data.py` `138 passed`. One edit was reverted to hold the byte-identical schema: `ScriptRunResult`'s class docstring is published as the command's `output.description`, so the pointer to the projection's new home lives in the call-site comment. Not changed: `runner.py`'s own pre-#862 vocabulary (`engine_data_path()` the function, `resolve_user_data_root`, the `log_file` attribute) — the launch's facts, not the contract's keys. The maintainer's re-verification on a detached checkout at this head follows.

qinhaihong-red:
## Maintainer re-verification — READY at `cea4a1fd0`

Independent run on a detached checkout of the pinned head (this Mac, macOS, Godot 4.6.3-stable, load 2.0–3.5):

| Check | Result |
| --- | --- |
| Commits base..head | seven, all `haihong.qin <haihongqin@gmail.com>`, Conventional with `(#862)`, no trailers; the two round-3 commits append to `686a760f5`, so #976's stack base is intact |
| PR body | `Closes #862`; no attribution; "Review round 1", "Review round 2" and "Review round 3 (external)" sections present |
| File scope (round 3) | 7 files; `runner.py` names no contract key (grep); `placement_fields` + `PLACEMENT_FIELD_NAMES` in `models.py`, the record imported under `TYPE_CHECKING` |
| `ruff check .` | All checks passed |
| `ruff format --check .` | 516 files already formatted |
| `pytest -m "not e2e" -n 4 --dist loadgroup` | 2771 passed, 3 skipped |
| `pytest -m e2e tests/script tests/runtime/test_e2e_launch_timeout.py tests/runtime/test_e2e_user_data.py` (machine-wide lock) | 138 passed in 3:43 |
| `gda schema` at `cea4a1fd0` vs `686a760f5` | byte-identical (`cmp` clean) |
| PR CI at this head | all jobs green, MERGEABLE / CLEAN |

Three review rounds closed (round 1: 0/2/6; round 2: Pass 0/0/4; external: 0/1/2 — all adopted). Merges second in the W4 order on the wave's authorization.

qinhaihong-red:
## Exact-head recheck — one ADR rationale remains (`cea4a1fd01b6ac543ab03475d23c6035f43f1b57`)

The timeout-history citation and placement-field ownership findings are closed. The public projection now lives in `models.py`; focused tests passed, and `gda schema` is byte-identical to the prior head.

ADR-0004:407-415 still says the *same classifier* refuses unusable placement before spawning and concludes that a signal death therefore cannot be caused by placement. The pre-spawn probe is in `runner.user_data_placement()`, not `classify_launch_or_crash`. Without an explicit root, it deliberately does not probe the engine's `user://` data directory; with a root, its write probe only establishes the state at that moment. The universal causal conclusion and the claimed failure of admission clause 3 are not established.

Keep the deliberate scope of three named, existing evidence builders, but correct this rationale without adding a new producer or changing the public schema. Focused exact-head validation: 287 tests passed; `gda schema` matched the previous head byte for byte. I will make this documentation correction on this PR branch.

qinhaihong-red:
## Resolution of the exact-head recheck (`5a250e107f01cec19d7f3c50721cf84213f6e446`)

**Fully adopted the remaining ADR rationale finding.** ADR-0004 now attributes the pre-spawn checks to `runner.user_data_placement`, states their default-root and point-in-time limits, and removes the unsupported conclusion that placement cannot contribute to a later signal death or necessarily fails admission clause 3. It retains #862's three named existing evidence builders; `engine_crashed` remains a shared verdict with no new evidence producer. The PR description's old round-3 claim was corrected in place.

Validation on this head: 167 focused CLI/script unit tests passed; `gda schema` is byte-identical to `cea4a1fd0`; `git diff --check` passed. Hosted CI is green, with the Godot e2e job skipped by its CI policy. This commit changes documentation only.

qinhaihong-red:
## Maintainer re-verification — READY at `5a250e107`

The exact-head recheck is right, and it corrects a claim the maintainer's own round-3 reply made: the pre-spawn checks live in `runner.user_data_placement`, not in `classify_launch_or_crash`; without an explicit root they do not probe the engine's default `user://` directory, and with one they establish the state at that moment only. So "a signal death cannot be the placement's doing" was not established, and the ADR-0004 note no longer says it. What the note keeps is what the decision rests on: the three named producers, `engine_crashed` as a shared verdict whose evidence would need its own admission and channel-scope decision, and `stdout_spill_failed` failing the third clause on its own message. The round-3 reply's third-clause sentence for `engine_crashed` is superseded by this note.

Independent run on a detached checkout of the pinned head (this Mac, macOS, Godot 4.6.3-stable):

| Check | Result |
| --- | --- |
| Commits base..head | eight, all `haihong.qin <haihongqin@gmail.com>`, Conventional with `(#862)`, no trailers; append-only, so #976's stack base `686a760f5` is intact |
| Diff since `cea4a1fd0` | one file, `docs/adr/0004-schema-flag-self-description.md`, the #862 note edited in place; no source, test or schema change |
| Stale-claim sweep (`placement's doing`, `refuses the launch before the spawn`) over `docs/`, `CONTEXT.md`, `src/`, `README.md` | no hit |
| `ruff check .` / `ruff format --check .` / `pyright` | clean / 516 files / 0 errors, 0 warnings |
| `pytest -m "not e2e" -n 4 --dist loadgroup` | 2771 passed, 3 skipped |
| `gda schema` at `5a250e107` vs `686a760f5` | byte-identical (`cmp` clean); the e2e set verified at `cea4a1fd0` (138 passed) covers this head, since no code changed |
| PR CI at this head | all jobs green, MERGEABLE / CLEAN |

Four review rounds closed. Second in the W4 merge order on the wave's authorization.

Issue #804, opened 2026-09-02
the res:// walk descends into a nested project and .gdignore directories the engine skips: script validate --all compiles inner scripts against the outer root


qinhaihong-red:
## Parent

Drafted during PR #799's review round (reproduced verbatim in that PR's "Follow-ups" section);
filed on user triage after the P4 promotion (#801).

## What to build

Align gda's shared `res://` traversal with the directories the engine's own scan skips.
`EditorFileSystem::_should_skip_directory` (Godot 4.6.3) skips a directory that holds a
`project.godot` (a NESTED project) and a directory that holds a `.gdignore` marker. gda's
walk — the one traversal every collector shares — descends into both.

The consequence is reproduced and stated as a known gap in ADR-0006/catalog/SKILL since
#799: `script validate --all` enumerates a nested project's scripts and compiles them
against the OUTER root, so the same file gets opposite answers depending on the selector
(`--all` vs naming it, which #697's ownership rule refuses). `script list`, `scene list`,
project analysis, and the import-gap listing likewise report files the engine's scan does
not see.

One shared change in the walk's descent rule, wide blast radius: every collector's universe
changes together, so this is its own slice with its own regressions. Do not copy PR #799's
original draft clause "`_should_descend` excludes only `res://.godot`" — stale since #795,
which added cache-alias and descent-cycle exclusions by filesystem identity; the walk's
exclusion set is ADR-0032's to state and must be re-read at implementation time.

## Acceptance criteria

- [ ] A directory holding `project.godot` (other than the root) and a directory holding `.gdignore` are not descended, matching the engine's scan; the rule lives in the ONE shared traversal, not per collector.
- [ ] `script validate --all` no longer compiles a nested project's scripts against the outer root — the `--all` and named-target answers agree (e2e regression with a nested-project fixture, red-proofed on the pre-fix walk).
- [ ] `script list`, `scene list`, project analysis, the import-gap listing, AND the ADR-0032 `class_name` index agree with the engine's universe on the same fixture — the index consequence is deliberate: a vendored sub-project's `class_name`s leave the index, so `node add --class-name` / `resource create` / `find-references` change answers there, and an `ambiguous_class_name` caused only by a vendored duplicate becomes unambiguous.
[...]

PR #981, opened 2026-09-15
fix(gda): export run reports the project-tree mutations of the native export (#839)

### LINKED ISSUE #839: export run: report every project-tree mutation the native export causes (cache, sidecars, rewritten files)
## Dogfooding source

- **GDA-DF-067** (2026-08-26, gda 0.11.0; reconfirmed on four later candidates with 0.11.0 and 0.12.0) — `gda export run --preset macOS --mode release --json` returned the `.app` with `warnings: []` while the native editor import rewrote two to four tracked binary `.translation` resources and created about 300 `.uid` files, 4,700–4,900 `.import` files, and a `.godot/` tree of 9,500–10,000 files (about 1.1 GiB). None of it appeared in the export result.
- **PIPE-DF-174** (2026-09-03, gda 0.12.0) — same shape. A partial cleanup that moved `.godot/` while leaving the sidecars active then made a later focused test emit about 96 MiB of engine log against the broken `res://.godot/imported/...` references.

Verified against gda 0.13.0 (`266c8009f`): `ExportRunResult` carries `preset`, `platform`, `mode`, `output_path`, `created_dirs` (only the output parents gda itself created), and `warnings` parsed from stderr; nothing observes the project tree — the one project read on that path is the harness snapshot's fixed file list (ADR-0028). The native `--export-*` path performs the editor import pass (the #486 triage established that it is what imports assets), so every export on a checkout without a warm cache creates the cache and the source-adjacent sidecars and rewrites generated resources. `resource import` (#668) already classifies every file its pass creates as `cache_owned` / `source_adjacent`; the export result does not.

## What to build

The export result reports what the export did to the project tree, so an agent can review, stage, or restore it without a manual git snapshot:

- Take a pre/post inventory of the project tree (path, size, mtime, and a content hash; the output artifact and its parents excluded) around the native export. The pre-export pass hashes every pre-existing file outside the classification root and the output path, because a rewritten file's earlier bytes exist only then; the post-export pass hashes only the candidates the next bullet names. That pre-export hashing is the cost the cost-bound criterion measures. Report `created` (each entry classified `cache_owned` / `source_adjacent`, the same vocabulary as `resource import`) and `modified`, with counts and total bytes.
- `modified` means the CONTENT changed: candidates are pre-existing files whose size or mtime changed, and a candidate enters `modified` only when its content hash differs from the pre-export hash. A changed mtime alone is not a modification — the import pass touches files it does not rewrite, and that noise must not bury the rewritten `.translation` resources the record is about.
- The result states the classification root (`.godot/`) so `created` can be cleaned as one unit. The root is the engine default, the constant `gda.import_evidence.CACHE_ROOT_REL` the reused classifier reads; a project that sets `application/config/use_hidden_project_data_directory=false` (cache under `godot/`) is out of scope for this slice — its cache files read as `source_adjacent` — and gets no follow-up unless a dogfooding case brings one. The report is disclosure: the export deletes and restores nothing.
- `created` is classified through `gda.import_evidence.classify_created_file` (`CreatedFileClass`, lifted by #741 in W2); no classification rule is restated on the export path. `resource import`'s own pre/post walker (`_project_files` / `_summarize`) stays in that command per Open item 9's decision on #741 — the shared part is the classification, not the walk.
- Human rendering summarizes the counts. `warnings` keeps its current meaning (engine advisories) and is not overloaded.

## Acceptance criteria

- [ ] Exporting a project whose `.godot/` cache is absent reports the created cache and sidecar files, classified, with counts and bytes.
- [ ] A second export on the warm cache reports zero `created` entries, or the PR names the exact set the engine recreates on every pass and the test asserts that set.
- [ ] A generated resource the import pass rewrites appears under `modified`; a file the pass only touches (mtime changed, content equal) does not.
- [ ] The output artifact and, for a directory artifact, everything under it are excluded from both lists. Files beside the artifact are reported whether its parent already existed or gda created it; `created_dirs` contains directory paths, not file mutations.
- [ ] `created` entries are classified by `gda.import_evidence.classify_created_file` and the reported root is `CACHE_ROOT_REL`; the export path restates no classification rule. The `use_hidden_project_data_directory=false` case is declared out of scope in the PR body, not handled.
- [ ] Existing consumers see only additive fields; `gda schema` and `--help` changes are limited to the new fields.
- [ ] Cost bound: measured on a GENERATED fixture of about 10,000 files (built by a test or script, never checked in), the two inventories together, pre-export hashing included, add at most 10 % to the warm macOS release export's own wall clock. The measured `--mode pack` case is an accepted exception at +16.2 % (+0.4 s absolute). The PR states both measurements; a different bound needs its own measurement and rationale.
- [ ] A failed native export carries no project-tree mutation report, even if it changed files before failing. A later run reports only changes after its own pre-export baseline.
- [ ] An e2e export on a fixture with one imported texture asserts the created and modified sets.

## Amendments

- **2026-09-15, W4 artifact review:** the inventory now carries a content hash and states which files are hashed pre-export (the earlier text asked `modified` to compare against a pre-export hash the inventory never computed); the classification root is bounded to the engine default `.godot/` with the `godot/` case declared out of scope; the reuse of `gda.import_evidence.classify_created_file` / `CACHE_ROOT_REL` is an acceptance criterion, not only the 2026-09-05 comment (whose `commands/resource.py:576` was the pre-#741 home — the rule lives at `src/gda/import_evidence.py` since W2); the cost bound names a generated fixture and a number; the title says "project-tree mutation" so the report has one name across this issue, the code comments, and #979.
- **2026-09-15, PR #981 review round 1:** the cost bound is read against the export the dogfooding record used — the real macOS release export (measured +1.8 % warm on a generated 10,005-file fixture); the cheapest export, `--mode pack`, measured +16.2 % (+0.4 s absolute) and is accepted as proportionate with the numbers disclosed in the PR. Cache-root rewrites are NOT reported under `modified` (the pre-hash scope excludes the classification root by design; a warm export rewrites `.godot/editor/filesystem_cache10` on every pass) — the published descriptions must say so. `skipped` stays a count: a bounded path list would reopen the truncation contract for a disclosure whose remedy is "fix permissions and rerun". An unreadable DIRECTORY counts as skipped too. `CONTEXT.md` gains the term `Project-tree mutation report`.

- **2026-09-16, PR #981 external review (acceptance criteria made executable):** the cost-bound criterion reads: measured on the warm macOS RELEASE export of the generated ~10,000-file fixture, the two inventories add at most 10 % to the export's own wall clock (measured +1.8 %); the cheapest export, `--mode pack`, is the accepted exception at +16.2 % (+0.4 s absolute), disclosed in the PR. A new criterion states the success-only boundary: a native export that FAILS carries no report even when it mutated the tree before failing — the failure answers through the `Error envelope`, and a later successful run cannot recover the failed run's mutations because it takes a new baseline; the caller must inspect the project state separately. AC4's exclusion is read literally — the artifact and anything under it; a directory in `created_dirs` is not a file and needs no exclusion, so a file the export writes beside the artifact is reported whether or not gda created the parent — and it applies to a `res://` output path resolved against the project. The walk follows directory links as the engine reads them, once each by filesystem identity (ADR-0032's #760 policy), so content reached through a link is inventoried. The report is the difference between gda's two walks under the `Concurrent external editor` assumption (ADR-0018); that assumption is disclosed where the report is described.

## Blocked by

None - can start immediately. #741 and #840 were the gates the comment and the wave plan named; both merged in W2, so the field stays as it is.

qinhaihong-red:
## Summary

`gda export run` now reports what the native export did to the project tree. The
export performs the editor import pass, so an export against a cold cache creates
the whole `.godot/` cache plus the `.import` and `.uid` sidecars beside the
sources, and a stale asset makes it rewrite generated resources that are tracked
in git. None of that was observable in the result before — in GDA-DF-067 about
14,000 new files and two to four rewritten `.translation` resources appeared on
disk while `warnings` stayed empty.

The result carries one new, additive field, `project_tree_mutations`:

- `created`: every file the tree gained ANYWHERE under the project — directory
  links walked as the engine reads them, once each — classified
  `cache_owned` / `source_adjacent` by
  `gda.import_evidence.classify_created_file` — the same function and the same
  vocabulary `resource import` reports — against the reported `cache_root`, which
  is `res://` + `gda.import_evidence.CACHE_ROOT_REL`. No classification rule is
  restated on the export path; the shared classifier is also what decides which
  files the pre-export walk hashes, and both call sites are pinned by a test.
- `modified`: every pre-existing file OUTSIDE the cache root whose CONTENT
  changed. A pre-existing file is a CANDIDATE only when its size or timestamp
  moved, and enters `modified` only when its digest then differs from the
  pre-export one, so a rewrite that preserves both is not seen. Rewrites INSIDE
  the cache root are not reported at all — a warm export rewrites its own
  bookkeeping there on every run — so an empty `modified` says nothing about the
  cache. The import pass touches far more files than it rewrites; a timestamp
  rule would bury the few rewrites the record is about.
- counts and total bytes for both, plus `skipped` — what neither walk could
  account for: an entry that is not a regular file (a FIFO, a socket, a device;
  gda never opens one), a file it could not read, or a directory whose whole
  subtree is then uncovered.

The report is the difference between gda's walk before the export and its walk
after; gda assumes it is the project's sole driver during the export (ADR-0018),
so a change another writer makes in that interval is attributed to the export.
The report is disclosure: the export deletes and restores nothing. A failed
export carries no report; the failure answers through the error envelope.

Closes #839

## What changed

**`src/gda/commands/export.py`** (the whole slice; ADR-0040 keeps a group's
models, renderers and operation in its group module)

- `ProjectTreeMutations` + `ExportCreatedFile` + `ExportModifiedFile`, and the
  additive `ExportRunResult.project_tree_mutations` field. A `model_validator`
  pins the counts against the lists (the `resource import` idiom).
- The exclusion is the artifact and its OWN subtree. A `res://` output path
  addresses the project tree; both walks identify the artifact by its parent
  directory's filesystem identity and its name. A directory link or another
  project-relative spelling cannot make the artifact appear in `created`.
  `created_dirs` holds directory paths, while both walks report files.
- `_walk_project` follows directory links and identifies each directory by
  filesystem identity (`st_dev`, `st_ino`), so a linked-in library is walked as
  the engine reads it, once each, and a cycle ends by rule (round 3, ADR-0032's
  #760 policy).
- The two walks: `_PreExportInventory.capture()` before the export,
  `.settle()` after it. `run_export_operation` captures OUTSIDE the harness
  snapshot region, so gda's own strip/restore (ADR-0028, byte-for-byte) is not
  reported as a mutation; `classify_export_run` settles on the SUCCESS branch
  only, so a failed export pays for no second walk. The walk passes an
  `os.walk(onerror=…)` so a directory it cannot list is disclosed instead of
  silently dropping its subtree, and the inventory reads REGULAR files only —
  it stats before it opens, so a FIFO is counted rather than blocking the
  command for ever.
- `render_export_run` appends one summary line — counts only, and
  `project tree: unchanged outside res://.godot` when nothing the report covers
  changed, because that is the scope it can vouch for.
- `gda export run --help` gains a paragraph for the new field.

**`src/gda/import_evidence.py`** — comments only: `CACHE_ROOT_REL`'s consumer note
now says what #839 does with it and uses the report's one name ("project-tree
mutation report"); `classify_created_file`'s docstring no longer names a single
caller. No logic change; the classifier needed none.

**`CONTEXT.md`** — one glossary entry, `Project-tree mutation report`, under
Operations beside `Import evidence`.

**`src/gda/skill/SKILL.md`** — one clause on the existing `export` row, the row
that already carries #840's two result fields.

**Tests** — `tests/export/test_export_tree_mutations.py` (new, 26 tests) drives the
real recipe with an export runner that mutates the project the way the pass does;
`tests/export/test_e2e_export_run.py` gains two real-engine tests (the report's
shape, and the mutations under a linked directory); three existing
assertions over the full result JSON gained the new key
(`tests/cli/test_models.py`, `tests/export/test_export_run_commands.py`).

**Docs** — `docs/command-catalog.md` export section; the README `export run` row
and its three translations (markers restamped with
`scripts/update_readme_i18n.py`).

## Schema / help diff

Baseline: a detached worktree at the fork point `86dae1f84`, `gda schema` plus
`--help` for the root, all 15 groups and all 77 commands (93 help dumps,
enumerated from the schema itself), diffed against this head. Re-measured after
round 3, which adds description text and nothing structural. Every difference:

1. `gda export run --help`: one added paragraph in the command description (the
   new field's contract). No option, argument or usage line changed. The other
   92 help dumps are byte-identical.
2. `gda schema`, command `export run`:
   - `description` (the command's): the same added paragraph.
   - `output.properties`: one added — `project_tree_mutations`. Nothing removed,
     nothing changed.
   - `output.$defs`: three added — `ProjectTreeMutations`, `ExportCreatedFile`,
     `ExportModifiedFile`.
   - `output.description`: the `ExportRunResult` docstring gains the paragraph
     naming the new field and stating that `warnings` keeps its meaning.
   - `output.required`: UNCHANGED (`['preset','platform','mode','output_path',
     'created_dirs']` on both sides) — the field has a default, so an existing
     consumer sees an addition and nothing else.
   - `input`: byte-identical.

No other command's `description`, `input`, `output` or `error` differs; no command
was added or removed. Round 3 moved only the TEXT inside those same nodes: the
`project_tree_mutations` field description gains the sole-driver sentence, the
`created` field description gains the directory-link clause, `ProjectTreeMutations`'s
own description carries the one exclusion rule plus the corrected GDA-DF-067
sentence, and the command description carries all three.

## Guards: PR CI vs wave e2e

- **PR CI** runs: `ruff check` + `ruff format --check`, `pyright` (root + the
  gda-balancing member), the harness-version guard, and
  `pytest -m "not e2e" -n 4 --dist loadgroup`. That covers all 26 new unit tests
  and the three updated result-shape assertions.
- **Only in the wave's e2e dispatch** (`workflow_dispatch` with `run-e2e`, or the
  nightly schedule): `tests/export/test_e2e_export_run.py`, including
  `test_export_run_reports_what_the_native_export_did_to_the_project` and
  `test_export_run_reports_the_mutations_under_a_linked_directory`. PR CI
  spawns no Godot, so the real-engine proof of this slice is dispatch-only and was
  run locally (below).

## Validation

Host: this Mac (macOS, Darwin 25.6.0), Godot 4.6.3-stable, engine-side claims from
the `4.6.3-stable` tag. NOT a quiet host: `uptime` load averages read 1.95 – 3.10
through the round. Each e2e invocation ran in the foreground under the
machine-wide Godot lock. Figures below are from this head (`b2252c12f`).

| Command | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `517 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2784 passed, 3 skipped, 1 warning in 22.51s` (the warning is the pre-existing `os.fork()` DeprecationWarning in `tests/daemon`) |
| `pytest -m e2e tests/export -q -rs` (under the lock) | `21 passed, 1 skipped, 95 deselected in 42.84s` (the skip is the pre-existing Linux-only `XDG_DATA_HOME` template test) |
| `pytest -m e2e tests/resource -q -rs` (under the lock) | `43 passed, 115 deselected in 124.69s` — `resource import`'s own walker is untouched by this PR, and its real-engine suite says so |
| `pytest tests/repo/test_readme_i18n_sync.py -q` | `7 passed` |

Fixture: generated by a script under a temp dir, never checked in — **10,005
files, 25.6 MiB** of content outside the cache (100 directories × 100 files of
256 B … 8 KiB, plus a PNG, a translation CSV, a script and the presets). The
non-cache files are what the pre-export pass hashes, so that is the number that
matters.

**Cost measurement conditions.** Same Mac, Godot 4.6.3-stable. NOT quiet — other
agents were running Godot e2e suites throughout (that run waited 128 s for the
machine-wide e2e lock); `uptime` load averages were 1.83 before and 3.27 after.
Each figure is the median of repeated, interleaved base/head runs (base = a
detached worktree at the fork point `86dae1f84`, which has no report). Measured at
`a82409164`; round 1 adds one callback per directory the walk cannot list, which
is zero in a healthy tree, so the figures stand.

**Round 3 did not move the walk's cost.** The link-following walk adds one
`stat` per directory reached and sorts each directory's entry names; it follows
no link on a fixture that has none. Measured in-process on a generated control
fixture (`/tmp`, 10,001 files, 40.3 MiB, 100 directories x 100 files), median of
three runs, the two walks interleaved on the same host minute:

```
round-2 walk (no followlinks, no identity):  capture 0.282s  settle 0.066s
this head:                                   capture 0.284s  settle 0.066s
```

So the export measurements below stand and were not re-run.

**The AC's measurement — a real `gda export run --preset macOS --mode release`:**

| Stage | base (no report) | head (report) | delta |
| --- | --- | --- | --- |
| cold cache (2 runs, fresh copies) | 13.79 s | 13.12 s | **-4.9 %** (-0.67 s) |
| warm cache (5 runs each) | 10.23 s | 10.42 s | **+1.8 %** (+0.18 s) |

Both are inside the 10 % bound. The run-to-run spread (±1 s) is larger than the
report's own cost, which is why the cold pair reads negative; for a figure free of
that noise, the two walks were also timed in-process on the same fixture:

```
capture 0.335s  settle 0.079s  files 10021      # first run in a fresh process
capture 0.276s  settle 0.069s  files 10021
walk only 0.017s for 10021 files
stat only 0.049s
stat+hash 0.222s                                 # 25.6 MiB, blake2b
```

So the two walks cost **0.35 – 0.41 s** on a 10,005-file project — about 3 % of the
release export above. A control confirms the delta is the walks and not startup:
`gda export get`, which runs no inventory, measured 0.54 s on base and 0.53 s on
head over 5 interleaved runs each.

**The worst case, stated rather than hidden.** `--mode pack` on the same fixture is
the cheapest export there is (no templates, no bundle), so the same 0.6 s is a much
larger fraction: warm base 3.95 s → head 4.59 s, **+16.2 %** (5 runs each,
interleaved). The cold pack pair was too noisy to separate (base 9.00/8.05 s, head
12.48/7.54 s). The bound therefore holds for the export the criterion names, and a
caller who runs repeated `--mode pack` exports over a 10,000-file tree pays about
half a second per export for the record. No further optimization was pursued: the
cost is per-file syscall overhead, not hashing — four digest implementations
(streaming, whole-file read, `read_bytes`, `hashlib.file_digest`) measured within
0.02 s of each other on the same 10,021 files.

## Mutant checks

Every rule test was re-run against a mutant that reverses its rule; each one went
RED, and each source file was restored from a SAVED COPY (`cp`, never
`git checkout --`) with `__pycache__` dropped after every write before the test was
confirmed GREEN again. 21 mutants over 22 tests — the classifier test carries two,
one per call site; three tests (the deletion boundary, the default report, and
the additive-field round-trip) pin boundaries rather than reversible rules.

| Test | Mutant applied | Verdict |
| --- | --- | --- |
| `…classified_against_the_shared_cache_root` | report `res://godot` instead of `res://` + `CACHE_ROOT_REL` | RED → restored GREEN |
| `…is_the_shared_function_not_a_local_rule` | settle site: classify with a local `startswith(".godot/")` | RED → restored GREEN |
| `…is_the_shared_function_not_a_local_rule` | capture site: gate the hash on a local `startswith(".godot/")` | RED → restored GREEN |
| `…is_modified_and_carries_both_sizes` | report the post-export size as `size_before` too | RED → restored GREEN |
| `…whose_content_is_equal_is_not_modified` | report a candidate without comparing its digest | RED → restored GREEN |
| `…keeps_the_size_and_the_timestamp_is_not_a_candidate` | hash every pre-existing file (drop the size/mtime gate) | RED → restored GREEN |
| `…cache_file_is_never_reported_as_rewritten` | compare a pre-existing cache file like any other | RED → restored GREEN |
| `…created_dirs_are_not_mutations` | exclude nothing but `.git` (re-checked against round 3's rewritten `_excluded_prefixes`) | RED → restored GREEN |
| `…res_output_artifact_is_the_output_not_a_mutation` | drop the `res://` resolution, so every `://` path is dropped as before round 3 | RED → restored GREEN |
| `…beside_the_artifact_is_reported_in_a_gda_created_parent` | put `created_dirs` back into the exclusion prefixes | RED → restored GREEN |
| `…directory_link_is_walked_as_the_engine_reads_it` | `followlinks=False` | RED → restored GREEN |
| `…link_that_leads_back_up_the_chain_is_not_re_entered` | drop the identity set | RED — the run terminates (macOS `PATH_MAX` ends it) and reports the loop's spellings, so the test reads red rather than hanging → restored GREEN |
| `…git_directory_is_not_walked` | walk the top-level `.git` | RED → restored GREEN |
| `…cannot_read_is_skipped_not_failed` | report a path the pre-export walk could not read as created | RED → restored GREEN |
| `…directory_the_walk_cannot_list_is_counted_not_ignored` | drop `os.walk`'s `onerror`, as before round 1 | RED → restored GREEN |
| `…under_a_locked_directory_is_not_announced_as_created` | let the settlement walk into a directory the first walk could not list | RED → restored GREEN |
| `…deleted_file_is_reported_nowhere` | none — see below | not mutant-checked |
| `…strip_and_restore_is_not_a_mutation` | capture the pre-export walk INSIDE the harness-stripped region | RED → restored GREEN |
| `…pays_for_no_second_walk` | settle before the success branch is decided | RED → restored GREEN |
| `…render_summarizes_the_counts` | never take the counted branch | RED → restored GREEN |
| `…quiet_render_names_the_scope_it_vouches_for` | print the bare word "unchanged" | RED → restored GREEN |
| `…unreadable_file_is_named_in_the_render` | leave the unreadable count off the line | RED → restored GREEN |
| `…counts_must_match_the_reported_lists` | drop the count/list check | RED → restored GREEN |
| `…non_regular_entry_is_counted_and_never_opened` | drop the `S_ISREG` check, so a FIFO is opened | RED (fails on its deadline, does not hang) → restored GREEN |
| `…same_size_rewrite_with_a_newer_timestamp_is_reported` | compare on size alone (drop the timestamp half) | RED → restored GREEN |
| `…exclusions_match_whole_path_components` | match a string prefix instead of a path component | RED → restored GREEN |

The one e2e mutant of round 3, run against the real engine under the lock:
`followlinks=False` makes `test_export_run_reports_the_mutations_under_a_linked_directory`
fail with `res://assets/sprite.png.import` absent from a `created` list holding
only cache files — the exact shape the reviewer measured on this branch's round-2
head. Restored and re-run GREEN.

Three findings came out of the earlier rounds, all fixed:

1. `…cannot_read_is_skipped_not_failed` originally proved nothing. A dangling
   symlink is unreadable in BOTH walks, so the settlement reached the same
   verdict with or without its guard. The test now also carries a link whose
   TARGET the export creates — unreadable before, readable after — which is the
   case the guard exists for, and it goes red without it.
2. A same-LENGTH mutant written in the same second as the previous write left
   CPython's `(mtime-seconds, size)` bytecode check satisfied, so the mutant never
   ran and read as "the test does not pin the rule". The harness now drops
   `__pycache__` after every write.
3. Round 1 (review finding P2-4): the classifier's CAPTURE-site call was pinned by
   nothing — the hash gate is behaviourally invisible, so only a spy can see it.
   The stub now records what it was asked about and the export runner snapshots
   that record before mutating, so whatever is in the snapshot was asked during
   the first walk.

`…deleted_file_is_reported_nowhere` pins a BOUNDARY (the report covers what the
pass adds and rewrites), not a rule with a reverse: any "mutant" for it would have
to add a deletions list the slice deliberately does not have. It is stated here
rather than given a contrived mutant.

## Deviations from the issue

None in scope or contract. Two judgment calls worth naming:

- `use_hidden_project_data_directory=false` is **declared out of scope**, as the
  issue requires: such a project keeps its cache under `godot/`, whose files then
  read as `source_adjacent`. Nothing is handled and nothing is guarded.
- `.godot` is spelled in three places already, with no guard tying them together:
  `src/gda/import_evidence.py:90` (`CACHE_ROOT_REL`), `src/gda/project.py:899`
  (the UID-cache probe's two spellings) and `src/gda/ops/operations.gd:174`
  (`ENGINE_CACHE_DIR`). This slice adds **no fourth spelling** in `src/` — every
  mention derives from the constant, and `grep -rn "\.godot" src/` returns the
  same 163 hits as the base — and it fixes none of the three, as instructed (the
  backlog carries it). The docs (README, catalog) name the directory in prose, as
  they already did.

- **A failed export reports no mutations.** A non-zero export can have mutated the
  tree (the import pass runs first), but the failure answers through the `Error
  envelope`, and that envelope's `evidence` key is #862's this wave. The second
  walk is therefore skipped on failure — a deliberate boundary, now stated in
  `--help` and the catalog. A later run takes a new baseline and cannot recover
  mutations from that failed interval.
- **The candidate rule's blind spot:** a rewrite that preserves both the size and
  the mtime is not reported. No engine pass does this; the rule is what bounds the
  cost, a test pins it, and since round 1 the published field description says so.
- **The cache root is reported as one unit.** A warm export rewrites its own cache
  bookkeeping (measured by the reviewer: `.godot/editor/filesystem_cache10`,
  `.godot/uid_cache.bin`) and those rewrites are deliberately not compared. Every
  public sentence now scopes the claim, and the human line says "unchanged outside
  res://.godot" rather than "unchanged".
- **Non-regular entries are counted, never opened.** A FIFO, a socket or a device
  inside the project is stat-ed and skipped; the inventory opens regular files
  only. What the report then cannot say about such an entry is the same thing it
  cannot say about an unreadable file: whether the export touched it.
- **`created` is unbounded.** A cold-cache export of a real project lists about
  14,000 entries (GDA-DF-067). That matches `resource import`, whose `created`
  list has the same property, and the counts make the summary readable without the
  list — but a caller that pipes the JSON around should know. Not capped here: a
  cap is a contract question of its own.
- **Sibling output files: one rule, both cases.** The exclusion is the artifact
  and everything under it — a directory artifact such as a macOS `.app` bundle
  included — and nothing else. So a Linux export with
  `binary_format/embed_pck=false`, which writes `game.pck` beside the binary, now
  reports that `.pck` as a created `source_adjacent` file whether or not gda made
  the parent directory. By design: it is a file the export left in the project,
  and the caller decides what to do with it. Before round 3 the same file was
  reported when the parent already existed and dropped when gda created it.
- **The walk does not apply the #804 skip rules.** The engine's own scan skips a
  directory holding a `project.godot` or a `.gdignore`; this walk still enters
  both, as it did before round 3. A vendored sub-project's tree therefore counts
  in `created` even where the export's import pass would not scan it. Not adopted
  here — it is a scope change to the report, not a fix to the link finding.
- **`resource import`'s own walker is still link-blind.** `_project_files` uses
  `Path.rglob`, which does not descend a directory link (verified on Python
  3.13.13), so that command's created-file accounting misses what an import pass
  writes under a linked-in directory. Untouched here: #741's open item 9 decided
  that the two walks stay separate and the shared part is the classification, so
  changing that walker is that decision's to revisit, not this PR's.
- **Following a link widens what the walk reads.** A link to a large tree outside
  the project is walked, because the engine's import scan reads it too. That is
  the `Trusted project` assumption (ADR-0009) and the engine's own cost, not a new
  exposure — but a project that links in a huge directory pays for it twice, once
  in the engine's scan and once in the report's two walks.
- **The reported spelling of a doubly-reachable directory.** A directory reachable
  under two spellings is walked once, under the first the sorted depth-first walk
  reaches. The two walks sort the same way on the same tree, so the pre- and
  post-export spellings agree; a tree whose link layout CHANGES during the export
  could shift a spelling, and the report would then read the same file as created
  under the new one. No engine pass rearranges links.
- **Not verified on Linux or Windows.** The walk uses `os.walk` + `stat` + a
  content digest, and the unit tests cover the rules; only the e2e ran against a
  real engine, and only on this Mac. The two locked-directory tests skip
  themselves where the platform lets the owner list a mode-000 directory. Round 3
  adds the one platform-sensitive piece: the identity pair is `st_dev` + `st_ino`,
  which CPython fills on Windows too (volume serial + file index), but a Windows
  directory link needs a privilege most checkouts do not have, so the link tests
  are exercised on POSIX only here.
- The e2e's touched-but-equal half depends on the engine touching a re-imported
  asset's `.import` sidecar without changing its bytes (measured on 4.6.3-stable).
  If a later engine stops doing that, the assertion fails with a message saying
  the case must be re-sourced, rather than passing vacuously.
- **GDA-DF-067 scale is not exercised.** The largest real export run here is the
  10,005-file generated cost fixture; the 14,000-entry `created` list is inferred
  from the dogfooding record, not reproduced.

## Review round 1

Round 1 (head `a82409164`) returned P1 0, P2 7, P3 8. Every adopted finding is
fixed in `21645fb26`; the map:

| Finding | Fix |
| --- | --- |
| P2-1 overclaim on the README row, its three translations and `--help` | `21645fb26` — all four rows, the `--help` paragraph, the catalog, the two field descriptions and the human line now scope the claim: created files anywhere under the project, rewritten files OUTSIDE the cache root, cache-root rewrites not reported. Translations re-stamped |
| P2-2 the published `modified` description omits the candidate gate | `21645fb26` — the `modified` field description and the model docstring both carry the size-or-timestamp rule and its consequence |
| P2-3 an unreadable DIRECTORY is invisible to both walks and to `skipped` | `21645fb26` — `_walk_project` takes an `os.walk(onerror=…)`; the directory is counted once and its subtree is passed over, so a file the project already had is never announced as created. Two tests, both mutant-checked |
| P2-4 the capture-site use of the shared classifier is pinned by nothing | `21645fb26` — the stub records what it was asked about and the export runner snapshots that record before mutating; the capture-site mutant now goes RED |
| P2-5 two new `.godot` literals in `src/`, against a PR-body claim | `21645fb26` — the field description is an f-string over `CACHE_ROOT_REL`, the help prose names the root by its result key instead; `grep -rn "\.godot" src/` is back to the base's 163 hits, and the deviation bullet above is corrected |
| P2-6 the SKILL omission's justification was false | `21645fb26` — one clause on the `export` row, the same row that carries #840's two result fields |
| P2-7 `CONTEXT.md` untouched and undeclared | `21645fb26` — one entry, `Project-tree mutation report`, under Operations beside `Import evidence` |
| P3-1 over-wide comment line | `21645fb26` — rewrapped; both changed files now have exactly the base's set of >88-character lines |
| P3-2 stale "`_project_files` yields them" note | `21645fb26` — "as the commands' project walkers yield it" |
| P3-3 the render printed a zero row when only `skipped` was non-zero | `21645fb26` — the quiet branch keys on `created`/`modified` alone and appends the unreadable count |
| P3-4 the failure boundary is undocumented publicly | `21645fb26` — one sentence in `--help` and in the catalog paragraph |
| P3-6 the sibling-output residual note is imprecise | `21645fb26` — the note above now states both outcomes and what decides them |
| P3-8 `.git` excluded by two different rules | `21645fb26` — one line at `_VCS_DIR` saying #741's open item 9 keeps the two walks separate |

Not adopted, recorded here by the wave's decision:

- **P3-5, `skipped` as a path list.** It stays a COUNT. A bounded path list would
  reopen the truncation contract (a cap, a truncated flag) for a disclosure whose
  remedy is "repair the permissions and run again" — which the field description
  and the `--help` sentence now say outright.
- **P3-7, the cost bound's mode scope.** The acceptance criterion is read against
  the export the dogfooding record used — the real macOS release export, measured
  at +1.8 % warm — and `--mode pack` at +16.2 % (+0.4 s absolute) is disclosed and
  accepted as proportionate. Issue #839 carries that dated decision; this PR only
  cites it.

**PR CI note (not this slice).** The `Python unit tests and package build` job on
`21645fb26` failed one unrelated test —
`tests/daemon/test_daemon_diag.py::test_launch_session_accepts_a_uid_scene_selector`,
`KeyError: 'argv'`. Its helper grants `launch_session` a **100 ms** readiness
deadline and then asserts the argv the spawn built, and `launch_session` returns
BEFORE spawning when that deadline has already expired ("the readiness deadline
expired before the engine was launched"), so the helper's `captured["argv"]` is
never set. I reproduced that branch directly against this head: an expired deadline
gives `returned None, Popen called: False` — exactly the CI symptom — while the
whole fast tier passes locally three times in a row at this head with the same
`-n 4 --dist loadgroup` settings, and `tests/daemon` passes on its own. Nothing in
this slice reaches `launch_session`. Not fixed here: the helper's deadline belongs
to the daemon suite.

## Review round 2

Round 2 (head `21645fb26`) re-verified every round-1 fix and found P1 0 / P2 1 /
P3 3. All four are fixed in `cbae85b45`; the map:

| Finding | Fix |
| --- | --- |
| P2-1 a FIFO in the project tree hangs `export run` for ever | `cbae85b45` — fixed as the GENERAL rule, not as one case: `_file_facts` stats before it opens and returns "unreadable" for ANY non-regular entry, so a FIFO, a socket, a device and anything else take the one path the socket already took by raising. gda never opens a non-regular entry. The rule is published in the `skipped` description, the model docstring, `--help` and the catalog. Test with a real `os.mkfifo` fixture, run on a thread with a deadline so a regression reads RED instead of wedging the suite |
| P3-1 the candidate gate's timestamp half was pinned by nothing | `cbae85b45` — one test: a pre-existing file rewritten to content of the SAME length with a newer mtime reaches `modified` with `size_before == size`. The size-only mutant, which survived everything including the real-engine e2e, now goes RED |
| P3-2 `_excluded`'s path-component rule was pinned by nothing | `cbae85b45` — one test with `.gitignore`, `.github/ci.yml` and a sibling `.pck` beside an artifact whose parent already existed: all three are reported, which the string-prefix mutant cannot do. This is also the test that holds the sibling-output claim in "Residual risk" |
| P3-3 `CONTEXT.md` attributed the created-file vocabulary to `Import evidence` | `cbae85b45` — the entry now says "classified `cache_owned` / `source_adjacent`, the same verdict `gda resource import` reports for the files its own pass creates, from the shared `import_evidence` module", and the closing guarantee is back to "the export deletes and restores nothing" |

**The FIFO measurement** (this Mac, Godot 4.6.3-stable, `--mode pack`, a 4-file
project plus one `mkfifo pipe.dat`, both runs under the machine-wide lock):

```
pre-fix (the S_ISREG guard removed from this head):  HUNG — killed at 45 s
                                                      (no result, no envelope, no exit)
this head (cbae85b45):                                exit=0 in 3.8 s
                                                      created 7, modified 0, skipped 1
```

The pre-fix run never reached the engine — it blocked in the pre-export walk,
before the native export, where no `--timeout` applies; killing the wrapper left
the child blocked on `open()` until it was killed by hand. The reviewer measured
the same shape against the branch point (`86dae1f84`), which exports normally
because nothing there opens a walked project file (the harness snapshot reads a
fixed list, ADR-0028) — so the hang was this slice's own regression, and the fix
restores the bounded behaviour with the entry counted in `skipped`.

**Schema / help re-diff after round 2** (the `skipped` description is published,
so the delta was rebuilt from a detached worktree at `86dae1f84`, 93 dumps each
side): still exactly `export run`, and the same six nodes as before — the command
`description`, `output.description`, `output.properties` + `project_tree_mutations`,
`output.$defs` + `ProjectTreeMutations` / `ExportCreatedFile` / `ExportModifiedFile`.
`output.required` identical, `input` byte-identical, no command added or removed,
and the other 92 help dumps byte-identical. Only the TEXT of the two descriptions
moved, by the `skipped` clause above.

Of the two nits round 2 recorded without requiring a change, the second — the
"GDA-DF-067 saw about 14,000 such files reported as `warnings: []`" construction —
is corrected in round 3 below. The README's "a cold cache leaves thousands behind"
subject is left as it is.

## Review round 3 (external)

An independent adversarial review of `cbae85b45` against the merge base returned
Standards 0, Artifact/issue 3, Architecture 2, Spec 1. Five findings are adopted;
the sixth is the maintainer's to answer, and has been.

| Finding | Decision | Fix |
| --- | --- | --- |
| **P1 (Architecture 1)** a `res://` output path is not excluded — a real pack export to `res://out2.pck` put the artifact in `created` as `source_adjacent` | Fully adopted | `caf61cad7` — the `res://` form is resolved against the project root before the prefix is derived, the way the engine resolves it; `user://` stays dropped, since it cannot name a path in the tree. Unit regression through the fake export runner, plus a real-engine `--mode pack --output res://out.pck` stage on the existing e2e fixture |
| **P2 (Artifact 2)** the exclusion prunes the whole subtree of every gda-created output parent, so a sibling file is reported in one case and not the other | Fully adopted | `caf61cad7` — one rule: the artifact and its own subtree. A `created_dirs` entry is a directory and both walks report files, so it needs no exclusion. New regression for the gda-created-parent case beside the existing pre-existing-parent one; the field description, the model, `--help` and the catalog state the one rule |
| **P2 (Architecture 2) + P2 (Spec 1)** directory links are not walked, so sidecars created and translations rewritten under `game/assets -> ../shared` were reported by neither list, with `skipped` at 0 | Fully adopted | `cdc3db9c8` — the walk follows directory links as the engine does and applies ADR-0032's #760 policy: identify by filesystem identity (`st_dev`, `st_ino`), walk once under the first spelling, never re-enter a link that leads back up the chain or to a directory already walked. A cycle is not unaccounted content, so it is not counted in `skipped`. Two unit tests (a link out of the project, a `sub/loop -> ..` cycle on a deadline) and one real-engine e2e modelled on the reviewer's repro |
| **P2 (Artifact 1)** state what the report can attribute (concurrent writer) | **Partially adopted** — the contract stays attributive ("what the export did"): the trust model already assumes gda is the project's sole driver (`Concurrent external editor`, ADR-0018), and "net changes observed during the interval" would rename the concept for a case the model excludes. What is added is the disclosure | `b2252c12f` — one sentence in the `project_tree_mutations` field description (so, the schema), in `export run --help` and in the catalog paragraph: the report is the difference between gda's two walks, gda assumes it is the sole driver (ADR-0018), so another writer's change in that interval is attributed to the export. `CONTEXT.md` gains one clause naming the assumption and no mechanism |
| **Prose (catalog)** "about 14,000 such files reported as `warnings: []`" — #839 says they were not reported at all | Fully adopted | `b2252c12f` — the catalog, the `ProjectTreeMutations` docstring and the test module docstring now say the files appeared on disk while `warnings` stayed empty. The four README rows carry no such construction and are unchanged, so no i18n restamp was needed (`tests/repo/test_readme_i18n_sync.py` green) |
| **P3 (Artifact 3)** make the issue's acceptance scope executable (cost bound by mode; the success-only boundary) | Maintainer's amendment, not a code change | Issue #839 carries the dated **2026-09-16** amendment: the cost bound reads against the warm macOS RELEASE export with `--mode pack` as the disclosed exception, and a new criterion states that a failed export carries no report even when it mutated the tree. This PR cites it |

Round 3 also confirms what it did NOT change: `resource import`'s own walker
(untouched, and its e2e suite re-run), the `skipped` count's shape, and the
success-only boundary.

## Review round 4 (exact-head recheck)

The recheck at `b2252c12f` found one composition failure: a pack export to
`res://assets/out.pck` through `game/assets -> ../shared` succeeded but reported
its own artifact as `source_adjacent`. `07ec84d67` fixes the shared boundary:
output exclusion compares the artifact's parent directory identity and final
name with each directory the existing walk reaches. This covers `res://` and
absolute output paths through a link, including an in-project alias whose first
walk spelling differs from the output spelling. It keeps sibling files visible
and prunes directory artifacts without walking their contents. New unit cases
were RED at `b2252c12f` and GREEN at this head; the linked-directory real Godot
4.6.3 pack export now excludes `res://assets/out.pck`.

Issue #839's acceptance checkbox now names the warm macOS release cost bound
and the accepted pack exception. Its failed-export clause now says that a rerun
cannot recover changes made during the failed interval. The success-only result
contract is unchanged.

Validation at `07ec84d67`: 29 export-tree unit tests and 98 export non-e2e
tests passed; the two focused real-engine e2e cases passed; Ruff and focused
Pyright passed; `gda schema` is byte-identical to `b2252c12f`. On the same
generated 10,000-file fixture, three direct capture/settle measurements were
0.237–0.240 / 0.065–0.067 s before and 0.236–0.237 / 0.064–0.067 s after.
This is a scanner comparison, not a fresh full-export cost measurement.


qinhaihong-red:
## Reply to review round 1 (head `a82409164` → fixes at `21645fb26`)

| P2-1 — README row, three translations and `--help` claim "every project file the export created or rewrote" while cache-root rewrites are not reported | Fully Adopt (description); boundary retained | Reviewer measured `.godot/editor/filesystem_cache10` rewritten on a warm export with the product printing "unchanged"; the pre-hash scope excludes the classification root by the issue's 2026-09-15 amendment (engine cache noise would bury the rewritten `.translation` resources the record is about) | Every public sentence now says: created files anywhere under the project (classified), rewritten files only OUTSIDE the cache root, cache-internal rewrites not reported; the human line reads "unchanged outside res://.godot"; translations redone and re-stamped | `tests/repo/test_readme_i18n_sync.py` 7 passed; help/catalog re-read |
| P2-2 — the `modified` `$def` says "content changed"; the size-or-mtime candidate gate and its blind spot live only in an internal docstring | Fully Adopt | Published description vs `_walk_project` | The field description and the model docstring state the candidate gate and its consequence | Schema re-diff (description text only) |
| P2-3 — `os.walk`'s default `onerror` hides an unreadable DIRECTORY from both walks and from `skipped` | Fully Adopt | Reviewer reproduced `skipped 0`, the locked file later announced as `created` | `onerror` counts the directory as skipped once; the settle skips its whole subtree by prefix so pre-existing files under it are not reported as created; two new tests | Both mutants (onerror removed; prefix guard removed) RED → GREEN |
| P2-4 — the pre-walk's classifier reuse is unpinned (capture-site local-rule mutant survives) | Fully Adopt | 178-test export+models suite green under the mutant at `a82409164` | The fake export runner snapshots a RECORDING stub before the tree changes, so the test sees what the first walk asked the shared classifier | Capture-site mutant now RED |
| P2-5 — two new `.godot` literals in `src/` | Fully Adopt | `grep -rn "\.godot" src/` 165 vs the base's 163 | Field description derived from `CACHE_ROOT_REL` (f-string); help prose refers to the `cache_root` result key | Count back to the base's 163 |
| P2-6 — SKILL omission justified by a false claim (the export row already carries #840's result fields) | Fully Adopt | `src/gda/skill/SKILL.md` export row | The report joins that row in the same style | Read |
| P2-7 — `CONTEXT.md` untouched and undeclared | Fully Adopt | Five of the last six slices amended it for the same kind of change; the report had four names across artifacts | New Operations entry `Project-tree mutation report` beside `Import evidence` (what it is, what it is not, `_Avoid_: inventory, diff, changeset`) | Read; `#862` rebases over a different section |
| P3-1/-2 — over-wide comment line; stale "`_project_files` is the only caller" note | Fully Adopt | `import_evidence.py:86`, caller note | Reflowed; "commands' project walkers" | Long-line set identical to the base |
| P3-3 — zero rows rendered when only `skipped` is non-zero | Fully Adopt | render path | The quiet branch keys on created/modified and appends the unreadable count | Mutant (quiet line back to bare "unchanged") RED → GREEN |
| P3-4 — failure boundary undocumented | Fully Adopt | help/catalog | One sentence each: a failed export carries no report, the error envelope answers | Read |
| P3-5 — `skipped` is a pathless count | Do Not Adopt | A bounded path list would reopen the truncation contract (cap, truncated flag) for a disclosure whose remedy is "fix permissions and rerun" | Count retained; the remedy is stated in the field description and help | Recorded on #839 (dated note, 2026-09-15) |
| P3-6 — sibling-output residual-risk note imprecise | Fully Adopt | PR body | Now states that whether gda created the parent decides if a sibling output is reported as created or pruned with the artifact, invisible to the caller | Read |
| P3-7 — the AC's cost bound is mode-agnostic while `--mode pack` exceeds it | Do Not Adopt (accepted as measured) | The AC is read against the export the dogfooding record used, the real macOS release export: +1.8 % warm; `--mode pack` +16.2 % is +0.4 s absolute | Bound retained; the worst case stays disclosed in the PR body | Recorded on #839 (dated note, 2026-09-15) |
| P3-8 — `.git` excluded by two rules | Fully Adopt (one-line note) | Open item 9 on #741 keeps the two walks separate | Comment at `_VCS_DIR` says so | Read |

Head `21645fb26` (one commit on top of `a82409164`). Gates on this Mac (macOS, Godot 4.6.3-stable): `ruff check` clean; `ruff format --check` 517 files; `pyright` 0/0/0; fast tier `2777 passed, 3 skipped`; `tests/export` e2e `20 passed, 1 skipped`; i18n sync 7 passed. Schema/help re-diff: still only `export run` (its description, `output.description`, `output.properties` +1, `output.$defs` +3; `required` and `input` byte-identical; the other 92 help pages byte-identical). 18 mutants RED → GREEN with `__pycache__` cleared around each edit. PR CI: one unrelated failure at this head — `tests/daemon/test_daemon_diag.py::test_launch_session_accepts_a_uid_scene_selector` (`KeyError: 'argv'`; its helper gives `launch_session` a 100 ms readiness deadline that can expire before the spawn on a loaded runner; reproduced by forcing the branch; three local `-n 4` runs at this head green) — re-run requested; the test is not on this slice's path and is tracked separately. Not verified: cost numbers carried from `a82409164` (round 1 adds one callback only for an unlistable directory); the two locked-directory tests self-skip where the owner can list a mode-000 directory; Linux/Windows. Round 2 follows at this head.

qinhaihong-red:
## Reply to review round 2 (head `21645fb26` → fixes at `cbae85b45`)

| P2-1 — a FIFO in the project tree hangs `export run` for ever, before the engine runs (a regression of this slice: the branch point opens no walked file) | Fully Adopt | Reproduced end to end: the pre-fix head stops after `export-get` and never reaches the native export; `open()` on a FIFO blocks until a writer appears, outside every timeout | Fixed as the general rule, not as the FIFO case: `_file_facts` stats before it opens and returns "unreadable" for ANY non-regular entry, so a FIFO, a socket and a device all take the one path the socket already took by raising; gda never opens a non-regular entry; the entry is counted once in `skipped`. Published in the `skipped` description, the model docstring, `--help` and the catalog | Regression test with a real `os.mkfifo` fixture, run on a thread with a deadline so a regression reads RED instead of wedging the suite; measured on the same 4-file project + `mkfifo pipe.dat`, `--mode pack`: pre-fix HUNG (killed at 45 s), this head `exit=0` in 3.8 s, `created 7, modified 0, skipped 1` |
| P3-1 — the candidate gate's timestamp half is pinned by nothing (size-only mutant survives the full fast tier and the real-engine e2e) | Fully Adopt | Reviewer's mutant table; the e2e's edited CSV makes both `.translation` resources grow, so the size half alone promotes them | One test: a pre-existing non-cache file rewritten to content of the SAME length with a newer mtime reaches `modified` with `size_before == size` | Size-only mutant now RED |
| P3-2 — `_excluded`'s path-component rule is pinned by nothing, and it decides a documented outcome (`.gitignore`, `.github/`, the sibling `.pck`) | Fully Adopt | Reviewer's fixture: head `created = ['res://.github/ci.yml', 'res://.gitignore', 'res://build/game.x86_64.pck']`, string-prefix mutant `created = []` | One test with `.gitignore`, `.github/ci.yml` and a sibling `.pck` beside an artifact whose parent already existed: all three reported; this test also holds the sibling-output claim in the PR body's residual-risk section | String-prefix mutant now RED |
| P3-3 — `CONTEXT.md` attributes `cache_owned` / `source_adjacent` to the `Import evidence` term, which defines a different vocabulary; closing sentence dropped "deletes and" | Fully Adopt | `grep -n "cache_owned\|source_adjacent" CONTEXT.md` had one hit, inside the new entry; the other four surfaces say "deletes and restores nothing" | The entry now says "classified `cache_owned` / `source_adjacent`, the same verdict `gda resource import` reports for the files its own pass creates, from the shared `import_evidence` module", and closes with "the export deletes and restores nothing" | Read; the `Import evidence` entry is untouched |
| Two nits recorded without a required change (README "a cold cache leaves thousands behind" subject; the "about 14,000 such files reported as `warnings: []`" construction) | Do Not Adopt (unchanged) | Round 2 asked for no change | Text retained | Say so if they should change |

Head `cbae85b45` (one commit on top of `21645fb26`; five commits over the branch point `86dae1f84`, all `haihong.qin`, all Conventional Commits with `(#839)`). Gates on this Mac (macOS, Godot 4.6.3-stable): `ruff check` clean; `ruff format --check` 517 files; `pyright` 0/0/0; fast tier `2780 passed, 3 skipped`; `tests/export` e2e under the machine-wide lock `20 passed, 1 skipped`. Schema/help re-diff after round 2: still exactly `export run`, the same six nodes; only the TEXT of the two descriptions moved, by the `skipped` clause; `output.required` and `input` byte-identical; the other 92 help pages byte-identical. PR CI at this head: all jobs green (the daemon flake from the previous head did not recur). Not verified: cost numbers carried from `a82409164` (round 2 adds one `stat()` per walked file, which the walk already paid inside `open()`); Linux/Windows (the FIFO test self-skips where `os.mkfifo` is absent).

qinhaihong-red:
## Maintainer re-verification — READY at `cbae85b45`

Independent run on a detached checkout of the pinned head (this Mac, macOS, Godot 4.6.3-stable, host load ~2.3, no sibling agent active):

| Check | Result |
| --- | --- |
| Commits base..head | five, all `haihong.qin <haihongqin@gmail.com>`, Conventional with `(#839)`, no trailers |
| PR body | `Closes #839`; no attribution; "Review round 1" and "Review round 2" sections present |
| File scope | 13 files; no file owned by a sibling W4 slice; `CONTEXT.md` edit confined to the new `Project-tree mutation report` entry |
| `ruff check .` | All checks passed |
| `ruff format --check .` | 517 files already formatted |
| `pytest -m "not e2e" -n 4 --dist loadgroup` | 2780 passed, 3 skipped |
| `pytest -m e2e tests/export` (machine-wide lock) | 20 passed, 1 skipped (the Linux-only templates case) |
| Round-2 mutants, re-run by hand with `__pycache__` cleared and the file restored from a saved copy | `S_ISREG` guard removed → 1 failed in 15 s (the FIFO test reads RED, the suite does not wedge); size-only candidate gate → 1 failed; `_excluded` string-prefix → 1 failed; restored → 22 passed |
| PR CI at this head | all jobs green |

Two independent review rounds closed (round 1: 0/7/8, all adopted but two declines recorded on #839; round 2: 0/1/3, all adopted). First in the W4 merge order on the wave's authorization; the wave's full e2e dispatch runs on the dev ref after the last merge.

qinhaihong-red:
Independent adversarial review of PR #981 at cbae85b451c1cfe477b814b5136a54aa2844fd50 against merge base 86dae1f84b522ea234339926a2c5b6dbbff908c5. I reviewed the issue and exact diff from first principles; earlier PR review conclusions were not evidence.

## Artifact and issue — Changes required

1. **[P2] State what the report can attribute.** Issue #839 and CONTEXT.md:90-98 say the report names files the export created or rewrote; README.md:452 likewise says what export left in the project. The implementation compares two filesystem walks (export.py:768-868), while accepted ADR-0018:17-21 states that Godot takes no project lock and another editor can write during export. A concurrent edit would be attributed to export and could be staged or restored on that false premise. Describe these as net changes observed during the export interval, and state the concurrent-writer limitation where agents are told to act on the result.

2. **[P2] Align the output exclusion with the promised created-file scope.** Issue #839 excludes the artifact, created_dirs, and paths under the output path. export.py:683-697 instead prunes the entire subtree of every newly created output parent. On this head, if gda created build/ for build/game.pck, a separate build/other.txt created during the native pass yields created=[], skipped=0; when build/ existed beforehand, a sibling is reported (the test at test_export_tree_mutations.py:434-461 covers only that case). This contradicts the “every file ... ANYWHERE” wording in CONTEXT.md:93-94 and the catalog:1321-1324. Exclude the artifact and its own subtree without suppressing sibling files, then add the new-parent regression; if the broader exclusion is intentional, amend #839 and the catalog to state it.

3. **[P3] Make the issue acceptance scope executable.** #839 still states an unqualified <=10% cost bound for an export, while its later amendment accepts a macOS release measurement and a pack measurement of +16.2%. Put the mode and accepted exception in the acceptance criterion. The issue also does not say that a native export that mutates the tree and then fails produces no report, although CONTEXT.md:105-107 and export.py:910-934 do. State that success-only boundary and its consequence in the issue, or require failure evidence.

**Terminology and prose assessment.** Checked all changed Markdown, the relevant full issue, CONTEXT.md glossary, catalog, bundled skill, and translated README rows against project terms and ASD-STE100 (word choice, technical nouns, sentence scope and causal wording). The causal wording above is substantive. Minor: catalog:1320-1321 says about 14,000 files were “reported as warnings: []”; #839 says they were not reported at all. Say that those files appeared on disk while warnings remained empty. I did not rerun the claimed 10,000-file performance benchmark.

## Architecture — Changes required

1. **[P1] Exclude a res:// export artifact from the mutation report** (export.py:683-685). The output model accepts virtual paths (lines 65-75), but this exclusion drops every path containing ://. A real Godot 4.6.3 pack export to res://out2.pck on this exact head succeeded and returned created=[{path: res://out2.pck, classification: source_adjacent, size: 1268}], skipped=0. That places the export artifact in the project-effects list despite #839 AC4. Resolve res:// against the project for exclusion, and cover it with a regression.

2. **[P2] Account for engine-visible symlink directories** (export.py:732-742). Godot imports resources through a plain-library directory symlink, while os.walk defaults to not descending it and the inventory does not count the omitted directory in skipped. The export-specific ownership remains in commands/export.py and reuse of import_evidence follows ADR-0040; no module split is indicated. The load-bearing completeness claim remains unestablished until this boundary is fixed or explicitly reported.

## Standards — Pass

Zero documented-standard violations and zero actionable Fowler baseline smells in the 13-file diff. ADR-0040 places the export result, recipe, and renderer in the command group; the shared cache classifier remains the single authority.

## Spec — Changes required (1 finding)

**[P2] Symlinked-in resource mutations are missing.** #839 requires a pre/post project-tree inventory and says a generated resource rewritten by import appears under modified. ADR-0006 treats a plain symlinked-in library as inside the project, and Godot follows it. In a real-engine pack export with game/assets linked to ../shared, the import created PNG/CSV sidecars and two translation files, but created_source_adjacent=0 and skipped=0. After changing the CSV, both translation SHA-256 values and sizes changed (484→520 and 507→547 bytes), yet modified=[] and skipped=0. Walk engine-visible directory links with cycle and boundary control, or explicitly disclose any unvisited tree as skipped and narrow the completeness contract.

Focused exact-head unit validation: 22 export-tree tests passed; git diff --check passed. The two real-engine counterexamples above were run at this SHA. The claimed performance figures were not independently benchmarked.

Axis totals: Standards 0 (none); Spec 1 (symlinked resources, P2).

qinhaihong-red:
## Reply to the external review (head `cbae85b45` → fixes at `b2252c12f`)

| Architecture P1 — a `res://` output path is not excluded; the artifact lands in `created` as `source_adjacent` | Fully Adopt | `_excluded_prefixes` dropped every path containing `://`; the reviewer's real pack export to `res://out2.pck` reproduced it | `caf61cad7` — a `res://` output path is resolved against the project root before the exclusion prefix is derived (`user://` stays dropped: it cannot be in the tree) | Unit test with a `res://` output through the fake runner; a fourth stage on the report e2e exports `--mode pack` to `res://out.pck` and asserts the artifact absent from `created`, `skipped 0`; reverse mutant RED |
| Artifact P2 — the exclusion prunes the whole subtree of every gda-created output parent, so a sibling output is reported only when the parent pre-existed | Fully Adopt | The PR's own residual bullet disclosed the two outcomes; AC4 names the artifact and what is under it, and a directory in `created_dirs` is not a file | `caf61cad7` — one rule: the artifact and everything under it are excluded, plus `.git`; `created_dirs` is no longer an exclusion input; a file the export writes beside the artifact (a Linux export's `game.pck` beside `game.x86_64`) is reported as created `source_adjacent` whether or not gda created the parent, by design | New-parent regression beside the pre-existing-parent test; reverse mutant (`created_dirs` back in the prefixes) RED; the residual bullet replaced |
| Architecture P2 + Spec P2 — directory symlinks are not walked while the engine's import pass follows them (sidecars and rewritten translations under `game/assets -> ../shared` invisible, `skipped 0`) | Fully Adopt | Reviewer's real-engine repro; ADR-0032's #760 amendment already decided the enumeration policy for the engine-side `res://` walk (follow a link as the engine does, identify by filesystem identity, do not re-enter a link back up the chain) | `cdc3db9c8` — `_walk_project` follows directory links, identifies each directory reached by `(st_dev, st_ino)`, walks it once under the first spelling, never re-enters a cycle (not counted in `skipped`); a link whose target cannot be stat'ed counts as unreadable once; paths are reported under the spelling the walk reached them by, as the engine names them; the rule is published in the `created` description, `--help` and the catalog. Not adopted here: #804's skip rules (nested `project.godot`, `.gdignore`) — the walk enters those directories as before; `resource import`'s own `rglob` walker is link-blind too (verified on Python 3.13) — both recorded as follow-ups under #741's open item 9, which keeps the two walks separate | Unit: a link to a directory outside the root with a sidecar created and a file rewritten under it (both reported under the link spelling); a cycle link (`sub/loop -> ..`) terminates with nothing reported twice; e2e modelled on the reviewer's repro (`assets -> ../shared`, PNG + CSV, `--mode pack`; the `.import` sidecar in `created`, both `.translation` files in `modified` after the CSV edit); mutants `followlinks=False` (unit AND e2e) and identity set removed → RED |
| Artifact P2 — a concurrent writer's change is attributed to the export; describe "net changes observed during the export interval" and state the limitation where agents act on the result | Partially Adopt | Godot takes no project lock (ADR-0018), and the trust model already assumes gda is the project's sole driver (`Concurrent external editor`) | `b2252c12f` — the contract stays attributive ("what the export did"); ONE disclosure sentence in the field description (so in the schema), `--help` and the catalog: the report is the difference between gda's walk before the export and its walk after, under the sole-driver assumption of ADR-0018, so another writer's change in that interval is attributed to the export; `CONTEXT.md`'s entry gains one clause naming the assumption. Not renamed to "net changes" | Read; schema delta still only `export run` (text inside the same nodes) |
| Prose — the catalog says GDA-DF-067's files were "reported as `warnings: []`"; #839 says they were not reported at all | Fully Adopt | The issue's record | `b2252c12f` — "appeared on disk while `warnings` stayed empty", in the catalog, the model docstring and the test docstring; the README rows carry no such construction | i18n sync 7 passed (no README change) |
| Artifact P3 — the issue's cost bound is unqualified against the accepted `--mode pack` measurement, and the success-only boundary is unstated | Fully Adopt (maintainer) | The 2026-09-15 amendment accepted the numbers without changing the criterion | Issue #839 carries a dated 2026-09-16 amendment: the cost criterion reads against the warm macOS release export with the pack exception named; a new criterion states that a failed export carries no report even when it mutated the tree; AC4 is read literally; links and the sole-driver assumption stated | Read |

Head `b2252c12f` (three commits on top of `cbae85b45`, all `haihong.qin`). Gates on this Mac (macOS, Godot 4.6.3-stable, load 2–3): `ruff check` clean; `ruff format --check` 517; `pyright` 0/0/0; fast tier `2784 passed, 3 skipped`; under the machine-wide lock `tests/export` e2e `21 passed, 1 skipped` and `tests/resource` e2e `43 passed` (the sibling walker untouched). Schema/help re-diff vs `86dae1f84`: still exactly `export run` — the same six nodes, text only; `output.required` and `input` byte-identical; the other 92 help pages byte-identical. Cost: the walk's own cost is unchanged (in-process control on a generated 10,001-file fixture: capture 0.282 s → 0.284 s, settle 0.066 s both); the fixture has no links, so following them adds nothing there. Not verified: Windows (a directory link needs a privilege most checkouts lack); GDA-DF-067's 14,000-entry scale is still inferred. The maintainer's re-verification on a detached checkout at this head follows.

qinhaihong-red:
## Maintainer re-verification — READY at `b2252c12f`

Independent run on a detached checkout of the pinned head (this Mac, macOS, Godot 4.6.3-stable, load 1.9–2.4):

| Check | Result |
| --- | --- |
| Commits base..head | eight, all `haihong.qin <haihongqin@gmail.com>`, Conventional with `(#839)`, no trailers |
| PR body | `Closes #839`; no attribution; "Review round 1", "Review round 2" and "Review round 3 (external)" sections present |
| File scope (round 3) | 5 files: `export.py`, the two export test modules, `CONTEXT.md` (the `Project-tree mutation report` entry only), the catalog |
| `ruff check .` | All checks passed |
| `ruff format --check .` | 517 files already formatted |
| `pytest -m "not e2e" -n 4 --dist loadgroup` | 2784 passed, 3 skipped |
| `pytest -m e2e tests/export` (machine-wide lock) | 21 passed, 1 skipped (the Linux-only templates case) — includes the new linked-directory export and the `res://` artifact stage |
| Round-3 mutants, re-run by hand with `__pycache__` cleared and the file restored from a saved copy | `followlinks=False` → 1 failed; identity check disabled (`if identity in walked` never true) → 1 failed, the cycle test, no hang; `res://` resolution disabled → 1 failed; restored → 26 passed |
| PR CI at this head | all jobs green, MERGEABLE / CLEAN |

Three review rounds closed (round 1: 0/7/8; round 2: 0/1/3; external: 1/3/1 + prose — all adopted, the concurrent-writer item partially with the attributive contract retained). First in the W4 merge order on the wave's authorization; the wave's full e2e dispatch runs on the dev ref after the last merge.

qinhaihong-red:
## Exact-head recheck — changes required (`b2252c12fd608ad45e64c64263585f36ee0c6e86`)

The previous plain `res://` output, symlinked resource, new output parent, and concurrent-writer disclosure counterexamples are addressed. The focused export-tree tests pass. One supported combination still violates #839 AC4:

- With `game/assets -> ../shared`, a real Godot 4.6.3 `--mode pack` export to `res://assets/out.pck` succeeds and writes `shared/out.pck`, but `project_tree_mutations.created` contains `res://assets/out.pck` as `source_adjacent` and `skipped` is zero. `_excluded_prefixes` resolves the output to a physical path outside the project and drops its prefix; `_walk_project` follows the link and reports the project-relative spelling. The output artifact must be excluded in this combination. An absolute output path through the same link has the same exclusion gap. Cover the combined path in the real-engine export test.

The local backtrace points to a path-coordinate mismatch: the separate `res://` and directory-link fixes each pass their own tests, but their physical and project-relative path rules do not compose. Give output exclusion one mapping into the walk's project-relative namespace; retain filesystem identity for traversal/cycle control. The current evidence does not support introducing a general filesystem platform or narrowing the already accepted link support.

Issue #839 also needs two textual corrections. Its original cost checkbox still says unconditional `<=10%`, while the dated amendment accepts the measured pack exception. The amendment says a caller can rerun after a failed export to obtain a report; the first failed run's mutations become the rerun's baseline, so that report cannot recover them. Keep the success-only contract and state this limit accurately.

Validation: 26 focused export-tree unit tests passed; the existing real-engine tests for plain `res://` output and linked resources passed, but the combined real-engine counterexample above fails. No full-suite claim. I will address these confirmed findings on this PR branch.

qinhaihong-red:
## Resolution of the exact-head recheck (`07ec84d67f1c4f10bf12397b87d76d8798125558`)

**Fully adopted the artifact-exclusion finding.** The old physical-path prefix was discarded when an output passed through a directory link; the walker then reported that output under a project-relative alias. The export-specific exclusion now identifies the artifact by its parent directory's filesystem identity and final name during both walks. That excludes a file or directory artifact under either the supplied or another link spelling, while keeping sibling files visible. The traversal's existing identity rule still handles cycles; no general filesystem layer was added.

Three new unit cases were RED at `b2252c12f` and GREEN here: `res://` and absolute output through an external directory link, plus an in-project alias where the walk picks another spelling. The real Godot 4.6.3 pack export to `res://assets/out.pck` through `game/assets -> ../shared` now writes the artifact without listing it in `created` (`skipped=0`). In all, 29 export-tree unit tests, 98 export non-e2e tests, and two focused real-engine export e2e tests passed. Ruff, focused Pyright, `git diff --check`, and a byte-for-byte `gda schema` comparison passed. Three direct capture/settle runs on the same generated 10,000-file fixture were 0.237–0.240 / 0.065–0.067 s before and 0.236–0.237 / 0.064–0.067 s after; this is a scanner comparison, not a fresh full-export cost claim.

Issue #839's acceptance checkbox now names the warm macOS release cost bound and accepted pack exception. It also states that a report from a later run cannot recover mutations from a failed export. The PR description reflects the current exclusion rule and those limits. Hosted CI is green; its Godot e2e job is skipped by CI policy, while the two focused engine tests above ran locally.

qinhaihong-red:
## Maintainer re-verification — READY at `07ec84d67`

The exact-head recheck's finding is confirmed as the right diagnosis: the round-3 exclusion resolved the destination to a physical path while the walk reports the first project-relative spelling, so the two coordinate systems did not compose through a directory link. The fix identifies the artifact by its parent directory's filesystem identity and final name at the one visit the walk makes to that directory, which is the walk's own coordinate; the parent exists before the pre-export walk because gda creates it first (`_ensure_output_parent_dirs` runs before `_PreExportInventory.capture`), so both walks exclude the same entry.

Independent run on a detached checkout of the pinned head (this Mac, macOS, Godot 4.6.3-stable, load 2.1–2.8):

| Check | Result |
| --- | --- |
| Commits base..head | nine, all `haihong.qin <haihongqin@gmail.com>`, Conventional with `(#839)`, no trailers |
| PR body | `Closes #839`; no attribution; rounds 1–3 and "Review round 4 (exact-head recheck)" present; the exclusion rule described as shipped |
| `ruff check .` / `ruff format --check .` / `pyright` | clean / 517 files / 0 errors, 0 warnings |
| `pytest -m "not e2e" -n 4 --dist loadgroup` | 2787 passed, 3 skipped |
| `pytest -m e2e tests/export` (machine-wide lock) | 21 passed, 1 skipped — includes the `res://assets/out.pck` export through `assets -> ../shared` |
| Mutants, re-run by hand with `__pycache__` cleared and the file restored from a saved copy | the artifact name never matched → 7 failed; the parent identity never computed → 7 failed; restored → 29 passed |
| Issue #839 | the cost checkbox names the warm macOS release bound and the pack exception; the new criterion states that a failed export carries no report and that a later run's baseline cannot recover it |
| PR CI at this head | all jobs green, MERGEABLE / CLEAN |

Four review rounds closed. First in the W4 merge order on the wave's authorization.

PR #987, opened 2026-09-17
feat(gda): export smoke runs a caller-selected export artifact headlessly and reports its completed process evidence (#979)

qinhaihong-red:
## Summary

`gda export run` reports whether Godot **built** an artifact; it never runs it. GDA-DF-072 is the gap that made this a slice: the first exported candidate loaded the whole game and then reported four leaked WAV resources at exit, while `export run` had returned `warnings: []`. The defect was observable only by launching the exported game.

This adds one bounded headless run of a caller-selected `Export artifact`, end to end:

```text
gda export smoke <artifact> [--arg VALUE ...] [--quit-after FRAMES] [--timeout SECONDS] [--strict]
```

It resolves the artifact to a Godot executable, launches it once through the shared `Headless launch` under a private `user://`, and returns the completed process as evidence. Projectless, `ExecutionKind.ARTIFACT_SMOKE`, three classifier-source `operation` codes at exit 4. ADR-0042 is the authority; where it and the issue differed, the ADR won.

Closes #979

### The command (`src/gda/commands/export.py`)

The whole vertical slice lives in the `export` group module (ADR-0040): `ExportSmokeParams` / `ExportSmokeResult`, `resolve_artifact_executable`, `smoke_args`, `run_export_smoke_operation`, `render_export_smoke`, the recipe, `EXPORT_SMOKE_COMMAND` (`inherits_project=False`, `kind=ARTIFACT_SMOKE`, recipe-routed like `SCRIPT_RUN` / `IMPORT`) and the Typer body. `dispatch.py` is untouched: the projectless recipe path already existed and was verified, not built — `_project_context` returns `None` for a non-inheriting command with no explicit `--project`, and the command declares none.

The module docstring's "the one command that cannot go through `operations.gd`" is corrected: `export run` cannot because the export subsystem is editor-only, and `export smoke` cannot because what it runs is the exported game itself.

### Shared seam 1 — one explicit placement input (`src/gda/runner.py`)

`launch()` gains `user_data_root: Path | None = None`. Omitted, the root is resolved process-wide exactly as before, so **every existing caller keeps its behaviour and its argv**; given, it is the root for that launch and the process-wide resolution is not consulted. Both inputs then share one preparation, spawn, refusal and report, extracted as `_launch_under` — a second copy is how the two would come to place a launch differently. `LaunchFn` learns the same optional keyword.

Lifetime stays with whoever supplied the root: the primitive still removes only the private temporary log directory it makes for itself.

### Shared seam 2 — the completed-run result base (`src/gda/completed_run.py`, new)

`script run` is no longer the only consumer of the promoted `Raw run`. Extracted: `STDOUT_CAP`, `bounded_stdout`, `spill_failure`, `completed_run_schema_extra` (the published truth table), `check_stdout_projection`, and the `CompletedRunResult` base that carries the schema extra and the matching validator.

**Where it lives, and why.** A new core module below both command groups, not a cross-group import from `commands/script.py`. `commands/export.py`'s own docstring declares that it "imports the shared machinery downward" and "is imported by nothing but the composition root"; importing the `script` group sideways would contradict that invariant and pull a whole sibling group (its Typer app and eight commands) into the `export` group's import graph. The shared half belongs to neither group, so it sits under both. No cycle: `gda.completed_run` imports `gda.errors`, which imports nothing from it.

**The base declares no fields, deliberately.** Pydantic orders a subclass's fields base-first, so declaring them there would move `script run`'s `path` out of first position and change its result bytes and its published schema. `gda.models.ProjectRootedResult` made exactly this choice for exactly this reason ("It declares no fields, so a subclass's schema — field order included — is exactly what it was"). What is shared is the RULE; each result spells its own fields with its own descriptions, which differ because the subjects differ — a user script's run and an exported game's. `script run`'s result key order is unchanged and verified below.

The `stdout_spill_failed` message now names WHAT ran (`the script ran …` / `the exported artifact ran …`); that is the whole generalization the second consumer needed. Its registry description is generalized to match and the ADR-0002 row with it.

`gda.completed_run` also owns `render_completed_run(run, *, lead)` (orchestrator re-verification): everything both commands show after their own opening line — the stdout block, the truncation note, the stderr block and the `script_error_line` loop — which `render_export_smoke` had repeated from `render_script_run` line for line. `script run`'s rendering is byte-identical, proved by a before/after dump over 24 fixtures.

The module also owns `DEFAULT_COMPLETED_RUN_TIMEOUT_SECONDS = 120.0` (round 1, P3-4): three public sentences and the issue state that the two commands' ceilings are the same number, so the number and its reasoning live once and each command keeps its public name as an alias (`DEFAULT_SCRIPT_RUN_TIMEOUT_SECONDS`, `DEFAULT_SMOKE_TIMEOUT_SECONDS`), which moves nothing else and keeps the existing 120.0 pin.

### Failure builders (`src/gda/errors.py`)

`export_artifact_not_found_failure`, `export_artifact_not_runnable_failure` (which names which resolution rule refused), and `smoke_exit_status_failure` — the `--strict` verdict, carrying `FailureEvidence.exit_status` and `FailureEvidence.script_errors` and nothing new. They live here because `gda.errors` is where every evidence-carrying builder lives by convention, which is what both AST producer guards in `tests/cli/test_error_registry.py` read. Two new fixed diagnostics headers, `--- artifact stdout ---` / `--- artifact stderr ---`: the same layout under a different subject.

### ADR notes (dated, appended at the end, no historical passage edited)

**ADR-0004**, after the #862 note:

> **Outcome (2026-09-17, #979): a TENTH producer.** `smoke_exit_status_failure` — `gda export smoke --strict`'s `smoke_failed` verdict (ADR-0042) — joins the set above […] It carries the two fields the `script run` verdict beside it already carries and no others […] The shape is unchanged: no field joins `FailureEvidence` and no command-specific envelope appears. […] The placement stays OUT, and the #862 note's three named builders stay three.

**ADR-0031**, at the end:

> **Outcome (2026-09-17, #979) — the promoted completed-run result is no longer this command's alone.** […] Two passages above are therefore HISTORICAL rather than current, and are kept as written: this ADR's claim that `script run` is the one operation, and the one command, whose success result can carry a non-zero `exit_status`. […] the `--help` and result-model sentences that stated the exclusivity are reworded in place, since those are read as current text. […] `script run`'s result bytes, its codes and its output schema shape are unchanged by the extraction.

**ADR-0042**, citation erratum only: the two `main.cpp` anchors now point at the real 4.6.3-stable lines — `--quit-after` is parsed at [L1741-L1748](https://github.com/godotengine/godot/blob/4.6.3-stable/main/main.cpp#L1741-L1748) and ends the main loop at [L5056-L5062](https://github.com/godotengine/godot/blob/4.6.3-stable/main/main.cpp#L5056-L5062) (verified against the `4.6.3-stable` tag in the local engine checkout), and the follow-up probe is dated 2026-09-15. No other word of those sentences changed.

### Everything else reconciled here

`error_codes.py` (three rows plus the generalized spill row) and ADR-0002's table; `models.py`'s two reused `FailureEvidence` descriptions; the guard tests (`_EVIDENCE_PRODUCERS`, the empty-evidence pin, `_RECIPE_OPERATIONS`, the execution-kind pins); `script run`'s two exclusivity sentences and the stale "second passthrough consumer" NOTE; the README row plus its bounded-support note, the three localized READMEs (translated first, markers refreshed after); the catalog's `export` section; the SKILL `export` row.

`CONTEXT.md` is untouched: it already defines `Export artifact` and `Artifact smoke`, and no term changed.

## Artifact resolution rule (as shipped)

| Input | Verdict |
| --- | --- |
| Path names nothing on disk (a broken symlink included, and a `~unknownuser/…` prefix gda cannot expand) | `export_artifact_not_found` |
| The host cannot inspect the artifact path (for example, its parent denies search permission) | `export_artifact_not_runnable` |
| A regular file this host may execute (`S_ISREG` + `os.access(X_OK)`, both following links) | accepted **as given** |
| A regular file without execute permission | `export_artifact_not_runnable` |
| A `.app` directory whose `Contents/Info.plist` names `CFBundleExecutable`, and `Contents/MacOS/<that name>` is a regular file this host may execute | resolves to that file |
| A `.app` whose `CFBundleExecutable` is not ONE filename — blank, a path separator, `.` or `..`, or a NUL | `export_artifact_not_runnable` |
| A `.app` directory without that plist, key, or file — or whose named file is not executable | `export_artifact_not_runnable` |
| Any other directory | `export_artifact_not_runnable` |

Nothing else is inspected: gda classifies no export platform and models no artifact format, and whether the resolved file is a Godot build is what the run shows. The refusal message names which rule spoke.

**The `.app` rule is not gated on the host platform**, deliberately: it reads a layout the artifact declares, and gating it would be exactly the platform classification ADR-0042 rejected.

The declared `CFBundleExecutable` must be **one filename** — not blank, no path separator, not `.` or `..`, no NUL — before it is joined to `Contents/MacOS`. That is what makes the published rule true rather than aspirational: `Path.joinpath` lets an absolute or climbing value out of the bundle, so an unchecked one would run a program the caller never selected. It is a filename test and nothing more: no sandbox, no containment check on the joined result, no identity or provenance notion.

**The path is made absolute once, before resolution.** `ExportSmokeParams.artifact` carries this module's own `SmokeArtifactPath` — `~` expanded, a relative path joined to the invocation cwd — so the resolver, `executable`, the human rendering, both refusal messages and the `smoke_failed` message all derive from one absolute value, and none of them can hand a consumer a string that means something else from another directory. It is absolute, not canonical: `..` is not folded and a symlink is not resolved, because the artifact stays the path the caller named. Unlike `export run --output`'s `ExportOutputPath` it has **no virtual-path exception**: a projectless command has nothing to resolve a `res://` against, so every input is a filesystem path, `://` or not. (Round 1's P2-1 absolutized only `artifact`, at the result site; the external review found the two remaining holes — see below.)

## Schema / help diff

Re-measured at the final head `70d98fc10` against a detached worktree at the base commit `0a749f116` (removed afterwards), and independently rebuilt by the round-2 reviewer at `70d7e6707` and by the external Spec reviewer at `8f33b3196`: `gda schema`, `gda schema --schema`, and every one of the 77 commands' `--help` and `--schema`, captured on both sides and diffed structurally. The complete delta, and nothing else:

1. **A new `export smoke` entry** in the aggregate manifest (`kind: "artifact_smoke"`), its own `--schema`, its own `--help`, and its line in the `export` group's help listing. No other command was added or removed.
2. **`ExecutionKind` gains `artifact_smoke`** — visible in `gda schema --schema` as one more `enum` member and as the enum's `description`, which documents the new member beside the other five. Nothing else in that meta-schema moved.
3. **Description-only changes to the two reused `FailureEvidence` fields.** `exit_status` now reads `(script run --strict, export smoke --strict)`; `script_errors` gains the one nuance both gates share — a `shutdown_leak` entry can be the trigger a caller opted into without becoming a verdict of its own. These are embedded in every command's `error` schema, which is why all 77 existing per-command schemas differ by exactly these two lines and by nothing else.
4. **Text-only changes to `script run`'s command description and its `ScriptRunResult` description.** The command description loses "This is the ONE command whose success result can carry a non-zero `exit_status`" and its paragraph is re-wrapped at the file's own width (round 1, P3-6: the reworded sentence left a 96-character source line that rich split into a 16-character orphan at 80 columns; the rendered words are unchanged). The result description loses "Unlike every other command" and "the **one** command", names `export smoke` as the other completed-run result, replaces the stale "a second passthrough consumer should promote…" NOTE with what actually happened, and updates the moved constant's `:data:` reference (`SCRIPT_STDOUT_CAP` → `STDOUT_CAP`). All of it is inside those two description fields.

Verified for `script run` specifically (a structural walk of its `--schema`): the ONLY changed nodes are `/error/$defs/FailureEvidence/properties/{exit_status,script_errors}/description` and `/output/description`. `output.properties` **key order**, `output.required` and `output.allOf` are byte-identical, and so are its codes and its result bytes.

Every other command's `--help` is byte-identical. Only `script run --help` and the `export` group listing differ, both of them item 1 or item 4 above.

**PR CI runs no Godot e2e.** It runs `ruff check`, `ruff format --check`, `pyright`, and the fast suite (`-m "not e2e"`) — which covers the whole artifact-resolution rule, the argv tail, the private-root lifecycle, both `--strict` triggers, the passthrough and the spill bound, the command surface and the negative gate, the launch seam's new input and its source guard, the ADR-0002 / ADR-0004 registry pins, the SKILL and README i18n sync tests, and the schema/kind pins.

**Only the wave's e2e dispatch (and the local runs below) exercise the engine**: the real `export run` → `export smoke` round trip, `OS.get_cmdline_user_args()`, `--quit-after` against the release template, the exit-time leak record, the timeout capture, and the private `user://` against a real game. `tests/export/test_e2e_export_smoke.py` is macOS-gated and additionally skips where the host has no export templates.

## Validation

Host: this Mac, macOS (arm64), Godot 4.6.3-stable, load average ~2.3–4.4 across the runs (other agents share the machine). Every Godot-spawning suite ran in the foreground under the machine-wide e2e lock. **The release-template export RAN — it did not skip:** the host has 4.6.3 templates installed, so `tests/export/test_e2e_export_smoke.py` built a real macOS `.app` with `gda export run` and smoked it.

The four fast gates below were re-run at the FINAL head after the last commit. They had once been run one commit early: `cf68a725b` therefore reached PR CI with 9 `reportArgumentType` errors in the truth-table fixture it added — a `dict(...)` row splatted into the model, whose inferred value union pyright rejected field by field — and the "0 errors" line first published here was measured before that commit existed. That fixture is now a plain dict validated through `model_validate`. Every line below is measured at the final head `70d98fc10`, except the four marked otherwise, which no later commit can affect.

| Command (at `70d98fc10`) | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `521 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2931 passed, 3 skipped in 26.09s` |
| e2e `tests/export` (incl. the new module, which RAN) | `36 passed, 1 skipped, 184 deselected in 78.31s` |
| e2e `tests/resource tests/project` (the shared normalizer's other callers) — at `a838a8ce9` | `161 passed, 408 deselected in 269.00s` |
| `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` (no harness file in the diff) |
| schema / help structural diff vs a detached base at `0a749f116` | still EXACTLY the four declared items; the annotation swap leaves `artifact`'s published property byte-identical |
| e2e `tests/script` (the renderer's other channel) — at `8f33b3196` | `126 passed, 335 deselected in 154.24s` |
| `COLUMNS=80 gda export smoke --help` — at `8f33b3196` | renders with no orphan line; `Default 120.0s` / `[default: 120.0]` still interpolated from the constant |
| `script run` / `export smoke` human rendering, 24 fixtures, before vs after the renderer extraction — at `8f33b3196` | `diff` reports no difference |

**The sibling e2e tiers.** Round 1 DID touch `commands/script.py` — which is `script run`'s own code, though only as a constant alias and a docstring reflow — so an earlier version of this sentence claiming "no sibling channel's code" was wrong. Round 2 re-ran all four tiers itself at `70d7e6707`, the head that carried those edits: e2e `tests/export` `36 passed, 1 skipped in 76.45s` (with `test_e2e_export_smoke.py` `15 passed in 30.56s`, the release-template module RAN), e2e `tests/script` `126 passed in 149.96s`, e2e `tests/scene tests/resource` `135 passed in 211.70s`, e2e `tests/project tests/runtime` `151 passed in 222.91s` — all matching the figures first published here.

`tests/export` and `tests/script` are re-run at the final head above, because the orchestrator's renderer extraction touches both commands' human output. `tests/scene tests/resource` (`135 passed in 211.70s`) and `tests/project tests/runtime` (`151 passed in 222.91s`) are not: nothing since `70d7e6707` touches the sentinel, import or preflight channels — the renderer extraction is confined to `completed_run.py` and the two completed-run renderers, and the rest is ADR/catalog/SKILL prose.

The one e2e skip is pre-existing and unrelated: `test_export_run_without_templates_yields_export_templates_missing` forces missing templates through `XDG_DATA_HOME`, which only the Linux engine honours.

The five launch-backed sibling channels (sentinel ops, export, import, `script run`, `scene preflight`) pass their unit **and** e2e tiers with unchanged assertions — the only edits in their suites are the renamed shared cap constant and one added keyword on `script run`'s `FakeLaunch` to satisfy the widened seam.

## Review round 1

Eight findings (0 P1 / 2 P2 / 6 P3), all handled at `37e08bd37` (with the e2e assertion it added corrected at `70d7e6707`). Seven fully adopted, one partially.

| Finding | Decision | Change (file) | Validation |
| --- | --- | --- | --- |
| **P2-1** `executable`, the human rendering and all three messages echoed the caller's RELATIVE path; only `artifact` was absolutized, at the result site | Fully adopt | `commands/export.py`: `ExportSmokeParams.artifact` carries `ExportOutputPath` (#403) instead of `NormalizedPath`, so the path is absolute BEFORE resolution; the result-site `Path(artifact).absolute()` is gone; the params docstring and the `artifact` field description say what derives from it | Both relative-path tests now assert `executable` is absolute and equals the cwd-joined path and that the launch got that same path; two new tests assert the `not_found` / `not_runnable` messages and the `smoke_failed` message carry the absolute path; the operation-tier implementation-shape pin (`resolve_artifact_executable("build/game") == Path("build/game")`) is replaced by a contract assertion through the operation. **Reverse mutant** (back to `NormalizedPath`) → RED on 5 tests. Published input schema for `artifact` verified identical under both annotations |
| **P2-2** ADR-0004's new note claimed "the first that is not `script run`'s", which the set disproves five times (`launch_timeout_failure` is the shared one) | Fully adopt | `docs/adr/0004-…md` headline is now "a TENTH producer."; the mirrored sentence in `tests/cli/test_error_registry.py` is deleted. No narrower "first" put in its place | `tests/cli/test_error_registry.py` green; the note's substance (tenth producer, two reused fields, no shape change, placement stays out) is untouched |
| **P3-1** nothing held ADR-0042's "private root created AFTER the artifact resolves" — reordering two blocks leaks a temp dir per refusal with every test green | Fully adopt (test only) | `tests/export/test_export_smoke_operation.py`: `test_a_refused_artifact_creates_no_private_root_at_all` redirects `tempfile.tempdir` (the module attribute, because `tempfile` caches its directory) and asserts a refused artifact leaves it empty | **Reverse mutant** (resolve block moved below the root creation) → RED |
| **P3-2** `dispatch.py` stated `inherits_project=False` as an equivalence with "a meta command"; `headless.py` enumerated the set without `export smoke` | Fully adopt | Docstring/comment only in both: `dispatch.py` now reads "(a meta command, or `export smoke`, which acts on a caller-selected path)"; `headless.py` names `export smoke` as the one DOMAIN command that sets it, with its reason | No behaviour change; fast suite green |
| **P3-3** two public sentences called the artifact a "build", which CONTEXT.md's `Export artifact` entry lists under `_Avoid_` | Fully adopt | `commands/export.py` help: "a smoked artifact"; `docs/command-catalog.md`: "a caller-selected artifact". Grepped the README, the three localized READMEs and the SKILL for the same word in this sense — no other hit. The unrelated "whether the file is a Godot build" stays: that is the other question | `gda export smoke --help` re-rendered; docs diff reviewed |
| **P3-4** the 120 s default was duplicated with nothing relating the two constants, while three public sentences assert they are the same number | **Partially adopt** | The review offered "assert equality" OR "drop the claim"; neither fits — RULES.md wants ONE authority for a fact three public sentences and the issue state. `completed_run.py` now owns `DEFAULT_COMPLETED_RUN_TIMEOUT_SECONDS = 120.0` with the ceiling's reasoning; `commands/script.py` and `commands/export.py` keep their public names as aliases, so nothing else moves. The "restated rather than imported" comment is deleted | The three public sentences are now structurally true rather than test-asserted, so no new test; `tests/script/test_script_run_operation.py`'s existing 120.0 pin still holds, and the schema/help diff is unchanged (the number did not move) |
| **P3-5** `smoke_exit_status_failure`'s `artifact` parameter receives the resolved executable | Fully adopt | `errors.py`: renamed to `executable`, docstring sentence added saying why the message names what RAN. Message text unchanged | Fast suite green; the strict-message test asserts the absolute executable |
| **P3-6** the reworded `script run --help` sentence left a 96-char source line that rich split into a 16-char orphan at 80 columns | Fully adopt | `commands/script.py`: that paragraph re-wrapped at the file's own width. Rendered words unchanged | `COLUMNS=80 gda script run --help` re-rendered: every line is a whole source line, no orphan. The delta stays inside declared item 4 |

Not changed on the reviewer's own reading: nothing. The review raised no P1.

## Review round 2

Passed: 0 P1 / 0 P2 / 3 P3, all record- or comment-level, none touching shipped behaviour, the schema or the contract. All three fully adopted at `56668790a`. The round-2 reviewer independently re-verified every round-1 fix (both reverse mutants RED again), rebuilt the schema/help delta from a base worktree, and re-ran all four e2e tiers.

| Finding | Decision | Change | Validation |
| --- | --- | --- | --- |
| **P3-1** the comment introducing the new kind calls `ARTIFACT_SMOKE` "the fifth execution kind"; `IMPORT` was the fifth (#668) and this is the sixth — which the same PR already says in `tests/meta/test_schema_aggregate.py` | Fully adopt | `commands/export.py`: one word, `fifth` → `sixth` | `execution.py` has six members and `gda schema --schema` publishes six; the test comment and `test_all_execution_kinds_appear_in_the_aggregate` ("All six kinds") already agreed |
| **P3-2** the 120 s ceiling is re-spelled as a bare literal in the catalog and the SKILL smoke rows, outside the single authority round 1 declared — three spellings of one fact, two unconstrained | Fully adopt | `docs/command-catalog.md` → "(the same completed-run ceiling `script run` uses; the number is stated where it is interpolated, in `--help`)"; `src/gda/skill/SKILL.md` smoke row → "`--timeout` (the same default ceiling as `script run`)". `script run`'s OWN SKILL row keeps its "default 120": that is the one prose statement of the number, beside the test that pins the constant | `grep "default 120"` → one hit left, `script run`'s row. `gda export smoke --help` still renders `Default 120.0s` / `[default: 120.0]` from the constant. The README rows never carried the literal, so no translation and no i18n marker refresh; `tests/repo` green |
| **P3-3** the PR body still quoted the PRE-fix ADR-0004 headline, said "four" launch-backed sibling channels while naming five, and justified not re-running the sibling e2e tiers with "no sibling channel's code" when round 1 edited `commands/script.py` | Fully adopt (record only) | This body: the ADR-0004 blockquote now reads as the file does ("a TENTH producer."); "four" → "five"; the sibling-e2e sentence replaced with what is true — round 1 did touch `commands/script.py` (an alias and a docstring reflow), round 2 re-ran all four tiers at `70d7e6707`, and the final commit changes no executable code | The live body; `git diff 70d7e6707..HEAD -- '*.py'` is one comment line |

## Orchestrator re-verification

Gates and the `tests/export` e2e tier reproduced at `56668790a`.

- **Design axis** — no change recommended.
- **Spec axis** — pass, no gap. Its one P3 is **not adopted**: a `res://` artifact passes the path normalizer unchanged and is refused as `export_artifact_not_found` echoing the virtual path. The typed refusal is correct, and a projectless command owes no meaning to a virtual path — resolving one would need a project, which this command deliberately does not have.
- **Standards axis** — five items adopted, four declined.

| Item | Decision | Change (commit) |
| --- | --- | --- |
| `render_export_smoke` repeats `render_script_run`'s tail line for line (Fowler, "Duplicated Code"); only the lead differs | Adopt | `958615860` — extracted as `gda.completed_run.render_completed_run(run, *, lead)`, beside the projection it renders; both renderers compose it with their own lead. `script run`'s output is **byte-identical**: both renderers dumped over 24 fixtures (six completed-run shapes × `script run`'s three placement shapes, plus the smoke) before and after, `diff` empty. A mutant dropping the lead turns the render tests red. `render_completed_run` joins `_HELPER_RENDERERS` for the same reason the five already there did |
| ADR-0042 has no record of the implementation outcome | Adopt | `8f33b3196` — one dated `> **Outcome (2026-09-17, #979 / PR #987)**` paragraph appended at the END: the base carries the RULE and declares no fields (pydantic orders base fields first; a field-carrying base would have moved `script run`'s `path` and broken the byte-identity the ADR's own validation list requires), where it lives, the `.app` rule as shipped and why it is not platform-gated, and the anchors corrected as a citation erratum. No historical passage edited |
| ADR-0040 §5 says a shape no single group owns stays in `gda.models`; `gda.completed_run` deviates | Adopt | `8f33b3196` — two sentences appended at the END: the shared half owns BEHAVIOUR (the spill's file IO, the bounded projection, the typed refusal, the default ceiling, the shared rendering), not only a shape; `gda.models` stays the home of field-carrying shared shapes |
| `64 KiB` re-spelled by hand in the smoke's catalog and SKILL rows, outside `completed_run.STDOUT_CAP` | Adopt | `8f33b3196` — same treatment as round 2's P3-2: the smoke rows keep the relation ("the same cap that command uses"), the number stays where `--help` interpolates it. `script run`'s own rows (SKILL, README, its catalog section) keep "64 KiB" as the one prose statement. The README smoke row never carried the literal, so no translation and no marker refresh |
| the new catalog text's one "behaviour" against the file's "behavior" | Adopt | `8f33b3196` |
| a `root_or_none()` seam for the empty `--user-data-root` in `run_export_smoke_operation` | **Decline** | The command reuses the existing public resolver and the refusal keeps ONE owner, in `launch`; a new seam for a single consumer is outside this slice's bound |
| the three-line `runnable_file` / `runnable` helpers duplicated across the two smoke test modules | **Decline** | Test-local scaffolding, no material cost |
| CONTEXT.md's "Their shared result base contains …" sentence | **Decline** | True of what both results carry; the mechanism is now recorded in ADR-0042's note instead |
| the `res://` refusal message | **Decline** | See the spec axis above |

## External review

Four independent external reviews at `8f33b3196`: **Standards — pass, 0 findings.** The other three reported three distinct defects, reproduced here before any edit and re-probed after. All adopted; nothing declined.

| Finding | Decision | Change (commit, file) | Validation |
| --- | --- | --- | --- |
| **A bundle's `CFBundleExecutable` could select a file OUTSIDE the artifact** (Artifact 1 / Architecture 2 / Spec 1). `/bin/echo` made `Path.joinpath` return that path verbatim, so the command LAUNCHED it and published it as `executable` with exit 0 — against the rule ADR-0042 and the help both publish, and against CONTEXT's "its trust subject is the caller-selected artifact" | Fully adopt | `a5088976f`, `commands/export.py`: `_one_filename` requires ONE usable filename — not blank, no `/` or `os.sep`, not `.` or `..`, no NUL — before the join. Every other value is the existing `export_artifact_not_runnable`, message naming the rule. No sandbox, no containment check on the joined result, no identity notion | Seven resolver-tier cases (absolute, `..`-bearing, separator, `..`, `.`, blank, NUL) plus a CLI-tier regression through `--json`. **Reverse mutant** (drop the check) → RED on 9 tests |
| **A NUL in `CFBundleExecutable` escaped as a traceback** (Architecture 2 / Spec 1): `os.stat` raises `ValueError`, not `OSError`, so the CLI exited 1 with no JSON envelope at all | Fully adopt | `a5088976f`, `commands/export.py`: the filename rule refuses it first, and `_is_runnable_file`'s guard widens to `(OSError, ValueError)` so a path the syscall cannot carry is honestly "not runnable" whatever a later caller hands in — closing the class, not the instance | The CLI regression proves the typed envelope and exit 4; the guard is pinned on the helper directly. **Reverse mutant** (narrow back to `OSError`) → RED |
| **A valid relative artifact could stay relative** (Architecture 1 / Spec 2): `ExportSmokeParams.artifact` reused `export run --output`'s normalizer, whose `"://" → unchanged` branch is for virtual paths. A real POSIX file under a directory named `foo:`, addressed as `foo://game`, kept its relative spelling in BOTH published addresses | Fully adopt | `301a713d7`, `commands/export.py`: the common half becomes `_absolute_filesystem_path`; `normalize_export_output_path` stays the virtual-path guard in front of it; the smoke gets `SmokeArtifactPath`, which calls the helper directly. The remedy **deletes** the inherited exception for this command rather than adding a rule on top — which is why this second round on the "addresses must be absolute" family converges | Unit + command-tier tests that `foo://game` yields an absolute `artifact` AND `executable`; `export run --output res://…` pinned unchanged and its own suites untouched. **Three reverse mutants** (point the smoke back at `ExportOutputPath`; give the smoke normalizer the `://` branch; remove `export run`'s branch) → all RED |
| **The glossary assigned fields to a fieldless base** (Artifact 2): `CONTEXT.md`'s `Raw run` entry said the shared base "contains" those fields, which the shipped `CompletedRunResult` does not | Fully adopt | `a838a8ce9` corrected field ownership; `da1b5560a` completes it: the fieldless base owns the validator and schema rule, while the `gda.completed_run` module owns rendering, the cap, projection, spill handling, and default timeout | `tests/repo` / `tests/meta` green |
| **Minor** — ADR-0042's note split `gda.models.ProjectRootedResult` across lines, rendering it as `gda.models .ProjectRootedResult` | Fully adopt | `a838a8ce9`, ADR-0042 — and the same note gains the clause that makes its resolution sentence true: the declared value must be one filename | Rendered check |
| **An unexpandable `~` escaped as a traceback** — found by this PR's own disclosure while fixing the normalizer, not by the reviewers. `Path.expanduser()` raises `RuntimeError` for `~unknownuser/…`, so `gda export smoke '~nosuchuser/x' --json` printed a traceback and exited 1 with no envelope — the same invariant the NUL refusal restores | Fully adopt, narrowly | `70d98fc10`, `commands/export.py`: `normalize_smoke_artifact_path` is total, following `gda.models.normalize_path`'s precedent for this input (#699). A `~` gda cannot expand names no user, so the value is kept, absolutized, and the ordinary resolution answers `export_artifact_not_found`. The guard is on the SMOKE's normalizer ALONE — not the shared helper, not `export run`'s | Operation-tier and CLI-tier `--json` regressions. **Reverse mutant** (remove the guard) → RED on both smoke tests; `36d0fa6f0` removes the test that pinned the unrelated `export run` defect |

### 2026-09-18 re-review closure

A backtrace found **local symptom-patch accumulation at artifact resolution**: `_is_runnable_file` guarded filesystem errors, but the preceding `Path.exists()` / `Path.is_dir()` probes could still raise `PermissionError`. The issue and accepted ADR already define the refusal contract; no broader support decision or new abstraction is needed. No NFR-driven platformization caused this failure, and this follow-up adds no runner, format classifier, result field, or platform obligation.

`36d0fa6f0` classifies confirmed absence (`ENOENT` / `ENOTDIR`) separately from paths the host cannot inspect, returning the existing typed `export_artifact_not_runnable`/4 for the latter. A real denied-directory CLI regression failed before the fix and passes after it; the complete export tier passes (221 passed, 1 skipped). Ruff and Pyright pass. `da1b5560a` corrects the class-versus-module ownership wording in `CONTEXT.md` and ADR-0042. The new review's Standards observation was also handled by deleting the test that pinned the old `export run --output` traceback.

The Spec reviewer's own first e2e attempt loaded an older installed CLI through a relative `PYTHONPATH` and passed on rerun — their environment, no action here.

**The three reproductions, after the fix** (real CLI, `--json`):

```text
$ gda export smoke /tmp/ext987/Abs.app --json          # CFBundleExecutable = /bin/echo
{"error":{"category":"operation","code":"export_artifact_not_runnable","message":"export artifact is not runnable: /tmp/ext987/Abs.app — the CFBundleExecutable it declares ('/bin/echo') is not one filename: it must name a single file directly under Contents/MacOS, so a value carrying a path separator, '.' or '..', or a NUL is refused","diagnostics":""}}   # exit 4

$ gda export smoke /tmp/ext987/Nul.app --json          # CFBundleExecutable = "Game\x00x"
{"error":{"category":"operation","code":"export_artifact_not_runnable","message":"export artifact is not runnable: /tmp/ext987/Nul.app — the CFBundleExecutable it declares ('Game\\x00x') is not one filename: …","diagnostics":""}}   # exit 4, no traceback

$ cd /tmp/ext987 && gda export smoke 'foo://game' --json
{"artifact":"/private/tmp/ext987/foo:/game","executable":"/private/tmp/ext987/foo:/game","exit_status":0,…}   # exit 0, both absolute
```

And the fourth, before and after `70d98fc10`:

```text
$ gda export smoke '~nosuchuser999/x' --json     # before
RuntimeError: Could not determine home directory.          # rich traceback, exit 1, no envelope

$ gda export smoke '~nosuchuser999/x' --json     # after
{"error":{"category":"operation","code":"export_artifact_not_found","message":"export artifact does not exist: /Users/…/~nosuchuser999/x","diagnostics":""}}   # exit 4
```

`export run`'s normalizer is untouched, asserted rather than assumed:
`normalize_export_output_path('~nosuchuser999/x')` still raises `RuntimeError`, and
`normalize_export_output_path('res://build/game')` still returns it verbatim.

## Mutant checks

Each mutant was applied to a saved copy of the file, `__pycache__` cleared, the named test run, then the file restored from that copy (never `git checkout --`) and `__pycache__` cleared again.

| Mutant | Test that went RED |
| --- | --- |
| An absent artifact returns `export_artifact_not_runnable` | `test_an_absent_artifact_is_not_found` |
| The `.app` suffix check is dropped (any directory accepted) | `test_a_plain_directory_is_not_runnable` |
| The execute-permission check is dropped | `test_a_file_without_execute_permission_is_not_runnable`, `test_a_bundle_whose_named_file_is_not_executable_is_not_runnable` |
| `--quit-after` is placed AFTER Godot's `--` | `test_a_positive_quit_after_is_placed_before_godots_separator` |
| `--quit-after 0` still emits the engine flag | `test_an_omitted_quit_after_adds_no_engine_flag` |
| `--strict` loses the `shutdown_leak` trigger | `test_strict_fails_a_zero_exit_that_leaked_at_shutdown` |
| `--strict` loses the non-zero-status trigger | `test_strict_fails_a_non_zero_exit_with_typed_evidence` |
| The `finally` cleanup is dropped | `test_a_private_root_is_created_before_the_launch_and_removed_after`, `test_the_private_root_is_removed_on_every_outcome_that_created_it` (4 params), `test_the_private_root_is_removed_when_the_launch_raises` |
| `ignore_errors=True` is dropped from the cleanup | `test_a_cleanup_failure_does_not_replace_the_outcome` |
| A caller's own root is replaced by a private one | `test_an_explicit_root_is_used_by_the_launch_and_is_not_removed`, `test_the_environment_twin_is_honoured_like_the_flag` |
| `launch()` ignores its explicit placement input | `test_an_explicit_root_is_not_overridden_by_the_flag_or_the_env`, `test_an_explicit_root_places_the_launch_like_the_process_wide_one` |
| An existing channel starts supplying a root | `test_only_the_artifact_smoke_hands_the_launch_an_explicit_root` |
| The shared completed-run validator is not applied | `test_result_truth_table_is_model_enforced`, `test_the_result_inherits_the_shared_bounded_stdout_truth_table` |
| The shared truth table is not published in the output schema | `test_the_schema_publishes_the_input_output_and_the_artifact_smoke_kind` |
| *(round 1, P2-1)* the artifact is not made absolute — back to `NormalizedPath` | `test_a_relative_artifact_resolves_against_the_invocation_cwd` (both tiers), `test_a_refusal_for_a_relative_artifact_names_the_absolute_path`, `test_the_strict_message_names_the_absolute_executable`, `test_a_relative_artifact_refusal_names_the_absolute_path` |
| *(round 1, P3-1)* the private root is created BEFORE the artifact resolves | `test_a_refused_artifact_creates_no_private_root_at_all` |
| *(external)* the `CFBundleExecutable` filename check is dropped | the 7 resolver cases, the NUL resolver test, and both CLI regressions (9 tests) |
| *(external)* the runnable guard is narrowed back to `OSError` | `test_a_path_the_syscall_cannot_carry_is_not_runnable` |
| *(external)* the smoke is pointed back at `export run`'s normalizer | `test_an_artifact_whose_name_contains_a_scheme_separator_is_still_a_file`, `test_a_scheme_like_artifact_name_is_a_filesystem_path` |
| *(external)* the smoke normalizer grows `export run`'s `://` branch | the same two |
| *(external)* `export run`'s own virtual-path convention is removed | `test_the_export_output_field_keeps_its_virtual_path_convention` plus its command suites |
| *(disclosure)* the smoke normalizer's totality guard is removed | `test_an_unexpandable_home_prefix_is_a_typed_refusal`, `test_an_unexpandable_home_prefix_answers_through_the_error_envelope` |

All 22 RED; the tree was verified clean afterwards.

## Deviations from the issue

- **The shared base is a field-less base class plus a core module, not a field-carrying base model.** ADR-0042 asks for "one shared result base"; AC10 asks for `script run`'s result bytes and schema shape to be byte-identical. Pydantic makes those two incompatible for a field-carrying base (base fields sort first, moving `path`), so the base carries the rule and each result carries its own fields — the repo's own `ProjectRootedResult` precedent, cited in the module docstring. The shared half (cap, projection, spill, truth table, validator) is genuinely shared, not copied.
- **`export smoke` declares no `--godot` either**, not just no `--project`. The engine it runs is the artifact's own, so the flag would name something the command never uses; accepting and ignoring it would be worse than refusing it.
- **The `stdout_spill_failed` registry description changed** (the issue made it conditional on "if its description generalizes"). It did: the code now has two consumers, so the row says "A completed run's stdout…" and names both. ADR-0002's row moved with it; this is not a schema delta (registry descriptions appear in no schema, and the byte diff above confirms it).

Nothing else deviates.

- **Linux and Windows are unmeasured**, per ADR-0042. The `.app` rule and the execute-permission rule are host-neutral code and are proved engine-free, but no exported Linux or Windows artifact was run. The README, the catalog, the SKILL and the command help all say so; the real-engine module skips off macOS rather than manufacturing evidence.
- **A `.app` bundle on a non-macOS host** would be resolved by the same declared-layout rule rather than refused as a platform mismatch. That is the deliberate no-platform-classifier choice, not an oversight, and it is unexercised on such a host.
- **The private-root cleanup is best-effort by contract.** A root that genuinely will not delete is left on disk; the outcome is unchanged and nothing reports it. The simulated-failure test pins that the outcome survives, but no test leaves a real undeletable directory behind.
- **`export run --smoke` composition, digests, PCK discovery, artifact-content inventory, provenance, receipts, a windowed mode, a completion marker and `smoke_aborted`** are all out, per ADR-0042; the negative gate reads only `export smoke`'s own surfaces, so #839's `export run` project-tree mutation report is untouched (asserted).
- The one pre-existing e2e skip named above is unrelated to this change.
- **`export run --output` keeps a pre-existing untyped traceback on `~unknownuser/…`**, and this PR deliberately does not touch it. `normalize_export_output_path` raises `RuntimeError` out of `Path.expanduser()` at the base commit already, so the exposure is a shipped command's, not this slice's; `gda.models.normalize_path` solved exactly this for every other path field in #699 by staying total. `export smoke`'s own normalizer is total (see above). The test that required the legacy `export run` behavior to remain broken was removed in `36d0fa6f0`. **A follow-up for the maintainer.**


qinhaihong-red:
## Reply to the adversarial review (round 1, head `4ddb49a35`)

Fixed at head `70d7e6707`. All eight findings were verified independently on a detached checkout before any change; the PR body has a "Review round 1" section mirroring this table.

| P2-1 `executable`, the human rendering, both refusal messages and the `smoke_failed` message echo the caller's RELATIVE path; only `artifact` was absolutized, and the PR body claimed the opposite | Fully Adopt | Reproduced by hand at `4ddb49a35`: `gda export smoke ../game.sh --json` → `"executable":"../game.sh"`; the module's own #403 rule (`normalize_export_output_path`) makes `export run`'s artifact path absolute for exactly this reason; the fix-shaped mutant survived the fast tier | `ExportSmokeParams.artifact` is now the module's `ExportOutputPath` (expanduser + cwd-absolute, applied BEFORE resolution), so `executable` and every message derive from an absolute path; the redundant `Path(artifact).absolute()` at the result site is gone; both relative-path tests assert `executable` and a refusal message; the implementation-shape pin on `resolve_artifact_executable` is replaced by a contract assertion (`37e08bd37`, plus `70d7e6707` for the e2e assertion that reads the bundle's declared `CFBundleExecutable` instead of the `.app` stem) | Reverse mutant (back to `NormalizedPath`) RED; the input schema is unchanged (both annotations are `Annotated[str, AfterValidator]`) — the schema/help delta re-measured against `0a749f116` is still exactly the four declared items |
| P2-2 ADR-0004's new note headline "a TENTH producer, and the first that is not `script run`'s" is false | Fully Adopt | `launch_timeout_failure` (shared by the sentinel / export / import channels), `export_templates_missing_failure`, the two `target_*` refusals and `path_case_mismatch_failure` precede it and are not `script run`'s; the test comment's narrower "or a pre-run refusal" still fails on `launch_timeout_failure` | The "first" clause is deleted from the ADR-0004 note (this PR's own text) and from the `test_error_registry.py` comment; no narrower "first" claim replaces it (`37e08bd37`) | Registry guard tests green |
| P3-1 nothing guards ADR-0042's "private root created AFTER the artifact resolves" | Fully Adopt | The reorder mutant survived 2910 fast tests and leaked a `gda-smoke-user-*` directory per refusal (reproduced with a redirected `TMPDIR`) | One operation-tier test: redirected temp dir, absent artifact, the directory stays empty (`37e08bd37`) | Reorder mutant RED |
| P3-2 `dispatch.py:157` "(a meta command)" and `headless.py:745` state an equivalence `export smoke` breaks | Fully Adopt | `grep -rn inherits_project=False src/gda` → four in `meta.py` plus the smoke; the routing is correct (recipe branch returns first), the prose is not | Docstring / comment edits only in both places (`37e08bd37`) | Behaviour unchanged; fast suite green |
| P3-3 two new public sentences call the `Export artifact` a "build" | Fully Adopt | `CONTEXT.md` `Export artifact`: `_Avoid_: build, bundle (macOS only), binary, release, receipt` | help "a smoked artifact"; catalog "a caller-selected artifact" (`37e08bd37`) | Help / catalog re-read |
| P3-4 the 120 s default is duplicated with nothing relating the two constants while three public sentences say they are the same | Partially Adopt | Both constants were `120.0` with a comment "restated rather than imported"; RULES.md's DRY rule wants ONE authoritative source for a fact the copy states as "the same completed-run ceiling" | Neither of the two offered options: one `DEFAULT_COMPLETED_RUN_TIMEOUT_SECONDS` in `gda.completed_run`, aliased by `script run`'s public name and used by the smoke; the "restated" comment is gone; no equality test (`37e08bd37`) | `tests/script/test_script_run_operation.py` still pins 120.0; help / catalog sentences unchanged and now structurally true |
| P3-5 `smoke_exit_status_failure`'s `artifact` parameter receives the executable | Fully Adopt | Called with `str(resolved)`; the result model distinguishes the two | Parameter renamed `executable`, docstring adjusted; message unchanged — it names what ran (`37e08bd37`) | Strict-path tests green |
| P3-6 the reworded `script run --help` sentence leaves a 16-character orphan line at 80 columns | Fully Adopt | `COLUMNS=80 gda script run --help` at `4ddb49a35` | Re-wrapped at the file's width (`37e08bd37`) | `script run --help` delta still inside declared item 4 |

Head after the fixes: `70d7e6707`. Gates at that head: ruff `All checks passed!`, format `521 files already formatted`, pyright `0 errors`, fast suite `2914 passed, 3 skipped`, e2e `tests/export` `36 passed, 1 skipped` (the smoke module ran, 15 tests), harness guard valid; the schema/help delta re-measured against `0a749f116` is still exactly the four declared items. The three sibling e2e tiers were not re-run at this head (the round-1 diff touches no sibling channel code; their figures stand at `4ddb49a35`, where the reviewer reproduced them) — round 2 re-runs them. Not verified: Linux / Windows artifacts remain unmeasured (ADR-0042's bound); nothing else outstanding from this round.

qinhaihong-red:
## Reply to the adversarial review (round 2, head `70d7e6707`)

Round 2 passed; its three P3 findings were verified and adopted at head `56668790a` (`56668790a`, prose and record only — no code path changed).

| P3-1 `commands/export.py` calls `ARTIFACT_SMOKE` "the fifth execution kind"; it is the sixth | Fully Adopt | `execution.py` has six members; `tests/meta/test_schema_aggregate.py` already says sixth | One word (`56668790a`) | Comment only |
| P3-2 the 120 s ceiling is re-spelled as a bare literal in the catalog and SKILL smoke rows, outside the single authority round 1 created | Fully Adopt | `docs/command-catalog.md` "(default 120, …)", `SKILL.md` "(default 120s)"; a `= 90.0` mutant makes `tests/script` RED but leaves the two rows silently false | The two smoke rows state the relation ("the same completed-run ceiling `script run` uses") without the number; `script run`'s own row keeps the one prose statement of it, beside the test that pins it; `--help` renders the constant (`56668790a`) | Catalog / SKILL re-read; fast suite green |
| P3-3 the PR body quoted the pre-fix ADR-0004 headline, counted "four" sibling channels (five), and called `commands/script.py` "no sibling channel's code" | Fully Adopt | PR body as of round 2 | Body corrected: the note quoted as it reads now, five channels, and the sibling e2e tiers stated as re-run in round 2 at `70d7e6707` (126 / 135 / 151 passed) | Record only |

Gates at `56668790a`: ruff `All checks passed!`, format `521 files already formatted`, pyright `0 errors`, fast suite `2914 passed, 3 skipped`; no executable code changed after `70d7e6707`, where round 2 re-ran the four e2e tiers and rebuilt the schema/help delta. The orchestrator's own re-verification on a detached checkout follows before the PR is marked ready.

qinhaihong-red:
## Orchestrator re-verification — READY at head `8f33b3196`

Independent re-run on a detached checkout at `8f33b3196` (this Mac, macOS arm64, Godot 4.6.3-stable, load 2.2–2.8; Godot suites under the machine-wide lock, foreground):

| Check | Result |
| --- | --- |
| `uv run --frozen ruff check .` / `ruff format --check .` | `All checks passed!` / `521 files already formatted` |
| `uv run --frozen pyright` (bare) | `0 errors, 0 warnings, 0 informations` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2914 passed, 3 skipped` |
| e2e `tests/export` (the smoke module ran against a real release-template `.app`) | `36 passed, 1 skipped` (the pre-existing Linux-only template skip) |
| `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` (no harness change) |
| PR CI at `8f33b3196` | ruff, pyright, unit: pass; `MERGEABLE` / `CLEAN` |

Review record: round 1 (0/2/6, all adopted, P3-4 partially) and round 2 (Pass, 0/0/3, adopted) by fresh independent reviewers at pinned heads; the orchestrator's own pass covered the design axis (no change), the spec axis (pass — every #979 criterion mapped to a named test or artifact) and the standards axis (five items adopted at `958615860` / `8f33b3196`, four declined with reasons in the PR body). `script run`'s result bytes, output schema shape and human rendering are byte-identical to the base; the schema/help delta is exactly the four declared items. Both ADR notes are dated `> **Outcome**` notes appended at the end; ADR-0042's two engine anchors are corrected in place as a citation erratum. Commits are Conventional with `(#979)`, author and committer `haihong.qin`, no attribution.

Not verified: Linux and Windows artifacts (ADR-0042's stated bound); `tests/scene tests/resource` and `tests/project tests/runtime` at this head (last run at `70d7e6707` by round 2; nothing since touches those channels).

Ready for the maintainer's merge decision: squash into `feat/gda-dogfooding-dev`, subject = the PR title, body blank.

qinhaihong-red:
## Artifact / Markdown review — Changes required

Reviewed PR #987 at `8f33b3196a6d3f685eb19cb77b963b71e831afb1` against `0a749f11673a1f2681169a5a82b90abafc23bac0`. Issue #979 is open with `ready-for-agent`; ADR-0042 and ADR-0040 are accepted. This is a fresh review of the artifacts and current implementation.

### Required changes

1. **The ADR's implementation outcome claims more than the resolver establishes.** ADR-0042:269-275 says bundle resolution shipped as declared, to `Contents/MacOS/<CFBundleExecutable>`. Its product contract (lines 63-66) and the new catalog text describe the executable inside the selected bundle. In `src/gda/commands/export.py:1430-1442`, an absolute plist value such as `/bin/echo` makes `Path.joinpath` return `/bin/echo`; I reproduced that result with a valid plist. The documented resolution rule is false for that accepted input. Validate `CFBundleExecutable` as a single executable filename before joining it to `Contents/MacOS`, and cover the refusal.

2. **The glossary now assigns fields to a fieldless base.** `CONTEXT.md:253-258` says the shared result base “contains” `exit_status`, stderr, diagnostics, and bounded stdout. ADR-0042:257-268 and `gda.completed_run.CompletedRunResult` explicitly say the base declares no fields; each command result declares them. This changes where readers think the public contract lives. Update the glossary's sentence to say that the two results share those fields while the base owns the common validator and schema rule.

### Minor

- ADR-0042:265-266 splits the inline code name across lines, rendering it as `gda.models .ProjectRootedResult`. Keep `gda.models.ProjectRootedResult` in one code span.

Sources: `CONTEXT.md`, issue #979, accepted ADR-0042/0040, the changed ADR notes, README translations, catalog, and bundled skill. Writing reference: ASD-STE100 Simplified Technical English. I read the complete issue and ADR-0042, inspected every changed prose passage and relevant surrounding sections, and checked term identity, technical nouns, scope, causal/order statements, and procedural wording. Apart from the findings above, no substantive terminology or prose issue was found. The two Godot 4.6.3 `main.cpp` anchors match the local tag; the README translation sync test passes. Linux and Windows exported-artifact behavior was not probed.

qinhaihong-red:
## Domain-modular architecture review — Changes required

**Authority and observed structure.** Issue #979 is open/`ready-for-agent`; ADR-0042 and ADR-0040 are accepted. `CONTEXT.md` defines Export artifact, Artifact smoke, Headless launch, and Raw run. The existing topology is retained: `commands/export.py` owns artifact resolution, private-root lifetime, strict policy, and the descriptor; `runner.py` owns placement and process launch; `completed_run.py` owns the shared bounded projection, schema rule, and rendering tail; `errors.py` owns failure builders. Imports point from command groups into the shared core, with no new group cycle or generic runner.

### Verified structural findings

1. **A projectful path convention leaked into the projectless artifact model.** `ExportSmokeParams.artifact` uses `ExportOutputPath` (`export.py:92-109,1463`), which preserves any string containing `://` for export's virtual-path convention. A real POSIX relative file addressed as `foo://game` remains `artifact="foo://game"` and resolves to relative `foo:/game`; I reproduced both values. The artifact result no longer owns a stable absolute filesystem address, contrary to issue #979 and ADR-0042. Give smoke a filesystem-only path normalizer in its own group.

2. **The artifact resolver does not enforce its executable boundary or typed refusal.** `export.py:1430-1442` treats any nonempty plist string as a bundle executable name. `CFBundleExecutable=/bin/echo` resolves outside the selected `.app`; a binary plist value containing NUL raises `ValueError` and the CLI exits 1 without a JSON envelope, rather than returning `export_artifact_not_runnable`/4. This leaves both the external effect and one failure path outside the command-owned rule. Require one usable filename before path joining, and map invalid names to the existing refusal.

**Structural checks.** Completeness fails at those addressing and refusal paths. Ownership, one-way dependencies, DRY, and the observed extension seam otherwise fit ADR-0040/0042; retain this topology and repair the two owners above.

**Load-bearing claims.** No project claim-evidence-state vocabulary was found, so both claims remain `open` under this review method: DDMA-987-1 (projectless input becomes an absolute artifact address; disconfirmed by `foo://game`) and DDMA-987-2 (bundle metadata selects an executable within the artifact or receives a typed refusal; disconfirmed by `/bin/echo` and NUL). The disconfirming inputs are the required regression cases. Focused tests passed (138 command/runtime/registry/i18n; 163 script/schema; 15 real macOS export-smoke e2e). Linux and Windows export runs were not measured.

qinhaihong-red:
## Standards — Pass

**0 actionable findings** at `8f33b3196a6d3f685eb19cb77b963b71e831afb1` versus `0a749f11673a1f2681169a5a82b90abafc23bac0`.

The independent Standards pass checked the changed production code and tests against `AGENTS.md`, `CLAUDE.md`, the workspace `RULES.md`, `CONTEXT.md`, ADR-0040/0042, `pyproject.toml`, and the review skill's Fowler smell baseline. Command policy remains in the group, shared launch and completed-run behavior are owned below it, and imports preserve the documented direction. The repeated result fields are deliberate under ADR-0042's fieldless-base decision, so they are not a Duplicated Code finding. No non-tooling-enforced standard breach or material smell was verified.

This axis was a source review; it did not independently run CI or runtime tests. Existing PR reviews and replies were not used.

qinhaihong-red:
## Spec — Changes required

Reviewed issue #979 against its higher-authority accepted ADR-0042, at head `8f33b3196a6d3f685eb19cb77b963b71e831afb1` and base `0a749f11673a1f2681169a5a82b90abafc23bac0`. **2 findings.**

1. **Bundle metadata can select a file outside the artifact or break the refusal contract.** Issue #979 requires that the `.app` executable resolve to `Contents/MacOS/<that name>` and that a present, unresolvable input return `export_artifact_not_runnable`. At `src/gda/commands/export.py:1430-1442`, `CFBundleExecutable=/bin/sh` resolves to `/bin/sh`, outside the selected bundle. A valid binary plist with a NUL character in `CFBundleExecutable` instead raises `ValueError` at `os.stat`; `gda export smoke --json` exits 1 without the typed JSON error or process exit 4. Require one valid executable filename before joining/stat, and add negative resolver/CLI regressions.

2. **A valid relative artifact path can stay relative.** Issue #979 says, “A relative `<artifact>` resolves against the invocation cwd.” `ExportSmokeParams.artifact` reuses `ExportOutputPath` (`src/gda/commands/export.py:92-109,1463`), which leaves any `://` string unchanged. With a runnable POSIX file at `foo:/game`, input `foo://game` remains the relative `artifact` value and resolves to relative `executable="foo:/game"`, so consumers outside the invocation cwd cannot use the promised absolute addresses. Use a smoke-specific filesystem normalizer.

The independent Spec pass compared the aggregate schema with the base and found only the declared deltas. Local validation passed 138 focused command/runtime/registry/i18n tests, 163 script/schema tests, and 15 real macOS export-smoke e2e tests. The first e2e invocation used a relative `PYTHONPATH` and loaded an older installed CLI in one subprocess; a full rerun with absolute `PYTHONPATH` passed. CI's Godot e2e job was skipped; Linux and Windows export runs were not measured.

Axis summary: Standards 0 (pass); Spec 2 (the bundle-resolution/refusal finding has the greater impact within Spec).

qinhaihong-red:
## Reply to the four independent reviews (Artifact, Architecture, Standards, Spec) at `8f33b3196`

Standards passed with no findings. The other three axes reported three distinct defects; I reproduced every one of them myself on a detached checkout at `8f33b3196` before deciding, and adopted all of them. Fixed at head `70d98fc10`.

| Bundle metadata can select a file outside the artifact, and one value breaks the refusal contract (Artifact 1 / Architecture 2 / Spec 1) | Fully Adopt | Reproduced: a `.app` declaring `CFBundleExecutable = /bin/echo` made `gda export smoke <app> --json` exit 0 and return `"executable":"/bin/echo"` with that program's own output — outside the artifact CONTEXT names as this command's trust subject, and the rule ADR-0042 publishes (`Contents/MacOS/<that name>`) is false for it. A binary plist declaring `Game\x00x` raised `ValueError` from `os.stat` inside `_is_runnable_file`, printed a traceback and exited 1 with no envelope | `_one_filename` requires ONE usable filename (non-blank, no separator, not `.`/`..`, no NUL) before the join; anything else is the existing `export_artifact_not_runnable` naming the rule. `_is_runnable_file` now guards `(OSError, ValueError)`, which closes the class rather than the instance. No sandbox, no containment check on the joined path, no identity notion (`a5088976f`) | Both inputs now return the typed refusal at exit 4; `../../../../bin/echo`, `sub/Game` and a blank value likewise. Reverse mutants: drop the filename check → 9 tests RED; narrow the guard back to `OSError` → RED |
| A projectful path convention leaked into the projectless artifact model (Architecture 1 / Spec 2) | Fully Adopt | Reproduced: a real POSIX file at `foo:/game`, addressed as `foo://game` from its parent directory, returned `"artifact":"foo://game","executable":"foo:/game"` — both relative — while the same file addressed as `foo:/game` returned both absolute. The cause is `normalize_export_output_path`'s `://` passthrough, which exists for `export run --output`'s virtual paths and has nothing to resolve against on a projectless command | The inherited exception is DELETED for this command rather than covered by a second rule: one shared `_absolute_filesystem_path`, `normalize_export_output_path` keeps its virtual-path guard in front of it, and the smoke gets `SmokeArtifactPath`, which has no exception to apply (`301a713d7`) | `foo://game` now yields both addresses absolute; `normalize_export_output_path('res://build/game')` is unchanged and `export run`'s own tests pass untouched. Reverse mutants: point the smoke back at `ExportOutputPath` → RED; give the smoke normalizer a `://` branch → RED; remove `export run`'s branch → its pin RED |
| The glossary assigns fields to a fieldless base (Artifact 2) | Fully Adopt | `CONTEXT.md`'s `Raw run` entry said "Their shared result base contains `exit_status`, stderr, diagnostics…", which is false of the shipped `CompletedRunResult` | The sentence now says both results carry those fields, each declaring them itself, while the base owns what is common: the validator, the schema rule and the human rendering. Nothing else in the entry changed (`a838a8ce9`) | Read back in place |
| ADR-0042 splits an inline code name across lines (Artifact, Minor) | Fully Adopt | Rendered as `gda.models .ProjectRootedResult` | One code span (`a838a8ce9`) | — |
| An unexpandable `~user` artifact escapes as a traceback | Fully Adopt (found by this round, not reported by the reviewers) | `gda export smoke '~nosuchuser/x' --json` raised `RuntimeError: Could not determine home directory.`, exit 1, no envelope — the same untyped-failure class the NUL refusal above restores | `normalize_smoke_artifact_path` is total for that input (the `gda.models.normalize_path` precedent, #699), so the ordinary `export_artifact_not_found` answers at exit 4 (``70d98fc10``) | Typed envelope at exit 4; `export run --output` is deliberately untouched and its behaviour is pinned by the same test |

Two notes on method. The second row is the SECOND round on the "published addresses must be absolute" family — round 1 chose to reuse `export run`'s normalizer and inherited its carve-out. I judge it converging because this remedy removes the inherited exception instead of laying another rule over it; a third occurrence in this family would go to a root-cause review rather than a third patch. And the Spec axis's own first e2e attempt loaded an older installed CLI through a relative `PYTHONPATH` and passed on rerun — no action taken.

One pre-existing defect is disclosed and NOT fixed here: `gda export run --output '~nosuchuser/x'` raises the same `RuntimeError` at the base commit `0a749f116`, so it is a shipped-command defect outside this slice. It is recorded for the maintainer as a follow-up.

Gates at `70d98fc10`: ruff `All checks passed!`, format `521 files already formatted`, pyright `0 errors`, fast suite `2931 passed, 3 skipped`, e2e `tests/export` `36 passed, 1 skipped` with the release-template smoke module running, harness guard valid, and the schema/help delta still exactly the four declared items — each re-measured by the maintainer on an independent detached checkout at this head.

qinhaihong-red:
## Re-verification after the external reviews — READY at head `70d98fc10`

Independent re-run on a detached checkout at `70d98fc10` (this Mac, macOS arm64, Godot 4.6.3-stable, load 2.6–3.7; Godot suites under the machine-wide lock, foreground):

| Check | Result |
| --- | --- |
| `ruff check .` / `ruff format --check .` | `All checks passed!` / `521 files already formatted` |
| `pyright` (bare) | `0 errors, 0 warnings, 0 informations` |
| `pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2931 passed, 3 skipped` |
| e2e `tests/export` (the smoke module ran against a real release-template `.app`) | `36 passed, 1 skipped` (the pre-existing Linux-only template skip) |
| PR CI at `70d98fc10` | ruff, pyright, unit: pass; `MERGEABLE` |

The four reported inputs, re-run through the real CLI at this head: an absolute `CFBundleExecutable` and a NUL-bearing one both return `export_artifact_not_runnable` at exit 4 naming the rule; `foo://game` returns absolute `artifact` and `executable`; an unexpandable `~user` returns `export_artifact_not_found` at exit 4. `export run`'s normalizer is unchanged — `res://build/game` passes through verbatim and the unknown-`~user` behaviour it has carried since before this branch is untouched, pinned by its own test.

Review record: two internal rounds (0/2/6 then 0/0/3) at pinned heads, the maintainer's design, spec and standards pass, and these four independent reviews — Standards passed; the three defects the others found were reproduced, adopted and fixed, and one more of the same class was found and fixed while verifying. Commits are Conventional with `(#979)`, author and committer `haihong.qin`, no attribution.

Disclosed and NOT fixed here: `gda export run --output '~nosuchuser/x'` raises the same `RuntimeError` at the base commit `0a749f116`, so it is a pre-existing defect on a shipped command and belongs to a follow-up, not to this slice.

Not verified: Linux and Windows exported artifacts (ADR-0042's stated bound); launching a `.app` off macOS; the best-effort private-root cleanup against a genuinely undeletable root. `tests/script`, `tests/scene`, `tests/resource` and `tests/project` were last run at earlier heads of this branch; the commits since touch `commands/export.py`'s smoke section and the two smoke test modules only, with the `export run` boundary pinned by test.

Ready for the maintainer's merge decision: squash into `feat/gda-dogfooding-dev`, subject = the PR title, body blank.

qinhaihong-red:
## Artifact / Markdown re-review — Changes required

Previous reviewed head: `8f33b3196`; current head: `70d98fc10d301a215c8270561debb05d71bdac9e`; base: `0a749f116`. Issue #979 is still open/`ready-for-agent`, and ADR-0042/0040 remain accepted. I checked the reply's claimed fixes against the new artifacts and executable behavior; the reply itself is not evidence.

**Verified fixes.** The new one-filename rule makes absolute, climbing, and NUL-bearing `CFBundleExecutable` values return `export_artifact_not_runnable`/4. `CONTEXT.md` now correctly says each command result declares its own completed-run fields. ADR-0042's split inline symbol is repaired. The `foo://game` path also becomes absolute.

### Required changes

1. **The revised ownership sentence still misidentifies the renderer's owner.** `CONTEXT.md:257-260` now says the shared result *base* owns the human rendering after the command's opening line. `CompletedRunResult` (`src/gda/completed_run.py:311-334`) only binds the schema hook and validator; rendering is the separate module function `render_completed_run` at line 278. ADR-0042:258-268 similarly attributes the cap, spill IO, timeout default, and rendering tail to the base rather than the module. This matters because these are architecture ownership statements. Say that the fieldless base binds schema validation, while `gda.completed_run` owns the projection, spill, default, and shared renderer. Split the long `CONTEXT.md` sentence accordingly.

2. **The artifact-refusal contract is still false for an inaccessible existing path.** Issue #979 and ADR-0042:63-68 require a present input that cannot resolve to a runnable file to return `export_artifact_not_runnable`. With an existing executable under a directory chmod 000, `export smoke --json` exits 1 with no JSON: `export.py:1495` lets `Path.exists()` raise `PermissionError` before the guarded runnable-file check. I reproduced this at the current head. Handle filesystem errors in the existence/type probe as the typed refusal, while keeping confirmed absence as `export_artifact_not_found`.

Sources: complete issue #979 and ADR-0042, `CONTEXT.md`'s affected glossary entries, ADR-0040, and the changed prose. Reference: ASD-STE100 Simplified Technical English. I checked term identity, class/module ownership, sentence structure, scope, and failure-path wording in every changed passage. Apart from finding 1, no substantive terminology or prose problem remains. Local export tests pass (220 passed, 1 skipped); Linux and Windows exported artifacts were not run.

qinhaihong-red:
## Domain-modular architecture re-review — Changes required

Pinned previous/current heads: `8f33b3196` → `70d98fc10d301a215c8270561debb05d71bdac9e`, against base `0a749f116`. Authority: `CONTEXT.md`, accepted ADR-0040/0042, and open/`ready-for-agent` issue #979. The existing topology remains: `commands/export.py` owns artifact addressing, bundle resolution, and command policy; `runner.py` owns process launch; `completed_run.py` owns shared completed-run mechanics; `errors.py` owns failure construction. The new helpers stay in the export group and introduce no cross-group cycle or generic runner.

**DDMA-987-1 (projectless absolute address): verified at this head.** `SmokeArtifactPath` removes the inherited virtual-path exception. A real `foo://game` relative file now yields absolute `artifact` and `executable` values; `export run` retains its own virtual-path rule.

**DDMA-987-2 (bundle metadata stays within its declared filename and refuses invalid values): verified at this head.** The one-filename check and guarded `os.stat` close the previous absolute-path and NUL counterexamples.

**DDMA-987-3 (every artifact-resolution failure has a command-owned outcome): open, with a counterexample.** `resolve_artifact_executable` calls `Path.exists()` at `export.py:1495` without catching `OSError`. An existing executable under a directory without search permission raises `PermissionError`; the CLI exits 1 with no JSON output instead of the typed `export_artifact_not_runnable`/4. Guard the existence and type probes in this resolver. No new owner or abstraction is needed.

**Structural checks and claim evidence.** Completeness still fails for that filesystem failure path. The ownership split, source-dependency direction, DRY, and observed extension seam otherwise fit ADR-0040/0042. No project claim-evidence-state vocabulary exists, so the skill's default claim state is `open`; this is distinct from the ADR's accepted lifecycle status. DDMA-987-1/2 have direct CLI and test evidence at this head; DDMA-987-3 is disconfirmed by the inaccessible-path input. Local validation passed 256 focused tests and 220 export tests (1 skipped). Linux and Windows exported artifacts remain unmeasured. The class-versus-module wording drift is reported in the Artifact comment.

qinhaihong-red:
## Standards re-review — 0 hard breaches, 1 minor judgement call

At `70d98fc10d301a215c8270561debb05d71bdac9e` versus `8f33b3196`, the independent Standards pass found no hard breach of `AGENTS.md`, `CLAUDE.md`, workspace `RULES.md`, ADR-0040/0042, or the documented code conventions. The new path and filename rules stay under the export group's ownership, and the shared dependency direction remains intact.

**Minor maintenance finding.** `tests/export/test_export_smoke_operation.py:442-448` asserts that `normalize_export_output_path("~nosuchuser999/x")` raises `RuntimeError`. This is an existing `export run` defect, yet the new test requires it to remain broken; a later fix must remove or rewrite the test. This works against the ease-of-change principle in `RULES.md:82`. It is a judgement call, not a hard violation. Remove this assertion while keeping the useful `res://` compatibility test at lines 451-456; track the legacy defect separately.

The Standards axis was a source review. The orchestrator independently ran the focused and export tests; existing review conclusions were not used as evidence.

qinhaihong-red:
## Spec re-review — Changes required

Reviewed issue #979 and higher-authority accepted ADR-0042 at current head `70d98fc10d301a215c8270561debb05d71bdac9e` (previous review `8f33b3196`, base `0a749f116`).

**Previously reported cases are fixed.** An absolute or NUL-bearing `CFBundleExecutable` now produces `export_artifact_not_runnable`/4; a valid relative `foo://game` path becomes absolute. I verified those inputs through the current CLI/model. The newly handled unexpandable `~user` path also gets a typed refusal.

### New required finding

**An existing but inaccessible artifact escapes the error contract.** Issue #979 says, “a present input that cannot be resolved to a host-runnable Godot executable is `export_artifact_not_runnable`” (classifier-source operation, exit 4). At `src/gda/commands/export.py:1495-1497`, `Path.exists()` runs before the guarded `_is_runnable_file`. On Python 3.13/macOS, I created an executable `blocked/game`, removed search permission from `blocked`, and invoked `gda export smoke <absolute-game> --json`: exit 1, empty stdout, uncaught `PermissionError`. Catch filesystem errors during the existence/type probes and return the existing typed not-runnable failure; reserve not-found for confirmed absence. Add one CLI regression.

Validation: 256 focused tests and the complete `tests/export` tier (220 passed, 1 skipped) passed at this head; the concrete inaccessible-path counterexample remains. CI lint, type check, and unit checks pass, while its Godot e2e job is skipped. Linux/Windows export behavior was not measured.

Axis summary: Standards 0 hard breaches plus 1 minor maintenance concern; Spec 1 required finding (the typed-refusal failure path).

qinhaihong-red:
## Re-review closure — `da1b5560a2a730a9cd39a59e1fdb0b4be9cfe208`

I took over the remaining Artifact, Architecture, Standards, and Spec findings from the re-review. The `backtrace-review` conclusion is **local symptom-patch accumulation at artifact resolution**: `_is_runnable_file` guarded filesystem errors, but the preceding `Path.exists()` / `Path.is_dir()` probes could still raise `PermissionError`. Issue #979 and accepted ADR-0042 already settle the refusal contract. I found no NFR-driven platformization causing this failure; the fix adds no generic runner, format classifier, result field, or new support promise.

- [`36d0fa6f0`](https://github.com/aigengame/godot-agent/commit/36d0fa6f0e9d3a1f0a7285fa873ddece5ca8dd40) uses one initial stat probe: confirmed absence (`ENOENT` / `ENOTDIR`) remains `export_artifact_not_found`; an uninspectable path returns the existing `export_artifact_not_runnable` envelope at exit 4. A real CLI test with denied parent-directory search permission failed before the change (exit 1, no JSON) and passes after it. The commit also removes the test that required the unrelated legacy `export run --output` traceback to persist.
- [`da1b5560a`](https://github.com/aigengame/godot-agent/commit/da1b5560a2a730a9cd39a59e1fdb0b4be9cfe208) corrects `CONTEXT.md` and ADR-0042: the fieldless `CompletedRunResult` base owns the validator and schema rule; the `gda.completed_run` module owns rendering, cap, projection, spill handling, and the shared timeout default.

**Validation:** 87 focused smoke tests passed; the complete export tier passed (221 passed, 1 skipped), including the real macOS export/smoke path. Ruff and Pyright passed locally; all required PR checks now pass. CI's Godot e2e job is skipped, and Linux/Windows exported artifacts remain unmeasured as already documented. The separate, pre-existing `export run --output` `~unknownuser` defect was not changed or pinned by this PR. The PR description has been reconciled with these changes.

qinhaihong-red:
## Re-verification of the 2026-09-18 re-review fixes — head `da1b5560a`

I took the three remaining findings and the two commits that answer them, and checked each against the running command rather than against the closure note.

| Finding | Verified at `da1b5560a` |
| --- | --- |
| An existing but inaccessible artifact escaped the error contract (Artifact 2 / Architecture DDMA-987-3 / Spec) | Fixed. An executable under a `chmod 000` directory returns `export_artifact_not_runnable` at exit 4, message `it could not be inspected ([Errno 13] Permission denied: …)`. The split is right where it matters: a path beneath a regular file and a broken symlink both still return `export_artifact_not_found` at exit 4, and an unreadable `Contents/Info.plist` returns the not-runnable refusal naming that rule |
| `CONTEXT.md` and ADR-0042 attributed module behaviour to the fieldless base (Artifact 1) | Fixed, and accurate against the code: `CompletedRunResult` binds only the schema hook and the projection validator, while `STDOUT_CAP`, `bounded_stdout`, `spill_failure`, `DEFAULT_COMPLETED_RUN_TIMEOUT_SECONDS` and `render_completed_run` are module-level in `gda.completed_run` |
| A test required `export run`'s legacy `RuntimeError` to persist (Standards, minor) | Removed, and nothing was lost: `normalize_smoke_artifact_path`'s docstring still records that `export run --output` keeps today's behaviour deliberately and why, so a later reader can still tell a deliberate line from an oversight — without a test that a future fix would have to delete |

No regression in the earlier fixes, re-run through the CLI at this head: an absolute and a climbing `CFBundleExecutable`, and a NUL-bearing one, all return `export_artifact_not_runnable`; an unexpandable `~user` returns `export_artifact_not_found`; `foo://game` returns absolute `artifact` and `executable`. `export run` is untouched — `res://build/game` passes through verbatim and its own unguarded `~unknownuser` behaviour is unchanged.

Gates measured by me on an independent detached checkout at `da1b5560a` (this Mac, macOS arm64, Godot 4.6.3-stable, load 2.1–2.6; the Godot tier under the machine-wide lock):

| Check | Result |
| --- | --- |
| `ruff check .` / `ruff format --check .` | `All checks passed!` / `521 files already formatted` |
| `pyright` (bare) | `0 errors, 0 warnings, 0 informations` |
| `pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2932 passed, 3 skipped` |
| e2e `tests/export` (the smoke module ran against a real release-template `.app`) | `36 passed, 1 skipped` |
| PR CI | ruff, pyright, unit: pass; `MERGEABLE` / `CLEAN` |

Two observations, neither of which asks for a change here:

- A NUL-bearing artifact PATH (reachable only through `--params-json`) shifts from `export_artifact_not_found` to `export_artifact_not_runnable`, because `Path.exists()` used to swallow the `ValueError` that `os.stat` now reports. Both are typed refusals at exit 4 inside the closed rule set, and no published sentence distinguishes them.
- From a deleted working directory every relative-path command raises `FileNotFoundError` out of `Path.cwd()` — `gda script run` and `gda export get` do the same at this head — so that exposure is repository-wide through `project.py`, not this slice's. It belongs with the `export run --output '~unknownuser/…'` follow-up.

Ready for the maintainer's merge decision: squash into `feat/gda-dogfooding-dev`, subject = the PR title, body blank.

Issue #985, opened 2026-09-16
Project tree inventory: one Python enumeration of the project tree, shared by export run and resource import


qinhaihong-red:
## Parent

Filed from the architecture review of promotion PR #984 (W4 of milestone #12, candidate 1). It supersedes the scope decision in #741's 2026-09-06 comment ("The accounting itself (`_project_files`, the before/after diff, `_summarize`) stays in the command as before"), which was taken by cost before the W4 rules existed; the other three review candidates were discarded by the user and are not filed.

## What to build

One `Project tree inventory`: the Python-side enumeration of a project's files has ONE owner, a shared core module beside `import_evidence`, and the two commands that inventory the tree read it instead of each walking on their own.

Today the same question is answered twice in Python. `gda export run`'s `Project-tree mutation report` walks with five rules (a directory link is followed as the engine reads it, once each by filesystem identity `(st_dev, st_ino)`, a cycle is not re-entered, only a regular file is opened, an unlistable or unreadable entry is counted once per project-relative spelling, the artifact is excluded by its parent's identity and name), and runs a capture → settle protocol (walk, run the engine, walk again, diff into created / modified / skipped). `gda resource import`'s `created` list runs the same before/after protocol on its own `rglob` walk, which is link-blind (a directory link is not descended on Python 3.13) and drops `.git` by its own rule — so on a project with a linked-in library, the same import pass reports different file sets to the two commands. Only `classify_created_file` / `CACHE_ROOT_REL` is shared. Two things answer a DIFFERENT question and stay where they are: the engine-side `res://` walk in the operations payload, which holds the decided policy for what the ENGINE reaches (ADR-0032 #760 / #804) and is not touched (ADR-0018); and `import_evidence`'s stale-sidecar gap scan, which enumerates `*.import` sidecars through `_engine_skips_directory_of` to predict what the engine's pass reaches — it keeps its own `rglob` and its predicate (the catalog records its link-blindness as an accepted under-promise).

[...]

PR #989, opened 2026-09-18
refactor(gda): one project tree inventory for export run and resource import (#985)

qinhaihong-red:
## Summary

The Python-side enumeration of a project's files has one owner. `gda export run`'s
`Project-tree mutation report` walked the tree under five rules (#839, PR #981);
`gda resource import` walked the same tree, around its own run of the same engine pass, with
its own `Path.rglob("*")`. One of those two walks descended a directory link and one did
not, so the same import pass reported two different file sets on a project with a linked-in
library.

Both now read the `Project tree inventory` — a new core module `gda.project_tree`, beside
`import_evidence` and below both command groups. `export run` is behaviour-preserving
(schema and all 94 `--help` pages byte-identical); `resource import` gains the rules it
lacked, which is the one behaviour change in this PR.

Closes #985

## What changed

**`src/gda/project_tree.py` (new).** The walk and the two-capture settlement, with the seven
rules stated once in the module docstring and one test naming each. What the two adapters
call:

- `ProjectTreeInventory.capture(project, *, artifact=None, detect_rewrites) ->
  …settle() -> ProjectTreeSettlement` — the capture → settle protocol, answering
  `created` / `modified` / `skipped` in project-relative paths. `artifact` is the file a run
  writes that must not read as a mutation of the project, excluded by its PARENT's
  `(st_dev, st_ino)` and its own name so an alias of that parent cannot smuggle it back in
  (`resource import` passes none — the pass writes no artifact). `detect_rewrites` is the ONE
  consumer-specific gate the scope guard allows: it decides whether the capture hashes the
  files outside the cache root, which is what `modified` needs and what a caller asking only
  "what did the pass create" must not pay for.
Turning an export DESTINATION STRING into that artifact `Path` — `res://` against the project,
another virtual scheme naming nothing here, a relative filesystem path against the project — is
export output-path policy and lives in `commands/export.py` as `_artifact_to_exclude`
(external review). The shared module takes the resolved `Path` and owns only what it does with
it: exclusion by the parent's filesystem identity and the file's own name.

Those two are **the whole of what a caller may say**, and the module docstring is where that is
stated — one home, so nothing can drift out of step with the signature (round 2, P3-1).
`_walk_project_files(project, *, artifact=None, on_unreadable_dir=None)` is the enumeration
underneath and is module-private: no adapter calls it (round 1, P3-1). Its `on_unreadable_dir`
sink is the module's own parameter — the capture and the settlement each supply one — and it
exists because `os.walk` swallows a listing error by default, so what an incomplete record
means is decided by the caller of the walk rather than by the walk.

No filter objects, no strategy seam, no walk options beyond those inputs, no filesystem
utility beyond the identity pair, and nothing for a third consumer.

**`src/gda/commands/export.py`** keeps only what the export knows: `ProjectTreeMutations` and
the two entry models, the renderer, the artifact it asks the inventory to exclude, and one
`_mutation_report()` that renders a settlement into the published `res://` shape and derives
the counts. `_walk_project`, `_file_facts`, `_project_tree_exclusions`, the
`_ProjectTreeExclusions` value object and `_PreExportInventory` are gone from it.

**`src/gda/commands/resource.py`** loses `_project_files` and reads the inventory instead.

**Tests**, drawn by reason to change (PR #836). `tests/project_tree/` holds 20: **18 of the 28
moved tests**, re-pointed at the module's own interface, plus 2 new ones (the rewrite gate, and
the engine skip markers with the `.gdignore`'d cache root as its case).
`tests/export/test_export_tree_mutations.py` holds 13 functions: **the other 11 moved tests**
plus 2 new (the group's destination resolution over every spelling, and the report's `res://`
projection with its counts). They drive `run_export_operation` / `render_export_run` and assert
on the published shape, counts, renderer and success-only boundary, so they change with
`commands/export.py`. The directory-link exclusion appears in both packages, because the one
original case covered two rules that now have different owners: the destination spelling here,
the identity exclusion there.
`tests/resource/` gains four unit tests (a file created under a link, a cycle, the `.git`
exclusion, and what the command asks the inventory for) and one real-engine e2e modelled on
PR #981's `assets -> ../shared` fixture.

**Docs.** `CONTEXT.md` gains the term. ADR-0040 gains a dated outcome note for the placement —
it needed one: point 5 sends a SHAPE no single group owns to `gda.models`, and this module
owns BEHAVIOUR, so the note records the `gda.completed_run` (#979) precedent applied a second
time. `classify_created_file`'s docstring named "the commands' project walkers", which no
longer exist as two. The dated supersession comment on #741 that AC6 asks for was posted by
the maintainer (comment 5726016530).

## The one behaviour change

`resource import`'s `created` list is no longer link-blind. Measured on one fixture, two
checkouts — a project with `assets -> ../shared` holding `sprite.png`, one `resource import
res://icon.png` run, Godot 4.6.3-stable:

```
BASE d715d50f9: created=10, nothing under the link
    res://.godot/imported/sprite.png-326bf2a…ctex     <- the linked asset's cache half
    res://.godot/imported/sprite.png-326bf2a….md5     <- reported
    …                                                  (assets/sprite.png.import: ON DISK,
                                                        reported nowhere)
HEAD:           created=11
    res://assets/sprite.png.import                     <- the other half, under the
                                                          link's spelling
```

The base reported the cache artefacts the pass wrote for the linked asset (they live inside
the project, so `rglob` saw them) and not the `.import` sidecar it wrote beside the source —
half of one asset's evidence. Pinned by a unit test and by
`test_the_pass_reports_what_it_created_under_a_directory_link` against the real engine.

**The `skipped` count is NOT published by `resource import`** — this is the one fact the
inventory holds that this command does not carry. The settlement computes it either way; the
result has no field for it, and adding one is a contract change #985 does not make.
`export run` publishes it as before.

## What deliberately did not change

- **The engine's two skip markers (#804) are not applied by the walk.** A nested
  `project.godot` and a `.gdignore` skip a directory in the engine's own scan; #804 gave the
  engine-side `res://` walk those same two markers, and `import_evidence`'s
  `_engine_skips_directory_of` predicts them plus the dot-prefix clause the engine adds. This
  walk takes none of it, and the reason is the engine's own bookkeeping: **Godot writes a
  `.gdignore` INTO the project data directory** — `res://.godot/.gdignore` is in `created` on
  every cold pass, visible in the base/head listing above — so the two markers ALONE would
  prune the cache root, empty the `cache_owned` half of both commands' `created` lists, and
  narrow the published "anywhere under the project" on the four surfaces that say it: the
  `created` field description and the `export run` help, the catalog, the SKILL `export` row,
  and `CONTEXT.md`. The first is proven unchanged by the byte-identical schema/help diff; the
  rest are untouched, and still true. A dot-prefixed directory stays in for the separate
  reason #54 and #712 decided, which is the engine-side walk's rule too.
  `test_the_engines_skip_markers_are_not_applied` carries the case — a `.gdignore` inside the
  cache root with a cache file created beside it — so both marker mutants below read RED in
  the fast tier.
- **`import_evidence`'s stale-sidecar gap scan** keeps its own `rglob("*.import")` and
  `_engine_skips_directory_of`: it predicts engine reachability, a different question, and the
  catalog already records its link-blindness as an accepted under-promise. It is the grep
  gate's one named exception.
- **The engine-side `res://` walk** in `operations.gd` is untouched (ADR-0018). No harness
  change: `scripts/harness_version_guard.py` reports "harness version transition is valid".
- **No published sentence changes.** The README and SKILL `resource import` rows state no walk
  rule for the `created` list, so no prose reconciliation is owed there. The catalog calls the
  real run's `created` list "the authoritative inventory" and the field says "Every file the
  pass created": this PR makes that list strictly MORE complete — a linked-in subtree is
  enumerated now and was not — but it does not make the unconditional wording verifiable. The
  inventory can report `skipped > 0` and `resource import` discards it, as AC3 requires (no new
  field), so when a subtree is unreadable the result carries no completeness signal. The
  mismatch predates this PR and this slice does not close it; it is recorded in Residual risk
  as an open contract question.

## Schema / help

No delta, and none was declared. `gda schema` plus all 94 `--help` pages are **byte-identical**
to a detached base checkout at `d715d50f9`: 95 pages dumped from each, `diff` empty. Re-run at
`51ce120f5`, after the external review moved `_artifact_to_exclude` between
modules.
`gda schema` was compared with the same base again at `9bfff827c`; the complete help
comparison remains the run at `51ce120f5`.

PR CI runs: ruff lint, ruff format, pyright, the harness version guard, and
`pytest -m "not e2e" -n 4 --dist loadgroup`. It runs **no Godot e2e** — the `godot-e2e` job is
nightly / `workflow_dispatch` only, so the three e2e suites below, the schema/help
byte-identity, the grep gate, the mutant checks and the cost numbers are local evidence from
this PR and are re-run in the wave's e2e dispatch.

## Review round 1

Six findings at `e4b229f4c`, all adopted.

| Finding | Decision | Change (commit, file) | Validation |
| --- | --- | --- | --- |
| P2-1 — AC6's dated #741 comment missing | Handled by the maintainer, not this PR | issue #741, comment 5726016530 | The "Docs" paragraph above now names it |
| P2-2 — `CONTEXT.md` mis-states what the engine-side walk answers, and the marker bound cites the wrong decision | Fully adopted | `db609e189` — `CONTEXT.md`, `project_tree.py` rule 7, and the rule's test comment. Says what #804 did (the two markers, dot-prefixed directories kept in), keeps `(#54, #712)` on the dot-prefix question only, and gives the bound its real reason: the engine's own `.gdignore` inside the project data directory. `import_evidence`'s sentence is accurate and untouched | `test_the_engines_skip_markers_are_not_applied` now seeds `.godot/.gdignore` and asserts the cache file created beside it is still `cache_owned`. The reviewer's mutant — prune a dot-prefixed directory that holds a `.gdignore` — was green in the fast tier and RED only against the real engine; it now reads **RED in the fast tier** |
| P3-1 — `walk_project_files` public with no adapter caller | Fully adopted, narrowly | `ef18cd799` — renamed `_walk_project_files`; one test monkeypatch updated; the PR's "two questions" claim dropped. `FileFacts` keeps its name: it is the element type of a public attribute, not an unused public function | Fast suite green; the grep below finds no consumer |
| P3-2 — `detect_rewrites=False` pinned by nothing | Fully adopted | `7ff1dd37a` — `tests/resource/test_resource_import_commands.py` records the kwargs the command passes to `ProjectTreeInventory.capture`, inside the test that already drives a real pass | Reverse mutant (flip to `True`) → **RED**, `AssertionError: [{'detect_rewrites': True}]`. It was green over the whole fast tier before |
| P3-3 — "a link **at** a regular file" | Fully adopted | `db609e189` — `project_tree.py`, "to" restored as the base sentence had it | Prose |
| P3-4 — the export report's tests split across two packages | Adopted; **#985's AC2 amended by the issue owner**, not deviated from | `2bbbb5f33` — the 10 export-adapter tests move back to `tests/export/test_export_tree_mutations.py`; `tests/project_tree/` keeps the 19 rule tests plus the 3 new ones | PR #836's rule: those 10 change when `commands/export.py`'s report changes. AC2's "not split across the two consumers' packages" still holds — no rule test lands in a consumer's package |

The same review listed four non-blocking observations rather than findings. All four were
real; all four are fixed in `48359eb46`, prose and one comment only, no code path touched.

| Observation | Change |
| --- | --- |
| The `CONTEXT.md` entry said "one set of rules" and listed six, while the module states seven | The lead-in no longer promises a count: "It walks under the rules the module states" |
| The same entry listed three per-call inputs, then introduced the rewrite gate as a fourth | The list now holds all four. It has to be complete: "the per-call inputs are the two questions' arguments, not options" is #985's own bound |
| ADR-0040's new note dropped the relative pronoun the #979 note above it uses | "the `gda.models` core **that** point 5 names" |
| `commands/export.py`'s section comment said "the two results the same engine pass produces" | Each command runs its own pass: "the two results each command's own engine pass produces" |

## Review round 2

Pass — 0 P1 / 0 P2 / 4 P3 at `48359eb46`, all non-blocking, all adopted.

| Finding | Decision | Change | Validation |
| --- | --- | --- | --- |
| P3-1 — `CONTEXT.md` and the module docstring each state a per-call input list that matches no actual call (one lists the walk's sink, the other omits the rewrite gate) | Fully adopted, and fixed as a DRY defect rather than as two typos: one fact, one home | `f1935fa86` — the module docstring is the authority and states `ProjectTreeInventory.capture`'s three arguments plus the one clause that the unreadable-directory sink is the private walk's own parameter. `CONTEXT.md` names the same three as concepts and enumerates nothing the module owns. The private walk's docstring describes its parameters as the module's, not as the caller bound | `grep` finds no second enumeration to drift; both adapters' call sites match the stated three |
| P3-2 — two relationships stated wrongly in the `CONTEXT.md` entry | Fully adopted | `f1935fa86` — "rather than under two different walks, one per command" (the inventory does still walk twice, once per capture); and "The inventory's walk takes neither marker", whose nearest antecedent had been the engine-side walk, which takes both | Prose |
| P3-3 — the engine-writes-its-own-`.gdignore` fact restated in three places with no pointer to where it was decided | Fully adopted | `f1935fa86` — rule 7 in the module is the statement and cites ADR-0032's #804 amendment, which carries the engine source; the glossary entry and the rule's test cite it instead of re-arguing it. No engine line numbers copied | Prose; the rule itself is still pinned by the two marker mutants below |
| P3-4 — an export-package test monkeypatches `gda.project_tree._walk_project_files`, a private name in another module | Fully adopted, fixed at the test | `366a21adc` — `test_a_failed_export_reports_no_mutations_and_pays_for_no_second_walk` counts `ProjectTreeInventory.settle`, the public method `classify_export_run` itself calls. It stays in `tests/export/` because it is an export-recipe test: the claim is the recipe's success-only boundary, not a walk rule. The evidence is the same one level up — no settlement on the failure branch, exactly one on the success branch | Mutant: settle before returning `export_failed` → **RED**, `assert [ProjectTreeI...] == []`. `grep -rn "_walk_project_files" tests/` now returns nothing |

The same review raised three non-blocking observations. One was a PR-body error and is fixed
above: `tests/project_tree/` holds 19 moved tests plus **2** new ones, not 3 — the `.gdignore`'d
cache root is a case inside the skip-markers test, added by round 1's P2-2 fix. The other two
are not changed here and both were reasoned: the destination conversion is a second `res://`
resolution beside `gda.project`'s canonical one, inherited verbatim from the base — reusing the
canonical resolver would not be behaviour-preserving, so it belongs in a follow-up, not in this
PR (the external review moved WHERE it lives, below, but not what it does); and `FileFacts`
stays public as the element type of `ProjectTreeInventory.files`.

## External review

Five independent reviews at `366a21adc` — Artifact, Domain-modular architecture, Standards,
Spec, and a Backtrace review. Six distinct findings were resolved or explicitly
deferred; the maintainer reproduced both code findings before deciding.

| Finding | Decision | Change | Validation |
| --- | --- | --- | --- |
| Spec 1 — an inaccessible directory reached through two spellings counts twice in `skipped`, while the module says "counted once" | Partially adopted: **prose, no behaviour change**. The maintainer reproduced the same count 2 from `_PreExportInventory` at the base, so this is `export run`'s shipped behaviour, not a regression — and `skipped` is a published field, so making the count identity-unique is an observable change that belongs in its own slice | `51ce120f5` — rule 4 states what the code counts (once per project-relative spelling; `os.walk` reports a listing error instead of yielding the directory, so rule 1's identity test is never asked about it) and says what the count is for. `CONTEXT.md` mirrors the one clause | Nothing executable changed, so the report stays byte-identical; the behaviour question is recorded in Residual risk |
| Architecture 1 / Standards / Spec 2 — the shared module interprets export destination syntax (`artifact_to_exclude`), with `commands/export.py` as its only caller | Fully adopted | `2d508bb45` — the conversion moves to `commands/export.py` as `_artifact_to_exclude`. `gda.project_tree` keeps `artifact: Path | None` and the identity-plus-name exclusion, which is the walk's rule. The tests follow: `tests/export/` gains a direct test of the conversion over every spelling and takes back the two destination-string cases; `tests/project_tree/` keeps the identity cases and passes the artifact as a `Path` | Reverse mutant: break the `res://` branch in `commands/export.py` → **RED**, 3 export tests. Schema/help still byte-identical; `tests/export` and `tests/resource` e2e green |
| Artifact 1 — the body's "it cannot exceed the old walk's cost" is unsupported | Fully adopted, body only | The general upper bound is deleted from Residual risk. The old `rglob` walk did not descend a directory link and the new `os.walk(followlinks=True)` does, so on a linked-in tree the new walk enumerates strictly more | The measured same-fixture numbers are kept and the linked-tree cost is stated as unmeasured |
| Spec 3 — #985's cost criterion cites a warm **macOS preset** release export; the benchmark used the `Linux/X11` preset on a macOS host | Fully adopted, closed by measuring | No code change. The same generated fixture was exported with a real macOS preset (universal, no codesign, `import_etc2_astc` on), under the lock | Warm macOS release export **13.523 s**, inventory **0.509 s** → **+3.9 %**, inside #839's +10 % bound. Both presets are now in the Cost section |
| Artifact Minor — `CONTEXT.md`'s "those three" lost its antecedent when the list moved into the module | Fully adopted | `51ce120f5` — the scope-guard sentence names them: the project, the artifact and the rewrite gate | Prose |
| Artifact 2 / Architecture 2 / Backtrace — the PR claimed this change makes the catalog's "authoritative inventory" wording true; it does not, because `resource import` discards the inventory's `skipped` | Fully adopted, body only. No field, no code, no description change — AC3 forbids them and the question is the maintainer's | The claim of repair is removed. The accurate statement replaces it: the list is strictly more complete than before, the completeness signal stays unpublished for this command as AC3 requires, and the unconditional wording is therefore unverifiable when `skipped > 0` | Recorded in Residual risk as an open contract question, with what a decision would have to settle |

The Backtrace review found no symptom-patch accumulation and no NFR platformization, and
recommends keeping the boundary as it is.

## Validation

Validation head `51ce120f5`. Host: macOS (Darwin 25.6.0), Godot 4.6.3-stable, load average 1.7–2.5
across the runs. Every Godot-spawning run serialized through the machine-wide e2e lock, in the
foreground.

The external review moved executable code (`_artifact_to_exclude`), so everything below was
re-measured at that head except the `tests/project` e2e suite, which the diff does not
reach and whose row names the head it was measured at.

The follow-up `9bfff827c` changes only the settlement docstring and its test comment.
At that head, 20 project-tree tests, Ruff check/format on the touched files,
`git diff --check`, and `gda schema` against the base passed. The engine and cost
measurements below remain attributed to `51ce120f5`.

| Command | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `524 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2940 passed, 3 skipped, 1 warning in 19.93s` |
| `pytest -m e2e tests/export -q -rs` | `36 passed, 1 skipped, 170 deselected in 75.79s` |
| `pytest -m e2e tests/resource -q -rs` | `44 passed, 118 deselected in 128.52s` |
| `pytest -m e2e tests/project -q -rs` (at `7ff1dd37a`) | `118 passed, 293 deselected in 132.35s` |
| `gda schema` + 94 `--help` vs `d715d50f9` | 95 pages, `diff` empty |

The one skip is the pre-existing Linux-only templates case.

Grep gate — `os.walk` / `rglob` under `src/gda`, code lines only:

```
src/gda/project_tree.py:255:    for dirpath, dirnames, filenames in os.walk(
src/gda/import_evidence.py:410:    for sidecar in sorted(project.rglob("*.import")):
```

The module and the sidecar scan, nothing else. The other matches are prose in docstrings and
comments that NAME the two (`commands/resource.py` records what the command used to do).

## Cost

Generated fixture, never checked in: 10,003 files — 10,000 `.tres` resources of ~0.9 KB in 100
directories, plus `project.godot`, `export_presets.cfg` and one PNG — and ~30,015 files after
the export has written its cache. `--mode release`, warm, medians of 3 runs, load 1.7–2.2. The
round-1 reviewer reproduced every `Linux/X11` figure independently on their own fixture.

- **`export run`, macOS preset** (universal, no codesign, `import_etc2_astc` on — what #985's
  criterion names). A warm release export takes **13.523 s** end to end, of which the inventory
  (capture + settle, hashing included) is **0.509 s**: **+3.9 % over the export's own 13.01 s of
  work**, inside #839's +10 % bound. Measured at `51ce120f5`, on the host's real 4.6.3 macOS
  templates.
- **`export run`, Linux/X11 preset** (the earlier measurement, kept because the base comparison
  was taken on it). Warm export 12.505 s, inventory 0.519 s: **+4.3 %**. The base's
  `_PreExportInventory` on the same fixture: **0.524 s** vs this module's **0.512 s** — the same
  work, moved. Taken at `e4b229f4c`; no commit since has changed a line the walk executes.
- **`resource import`.** The command with a real pass: 3.215 s. Its accounting went from
  0.780 s (`rglob` × 2) to **0.262 s** (one capture that stats each file once, and a settlement
  that stats only what is new): **−0.519 s, −16.1 %** of the command's wall clock. Not slower at
  all, so the "no more than the walk's own share" bound is met with room; the walk is 8.1 % of
  the command.

## Mutant checks

Last run at `51ce120f5`. Each mutant applied to one file alone, with a saved-copy restore and `__pycache__` cleared after every edit, run against `tests/project_tree`, `tests/resource/test_resource_import_commands.py`
and `tests/export`; `git status` empty after each. All eleven read RED — two of them applied to
a command module rather than to the shared one.

| Mutant | Dead test(s) | |
| --- | --- | --- |
| `followlinks=True` → `False` | 5: `test_an_artifact_under_a_directory_link_is_excluded_by_identity`, `test_a_directory_link_is_walked_as_the_engine_reads_it`, resource `test_a_file_created_under_a_directory_link_is_reported`, export `…excluded_by_identity[True]`, `[False]` | RED |
| identity set off (`walked.add` → `pass`) | 2: `test_a_link_that_leads_back_up_the_chain_is_not_re_entered`, resource `…created_under_a_directory_link…` | RED |
| `S_ISREG` guard removed | 1: `test_a_non_regular_entry_is_counted_and_never_opened` | RED |
| `res://` resolution off — now in `commands/export.py` (external review) | 3, all in `tests/export/`: `test_the_export_destination_resolves_to_the_artifact_kept_out`, `test_a_res_output_artifact_is_the_output_not_a_mutation`, `…excluded_by_identity[True]` | RED |
| artifact parent matched by path spelling instead of `(st_dev, st_ino)` | 1: `test_an_artifact_is_excluded_when_the_walk_uses_another_link_spelling` | RED |
| `.git` exclusion off | 3: `test_a_top_level_git_directory_is_not_walked`, `test_the_exclusions_match_whole_path_components`, resource `test_a_top_level_git_directory_is_still_excluded` | RED |
| the engine's two skip markers applied | 1: `test_the_engines_skip_markers_are_not_applied` | RED |
| a `.gdignore` prunes a dot-prefixed directory (the reviewer's mutant) | 1: `test_the_engines_skip_markers_are_not_applied` | RED |
| rewrite gate ignored inside the module | 1: `test_a_capture_without_rewrites_reports_only_what_was_created` | RED |
| `resource import` asks for rewrites (round 1, P3-2) | 1: `test_missing_asset_runs_the_pass_and_reports_created_classified` | RED |
| the recipe settles before returning `export_failed` (round 2, P3-4) | 1: `test_a_failed_export_reports_no_mutations_and_pays_for_no_second_walk` | RED |

The first five are #839's round-3 mutants, re-read against the module. The two marker mutants
kill exactly the test that names rule 7, and they kill it through the cache root: a project
whose `.godot` holds no `.gdignore` is untouched by a marker rule, which is why the test seeds
the marker the real engine always writes.

## Deviations from the issue

None. One naming judgment worth stating: the module is `gda/project_tree.py` with the class
`ProjectTreeInventory`, rather than a module spelled for the whole term. The term is in the
class name and the module docstring's first line, and the test package is the
`tests/project_tree/` the issue's AC names as its example.

- The cost numbers are from a fixture generated for this PR, not #839's, so +3.9 % / +4.3 % are
  not directly comparable to that PR's +1.8 % — the fixture's file sizes and directory shape
  differ. What IS comparable is the base-vs-head inventory on the same tree (0.524 s vs
  0.512 s), which is what "behaviour-preserving" is being claimed about; that pair was taken at
  `e4b229f4c` and not re-run.
- The real-engine evidence is macOS-only. The link-following behaviour rests on `os.walk` and
  `stat`, which are POSIX-portable; Windows directory junctions are covered by no test here and
  were not covered before.
- The `resource import` accounting is faster on this fixture partly because the settlement stats
  only new files. A tree where the pass creates a very large share of the files would narrow
  that margin. **No general upper bound is claimed**: the old walk used `Path.rglob`, which does
  not descend a directory link, and the new one uses `os.walk(followlinks=True)`, so on a
  linked-in tree the new walk enumerates strictly more files than the old one ever did. The cost
  on such a tree is unmeasured — the measurement above is same-fixture, and that fixture holds
  no link.
- **Open contract question, for the maintainer, not closed here:** `resource import`'s published
  `created` says "Every file the pass created" and the catalog calls it authoritative, while the
  inventory's `skipped` stays unpublished for this command (AC3). When a subtree is unreadable
  the result therefore carries no signal that the list is partial, and the absolute claim cannot
  be checked from the result. It predates this PR, the link fix does not close it, and closing
  it needs a decision — qualify the promise, or publish a completeness signal — that #985's
  "no new field / no published sentence changes" constraint currently forbids.
  Tracked by #990.
- An inaccessible directory reached through two spellings counts twice in `skipped`, for
  `export run`'s published field. The base counts the same 2 on the same tree, so this PR
  changes nothing; rule 4 now STATES it rather than claiming "counted once". Making the count
  identity-unique is an observable change to a published field and belongs in its own slice.


qinhaihong-red:
## Orchestrator re-verification — READY at head `366a21adc`

Independent re-run on a detached checkout at `366a21adc` (this Mac, macOS arm64, Godot 4.6.3-stable, load 1.8–2.2; the Godot tiers under the machine-wide lock, foreground):

| Check | Result |
| --- | --- |
| `ruff check .` / `ruff format --check .` | `All checks passed!` / `524 files already formatted` |
| `pyright` (bare) | `0 errors, 0 warnings, 0 informations` |
| `pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2938 passed, 3 skipped` |
| e2e `tests/export` + `tests/resource` | `80 passed, 1 skipped` (the skip is the pre-existing Linux-only templates case) |
| `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` — no harness change |
| `gda schema`, root help and all 78 command helps, rebuilt on BOTH sides | **byte-identical** to the base `d715d50f9` |
| Grep gate | `src/gda` holds the module's own `os.walk` and `import_evidence`'s `rglob("*.import")` sidecar scan, the one named exception |

The declared-nothing surface is the load-bearing claim of this slice, so I rebuilt it rather than trusting the diff: `diff -r` over both dumps is empty.

The one behaviour change is real and bounded. On identical fixtures at base and head — a project whose `assets` is a link to a directory outside it, then a real `gda resource import` of an asset under that link — the base reports `created` without the `.import` sidecar the pass had just written to disk, and the head reports it under the link's spelling. No new field appears; `skipped` stays unpublished for that command.

The bound the slice exists to draw holds: the engine's two skip markers are not applied, so the cache root is still walked and the `cache_owned` half of both commands' `created` lists is intact, and the four published sentences that say the report names files "anywhere under the project" stay true. `import_evidence`'s reachability predicate and its sidecar scan are untouched.

Review record: two independent rounds at pinned heads — round 1 (0/2/6 with four observations) and round 2 (Pass, 0/0/4) — every finding adopted. Round 2's interface finding was fixed by giving the fact one authority (the module states `ProjectTreeInventory.capture`'s three arguments; the glossary defers to it) rather than editing two drifting prose lists a third time. Commits are Conventional with `(#985)`, author and committer `haihong.qin`, no attribution.

Not verified: Linux and Windows (`os.walk(followlinks=True)` and `stat` are POSIX-portable, but Windows directory junctions are covered by no test, before or after); the cost numbers were measured by the implementer and independently reproduced by both reviewers, not re-measured by me at this head; the daemon, live and harness e2e tiers were not run — nothing in the diff reaches them and the harness guard reports no change.

Ready for the maintainer's merge decision: squash into `feat/gda-dogfooding-dev`, subject = the PR title, body blank.

qinhaihong-red:

Reviewed issue #985 (including its amendment), the PR description, the changed `CONTEXT.md` term, and ADR-0040 at head `366a21adc`. I checked their factual claims against #668/#839 and the current code; existing PR reviews/replies were not evidence.

### Required changes

1. **The PR's cost upper bound is unsupported.** “It cannot exceed the old walk's cost” (PR description, Residual risk) does not follow from the implementation. The base `resource.py::_project_files` uses `Path.rglob` and does not descend directory links; `project_tree.py::_walk_project_files` uses `os.walk(..., followlinks=True)`. In a 300-file linked-directory fixture, the old walk enumerates 0 files through the link and the new walk enumerates 300. Keep the measured same-fixture numbers, remove the general upper bound, and treat linked-tree cost as unmeasured unless it is measured separately.

2. **The completeness claim conflicts with the accepted scope.** The PR says this change makes the catalog's “authoritative inventory” claim true, while #985 AC3 expressly leaves `resource import`'s `skipped` count unpublished. `ProjectTreeInventory.settle()` can return `skipped > 0`, and `resource.py:1101-1108` discards it; the published `created` description still says “Every file the pass created” and the catalog calls it authoritative. With a mode-000 directory, the inventory reports `skipped=1`; a real Godot `resource import` succeeds but returns no completeness signal. This does not prove that Godot created an omitted file in that fixture; it leaves the unconditional completeness claim unsubstantiated. The wording predates this PR, but the PR explicitly claims to have repaired it. The issue owner should decide the required behavior for an incomplete walk, then reconcile #985's “no published sentence changes” constraint and the public descriptions.

### Minor

- `CONTEXT.md:116-117`: “those three” has no nearby clear antecedent. Name the three capture inputs directly so the scope rule is actionable.

**Terminology and prose assessment:** I used `CONTEXT.md`, ADR-0040/0032, #985, and the published command descriptions as term sources, with ASD-STE100 as the technical-prose reference. I inspected all changed document passages, the full issue and PR description, and the claims they make about code and tests; I checked term consistency, referents, and causal/scope wording. “Project tree inventory” and the separate engine-side “project walk” are otherwise consistent; no other substantive terminology or prose issue found. I did not rerun the 10,000-file cost benchmark or every real-engine suite.

qinhaihong-red:
## Domain modular architecture review — Changes required

At `366a21adc`, the observed dependency path is `commands/export.py` and `commands/resource.py` → `gda.project_tree` → `gda.import_evidence`. This follows ADR-0040's one-way group-to-core direction; the shared walk and settlement remove the duplicate file-set owner. I found no group-to-group dependency or need for a new context, plugin seam, or generic filesystem layer.

1. **Export output syntax has moved into the shared owner.** `project_tree.py:200-220` exports `artifact_to_exclude(project, output_path)` and interprets `res://`, other schemes, and relative export destinations. Only `export run` calls it (`commands/export.py:1024-1027`). #985 says the shared capture receives an optional artifact `Path` while the export group keeps the artifact; ADR-0040 assigns group-specific behavior to its command module. An export output-policy change now edits the shared inventory. Resolve the destination to a `Path | None` in `commands/export.py`, then pass that path to the existing shared capture interface. Keep the identity-based artifact exclusion inside `project_tree`.

2. **The completeness decision has no public owner in the resource composition.** The core computes `settlement.skipped`; export publishes it, but `resource import` discards it while its published `created` contract says every file. The shared inventory cannot make that public guarantee by itself. Before choosing an adapter change, settle whether `resource import` promises a complete list when a subtree is unreadable; #985 currently combines that guarantee with “no new field/no published sentence changes.”

Structural checks: completeness is open at the resource result boundary; orthogonality and DRY improve for the two walks but are weakened by export-only path interpretation in core; the existing three-input capture is sufficient for current extensibility. Load-bearing claims (the repository defines no claim-evidence-state vocabulary, so both are `open`): **PT-1**, the core is the sole Python inventory owner — supported by imports and the source scan, disconfirmed if another command walk remains; **PT-2**, both public adapters preserve the inventory's completeness meaning — contradicted by the discarded `skipped`, and requiring the issue owner's scope decision. Sources: current #985, `CONTEXT.md`, and accepted ADR-0040; the last is a decision status, not evidence that PT-2 holds.

qinhaihong-red:
## Standards — 1 judgement call; 0 hard violations

At `366a21adc`, `src/gda/project_tree.py:200-220` adds `artifact_to_exclude(project, output_path)`, which interprets export-only output-path syntax. `commands/export.py:1024-1027` is its only caller. This is a possible **Divergent Change** smell: a shared inventory must now change when export destination semantics change. ADR-0040:64-67 and :222-231 place group-specific behavior, including export's artifact, in the command group. Keep output-path-to-`Path` conversion in `commands/export.py`; the shared inventory can retain its `artifact: Path | None` input and identity-based exclusion. The export-destination tests currently in `tests/project_tree/` would then follow that responsibility.

This is a maintainability judgement call, not a demonstrated runtime failure. No other documented-standard breach or baseline smell warrants a finding. The focused 79 tests passed; `git diff --check` is clean. Existing review comments/replies were not used as evidence.

qinhaihong-red:
## Spec — Changes required

Reviewed `d715d50f9...366a21adc` against issue #985's current acceptance criteria, including the test-package amendment. Existing review comments/replies were not used as evidence.

1. **An inaccessible directory reached through two spellings is counted twice.** #985 requires directory links to be followed once by filesystem identity and an unlistable entry to be counted once in `skipped`. In `project_tree.py:257-296`, the `os.walk` error callback records each path before the directory-identity check can run. A project with `locked/` at mode 000 and `alias -> locked` produces `unreadable={'alias','locked'}` and `settle().skipped == 2` for one directory inode. Add a regression case for aliases of an unlistable directory and account for that identity once. This is a current behavioral mismatch, not a test failure in the existing 79 focused tests.

2. **Export destination conversion is outside its specified owner.** #985 AC1 says `commands/export.py` keeps the export-specific artifact and the shared module takes an optional artifact to exclude. `project_tree.py:200-220` instead interprets `output_path` syntax in an additional public helper, called from `commands/export.py:1024-1027`. Resolve the export destination in the command group and pass its `Path | None` to `capture`; keep exclusion by filesystem identity in the shared module.

3. **The specified cost gate has not been measured.** #985 cites #839's +10% bound for a warm **macOS preset** release export on a generated ~10,000-file fixture. The PR's benchmark ran on a macOS host, but its Cost section says **Linux/X11 preset**, `--mode release`. The reported +4.3% and base/head inventory times are useful evidence, but they do not establish the specified macOS-export target. Run or cite that measurement, or amend the acceptance criterion with its rationale. This is an evidence gap, not a demonstrated regression.

Focused tests: 79 passed; `git diff --check` clean. I did not independently rerun the full engine suites or the 10,000-file benchmark.

qinhaihong-red:
## Backtrace review — Reconfirm the completeness contract

The user outcome in #668 is a trustworthy account of files a project-wide import creates, so an agent can review its residue. #839 gave `export run` a `skipped` count and states that repairing an unreadable tree and rerunning is needed for a complete record. #985 correctly converges the two Python walks and fixes the linked-directory omission, but also requires `resource import` to discard that same completeness signal and leave the “every file” / “authoritative inventory” descriptions unchanged.

**Confirmed upstream mismatch:** the core can say “some subtree was unobserved”; the resource result cannot. In a real Godot fixture with a mode-000 directory, `resource import` exits 0 and reports an imported asset and created files, with no completeness field; the shared inventory on that tree reports `skipped=1`. This does not establish that the engine created an omitted file there. It establishes that the public absolute claim cannot be verified from the returned result when enumeration is incomplete. The mismatch predates #985; the PR's link fix does not close it.

**Causal classification:** I found no evidence that repeated patches to *this* completeness boundary have accumulated, so symptom-patch accumulation is **not confirmed**. The shared module serves two current consumers with one stable walk, and adds no registry, variants, compatibility layer, or open-ended platform obligations; NFR-driven platformization is **not found**. The architectural boundary itself is useful and should be retained.

**Decision needed from the issue/product owner:** Should `resource import.created` remain an exhaustive guarantee when `skipped > 0`? If yes, choose the smallest explicit incomplete-result behavior (for example a completeness signal or typed refusal), amend #985's no-public-change constraint, and test the actual result boundary. If no, qualify the public promise and acceptance criteria so `created` means observed files. Keep the shared walk, its `skipped` fact, and the linked-file behavior in either case. A local walker patch alone cannot settle this product-contract choice.

qinhaihong-red:
## Re-verification after the external reviews — READY at head `51ce120f5`

I took the six decisions and the two commits that answer them, and checked each against the running code rather than against the report.

| Finding | Verified at `51ce120f5` |
| --- | --- |
| The `skipped` count is not identity-unique (Spec 1) | Confirmed, and confirmed PRE-EXISTING: a directory at mode 000 also reachable through a link gives `skipped == 2` here AND at the base `d715d50f9`. Handled as prose: rule 4 now states what the settlement counts, names the mechanism, and records that `export run` has counted it this way since #839. The behaviour change belongs to #990, not to a refactor that promises a byte-identical report |
| Export destination syntax lived in the shared core (Architecture 1 / Standards / Spec 2) | Fixed. `artifact_to_exclude` is gone from `gda.project_tree`; `commands/export.py` owns the conversion, and I re-ran every spelling through it: `res://build/game.pck` and `res:///build/g` resolve against the project, a bare `res://`, another scheme and an empty string give none, a relative path resolves against the project and an absolute one passes through. The module keeps `artifact: Path | None` and the identity-plus-name exclusion |
| "It cannot exceed the old walk's cost" (Artifact 1) | Removed; the reason it cannot hold is stated, and linked-tree cost is declared unmeasured |
| The cost gate cited a macOS preset and the benchmark used Linux/X11 (Spec 3) | Closed by measuring: a macOS-preset warm release export of the same fixture is 13.523 s with 0.509 s of inventory, +3.9 %, inside #839's +10 % bound. Both presets are in the body |
| "those three" had no antecedent (Artifact Minor) | Named |
| The claim to have repaired the "authoritative inventory" wording (Artifact 2 / Architecture 2 / Backtrace) | Removed. The body now says what is true — the list is strictly more complete because linked subtrees are enumerated, the completeness signal stays unpublished as AC3 requires, and the unconditional wording is therefore unverifiable when `skipped > 0` — and carries it as an open contract question. That question is now issue #990 |

Gates measured by me on an independent detached checkout at `51ce120f5` (this Mac, macOS arm64, Godot 4.6.3-stable, load 2.0–3.0; the Godot tiers under the machine-wide lock, foreground):

| Check | Result |
| --- | --- |
| `ruff check .` / `ruff format --check .` | `All checks passed!` / `524 files already formatted` |
| `pyright` (bare) | `0 errors, 0 warnings, 0 informations` |
| `pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2940 passed, 3 skipped` |
| e2e `tests/export` + `tests/resource` | `80 passed, 1 skipped` (the pre-existing Linux-only templates case) |
| `gda schema`, root help and all 78 command helps, rebuilt on BOTH sides | **byte-identical** to `d715d50f9` |
| PR CI | ruff, pyright, unit: pass; `MERGEABLE` / `CLEAN` |

Three review rounds now stand behind this branch: two internal at pinned heads, the maintainer's own re-verification, and these five external axes. The external backtrace review found no symptom-patch accumulation and no platformization, and recommended keeping the shared module — which is what shipped.

Not verified: Linux and Windows, as before; the linked-tree walk cost; `tests/project` e2e at this exact head, since the diff since `7ff1dd37a` does not reach it.

Ready for the maintainer's merge decision: squash into `feat/gda-dogfooding-dev`, subject = the PR title, body blank.

Issue #988, opened 2026-09-18
path options: an unresolvable ~user is a literal path, never a traceback


qinhaihong-red:
## What to build

Every `gda` failure is a typed `Error envelope` (ADR-0002, ADR-0004). Four caller-supplied path options break that invariant: when the value starts with a `~user` prefix the host cannot resolve, `Path.expanduser()` raises `RuntimeError` and the process prints a traceback and exits 1 with no envelope at all.

Verified on `feat/gda-dogfooding-dev` at `d715d50f9` with the real CLI. Each of these exits 1 with a traceback:

- `gda export run --output '~unknownuser/x' --preset <name>`
- `gda info --godot '~unknownuser/godot'`
- `gda --user-data-root '~unknownuser/data' info`
- `gda skill --install --dir '~unknownuser/skills'`

`gda scene list --project '~unknownuser/p'` is already correct and returns the typed `project_not_found`, because the project resolver expands through a helper that is total by construction. The model-side `NormalizedPath` is total for the same reason (#699): an unresolvable `~user` passes through unchanged, and whether the path is usable is decided by whoever consumes it. The four options above simply never got that treatment.

The fix is to give them the same one, and nothing more. Route each of the four through the existing total expansion, so the refusal that already waits downstream can speak: a missing binary, an unwritable data root, an absent export destination, a directory the install cannot create. An unresolvable `~user` is a path that names nothing, not a crash.

**Keep the change minimal.** Reuse the total helper the project resolver already uses. Do not add an error code, a path-existence check, a new validator, a shared normalization layer, or a per-command guard. Do not change what `gda.models.normalize_path` or the project resolver already do.

## Acceptance criteria

- [ ] Each of the four options above answers an unresolvable `~user` value with the typed `Error envelope` its own failure path already produces, at that path's existing exit code. No new error code is registered.
- [ ] One CLI regression per option asserts the envelope and the exit code, and fails before the change.
- [ ] `--project` keeps its current `project_not_found` answer, and the model-side normalization keeps its current pass-through. Both are pinned by tests that already exist.
- [ ] Every other `expanduser` call in `src/gda` is either routed through the same total helper or recorded in the pull request body with the reason no caller can reach it with a `~user` value.
- [ ] `gda schema`, every `--help` page, and every existing result stay byte-identical: this changes a failure path only.

## Out of scope

[...]

PR #993, opened 2026-09-19
fix(gda): path options take an unresolvable ~user as a literal name, not a traceback (#988)

qinhaihong-red:
## Summary

Eight sites expanded a tilde on their own, so a `~user` prefix this host cannot
resolve escaped as a Rich traceback at exit 1 with no `Error envelope` at all —
against ADR-0002 / ADR-0004, under which every gda failure is typed.
`gda.project._expand_user` (total since #699, and what `--project` has always
used) becomes public as `expand_user`. An unresolvable `~user` is now a literal
name, as bash and `os.path.expanduser` treat it, and each consumer answers for
itself.

- **Four** are the caller-supplied path OPTIONS the issue names, and they call the
  helper.
- **Three** are project RE-EXPANSIONS the AC4 sweep found reachable — `scene
  validate`, `scene preflight` and `daemon_paths` expand the already-resolved
  project a second time, and `resolve_project_dir` hands them the `~unknownuser…`
  name it kept literal. They call the helper too.
- **One** is gda-mcp's `$GDA_PROJECT` pin. ADR-0011 keeps that package free of any
  `gda` internal symbol, so the same rule is written out in its own resolver — the
  pattern that package already uses for `GDA_PROJECT_ENV` and `PROJECT_MARKER`.

No new error code, no path-existence check, no new validator, no per-command
guard. `gda.models.normalize_path`'s results and the project resolver are
unchanged; the normalizer now takes its decision from `expand_user_or_none`, the
one in-process statement of the rule (the issue's 2026-09-20 amendment).
`gda schema` and every `--help` page are byte-identical to the base.

Closes #988

## What changed

| Change | File |
| --- | --- |
| `_expand_user` → `expand_user` (public; docstring says why it is shared) | `src/gda/project.py` |
| `expand_user_or_none`: the ONE in-process decision (the expansion, or `None` where the host cannot resolve the `~user`); `expand_user` and `normalize_path` are its two callers, each keeping its own answer | `src/gda/project.py`, `src/gda/models.py` |
| `--godot` / `$GDA_GODOT` resolution calls it | `src/gda/binary.py` |
| `--user-data-root` / `$GDA_USER_DATA_ROOT` resolution calls it | `src/gda/runner.py` (`resolve_user_data_root`) |
| `skill --install --dir` calls it | `src/gda/commands/meta.py` |
| The half `export run --output` and `export smoke <artifact>` share calls it | `src/gda/commands/export.py` (`_absolute_filesystem_path`) |
| `project_root` stamping in `scene validate` and `scene preflight` calls it | `src/gda/commands/scene.py` (`:965`, `:1093`) |
| The daemon's project canonicalization calls it | `src/gda/daemon/discovery.py` (`daemon_paths`) |
| The `$GDA_PROJECT` pin gets the same rule, written locally (ADR-0011) | `src/gda/mcp/project_context.py` (`resolve_project_dir`) |
| Eight regressions, each RED without its site's routing | `tests/cli/test_tilde_path_options.py` (new) |
| Two gda-mcp regressions, beside that resolver's own precedence tests | `tests/mcp/test_mcp_project_context.py` |

The three re-expansions are ROUTED, not deleted. Routing keeps every current
input byte-identical; deleting would rest on a claim about what
`resolve_project_dir` returns that no test pins.

`normalize_smoke_artifact_path` guarded this input with its own
`try/except RuntimeError` (#979) while `export run --output` still crashed. The
guard moves onto the shared half and the wrapper's now-redundant copy is deleted,
so the fact has one home and both wrappers get it. `export smoke` behaviour is
unchanged, and its existing `~unknownuser` test stays green untouched.

## Outcome per option

Probed at the head of this branch from an empty scratch directory (values quoted,
because zsh expands a bare `~unknownuser/...` before gda sees it). "Before" is
`27c04d769`, where each of the four printed
`RuntimeError: Could not determine home directory.` and exited 1.

| Option | Before | After (probe line, truncated) |
| --- | --- | --- |
| `gda info --godot '~unknownuser/godot' --json` | traceback, exit 1 | exit 127 — `{"error":{"category":"environment","code":"binary_not_found","message":"Godot binary could not be launched: ~unknownuser/godot",…}}` |
| `gda --user-data-root '~unknownuser/data' info --godot /nonexistent/godot --json` | traceback, exit 1 | exit 127 — `{"error":{…,"code":"binary_not_found","message":"Godot binary could not be launched: /nonexistent/godot",…}}`, after creating `./~unknownuser/data/logs/godot.log` |
| `gda skill --install --dir '~unknownuser/skills' --json` | traceback, exit 1 | exit 0 — `SKILL.md` written to `./~unknownuser/skills/SKILL.md`, `installed_path` reporting it |
| `gda export run --preset nosuch --output '~unknownuser/x' --json` | traceback, exit 1 | exit 4 — `{"error":{"category":"operation","code":"project_not_found","message":"export get requires a Godot project; none was resolved …"}}` (on a host with an engine; without one the same invocation reaches `binary_not_found` first — see the regression note below) |

The three re-expansion sites, probed with a directory literally named
`~unknownuser988/p` that holds a `project.godot` — the only shape that reaches
them, because `resolve_project_dir` refuses anything else:

| Invocation | Before | After |
| --- | --- | --- |
| `gda scene validate res://x.tscn --project '~unknownuser988/p' --json` | traceback, exit 1 | exit 4 — `{"error":{"category":"operation","code":"path_not_found","message":"scene file does not exist: res://x.tscn"}}` |
| `gda scene preflight res://x.tscn --project '~unknownuser988/p' --json` | traceback, exit 1 | exit 4 — the same `path_not_found`, from `scene-preflight` |
| `gda daemon status --project '~unknownuser988/p' --json` | traceback, exit 1 | exit 0 — `{"running":false,"pid":null,"socket_path":"…/.gda/run/<slug>.cli.sock",…}` |

gda-mcp's `$GDA_PROJECT` pin needed no such precondition — it raised for any
`~user` value, whatever was on disk. Measured on the pure resolver:

| Call | Before | After |
| --- | --- | --- |
| `resolve_project_dir({'GDA_PROJECT': '~nosuchuser999/p'}, [], cwd)` | `RuntimeError` | `None` — the strict-pin answer this module already documents, so gda inherits the pin and surfaces its own typed error |
| the same with a literal `~unknownuser988/p` holding a `project.godot` | `RuntimeError` | `Path('~unknownuser988/p')` — a literal name is a real address, which is what tells the guard from a blanket refusal |

The two already-correct references, re-probed at the head and unchanged:

| Reference | After |
| --- | --- |
| `gda export smoke '~unknownuser/game' --json` (already total, #979) | exit 4 — `export_artifact_not_found: <cwd>/~unknownuser/game` |
| `gda scene list --project '~unknownuser/p' --json` | exit 4 — `project_not_found: --project is not a Godot project (no project.godot): ~unknownuser/p` |

A literal `~unknownuser` directory beside the caller is the outcome for the three
write destinations, not a defect: it is what the shell does with the same value.

## The `expanduser` ledger (AC4)

Every `expanduser` in `src/gda` at this head. AC4 names two dispositions: routed
through the same total helper, or recorded here with the reason no caller can
reach the site with a `~user` value. Seven sites take the first and the
unreachable ones the second. ONE site, gda-mcp's `$GDA_PROJECT` pin, takes the
third disposition the issue's 2026-09-20 amendment added to AC4: it is reachable
and it is not routed, because ADR-0011 forbids the import, so it states the same
rule locally with its own regressions.

| Site | Verdict |
| --- | --- |
| `binary.py:45`, `runner.py:244`, `commands/meta.py:239`, `commands/export.py:122` | Routed through `gda.project.expand_user` — the four caller-supplied options. |
| `commands/scene.py:965`, `commands/scene.py:1093`, `daemon/discovery.py:89` | Routed through the same helper — the three project re-expansions. |
| `mcp/project_context.py:60` | Guarded LOCALLY, with the same rule and no `gda` import. ADR-0011 keeps gda-mcp free of any `gda` internal symbol — which is why that module already redefines `GDA_PROJECT_ENV` and `PROJECT_MARKER` for itself — so a three-line `try/except RuntimeError` inside its own resolver is that package's pattern, not the shared normalization layer the issue rules out. Measured after: `resolve_project_dir({'GDA_PROJECT': '~nosuchuser999/p'}, [], cwd)` → `None` (the strict-pin path the module already documents), and with a literal `~unknownuser988/p` holding a `project.godot` → that path. `grep` confirms `src/gda/mcp` still imports nothing from `gda.*`. |
| `project.py:103` | The helper itself. |
| `models.py` (`normalize_path`) | Takes its decision from `expand_user_or_none` and keeps its own answer — the caller's raw string, double slash and all, pinned by `test_normalize_path_is_total_for_an_unexpandable_tilde`; the virtual-path pass-through is untouched, and `gda schema` plus every result stay byte-identical. |
| `daemon/discovery.py:65` | `Path(HOME_RUNTIME_DIR).expanduser()` where `HOME_RUNTIME_DIR` is the module constant `'~/.gda/run'`. A bare `~`, never caller-supplied, so no `~user` form reaches it. Measured: it resolves to `~/.gda/run`. |
| `daemon/server.py:557` | `(self.paths.project / rel).expanduser()`. `paths.project` is the canonical ABSOLUTE root, so the join is absolute and `expanduser` only ever expands a LEADING `~`. Measured: `Path('/abs/project') / '~unknownuser988/x.tscn'` expands to `/abs/project/~unknownuser988/x.tscn`. Structurally unreachable. |
| `commands/script.py:1156` | A comment, not a call. `script run` refuses a leading `~` before any expansion. Measured inside a project: `gda script run '~unknownuser988/s.gd' --json` → exit 4, `invalid_path`. |

The three that are now routed re-expand a path the resolver already expanded
(`resolve_project_dir` returns an `expand_user`-ed `Path`). They were reachable
only when a directory LITERALLY named `~unknownuser…` exists and holds a
`project.godot`, so the resolver hands the literal on and the second
`expanduser()` raised — reproduced that way before the fix, and re-probed after
it in the table above. The gda-mcp pin needed no such precondition: it raised for
any `~user` value, whatever was on disk.

## Schema / help

Measured, not asserted: `gda schema`, the root help page, and all 93
group/command `--help` pages were captured from a detached base checkout at
`27c04d769` and again from this head, and `diff -r` over the two trees reports no
difference. This changes a failure path only.

PR CI (`.github/workflows/ci.yml`, `pull_request`) runs: `ruff check`,
`ruff format --check`, `pyright` (root and `libs/gda-balancing`), the harness
version-identity guard, the unit tier `pytest -m "not e2e" -n 4 --dist loadgroup`,
the Panda pure-Python tier, and `uv build`. All ten regressions here are in the
unit tier, so the contract is gated on every PR run.

The `godot-e2e` job is `schedule` / `workflow_dispatch` only, so **no Godot e2e
runs on this PR**. The e2e suites below were run locally under the machine-wide
lock and repeat in the wave's e2e dispatch.

## Validation

All at the final head `b13793eb1`, on macOS 27.0 (arm64), Godot 4.6.3. Host load
named beside each timing, from `uptime`; the box is shared with other agents.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `525 files already formatted` |
| `uv run --frozen --project libs/gda-balancing pyright --project libs/gda-balancing` | `0 errors, 0 warnings, 0 informations` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2950 passed, 3 skipped, 1 warning in 18.27s` (load 4.5) |
| `pytest -m e2e tests/cli tests/mcp tests/scene tests/export -q -rs` (under the lock) | `140 passed, 1 skipped, 729 deselected in 183.05s` (load 4.58); the skip is the pre-existing Linux-only template gate |
| `pytest -m e2e tests/daemon tests/runtime tests/meta -q -rs` (under the lock) | `77 passed, 574 deselected in 155.03s` (load 4.66) |
| `gda schema` + 95 help pages (root + 94) vs `27c04d769` | byte-identical, 96 files under `diff -r` |

The changed code sits on every headless launch (`--godot`, `--user-data-root`),
on both `scene` verdict channels, on every daemon command's on-disk identity, and
on the MCP server's project resolution, so the e2e tier for `runtime`, `export`,
`scene`, `daemon`, `meta` and `mcp` is part of this slice's evidence and not only
the unit tier.

Each arm was run against a checkout that lacks its own fix, and fails there with
`RuntimeError: Could not determine home directory.`: the five option arms against
a detached `27c04d769` (all five fail); the three re-expansion arms against a
detached `f799afcec`, which already carries the option fixes (exactly those three
fail, the option arms pass); the two gda-mcp arms against a detached `2581726b8`
(exactly those two fail, the other seven in that module pass).

The export-output arm first failed on CI (run 35435970121) while passing locally,
because it let the command run PROJECTLESS: with an engine installed that reaches
the operation's own `project_not_found`, and on a runner without one it reaches
`binary_not_found` first. It now names a `--project` that is not a Godot project,
a refusal `resolve_project_dir` raises CLI-side before the engine is asked for
anything. What it proves is unchanged, and the whole file was re-run with
`GDA_GODOT=/nonexistent/godot` to confirm every arm is engine-independent.

The `--output` amendment ("creates it where it can") is also measured end to end
at this head, outside the test suite: from a scratch cwd, `gda export run
--preset Linux/X11 --mode pack --output '~unknownuser988/out/packed.pck'
--project <a project with the e2e preset> --json` under the e2e lock exits 0,
`output_path` is `<cwd>/~unknownuser988/out/packed.pck`, `created_dirs` names
`<cwd>/~unknownuser988` and `<cwd>/~unknownuser988/out`, and the `.pck` (796
bytes) is on disk there.

## Mutant checks

Each site's unguarded `Path(...).expanduser()` was restored from a saved copy of
the head file (never `git checkout --`), `__pycache__` cleared, the regression
run, then the saved copy put back and the tree confirmed clean.

| Site restored | Regression | Result |
| --- | --- | --- |
| `binary.py` | `test_an_unresolvable_home_on_godot_is_the_binary_not_found_envelope` | RED |
| `runner.py` | `test_an_unresolvable_home_on_the_user_data_root_is_a_directory_under_the_cwd` | RED |
| `commands/meta.py` | `test_an_unresolvable_home_on_the_skill_install_dir_writes_under_the_cwd` | RED |
| `commands/export.py:122` | the two `--output` / normalizer arms **plus** the two pre-existing #979 smoke tests | RED — `4 failed, 91 passed` |
| `commands/scene.py:965` alone | `test_scene_validate_accepts_a_project_whose_literal_name_starts_with_a_tilde` | RED (the preflight arm stays green) |
| `commands/scene.py:1093` alone | `test_scene_preflight_accepts_a_project_whose_literal_name_starts_with_a_tilde` | RED (the validate arm stays green) |
| `daemon/discovery.py:89` | `test_daemon_paths_accept_a_project_whose_literal_name_starts_with_a_tilde` | RED |
| `mcp/project_context.py` (guard deleted) | both `tests/mcp/test_mcp_project_context.py` tilde arms | RED — `2 failed, 7 passed` |

The two `scene.py` rows were mutated ONE LINE at a time, and each run asserts the
other arm stays green, so neither regression is standing in for the other.

The `export.py` row is the one that proves the deleted wrapper guard was MOVED and
not lost: the two pre-existing #979 smoke tests go RED with it.

## Deviations from the issue

One, disclosed here rather than covered by the issue's words. (The gda-mcp
site's local statement of the rule was a second one until the issue's
2026-09-20 amendment added that disposition to AC4.)

- **A touch on an out-of-scope command.** The issue keeps `gda export smoke` out
  of scope because it already answers this input with a typed refusal (#979).
  Its wrapper's own `try/except RuntimeError` is deleted and the guard moves onto
  the shared `_absolute_filesystem_path`, which `export run --output` needed. The
  smoke's behaviour is byte-identical head vs base, and the `://` pass-through
  stays in the `--output` wrapper (#987's lesson).

- The gda-mcp guard is measured on the pure resolver, not against a running MCP
  server: `tests/mcp` e2e passes at this head, but no probe drove a live server
  with an unresolvable `$GDA_PROJECT`.
- The rule has two statements in `src/gda`: `gda.project.expand_user_or_none`,
  the one in-process decision that `expand_user` and `gda.models.normalize_path`
  share, and gda-mcp's local copy in `gda.mcp.project_context`, which ADR-0011
  keeps from importing it. Each is pinned by its own reverse-mutant; the gda-mcp
  copy is kept in step by hand and can drift.
- A DELETED working directory stays out of scope, per the issue: every
  relative-path option then raises `FileNotFoundError` out of `Path.cwd()`, which
  belongs to the shared project resolution rather than to these sites.
- Whether an expanded path exists, is writable, or is of the right kind is still
  the consumer's call, unchanged.
- The three routed re-expansions are ROUTED, not removed: whether
  `resolve_project_dir` can ever hand on a path that still needs expanding is not
  pinned by any test, so the redundant-looking call is kept rather than deleted on
  that unpinned premise.
- Rendered and Windows/Linux behaviour is not verified here: all evidence is
  macOS. Nothing in the change is platform-specific — `expanduser`'s `RuntimeError`
  arm is CPython's, not the platform's.


qinhaihong-red:
## Round-1 review — all six findings adopted

Head after the fixes: **`709403dab5004723e897d9828a7168f59f1bb26a`**
(`a5e18b7bf` for P2-1, `709403dab` for P3-1 to P3-5).

| # | Claim | Decision | Evidence | Change (sha) | Validation |
| **P2-1** | `mcp/project_context.py:50` is reachable and unrouted, so AC4 is unmet, and the "second total helper" half of the stated reason does not follow | **Adopted — fix 1, the local guard.** Your reading is right: ADR-0011 refuses the IMPORT, and that is exactly why this module already redefines `GDA_PROJECT_ENV` and `PROJECT_MARKER` for itself. A `try/except RuntimeError` inside one resolver is that package's own pattern, not the shared normalization layer the issue rules out. The literal is kept, so the strict-pin path the module already documents decides. | Before: `resolve_project_dir({'GDA_PROJECT': '~nosuchuser999/p'}, [], cwd)` raised. After: `None` for a name nothing carries, `Path('~nosuchuser999/p')` for a literal directory that really holds a `project.godot` — the second case is what tells the guard from a blanket refusal of every `~` value. `grep` confirms `src/gda/mcp` still imports nothing from `gda.*`. | `a5e18b7bf` | Two regressions appended to `tests/mcp/test_mcp_project_context.py`, beside that resolver's own precedence tests. Both RED at `2581726b8` with the `RuntimeError` (the other seven in the module pass). Reverse-mutant (guard deleted): `2 failed, 7 passed`. AC4 now reads: seven sites routed, one guarded locally under ADR-0011, the rest unreachable with measured reasons — **Deviations is now "none"**. |
| **P3-1** | The authority's docstring names four of its seven callers and says "four options" while listing five items | **Adopted, your smallest fix.** The enumeration is gone; the paragraph states the SHAPE of the caller set ("the caller-supplied path options, and the project re-expansions that stamp a resolved root onto a result") and says outright that the per-site consequence stays at each call site. That is the enumeration that cannot drift again. | `grep -rn expand_user src` → seven callers; the third commit added three and left the prose at four. | `709403dab` | Prose only; `gda schema` + 94 help pages still byte-identical to `27c04d769`. |
| **P3-2** | "so no file carries that name" is false — a file literally named `~unknownuser/godot` IS launched, and this PR's own tests build the counter-example | **Adopted, your wording.** `resolve_godot_binary` now states the rule rather than an absolute: the value is resolved as the ordinary path it names, exactly as a shell would; where nothing carries that name the caller gets `binary_not_found`, where something does it is launched like any other binary. Same correction in both test-file passages (`:14-15` and the `--godot` arm's comment), which now say it is the ABSENCE that decides. | Your `FAKE-GODOT INVOKED` probe. The PR's own `_literal_project` helper relies on the same possibility. | `709403dab` | Prose only; the `--godot` arm still RED under its reverse-mutant. |
| **P3-3** | "reads the same wherever gda accepts a path" over-states — `normalize_path` differs | **Adopted, your wording.** Narrowed to "so a CLI path option and `--project` answer an unresolvable `~user` the same way." | Your measurement: `normalize_path('~nosuchuser999//x/')` = `'~nosuchuser999//x/'` vs `str(expand_user(Path(...)))` = `'~nosuchuser999/x'` — the ledger's own row. | `709403dab` | Prose only. (The second counter-example you cited, `gda.mcp`, is closed by P2-1 — but the `normalize_path` one stands, so the sentence stays narrow.) |
| **P3-4** | "total" predicated of a VALUE, and the wrapper "states nothing about it" inside the sentence that states it | **Adopted, your wording.** "The shared half above is total for an unresolvable `~user`. The guard this wrapper carried alone (#979) lives there now, so `--output` gets the same one rule (#988)." | Totality is the function's property; the sentence contradicted itself. | `709403dab` | Prose only; `export smoke` behaviour unchanged, and both pre-existing #979 smoke tests still go RED under the `export.py` mutant. |
| **P3-5** | Stale `_expand_user` left by the rename | **Adopted.** Renamed in place. | Only surviving reference to the pre-rename symbol. | `709403dab` | `grep -rn _expand_user src tests libs` → no match. |

## Gates at `709403dab`

macOS 27.0 (arm64), Godot 4.6.3. Host shared with other agents; load from `uptime` beside each timing.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `525 files already formatted` |
| `uv run --frozen pyright` (bare) | `0 errors, 0 warnings, 0 informations` |
| `uv run --frozen --project libs/gda-balancing pyright --project libs/gda-balancing` | `0 errors, 0 warnings, 0 informations` (you listed this as not run) |
| `pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2950 passed, 3 skipped, 1 warning in 24.01s` (load 4.49) — 2948 + the two gda-mcp arms |
| `pytest -m e2e tests/runtime tests/export tests/scene -q -rs` (under the lock) | `161 passed, 1 skipped, 355 deselected in 253.10s` (load 5.24); the skip is the pre-existing Linux-only export-template gate |
| `pytest -m e2e tests/daemon tests/meta tests/mcp -q -rs` (under the lock) | `54 passed, 516 deselected in 80.19s` (load 3.26) — `tests/mcp` e2e added for P2-1 |
| `gda schema` + 94 help pages (root + 93) vs a detached `27c04d769` | `diff -r` reports no difference |

All eight reverse-mutants re-run at this head, each restored from a saved copy and
the tree confirmed clean after: `binary.py:45`, `runner.py:244`,
`commands/meta.py:239`, `commands/export.py:122` (`4 failed, 91 passed` — the two
new arms plus both pre-existing #979 smoke tests), `commands/scene.py:965` ALONE,
`commands/scene.py:1093` ALONE (each fails only its own arm), `daemon/discovery.py:89`,
and `mcp/project_context.py` with the guard deleted (`2 failed, 7 passed`).

## Still not verified

- Platforms other than macOS.
- The gda-mcp guard is measured on the pure resolver and through `tests/mcp` e2e; no
  probe drove a live MCP server with an unresolvable `$GDA_PROJECT`, so whether the
  old `RuntimeError` reached the serve loop or the SDK is still unestablished — and
  now unreachable.
- The rest of the e2e tier (`tests/live`, `tests/script`, `tests/node`,
  `tests/resource`, `tests/asset_file`, `tests/project`, `tests/harness`,
  `tests/value_projection`). PR CI runs no Godot e2e.
- The rule now has two homes — the helper and four lines in `gda.mcp.project_context`
  — because ADR-0011 forbids the import. Each is pinned by its own reverse-mutant,
  but they can drift.

qinhaihong-red:
## Round-2 review — findings handled at `ac5d8aa8c`

All five findings were reproduced before they were answered. Four are adopted in
one prose-only commit; the fifth is not adopted, as the review itself proposed.
Head after the fix: **`ac5d8aa8c`** (was `709403dab`). The PR body's residual-risk
bullet now says three statements of the rule, not two, and names the third.

| P3-1 | `expand_user`'s docstring says the rule "has a single home" and `runner.py:30` opens "The one total `~` expansion", while three statements of the rule live in `src/gda` and the PR body itself said two | **Fully adopt** | Confirmed: `project.py` (`try/except RuntimeError`), `models.py`'s `normalize_path` (its own `try/except`, answering with the raw string, #699), `mcp/project_context.py` (the local guard). The body's "two homes" was wrong as well | `ac5d8aa8c` — both uniqueness claims removed; the authority now names the other two and why each keeps its own (`normalize_path` also carries the virtual-path pass-through and answers with the caller's raw string; gda-mcp cannot import it under ADR-0011). PR body residual risk: "three statements in `src/gda` … kept in step by hand" | Prose only; gates below |
| P3-2 | The caller-set shape "stamp a resolved root onto a result" does not describe `daemon_paths` | **Fully adopt** | Confirmed: `daemon status --json` carries no project-root field; `daemon_paths` canonicalizes the daemon's on-disk identity | `ac5d8aa8c` — "…that stamp a resolved root onto a verdict or onto the daemon's on-disk identity" | Prose only |
| P3-3 | "escaped both wrappers" is disproved by the base and by the paragraph's own last line | **Fully adopt** | Confirmed: #979 introduced `normalize_smoke_artifact_path` with its guard already in place; only `export run --output` crashed | `ac5d8aa8c` — "which escaped `export run --output` as a traceback at exit 1 with no envelope at all" | Prose only |
| P3-4 | The test module counts seven sites where the PR fixes eight | **Fully adopt** | Confirmed: the eighth is gda-mcp's `$GDA_PROJECT` pin, guarded locally rather than through the helper | `ac5d8aa8c` — "eight sites … The seven arms HERE share the project resolver's total expansion … The eighth, gda-mcp's `$GDA_PROJECT` pin, states the same rule locally under ADR-0011 and is pinned in `tests/mcp/test_mcp_project_context.py`" | Prose only |
| P3-5 | `709403dab`'s subject lacks `(#988)` | **Not adopt** | Agreed, and as the review proposes: the squash subject is the PR title, which carries `(#988)`; the branch is not rewritten for a subject | none | — |

Why the uniqueness claim came back once (round-1 P3-3, then round-2 P3-1): the
round-1 fix narrowed one clause and left the claim's premise standing. The premise
is what is wrong — three statements are structural, because `normalize_path`
answers a different question and gda-mcp cannot import — so this round removes
the claim everywhere and makes the authority list the other two. There is no
third wording to try.

### Gates at `ac5d8aa8c`

Host macOS (Darwin 27.0.0), Godot 4.6.3-stable, shared machine, `uptime` load
3.84–5.00 across the runs. Every line measured after the last commit.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `525 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2950 passed, 3 skipped, 1 warning in 19.89s` |
| `pytest -m e2e tests/cli tests/mcp tests/scene tests/export -q -rs` (under the e2e lock) | `140 passed, 1 skipped, 729 deselected in 180.26s` |
| `pytest -m e2e tests/daemon -q -rs` (under the lock) | `40 passed, 221 deselected in 62.94s` |
| `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| `gda schema` + every `--help` page vs base `27c04d769` | 96 files, `diff -r` empty |

qinhaihong-red:
## READY — orchestrator re-verification at `26b89db40`

Three fresh axes were run on a detached checkout at the round-2 head
`ac5d8aa8c`: Standards (documented rules plus a smell baseline), Spec (every
amendment outcome reproduced head vs base with the real CLI, a seven-site
reverse-mutant, and the surface rebuilt from both checkouts), and
Domain-modular architecture (dependency graph, DRY, deletion test,
extensibility). Their findings and what happened to each:

| Axis | Finding | Disposition |
| --- | --- | --- |
| Spec | The body said "Deviations from the issue: None", but the gda-mcp site takes a disposition AC4 does not list — reachable, not routed, the rule stated locally under ADR-0011 | Body: the AC4 ledger intro names the third disposition, and "Deviations" now records two (this one, and the `export smoke` wrapper touch, byte-identical) |
| Spec | The `--output` amendment ("creates it where it can") was pinned only by a normalizer call, never by an export | Measured end to end at this head: a pack export into `'~unknownuser988/out/packed.pck'` from a scratch cwd exits 0, `created_dirs` names both literal directories, the `.pck` is on disk; in the body |
| Spec | The three re-expansion routings and the gda-mcp guard | Judged INSIDE AC4, not creep; all load-bearing (reverse-mutant) |
| Standards | One twenty-word statement of the rule repeated verbatim in three call-site docstrings | `26b89db40` — each points at `expand_user`, which owns the rule, and states only its own outcome |
| Standards | `export.py`: "so both wrappers state it once" — only the smoke wrapper points at the rule | `26b89db40` — "neither wrapper carries a guard of its own" |
| Standards | Two sentences of 47–62 words; one garden-path clause | `26b89db40` — split; "A `~` that gda cannot expand names no user" |
| Architecture | `normalize_path` names no sibling, while the authority names both copies | `26b89db40` — it names `expand_user` and says why it stays its own |
| Architecture | A ninth `expanduser` site has no gate | Not adopted, as the axis itself recommends: a gate is the mechanism the issue rules out; recorded as a validation need |
| Standards | `UNKNOWN_USER` could be `UNRESOLVABLE_HOME`; `gda.project` as the home of a generic `~` primitive | Not adopted: the constant names the user half; the home is the issue amendment's, and a `paths` module for one function is the abstraction the standing bound forbids |
| Standards | Pre-existing: `daemon/discovery.py:65` expands a bare `~`, which raises only when no home can be determined at all | Out of #988's `~user` scope; noted for the maintainer, not filed |

Nothing in this round changed behaviour: `gda schema` and every `--help` page
stay byte-identical to the base (96 files under `diff -r`). The validation
table in the body is re-measured at `26b89db40`; PR CI is 7 pass / 4 skip at
that head; `Closes #988` is in the body; author and committer are
`haihong.qin` on all seven commits; no attribution anywhere.

qinhaihong-red:
## Independent adversarial review — Changes required

Reviewed exact range `27c04d76969b06733e80c4d27895aa7fd982d66f...26b89db400031a98eae10156afe42453a8131ac1`. I did not read or use existing PR reviews or PR comments.

### Artifact review

**Changes required.**

1. **[P2] Issue #988 does not authorize the architecture-required third AC4 disposition.** AC4 requires every other `expanduser` call either to use the shared helper or to be documented as unreachable from `~user`. [`gda.mcp.project_context`](https://github.com/aigengame/godot-agent/blob/26b89db400031a98eae10156afe42453a8131ac1/src/gda/mcp/project_context.py#L51-L63) is reachable and deliberately implements the rule locally because ADR-0011 prohibits importing a `gda` internal symbol. The PR discloses that deviation, but disclosure does not amend the issue. The smallest coherent fix is to amend [#988](https://github.com/aigengame/godot-agent/issues/988) so AC4 explicitly permits an equivalent context-local implementation where ADR-0011 requires it, with its own regressions.
2. **[P3] The issue title and opening still describe all four options as a typed refusal.** The dated amendment instead makes three write destinations valid literal relative paths that can succeed. Update the title/opening so the current goal is “literal path, never traceback”; otherwise the artifact's summary contradicts its controlling amendment.

**Terminology and prose assessment.** Sources: `CONTEXT.md`, ADR-0002, ADR-0004, ADR-0006, ADR-0011, ADR-0040, the complete current issue, and the complete PR body. Reference: ASD-STE100. I inspected all issue/PR requirements, decisions, procedures, tables, and changed technical prose for stable terms, causal/scope wording, and alternate names. The two inconsistencies above are substantive; no other terminology or prose defect was found. I did not verify Windows/Linux shell behavior.

The observed direction is otherwise sound: in-process callers depend on the existing `gda.project` path authority, and the MCP adapter keeps its accepted subprocess/no-internal-import boundary. No cycle or group-to-group dependency was added.

- **[P3] The in-process fallback rule still has two editable authorities.** [`gda.project.expand_user`](https://github.com/aigengame/godot-agent/blob/26b89db400031a98eae10156afe42453a8131ac1/src/gda/project.py#L103-L133) owns “unresolvable `~user` stays literal”, but [`gda.models.normalize_path`](https://github.com/aigengame/godot-agent/blob/26b89db400031a98eae10156afe42453a8131ac1/src/gda/models.py#L808-L832) retains the same `expanduser()`/`RuntimeError` decision, and the new owner says the copies must be kept in step by hand. That leaves the load-bearing DRY claim contradicted at this head. `models.py` already depends on `gda.project`; delegate the expansion/fallback decision while retaining `normalize_path`'s virtual-path pass-through and exact raw-string result. Preserve the double-slash/raw-string regression when doing so. The MCP-local copy is retained because it belongs to the separate ADR-0011 boundary.

Claim evidence state for “one in-process authority”: `open` (current code contains two handlers). Disconfirming evidence would be a demonstrated inability to delegate without changing `normalize_path`'s byte-level result; the existing raw-string tests are the required validation.

### Standards

**1 judgement-call finding.** The duplicate in-process `expanduser()`/`RuntimeError` fallback above is possible **Duplicated Code** and conflicts with ADR-0040 §4's single-authority direction; `tests/cli/test_params_json.py` still calls `normalize_path` the “single home.” No other documented-standard breach or baseline smell was found.

### Spec

**Changes required — 1 finding.** AC4 is unsatisfied at [`src/gda/mcp/project_context.py:60`](https://github.com/aigengame/godot-agent/blob/26b89db400031a98eae10156afe42453a8131ac1/src/gda/mcp/project_context.py#L60): the caller-reachable site is neither routed through `gda.project.expand_user` nor unreachable. Amend #988 to authorize the ADR-0011-required local rule, or redesign the authority without violating ADR-0011. No other Spec finding: the four amended outcomes, three re-expansions, unchanged `normalize_path` behavior, remaining ledger, and `export smoke` behavior all match the issue.

### Validation

- `git diff --check`, Ruff check/format, Pyright, and the harness version guard: pass.
- Full non-e2e tier: `2948 passed, 5 skipped` locally.
- Focused tilde/MCP/containment coverage: `75 passed`; the independent Spec axis also ran 17 tilde/MCP and 121 export-smoke/containment checks successfully.
- `gda schema` and all 94 root/group/command help surfaces are byte-identical to the base (independent SHA-256 comparison).
- Hosted CI is green, but its Godot e2e job is skipped; no new real-engine run was needed to establish the findings above.

qinhaihong-red:
## Independent adversarial review — findings handled at `b13793eb1`

Every finding was reproduced before it was answered: two are adopted in the
issue, one in a refactor commit, one is not adopted with the reason stated.
Head after the fixes: **`b13793eb1`** (was `26b89db40`). Issue #988 and the PR
body are updated.

| Artifact P2 / Spec P2 | AC4 offers two dispositions; `gda.mcp.project_context` takes a third (reachable, not routed, ADR-0011), and disclosure does not amend the issue | **Fully adopt** | Confirmed: ADR-0011 forbids the import and the site is reachable at the base | Issue #988, `## Amendments (2026-09-20)`: AC4 restated with the third disposition — stated locally with its own regressions where ADR-0011 forbids the import. PR body: the ledger intro and "Deviations" cite the amendment; one deviation remains, the `export smoke` wrapper touch | The two regressions in `tests/mcp/test_mcp_project_context.py`, unchanged |
| Artifact P3 | The title and opening still say "typed refusal" while the 2026-09-19 amendment makes three options succeed on a literal path | **Fully adopt** | Confirmed | Title → "path options: an unresolvable ~user is a literal path, never a traceback"; the amendment records that the opening's framing is superseded | — |
| Architecture P3 / Standards | `expand_user` and `normalize_path` are two editable copies of one in-process decision; delegate while keeping `normalize_path`'s byte-level result | **Fully adopt** | Confirmed: both held the same `expanduser()`/`RuntimeError` decision. Delegating to `expand_user` itself would re-spell the raw string (`'~u//x/'` → `'~u/x'`), so what is shared is the decision, not the answer | `b13793eb1` — `gda.project.expand_user_or_none` is the one in-process decision (the expansion, or `None` where the host cannot resolve the `~user`); `expand_user` and `normalize_path` are its two callers and keep their own answers. The issue amendment records that `normalize_path` may share the decision while its results stay byte-identical. The normalizer's regression now also pins the double-slash raw string | Reverse-mutant (the decision re-raises): 9 RED across both callers — the normalizer's regression and every routed option arm; `gda schema` + 95 help pages byte-identical to the base (96 files) |
| Standards | `tests/cli/test_params_json.py` still calls `normalize_path` "the single home" | **Not adopt** | That sentence is ADR-0015's claim about normalization being model-side, not a claim about the tilde rule; it stays true | none | — |

The in-process rule now has ONE decision site. gda-mcp's local copy is the only
other statement, and the authority names it.

### Gates at `b13793eb1`

Host macOS (Darwin 27.0.0), Godot 4.6.3-stable, shared machine, `uptime` load
4.5–4.8 across the runs. Every line measured after the last commit.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `525 files already formatted` |
| `uv run --frozen --project libs/gda-balancing pyright --project libs/gda-balancing` | `0 errors, 0 warnings, 0 informations` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2950 passed, 3 skipped, 1 warning in 18.27s` |
| `pytest -m e2e tests/cli tests/mcp tests/scene tests/export -q -rs` (under the e2e lock) | `140 passed, 1 skipped, 729 deselected in 183.05s` |
| `pytest -m e2e tests/daemon tests/runtime tests/meta -q -rs` (under the lock) | `77 passed, 574 deselected in 155.03s` |
| `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| `gda schema` + every `--help` page vs base `27c04d769` | 96 files, `diff -r` empty |

qinhaihong-red:
## Independent adversarial re-review — Pass

Re-reviewed exact head `b13793eb126f8e6c861839936da8a2948739d75d` against base `27c04d76969b06733e80c4d27895aa7fd982d66f` and the previously reviewed head `26b89db400031a98eae10156afe42453a8131ac1`. The response to the prior review was treated as a set of claims; the conclusions below come from the current issue, diff, repository authority, and independent checks.

### Artifact review

**Pass.** Issue #988 now explicitly permits the ADR-0011-required context-local gda-mcp implementation, changes the title to the literal-path outcome, and explicitly supersedes the opening's old all-refusal framing. The PR body and ledger now agree with that authority. No remaining actionable terminology, scope, or internal-consistency finding was found.

**Pass.** `gda.project.expand_user_or_none` is now the sole in-process decision for `Path.expanduser()`'s unresolved-user case. `expand_user` and `gda.models.normalize_path` delegate to it while preserving their distinct result representations; the only other statement remains inside the ADR-0011-isolated gda-mcp context. No new cycle, ownership split, or inappropriate cross-context dependency was introduced.

### Standards / Spec

**Pass.** The prior duplicated-authority concern is closed, AC4 is satisfied as amended, and the byte-level `normalize_path` behavior remains pinned. No new documented-standard breach, baseline smell, or spec deviation was found in `26b89db4..b13793eb` or the full base-to-head change.

### Validation

- Focused tilde, MCP, params-JSON, export-smoke, and containment checks: `243 passed` in the broadest independent run; the changed-area slice also passed `180` tests.
- Ruff and Pyright: pass; `git diff --check`: pass.
- `gda schema` is byte-identical before and after the fix commit; the complete root/group/command help capture remains byte-identical to the base.
- Required hosted CI checks are green. The hosted Godot e2e job is skipped by workflow design; no additional Godot-backed e2e was needed for this pure-Python delegation change.

The previous review findings are closed at this exact head. I found no new actionable issue.

Issue #990, opened 2026-09-18
resource import: settle what an incomplete project tree inventory promises, and count one inode once


qinhaihong-red:
## What to build

Two facts about an incomplete enumeration, settled together because they are the same fact seen from two sides: what the count MEANS, and who is allowed to see it.

Since #985 both `gda export run` and `gda resource import` read one `Project tree inventory`, which settles two captures and can report `skipped` — the entries neither capture could account for, an unreadable directory among them. `export run` publishes that count. `resource import` does not, and its published `created` field says "Every file the pass created, classified; empty on a dry run." The command catalog calls the same list "the authoritative inventory". When a subtree is unreadable, the pass still exits 0 and the result carries no signal that the enumeration was partial, so the unconditional claim cannot be verified from the result. This predates #985: the command's own walk had the same silence, and #985's acceptance criteria forbade a new field, so the slice left it alone. What changed is that the fact now exists and is discarded rather than never computed.

The current inventory rule and walk count an unlistable or unreadable entry once per project-relative spelling, not once per filesystem inode. A directory at mode 000 that is also reachable through a directory link inside the project is counted twice for one filesystem inode. Reproduced on the current head and identically on the base commit `d715d50f9`, so this is long-standing behaviour, not a regression; #985 corrected the rule's wording to describe it rather than change a published count inside a refactor.

**The product decision comes first, and it is the maintainer's.** Does `resource import`'s `created` promise an exhaustive list when the walk could not read part of the tree?

- **If yes**, the result needs the smallest explicit signal that the enumeration was partial, and `#985`'s "no new field" constraint is superseded for this one field.
- **If no**, the published description and the catalog sentence are qualified so that `created` means the files the pass created that gda observed, and the exhaustive reading is retired.

Either way the shared walk, its `skipped` fact and the linked-directory behaviour stay as they are. Record the decision before implementing; the rest of this issue follows from it.

Whichever branch is chosen, also make the count identity-unique: one filesystem inode that cannot be listed is one entry in `skipped`, whatever spelling reached it. This changes `export run`'s published count in that corner, which is why it did not belong inside a behaviour-preserving refactor; declare it in the implementing pull request's schema and behaviour delta.

**Keep the change minimal.** No new error code, no completeness policy layer, no per-command configuration, no change to what the walk enumerates. If a signal is added, it is one field on one result, described in one sentence.

## Acceptance criteria

- [ ] The decision above is recorded on this issue with its reason before any code changes.
[...]

PR #994, opened 2026-09-19
feat(gda): resource import reports the entries its inventory could not see, and one unlistable inode counts once (#990)

qinhaihong-red:
## Summary

Two facts about an incomplete enumeration, settled together.

`gda resource import` now publishes `skipped` — one integer beside `created`,
with the meaning `gda export run`'s field already has. Until now the shared
`Project tree inventory` computed that count on every real run and this command
threw it away, so `created`'s "Every file the pass created" could not be checked
from the result: an unreadable subtree left the pass at exit 0 with nothing said.

And one unreadable inode is now counted once. `os.walk` reports a listing error
INSTEAD of yielding the directory, and a per-file failure never reaches rule 1 at
all, so two names for one unreadable inode were counted twice. Each failing
spelling is identified the moment its failure is observed now, at every
observation and in whichever capture observes it, and the settlement counts on
those identities.

Closes #990

## The decision it implements

The issue's dated **Decision (2026-09-19)** section is the contract. The
maintainer decided YES: `created` keeps its exhaustive promise, and the result
says when the inventory could not keep it. One new field, one sentence, no
completeness policy layer, no new error code, no change to what the walk
enumerates. #985's "no new field" constraint is superseded for this one field.

## What changed

**Part 1 — `resource import` publishes `skipped`.**

- `src/gda/commands/resource.py`
  - `ResourceImportResult` gains `skipped: int = 0`, placed directly after
    `created` because it qualifies that list, described in two sentences that
    mirror its twin on `export run`.
  - `created`'s description is the one authority for what the pair means: "the
    list is complete when `skipped` is 0".
  - Every other published home of that promise defers to it in one clause — the
    result model's own docstring (twice) and `pass_will_also_import`'s
    description, both of which a machine reads from `gda schema`, plus the
    command docstring that is both the help text and the schema's command
    `description`.
  - `_mode_fields` treats the field as it treats `created`: a dry run walks
    nothing, so a dry-run payload with `skipped != 0` is refused; a real run may
    carry any count.
  - The recipe keeps the settlement it already took (`inventory.settle()`) and
    reads both halves from it instead of the `created` list alone.
  - `render_resource_import` names the count on the created line when it is not
    `0` (`  created: 2 cache-owned, 1 source-adjacent, 1 unreadable`), and prints
    that line for the count alone, so a record that could not read part of the
    tree never prints as a complete one — `export run`'s rule for the same fact.

**Part 2 — one unreadable inode counts once.**

- `src/gda/project_tree.py`
  - New module-private `_SkippedEntries`: adds arrive as project-relative
    spellings and each is identified WHEN its failure is observed, by a `stat`
    of the failing path at that moment — the `(st_dev, st_ino)` pair, or `None`
    for a path `stat` cannot answer for (a dangling link, an entry that
    vanished), which is then counted under its spelling. The two questions sit
    on two records: `covers(rel)` answers on the spelling — whether a capture
    already failed to read this path, the question the settlement asks before
    it calls a file created — and `count` answers on the set of identities
    every observation reached, so a spelling is identified at EVERY
    observation and not only at its first: the settlement's walk can reach a
    second inode through a spelling the first capture recorded (a link
    retargeted between the captures), and that inode was observed failing too
    (external re-review counterexample: `alias` retargeted from `locked/` to a
    second mode-000 directory, was 1).
  - `capture()` records its failures through the helper and keeps the
    observations: `unreadable` is now `dict[str, tuple[int, int] | None]`,
    spelling → identity at observation, and stays internal. `settle()` continues
    from those observations and passes the helper's `add` as its own walk's
    sink, so a spelling that vanishes between the captures does not split one
    observed inode into two entries (external review counterexample: `locked/`
    + `alias -> locked/`, `alias` removed after the capture, was 2).
  - What the settlement observes is decided by the first capture's record,
    what it may call created or modified by coverage. A path the record does
    not hold is read — one that was not there with a `stat`, one the first
    capture could not account for (a spelling it could not read, or anything
    beneath a directory it could not list, under every spelling that reaches
    it) under the capture's own criterion — and a failure observed now is
    counted on the identity the path reaches now; a path the record holds is
    re-read only as far as the rewrite-candidate rule asks. `_hashed(rel,
    detect_rewrites)` is the one rule for every read that asks a file for its
    bytes — only when rewrites are asked for, never under the cache root — at
    the first capture, the rewrite candidate and the second look. Before this, a covered file was looked at
    under a criterion of its own (`digest=self.detect_rewrites`, without the
    cache-root clause, pinned by no test) and a file beneath a directory that
    opened up was passed over while a locked subdirectory beside it was counted
    through the walk's error sink. Five regressions pin the rule: the two POSIX
    FIFO races (inode A → inode B, unidentified spelling → inode), a FIFO
    beneath a directory that opens up, a covered file outside the cache root
    replaced by a second mode-000 inode (counts both, with rewrites on), and a
    covered cache FIFO replaced by a mode-000 regular file (one entry).

Cost: one `stat` per entry the walk could not read, and — where rewrites are
asked for and the path is outside the cache root — one read of a covered path
the settlement can read now. Both sets are normally empty and bounded by what a
run failed to read rather than by tree size, and the walk already stats every
entry it does reach — so no measurable change is expected and none was measured.

## Behaviour delta

`gda resource import` gains `skipped`, which it never published.

`gda export run`'s published `skipped` changes for a shape reached by two names,
and only there. Measured through the inventory, base `27c04d769` vs this head,
on the same five fixtures; `created` and `modified` are empty and identical on
every row:

| Fixture | base `skipped` | head `skipped` |
| --- | --- | --- |
| mode-000 directory + a directory link to it | 2 | 1 |
| FIFO + a symlink to the same FIFO | 2 | 1 |
| mode-000 regular file + a symlink to it | 2 | 1 |
| two distinct mode-000 directories | 2 | 2 |
| two dangling links | 2 | 2 |

The rule is per filesystem identity, so the first three collapse and the last two
do not: two inodes stay two, and a dangling link has no inode to ask for and is
still counted by its spelling. No `created` or `modified` list changes, on these
fixtures or on the real-engine runs below.

## Schema / help

`gda schema` against a detached checkout of the base `27c04d769`, normalized and
diffed. **Five hunks, nothing else**, all on `resource import`:

1. the command `description` (which is also the `--help` text)
2. the result model's own `description` (its docstring)
3. `created`'s description — the authority sentence
4. `pass_will_also_import`'s description — the one-clause deferral
5. the new `skipped` property

`ProjectTreeMutations.skipped`'s description is unchanged, so `export run`'s
schema is byte-identical.

```diff
--- gda schema @ 27c04d769
+++ gda schema @ HEAD
@@ -29305,5 +29305,5 @@
       ],
       "constraints": null,
-      "description": "Ensure assets are imported into the project cache (clean-worktree loading).\n\nA clean worktree has the sources and their committed .import sidecars but\nnot the gitignored .godot/ cache, so a one-shot run's preload() of e.g. a\nPNG fails with \"no recognized resource loader\" (GDA-DF-010). This command\nreads each requested asset's evidence state as the engine's own reimport\ntest reads its artifacts (`cached` / `missing` / `stale` / `invalid`) and,\nonly when a request is missing or stale, runs the engine's import pass —\nwhich is PROJECT-WIDE, the engine's one scriptable import primitive; an\ninvalid request settles `failed` without a pass (the engine skips failed\nimports and parse errors; gda conservatively skips unsupported receipt\nsyntax — delete the sidecar to retry). It then reports\nevery file the pass created, classified against the cache root\n(cache-owned under .godot/ vs source-adjacent, e.g. .import and .uid\nsidecars). An `invalid` or `failed` asset says why: `reason` names the\ncheck that decided it (an invalid one survives into the settled `failed`),\n`detail` the offending line or path when there is one, and `engine_output`\nthe pass's own stderr lines naming that asset when this request ran a pass\n(at most 20, with `engine_output_truncated` when more matched; empty when\nno pass ran). `--dry-run` reports the states and the decidable predictions\n(including `pass_will_also_import`) and writes nothing. An asset whose CASE does\nnot match the stored file is refused with `path_case_mismatch` naming the stored\nres:// spelling, because such a path opens on a case-insensitive filesystem and\nfails on a case-sensitive one. Plain `gda script\nrun` never triggers an import pass. The pass\nexecutes engine importer code over project content (the Trusted project\nassumption, ADR-0009).",
+      "description": "Ensure assets are imported into the project cache (clean-worktree loading).\n\nA clean worktree has the sources and their committed .import sidecars but\nnot the gitignored .godot/ cache, so a one-shot run's preload() of e.g. a\nPNG fails with \"no recognized resource loader\" (GDA-DF-010). This command\nreads each requested asset's evidence state as the engine's own reimport\ntest reads its artifacts (`cached` / `missing` / `stale` / `invalid`) and,\nonly when a request is missing or stale, runs the engine's import pass —\nwhich is PROJECT-WIDE, the engine's one scriptable import primitive; an\ninvalid request settles `failed` without a pass (the engine skips failed\nimports and parse errors; gda conservatively skips unsupported receipt\nsyntax — delete the sidecar to retry). It then reports\nevery file the pass created, classified against the cache root\n(cache-owned under .godot/ vs source-adjacent, e.g. .import and .uid\nsidecars) — a list `skipped` qualifies: it counts what the inventory could not\nread, so `created` is complete when `skipped` is 0. An `invalid` or\n`failed` asset says why: `reason` names the\ncheck that decided it (an invalid one survives into the settled `failed`),\n`detail` the offending line or path when there is one, and `engine_output`\nthe pass's own stderr lines naming that asset when this request ran a pass\n(at most 20, with `engine_output_truncated` when more matched; empty when\nno pass ran). `--dry-run` reports the states and the decidable predictions\n(including `pass_will_also_import`) and writes nothing. An asset whose CASE does\nnot match the stored file is refused with `path_case_mismatch` naming the stored\nres:// spelling, because such a path opens on a case-insensitive filesystem and\nfails on a case-sensitive one. Plain `gda script\nrun` never triggers an import pass. The pass\nexecutes engine importer code over project content (the Trusted project\nassumption, ADR-0009).",
       "error": {
         "$defs": {
@@ -29928,5 +29928,5 @@
           }
         },
-        "description": "The result of ``gda resource import`` (#668).\n\nThe report IS the scoping: the engine's import primitive is project-wide,\nso gda runs it only when a requested asset's evidence state is ``missing``\nor ``stale`` (``engine_pass``; an ``invalid`` request settles ``failed``\nwithout a pass — the engine skips failed imports and parse errors, while\ngda conservatively skips unsupported receipt syntax), and accounts\nfor everything it touched — ``created`` lists every new file, classified\nagainst ``cache_root``. On a dry run nothing runs and nothing is written:\n``assets`` carry the ``cached`` / ``missing`` / ``stale`` / ``invalid``\nstates, ``engine_pass`` says whether a real run WOULD run the pass,\n``predicted_source_adjacent`` lists the requested assets' sidecars-to-be,\nand ``pass_will_also_import`` the other stale assets the pass will\nre-import (the remaining inventory — engine hash-named cache files under\n``cache_root``, sidecars for no-sidecar assets, generated ``.uid`` files —\nis the engine's to decide, and the real run's ``created`` is the\nauthoritative list). The mode's field set is validated, not merely\ndescribed.",
+        "description": "The result of ``gda resource import`` (#668).\n\nThe report IS the scoping: the engine's import primitive is project-wide,\nso gda runs it only when a requested asset's evidence state is ``missing``\nor ``stale`` (``engine_pass``; an ``invalid`` request settles ``failed``\nwithout a pass — the engine skips failed imports and parse errors, while\ngda conservatively skips unsupported receipt syntax), and accounts\nfor everything it touched — ``created`` lists every new file, classified\nagainst ``cache_root``, and ``skipped`` says how much of the tree the\ninventory could not see. On a dry run nothing runs and nothing is written:\n``assets`` carry the ``cached`` / ``missing`` / ``stale`` / ``invalid``\nstates, ``engine_pass`` says whether a real run WOULD run the pass,\n``predicted_source_adjacent`` lists the requested assets' sidecars-to-be,\nand ``pass_will_also_import`` the other stale assets the pass will\nre-import (the remaining inventory — engine hash-named cache files under\n``cache_root``, sidecars for no-sidecar assets, generated ``.uid`` files —\nis the engine's to decide, and the real run's ``created`` is the\nauthoritative list, complete when its ``skipped`` is 0). The mode's field\nset is validated, not merely described.",
         "properties": {
           "assets": {
@@ -29944,5 +29944,5 @@
           },
           "created": {
-            "description": "Every file the pass created, classified; empty on a dry run.",
+            "description": "Every file the pass created, classified; the list is complete when `skipped` is 0; empty on a dry run.",
             "items": {
               "$ref": "#/$defs/ImportCreatedFile"
@@ -29962,5 +29962,5 @@
           },
           "pass_will_also_import": {
-            "description": "Dry run only: OTHER assets whose committed sidecar fails an engine check (stale) — the project-wide pass WILL re-import these. Invalid assets are excluded (the engine skips them), and assets with no sidecar or generated .uid sidecars cannot be predicted; the real run's `created` list is the authoritative inventory.",
+            "description": "Dry run only: OTHER assets whose committed sidecar fails an engine check (stale) — the project-wide pass WILL re-import these. Invalid assets are excluded (the engine skips them), and assets with no sidecar or generated .uid sidecars cannot be predicted; the real run's `created` list is the authoritative inventory, complete when its `skipped` is 0.",
             "items": {
               "type": "string"
@@ -29976,4 +29976,10 @@
             "title": "Predicted Source Adjacent",
             "type": "array"
+          },
+          "skipped": {
+            "default": 0,
+            "description": "What `created` could not account for: entries that are not regular files, or could not be read, including a directory whose whole subtree is then uncovered. A count only, 0 on a dry run; repair the tree and run again for a complete record.",
+            "title": "Skipped",
+            "type": "integer"
           },
           "summary": {
```

`--help` was dumped for the root, all 15 groups and all 78 commands (94
surfaces) at both heads, at `COLUMNS=100`. One hunk, on the command that gained
the field:

```diff
--- help-base.txt	2026-09-19 17:44:14
+++ help-head.txt	2026-09-19 18:45:45
@@ -2324,7 +2324,9 @@
  syntax — delete the sidecar to retry). It then reports                                             
  every file the pass created, classified against the cache root                                     
  (cache-owned under .godot/ vs source-adjacent, e.g. .import and .uid                               
- sidecars). An `invalid` or `failed` asset says why: `reason` names the                             
+ sidecars) — a list `skipped` qualifies: it counts what the inventory could not                          
+ read, so `created` is complete when `skipped` is 0. An `invalid` or                                
+ `failed` asset says why: `reason` names the                                                        
  check that decided it (an invalid one survives into the settled `failed`),                         
  `detail` the offending line or path when there is one, and `engine_output`                         
  the pass's own stderr lines naming that asset when this request ran a pass                         
```

Hunks 1–4 are declared rather than avoided. Each of those four texts stated the
old unconditional claim, and three of them are in the machine-readable channel a
caller branches on. Leaving them is the drift this change exists to remove.

## Prose reconciliation

One fact, one authority, seven one-sentence restatements synchronized here.
`ResourceImportResult.created`'s field description is the authority: the list is
complete when `skipped` is `0`. Each other home says the same rule in one
sentence of its own — the catalog's first also names the authority — so the
drift surface is those seven sentences, not one text. The last two rows carry a
different fact, the identity rule, whose authority is the module docstring:

| Home | Published? | Change |
| --- | --- | --- |
| `resource.py` — `created` field description | schema | the authority: "the list is complete when `skipped` is 0" |
| `resource.py` — `ResourceImportResult` docstring, "`created` lists every new file" | schema | "…, and `skipped` says how much of the tree the inventory could not see" |
| `resource.py` — same docstring, "the real run's `created` is the authoritative list" | schema | "…, complete when its `skipped` is 0" |
| `resource.py` — `pass_will_also_import` description | schema | "…the authoritative inventory, complete when its `skipped` is 0" |
| `resource.py` — command docstring | schema + `--help` | "a list `skipped` qualifies … complete when `skipped` is 0" |
| `docs/command-catalog.md` — the real run's created list | docs | "`skipped` qualifies that list … and `created`'s own description is the rule" |
| `docs/command-catalog.md` — `pass_will_also_import`'s "authoritative inventory" | docs | "…, complete when its `skipped` is `0`" |
| `src/gda/import_evidence.py` — `project_import_gaps`'s contract note | source | "…, complete when that result's `skipped` is 0 (#990)" |
| `CONTEXT.md` — `Project tree inventory` rule for an unlistable entry | docs | the identity rule, short form, deferring to the module for the rest |
| `src/gda/project_tree.py` — module docstring rule 4 | source | the identity rule in full |

`ProjectTreeMutations.skipped`'s description claims nothing about spellings or
inodes, so the identity rule leaves it true and `export run`'s schema untouched.
`README.md` with its three translations (`docs/README.es.md`, `.ja.md`,
`.zh-CN.md`) and `src/gda/skill/SKILL.md` make no completeness claim about
`created` — checked, all unchanged, so no i18n re-stamp is owed.

Runs on this PR's CI (`.github/workflows/ci.yml`, `pull_request`):

- `ruff check .` and `ruff format --check .`
- `pyright` (root and `libs/gda-balancing`)
- `scripts/harness_version_guard.py <base> <head>`
- `pytest -m "not e2e" -n 4 --dist loadgroup` — this carries the two-spellings
  regression for both commands, seven capture races (a spelling removed, a
  directory spelling retargeted, a file spelling retargeted, an unidentified
  file spelling gaining an inode, an entry beneath a directory that opens up, a
  covered file under the capture's hash rule, a covered cache entry without
  one), the mode-validator pair and the three render assertions, including the
  empty-`created` one
- `uv build`

Does NOT run on PR CI, and was run locally instead (below): every `-m e2e` suite.
The `godot-e2e` job is `schedule`/`workflow_dispatch` only, so the new
real-engine proof
(`tests/resource/test_e2e_resource_import.py::test_an_unreadable_subtree_is_disclosed_beside_what_the_pass_created`)
first runs in CI at the wave's e2e dispatch. The schema and `--help` diffs and
the `os.walk`/`rglob` grep gate have no CI job at all and are the local evidence
below.

## Validation

Host: macOS (Darwin 27.0.0), Godot 4.6.3-stable
(`~/Applications/Godot.app/Contents/MacOS/Godot`), shared with other agents;
`uptime` load averages 2.5–12 across these runs (sibling agents' load). Every
line measured at the final head `147674475132c52f5c92db7aaf4ce338c5bd9d3a`,
after the last commit, which changes comments and docstrings only.

| Gate | Command | Result |
| --- | --- | --- |
| Format | `uv run --frozen ruff format --check .` | `524 files already formatted` |
| Unit | `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2953 passed, 3 skipped, 1 warning in 36.71s` |
| e2e resource + export + project_tree | `pytest -m e2e tests/resource tests/export tests/project_tree -q -rs` (under the machine-wide e2e lock) | `81 passed, 1 skipped, 321 deselected in 254.20s` |
| e2e project | `pytest -m e2e tests/project -q -rs` (under the lock) | `118 passed, 293 deselected in 144.30s` |
| Harness | `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| Schema | `gda schema` vs `4a51d9153` | byte-identical; still 5 declared hunks vs base |
| Help | 94 `--help` surfaces vs `4a51d9153` | byte-identical; still 1 declared hunk vs base |

The one export e2e skip is the pre-existing `XDG_DATA_HOME` template case that
only the Linux engine honors.

#985's grep gate still holds — the only `os.walk` / `rglob` CALLS under
`src/gda` are the inventory's own walk and `import_evidence`'s sidecar scan:

```
src/gda/project_tree.py:259:    for dirpath, dirnames, filenames in os.walk(
src/gda/import_evidence.py:410:    for sidecar in sorted(project.rglob("*.import")):
```

## Mutant checks

Each mutant was applied at `1a7fe67e9`, the last head that changes code, in a
detached checkout, `__pycache__` cleared, the WHOLE non-e2e tier run, and the
file restored from a saved copy (never `git checkout --`).

| Mutant | Tier result | Tests that read RED |
| --- | --- | --- |
| covered paths not observed (`covers(rel) or _under(...)` → `continue`) | 4 failed, 2949 passed | `test_a_covered_file_is_observed_as_the_capture_reads_it`, `test_a_file_spelling_retargeted_to_a_second_unaccounted_inode_counts_both`, `test_a_file_spelling_that_gains_an_identity_counts_both_observations`, `test_an_entry_beneath_a_directory_that_opens_up_is_observed_not_created` |
| subtree beneath an unlistable directory not observed (the `84e0d4d80` order: `covers` observed, `_under` passed over) | 1 failed, 2952 passed | `test_an_entry_beneath_a_directory_that_opens_up_is_observed_not_created` |
| observation criterion `digest=True` (hashes under the cache root) | 1 failed, 2952 passed | `test_a_covered_cache_entry_is_observed_without_a_hash` |
| observation criterion `digest=False` (never hashes) | 1 failed, 2952 passed | `test_a_covered_file_is_observed_as_the_capture_reads_it` |
| `_hashed` without its cache-root clause (returns `detect_rewrites` alone) | 3 failed, 2950 passed | `test_the_classification_is_the_shared_function_not_a_local_rule`, `test_a_pre_existing_cache_file_is_never_reported_as_rewritten`, `test_a_covered_cache_entry_is_observed_without_a_hash` |
| early return restored (`_SkippedEntries.add` identifies a spelling only at its first observation) | 4 failed, 2949 passed | `test_a_spelling_retargeted_to_a_second_unreadable_inode_counts_both`, the two FIFO races, `test_a_covered_file_is_observed_as_the_capture_reads_it` |
| identity dedupe off (`count` counts every spelling) | 7 failed, 2946 passed | `test_a_directory_the_walk_cannot_list_is_counted_once_per_inode`, `test_a_spelling_that_vanishes_between_the_captures_does_not_split_one_inode`, `test_a_file_spelling_that_gains_an_identity_counts_both_observations`, `test_a_covered_file_is_observed_as_the_capture_reads_it`, `test_one_unreadable_inode_is_counted_once_in_the_published_report`, `test_an_unreadable_subtree_is_counted_once_beside_the_created_list`, `test_the_render_names_the_unreadable_count_beside_the_created_line` |
| identity re-taken at settle time (`settle()` re-`stat`s the first capture's spellings instead of continuing from its observations) | 3 failed, 2950 passed | `test_a_spelling_that_vanishes_between_the_captures_does_not_split_one_inode`, `test_a_file_spelling_that_gains_an_identity_counts_both_observations`, `test_a_covered_file_is_observed_as_the_capture_reads_it` |
| field not fed from the settlement (`skipped = 0` after `settle()`) | 2 failed, 2951 passed | the two `resource import` regressions above |
| dry-run validator branch removed | 1 failed, 2952 passed | `test_result_model_validates_its_mode_fields` |
| render gate `if outcome.created or outcome.skipped:` → `if outcome.created:` | 1 failed, 2952 passed | `test_the_render_prints_the_created_line_for_the_count_alone` |

The first five rows are this round's: the observation rule and its criterion
are each pinned from both sides, where at `84e0d4d80` `digest=False` and
`digest=True` both survived the tier, and the cache-root clause is pinned at the
first capture and at the second look. The last row's branch survived the whole
tier at the first head, and the assertion added for it is the one case the
disclosure rule exists for — a pass that creates nothing and still leaves part
of the tree unread.

## Deviations from the issue

One, declared rather than silent: **four published texts beyond the new field
changed**, so the schema diff carries five hunks where the issue's AC5 reads
literally as "the new field". All four stated the old unconditional promise
("every file the pass created", "the authoritative list", "the authoritative
inventory") — the exact claim the issue calls the defect — and three of them are
what `gda schema` publishes to a machine. Each carries one clause now; every
diff is pasted above.

Nothing else: no completeness policy, no per-command configuration, no path list
for skipped entries, no change to what the walk enumerates (#804's markers stay
unapplied, the cache root stays walked, the top-level `.git` stays out), no
change to `import_evidence`'s sidecar scan, and no new error code.

- **The identity is taken at observation time, and only then.** A spelling
  `stat` could not answer for when its failure was observed (a dangling link)
  stays counted under its spelling even if a target appears before the
  settlement; the settlement never re-identifies what the first capture saw,
  and a second observation of the same spelling is identified on its own. So a
  spelling unidentified in the first capture and reaching a real unreadable
  inode in the second counts twice — once as the spelling, once as the inode —
  because nothing observed says the two were one entry. A path the first
  capture covers — a spelling it could not read, or anything beneath a directory
  it could not list — is observed again under the capture's own rule and never
  called created, so a file retargeted from unreadable inode A to unreadable
  inode B counts both, while one that becomes readable keeps only its first
  failure. Seven fixtures race the two walks: spelling removal, directory
  retarget, file retarget, an unidentified file spelling gaining an inode, an
  entry beneath a directory that opens up, and the two criterion shapes.
- **A path the first capture recorded is not re-observed.** It is read again
  only as far as the rewrite-candidate rule asks, so a recorded entry that
  changed kind between the captures under the cache root (rewrites on), or
  anywhere under a caller that asked for no rewrites (`resource import`), is
  not counted; a recorded file that lost its read permission with size and
  mtime unchanged is `stat`ed and not hashed. That is the rule `CONTEXT.md`
  states — how much neither capture could account for — and making
  observation universal would move published `skipped` for shapes #990 never
  asked about, so the prose was corrected to the code, not the code to the
  prose (review at `1a7fe67e9`).
- **Hard links.** Two hard links to one unreadable inode are now one entry, by
  the same rule; the measured shapes above use symlinks and a directory link.
- **Cost is reasoned, not measured.** Each path the first capture could not
  account for — including everything beneath a directory it could not list
  that opens up before the settlement — is classified once more, under the
  capture's rule: hashed where the capture hashes, `stat`ed where it only
  `stat`s, and the hash is discarded. Bounded by the covered subtree, not by the
  tree: the review at `1a7fe67e9` measured 64 files / 64 MiB hashed and thrown
  away when a mode-000 directory holding them opens up between the captures.
  The set is empty unless a run failed to read something. No number is claimed
  for the common case.
- **The mode-000 fixtures skip where the platform lets the owner list such a
  directory**, which is also how a suite running as root skips instead of reading
  RED. Not verified as root.
- **macOS only.** The real-engine proof and every e2e line above are Godot
  4.6.3-stable on macOS. Windows was not verified; where `stat` cannot answer for
  a path there, the count falls back to the previous per-spelling behaviour by
  the same rule.
- **`gda-mcp` regeneration against the changed schema** was not exercised.
- **`export run`'s changed count reaches no consumer this repository owns** —
  it is published output. A caller that counted spellings sees one number move.


qinhaihong-red:
## Round-1 review — findings handled at `779451ec9`

Every one of the ten findings was reproduced before it was answered. Three
are fully adopted with code, five with prose, one is partially adopted, and two
are not adopted with the reason stated. Head after the fixes: **`779451ec9`**
(was `fe6ccc610`). The PR title and body are updated; the body's declared deltas
are now five schema hunks, and its "Prose reconciliation" section no longer
makes the claim P2-1 refuted.

| P2-1 | Two unqualified restatements of `created`'s promise survive in the PUBLISHED schema (the result model's docstring, `pass_will_also_import`'s description), and the PR body says they do not exist | **Fully adopt** | Confirmed by walking the head schema dump: both are real `description` values, and the model docstring's "`created` lists every new file" is the unconditional claim itself. The PR body's "point AT `created` rather than restate its promise" was wrong about both — my error, not a judgement call | `5f0e4c2f6` — three one-clause deferrals: the docstring's "lists every new file" gains "…, and `skipped` says how much of the tree the inventory could not see"; its "the real run's `created` is the authoritative list" and `pass_will_also_import`'s "the authoritative inventory" each gain "…, complete when its `skipped` is 0". The PR body's Prose reconciliation is rewritten as a ten-row table marking which homes are published | Schema diff vs base `27c04d769` now shows **5 hunks and nothing else** (command description, model description, `created`, `pass_will_also_import`, the new `skipped`); `ProjectTreeMutations` still byte-identical; `--help` still 1 hunk over 94 surfaces |
| P2-2 | The "print the created line for the count alone" branch is unpinned — `if outcome.created or outcome.skipped:` → `if outcome.created:` leaves the whole fast tier green | **Fully adopt** | Reproduced: the mutant passed 2945/3 at `fe6ccc610`. The new render test's fixture does create files, so it never exercised the `or` | `779451ec9` — `test_the_render_prints_the_created_line_for_the_count_alone`, model level, `created=[]` and `skipped=1`, asserting the exact line `  created: 0 cache-owned, 0 source-adjacent, 1 unreadable` | The same mutant now reads **RED** against the whole non-e2e tier (1 failed, 2945 passed), and is a permanent row in the PR body's mutant table |
| P2-3 | The PR title types a published-field addition as `fix`, and the squash subject IS the title | **Fully adopt** | Confirmed on the two prior slices of this milestone: `d715d50f9`'s subject is PR #987's title verbatim, and the branch's own commit types the field `feat` | PR retitled `feat(gda): resource import reports the entries its inventory could not see, and one unlistable inode counts once (#990)`. Commits unchanged — the branch's `feat` commit and the title now agree | Live title re-fetched and confirmed |
| P3-1 | The inserted sentences were not re-wrapped: one short line in `CONTEXT.md`, two long ones in the catalog | **Fully adopt** | Confirmed by line-length measurement | `5f0e4c2f6` — `CONTEXT.md:96-98` now 76/78/79 against neighbours at 76-81; `command-catalog.md` around `:1265` and `:1294` re-flowed to 78-88, the file's range, with the one short line the re-flow itself created merged away | `ruff format --check` clean (Markdown is hand-wrapped, so this is a measured check, not a tool's) |
| P3-2 | `created`'s description attaches "complete" to "Every file" rather than to the list | **Fully adopt** | Confirmed — the three post-modifiers hang off the wrong noun, in the sentence this PR declares to be the authority | `5f0e4c2f6` — "Every file the pass created, classified; the list is complete when `skipped` is 0; empty on a dry run." | In the schema diff, hunk 3 |
| P3-3 | The catalog's authority pointer does not name the field it means | **Fully adopt** | Confirmed — the nearest antecedent of "the field" is `skipped`, not `created` | `5f0e4c2f6` — "…and `created`'s own description is the rule." | In the re-wrapped catalog paragraph |
| P3-4 | `skipped`'s description is one 45-word sentence where the field it mirrors uses two, with an abstract-noun remedy | **Fully adopt** | Confirmed against `export.py:497-505` | `5f0e4c2f6` — two sentences with the imperative: "…including a directory whose whole subtree is then uncovered. A count only, 0 on a dry run; repair the tree and run again for a complete record." | In the schema diff, hunk 5 |
| P3-5 | The shipped prose describes the identity clause only for directories, while the code collapses three shapes | **Partially adopt** — prose and the declared delta, no new tests | I re-measured your five fixtures myself through the inventory at base and head: dir+directory-link 2→1, FIFO+symlink 2→1, mode-000 file+symlink 2→1, two distinct mode-000 dirs 2→2, two dangling links 2→2. `created` and `modified` empty and identical on every row | `5f0e4c2f6` — rule 4 now reads "once per filesystem identity whatever spelling reaches it" and explains it for the identity ("a ``stat`` of the failing path answers for every shape a second name can reach — a mode-000 directory (its PARENT is listable), a FIFO, an unreadable file"); `_SkippedEntries`' docstring gains your clause; `CONTEXT.md` matches. The PR body's delta table is now all five measured rows. No test added: one shape pins the rule, and three shapes of one rule would be three tests of the same branch | The five-fixture measurement is in the PR body's Behaviour delta section |
| P3-6 | The empty-`created` rendering says "created: 0 cache-owned, 0 source-adjacent" | **Not adopt** | The line is true, and the rule this PR states is that the count rides the created line — the same line, whether or not the list is empty. `export run` switches its head clause because its quiet line already says something ("unchanged outside `<cache_root>`"); this command's line has no quiet form to switch to, so a second head clause would be a branch with no fact behind it. P2-2's new test pins the line exactly as it prints | — | `test_the_render_prints_the_created_line_for_the_count_alone` asserts the full line |
| P3-7 | Pre-existing: #839's artifact exclusion has a name-only mutant that survives the fast tier | **Not adopt (out of scope)** | Reproduced and agreed: the converse of #839's gate — a file whose basename equals the artifact's, in a DIFFERENT directory — is unpinned. It is untouched by this PR (`project_tree.py:279-283` is base code) and pinning it belongs to the module's own suite | — | Carried to the maintainer as a candidate; nothing filed from this PR |

### Gates at `779451ec9`

Host macOS (Darwin 27.0.0), Godot 4.6.3-stable, shared machine, `uptime` load
3.04–3.97 across the runs. Every line measured after the last commit.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `524 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2946 passed, 3 skipped, 1 warning in 17.59s` |
| `pytest -m e2e tests/resource -q -rs` (under the e2e lock) | `45 passed, 123 deselected in 133.52s` |
| `pytest -m e2e tests/export -q -rs` (under the lock) | `36 passed, 1 skipped, 171 deselected in 77.13s` |
| `pytest -m e2e tests/project -q -rs` (under the lock) | `118 passed, 293 deselected in 144.18s` |
| `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| `gda schema` vs base `27c04d769` | 5 hunks, all declared in the PR body |
| `--help`, 94 surfaces vs base | 1 hunk, on `resource import`, declared |
| grep gate (#985) | only `project_tree.py:259` (`os.walk`) and `import_evidence.py:410` (`rglob("*.import")`) |

Mutants, each applied at `779451ec9` with `__pycache__` cleared, run against the
WHOLE non-e2e tier, restored from a saved copy: identity dedupe off → 4 RED;
field not fed from the settlement → 2 RED; dry-run validator branch removed →
1 RED; **render gate on the list alone → 1 RED** (green at the previous head —
P2-2).

Still not verified, unchanged from the PR body: anything but macOS, root, the
new e2e in CI (`godot-e2e` is dispatch-only), a real race between the two walks,
`gda-mcp` regeneration, and performance.

qinhaihong-red:
## Round-2 review — findings handled at `458e0d090`

All three findings were reproduced before they were answered, and all three are
adopted: one docs commit, one test commit, and one PR-body correction. Head after
the fixes: **`458e0d090`** (was `779451ec9`). The published surface is unchanged
by this round: `gda schema` and every `--help` page are byte-identical to
`779451ec9`, so the declared deltas against the base stay five schema hunks and
one help hunk.

| P3-1 | Rule 4 says `stat` "answers for every shape a second name can reach", and its next sentence (a dangling link: two spellings, two entries) is the counter-example | **Fully adopt** | Confirmed by reading the two sentences together; the behaviour is the pinned one (the review's `_unidentified` mutant reads RED) | `71d095d0a` — rule 4 now uses the class docstring's wording: "A `stat` of the failing path names that inode, so any unreadable inode a second name reaches — a mode-000 directory (its PARENT is listable), a FIFO, an unreadable file — is one entry." | Prose only; `CONTEXT.md`'s short form already carried the fallback clause and is untouched |
| P3-2 | `_unlistable` is four byte-identical copies, three added by this branch | **Fully adopt** | Confirmed: one distinct body across the four modules | `458e0d090` — one `unlistable` in `tests/support.py`, beside the other probes that decide a skip; the four modules import it; one now-unused `import os` dropped | Fast tier and the three e2e suites below |
| P3-3 | The body's "one authority, six deferrals … restates nothing" is not what the files say | **Fully adopt** | Confirmed by counting at the head: the authority plus seven one-sentence restatements (`resource.py` ×4, the catalog ×2, `import_evidence.py` ×1), the catalog's first also naming the authority; the table's last two rows are the identity rule, a different fact | PR body — "one authority, seven one-sentence restatements synchronized here … the drift surface is those seven sentences, not one text. The last two rows carry a different fact, the identity rule, whose authority is the module docstring" | Re-fetched after `gh pr edit`; the edit landed |

The optional follow-up the review names — giving the `authoritative list` /
`authoritative inventory` homes the catalog's pointer form so the rule has one
place to change — is not taken here: issue #990's Decision asks each home to say
so in one sentence, and whether to tighten that is the maintainer's call.

### Gates at `458e0d090`

Host macOS (Darwin 27.0.0), Godot 4.6.3-stable, shared machine, `uptime` load
3.58–5.00 across the runs. Every line measured on the final tree, after the last
edit.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `524 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2946 passed, 3 skipped, 1 warning in 19.89s` |
| `pytest -m e2e tests/resource tests/export tests/project_tree -q -rs` (under the e2e lock) | `81 passed, 1 skipped, 314 deselected in 206.29s` |
| `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| `gda schema` vs `779451ec9` | byte-identical |
| every `--help` page vs `779451ec9` (95 pages) | byte-identical |

qinhaihong-red:
## READY — orchestrator re-verification at `c45c61a9b`

Three fresh axes were run on a detached checkout at the round-2 head
`458e0d090`: Standards, Spec (its own real-engine fixtures under the e2e lock,
the schema and 98 help surfaces rebuilt from both checkouts), and
Domain-modular architecture; plus the orchestrator's own probes (inode-once on
three shapes in-process, the dry-run refusal's mutant, a real `resource import`
with a mode-000 directory). Their findings and what happened to each:

| Axis | Finding | Disposition |
| --- | --- | --- |
| Standards | The published command description said `skipped` "counts what the walk could not read"; in this command's catalog section "the walk" is the engine-side walk, and in `--help` it had no antecedent | `c45c61a9b` — "the inventory"; the declared delta is unchanged, 5 schema hunks and 1 help hunk against the base |
| Standards + Architecture (independently) | `_SkippedEntries` answered `len()` on the identity and `in` on the spelling | `29d0e4738` — `count` and `covers(rel)`; one settle site; no behaviour change |
| Standards, Spec | CONTEXT.md called the inventory the fact behind `resource import`'s `created` list only | `c45c61a9b` — "`created` list and `skipped` count", re-wrapped to the sentence boundary |
| Standards | The SKILL.md `resource` row documented neither `created` nor `skipped` while the `export` row documents its `skipped` | `c45c61a9b` — one clause names the pair and the completeness rule |
| Standards | The body's validation table was pinned to `779451ec9` | Body re-pinned to `c45c61a9b`; every line re-measured, and the quoted hunks updated |
| Spec | Three schema hunks beyond the Decision's list (the result docstring, the command description, `pass_will_also_import`) | Declared since round 1: each deletes the unconditional promise the issue calls the defect; the maintainer's call stands |
| Spec | The identity rule collapses any unreadable inode, wider than "cannot be listed" | Declared and tabled in the body; reproduced 2→1 through a real export on the file shape |
| Spec | Both `skipped` descriptions say "entries" while the count is per inode, and no machine channel states the identity rule | Not adopted: neither description names a unit, the reliable branch is `== 0`, and `export run`'s description stays byte-identical by the Decision's "if its wording changes" |
| Standards | Twin descriptions and the `, N unreadable` phrase in both renderers | Not adopted: the mirror is the Decision's contract; two consumers, one word |
| Standards | `unlistable()` restores only on `False`, so each site keeps its own `try/finally` | Not adopted here: test-only tidy-up, a candidate for a later pass |
| Architecture | No test guards the seven one-sentence restatements | Recorded as a validation need; the maintainer's call whether to guard |
| Spec (note) | A mode-000 DIRECTORY makes the engine itself crash during an export's scan (`engine_crashed`, signal 11, `_scan_new_dir`), with or without a link | Engine behaviour, pre-existing, correctly reported as a typed envelope; the export test's fake runner is the right instrument for that shape; noted for the maintainer |

PR CI is 7 pass / 4 skip at `c45c61a9b`; `Closes #990` is in the body; author
and committer are `haihong.qin` on all eight commits; no attribution anywhere.

qinhaihong-red:
## Independent adversarial review — Changes required

Reviewed exact range `27c04d76969b06733e80c4d27895aa7fd982d66f...c45c61a9b85f827cebd61d12fb5462d7ae7bdfda`. I did not read or use existing PR reviews or PR comments.

### Artifact review

**Changes required.**

1. **[P2] The agent skill overstates `created` as every file the pass “wrote”.** [`src/gda/skill/SKILL.md:192`](https://github.com/aigengame/godot-agent/blob/c45c61a9b85f827cebd61d12fb5462d7ae7bdfda/src/gda/skill/SKILL.md#L192) says `created` reports “every file the pass wrote”. The authoritative model says “every file the pass created”, and `resource import` deliberately calls `ProjectTreeInventory.capture(..., detect_rewrites=False)`, so pre-existing files rewritten by the import pass are not reported. In this repository, `created` and `modified` are distinct established terms. Change “wrote” to “created” (or “new file the pass created”) so an agent does not mistake this for a mutation-completeness record.
2. **[P2] Issue #990 gives mutually incompatible schema-delta instructions.** Its decision requires the `created` description and catalog to explain completeness through `skipped`, but later says the exact declared deltas exclude those schema descriptions and that “Nothing else” may appear in the schema diff. The head necessarily changes four existing schema descriptions plus the new property. Amend [#990](https://github.com/aigengame/godot-agent/issues/990) to list the required reconciliation explicitly; otherwise reviewers cannot satisfy both acceptance statements.

**Terminology and prose assessment.** Sources: `CONTEXT.md`, ADR-0040, the `Project tree inventory` module contract, the command catalog, the built-in skill, and the complete issue/PR body. Reference: ASD-STE100. I inspected every changed Markdown passage and every requirement/decision/procedure, checking the established distinctions among `Project tree inventory`, `created`, `modified`, `skipped`, “walk”, and “inventory”, plus causal/scope wording. Finding 1 is a substantive terminology error and finding 2 is an acceptance ambiguity. No other substantive prose issue was found. Windows and root-mode permission behavior were not tested.

Ownership and dependency direction otherwise pass: `gda.project_tree` remains the single owner of walk/settlement behavior; `resource` owns only its result projection and renderer; both command groups depend downward on the shared owner, with no new cycle or generic policy layer.

- **[P2] Identity is captured too late for a two-capture settlement.** [`_SkippedEntries.add`](https://github.com/aigengame/godot-agent/blob/c45c61a9b85f827cebd61d12fb5462d7ae7bdfda/src/gda/project_tree.py#L319-L333) re-`stat()`s first-capture failures only during `settle()`. Counterexample at this exact head: capture a mode-000 `locked/` plus `alias -> locked/`, remove `alias` before settlement, then settle. The first capture observed one inaccessible inode through two spellings, but the result is `skipped == 2`: vanished `alias` becomes an unidentified spelling while `locked` becomes the inode identity. Preserve the identity when each failure is observed; use the spelling fallback only when identity was unavailable at that observation.

Load-bearing claim `A994-identity-unique` has evidence state `open`/contradicted: stable-tree fixtures pass, while the mutation-between-captures counterexample fails. The claim is disconfirmed whenever a spelling disappears or retargets before settlement; add that regression for the shared owner, then replay both `resource import` and `export run` projections.

### Standards

**Pass — 0 findings.** Checked `AGENTS.md`, `CONTEXT.md`, ADR-0040/0032, the complete diff, and all twelve baseline smells. `_SkippedEntries` stays inside the shared owner, and `resource` keeps only its vertical public slice; no standard breach or actionable smell was found.

### Spec

1. **[P2] Identity deduplication is evaluated too late.** The counterexample above violates “one filesystem inode that cannot be listed is one entry in `skipped`, whatever spelling reached it.” Store the observation-time identity rather than recomputing it after the run.
2. **[P3] The new field exceeds the explicit one-sentence contract.** [`ResourceImportResult.skipped`](https://github.com/aigengame/godot-agent/blob/c45c61a9b85f827cebd61d12fb5462d7ae7bdfda/src/gda/commands/resource.py#L739-L746) uses two sentences, while #990 says the one field is “described in one sentence” and repeats “one integer field and one sentence.” Combine the meaning, dry-run value, and remedy into one sentence.

There is no separate scope finding for the five schema hunks: the issue itself requires reconciliation of the old unconditional descriptions, which is why the contradictory issue text in Artifact finding 2 must be repaired.

### Validation

- `git diff --check`, Ruff check/format, Pyright, and the harness version guard: pass.
- Full non-e2e tier: `2944 passed, 5 skipped` locally; focused changed suites: `87 passed`.
- New real-engine unreadable-subtree acceptance test: `1 passed` on Godot 4.6.3.
- Independent FIFO+symlink, FIFO+hardlink, and two-dangling-link probes returned `1`, `1`, and `2` as documented.
- Normalized schema comparison confirms exactly five hunks; hosted CI is green, but its Godot e2e job is skipped.

qinhaihong-red:
## Independent adversarial review — findings handled at `4a51d9153`

Every finding was reproduced before it was answered: two are adopted in the
issue, one in code with a regression, one in prose. Head after the fixes:
**`4a51d9153`** (was `c45c61a9b`). Issue #990 and the PR body are updated.
`gda schema` and every `--help` page are byte-identical to `c45c61a9b`, so the
declared delta stays five schema hunks and one help hunk against the base.

| Artifact P2 | SKILL.md says `created` is "every file the pass wrote"; `created` and `modified` are distinct terms, and `resource import` asks for no rewrite detection | **Fully adopt** | Confirmed | `4a51d9153` — "every file the pass created" | Prose only |
| Artifact P2 | Issue #990's Decision asks the `created` and catalog sentences to change, then lists deltas that exclude them and says "Nothing else" | **Fully adopt** | Confirmed: the two statements cannot both hold | Issue #990, `## Amendments (2026-09-20)`: the declared delta restated as the five hunks (the new field; `created`; the result docstring; `pass_will_also_import`; the command description, which is the help hunk), hunks 3 to 5 each removing the unconditional promise the issue calls the defect | Schema recount at the head: 5 hunks / 1 help hunk |
| Architecture P2 / Spec P2 | Identity is captured too late: `locked/` + `alias -> locked/`, `alias` removed before the settlement → `skipped == 2` | **Fully adopt** | Reproduced at `c45c61a9b`: the new regression reads 2 there | `945c9fecd` — each failing spelling is identified when its failure is observed, in whichever capture observes it; the first capture keeps its observations (`unreadable`: spelling → identity, or `None` where `stat` could not answer) and the settlement continues from them. Rule 4, the helper's docstring and the issue amendment state the rule | New regression `test_a_spelling_that_vanishes_between_the_captures_does_not_split_one_inode`; reverse-mutant (the settlement re-`stat`s the first capture's spellings): 1 RED, 314 green; e2e resource + export + project_tree and project green at the head |
| Spec P3 | `skipped`'s description is two sentences; the issue says one | **Adopt in the issue** | Confirmed: the field mirrors `ProjectTreeMutations.skipped`'s two sentences, whose second sentence is the remedy | Issue amendment: the Decision's "one sentence" counts the qualification each other home adds; the field's own description mirrors `export run`'s two | — |

### Gates at `4a51d9153`

Host macOS (Darwin 27.0.0), Godot 4.6.3-stable, shared machine, `uptime` load
4.13–4.84 across the runs. Every line measured after the last commit.

| Gate | Result |
| --- | --- |
| `uv run --frozen ruff format --check .` | `524 files already formatted` |
| `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2947 passed, 3 skipped, 1 warning in 21.27s` |
| `pytest -m e2e tests/resource tests/export tests/project_tree -q -rs` (under the e2e lock) | `81 passed, 1 skipped, 315 deselected in 209.83s` |
| `pytest -m e2e tests/project -q -rs` (under the lock) | `118 passed, 293 deselected in 135.88s` |
| `python scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| `gda schema` / 95 `--help` pages vs base `27c04d769` | 5 hunks / 1 hunk, all declared; byte-identical to `c45c61a9b` |

qinhaihong-red:
## Independent adversarial re-review — Changes required

Re-reviewed exact head `4a51d9153c57ba78294dc2396d5b4cd85ea1a3e5` against base `27c04d76969b06733e80c4d27895aa7fd982d66f` and the previously reviewed head `c45c61a9b85f827cebd61d12fb5462d7ae7bdfda`. The response to the prior review was treated as a set of claims; the conclusion below comes from the current issue, diff, repository authority, and an independent counterexample.

### Artifact review

**Pass.** The SKILL row now says `created` means files the pass created. Issue #990's amendment consistently declares the five schema hunks and explicitly authorizes the shared two-sentence `skipped` description. The prior artifact findings are closed.

### Domain modular architecture / Standards / Spec

1. **[P2] A spelling retargeted to a second unreadable inode loses that second identity.** [`_SkippedEntries.add`](https://github.com/aigengame/godot-agent/blob/4a51d9153c57ba78294dc2396d5b4cd85ea1a3e5/src/gda/project_tree.py#L337-L341) returns immediately when the project-relative spelling is already present in `_observed`. That conflates two different questions the helper itself says are separate: whether the first capture covered this spelling, and which filesystem identities were observed failing.

   Reproduced at this exact head:

   - First capture: `locked` and `alias -> locked` both reach unreadable inode A, so the saved map correctly holds A under both spellings.
   - Between captures: retarget `alias` to a different unreadable inode B. `(st_dev, st_ino)` for A and B are distinct.
   - Settlement observes the new failure through `alias`, but `add("alias")` returns before sampling B. The result is `skipped == 1`, although two inaccessible identities were observed.

   This contradicts the current issue amendment (“identified at the moment its failure is observed, in whichever capture observes it”) and the `Project tree inventory` rule in `CONTEXT.md` (“once per filesystem identity whatever spelling reaches it”). It is also the retarget half explicitly named by the new module prose.

   Separate spelling coverage from observed identity accumulation: `covers()` should retain the first-capture spelling set needed to prevent a false `created`, while each actual failure observation should sample and union its inode identity (with spelling-based handling where `stat` cannot answer). Add the retarget-A-to-B regression beside the new vanishing-alias regression.

### Validation

- The exact-head counterexample above reports distinct inode pairs and `settled_skipped: 1`.
- Existing changed-area suites pass (`117 passed` in the broad local slice); the existing real-engine unreadable-subtree test also passes, but neither exercises retargeting to a second unreadable inode.
- Ruff and Pyright pass; `git diff --check` passes.
- `gda schema` is byte-identical before and after the fix commits. Required hosted CI checks are green; hosted Godot e2e remains skipped by workflow design.

The old vanishing-alias counterexample is fixed, but the amended identity-at-observation rule is not yet complete at this head.

qinhaihong-red:
## Re-review handled — head `65c07340df4898a91fa996f3413f4d5c49526cc8`

| Finding | Decision | Change |
| --- | --- | --- |
| [P2] A spelling retargeted to a second unreadable inode loses that second identity | Accepted. `_SkippedEntries.add` recorded a spelling once, so the second observation of `alias` was never sampled and inode B went uncounted: the spelling record and the identity count were one dict. | `1abb21f2b`: the helper keeps the two questions on two records. `covers(rel)` still answers on the spelling (whether a capture already failed to read this path); `count` answers on the set of identities every observation reached — the `(st_dev, st_ino)` pair, or the spelling where `stat` cannot answer — so every observation is identified and a retargeted spelling adds the inode it reaches now. Regression `test_a_spelling_retargeted_to_a_second_unreadable_inode_counts_both` beside the vanishing-alias one: `locked/` + `alias -> locked/`, `alias` retargeted after the capture to a second mode-000 directory OUTSIDE the project so only the link reaches it; reads 1 at `4a51d9153`, 2 at head. With the early return restored, that test alone reads RED across the whole non-e2e tier. |

Rule 4, the `CONTEXT.md` entry and issue #990's amendment already stated the identity as taken "at the moment the failure is observed, in whichever capture observes it"; the code fell short of that statement. A fresh adversarial pass at `1abb21f2b` (14 hand-built shapes, the PR's five export fixtures re-measured 1/1/1/2/2) found one P3: the `unreadable` field comment and the issue amendment carried only the "does not split" half of the retarget rule — `65c07340d` adds the clause to the comment, and the amendment now says the same and names both race regressions.

Two consequences of the rule are declared under residual risk rather than changed: a spelling `stat` could not identify in the first capture and a real unreadable inode reached through it in the second count twice, because nothing observed says they were one entry; and a covered spelling is not re-read by the settlement, so a FILE retargeted to a second unreadable inode is never observed failing again and stays one entry (a directory is, because the walk's listing is the observation). Re-reading covered spellings for it would spend a read to refine a disclosure that is not a measure.

Gates at the head: ruff check / format clean, pyright 0 errors, non-e2e tier `2948 passed, 3 skipped`, e2e `tests/resource tests/export tests/project_tree` `81 passed, 1 skipped` and `tests/project` `118 passed` under the machine-wide lock, `gda schema` and every `--help` page byte-identical to `4a51d9153` (still five schema hunks and one help hunk against the base). All six mutant rows re-run at this head against the whole tier (table in the body). PR body validation, mutant and residual-risk sections re-measured and re-stated at this head.

qinhaihong-red:
## READY — head `65c07340df4898a91fa996f3413f4d5c49526cc8`

The re-review's P2 is fixed at `1abb21f2b` (reply above). A fresh adversarial
pass was then run at that head on a detached read-only checkout: 14 hand-built
shapes against the identity rule, the regression and three mutants re-run, the
PR's five export fixtures re-measured (1/1/1/2/2, unchanged), ruff / pyright /
fast tier / e2e under the lock, and `gda schema` plus every `--help` surface
byte-identical to `4a51d9153`. Verdict: Pass, one P3 — the `unreadable` field
comment and the issue amendment carried only the "does not split" half of the
retarget rule — landed as `65c07340d`, a comment-only commit; every gate and all
six mutant rows were re-run at that head and the body's validation section is
pinned to it.

PR CI is 7 pass / 4 skip at `65c07340d`; `Closes #990` is in the body; author
and committer are `haihong.qin` on every commit of the branch; no attribution
anywhere.

qinhaihong-red:
## Independent adversarial re-review (round 2) — Changes required

Re-reviewed exact head `65c07340df4898a91fa996f3413f4d5c49526cc8` against base `27c04d76969b06733e80c4d27895aa7fd982d66f` and the previously reviewed head `4a51d9153c57ba78294dc2396d5b4cd85ea1a3e5`. The author replies were treated as claims to verify, not as review evidence.

### Artifact review

**Changes required.** The PR body's residual-risk section says that a spelling unidentified in the first capture and reaching a real unreadable inode in the second “counts twice.” At this head, a dangling file symlink retargeted to a mode-000 regular file returns `skipped == 1`, not `2`, because the second capture exits before recording the current identity. The same section separately discloses the early exit for file retargets, so the artifact is internally inconsistent about the behavior.

**Terminology and prose assessment.** Sources: issue #990, `CONTEXT.md`'s `Project tree inventory`, and `src/gda/project_tree.py`; writing reference: ASD-STE100. I inspected the complete current issue, all PR-body requirements/decision/validation/residual-risk prose, and the changed module comments for identity, observation, spelling, capture, and settlement usage. Apart from the factual inconsistency above and its matching code finding below, I found no substantive terminology or prose issue. I did not re-verify every hand-built fixture claimed in the PR body.

The rule remains under the correct owner, `gda.project_tree`, and the fix usefully separates spelling coverage from accumulated identities. However, the execution paths are still incomplete: directory failures call `add()` through `os.walk.onerror`, while a yielded file spelling already present in the first-capture record takes [`covers(rel) -> continue`](https://github.com/aigengame/godot-agent/blob/65c07340df4898a91fa996f3413f4d5c49526cc8/src/gda/project_tree.py#L475-L480) before its second-capture failure can reach the identity owner. Claim evidence state for “every failure observation contributes its identity”: **open**; the directory counterexample is closed, but the file counterexamples below disconfirm the general claim.

### Standards

1. **[P2] Covered file entries still bypass identity-at-observation.** The early `continue` at `src/gda/project_tree.py:479` means the new two-record design is used by directory failures but not by file-shaped entries yielded by the walk. This conflicts with the repository's `Project tree inventory` rule that an unaccounted entry is counted once per filesystem identity. No additional documented-standard breach or actionable baseline smell was found.

### Spec

1. **[P2] The identity-at-every-observation amendment is only implemented for directory failures.** Issue #990 requires each failing spelling to be identified “at the moment its failure is observed, in whichever capture observes it,” and says an inode reached through a retargeted spelling is its own entry. Independent exact-head probes produced:

   | Transition between captures | Distinct identities | Expected | Actual |
   | --- | ---: | ---: | ---: |
   | `alias -> unreadable file A` to `alias -> unreadable file B` | 2 | `skipped == 2` | `1` |
   | `alias -> missing` to `alias -> unreadable file B` | spelling + inode | `skipped == 2` (also claimed by the PR body) | `1` |
   | directory inode A to directory inode B | 2 | `skipped == 2` | `2` |

   Preserve spelling-based coverage so a pre-existing path is not misreported as `created`, but independently re-observe a covered yielded entry before skipping it and add its current failure identity when that second observation also fails. Add file-transition regressions beside the directory-retarget regression.

### Validation

- Prior directory A→B counterexample: fixed (`2`). The two file-transition counterexamples above: reproduced (`1`).
- Changed-area suites: `118 passed`; full non-e2e tier: `2946 passed, 5 skipped`.
- Ruff check/format, Pyright, and `git diff --check`: pass.
- `gda schema` is byte-identical to `4a51d9153`; required hosted CI checks are green. Hosted Godot e2e is skipped by workflow design; the existing real-engine unreadable-subtree test passes but does not exercise a file retarget between captures.

The previous directory-retarget finding is closed, but the same rule remains incomplete for file entries at this exact head.

qinhaihong-red:
## Review handling complete — `84e0d4d8012ace1c33c835a904d6173cf0046201`

| Covered file entries bypass the second capture's identity observation | **Fully Adopt** | At `65c07340d`, unreadable file A → B and dangling spelling → unreadable file both returned `skipped == 1`; the two new regressions were RED before the code change | `84e0d4d80`: a covered yielded entry is classified again before spelling coverage suppresses `created`; a second failure adds its current identity. Added FIFO regressions for inode A → B and spelling → inode | Both probes now return `2`; focused `120 passed`; full non-e2e `2948 passed, 5 skipped` |
| PR body says spelling → inode counts twice while also saying covered files are not re-read | **Fully Adopt** | The two statements described incompatible behavior, and the executable probe read `1` | Reconciled the implementation/validation/residual-risk sections with the final behavior and exact head | Remote PR body read back after update; scope guard passed |

### Backtrace result

Repeated review did expose **symptom-patch accumulation**, but not a need for a rewrite or a broader platform mechanism. The upstream cause was one incomplete composition boundary: spelling coverage owns suppression of a false `created`, while failure identity owns `skipped`; the settlement applied coverage before the file path completed its second observation. The current `gda.project_tree` owner and two-record design are retained. The fix changes only that ordering and adds no new public surface, state kind, compatibility policy, or NFR obligation. **NFR-driven platformization: absent for this recovery.**

### Final validation

- Ruff check: pass; Ruff format: `524 files already formatted`.
- Pyright, root and `libs/gda-balancing`: `0 errors`.
- Harness version guard: pass; `git diff --check`: pass.
- `gda schema`: byte-identical to `65c07340d` and still the same five declared hunks against the base.
- Hosted checks at the final head: **8 pass / 4 skip**. The skipped Godot e2e job remains workflow policy.
- Real-engine e2e was not repeated for this pure-Python settlement control-flow fix; the last PR-body e2e evidence remains pinned to `65c07340d`, and the exact changed boundary is covered by the two RED→GREEN unit regressions.

The branch and PR body are now at the final head above; no unresolved review finding remains from this thread.

qinhaihong-red:
## Round-2 fix evaluated and completed — head `1a7fe67e965097677b2b214a99b93607413a4ab7`

`84e0d4d80` was correct for the two counterexamples it answered and its two FIFO regressions pin that boundary. Evaluated against its own principle — coverage suppresses `created`, never observation — it left two gaps, both reproduced by probe at that head:

| Gap at `84e0d4d80` | Evidence | Change at `1a7fe67e9` |
| --- | --- | --- |
| The same directory/file asymmetry one level down: beneath a directory the first capture could not list that opens up before the settlement, a locked subdirectory is observed through the walk's error sink and counted, while a FIFO beside it is passed over unobserved | probe: locked subdir beneath → `skipped == 2`; FIFO beneath → `1`; `settle()`'s docstring said "everything beneath … is passed over", which the error-sink route contradicted | One coverage branch: `covers(rel) or _under(rel, unlistable_dirs)` observes the path under the capture's criterion and never calls it created. Regression `test_an_entry_beneath_a_directory_that_opens_up_is_observed_not_created` (2, `created == []`); the `84e0d4d80` order as a mutant reads RED on it alone |
| The second look's failure criterion was a copy of the capture's rule minus its cache-root clause (`digest=self.detect_rewrites`), and nothing pinned it: `digest=False` and `digest=True` both survived the whole tier | probe: a covered cache FIFO replaced by a mode-000 regular file counted 2 with rewrites on, under a criterion the capture never applies below the cache root, and 1 without | `_hashed(rel, detect_rewrites)` is the one rule for every read a capture takes — the first capture's read, the settlement's rewrite candidate, the second look. Two regressions pin it from both sides: a covered file outside the cache root replaced by a second mode-000 inode counts both (rewrites on, the capture reads it in full; `digest=False` mutant RED), a covered cache FIFO replaced by a mode-000 regular file stays one entry (`digest=True` mutant RED) |

Evidence the round-2 reply left at earlier heads, now measured at `1a7fe67e9`: e2e `tests/resource tests/export tests/project_tree` `81 passed, 1 skipped` and `tests/project` `118 passed` under the machine-wide lock; all ten mutants (the five original rows, early return, observation removed, subtree not observed, the two criterion mutants) RED against the whole non-e2e tier, each on the test meant to catch it; fast tier `2953 passed, 3 skipped` (the round-2 reply's `2948 passed, 5 skipped` does not match this host at `84e0d4d80`, where it reads `2950 passed, 3 skipped` — the two FIFO regressions look skipped in that run); ruff / format / pyright clean; `gda schema` and every `--help` page byte-identical to `4a51d9153` (still five schema hunks and one help hunk against the base); PR CI 7 pass / 4 skip.

Prose: `settle()`'s docstring states the observation rule and the coverage rule in the same paragraph; the `unlistable_dirs` comment says what coverage suppresses and that the subtree is still observed; `_hashed`'s docstring carries the cache-root sentence that `capture()`'s comment carried (#741). Rule 4, the `CONTEXT.md` entry and issue #990's amendment are unchanged: the code now does what they say. The PR body's What changed, validation, mutant table (ten rows) and residual risk are re-pinned to this head.

qinhaihong-red:
## READY — head `147674475132c52f5c92db7aaf4ce338c5bd9d3a`

A fresh adversarial pass was run at `1a7fe67e9` on a detached read-only checkout: 20-odd hand-built shapes against the observation rule, six mutants against the whole tier, `tests/export` non-e2e (the export report unchanged), ruff / pyright / fast tier / e2e under the lock, `gda schema` and all 94 `--help` pages byte-identical to `4a51d9153`. Verdict: the code model is closed — the directory/file asymmetry is gone and the `84e0d4d80` order reads RED as a mutant — but the prose over-claimed, one P2 and three P3:

| Finding | Disposition |
| --- | --- |
| [P2] "Every path this walk reaches is observed" was false: a path the first capture RECORDED is re-read only as far as the rewrite-candidate rule asks, so a recorded cache file replaced by a FIFO (rewrites on) or any recorded file (rewrites off — `resource import`'s settlement) is not counted; `_hashed`'s "every read the two captures take" was false of the two `stat`-only reads | Adopted as the reviewer asked: prose, not code — universal observation would move published `skipped` for shapes #990 never asked about, and `CONTEXT.md`'s "how much neither capture could account for" is already the true rule. `147674475`: rule 4 counts "an entry a capture tried to read" and names the recorded-path exception; `settle()` states the two-part rule (the record decides what is observed, coverage decides what may be called created or modified); `_hashed` governs "every read that asks for BYTES" and no other; the prefix comment says "like every path the first capture could not account for" |
| [P3] The second look's discarded hash on a locked subtree that opens up was undeclared (measured: 64 files / 64 MiB hashed and thrown away) | `147674475`: declared beside the branch and in `settle()`'s docstring — the second look pays the capture's read so "could not be read" means the same both times; its cost is the covered subtree the first capture never paid for, empty unless a run failed to read something. Also in the body's residual risk |
| [P3] "everything beneath either spelling" — `_under` tests every recorded spelling | `147674475`: "EVERY spelling that reaches it" |
| [P3] `tests/support.py`'s `unreadable()` leaves each site its own restore burden, like `unlistable()` | Not changed here: test hygiene, no defect; the backlog item for `unlistable()` now covers both helpers |

The reviewer's own extra mutant — `_hashed` without its cache-root clause — was re-run at `1a7fe67e9`: `3 failed` (`test_the_classification_is_the_shared_function_not_a_local_rule`, `test_a_pre_existing_cache_file_is_never_reported_as_rewritten`, `test_a_covered_cache_entry_is_observed_without_a_hash`), so the clause is pinned at the first capture and at the second look.

`147674475` changes comments and docstrings only (verified by diff). At it: ruff / format clean, pyright 0, non-e2e tier `2953 passed, 3 skipped`, e2e `tests/resource tests/export tests/project_tree` `81 passed, 1 skipped` and `tests/project` `118 passed` under the lock, `gda schema` and every `--help` page byte-identical to `4a51d9153`; the eleven-row mutant table in the body is measured at `1a7fe67e9`, the last code head. PR CI 7 pass / 4 skip; `Closes #990` is in the body; author and committer are `haihong.qin` on every commit of the branch; no attribution anywhere.

Issue #1001, opened 2026-09-21
export run: output_path's three descriptions call it a resolved absolute path, which a res:// or user:// output is not


qinhaihong-red:
## What to build

`gda export run` publishes `output_path` as the destination the caller named (#403): a filesystem `--output` made absolute against the invoker's cwd, a virtual address (`res://…`, `user://…`) passed through as spelled, and a preset's configured `export_path` resolved against the project. Three texts say it is "the resolved absolute path": the published field description of `ExportRunResult.output_path` (what `gda schema` emits), the result's class docstring, and the `export run` section of `docs/command-catalog.md`. An agent that reads the schema and receives `res://build/game.x86_64` is told it holds a filesystem path it does not hold.

Make the three texts state the rule #403 implements, in one sentence each, without changing what the command publishes.

## Acceptance criteria

- [ ] The field description, the class docstring and the catalog say what `output_path` is: the caller's spelling — a filesystem `--output` absolute against the invoker's cwd, a `res://` or `user://` address as given, a configured `export_path` resolved against the project.
- [ ] The field-description change is the only schema hunk and is declared in the PR; `--help` is byte-identical.
- [ ] No change to the published value, the CLI normalizers or the export flow.

## Blocked by

None - can start immediately

## Amendments (2026-09-21)

**The declared schema delta is two hunks, not one.** `ExportRunResult`'s class docstring is published as the output schema's own `description`, so correcting its sentence is a second hunk beside the field description. AC2 reads: the field description and the result docstring are the only two schema hunks, both declared in the PR; `--help` is byte-identical.

**A fourth text carried the same claim** — the `export smoke` section of `docs/command-catalog.md` said "the absolute `output_path` from `export run` passes straight through"; it now says a filesystem one does, and that a `res://` one names the artifact only inside its project. The recipe in the `gda skill` `export` row ("feed `smoke` `run`'s `output_path`") is left as it is: it holds for the configured `export_path` and for a filesystem `--output`, which are the common cases.

## Amendments (2026-09-21, second)

[...]

PR #999, opened 2026-09-21
fix(gda): a res:// export output resolves through the one res:// authority for parent creation and artifact exclusion (#997)

qinhaihong-red:
## Summary

`gda export run` answered "where does a `res://` destination land on disk" in
three places with three rules. The mutation report joined the RAW remainder, so
`--output res://build\game.x86_64` kept out a file named `build\game.x86_64` that
nothing ever wrote while the engine wrote `build/game.x86_64`. Parent creation
dropped every `://` address, so a `res://` destination with a missing parent came
back as an opaque `export_failed` although `docs/command-catalog.md:1331` promises
the parents are made and reported. And the native export received the caller's raw
spelling, so a `..` that steps through a directory link was walked PHYSICALLY
there while gda collapsed it lexically here — two different files.

There is now ONE location, owned by the path authority, and the engine is handed
it too.

Closes #997

## What changed

**`gda.project.res_location(path, project)` — the one owner.** The canonical
`res://` address (`canonical_res_path`, the engine's own `String::simplify_path`
semantics) under the absolute project (`project_absolute`). A bare `res://`, and
any spelling that collapses to it, is the project directory. Its docstring states
the rule AND the boundary: the reading is lexical, so a `..` that would step
through a directory link lands where the address canonicalizes to, not where the
link leads — which is why the location has to be handed on rather than the address
re-spelled. An escaping remainder is joined as written, so its surviving `..` is
resolved by whoever walks the result. It owns the export destination and the
location a refusal reports (`_anchored_target`), not every `res://` join in the
codebase — the import-evidence reads, `resource import`'s asset mapping and the
daemon's session paths still build their own, and the docstring says so.
`_anchored_target`'s `res://` branch reads it, and `target_location`'s and
`owning_project`'s suites are green unchanged.

**`commands/export.py` keeps only policy.** `_res_output_location` is the `res://`
scheme check, the empty-remainder answer (a bare `res://` names no artifact and
has no parent to make) and "another scheme is not this namespace"; everything else
is the authority's. The duplicate `_absolute_project` is DELETED — the
configured-`export_path` channel reads `gda.project.project_absolute`, the same
cwd rule plus the module's total `~`.

**The native export consumes the same location.** `export_runner.run` is given the
resolved absolute path for a `res://` destination, as it is already given the
absolute path of a filesystem `--output`; another virtual scheme still passes
through as the engine's to resolve. What the engine writes, what the preflight
prepares and what the walk keeps out are now the same file by construction. The
[...]

Issue #997, opened 2026-09-21
export run: a res:// output resolves through the one res:// authority for parent creation and artifact exclusion


qinhaihong-red:
## What to build

`gda export run --output res://…` answers "where does this land on disk" twice with its own rules instead of `gda.project`'s `res://` primitive. Measured against Godot 4.6.3: `--output res://build\game.x86_64` is written by the engine to `build/game.x86_64` (the engine folds `\` to `/`), but the export excludes its artifact by the raw spelling, so the artifact is reported in `project_tree_mutations.created`; and `--output res://build/game.x86_64` with no `build/` directory fails inside the engine ("export path does not exist") because parent creation skips every `://` address, while the command catalog promises that missing output parent directories are created and reported in `created_dirs`.

Resolve a `res://` output once, through the canonical `res://` form the path authority already owns, and use that location for both parent creation and artifact exclusion. A filesystem output keeps its behaviour; `output_path` in the result keeps the caller's spelling.

## Acceptance criteria

- [ ] `--output res://<dir>/<file>` with a missing `<dir>` creates it before the export and reports it in `created_dirs`, as a filesystem output does.
- [ ] The artifact of a `res://` output spelled with `\`, `//` or `.` segments is excluded from the mutation report, as the engine writes it.
- [ ] The `res://` resolution has one owner in `gda.project`; a reverse mutant of the export's exclusion reads RED on a unit regression.
- [ ] Schema and help unchanged, or the delta declared.

## Blocked by

None - can start immediately

## Amendments (2026-09-21)

[...]

Issue #403, opened 2026-07-04
export run: relative --output resolves against the PROJECT dir, and the result echoes the unresolved path — resolve against the invoker's cwd and report the absolute artifact path


qinhaihong-red:
## Summary

`gda export run --output <relative-path>` silently writes the artifact relative to the **project** directory, not the invoker's cwd — and the success result echoes the **unresolved** input string as `output_path`, so a programmatic consumer cannot locate the artifact at all. Resolve `--output` against the invoker's cwd client-side (standard CLI convention) and always report the resolved absolute path.

## Current behavior

Reproduced on gda 0.2.1 (Godot 4.6, macOS), invoking from an unrelated directory `A`:

```
A$ gda export run --preset macOS --output ./PA-test.zip --project <game> --json
{"preset": "macOS", "platform": "macOS", "mode": "release",
 "output_path": "PA-test.zip", "warnings": []}
```

- `A/PA-test.zip` does not exist; the artifact landed at `<game>/PA-test.zip` (the user's real-world hit: `--output ./PA.app` littered `PA.app/` into the project working tree).
- Mechanics: the export channel is the one native `--export-<mode>` invocation, run through the `Headless launch` primitive with the **export-only cwd** = the project dir (ADR-0031); `--output` is passed through verbatim, so the engine resolves a relative path against *its* cwd, the project root.
- The result's `output_path` is the raw input string — with a relative input it describes a location relative to whoever reads it, which is wrong for every reader except the engine.

## Assessment

The project-relative frame alone would be *defensible* (a preset's configured `export_path` is project-relative by Godot convention, and `--output` overrides that value), but it is (a) documented nowhere — help/schema say only "write the artifact here instead" — and (b) compounded by the unresolved echo, which breaks structured consumption regardless of which frame is chosen: an agent checking `output_path` for the artifact looks in its own cwd and finds nothing.

## Desired behavior

- `--output` given as a relative path is resolved against the **invoker's cwd** at param-build time (client-side abspath before the runner sees it) — the universal CLI convention; project-relative intent stays expressible via an absolute path or invoking from the project dir.
- A preset-configured `export_path` (no `--output`) keeps its Godot project-relative convention.
- In **both** cases the success result's `output_path` reports the **resolved absolute path** of the written artifact, so the result is self-describing for any consumer.
- `--schema` / result-model descriptions state the resolution rule (ADR-0004 chain).

## Acceptance criteria

[...]

PR #1002, opened 2026-09-21
docs(gda): output_path is the destination as the caller named it, in every text that describes it (#1001)

qinhaihong-red:
## Summary

`gda export run` publishes `output_path` as the destination the caller named (#403). Six texts said otherwise — the published field description (what `gda schema` emits), the result's class docstring, the command catalog's `export run` paragraph, `export run --help` ("the resolved artifact path"), and the catalog's and `export smoke --help`'s "the absolute `output_path` from `export run` passes straight through". An agent that reads the schema and receives `res://build/game.x86_64` was told it holds a filesystem path it does not hold. The six texts now state the rule the code implements: a `res://` or `user://` address as given, from `--output` or the preset; a filesystem `--output` with `~` expanded and made absolute against the invoker's current working directory; or a filesystem `export_path` from the preset, resolved against the project. No published value changes.

Closes #1001

## What changed

- `src/gda/commands/export.py`
  - `ExportRunResult.output_path`'s field description: "The destination as the caller named it: a `res://` or `user://` address as given, from `--output` or the preset; a filesystem `--output` with `~` expanded and made absolute against the invoker's current working directory; or a filesystem `export_path` from the preset, resolved against the project."
  - `ExportRunResult`'s class docstring says the same, word for word (it is published as the output schema's own description — the second schema hunk).
  - `export run`'s command docstring (the `--help` text and the command description in the schema): "the destination as the caller named it — a `res://` or `user://` address as given, a filesystem path made absolute".
  - `export smoke`'s command docstring: "a filesystem `output_path` from `export run` passes straight through; a `res://` one names the artifact only inside its project — give the file's path".
- `docs/command-catalog.md`
  - The `export run` paragraph: the same sentence.
  - The `export smoke` section: "a filesystem `output_path` from `export run` passes straight through (a `res://` one names the artifact only inside its project; give `smoke` the file's path)".
- `tests/export/test_export_run_commands.py`
  - The schema test that pinned "resolved absolute path" in the field description now pins "as the caller named it".

Not changed, each with its reason: the `gda skill` `export` row's recipe "feed `smoke` `run`'s `output_path`" (and the same recipe in `export smoke`'s docstring and `errors.py`'s `export_artifact_not_found` text) holds for the configured `export_path` and a filesystem `--output`, the common cases; the two test comments that call the configured channel's result "the resolved absolute path" describe that channel, where it is true; `docs/adr/0042-artifact-smoke-runs-the-exported-build.md` line 57 ("the absolute `output_path` reported by `export run` therefore passes through directly") is a point-in-time decision record — its premise was already inexact for a `res://` `--output` when it was written (#403 predates it) — and ADRs are not rewritten; the living texts that describe the surface are the schema, the help and the catalog, all corrected here.

How the inventory was built, and where it went wrong: the issue's three texts came from a search for the literal phrase "resolved absolute"; the two `--help` texts phrase the same claim as "the resolved artifact path" and "the absolute `output_path` … passes straight through", and the ADR sentence differently again. A sweep by MEANING at the head (every `output_path` mention read, in `docs/`, `src/gda/`, README and SKILL) finds no other home of the claim.

## Schema / help

`gda schema` pretty-printed and diffed against the base `63b4f96cc`: **4 hunks**, all declared in issue #1001's second amendment — the field description, the result docstring, and the `export run` and `export smoke` command descriptions.

```diff
-  … reported ``output_path`` is the resolved artifact path, and missing output …
+  …             "description": "Export a named preset to a destination and report the artifact.\n\nUnlike every other comma … …
-  …                 "description": "The result of ``gda export run``: the artifact that was produced (issue #121).\n\nEchoe … …
+  …                 "description": "The result of ``gda export run``: the artifact that was produced (issue #121).\n\nEchoe … …
-                        "description": "The resolved absolute path the export artifact was written to.",
+                        "description": "The destination as the caller named it: a `res://` or `user://` address as given, from `--output` or the preset; a filesystem `--output` with `~` expanded and made absolute against the invoker's current working directory; or a filesystem `export_path` from the preset, resolved against the project.",
-  … ``output_path`` from ``export run`` passes straight through. …
+  … ``output_path`` from ``export run`` passes straight through; a …
```

Every command's `--help` against the base: **2 hunks**, `export run` and `export smoke`; the other 93 pages are byte-identical.

```diff
- reported ``output_path`` is the resolved artifact path, and missing output
- parent directories are created and reported in ``created_dirs`` (#402/#403).
+ reported ``output_path`` is the destination as the caller named it — a
+ ``res://`` or ``user://`` address as given, a filesystem path made absolute
+ — and missing output parent directories are created and reported in
+ ``created_dirs`` (#402/#403).
- relative artifact path resolves against the current directory, so the absolute
- ``output_path`` from ``export run`` passes straight through.
+ relative artifact path resolves against the current directory, so a
+ filesystem ``output_path`` from ``export run`` passes straight through; a
+ ``res://`` one names the artifact only inside its project — give the file's
+ path.
```

PR CI runs ruff, pyright and the non-e2e tier, which carries the schema test's new pin and the help test's existing pins. The export e2e was run locally under the machine-wide lock; the schema and help captures are local evidence.

## Validation

Host: macOS (Darwin 27.0.0), Godot 4.6.3-stable, shared with other agents; `uptime` load averages 3.8–19.1 (sibling agents' load during the last runs); every line at the final head `67aaf7e98c2c788ba31efc18d9e46418aee6bf55`.

| Gate | Command | Result |
| --- | --- | --- |
| Format | `uv run --frozen ruff format --check .` | `525 files already formatted` |
| Unit | `uv run --frozen pytest -m "not e2e" -n 4 --dist loadgroup -q` | `2963 passed, 3 skipped, 1 warning in 74.97s` |
| e2e export | `pytest -m e2e tests/export -q -rs` (under the lock) | `36 passed, 1 skipped, 171 deselected in 167.90s` |
| Harness guard | `scripts/harness_version_guard.py origin/feat/gda-dogfooding-dev HEAD` | `harness version transition is valid` |
| Schema | `gda schema` vs base, pretty-printed | 4 hunks, pasted above |
| Help | every `--help` vs base | 2 hunks, pasted above |

## Deviations from the issue

- The declared delta grew twice, each time by a dated amendment on the issue: the result docstring is published too (two schema hunks), and the two `--help` texts carried the same claim (four schema hunks, two help hunks). The issue's first draft required byte-identical help; leaving those two texts would have fixed the defect in half its homes, so the second amendment declares them instead.
- AC1's first wording missed a preset `export_path` spelled `res://…` or `user://…`, which is published as given too (review round 1 measured it with a real `--mode pack` export); the second amendment restates the rule, and the field description and result docstring carry it word for word.

- Text only: nothing changes what the command publishes, and the e2e that asserts `output_path` spellings is unchanged.
- PR #999 (#997) also edits `src/gda/commands/export.py`, on other lines; the two squash cleanly in either order (both fork from `63b4f96cc`).


qinhaihong-red:
## Review round 1 handled — head `67aaf7e98c2c788ba31efc18d9e46418aee6bf55`, READY

Fresh adversarial review at `b4fb1c586` (13 real-engine calls, every input kind, both schema hunks rebuilt, 95 help pages compared): no P1, two P2, three P3.

| Finding | Decision | Change |
| --- | --- | --- |
| [P2] The configured-channel clause was false: a preset `export_path` spelled `res://…` or `user://…` is published as given too (measured: `output_path: 'res://build/res.pck'`, `'user://cfg.pck'`); the issue's own AC1 carried the gap | Adopted; issue #1001's second dated amendment restates AC1 | `67aaf7e98`: "a `res://` or `user://` address as given, from `--output` or the preset; a filesystem `--output` with `~` expanded and made absolute against the invoker's current working directory; or a filesystem `export_path` from the preset, resolved against the project" — word-identical in the field description and the result docstring, in prose in the catalog |
| [P2] `export run --help` ("the resolved artifact path") and `export smoke --help` ("the absolute `output_path` … passes straight through") still carried the claim, the channel the skill points agents at; AC2's byte-identical help blocked fixing them | Adopted as a declared delta rather than a follow-up: leaving them would fix the defect in half its homes. The same amendment declares four schema hunks and two help hunks | `67aaf7e98`: both command docstrings corrected; the hunks are pasted in the body |
| [P3] `~` expansion omitted from the `--output` clause | Adopted | "with `~` expanded and" |
| [P3] Ragged wraps at the two edit sites | Adopted | Reflowed |
| [P3] "the invoker's cwd" against the surface's "the invoker's current working directory" | Adopted | The surface's phrase everywhere |

Left as the review agreed: the `gda skill` `export` row's smoke recipe (true for the common cases), the test comments and name about the configured channel (true of their own arm).

Gates at `67aaf7e98`: ruff / format clean, pyright 0, non-e2e tier `2963 passed, 3 skipped`, export e2e `36 passed, 1 skipped` under the lock, harness guard valid; `gda schema` vs base 4 hunks and `--help` 2 hunks, exactly the declared ones, the other 93 pages byte-identical; the schema test's pin reads RED when the description is reverted. PR CI 7 pass / 4 skip; `Closes #1001` in the body; author and committer `haihong.qin` on both commits; no attribution anywhere. One fresh adversarial round plus the orchestrator's own verification (each new clause derived from `normalize_export_output_path`, `_absolute_filesystem_path` and `_resolve_configured_export_path`; the hunk counts rebuilt).

qinhaihong-red:

Reviewed exact range `63b4f96cc9b1fc236fa2d6f3bfa6a46de1a5fd19...67aaf7e98c2c788ba31efc18d9e46418aee6bf55`. I did not read or use existing PR reviews or review comments.

### 1. Artifact review — Changes required

#### Required changes

1. **The replacement definition still says something that the detailed rule disproves.** The result model and catalog call `output_path` “the destination as the caller named it” ([model](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/src/gda/commands/export.py#L540-L578), [catalog](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/docs/command-catalog.md#L1326-L1337)). That umbrella phrase is true only for virtual addresses. A filesystem `--output` is changed by `~` expansion and absolutization, and a preset destination is configured by the preset rather than named by the caller. The case list immediately after the phrase is accurate, but it does not make the lead definition accurate. This matters because the field description is the machine-readable contract that the PR is intended to correct. Replace the umbrella phrase with a neutral term such as “the effective export destination,” then state the three cases. Issue #1001 must be amended because its current AC requires the inaccurate phrase.

2. **The inventory leaves agent-facing and architecture guidance with the old unconditional handoff.** The bundled `gda skill` still tells an agent to “feed [smoke] `run`'s `output_path`” without a filesystem-only condition ([SKILL.md:193](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/src/gda/skill/SKILL.md#L193)). `export smoke --help` also opens with the unconditional instruction and only contradicts it later on the same page ([export.py:1885-1904](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/src/gda/commands/export.py#L1885-L1904)). The accepted ADR still says the absolute result passes through and requires feeding the result directly to smoke ([ADR-0042:56-61](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/docs/adr/0042-artifact-smoke-runs-the-exported-build.md#L56-L61), [validation](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/docs/adr/0042-artifact-smoke-runs-the-exported-build.md#L149)). At this head, `gda export smoke 'res://out.pck' --json` normalizes the operand to `<cwd>/res:/out.pck` and returns `export_artifact_not_found`. The issue's exception is also internally stale: it says the skill recipe holds for “the configured `export_path`,” while the second amendment establishes that a configured path can itself be `res://` or `user://`. Expand the declared delta, make the skill and the opening help instruction conditional, and add a dated ADR amendment rather than rewriting its historical text.

#### Terminology and prose assessment

I used `CONTEXT.md`'s **gda skill**, **Export artifact**, **Artifact smoke**, and **Access-path message** definitions, plus ADR-0004, ADR-0015, and ADR-0042. I applied ASD-STE100 as the writing reference and inspected every changed prose passage plus every `output_path`/smoke-handoff mention under `docs/`, `src/gda/`, `README.md`, the bundled skill, and relevant tests. The two findings above are substantive accuracy and consistency defects. I found no other material terminology or prose issue.

### 2. Domain modular architecture review — Changes required

Observed ownership is otherwise sound: `gda.commands.export` owns normalization and execution; `ExportRunResult` owns the output model; schema and MCP descriptions derive from that model. CLI help, the command catalog, and the bundled skill are hand-maintained guidance over the same operation surface. `CONTEXT.md` explicitly defines CLI, Agent Skill, and MCP as three complementary access paths.

**[A1] The handoff rule has no complete semantic projection across those access paths.** This PR updates the model/schema, two help descriptions, and the catalog, but leaves the bundled skill and accepted ADR inconsistent. The existing skill synchronization gate checks command tokens, not this relationship ([test_skill_surface_sync.py:1-14](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/tests/meta/test_skill_surface_sync.py#L1-L14)). This is a completeness and DRY failure in knowledge ownership, not a reason to add a new module or generalized documentation framework. The smallest durable fix is to update the remaining projections and add one focused semantic guard: filesystem results may pass directly; virtual results require the artifact's filesystem path.

Load-bearing claim `DDMA-1002-1` — “all three access paths describe the handoff consistently” — has claim evidence state `open` (the repository defines no stronger evidence-state vocabulary). Current evidence is the skill/help/ADR contradiction and the executable `res://` counterexample. The disconfirming condition for the open risk is a regression test that mutates either side of the filesystem/virtual conditional and fails. No module-boundary or dependency change is otherwise needed.

### 3. Two-axis code review

#### Standards — 1 finding

- **[P2] Synchronize the accepted ADR with the corrected `output_path` contract.** The changed catalog and help correctly limit direct smoke handoff to filesystem results, while ADR-0042 still states that the absolute result passes through directly and its product contract does not limit that statement. This violates the repository rule that related authoritative artifacts must be checked and synchronized when behavior or design facts change. Preserve the point-in-time body, but add an explicit amendment.

No other documented-standard violation or actionable Fowler smell was found. Tool-enforced lint/format matters were excluded from this axis.

#### Spec — pass, 0 findings

Against issue #1001 **as currently amended**, the diff implements the declared four schema-description changes and two help changes, keeps all other help pages byte-identical, and does not change runtime behavior. The unchanged skill recipe was not counted as a Spec omission because the issue explicitly requires leaving it unchanged; that exception is instead challenged by the Artifact and Architecture reviews above.

**Axis summary:** Standards — 1 finding, worst P2; Spec — 0 findings, pass.

### Validation and limits

- `93 passed` across focused export-run, export-smoke, mutation, and skill-synchronization tests.
- Ruff lint passed; both changed Python files are formatted.
- Independent comparison found exactly four schema hunks and two changed help pages out of 95; the other 93 help pages were byte-identical.
- Runtime contract probe: `gda export smoke 'res://out.pck' --json` returned `export_artifact_not_found` for `<cwd>/res:/out.pck`.
- Remote CI is green for lint, types, package/unit tests, co-install, and scope guards. The Godot e2e job is skipped. I did not repeat the full non-e2e suite or a real Godot export locally; the review relies on focused tests, the checked CI result, and the deterministic projectless smoke probe.

qinhaihong-red:
## Backtrace review — stop and reframe before merging

Reviewed at base `63b4f96cc9b1fc236fa2d6f3bfa6a46de1a5fd19`, head `67aaf7e98c2c788ba31efc18d9e46418aee6bf55`.

### Conclusion

The underlying problem is real, but this PR is a symptom-level solution to the wrong requirement. **Do not merge #1002 as-is.** Its short-term value is that it stops the schema from making one false statement; its larger effect is to ratify an unresolved representation accident as a durable product contract.

The real user outcome is not “describe every spelling that the current implementation happens to echo.” It is: **an `export run` result must let a programmatic consumer locate the artifact that was produced, and that result must compose with `export smoke`.**

### Backtrace

1. [Issue #403](https://github.com/aigengame/godot-agent/issues/403) came from a reproduced dogfooding failure: the result echoed an unlocatable relative string. Its desired behavior and AC require `output_path` to be the resolved absolute filesystem path in both override and preset cases, so the result is self-describing.
2. [PR #411](https://github.com/aigengame/godot-agent/pull/411) implemented that fix but also introduced an unrequested exception: any value containing `://` passes through unchanged ([normalizer](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/src/gda/commands/export.py#L129-L140), [preset resolution](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/src/gda/commands/export.py#L890-L900)). #403 did not require caller-spelling preservation, and #411 added no virtual-output acceptance test. The implementation comment nevertheless continued to claim that the result is self-describing.
3. [Issue #979](https://github.com/aigengame/godot-agent/issues/979) and ADR-0042 then built the end-to-end `export run → output_path → export smoke` flow on the original absolute-path assumption ([ADR contract](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/docs/adr/0042-artifact-smoke-runs-the-exported-build.md#L53-L62), [required validation](https://github.com/aigengame/godot-agent/blob/67aaf7e98c2c788ba31efc18d9e46418aee6bf55/docs/adr/0042-artifact-smoke-runs-the-exported-build.md#L147-L150)). A projectless smoke command cannot resolve `res://` or `user://`.
4. [Issue #1001](https://github.com/aigengame/godot-agent/issues/1001) and this PR respond by changing public prose to match the exception, rather than deciding which representation the result must own. Review expanded the patch from three texts and one schema hunk to six published texts, four schema hunks, and two help pages; the bundled skill and accepted ADR still expose further contradictions.

This is **confirmed symptom-patch accumulation**: each edit is locally small, but each one preserves the unresolved boundary and produces another synchronization obligation.

### Product-promise expansion

This PR would turn a technically possible input form into an explicit support policy. Schema, CLI help, the command catalog, Agent Skill guidance, and MCP tools would need to preserve and explain a matrix across:

- filesystem vs `res://` vs `user://`;
- CLI override vs preset configuration;
- invoker cwd vs project root;
- `~` expansion vs literal `~`;
- values that can be handed directly to `export smoke` vs values that require caller repair.

There is evidence that Godot accepts at least a measured `res://` pack destination. There is **no recorded user requirement** that the public result preserve that spelling. Technical possibility does not establish a product support obligation. Conversely, this behavior has existed in released versions since the #411 implementation, so removing or changing it now is a compatibility decision for the product owner, not a cleanup to make implicitly.

NFR/platformization status: **support-scope decision unresolved**. “Preserve every virtual output spelling as a stable public result” is the candidate obligation; its causal relationship to the present rework is confirmed, but its product value is not. Human direction is required before dependent patches continue.

### Entropy verdict

**Remove this solution; preserve the underlying need.** The current PR has no sufficient standalone value because it makes the public contract more precise while leaving the core consumer outcome worse than #403 and ADR-0042 require.

Essential complexity to retain:

- one owner for export-destination resolution;
- a truthful, self-describing result contract;
- a real `export run → export smoke` verification path;
- a typed refusal for unsupported destination forms.

Complexity to avoid:

- one field that alternates between an input address and a locatable artifact path;
- caller-side repair rules;
- semantic synchronization across every documentation surface for an unconfirmed support matrix.

### Recommended recovery

1. Pause or close #1002 in its current form.
2. Make an explicit product decision about virtual export destinations. The recommended invariant is: **`output_path` names the actual artifact as an absolute filesystem path.**
3. Keep `res://` as an input syntax only if useful, but translate its result through the existing project/path authority before publication.
4. Treat `user://` support as unverified. Either resolve it with real end-to-end evidence or reject it with a typed error; do not create a permanent promise by default.
5. If released compatibility prevents changing `output_path` directly, add an unambiguous `artifact_path` and deprecate the mixed-semantics field. Preserve the caller's original address in a separate field only if an independent requirement establishes value for it.
6. Reconcile #403, #979/ADR-0042, schema, help, catalog, and skill only after that decision. Completion must include a real export whose published artifact path is passed unchanged to `export smoke`.

The decision owner is the maintainer/product owner. Until that decision is made, further wording amendments would deepen the current patch cycle rather than resolve it.

qinhaihong-red:
Closed without merging, on the maintainer's backtrace review: the texts this PR rewrote are #403's requirement (`output_path` is the absolute path of the artifact the engine wrote, for both destination channels), and the behaviour they contradict — the `://` pass-through PR #411 added without a requirement — is the drift. Documenting the pass-through would have ratified it as a contract and grown a support matrix across scheme, channel, anchor, `~` and what `export smoke` can take. The fix is to the behaviour, on top of PR #999's one `res://` location; the texts then stay as they are. Issue #1001 is closed as superseded; the replacement slice is filed separately.

